PR #3, opened 2025-05-12
Add pre-commit and update .gitignore.

andrewkvuong:
Adds pre-commit and updated gitignore. Mostly copied from vLLM, removed pre-commit checks that current depend on specific tools defined in the vLLM repo.

Python changes are from running lint + precommit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,5 +51,11 @@ pytest -v ./tests/ragged_paged_attention_test.py
 ## How to format the code?
 
 ```
-yapf -ir .
+pip install pre-commit
+
+# Linting, formatting and static type checking
+pre-commit install --hook-type pre-commit --hook-type commit-msg
+
+# You can manually run pre-commit with
+pre-commit run --all-files
 ```
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,8 @@ setup(
     name="tpu_commons",
     version="0.1.0",
     description="",
-    long_description=open("README.md").read() if hasattr(open("README.md"), "read") else "",
+    long_description=open("README.md").read() if hasattr(
+        open("README.md"), "read") else "",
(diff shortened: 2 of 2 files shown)


yarongmu-google:
LGTM - @bythew3i can you please review the kernel part?

README.md
```diff
@@ -36,16 +37,17 @@ pip install -r requirements.txt
...
 ```
 
 Run the test:
+
 ```
 pytest -v ./tests/ragged_paged_attention_test.py
 ```
 
-
 ## How to format the code?
```
bythew3i, line 51: It seems we will auto-format the code when pre-commit...can we either remove or update this section "How to format the code?"?

tests/ragged_paged_attention_test.py
```diff
@@ -1,16 +1,13 @@
 import random
 
-from absl.testing import absltest
-from absl.testing import parameterized
 import jax
-from jax._src import test_util as jtu
 import jax.numpy as jnp
+from absl.testing import absltest, parameterized
+from jax._src import test_util as jtu
```
bythew3i, line 6: is this generated by formatter? How come it is not sorted alphabetically?

README.md
andrewkvuong: Updated section on formatting. 

Should be same as vLLM https://docs.vllm.ai/en/stable/contributing/overview.html.

tests/ragged_paged_attention_test.py
andrewkvuong: It sorts by default `import` first then `from` then alphabetically within. 

https://pycqa.github.io/isort/

bythew3i: Ah it starts with "from". NVM

PR #20, opened 2025-06-04
The first version of model implementation with top-level structure desi…

bzgoogle:
# Description

The first version of model implementation with top-level structure design. 
These include blocks to build up the transformer block, i.e.  Attention, MoE, KVCache, FFW, Embedder, etc. 
It aims to have improved control of sharding strategy and better code structures in Prefill/Generate modes

# Tests

Unit tests wrote for each block to verify it generates expected output

# Checklist

### DIFF
--- a/tpu_commons/models/jax/common/layers.py
+++ b/tpu_commons/models/jax/common/layers.py
@@ -25,61 +25,6 @@ class FlaxUtils:
 
 modeling_flax_utils = FlaxUtils()
 
-@dataclasses.dataclass
-class ParamFactory:
-    """A factory for creating nnx.Param objects with shared RNGs and initializers.
-
-    This class simplifies the creation of parameters by holding common
-    configuration like the RNG stream and the weight initialization function.
-
-    Attributes:
-        rngs: An `nnx.Rngs` object to provide RNG streams for parameter initialization.
-        initializer: A callable (e.g., a kernel initializer from JAX) used to
-            generate parameter data.
-    """
-    rngs: nnx.Rngs
-    initializer: Initializer
-
-    def create_kernel_init(self,
-                           shape: tuple[int, ...],
-                           sharding: NamedSharding,
-                           dtype: Any = jnp.float32) -> nnx.Param:
-        """Creates an nnx.Param using the factory's RNG stream and initializer.
-
-        Args:
-            shape: The shape of the parameter tensor to create.
-            sharding: The `NamedSharding` object that specifies how the parameter
-                should be distributed across devices.
-            dtype: The data type of the parameter.
-
-        Returns:
-            An `nnx.Param` instance containing the initialized data and sharding info.
-        """
-        param_data = self.initializer(self.rngs.params(), shape, dtype)
-        return nnx.Param(param_data, sharding=sharding)
-
-from flax import nnx
-from jax.sharding import Mesh, NamedSharding
-from jax.sharding import PartitionSpec as P
-from jaxtyping import Float, Int
-
-from tpu_commons.models.jax.common.base import Config, ParamFactory
-from tpu_commons.models.jax.common.sharding import ShardingConfig
-
-
-# A dummy for modeling_flax_utils which might contain activation functions
-class FlaxUtils:
-    """A dummy class to namespace activation functions, mimicking external utilities."""
-    ACT2FN = {
-        'silu': nnx.silu,
-        'gelu': nnx.gelu,
-        'relu': nnx.relu,
-    }
-
-
-modeling_flax_utils = FlaxUtils()
-
-
 @dataclass
 class RuntimeParams:
     """A container for runtime parameters needed by neural network blocks.


models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
+@dataclass
+class TransformerConfig:
```
gpolovets1: Test comment.

models/attention/attention.py
```diff
@@ -0,0 +1,75 @@
...
+class AttentionConfig(Config):
+    d_model: int
+    num_heads: int
+    num_kv_heads: int   
+    qk_dim: int
+    v_dim int
+    #...
+
+    @classmethod
+    def from_cfg(cls, flags_cfg: dict):
```
gpolovets1: Probably do not need to redefine here since inherited from Config?

models/sharding/sharding.py
```diff
@@ -0,0 +1,149 @@
+from constants import LOGICAL_MESH_AXIS_NAME
+@dataclass
+class Sharding_Strategy:
```
gpolovets1: Maybe sharding_strategy = prefill_strategy + decode_strategy?

models/sharding/sharding.py
```diff
@@ -0,0 +1,149 @@
...
+              ...
+          if shading_strategy.sequence_parallelism > 0:
+              ...
+          generate_sharding_cfg = OpSpecificShardingConfig(...)
+
+        self.sharding_cfg = ShardingConfig(
+          prefill=prefill_sharding_cfg,
+          generate=generate_sharding_cfg)
+        return self.sharding_cfg
+
```
gpolovets1: We may also need to add some validation methods here (e.g. each shard needs to multiply to the total number of devices).
Maybe we can also print the full sharding or write it to a (local or gcs) experiment folder.

models/moe/moe.py
```diff
@@ -0,0 +1,88 @@
...
+    def __call__(self,):
+        #TODO implement the logics here
+        ...
+    def get_cfg(self) -> MoEConfig: 
+        ...
+    def sharding(op_mode):
+        ...
+
+@dataclass
+class RoutingConfig:
```
gpolovets1: Do you mean `class Router`? This seems to be more than just a config.

models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
...
+        provided_params = set(flags_cfg.keys())
+        missing_params  = required_params - provided_params
+
+        if missing_params:
+            ...
+
+        transformer_flags = {k: flags_cfg[k] for k in required_params}
+        return cls(**transformer_flags)
+
+    def make(self, runtime_param: Optional[layer.RuntimeParams] = None) -> TransformerBlock:
```
gpolovets1: I think maybe it's a little confusing for a Config class to create the actual model? I would imagine that make() should combine all of the component configs, runtime args, and default args. And then a Transformer class can use this config in its setup() method.

models/quantization/quantization.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Quantization:
+    cfg: QuantizationConfig
+
+    def __init__(self):
+        ...
+    def get_quantization_config(self) -> QuantizationConfig:
+        ...
+    def create_weight(self):
+        ...
```
gpolovets1: Wonder if it would be feasible to have a quantize(self, model: Transformer) method that applies quantization on the pytree. We can specify which layers to apply based on the pytree.
Then in the driver code code we could do something like model = quantizer.quantize(model)

models/sharding/sharding.py
```diff
@@ -0,0 +1,149 @@
...
+    norm_scale: LOGICAL_MESH_AXIS_NAME | None = None
+    ...
+
+@dataclasses.dataclass
+class ShardingConfig:
+    prefill: OpSpecificShardingConfig
+    generate: OpSpecificShardingConfig
+
+
+class Sharding:
```
gpolovets1: Should we also consider adding a ShardingInfo class? https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/core/compilation/partitioning_stats.py;rcl=756952563;l=43

We can store it as a part of ArrayInfo.

models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
+@dataclass
```
gpolovets1: We can also consider creating an ArrayInfo class to store the Transformer tensor information: https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/experimental/minmax/transformer.py;l=170

Or jax.ShapeStruct: https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/experimental/minmax/transformer.py;l=135

It would also be great to store the variable name in this metadata (e.g. information stored in tree_utils.tree_flatten_with_path).

Alternatively, we can place this in a utils.py.

gpolovets1: We can also consider adding metrics to each tensor (e.g. latency, memory, flops, etc) to help with rooflining.
Some examples: 
  - hex-llm: https://source.corp.google.com/h/vertex-model-garden/hex-llm/+/main:hex_llm/engine/llm_engine.py;l=747;bpv=0;bpt=0
  - Gemax tracked metrics: https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/experimental/minmax/train.py;l=380?q=count_activated_params&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental
  - MoEMetrics in Gemax: https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/core/models/gemini2/models/routed_layers.py;l=39;rcl=733860666

models/layers.py
```diff
@@ -0,0 +1,101 @@
+@dataclass
```
gpolovets1: If you anticipate a common interface, may want to create an abstract Layer class that all others implement (including attention and moe).

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
+# Input flags are stored in below configs
```
gpolovets1: One suggestion is that you may not need to create a new model.py per model.. instead you could have a builder/factory class to assemble models. Everything can inherit from Transformer/TransformerConfig but specialized models may have their own version of attention, MoE, etc. So you can have a builder class instantiate the corresponding specialized configs based on the model type.
Instead, maybe you can add a config/ directory with predefined configs for each type of model.

models/attention/attention.py
```diff
@@ -0,0 +1,75 @@
...
+        #q, k, v = kernel * x
+        return (q,k,v)
+    def sharding(op_mode):
+        ...
+    def __call__(
+        self, 
+        x,
+        attn_cache
+    ):
+        (self.q, self.k, self.v) = self.input_projection(x, self.kernel)
```
gpolovets1: Maybe we could use the variable naming convention in our code?
E.g. self.q_btkgd, self.k_bskd,... =

models/attention/attention.py
```diff
@@ -0,0 +1,75 @@
...
+    def setup(self):
+        ...
+    def input_projection(self, x):
+        #q, k, v = kernel * x
+        return (q,k,v)
+    def sharding(op_mode):
+        ...
+    def __call__(
+        self, 
+        x,
```
gpolovets1: Should we consider jaxtyping? https://docs.kidger.site/jaxtyping/#example

models/attention/attention.py
```diff
@@ -0,0 +1,75 @@
...
+        attn_cache
+    ):
+        (self.q, self.k, self.v) = self.input_projection(x, self.kernel)
+        self.sharding()
+        attn_weight = jnp.einsum("btkgd,bskd->bkgts", self.q, self.k)
+        attn_weight = self.sharding(attn_weight)
+        logits = einsum("bkgts,bskd->btkgd", attn_weights, self.v)
+        out = jnp.einsum("btkgd,btkgdh->bth", logits, self.kernel)
+        self.kv_cache.update(self.k, self.v, ...)
```
gpolovets1: Should we use kv_cache.py logic here?

models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
...
+    it uses make() to create the live module from this config
+    """
+    embeddingsConfig: EmbedderConfig 
+    attention: AttentionConfig
+    moe: MoEConfig = None
+    ffw: FFWConfig = None
+
+    num_groups: int
+    routing: bool = False
+    ...
```
gpolovets1: Maybe also mode = str: (prefill/decode) since it influences the ops in the Transformer.

models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
+@dataclass
+class TransformerConfig(Config):
+    """
+    light weighted transformer config, which includes config for all sub-modules
+    it uses make() to create the live module from this config
```
jrplatin: Maybe you'll get to this but one pain point of MaxText I want to improve is the lack of docstrings (I assume they're missing here since this is early-stage code)

models/attention/attention.py
jrplatin: +1, having the shaping embedded in the var names is such a major help

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
...
+#
+# Each Block(attn, mlp etc) has Config Class and Live Class
+# There're 2 ways to initialize a Config:
+# 1. manual specify as Config(d_model=, num_layer=, ..)
+# 2. auto assignment from a config as Config.from_cfg(cfg)
+
+
+# The foundation Model/ModelConfig class is to-be-implemented
+
+class LlamaConfig(ModelConfig):
```
jrplatin: Just an idea but might make sense to have Llama/DeepSeek/Mixtral all inherit from a common `MoEModelConfig`? Depends on how the `ModelConfig` itself is defined though

models/TransformerBlock.py
```diff
@@ -0,0 +1,94 @@
+@dataclass
+class TransformerConfig(Config):
```
jrplatin: Kind-of a nit, but I'd change this to `TransformerBlockConfig`, since `TransformerCofnig` to me sounds like an entire _model_

models/llama4/model.py
jrplatin: Yeah this is interesting discussion I think -- on one hand, I really like how HexLLM does it where each model is very cleanly defined and almost everything is in one-or-two files, but then you end up duplicating a lot of code (or doing weird things like using the DeepSeek MoE block in the Llama4 model).  On the other hand, while it's a bit more scattered and hard-to-read, I think the "plug-in" style approach is probably better SWE practice, especially for modularization.

models/sharding/sharding.py
jrplatin: What do you mean? Not sure I follow

gpolovets1: We may have different sharding strategies during Prefill and Decode. e.g. Prefill has bs=1 but seq > 1 while Decode has bs > 1 and seq = 1 so we cannot use data parallel for both phases.

models/TransformerBlock.py
bzgoogle: agree, this is just for a demo purpose for the first version. I will workout a more clear docstrings for the major blocks/functions

models/sharding/sharding.py
bzgoogle: Yes, i think we should have 2 sets. of strategy for prefill and decode. Then the downstream blocks will consume it and create sub-class for prefill & decode whenever necessary.

models/moe/moe.py
bzgoogle: This is for config, the related Router is declared below. I used to put TopkRouter below, but corrected it to Router

models/sharding/sharding.py
bzgoogle: That's a good idea for visualizing & debugging the sharding performance. As it won't impact the functionality, i will add it in later commits/PRs

models/sharding/sharding.py
bzgoogle: yes, i put a placeholder, i.e. validate_sharding_strategy, which validates the parallelism fitting into the devices

models/layers.py
bzgoogle: Done

models/TransformerBlock.py
bzgoogle: Per offline discussion, I removed the make() function in config class, so that it only processes and holds parameters. And the live class is initialized with the config class

models/quantization/quantization.py
mitalisi: We will likely not use simple quantized tensors for model quantization unless we only use weight-only quantization. Qwix does something like quantizing the whole model and specifying which layers to apply quant to. We should think of this more as configuring the library or another algorithm.

models/sharding/sharding.py
mitalisi: It is OK to add the class with comments on intention and TBD.

models/sharding/sharding.py
```diff
@@ -1,83 +1,102 @@
...
+         if op_mode == OPERATION_MODE.PREFILL:
+            return self.logical_mesh_axis_name_prefill
+        elif op_mode == OPERATION_MODE.DECODE:
+            return self.logical_mesh_axis_name_decode      
 
 @dataclass
-class AttentionShardingConfig:
-    # demo only, suffix needs to be verified
-   
+class ShardingConfig:
```
gpolovets1: We can play around with this a bit more, but wondering if we should divide this up into sections - e.g. attention_sharding: AttentionShardding()
...
That way in the model we can access the sharding in respective layers using something like self.sharding.activation_sharding.activation_attention_bsd...

models/sharding/sharding.py
```diff
@@ -99,49 +118,76 @@ class Sharding:
...
-    logical_mesh: Mesh
+    LOGICAL_MESH_AXIS_NAME: LOGICAL_MESH_AXIS_NAME
 
     def __init__(self, flags_cfg):
       self.sharding_strategy = Sharding_Strategy(flags_cfg)
-      self.logical_mesh = get_logical_mesh()
-      ...
+    def __post__init__(self):
+      self.mesh(
+        prefill_mesh=self.build_mesh_per_op(self.sharding_strategy.prefill_Sharding_Strategy),
```
gpolovets1: I'm not sure if this messes things up having two meshes simultaneously.. maybe we just want two sets of logical rules?

models/layers.py
```diff
@@ -32,47 +21,110 @@ def from_cfg(cls, flags_cfg: dict):
...
+    ...
+
+
+class FFWConfig(Config):
+  d_model: int
+  hidden_size: int
+  act: str
+  dtype: Any = jnp.float32
 
 class FFW(nnx.Module):
```
gpolovets1: By the way, do you know what this type of MLP is called? Traditionally there is no gating term, so at first I was a little bit confused when saw this in MaxText. Wondering if we should somehow clarify in the name

models/attention/attention.py
bzgoogle: Yes, a base class of Config is added

models/attention/attention.py
bzgoogle: Yes, this is one of the design proposals. And now the variable for tensors(kernel, weights, activations) are named with its shape as a suffix

models/layers.py
bzgoogle: it a gated MLP, or gMLP.  From other stack, including MaxText, people may use MLP, FFW. It's a good idea to clarify the name some where, i.e. docstrings for the class

models/sharding/sharding.py
bzgoogle: agree, there could be much overhead with 2 mesh(logic topology), with a vague potential gains. I updated with using a single mesh but different logical rules for prefill and decode. We also keep the flexibility that each layer/module could take external mesh for sharding, in case any future needs for multiple mesh

models/attention/attention.py
bzgoogle: yes, I updated with jaxtyping, there's some pre-commit format check error,  but will address it later

models/attention/attention.py
bzgoogle: Done

models/TransformerBlock.py
bzgoogle: added comments/docstrings for the blocks and major functions.

models/sharding/sharding.py
bzgoogle: per offline discussion, we will use a single mesh, but will map the mesh axis to a clear names for prefill and decode

models/llama4/model.py
bzgoogle: The model level binaries will be addressed in a separate PR

models/quantization/quantization.py
bzgoogle: per offline discussion, the quantization part will be added in later PRs. An interface has been kept for each block in this PR

models/llama4/model.py
bzgoogle: The model level files will be addressed in a separate PR. We will take the above comments in.

models/sharding/sharding.py
vipannalla: nit:  remove underscore and use`ShardingStrategy` for class name please

models/sharding/sharding.py
```diff
@@ -0,0 +1,149 @@
...
+        elif op_mode == 'generate'"
+          if shading_strategy.data_parallelism > 0:
+              ...
+          if shading_strategy.tensor_parallelism > 0:
+              ...
+          if shading_strategy.expert_parallelism > 0:
+              ...
+          if shading_strategy.sequence_parallelism > 0:
+              ...
+          generate_sharding_cfg = OpSpecificShardingConfig(...)
```
vipannalla: We also need to support a `mixed` op_mode which can do both prefill and decode.

models/constants.py
```diff
@@ -0,0 +1,31 @@
...
+    'per_layer_max_router_logits',
+    'over_capacity',
+    'expert_assignment_fraction',
+    'dispatched_to_0',
+    'dispatched_to_1',
+    'dispatched_to_2',
+    'total_dispatch_weight',
+    'router_w_clusterfactor',
+    'average_entropy',
+    'entropy_average',
```
vipannalla: Where are these used? If not used yet, should remove for now and make this PR about model configs to make it simpler.

models/sharding/sharding.py
```diff
@@ -0,0 +1,149 @@
...
+class Mesh:
+    def __init__(self, device_grid , axis_names: Tuple[str, ...]):
+        self.device_grid = device_grid
+        self.axis_names = axis_names
+        ...
+
+
+@dataclass
+class AttentionShardingConfig:
+    # demo only, suffix needs to be verified
```
vipannalla: Lets use this naming convention in comments, inspired from https://source.corp.google.com/piper///depot/google3/learning/gemini/gemax/core/models/gemini3/models/transformer.py;l=342?q=ShardingConfig&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental :


```
"""
 We follow this einsum axis naming convention:
    b: batch
    t: query length
    s: k/v length
    n: num query heads
    k: num k/v heads
    h: head dim
    d: d_model
    f: d_ff
    v: vocab size
    
   And these are for MoE specific axes:
    g: num groups
    s: group size
    e: num experts
    c: expert capacity
    x: num expert shards
    r: num replicas
 """
```

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
...
+        self.embedder = self.create_embedder()
+        # a better way to guarantee the order of initialization 
+        # i.e. sharding_cfg, quantization should be ready before global_KV_cache etc
+        self.sharding_cfg = self.create_sharding_cfg()
+        self.quantization = self.create_quantization()
+        self.global_KV_cache = self.create_KV_cache()
+        self.global_runtime_params = create_global_runtime_params()
+
+        self.dense_blocks = [
+            self.cfg.transformer_moe_blocks_config.make(
```
vipannalla: you meant to use `transformer_dense_blocks_config` here?

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
...
+                name=f'dense_layer_{i}',
+                runtime_param=self.global_runtime_params[i])
+            for i in self.cfg.dense_layers
+        ]
+        self.moe_blocks = [
+            self.cfg.transformer_dense_blocks_config.make(
```
vipannalla: same here, use `transformer_moe_blocks_config`

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
...
+                name=f'moe_layer_{i}',
+                # demo only, retrieval of runtime_param should not be upon index
+                runtime_param=self.global_runtime_params[self.cfg.dense_layers + i])
```
vipannalla: nit: self.cfg.moe_layers

models/llama4/model.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+    def __call__(
+        self,
+        inputs,
+    ):
+        x = self.embedder.encode(inputs)
+        for i, block in self.dense_blocks + self.moe_blocks:
+            x, new_cache = block(x)
+            self.global_KV_cache.append(new_attn_cache)
```
vipannalla: nit: new_cache

models/llama4/model.py
vipannalla: By default, vLLM will the hf.LLamaConfig to tpu_commons. How are you going to create and pass this new config and where would sharding details be set?

models/TransformerBlock.py
bzgoogle: the op mode is not a attributes to transformer but a running state, so i would think adding it in the __call__() to control the ops in Transformer

models/sharding/sharding.py
bzgoogle: Yes, the naming pattern is currently from internal stack, which is also consistent dragon book

models/sharding/sharding.py
bzgoogle: Done

models/constants.py
bzgoogle: removed for now. And will add back when we add the metrics visualization methods

models/TransformerBlock.py
bzgoogle: Done

models/sharding/sharding.py
bzgoogle: Done, added a empty class as a placeholder

models/sharding/sharding.py
bzgoogle: For this v1 version ,we will support prefill/generate modes only, but will add support for more modes in the next version

models/llama4/model.py
bzgoogle: fixed

models/llama4/model.py
bzgoogle: fixed

models/llama4/model.py
bzgoogle: fixed in llama4.py

models/llama4/model.py
bzgoogle: Currently we rely on vllm_config.model_config.hf_config to get model parameters and will rely on the vllm_config.additional_config for overwriting optimization related configs.

models/llama4/model.py
bzgoogle: it's fixed in the new llama4.py

tpu_commons/models/jax/common/sharding.py
```diff
@@ -0,0 +1,398 @@
+from dataclasses import dataclass
+
+import jax
+import numpy as np
+from jax.sharding import Mesh
+
+
+BATCH_AXIS_NAME = 'data'
+SEQUENCE_AXIS_NAME = 'data'
```
mitalisi, line 9: Should be seq and not data

tpu_commons/models/jax/common/moe/moe.py
```diff
@@ -0,0 +1,286 @@
...
+        shape = (self.cfg.d_model, self.cfg.num_experts)
+        self.kernel_DE = self.param_factory.create_kernel_param(
+            rngs, shape=shape, dtype=self.cfg.dtype, sharding=self.ed_sharding)
+
+    def create_sharding(self):
+        """Creates sharding configurations for activations and kernel."""
+        mode_dependent_attrs = [
+            "activation_ffw_btd",
+        ]
+        for attr_name in mode_dependent_attrs:
```
mitalisi, line 109: This is being repeated as well. Perhaps can be extracted out.

tpu_commons/models/jax/common/moe/moe.py
```diff
@@ -0,0 +1,286 @@
...
+            prefill_sharding_config = getattr(
+                self.sharding_cfg.prefill_sharding_cfg, attr_name)
+            generate_sharding_config = getattr(
+                self.sharding_cfg.generate_sharding_cfg, attr_name)
+
+            sharding_dict = {
+                'prefill': NamedSharding(self.mesh,
```
mitalisi, line 116: Define constants and use those instead of strings in code.

tpu_commons/models/jax/common/moe/moe.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def generate_kernel(self, rngs: nnx.Rngs):
+        """Generates the router kernel (weights) for routing."""
+        shape = (self.cfg.d_model, self.cfg.num_experts)
+        self.kernel_DE = self.param_factory.create_kernel_param(
+            rngs, shape=shape, dtype=self.cfg.dtype, sharding=self.ed_sharding)
+
+    def create_sharding(self):
+        """Creates sharding configurations for activations and kernel."""
+        mode_dependent_attrs = [
```
mitalisi, line 106: Add a comment what this variable means - attributes which have different sharding based on inference mode.

tpu_commons/models/jax/common/moe/moe.py
```diff
@@ -0,0 +1,286 @@
...
+        if self.cfg.apply_expert_weight_before_computation:
+            with jax.named_scope("pre_computing_weight"):
+                # need optimization for the out-product
+                weighted_x_BTXD = jnp.einsum('BTD,BTX->BTXD', x, weights_BTX)
+            return self._moe_fwd(weighted_x_BTXD,
+                                 jnp.ones_like(full_weights_BTE), op_mode)
+        else:
+            return self._moe_fwd(x, full_weights_BTE, op_mode)
+
+    def generate_kernel(self, rngs: nnx.Rngs):
```
mitalisi, line 205: init_kernel? SInce we use generate for a mode or functions which are specific to generate.

tpu_commons/models/jax/common/sharding.py
```diff
@@ -0,0 +1,398 @@
...
+    Attributes:
+        tensor_parallelism: The degree of tensor parallelism (e.g., splitting
+            weights of a single layer).
+        expert_parallelism: The degree of expert parallelism for MoE models.
+        sequence_parallelism: The degree of sequence parallelism (splitting
+            activations along the sequence length dimension).
+        data_parallelism: The degree of data parallelism (splitting the batch
+            across devices).
+    """
+    tensor_parallelism: int = 1
```
mitalisi, line 34: Do we do sanity checks  on these based on number of chips?

tpu_commons/models/jax/common/sharding.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        # Use a factory pattern to avoid mutable default arguments
+        self.prefill_sharding_cfg = prefill_sharding_cfg if prefill_sharding_cfg is not None else OpShardingConfig(
+        )
+        self.generate_sharding_cfg = generate_sharding_cfg if generate_sharding_cfg is not None else OpShardingConfig(
+        )
+
+
+@dataclass
+class Sharding:
```
mitalisi, line 157: Should this be model specific. Base class only with mlp and attention and extended to moe.

tpu_commons/models/jax/common/sharding.py
```diff
@@ -0,0 +1,398 @@
...
+    overrides of these rules.
+
+    Attributes:
+        sharding_strategy: The high-level `ShardingStrategy` instance.
+        sharding_cfg: The generated `ShardingConfig` with detailed rules.
+        mesh: The JAX `Mesh` object representing the device grid.
+    """
+    def __init__(self,
+                 strategy_dict: dict,
+                 prefill_sharding_cfg: dict | None = None,
```
mitalisi, line 172: Should this be a dict itself sharding_cfg[key] where key is DEFAULT, PREFILL, GENERATE.

tpu_commons/models/jax/common/sharding.py
```diff
@@ -0,0 +1,398 @@
...
+        """
+        #TODO: check num_devices % parallelism == 0
+        #TODO: check num_devices == multiply(parallelism(with inferred))
+        return
+
+    def get_sharding_cfg(self) -> ShardingConfig:
+        """Returns the generated sharding configuration."""
+        return self.sharding_cfg
+
+    def build_mesh(self, strategy: ShardingStrategy) -> Mesh:
```
mitalisi, line 205: For trillium we had to optimize the mesh topology - ideally jax should figure out the topology but often the support is not in for new chips. Some functions to support that. See optimize_mesh_for_tpu_v6e=True in MaxText.

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -337,18 +338,37 @@ def execute_model(
...
-        axis_names = ("data", "model")
-        # In case we are in disagg mode, the number of devices can exceed 8.
-        # TODO(fhzhang): fix this properly as we implement disagg serving.
-        if len(self.devices) > 8:
-            self.devices = self.devices[:8]
-        mesh_shape = (1, len(self.devices))
-        self.mesh = jax.make_mesh(mesh_shape, axis_names, devices=self.devices)
-
-        logger.info(f"Init mesh | mesh={self.mesh}")
+        hf_config = self.model_config.hf_config
```
mitalisi, line 345: We also would like to store xla flags somewhere by model and architecture.

PR #21, opened 2025-06-05
Add MMLU/MLPerf benchmarking scripts + BUILDKITE E2E CI Test + README on how to run

jrplatin:
# Description

* Add vLLM-based benchmarking scripts that add support for MMLU/MLPerf (specifically, these scripts should replace those found in vLLM [`benchmarks`](https://github.com/vllm-project/vllm/tree/main/benchmarks))
* Add an E2E benchmark test that will spin up a vLLM server with Llama 3.1 8B, run a single request from the MLPerf dataset against it, and check that the throughput/ROUGE1 score is reasonable
* Add the test to our BUILDKITE/CI

# Tests

N/A

# Checklist


tests/e2e/benchmarking/llama3.1_8b_mmlu_single_prompt.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+
+# This script will test running Llama3.1-8B-Instruct on a single prompt to check that the ROUGE score and overall throughput are reasonable.  Specifically, it will do the following:
+# 1. Spin up a vLLM server
+# 2. Once the server is ready, run the single prompt MMLU benchmark
```
mitalisi: Single prompt is not very useful. We can add a unit test for a single prompt. Why not add our benchmarking serving here for the first iteration.

jrplatin: What do you mean by "our benchmark serving" -- as in the entire set?

requirements.txt
```diff
@@ -10,3 +10,6 @@ google-cloud-storage
 --find-links https://storage.googleapis.com/jax-releases/libtpu_releases.html
 jaxtyping
 flax
+nltk
```
xiangxu-google: Since these libs are for benchmark/CI only, it's better to separate them into different requirements.
It's fine to do this in another PR.

.buildkite/pipeline_jax.yml
```diff
@@ -12,6 +12,16 @@ steps:
...
 
+   - label: "TPU Test 14: E2E vLLM MLPerf Benchmarking Run"
+     key: tpu_test_1
+     soft_fail: true
+     env:
+       TPU_BACKEND_TYPE: "jax"
+     agents:
+       queue: tpu_v6e_queue
+     commands:
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/llama3.1_8b_mmlu_single_prompt.sh
```
xiangxu-google: how long does this test run

tests/e2e/benchmarking/llama3.1_8b_mmlu_single_prompt.sh
mitalisi: Yes we can run entire dataset or a set of sample prompts which shows similar accuracy/performance as the entire set reliably.

scripts/vllm/benchmarking/README.md
```diff
@@ -0,0 +1,82 @@
...
+    --run_eval
+```
+
+Note that you can also specify `mmlu` for the `dataset-name` (but your results will be different).
+
+If all goes well, you should an output similar to (noting that your numbers might be different):
+
+```
+============ Serving Benchmark Result ============
+Successful requests:                     50
```
mitalisi: We should just replace all the numbers with XX if we want reader to ignore them.

scripts/vllm/benchmarking/backend_request_func.py
```diff
@@ -0,0 +1,642 @@
+# SPDX-License-Identifier: Apache-2.0
```
mitalisi, line 3: Where is this code copied from. It is easier to see diffs if you could commit the base code and then a successive PR with your changes. Also perhaps add comments on where this code is copied from in the file.

scripts/vllm/benchmarking/backend_request_func.py
```diff
@@ -0,0 +1,642 @@
...
+    ttft: float = 0.0  # Time to first token
+    itl: list[float] = field(
+        default_factory=list)  # list of inter-token latencies
+    tpot: float = 0.0  # avg next-token latencies
+    prompt_len: int = 0
+    error: str = ""
+    input_request: Optional[RequestFuncInput] = None
+
+
+async def async_request_tgi(
```
mitalisi: We dont run most of these benchmarks. Also see a lot of code related to lora. What is the context on adding these?

scripts/vllm/benchmarking/benchmark_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        json.dump(
+            records,
+            f,
+            cls=InfEncoder,
+            default=lambda o:
+            f"<{type(o).__name__} object is not JSON serializable>",
+        )
+
+
+def postprocess_text(preds, targets):
```
bzgoogle: this func is only used for mmlu dataset so far, I see a similar function for mlperf below. Do we make this one as a local func in eval_accuracy_mmlu? Or we could rename it to postprocess_text_mmlu?

scripts/vllm/benchmarking/benchmark_dataset.py
```diff
@@ -0,0 +1,1386 @@
...
+                    prompt_len=prompt_len,
+                    expected_output_len=new_output_len,
+                    completion=completion,
+                ))
+        self.maybe_oversample_requests(samples, num_requests)
+        return samples
+
+
+# -----------------------------------------------------------------------------
+# MMLUDataset Implementation
```
bzgoogle: nit: would be MLPerf

jrplatin:
I am going to update the benchmarking scripts to remove all of the non-MMLU/MLPerf datasets

.buildkite/pipeline_jax.yml
jrplatin: About [three minutes](https://screenshot.googleplex.com/4vSGZR6MZciT6m6)

requirements.txt
jrplatin: Good point -- will separate out

scripts/vllm/benchmarking/benchmark_utils.py
jrplatin: Good catch -- let me update it so it's more clear, thanks!

scripts/vllm/benchmarking/backend_request_func.py
jrplatin: I just copy and pasted this from vLLM directly, but I'll clean these scripts like you mentioned

scripts/vllm/benchmarking/backend_request_func.py
jrplatin: Good point, let me add a comment

tests/e2e/benchmarking/llama3.1_8b_mmlu_single_prompt.sh
jrplatin: @mitalisi  Ah, I think there was some confusion here -- this is the unit/E2E test

jrplatin: I've added an `all_prompt` script to serve as the full benchmark

scripts/vllm/benchmarking/README.md
```diff
@@ -8,6 +8,8 @@ In order to begin benchmarking TPU Commons (via vLLM), you'll want to copy all o
 * `scripts/vllm/benchmarking/benchmark_serving.py`
 * `scripts/vllm/benchmarking/benchmark_utils.py`
 
+You'll also wnat to install all the necessary requirements via `pip install -r requirements_benchmarking.txt`
```
gpolovets1: nit: typo on 'wnat'

scripts/vllm/benchmarking/README.md
```diff
@@ -0,0 +1,84 @@
...
+---------------Inter-token Latency----------------
+Mean ITL (ms):                           XX
+Median ITL (ms):                         XX
+P99 ITL (ms):                            XX
+==================================================
+Evaluating MLPerf...
+
+Results
+
+{'rouge1': 35.5177, 'rouge2': 16.4534, 'rougeL': 24.0036, 'rougeLsum': 34.2323, 'gen_num': 50}
```
mitalisi: Replace these numbers with XX as well.

tests/e2e/benchmarking/llama3.1_8b_mmlu_single_prompt.sh
mitalisi: > @mitalisi Ah, I think there was some confusion here -- this is the unit/E2E test

Hmm  - we should be clear on the goal. Unit tests are intented to be lightweight more to test functionality. Benchmarks are to test performance regressions. They will require more resources, take more time and run on hardware that we aim to optimize on. Lets discuss this offline.

PR #28, opened 2025-06-10
Update buildkite organization.

andrewkvuong:
# Description

Update buildkite README to reference new buildkite organization.

### DIFF
--- a/.buildkite/README.md
+++ b/.buildkite/README.md
@@ -1,5 +1,7 @@
 # Buildkite
 
+https://buildkite.com/tpu-commons
+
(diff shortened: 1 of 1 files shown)


.buildkite/README.md
```diff
@@ -1,5 +1,7 @@
 # Buildkite
 
+https://buildkite.com/tpu-commons
```
QiliangCui, line 3: is this the final project? 

Do we want to use buildkite/vllm?

andrewkvuong: We will use this for now. Upto the rest of the team to decide whether long term we want our own or share buildkite orgs.

PR #35, opened 2025-06-11
JetStream integration: put execute_model in js threads

sixiang-google:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_commons/core/jetstream_commons/core/orchestrator.py
+++ b/tpu_commons/core/jetstream_commons/core/orchestrator.py
@@ -269,6 +269,7 @@ class Driver:
 
     def __init__(
         self,
+        execute_model_f,
         prefill_engines: Optional[list[engine_api.Engine]] = None,
         generate_engines: Optional[list[engine_api.Engine]] = None,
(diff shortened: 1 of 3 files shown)


tpu_commons/core/jetstream_commons/core/orchestrator.py
```diff
@@ -269,6 +269,7 @@ class Driver:
 
     def __init__(
         self,
+        execute_model_f,
```
jrplatin, line 272: Nit: can you make this `execute_model_fn` just so it's a bit more clear what the `f` is?

tpu_commons/core/jetstream_commons/core/orchestrator.py
jrplatin, line 1: Is there a reason you're commenting out a ton of code here and not just deleting it?

sixiang-google: Might be useful in disagg. Also some of them are orthogonal (lora etc) so will be easier to add back later.

PR #39, opened 2025-06-13
Consolidate + improve E2E benchmarking script + disable sampling if using Greedy decoding

jrplatin:
# Description

* Consolidate the E2E benchmarking scripts into a single script
* Update the MMLU scoring to handle the case where the output is in the form of `(C` (in addition to `(C)`
* Update the E2E benchmarking script to take in arguments for dataset name and number of prompts to run + fix a bug where arguments weren't being parsed correctly
    * These changes also enable running the script locally (as a standalone script (not test))
* Fix a minor "bug" where `do_sampling` was being set to `True` even if the temperature was 0 (i.e. greedy decoding)

# Tests

Verified CI [here](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/46#01976aef-5b85-40ef-8ccf-988e4b60af8c)

# Checklist

### DIFF
--- a/tpu_commons/runner/tpu_jax_runner_v2.py
+++ b/tpu_commons/runner/tpu_jax_runner_v2.py
@@ -21,6 +21,7 @@ from tpu_commons.logger import init_logger
 from tpu_commons.models.jax.attention_metadata import AttentionMetadata
 from tpu_commons.models.jax.model_loader import get_model
 from tpu_commons.runner.tpu_torch_xla_runner import _get_token_paddings
+from tpu_commons.runner.utils import determine_do_sampling
 from tpu_commons.worker.input_batch_jax import CachedRequestState, InputBatch
 
 logger = init_logger(__name__)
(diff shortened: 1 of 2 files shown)


tpu_commons/runner/tpu_jax_runner_v2.py
```diff
@@ -633,7 +633,8 @@ def _prepare_decode(self, scheduler_output: VllmSchedulerOutput) -> Any:
             top_ps[i] = self.input_batch.top_p_cpu[seq_index]
             top_ks[i] = self.input_batch.top_k_cpu[seq_index]
             output_token_indices[i] = input_token_indices[i] + 1
-            if top_ks[i] != 1:
+            # # If the temperature is 0.0 (i.e. greedy decoding), then we don't need to sample
+            if top_ks[i] != 1 and temperatures[i] != 0.0:
```
gpolovets1: You could consider making this a function so that if you update the criteria, you don't have to make manual changes and possibly forget in some of the sections.

jrplatin: Ah good point, will update

PR #41, opened 2025-06-16
Integrate an inference microbenchmark

jrplatin:
# Description

In this PR, we add a inference microbenchmark,  which is designed as a vLLM analog to MaxText's own [inference microbenchmark](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/inference_microbenchmark.py).  Specifically, it attempts to isolate and benchmark (with as little overhead as possible) prefill and decode for a given number of benchmark iterations (10 by default) after running a brief warmup.

Please see the README (`scripts/inference/README.md`) for more specific details.


scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+    bw_per_device = (GB_per_step_per_device /
+                     seconds_per_step) if seconds_per_step > 0 else 0
+
+    if verbose:
+        print(
+            f"\nDecode benchmark results:\n"
+            f"WARNING: these results include some overhead from the scheduler, so they may not reflect true prefill performance and we recommend checking the profiles directly!\n"
+            f"\tAR step average time: {decode_average_ms:.3f} ms\n"
+            f"\tAR step average time per seq: {ar_average_ms_per_seq:.3f} ms\n"
+            f"\tAR global batch size: {ar_global_batch_size}\n"  # TODO: is this really 1? Or is it 8?
```
jrplatin, line 381: This is one thing I am curious about, especially since we pad the number of decoding sequences [here](https://github.com/vllm-project/tpu_commons/blob/1bab139/tpu_commons/runner/tpu_jax_runner_v2.py#L599)

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+    num_model_params = engine_core.model_executor.driver_worker.worker.model_runner.total_model_params_num
+    model_size_bytes = num_model_params * jnp.dtype(
+        vllm_model_config.dtype).itemsize
+    # This will be a list of length num decoder layers where each entry is a tuple of (key cache, value cache)
+    kv_caches = engine_core.model_executor.driver_worker.worker.model_runner.kv_caches
+    # total KV cache size is number of layers * 2 (for key and value caches) * KV cache size
+    kv_cache_num_params = len(kv_caches) * 2 * kv_caches[0][0].size
+    kv_cache_size_bytes = kv_cache_num_params * kv_caches[0][1].dtype.itemsize
+
+    prefill_len_padding = engine_core.model_executor.driver_worker.worker.model_runner.scheduler_config.prefill_len_padding
```
jrplatin, line 499: Looks like we are also padding the prefill batch size [here](https://github.com/vllm-project/tpu_commons/blob/1bab139/tpu_commons/runner/tpu_jax_runner_v2.py#L749), so I think we'll need to account for this?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+               prompt_ids: list[int]):
+    """
+    Run a warmup pass before the actual benchmark.
+
+    Args:
+        engine_core: EngineCore instance.
+        tokenizer: Tokenizer instance.
+        prompt_ids: input prompt tokens.
+    """
+    for _ in range(WARMUP_ITERS):
```
mailvijayasingh, line 396: not super necessary - but do we want to get some stats on compile/warmup also?

scripts/inference/inference_microbenchmark.py
mailvijayasingh: IMO, we should account for this. We can start by counting this in, it should not add a lot. WDYT?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
+# SPDX-License-Identifier: Apache-2.0
```
mailvijayasingh, line 1: generic comment - lets create a todo on profiling subsequent prefill also? This will help in understanding any perf gap.

scripts/inference/inference_microbenchmark.py
jrplatin: I'd lean towards no and add it as a TODO since I don't foresee any use case for that?

scripts/inference/inference_microbenchmark.py
jrplatin: What exactly do you mean, sorry?

scripts/inference/README.md
```diff
@@ -0,0 +1,42 @@
+# Inference Microbenchmark Script
+
+This script is designed as a vLLM analog to MaxText's own [inference microbenchmark](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/inference_microbenchmark.py).  Specifically, it attempts to isolate and benchmark (with as little overhead as possible) prefill and decode for a given number of benchmark iterations (10 by default) after running a brief warmup.
+
+Note that our main entrypoint to vLLM is the [`EngineCore`](https://github.com/vllm-project/vllm/blob/799397e/vllm/v1/engine/core.py#L55) class, rather than something higher level like the [`LLM`](https://github.com/vllm-project/vllm/blob/799397ee4f57b90ee1b5f12f88b12f4de0de0d1d/vllm/entrypoints/llm.py#L60) class, which gives us a bit more finegrained control, especially with accessing the TPU model runner and scheduler.
+
```
gpolovets1, line 6: Wondering if these comments regarding design motivation should be published in the public-facing readme or if there is a better avenue to share this..

scripts/inference/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Command-Line Options
+
+This script accepts arguments specific to benchmarking but can also take in any vLLM-specific argument as well (note the defaults of a few the relevant settings below).
+
+### Benchmark-Specific Options
+
+| Argument | Type | Default | Description |
+|---|---|---|---|
+| `--prompt` | `str` | `"I love to"` | The initial prompt to use for the benchmark. This will be padded to the nearest multiple of the `prefill_len_padding` setting. |
```
gpolovets1, line 26: Can you also pass a batch/multiple requests to microbenchmark?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+from vllm.utils import FlexibleArgumentParser
+from vllm.v1.engine import EngineCoreRequest
+from vllm.v1.engine.core import EngineCore
+from vllm.v1.executor.abstract import Executor
+from vllm.v1.request import Request
+
+from tpu_commons.utils_jax import (calculate_prefill_tflops_per_device,
+                                   pad_tokens)
+
+MODEL_NAME = "meta-llama/Llama-3.3-70B-Instruct"
```
gpolovets1: nit: Maybe more informative to name these "DEFAULT_*"? E.g. DEFAULT_MODEL

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+        engine_core.scheduler.running.clear()
+
+    if len(engine_core.scheduler.requests) > 0:
+        engine_core.scheduler.requests.clear()
+
+    if hasattr(engine_core.scheduler, 'swapped') and len(
+            engine_core.scheduler.swapped) > 0:
+        engine_core.scheduler.swapped.clear()
+
+    assert len(engine_core.scheduler.waiting) == 0
```
gpolovets1: nit: Ran previously suggested replacing these with raise conditions. Also you can probably just have one raise condition for these checks.

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+    assert len(engine_core.scheduler.running) == 0
+    assert len(engine_core.scheduler.requests) == 0
+
+
+def run_prefill(engine_core: EngineCore,
+                tokenizer,
+                prompt_ids: list[int],
+                should_profile: bool,
```
gpolovets1: nit: how about naming something like `is_jax_profiler`?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+
+    vllm_model_config = engine_core.vllm_config.model_config
+
+    total_time_s = 0.0
+
+    prefill_len_for_tflops = len(prompt_ids)
+
+    for i in range(benchmark_iters):
+        # Ensure scheduler is clean for each iteration's fresh start.
+        clear_scheduler_state(engine_core)
```
gpolovets1, line 153: Good practice! Curious if the motivation was avoiding pre-emption and/or avoiding running into memory limits?

scripts/inference/README.md
jrplatin: How do you mean, sorry? When TPU commons goes public, this should be available, no?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+        assert len(engine_core.scheduler.waiting) == 1
+        assert len(engine_core.scheduler.running) == 0
+
+        # Profile only for the first iteration
+        if should_profile and i == 0:
+            engine_core.profile(is_start=True)
+
+        start_time_iter = time.perf_counter()
+        _ = engine_core.step()
+        jax.block_until_ready(engine_core.scheduler.running)
```
gpolovets1, line 186: Just confirming that engine_core.scheduler.running contains Jax objects?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+        num_model_params, prefill_len_for_tflops, vllm_model_config, log=False)
+
+    tflops_per_sec_per_device = total_tflops_per_device_value / (
+        prefill_average_ms / 1000.0)
+
+    if verbose:
+        print(
+            f"\nPrefill benchmark results for length {prefill_len_for_tflops}:\n"
+            f"WARNING: these results include some overhead from the scheduler, so they may not reflect true prefill performance and we recommend checking the profiles directly!\n"
+            f"\tPrefill step average time: {prefill_average_ms:.3f} ms\n"
```
gpolovets1, line 240: Would it also be worth printing tokens/s?

scripts/inference/inference_microbenchmark.py
gpolovets1: engine_core.scheduler.running is probably the batch size? Although this may not be static.
You could also print something like "implied batch size" which is max_tokens / max_seq_len?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+
+    Args:
+        num_model_params: Number of model parameters.
+        model_size_bytes: Size of the model in bytes.
+        num_kv_cache_params: Number of KV cache parameters.
+        kv_cache_size_bytes: Size of the KV cache in bytes.
+    """
+    num_model_params_in_billions = num_model_params / 1e9
+    total_model_param_size_in_gb = model_size_bytes / 1e9
+    avg_model_param_size = model_size_bytes / num_model_params
```
gpolovets1: Is this just the dtype of the weights?

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+
+    vllm_config.cache_config.block_size = block_size
+
+    engine_core = EngineCore(vllm_config=vllm_config,
+                             executor_class=executor_class,
+                             log_stats=True)
+    global kv_cache_size_bytes, model_size_bytes, num_model_params
+    num_model_params = engine_core.model_executor.driver_worker.worker.model_runner.total_model_params_num
+    model_size_bytes = num_model_params * jnp.dtype(
```
gpolovets1, line 491: You probably need to take into account the sharding as well.

scripts/inference/inference_microbenchmark.py
```diff
@@ -0,0 +1,515 @@
...
+        # This is a bit hacky, but this logic will ensure that the prefill traces are saved to a subdirectory
+        # of the profile_dir for prefill, called "prefill_length_XXX_YYYY_MM_DD_HH_MM_SS"
+        if should_profile:
+            profile_dir_for_prefill = os.path.join(
+                profile_dir,
+                f"prefill_length_{prefill_length}_{current_timestamp}")
+            update_vllm_profile_dir(engine_core, profile_dir_for_prefill)
+        run_warmup(engine_core, tokenizer, prompt_ids_padded_to_prefill_length)
+
+        clear_scheduler_state(engine_core)
```
gpolovets1: nit: These seem to be in both the profiling step methods and in the driver code. Would it make sense to just add these in one consistent place like having each profiling step method take care of clearing?
Thinking from the perspective of debuggibility.

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+  Returns:
+    A tuple containing: (total_tflops, learnable_weight_tflops, causal_attention_tflops)
+  """
+    num_query_heads = vllm_model_config.hf_config.num_attention_heads
+    num_decoder_layers = vllm_model_config.hf_config.num_hidden_layers
+    head_dim = vllm_model_config.hf_config.hidden_size // vllm_model_config.hf_config.num_attention_heads
+
+    learnable_weight_tflops = 2 * num_model_parameters * prefill_length / jax.device_count(
+    ) / 1e12
+    noncasual_attention_flops = (4 * num_query_heads * num_decoder_layers *
```
gpolovets1: typo: noncausal_attention_flops

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+    log: If True, prints the TFLOPs breakdown.
+
+  Returns:
+    A tuple containing: (total_tflops, learnable_weight_tflops, causal_attention_tflops)
+  """
+    num_query_heads = vllm_model_config.hf_config.num_attention_heads
+    num_decoder_layers = vllm_model_config.hf_config.num_hidden_layers
+    head_dim = vllm_model_config.hf_config.hidden_size // vllm_model_config.hf_config.num_attention_heads
+
+    learnable_weight_tflops = 2 * num_model_parameters * prefill_length / jax.device_count(
```
gpolovets1, line 98: Why are we dividing by the device_count? Isn't it sharding-specific?

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+        )
+    return total_tflops, learnable_weight_tflops, causal_attention_tflops
+
+
+def count_model_parameters(params: Any) -> int:
+    """
+    Counts the total number of parameters (individual scalar values) in a JAX PyTree.
+    This includes all arrays within the PyTree.
+    """
+    return jax.tree_util.tree_reduce(
```
gpolovets1: This may also be simpler: sum(x.size for x in jax.tree_leaves(params))

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+
+  Args:
+    lengths: A list of integers.
+    length: An integer.
+
+  Returns:
+    The nearest length to the right in the list.
+  """
+    pos = bisect_left(lengths, length)
+    if pos == len(lengths):
```
gpolovets1, line 138: Doesn't this imply that length > max(lengths)? Wondering if it should be an error or something in this case.

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+
+  Returns:
+    tokens: Tokenized into integers.
+    true_length: Actual length of the non-padded sequence.
+  """
+    assert pad_id == 0, "Further logic required if pad_id not 0."
+    if isinstance(tokens, list):
+        tokens = np.array(tokens)
+    true_length = tokens.shape[-1]
+    padded_length = take_nearest_length(prefill_lengths, true_length)
```
gpolovets1, line 166: Does padding also have to be a multiple of 2 (e.g. 256) for MXU optimality?
If os, maybe we should at least add a print statement notifying the user that their requested prefill_lengths were implicitly modified.

tpu_commons/utils_jax.py
```diff
@@ -67,3 +71,108 @@ def hbm_usage_gb(devices: Any) -> List[Tuple[float, float]]:
...
+    padding = padded_length - true_length
+    if padding < 0:
+        print("Provided sequence longer than available.")
+        # Take the last N tokens if we have too many.
```
gpolovets1, line 172: May want to include this in the print and/or track statistics of edge cases in the benchmarking (e.g. could print #sequences that were truncated / total # sequences).

scripts/inference/inference_microbenchmark.py
jrplatin: So running will be the actual user requests, but that won't account for this padding we are adding -- in terms of `max_tokens / max_seq_len`, I don't think that's really the BSZ though, right? That's more like the average tokens per sequence?

scripts/inference/README.md
jrplatin: That's a good call-out and theoretically, yes but we don't have the logic right now -- do you think that's important enough to add now?

scripts/inference/inference_microbenchmark.py
jrplatin: Ah good point, yep -- I know our team uses a mix of `assert`s and `raises` but I think the latter is more standard practice, so I would switch!

scripts/inference/inference_microbenchmark.py
jrplatin: What about `should_enable_jax_profiler`?

scripts/inference/inference_microbenchmark.py
jrplatin: Basically, if we don't do this, there will be requests from the previous iteration, which will be scheduled ahead of the new requests

scripts/inference/inference_microbenchmark.py
jrplatin: Also good catch -- I don't think it does (it's just a list of `Requests`), so I guess this will be a no-op and we can remove it

scripts/inference/inference_microbenchmark.py
jrplatin: We print that for `decode` but I can add if you think it's important -- we don't currently do that in MaxText though

scripts/inference/inference_microbenchmark.py
jrplatin: Yep, it's a good sanity check in a way I think (plus we do it in MaxText)

scripts/inference/inference_microbenchmark.py
jrplatin: Interesting -- let me think on this because I don't think we do that in MaxText

scripts/inference/inference_microbenchmark.py
jrplatin: Oh good catch -- let me just move all of `clear_scheduler_state` calls to be in the `prefill` / `decode` functions

tpu_commons/utils_jax.py
jrplatin: We took this code from [MaxText](https://github.com/AI-Hypercomputer/maxtext/blob/d5ecf6d042ca20ea6b0a5d81671289926d91f0d7/MaxText/maxtext_utils.py#L347) -- is there something different here?

tpu_commons/utils_jax.py
jrplatin: I took this from [JetStream](https://github.com/AI-Hypercomputer/JetStream/blob/29329e8e73820993f77cfc8efe34eb2a73f5de98/jetstream/engine/token_utils.py#L54) so I think it should be fine as is?

tpu_commons/utils_jax.py
jrplatin: Sorry, not sure I follow you here

scripts/inference/inference_microbenchmark.py
gpolovets1: #User requests that are 'running' should roughly be batch size? I'm not sure if you mean that padding can cause some pre-emption which may be able to take into account?
For the second point, I don't think we can exceed max_seq_len or max_tokens when we perform padding so it should still stand if that's something you want to report (as a rough correlate with batch size).

scripts/inference/README.md
gpolovets1: I mean, would it benefit users of tpu_commens to know about the related MaxText code pointers and which entrypoint we decided to place the benchmarking code? Feels a bit tmi/may be more suited for internal design discussion but ultimately it's somewhat of a nit =]

scripts/inference/README.md
gpolovets1: Will this break if we use data parallelism in our sharding?

scripts/inference/inference_microbenchmark.py
gpolovets1: How about `enable_jax_profiler` as a middle ground =]

scripts/inference/inference_microbenchmark.py
gpolovets1: Makes sense, thanks!

scripts/inference/inference_microbenchmark.py
gpolovets1: You probably don't want to remove it since block_until_ready is required. But maybe you can apply it to the array output from engine_core.step()?

scripts/inference/inference_microbenchmark.py
gpolovets1: It probably doesn't hurt, to avoid having to do an unnecessary mental translation. I think all the ingredients are already here to just add a print statement.

scripts/inference/inference_microbenchmark.py
gpolovets1: If you plan on using it as a sanity check then may be worth printing or throwing an error if it doesn't match? But I would probably rename to "Weight dtype/precision" - avg_model_param_size is kind of an unorthodox way of stating this IMO.

scripts/inference/inference_microbenchmark.py
gpolovets1: Yea it could be a bit misleading if you're not sharding all of TP!= #devices.

scripts/inference/inference_microbenchmark.py
gpolovets1: Nice - thank you!

tpu_commons/utils_jax.py
gpolovets1: Just wanted to confirm whether you can set prefill_lenghths to anything (e.g. 117) or if there is somewhere in the code that enforces it to be a power of 2.

scripts/inference/README.md
jrplatin: Oh I see what you're getting at -- I can remove the references

scripts/inference/README.md
jrplatin: I am not sure, but I think we can deal with that in the future

tpu_commons/utils_jax.py
jrplatin: Oh yeah, in the code it'll be forced to a multiple of 128 (see [here](https://github.com/vllm-project/tpu_commons/blob/a1fd8e7b14eda7390bc3abbd4a5a00fcc24ae27c/tpu_commons/runner/jax/tpu_jax_runner.py#L71))

jrplatin:
Closing for now since @mailvijayasingh  is working on  a successor that removes the vLLM dependency

PR #55, opened 2025-06-18
Add expert parallelism support in JaxFusedMoE.

hfan:
# Description

Shard JaxFusedMoE weight tensors by the expert dim and use `ragged_all_to_all` to send/recv the `gmm` results from all experts.

# Tests

Ran the Qwen3-30B-A3B and Mixtral-8x7B models, note the `--enable-expert-parallel` command line argument:

```
MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python examples/offline_inference/basic/generate.py     --model=Qwen/Qwen3-30B-A3B     --tensor_parallel_size=4     --task=generate     --max_model_len=1024     --max_num_seqs=1 --enable-expert-parallel
```

```
MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python examples/offline_inference/basic/generate.py     --model=mistralai/Mixtral-8x7B-Instruct-v0.1     --tensor_parallel_size=8     --task=generate     --max_model_len=1024     --max_num_seqs=1  --enable-expert-parallel
```

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/jax_fused_moe.py
+++ b/tpu_commons/models/vllm/jax_fused_moe.py
@@ -9,6 +9,7 @@ from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torch.nn.parameter import Parameter
 from torchax.interop import jax_view, torch_view
 from torchax.tensor import t2j
+from vllm.config import ParallelConfig
 from vllm.model_executor.layers.fused_moe import FusedMoE
 
 P = PartitionSpec
@@ -79,7 +80,7 @@ def _get_tiling_size_for_gmm_kernel(m: int, k: int, n: int,
     return tm, tk, tn
 
 
-def sharded_gmm(
+def tensor_sharded_gmm(
     lhs: jax.Array,
     rhs: jax.Array,
     group_sizes: jax.Array,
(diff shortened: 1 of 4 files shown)


tpu_commons/models/vllm/jax_fused_moe.py
```diff
@@ -218,6 +381,7 @@ def __init__(self, fused_moe: torch.nn.Module, mesh: Mesh):
         self.global_num_experts = fused_moe.global_num_experts
         self.renormalize = fused_moe.renormalize
         self.reduce_results = fused_moe.reduce_results
+        self.use_ep = vllm_parallel_config.enable_expert_parallel
```
lsy323, line 384: We can get `use_ep` from the `FusedMoE` layer directly https://github.com/vllm-project/vllm/blob/8b6e1d639c66d5828d03a7df2c3a500030a5c5cd/vllm/model_executor/layers/fused_moe/layer.py#L865. So that we don't need to pipe the vllm parallel config through.

lsy323:
Left a comment to avoid piping parallel config through the module swap workflow. Otherwise lgtm!

tpu_commons/models/vllm/jax_fused_moe.py
hfan: I thought about that too, but then the use_ep in FusedMoEParallelConfig is determined also by tp_size and dp_size, which we overwrite as 1 in order to load the vllm model into CPU before moving and sharding the model to TPU: https://github.com/vllm-project/vllm/blob/8b6e1d639c66d5828d03a7df2c3a500030a5c5cd/vllm/model_executor/layers/fused_moe/layer.py#L171

So I figured more reliable way to get the flag is from the config

PR #69, opened 2025-06-21
Optimize nnx model loading

xiangxu-google:
# Description

```
    # We first create an abstract model without allocating any weights,
    # then fill in its weigths during load_weights from HF.
    # This shows 3 advantages than the normal way:
    # 1. The model weights will only be allocated once. Otherwise the normal way
    #    will random-init the model weights first, then load the real weights.
    #    The two pass weights allocation causes model loading slow.
    # 2. The model loading won't be OOM. Otherwise the normal way will hold
    #    a full model weights after random-init, then duplicate a layer during
    #    the load_weights. This would be easy to OOM if the layer is super large.
    # 3. The model architecture definition won't need to worry about the sharding.
    #    The sharding definition is taken over by the load_weights instead.
```

# Tests

Llama3.1 8B, 70B
Qwen2.5 1.5B

# Checklist

### DIFF
--- a/tpu_commons/worker/tpu_worker_jax.py
+++ b/tpu_commons/worker/tpu_worker_jax.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 from typing import Optional
 
 import jax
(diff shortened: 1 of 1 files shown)


tpu_commons/worker/tpu_worker_jax.py
```diff
@@ -107,7 +107,11 @@ def execute_model(
 
     def profile(self, is_start: bool = True):
         if is_start:
-            jax.profiler.start_trace(self.profile_dir)
+            options = jax.profiler.ProfileOptions()
+            # Set this to 1 to enable python stack trace
+            options.python_tracer_level = 0
```
jrplatin: Is it worth making this read from a config variable/env var (e.g. `options.python_tracer_level` = os.environ.get("PYTHON_TRACER_LEVEL", 0`)?

xiangxu-google: done, thanks!

PR #73, opened 2025-06-24
Improve step time by 5% by avoiding useless all-gather on logits

py4:
# Description

Returning full unsharded logits is costly and results in expensive all-gather. It's ~6ms for 2K tokens with 128K vocab.
        # In future, if we need returning logits, it should be through topK not the entire logits, or at least through a flag not a default.

# Tests

local tests

# Checklist


wwl2755-google:
LGTM!

PR #84, opened 2025-06-26
Refactor update states among existing runners.

py4:
# Description

# Tests

```
PYTHONPATH=$PYTHONPATH:/home/pooyam/vllm TPU_BACKEND_TYPE=jax python tpu_commons/examples/offline_inference.py      --task=generate     --max_model_len=16384 --model=/dev/shm/Llama-3.1-8B-Instruct/ --tensor-parallel-size=4 --max-num-batched-tokens=4096
```

passed

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/input_batch_jax.py
+++ b/tpu_commons/runner/jax/input_batch_jax.py
@@ -9,6 +9,7 @@ import jax.numpy as jnp
 import numpy as np
 from vllm.lora.request import LoRARequest
 from vllm.multimodal.inputs import MultiModalKwargs, PlaceholderRange
+from vllm.pooling_params import PoolingParams
 from vllm.sampling_params import SamplingParams, SamplingType
 from vllm.utils import swap_dict_values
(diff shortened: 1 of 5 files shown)


lsy323:
Hi @py4, thanks for the PR, right now, we sync ptxla / torchax path from vLLM -> tpu_commons by copy pasting, this sync is manual at this moment, so I'd rather keep them in the torchxla / torchax for now,  to make the sync easier.

I agree this is a great initiative, and we can keep the changes in JAX runner at this moment.

Whenever we move all dev works from vLLM to tpu_commons for torchxla / torchax work, we won't need to manually sync between the 2 repos, at that time, we can use the common function.

py4:
@lsy323 Thanks for your review! We can still sync manually. It's mostly for getting rid of duplicity. Are you saying the `_update_states` for torchax runner and `_update_states` for ptxla_runner will be evolving, and hence syncing differently? Otherwise, if there is a single `_update_states` in the repo, we can keep manually syncing still

PR #87, opened 2025-06-27
Support dynamic grid in ragged kv cache update kernel

yaochengji:
# Description

Add dynamic grid support in ragged kv cache update kernel, it provides about 1% e2e performance improvement.

# Tests

pytest -s -v tests/ragged_kv_cache_update_test.py

### DIFF
--- a/tpu_commons/attention/backends/pallas_torchax.py
+++ b/tpu_commons/attention/backends/pallas_torchax.py
@@ -119,11 +119,11 @@ class PallasMetadata:
     context_lens: torch.Tensor
     query_start_loc: torch.Tensor
     num_seqs: torch.Tensor
-    num_pages: torch.Tensor
+    num_slices: torch.Tensor
 
     def tree_flatten(self):
         children = (self.slot_mapping, self.block_tables, self.context_lens,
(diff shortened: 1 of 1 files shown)


yaochengji:
cc @xiangxu-google @py4

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -108,7 +114,7 @@ def kv_cache_update(
             num_scalar_prefetch=len(scalar_prefetches),
             in_specs=in_specs,
             out_specs=out_specs,
-            grid=(slices.shape[1] // num_slices_per_block, ),
+            grid=(_ceil_div(num_slices[0], num_slices_per_block), ),
```
vanbasten23, line 117: What's the motivation of using dynamic grid?

vanbasten23: Is the reason what we can do less iteration because the dynamic num_slices could be smaller than the previous static and padded num_slices?

yaochengji: yes, then we can reduce the number of kernel blocks.

PR #102, opened 2025-07-01
[Kernel] Make kv cache write kernel work with SPMD

lsy323:
Add a shard_map wrapper for the kv cache write kernel

Land after https://github.com/vllm-project/tpu_commons/pull/101

Include the change in the above PR for testing

### DIFF
--- a/tests/ragged_kv_cache_update_test.py
+++ b/tests/ragged_kv_cache_update_test.py
@@ -3,7 +3,7 @@ import jax.numpy as jnp
 import numpy as np
 from absl.testing import parameterized
 from jax._src import test_util as jtu
-from jax.sharding import NamedSharding
+from jax.sharding import Mesh, NamedSharding
 from jax.sharding import PartitionSpec as P
 
 from tpu_commons.kernels.ragged_kv_cache_update import kv_cache_update
@@ -21,8 +21,8 @@ def kv_cache_update_ref(new_kv, slot_mapping, kv_cache):
 @jtu.with_config(jax_numpy_dtype_promotion="standard")
 class KVCacheUpdateTest(jtu.JaxTestCase):
 
-    def _generate_input_data(self, page_size, combined_kv_head_num, head_dim,
-                             num_slices_per_block):
+    def _generate_data(self, page_size, combined_kv_head_num, head_dim,
+                       num_slices_per_block):
         page_num = 20
         padded_num_tokens = 128
         prng_key = jax.random.key(1234)
@@ -94,7 +94,7 @@ class KVCacheUpdateTest(jtu.JaxTestCase):
             page_size, combined_kv_head_num, head_dim, num_slices_per_block)
         old_kv_cache_copy = kv_cache.copy()
 
-        mesh = jax.sharding.Mesh(jax.devices(), 'x')
+        mesh = Mesh(jax.devices(), 'x')
         kv_cache_pspec = P(None, 'x', None)
 
         new_kv = jax.device_put(new_kv, NamedSharding(mesh, kv_cache_pspec))


lsy323:
cc @bythew3i as well somehow I can only request 1 reviewer in the PR

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -122,3 +119,46 @@ def kv_cache_update(
...
+    slices: jax.
+    Array,  # [3, slices], list of (kv_cache_start, new_kv_start, slice_len)
+    kv_cache: jax.
+    Array,  # [total_num_pages * page_size, num_combined_kv_heads, head_dim]
+    num_slices: jax.Array,  # [1]
+    *,
+    page_size: int = 32,
+    num_slices_per_block: int = 8,
+    mesh: Mesh | None = None,
+    kv_cache_pspec: P | None = None,
```
bythew3i: Looks like only num_heads sharding is supported. Please add the comment or check to Note that!

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -122,3 +119,46 @@ def kv_cache_update(
...
+):
+    if mesh is None:
+        return _kv_cache_update(new_kv, slices, kv_cache, num_slices,
+                                page_size, num_slices_per_block)
+
+    assert kv_cache_pspec is not None, \
```
bythew3i: Nit: user of `kv_cache_update` can pass bad `kv_cache_pspec`, prefer error throw than assert.

tpu_commons/kernels/ragged_kv_cache_update.py
lsy323: Thanks @bythew3i! May I ask what would be a good way to assert only num_heads sharding is supported? I was thinking about doing something like `kv_cache_pspec == P(None, 'x', None)`, but it might fail if caller uses a different axis name, or the last `None` is omitted.

tpu_commons/kernels/ragged_kv_cache_update.py
lsy323: Thanks! I just updated to raise a `ValueError` instead of assert

tpu_commons/models/torchax/torchax_wrapper.py
```diff
@@ -193,3 +193,43 @@ def wrap_shard_map(q, kv_pages, kv_lens, page_indices, cu_q_lens,
...
+    slices: jax.
+    Array,  # [3, slices], list of (kv_cache_start, new_kv_start, slice_len)
+    kv_cache: jax.
+    Array,  # [total_num_pages * page_size, num_combined_kv_heads, head_dim]
+    num_slices: jax.Array,  # [1]
+    *,
+    page_size: int = 32,
+    num_slices_per_block: int = 8,
+) -> Array:
+    # TODO: Get rid of this wrapper and call from pallas.py directly. Need to
```
yaochengji, line 214: I think we can have a global context and get the mesh information from there later

lsy323: That's a good idea! Thanks @yaochengji, since right now the multi-chip path is broken, so let me do that properly in a separate PR.

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: I did not know the last `None` can be omitted? Are you saying `kv_cache_pspec == P(None, 'x', None)==P(None, 'x')`?

Can we check if size == 3, the kv_cache_pspec[0] and kv_cache_pspec[2] are None? If not, throw error?

lsy323: > I did not know the last None can be omitted?

Yes, trailing `None`s can be ommitted. Please see the following from https://docs.jax.dev/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization.html#namedsharding-gives-a-way-to-express-shardings-with-names

image

We can check the pspec has a length of 2 or 3, and the 2nd dim is not None.

lsy323: Took another thought, actually I think we should not expect the head dim is partitioned. Since it should be legal to have `num_head` dim replicated (from the kernel's perspective)

lsy323: @bythew3i Thanks a lot for the review!

lsy323:
@bythew3i @yaochengji Thanks for the review!

PR #111, opened 2025-07-02
[KV cache] Optimize kv cache allocation logic

lsy323:
Land after https://github.com/vllm-project/tpu_commons/pull/110 lands.

- Create KV cache tensors on device, instead of creating on CPU and then transfer to TPU, this is more efficient
- When determine the amount of memory allocated for KV cache, the existing logic doesn't fully utilize the memory under SPMD. Optimize the memory determination logic.

### DIFF
--- a/tpu_commons/worker/tpu_worker_torchax.py
+++ b/tpu_commons/worker/tpu_worker_torchax.py
@@ -249,7 +249,8 @@ class TPUWorker:
             tpu_kv_cache_bytes = (tpu_kv_cache_bytes * head_size //
                                   padded_head_size)
         if self.original_parallel_config is not None:
-            tpu_kv_cache_bytes *= self.original_parallel_config.world_size
+            tpu_kv_cache_bytes *= \
+                self.original_parallel_config.tensor_parallel_size
 
         return int(tpu_kv_cache_bytes)
 


tpu_commons/worker/tpu_worker_torchax.py
```diff
@@ -247,6 +248,9 @@ def determine_available_memory(self) -> int:
             # errors, even after padding the head_size.
             tpu_kv_cache_bytes = (tpu_kv_cache_bytes * head_size //
                                   padded_head_size)
+        if self.original_parallel_config is not None:
+            tpu_kv_cache_bytes *= self.original_parallel_config.world_size
```
yaochengji: How about kv_head_num = 4, and world_size = 8?

yaochengji: Or if it's not TP?

tpu_commons/worker/tpu_worker_torchax.py
```diff
@@ -69,7 +70,7 @@ def __init__(
         if self.use_spmd:
             # Under SPMD mode, distributed env is initialized as if there is
             # only one worker/device.
-            self.original_parallel_config = self.parallel_config
+            self.original_parallel_config = copy.deepcopy(self.parallel_config)
```
yaochengji, line 73: What do we need a deepcopy here?

tpu_commons/worker/tpu_worker_torchax.py
lsy323: > How about kv_head_num = 4, and world_size = 8?

Right now it's not supported. See assertion here https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/runner/tpu_torchax_runner.py#L1653-L1655

lsy323: > if it's not TP?

That's a good point. I think we should use TP size here instead of world size. Because only tp will require kv cache to be sharded

tpu_commons/worker/tpu_worker_torchax.py
lsy323: I found if I don't do deepcopy, `original_parallel_config` will be a reference to `parallel_config`. This should be a bug in the previous PR.

tpu_commons/worker/tpu_worker_torchax.py
lsy323: Updated to tp size here

tpu_commons/worker/tpu_worker_torchax.py
lsy323: `parallel_config` is overwritten to have no parallelism, for model init. https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/worker/tpu_worker_torchax.py#L72-L75

tpu_commons/distributed/tpu_distributed_utils.py
```diff
@@ -26,6 +27,25 @@
...
+
+    jax_dtype = t2j_dtype(dtype)
+    if mesh is None:
+        device = jax.devices()[0]
+    else:
+        sharding = NamedSharding(mesh, P(*partition_spec))
+        device = sharding
+
+    jax_t = jnp.zeros(shape, dtype=jax_dtype,
+                      device=device).block_until_ready()
```
vanbasten23, line 44: if mesh is not None, why does it make sense that `device=NamedSharding`? @lsy323

PR #115, opened 2025-07-02
Delete more dead code from InputBatch [Gemini Generated]

fenghuizhang:
# Description

Deleted some dead code from InputBatch, expose some issues:
- Fixed some hacks I left
- Had to reserve 50% memory to avoid an OOM in the disaggregated inference code path.

Issues observed:
- With this change, we start to see the following warning and the system ran out of memory in the DI code path. 

```
/home/fhzhang/miniconda3/envs/vllm/lib/python3.10/site-packages/jax/_src/interpreters/mlir.py:1184: UserWarning: Some donated buffers were not usable: bfloat16[4,2737,32,128]....
See an explanation at https://docs.jax.dev/en/latest/faq.html#buffer-donation.
  warnings.warn("Some donated buffers were not usable:"
```
- The non-DI code path doesn't work when `tensor_parallel_size` is set to 2.

This happens with both DI and non-DI code paths when there are more than 1 TPU involved in the computation.

# Tests

```
PREFILL_SLICES=2 DECODE_SLICES=2 PYTHONPATH=$(pwd)/vllm:$(pwd)/tpu_commons TPU_BACKEND_TYPE=jax python tpu_commons/examples/offline_inference.py --task=generate --model=meta-llama/Meta-Llama-3-8B-Instruct --max_model_len=1024 --max_num_seqs=8

TPU_BACKEND_TYPE=jax python tpu_commons/examples/offline_inference.py --task=generate --model=meta-llama/Meta-Llama-3-8B-Instruct --max_model_len=1024 --max_num_seqs=8 --tensor_parallel_size=4
```

# Checklist

### DIFF
--- a/tpu_commons/core/core_tpu.py
+++ b/tpu_commons/core/core_tpu.py
@@ -118,7 +118,7 @@ class EngineCore:
         # Setup KV Caches and update CacheConfig after profiling.
         num_gpu_blocks, num_cpu_blocks, kv_cache_config = \
             self._initialize_kv_caches(vllm_config, self.prefill_executor)
-        if len(devices) == 8:
+        if self.decode_executor:
             self._initialize_kv_caches(vllm_config, self.decode_executor)
 
         vllm_config.cache_config.num_gpu_blocks = num_gpu_blocks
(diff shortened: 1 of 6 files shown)


fenghuizhang:
It turns out the donation issue is red-herring, it happens at HEAD as well. I am merging this now.

PR #136, opened 2025-07-07
Fix incorrect argument being passed to `sample`

jrplatin:
# Description

In this PR, I fix a bug where an extra `attention_metadata.chunked_prefill_enabled` argument was being passed to `sample` after the latter was refactored.

image


py4:
@jrplatin Thanks for the PR! I had a fix here: https://github.com/vllm-project/tpu_commons/pull/127. I'll get it in today

PR #157, opened 2025-07-09
Deepseek V3 router

wenxindongwork:
# Description

This pull request introduces the `DeepSeekV3Router` module for the DeepSeekV3 MoE layer, along with its config and tests.

### Router Implementation and Configuration:

* [`tpu_commons/models/jax/deepseek.py`](diffhunk://#diff-50574442dcbd5aef9a69d5986c75013bbd84003bda5772891ca9709b35539ee0R1-R184): Added the `DeepSeekV3Router` class, which routes tokens to experts based on input scores, and the `DeepSeekV3RoutingConfig` dataclass for defining the router's configuration. 

### Unit Tests for Router:

* [`tests/models/jax/common/deepseek.py`](diffhunk://#diff-48af42eefb14b7d27b1299e6f0dfb96f58d9430559f29d859e4438786234872eR1-R109): Added unit tests for `DeepSeekV3Router`.

### Updates to Constants:

* [`tpu_commons/models/jax/common/constants.py`](diffhunk://#diff-b8b0ae6d2261ee085eaf43bedb7e3c18c8c01b1f3d1f8c0709d6bfa65b396af7L52-R61): Extended `HuggingFaceArgNames` with new attributes for DeepSeek MoE routing.

### Sharding Configuration Updates:

* [`tpu_commons/models/jax/common/sharding.py`](diffhunk://#diff-9443359725b7cdd9b158325ae4304ed5abd2b4ce1b60688967ce0f0c07d111ecR96-R97): Added `moe_router_bias_e` to `ShardingRulesConfig` for specifying sharding rules for router bias weights.

# Tests

Added unit test. 

# Checklist

### DIFF
--- a/tpu_commons/models/jax/common/moe/deepseek_moe.py
+++ b/tpu_commons/models/jax/common/moe/deepseek_moe.py
@@ -1,4 +1,4 @@
-from dataclasses import field, make_dataclass
+from dataclasses import field, make_dataclass, dataclass
 from typing import Any, Tuple
 
 import jax
@@ -41,6 +41,7 @@ DeepSeekV3RoutingConfig.__doc__ = f"""Configuration for the Router module.
         {HuggingFaceArgNames.NORM_TOPK_PROB.value}: Whether to normalize the top k groups."""
 
 
+@dataclass
 class DeepSeekV3Router(nnx.Module):
     """Router module for Mixture-of-Experts (MoE) layers.
 
@@ -54,21 +55,16 @@ class DeepSeekV3Router(nnx.Module):
         quant: Optional configuration for quantization.
     """
 
-    def __init__(
-        self,
-        cfg: DeepSeekV3RoutingConfig,  # type: ignore
-        mesh: Mesh,
-        param_factory: ParamFactory,
-        sharding_cfg: ShardingConfig,
-        quant: Any | None = None,
-    ):
-        self.cfg = cfg
-        self.mesh = mesh
-        self.param_factory = param_factory
-        self.sharding_cfg = sharding_cfg
-        self.quant = quant
-        self.dtype = self.cfg.dtype
+    cfg: DeepSeekV3RoutingConfig
+    mesh: Mesh
+    param_factory: ParamFactory
+    sharding_cfg: ShardingConfig
+    quant: Any | None = None
 
+
+    def __post_init__(self):
+        """Initializes the Router module by creating sharding configurations and generating the router kernel."""
+        
         self.hidden_size = getattr(self.cfg, HuggingFaceArgNames.HIDDEN_SIZE.value)
         self.num_experts = getattr(
             self.cfg, HuggingFaceArgNames.NUM_ROUTED_EXPERTS.value
@@ -84,11 +80,8 @@ class DeepSeekV3Router(nnx.Module):
         self.routed_scaling_factor = getattr(
             self.cfg, HuggingFaceArgNames.ROUTED_SCALING_FACTOR.value
         )
+        self.dtype = self.cfg.dtype
 
-        self.__post_init__()
-
-    def __post_init__(self):
-        """Initializes the Router module by creating sharding configurations and generating the router kernel."""
         self.create_sharding()
 
     def get_topk_indices(self, scores: Float) -> Float:


tpu_commons/models/jax/deepseek.py
gpolovets1, line 1: Per our code design, maybe it would be better to move this file under moe/deepseek_moe.py

tpu_commons/models/jax/common/constants.py
```diff
@@ -35,23 +36,30 @@ class HuggingFaceArgNames(enum.Enum):
     RMS_NORM_EPS: str = "rms_norm_eps"
     ROPE_SCALING: str = "rope_scaling"
     ROPE_THETA: str = "rope_theta"
-    VOCAB_SIZE="vocab_size"
-    
+    VOCAB_SIZE = "vocab_size"
```
gpolovets1: Maybe we should also add ": str" for consistency.

tpu_commons/models/jax/deepseek.py
```diff
@@ -0,0 +1,183 @@
...
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
+from jaxtyping import DTypeLike, Float
+from vllm.config import VllmConfig
+
+from tpu_commons.models.jax.common.base import Config, ParamFactory
+from tpu_commons.models.jax.common.constants import HuggingFaceArgNames
+from tpu_commons.models.jax.common.sharding import ShardingConfig
+
+DeepSeekV3RoutingConfig = make_dataclass(
```
gpolovets1: You can probably also inherit from "RoutingConfig" to avoid having to redefine attributes. You could do this by setting bases=(RoutingConfig,)

tpu_commons/models/jax/deepseek.py
```diff
@@ -0,0 +1,183 @@
...
+        quant: Any | None = None,
+    ):
+        self.cfg = cfg
+        self.mesh = mesh
+        self.param_factory = param_factory
+        self.sharding_cfg = sharding_cfg
+        self.quant = quant
+        self.dtype = self.cfg.dtype
+
+        self.hidden_size = getattr(self.cfg,
```
gpolovets1: Do we need to add these as instance variables since these are all in the config already?

tpu_commons/models/jax/deepseek.py
```diff
@@ -0,0 +1,183 @@
...
+            dtype=self.cfg.dtype,
+            sharding=self.ed_sharding)
+        self.bias_E = self.param_factory.create_scale_param(
+            rngs,
+            shape=(E, ),
+            dtype=self.cfg.dtype,
+            sharding=None,  # TODO: check if this is correct
+        )
+
+    def create_sharding(self):
```
gpolovets1: Is this a duplicate of moe.py? If so, you can call super().create_sharding()

tpu_commons/models/jax/deepseek.py
```diff
@@ -0,0 +1,183 @@
...
+            self.cfg, HuggingFaceArgNames.NUM_EXPERTS_PER_TOKEN.value)
+        self.n_groups = getattr(self.cfg, HuggingFaceArgNames.NUM_GROUPS.value)
+        self.topk_groups = getattr(self.cfg,
+                                   HuggingFaceArgNames.TOPK_GROUP.value)
+        self.norm_topk_prob = getattr(self.cfg,
+                                      HuggingFaceArgNames.NORM_TOPK_PROB.value)
+        self.routed_scaling_factor = getattr(
+            self.cfg, HuggingFaceArgNames.ROUTED_SCALING_FACTOR.value)
+
+        self.__post_init__()
```
gpolovets1: This is called implicitly - I don't think you need to add this call.

tpu_commons/models/jax/deepseek.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __post_init__(self):
+        """Initializes the Router module by creating sharding configurations and generating the router kernel."""
+        self.create_sharding()
+
+    def get_topk_indices(self, scores):
+        """Get the topk indices of the scores."""
+
+        scores = scores + self.bias_E
```
gpolovets1: Could you include a small comment on the starting scores shape?

tests/models/jax/common/deepseek.py
gpolovets1, line 1: Thanks for adding this unit test! It's a nice way to communicate our assumptions as well =]

wenxindongwork: of course!

tpu_commons/models/jax/common/constants.py
wenxindongwork: good catch!

tpu_commons/models/jax/deepseek.py
wenxindongwork: Since the RoutingConfig contained attributed not present in the deepseekv3routeconfig, I thought it would be unclear for the user which attributes are required and which are irrelevant if we use inheritance here. Would you agree with this?

tpu_commons/models/jax/deepseek.py
wenxindongwork: they are not strictly necessary, I felt it was cleaner to have explicit variable naming in the init function, but I also think it's fine moving these lines into the call function, also open to other ideas!

tpu_commons/models/jax/deepseek.py
wenxindongwork: i thought that too but actually in this case the post_init function will not be automatically called, because this class is not a dataclass.  Do you think it will be better to make this class a dataclass?

tpu_commons/models/jax/deepseek.py
wenxindongwork: yup for sure

tpu_commons/models/jax/deepseek.py
wenxindongwork: This DeepSeekV3Router class is not a child another class. It's not a child of Router because there is not enough commonalities - the kernels and forward pass are both different.

tpu_commons/models/jax/deepseek.py
gpolovets1: Sure, I think that's a reasonable tradeoff =]

tpu_commons/models/jax/deepseek.py
gpolovets1: My personal take is that it's a bit redundant to copy the config entries into the model class (essentially just removing self.cfg from the references) but I will leave it up to you!

tpu_commons/models/jax/deepseek.py
gpolovets1: I think you could inherit the shardings and generate_kernel from Router?

wenxindongwork: the kernels are different -- there is a bias term in the deepseek router that is not present in the default router. Depending on whether we shard the bias or not, the sharding could also be different.

tpu_commons/models/jax/deepseek.py
gpolovets1: Maybe make it a dataclass for now for consistency with the other layers? I think there is still a bit of debate about this in the context of nnx but at least for now, it using dataclasses gives you some benefits like this post_init functionality.

tpu_commons/models/jax/deepseek.py
wenxindongwork: let me move these entries into the post_init function, and make this a dataclass again.

PR #175, opened 2025-07-11
Extract select_hidden_states and sample from model to reduce warmup from O(m*n) to O(m+n)

py4:
# Description

Reduces warmup time from O(m*n) to O(m+n). This is also what ptxla does.

buildkite passed: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/523

warmup time for llama:
* after: 2 min
* before: 16 min

# Tests

on both vllm and jax model impl i get:

```
-------------------------------------------------
Prompt: 'Hello, my name is'
Generated text: ' Emily and I am a 25-year-old freelance writer and editor. I have'
Generated text: ' a city of romance, art, fashion, and cuisine. Paris is a must'
Generated text: ' often remembered using the acronym ROYGBIV, which stands for Red, Orange'
Prompt: 'The future of AI is'
Generated text: ' bright, but it also raises concerns about bias, accountability, and the impact on'
Generated text: ' the head of state and head of government of the United States. The president serves'
```

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -334,11 +334,17 @@ class TPUModelRunner():
             logger.info(f"Precompile sampling --> num_reqs={num_reqs}")
             start = time.perf_counter()
             for do_sampling in (True, False):
-                temperature = np.full((num_reqs, ), 0.7, dtype=np.float32)
-                top_k = np.full((num_reqs, ), 20, dtype=np.int32)
-                top_p = np.full((num_reqs, ), 0.8, dtype=np.float32)
-                (temperature, top_k, top_p) = self._device_array(
-                    (temperature, top_k, top_p))
+                if do_sampling:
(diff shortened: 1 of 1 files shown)


wwl2755-google:
Just curious, would there be any performance degradation on splitting the jitted kernel on forward pass from one to multiple? (Like one fused cudagraph vs multiple small cudagraphs on GPU) vLLM adopted piece-wise cudagraph because of flexibility. How would we prioritize it, if it sacrifices performance?

py4:
Yeah that’s a great point and I also have thought about it. One big graph
gives XLA flexibility to compile better. I’m planning to benchmark this to
understand the trade off and see how things are

py4:
I profiled both codes. 

Before (decode to decode is ~15.9 ms):

image

After (decode to decode is ~15.9ms):

image

The conclusion is i don't observe a performance degradation. @hfan 's intuition was that sampling has data dependency on logits, so there is no much room for "smart" planning/fusing of XLA.

PR #176, opened 2025-07-11
Fix recompile in torchax pt runner

lsy323:
# Description

If a jitted module is called with args placed onto device with `jax.device_put`, the HLO graph would contain replicated sharding annotation on the input arg if it's placed on a single TPU device.

By looking into the HLO graph dumps, I found during serving, all the HLO graphs contain the `replicated` sharding annotation. So during warmup time, we use `device_put` to enforce the sharding annotation on the graph inputs. Therefore we can avoid recompilation during serving.

Some minor clean up.

# Tests

Ran benchmarking, compare number of XLA graph dumps after server warm up and after a benchmarking run, the XLA graph dump size is the same before and after.

# Checklist

- [ x ] I have performed a self-review of my code.
- [ x ] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ x ] I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/tpu_commons/runner/tpu_torchax_runner.py
+++ b/tpu_commons/runner/tpu_torchax_runner.py
@@ -547,10 +547,6 @@ class TPUModelRunner(LoRAModelRunnerMixin):
 
         return kv_cache_spec
 
-    def _create_torchax_array(self, torch_tensor, partition_spec=()):
-        return create_torchax_tensor_with_partition_spec(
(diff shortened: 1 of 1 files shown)


lsy323:
I'll just merge the change, since torchax + PT runner is meant to be experimental (Analyzing torchax model wrapper purpose), and it doesn't affect other code path. Merge the change so that we can track the torchax performance on the dashboard

PR #178, opened 2025-07-11
Fix the wrong warmup input for sampling.

py4:
# Description

# Tests

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [ ] I have performed a self-review of my code.
- [ ] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ ] I have made or will make corresponding changes to any relevant documentation.

PR #180, opened 2025-07-11
Add the ability to catch recompilation in JAX path.

py4:
# Description

1. Guarded with existing `VLLM_XLA_CHECK_RECOMPILATION` so doesn't run by default.
2. Catches recompilation in CI (The flag is enabled by default in docker)
3. That's how I discovered and fixed #178 

When there is a failure:

image

# Tests

```USE_JAX_V1=1 PYTHONPATH=$PYTHONPATH:/home/pooyam/vllm:/home/pooyam/tpu_commons TPU_BACKEND_TYPE=jax python examples/offline_inference.py     --task=generate     --max_model_len=1024 --model=/dev/shm/Llama-3.1-8B-Instruct --tensor-parallel-size=1```

```
Prompt: 'Hello, my name is'
Generated text: ' Emily and I am a 25-year-old freelance writer and editor. I have'
Generated text: ' a city of romance, art, fashion, and cuisine. Paris is a must'
Generated text: ' often remembered using the acronym ROYGBIV, which stands for Red, Orange'
Prompt: 'The future of AI is'
Generated text: ' bright, but it also raises concerns about bias, accountability, and the impact on'
Generated text: ' the head of state and head of government of the United States. The president serves'
```

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -2,6 +2,7 @@ import functools
 import os
 import random
 import time
+from contextlib import nullcontext
 from dataclasses import asdict
 from typing import Any, List, Optional, cast
 
@@ -31,7 +32,7 @@ from tpu_commons.runner.jax.input_batch_jax import (CachedRequestState,
 from tpu_commons.runner.tpu_torch_xla_runner import (_get_padded_token_len,
                                                      _get_req_paddings,
                                                      _get_token_paddings)
-from tpu_commons.runner.utils import (LatencyTracker,
+from tpu_commons.runner.utils import (ForbidCompile, LatencyTracker,
                                       get_padded_num_reqs_with_upper_limit)
 from tpu_commons.sample.metadata_jax import TPUSupportedSamplingMetadata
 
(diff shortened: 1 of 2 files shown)


wwl2755-google:
Just a nit: would it cause any misunderstanding by reusing `VLLM_XLA_CHECK_RECOMPILATION` ?

py4:
@wwl2755-google  `VLLM_XLA_CHECK_RECOMPILATION` is used by PT path to check for recompilation. That's why i wanted to reuse it and not introduce something else for JAX. Do you think it causes confusion?

wwl2755-google:
> @wwl2755-google `VLLM_XLA_CHECK_RECOMPILATION` is used by PT path to check for recompilation. That's why i wanted to reuse it and not introduce something else for JAX. Do you think it causes confusion?

The confusion I meant is: others may think this variable is PTXLA-path-only. I might be wrong, but IIUC, this variable is not make effects on JAX path for now, and should not in the future. My comment is a separate variable like `JAX_CHECK_RECOMPILATION`.

py4:
@wwl2755-google The `VLLM_XLA_CHECK_RECOMPILATION` is a r[egistered env in vllm ](https://screenshot.googleplex.com/3QczVR8JA4UN5nf) and is generic. It's not meant to be used for ptxla. It can be used for any XLA compilation, including JAX. The value is already set to `1` in `run_docker.sh` so it's enabled in JAX path as well in CI. For this reason, i'm not a huge fan of adding a seperate flag for JAX. Why can't the same flag be used for both?

wwl2755-google:
Oh. I see. Thanks for clarification. As this is not ptxla-path only, then it sounds good to me!

PR #185, opened 2025-07-14
create a rope class and deepseek rope inherit from it

Chenyaaang:
# Description

Implement original rotary positional embedding and specific rope for deepseek. 

# Tests

Unit tests verify the cached sin/cos matrix are as expected shape, the output from rope are as expected shape and different from the original input. 

# Checklist

- [X] I have reviewed the redesigned uLLM playbook.

### DIFF
--- a/tests/models/jax/common/test_rope.py
+++ b/tests/models/jax/common/test_rope.py
@@ -8,30 +8,36 @@ from tpu_commons.models.jax.common.rope import (DeepseekScalingRotaryEmbedding,
 class RotaryEmbeddingTest(jtu.JaxTestCase):
 
     def test_apply_rope(self):
-        head_dim = 128
+        head_dim = 2
         rope_theta = 10000
-        original_max_position_embeddings = 4096
+        original_max_position_embeddings = 2
         rope = RotaryEmbedding(head_dim, rope_theta,
                                original_max_position_embeddings)
         self.assertTrue(
             rope.sin_cos_cache.shape == (original_max_position_embeddings,
                                          head_dim))
-        num_tokens = 100
-        num_heads = 8
+        expected_sin_cos = jnp.array([[1, 0], [0.5403023, 0.841471]],
+                                     dtype=jnp.float32)
+        self.assertArraysAllClose(rope.sin_cos_cache, expected_sin_cos)
+
+        num_tokens = 2
+        num_heads = 1
         positions = jnp.arange(num_tokens)
         x = jnp.ones((num_tokens, num_heads, head_dim))
         x_rope = rope.apply_rope(positions, x)
+        expected_x_rope = jnp.array([[[1, 1]], [[-0.30116874, 1.3817732]]],
+                                    dtype=jnp.float32)
         self.assertTrue(x_rope.shape == x.shape)
-        self.assertFalse(jnp.array_equal(x_rope, x))
+        self.assertArraysAllClose(x_rope, expected_x_rope)
 
 
 class DeepseekScalingRotaryEmbeddingTest(jtu.JaxTestCase):
 
     def test_apply_rope(self):
-        head_dim = 128
+        head_dim = 2
         rope_theta = 10000
-        original_max_position_embeddings = 4096
-        scaling_factor = 4
+        original_max_position_embeddings = 1
+        scaling_factor = 2
         rope = DeepseekScalingRotaryEmbedding(
             head_dim, rope_theta, original_max_position_embeddings,
             scaling_factor)
@@ -39,10 +45,16 @@ class DeepseekScalingRotaryEmbeddingTest(jtu.JaxTestCase):
             rope.sin_cos_cache.shape == (scaling_factor *
                                          original_max_position_embeddings,
                                          head_dim))
-        num_tokens = 100
-        num_heads = 8
+        expected_sin_cos = jnp.array([[1.0693147, 0], [0.5777532, 0.8997973]],
+                                     dtype=jnp.float32)
+        self.assertArraysAllClose(rope.sin_cos_cache, expected_sin_cos)
+        num_tokens = 2
+        num_heads = 1
         positions = jnp.arange(num_tokens)
         x = jnp.ones((num_tokens, num_heads, head_dim))
         x_rope = rope.apply_rope(positions, x)
+        expected_x_rope = jnp.array(
+            [[[1.0693147, 1.0693147]], [[-0.32204413, 1.4775505]]],
+            dtype=jnp.float32)
         self.assertTrue(x_rope.shape == x.shape)
-        self.assertFalse(jnp.array_equal(x_rope, x))
+        self.assertArraysAllClose(x_rope, expected_x_rope)
--- a/tpu_commons/models/jax/common/rope.py
+++ b/tpu_commons/models/jax/common/rope.py
@@ -35,16 +35,20 @@ class RotaryEmbedding:
         cache = jnp.concatenate((cos, sin), axis=-1)
         return cache
 
-    def apply_rope(self, positions: jax.Array, x: jax.Array):
-        # positions should be 1d according to vllm.
-        assert x.ndim == 3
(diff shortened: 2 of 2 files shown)


tpu_commons/models/jax/common/rope.py
```diff
@@ -0,0 +1,135 @@
+import math
+
+import jax
+from jax import numpy as jnp
+
+
+class RotaryEmbedding:
```
gpolovets1, line 7: Thanks for porting this into our code! Curious if you could do a quick sanity check to see if you get the same quality/perf when pointing the models in our dashboard to this file instead of tpu_commons/tpu_commons/models/jax/layers/rope.py.
You can push the changes and run build kite as explained in the playbook (and then revert them back before we merge your PR): https://docs.google.com/document/d/1DGQBVvr2bh4G8tBUO1YH8pO7Dd_myw5rfEMfVDymEk8/edit?resourcekey=0-V7MGHu3aQjJH6YrI3-y8Hg&tab=t.0#heading=h.j4pastholmg5

tpu_commons/models/jax/common/rope.py
```diff
@@ -0,0 +1,135 @@
...
+        t = jnp.arange(self.original_max_position_embeddings *
+                       self.scaling_factor,
+                       dtype=jnp.float32)
+        freqs = jnp.einsum("...T,k->...Tk", t, inv_freq)
+        sin, cos = jnp.sin(freqs) * self.mscale, jnp.cos(freqs) * self.mscale
+        cache = jnp.concatenate((cos, sin), axis=-1)
+        return cache
+
+
+def _yarn_get_mscale(scale, mscale):
```
gpolovets1, line 122: Would it be possible to use more informative naming and/or documentation/references? It's a bit hard to follow what/why the different expressions are being performed.

tpu_commons/models/jax/common/rope.py
```diff
@@ -0,0 +1,135 @@
...
+        return cache
+
+    def apply_rope(self, positions: jax.Array, x: jax.Array):
+        # positions should be 1d according to vllm.
+        assert x.ndim == 3
+        cos_sin = self.sin_cos_cache[positions]
+        cos, sin = jnp.split(cos_sin, 2, axis=-1)
+        assert sin.ndim == 2 and cos.ndim == 2
+        cos, sin = cos[:, None, :], sin[:, None, :]
+        x1, x2 = x[..., ::2], x[..., 1::2]
```
gpolovets1: Should this actually be x[..., :x.shape[-1]//2] and x[..., x.shape[-1]//2:]? (https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/layers/embeddings.py#L307)

tests/models/jax/common/test_rope.py
```diff
@@ -0,0 +1,48 @@
...
+        self.assertTrue(
+            rope.sin_cos_cache.shape == (original_max_position_embeddings,
+                                         head_dim))
+        num_tokens = 100
+        num_heads = 8
+        positions = jnp.arange(num_tokens)
+        x = jnp.ones((num_tokens, num_heads, head_dim))
+        x_rope = rope.apply_rope(positions, x)
+        self.assertTrue(x_rope.shape == x.shape)
+        self.assertFalse(jnp.array_equal(x_rope, x))
```
gpolovets1: I think these tests may not necessarily validate the correctness of the RoPE computations. Maybe you can either hand-calculate a simple example, try find an app/implementation online that can calculate it for you, or ask Gemini.
You can also test individual components/calculations of the rope implementations to make calculations easier (it may require breaking your calculations a bit more into functions).

tpu_commons/models/jax/common/rope.py
bzgoogle: the interleaved partition is used in Llama4, and we will need it soon, but I would think use a traditional partition (:h/2) (h/2:h) as the default RopE?

bzgoogle: nit: could we rename the x1,x2 as the first_part, first_half etc, as used in maxtext or the legacy rope?

tpu_commons/models/jax/common/rope.py
```diff
@@ -0,0 +1,135 @@
...
+                           t,
+                           inv_freq,
+                           precision=jax.lax.Precision.HIGHEST)
+        sin, cos = jnp.sin(freqs), jnp.cos(freqs)
+        cache = jnp.concatenate((cos, sin), axis=-1)
+        return cache
+
+    def apply_rope(self, positions: jax.Array, x: jax.Array):
+        # positions should be 1d according to vllm.
+        assert x.ndim == 3
```
bzgoogle: for the apply_rope, as we check the dim constrains for x and sin, cos, could we add shape suffix to the major variables, like x, cos, sins, x1, x2? like x -> x_TNH?

tpu_commons/models/jax/common/rope.py
Chenyaaang: Actually I think both are fine, as long as q and k are applied on the same rope.

tpu_commons/models/jax/common/rope.py
Chenyaaang: Actually I referred to a few implementations and they all referred from Peng et al. github.com/jquesnelle/yarn, which also doesn't have name/documentations on all functions, but I just added as much as I can. 

my referral1: https://github.com/gpolovets1/jax-llm-examples/blob/v6e_32_prof/deepseek_r1_jax/deepseek_r1_jax/model.py#L684
my referral2: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/rotary_embedding.py#L799

tpu_commons/models/jax/common/rope.py
Chenyaaang: what are "the models in our dashboard"?

tpu_commons/models/jax/common/rope.py
Chenyaaang: If we select as interleaved, are they still called first/second?

gpolovets1: I think your code divides x1 and x2 into even and odd entries respectively while we actually need x1 and x2 to be the first and second halves.

tpu_commons/models/jax/common/rope.py
gpolovets1: Sure, I just mean if you are pretty familiar with the calculations now, please consider adding more explanation. But if some of it you did not spend too much time understanding then we can leave it as is.

tpu_commons/models/jax/common/rope.py
gpolovets1: Chatted offline - Beinuo added our Llama3 implementation to build kite so you should probably point the attention code to this rope implementation and confirm that everything still works as expected: https://github.com/vllm-project/tpu_commons/pull/179
I believe the build kite output will also have accuracy and perf numbers so we should make sure that we don't introduce regressions there.

tpu_commons/models/jax/common/rope.py
Chenyaaang: ok, fixed.

tpu_commons/models/jax/common/rope.py
Chenyaaang: Added in comments.

tests/models/jax/common/test_rope.py
Chenyaaang: Thanks, I hand-calculate simple examples.

tpu_commons/models/jax/common/rope.py
Chenyaaang: This is currently blocked as I still don't have access to initialize a buildkite. Can any of you help me trigger it, or I can wait until I have access.

gpolovets1: Sure, I just triggered it.

tests/models/jax/common/test_rope.py
gpolovets1: Thanks a lot this is very helpful!

PR #186, opened 2025-07-14
[torchax] Fix import failure due to upstream change

lsy323:
`bind_kv_cache` path is changed in https://github.com/vllm-project/vllm/pull/20900. Accommodate the change

### DIFF
--- a/tpu_commons/runner/tpu_torchax_runner.py
+++ b/tpu_commons/runner/tpu_torchax_runner.py
@@ -77,10 +77,10 @@ from vllm.v1.outputs import (EMPTY_MODEL_RUNNER_OUTPUT, LogprobsTensors,
                              ModelRunnerOutput)
 from vllm.v1.sample.tpu.metadata import TPUSupportedSamplingMetadata
 from vllm.v1.sample.tpu.sampler import Sampler as TPUSampler
-from vllm.v1.utils import bind_kv_cache
 from vllm.v1.worker.lora_model_runner_mixin import LoRAModelRunnerMixin
 from vllm.v1.worker.tpu_input_batch import CachedRequestState, InputBatch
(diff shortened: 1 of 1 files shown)


wwl2755-google:
I think the pinned version of vLLM should be changed at the same time? Assuming no other mismatch from current pinned version and the new version?

lsy323:
@wwl2755-google Thanks for the comment, I think we always use vLLM HEAD, at least in the auto benchmarking. Do you happen to know how the vLLM is pinned? Thanks

lsy323:
I'm landing this to bring back the torchax benchmarking workflow

wwl2755-google:
> @wwl2755-google Thanks for the comment, I think we always use vLLM HEAD, at least in the auto benchmarking. Do you happen to know how the vLLM is pinned? Thanks

Okay. I see. All the numbers are zero now because of this. Then it makes sense. Thanks for the quick fix!

Here it is: https://github.com/vllm-project/tpu_commons/blob/main/docker/Dockerfile#L17. I think the pinned version is used in the CI, unless this torchax path has not been covered in CI yet. Let's wait to see if there is any error in the buildkite. 😂

lsy323:
Hi @wwl2755-google, the dashboard is not using this docker image for torchax path, it's always using vLLM main. Also torchax e2e test hasn't been added to CI yet, therefore we didn't see the CI failure

wwl2755-google:
> Hi @wwl2755-google, the dashboard is not using this docker image for torchax path, it's always using vLLM main. Also torchax e2e test hasn't been added to CI yet, therefore we did see the CI failure

Sounds good. Thanks for the clarification!

lsy323:
@wwl2755-google thank you for calling this out! I wasn't aware that we pin vLLM main in CI and docker

PR #188, opened 2025-07-14
Deepseek MLA Implementation 

wenxindongwork:
# Description

This PR adds skeleton implementation for Multi-Head Latent Attention (MLA). It includes MLA Config, new sharding rules, and unit tests to verify functionality. 

### Caveats 

**KV cache**: This PR does not include MLA KV cache implementation. For now,  instead of the latent vector, the full KV tensors are cached. Supporting MLA KV cache requires significant code change and will be supported after the functional phase. 

Moreover, there are limitations in the current kernel design that expects same head dimension for K, V tensors, which is not the case in DeepSeek V3. The current kernel design also expects the head dimension to be a multiple of 128, which is also not the case in DeepSeek V3. Kernel changes are also required to support an efficient implementation. The current implementation is expected to be mathematically correct but inefficient (to be verified). 

### MLA Implementation:
* Added the `MLA` class in `tpu_commons/models/jax/common/attention/attention.py`.

### Configuration Updates:
* Introduced the `MLAConfig` dataclass in `tpu_commons/models/jax/common/attention/attention.py` to define MLA-specific parameters including `q_lora_rank`, `kv_lora_rank`, `qk_nope_head_dim`, and `v_head_dim`.
* Updated `HuggingFaceArgNames` in `tpu_commons/models/jax/common/constants.py` to include MLA-specific attributes.

### Sharding Enhancements:
* Added MLA-specific sharding rules in `ShardingRulesConfig` and `_make_default_sharding_config` in `tpu_commons/models/jax/common/sharding.py`. These rules define tensor layouts for MLA's query and key-value projections. 

### Minor Fixes:
* Updated `maybe_apply_overrides` in `tpu_commons/models/jax/common/base.py` to handle cases where `vllm_config` is `None`.

# Tests

* Added `TestMLA` in `tests/models/jax/common/attention/mla.py` to validate MLA's forward pass. 

# Checklist

- [x] I have reviewed the uLLM modeling code checklist: https://docs.google.com/document/d/1DGQBVvr2bh4G8tBUO1YH8pO7Dd_myw5rfEMfVDymEk8/edit?resourcekey=0-V7MGHu3aQjJH6YrI3-y8Hg&tab=t.t91cyovog2mr#heading=h.cqdzv8mlszca
- [x] I have received at least 1 readability approval and 1 correctness approval.

### DIFF
--- a/tpu_commons/models/jax/common/base.py
+++ b/tpu_commons/models/jax/common/base.py
@@ -68,7 +68,7 @@ class Config:
         If there is overlap in overrides between the configs, then print a warning declaring which
         overrides will take precedent."""
 
-        if not getattr(self, "vllm_config", None):
+        if not getattr(self, "vllm_config"):
             return
 
         def _overrides_str(original: str, original_val: Any,
--- a/tests/models/jax/common/attention/deepseek_v3_attention.py
+++ b/tests/models/jax/common/attention/deepseek_v3_attention.py
@@ -10,7 +10,6 @@ from tpu_commons.models.jax.attention_metadata import AttentionMetadata
 from tpu_commons.models.jax.common.attention.deepseek_v3_attention import (
     MLA, MLAConfig)
 from tpu_commons.models.jax.common.base import ParamFactory
-from tpu_commons.models.jax.common.kv_cache import KVCacheConfig
 from tpu_commons.models.jax.common.sharding import ShardingConfig
 from tpu_commons.models.jax.recipes.deepseek_v3 import (
     DeepSeekV3GenerateShardingRulesConfig,
@@ -59,7 +58,7 @@ class TestMLA(unittest.TestCase):
                 "original_max_position_embeddings": 4096,
                 "type": "yarn",
             },
-        )
+            vllm_config={})
 
         mla = MLA(
             cfg=mla_config,
@@ -74,27 +73,20 @@ class TestMLA(unittest.TestCase):
         seq_len = 32
         x = jnp.ones((seq_len, mla_config.hidden_size), dtype=jnp.bfloat16)
 
-        # Create KV cache configuration and cache
+        # Create KV cache
         # TODO(wenxindongwork): test with unpadded head dimension once
         # MLA kv cache implementation is added.
         qk_head_dim = mla_config.qk_nope_head_dim + mla_config.qk_rope_head_dim
         multiple_of_128 = ((qk_head_dim - 1) // 128 + 1) * 128
         block_size = 16
-        kv_cache_config = KVCacheConfig(
-            batch_size=1,
-            cache_len=128,
-            num_kv_heads=mla_config.num_key_value_heads,
-            head_dim=multiple_of_128,
-            dtype=jnp.bfloat16,
-        )
-        num_blocks = kv_cache_config.cache_len // block_size
+        num_blocks = 8
         cache_shape = (
             num_blocks,
             block_size,
-            kv_cache_config.num_kv_heads * 2,
-            kv_cache_config.head_dim,
+            mla_config.num_key_value_heads * 2,
+            multiple_of_128,
         )
-        kv_cache = jnp.zeros(cache_shape, dtype=kv_cache_config.dtype)
+        kv_cache = jnp.zeros(cache_shape, dtype=jnp.bfloat16)
 
         # Create attention metadata
         num_slices = 8


tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -44,6 +50,39 @@
...
+        (HuggingFaceArgNames.QK_ROPE_HEAD_DIM.value, int),
+        (HuggingFaceArgNames.V_HEAD_DIM.value, int),
+        (HuggingFaceArgNames.RMS_NORM_EPS.value, float),
+        (
+            "dtype",
+            DTypeLike,
+        ),
+        ("vllm_config", VllmConfig, field(repr=False, default=None)),
+    ],
+    bases=(Config,),
```
wenxindongwork: Re potential question on why not inherent `MLAConfig` from `AttentionConfig`:

1. `vllm_config` is an optional param. Optional params are expected to be ordered to the end of the list of params in a dataclass. Inheritance will sandwich the `vllm_config` between non-optional params, which throws an error.  
2. `AttentionConfig` expects the `head_dim` parameter but `MLAConfig` does not. If we want to keep MLAConfig the minimal set, we cannot inherit from AttentionConfig.

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -23,15 +22,22 @@
...
+        (HuggingFaceArgNames.ROPE_SCALING.value, Dict[str, Any]),
+        (
+            "dtype",
+            DTypeLike,
+        ),
+        ("vllm_config", VllmConfig, field(repr=False, default=None)),
+    ],
+    bases=(Config,),
+)
+
```
wenxindongwork: I was noticing formatting differences. How can we ensure that we are using the same Ruff formatter?

tpu_commons/models/jax/common/base.py
```diff
@@ -68,7 +68,7 @@ def maybe_apply_overrides(self):
         If there is overlap in overrides between the configs, then print a warning declaring which
         overrides will take precedent."""
 
-        if not getattr(self, "vllm_config"):
+        if not getattr(self, "vllm_config", None):
```
wenxindongwork: I was getting the error "this class does not have the `vllm_config` attribute", hence patching this line.

tpu_commons/models/jax/common/attention/attention.py
gpolovets1: I think when you run with pre-submit checks (tpu_commons/.pre-commit-config.yaml) it should be running ruff but feel free to double check as well.

tpu_commons/models/jax/common/attention/attention.py
gpolovets1: 1. You can use kw_only=True to avoid the optional arguments issue.
2. Don't you still have to project the K & V dimensions back to the original head dimension with MLA?

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -293,3 +337,263 @@ def _ragged_paged_attention(*args):
...
+        x_q_TD = nnx.with_sharding_constraint(x, self.activation_q_td[op_mode])
+
+        with jax.named_scope("q_proj"):
+            # Query down projection.
+            q_TA = jnp.einsum("TD,DA -> TA", x_q_TD, self.kernel_q_down_proj_DA.value)
+            q_TA = self.q_rms_norm(q_TA)
+            # Query up projection.
+            q_TNH = jnp.einsum("TA,ANH -> TNH", q_TA, self.kernel_q_up_proj_ANH.value)
+            # Split the query into nope and rope.
+            q_nope_TNH = q_TNH[..., : self.qk_nope_head_dim]
```
gpolovets1: I think this may need to be done over q_TA (down-projected query).

gpolovets1: I think if you wanted to do it this way your kernel_q_down_proj_DA would need to be (self.D, kernel_q_down_proj_DA + qk_rope_head_dim)

tpu_commons/models/jax/common/attention/attention.py
wenxindongwork: 1. thanks! that's helpful to know
2. There is no `head_dim` in the MLA config. The head dim of K and Q is determined by `qk_rope_dim + qk_nope_dim` and head dim of V is determined by `v_head_dim`.

tpu_commons/models/jax/common/attention/attention.py
wenxindongwork: do you mean the splitting part? I think the splitting should be done on the up projected query (q_TNH), otherwise the dimension will not match. 

Here is the reference implementation: https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/model.py#L461-L463

wenxindongwork: I think 
kernel_q_down_proj should be (self.D, q_lora_rank) and 
kernel_q_up_proj should be (q_lora_rank, qk_rope_head_dim +qk_nope_head_dim)

ref:  https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/model.py#L424-L426

gpolovets1: Yes but I think in your code you are doing the splitting on the up-projected query. I think splitting should be done on the down-projected query?

gpolovets1: Ah I see my confusion now - the comment `# Query up projection.` is a little bit misleading. This matmul is used to both up-project the query but also perform the RoPE projection. Makes sense now!

tests/models/jax/common/attention/mla.py
```diff
@@ -0,0 +1,114 @@
...
+        mla_config = MLAConfig(
+            hidden_size=256,
+            num_attention_heads=32,
+            num_key_value_heads=32,
+            rope_theta=10000,
+            dtype=jnp.bfloat16,
+            q_lora_rank=512,
+            kv_lora_rank=512,
+            qk_nope_head_dim=64,  # Half of DeepSeek v3's real values
+            qk_rope_head_dim=32,  # Half of DeepSeek v3's real values
```
gpolovets1: What do you mean in your comment?

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -293,3 +337,263 @@ def _ragged_paged_attention(*args):
...
+        self.kernel_q_down_proj_DA = self.param_factory.create_kernel_param(
+            rngs, (self.D, self.query_lora_rank), self.q_da_sharding, self.cfg.dtype
+        )
+        self.kernel_q_up_proj_ANH = self.param_factory.create_kernel_param(
+            rngs,
+            (self.query_lora_rank, self.N, self.qk_head_dim),
+            self.anh_sharding,
+            self.cfg.dtype,
+        )
+        self.kernel_kv_down_proj_DA = self.param_factory.create_kernel_param(
```
gpolovets1: Is this actually self.kernel_k_down_proj_DA? It's just used for the keys and could help differentiate from kernel_kv_up_proj_ANH which will have 2x final dimension.

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -293,3 +337,263 @@ def _ragged_paged_attention(*args):
...
+        with jax.named_scope("attn_op"):
+            # TODO(wenxindongwork): K and V have different head dimension,
+            # which is not supported by the current kv cache implementation.
+            # For now we are padding the v dimension to match the k dimension.
+            # Furthermore, deepseekv3 k head dimension is 192, which is
+            # not supported by the current attention kernel, which expects
+            # q, k, v head dimension to be multiple of 128. For now, we will
+            # pad the q, k, v dimension to multiple of 128.
+            # We should update the MLA kv cache implementation in the future.
+            multiple_of_128 = ((self.qk_head_dim - 1) // 128 + 1) * 128
```
gpolovets1: Thanks for the rich commentary!

tpu_commons/models/jax/common/sharding.py
gpolovets1, line 1: I think these are actually better suited in recipes/deepseek_v3.py. You can create a custom ShardingRulesConfig where you specify this.
Maybe you can make a placeholder recipes/deepseek_v3.py where you add this section? If you need to add separate prefill/decode shardings, you can consider adding the overrides in a _post_init or something.

tpu_commons/models/jax/common/sharding.py
```diff
@@ -351,6 +361,10 @@ def _make_default_sharding_config(self, prefill_rules, generate_rules):
                                             ATTN_TENSOR_AXIS_NAME)
         generate_rules.attn_o_weight_nhd = (None, ATTN_HEAD_AXIS_NAME,
                                             ATTN_TENSOR_AXIS_NAME)
+        # MLA Query up projection weight: (QueryLoraRank, NumHeads, HeadDim)
+        generate_rules.attn_mla_qb_weight_anh: tuple = (None, ATTN_HEAD_AXIS_NAME, ATTN_TENSOR_AXIS_NAME)
```
gpolovets1: Do we need to shard these given they are already relatively small matrices (due to QueryLoraRank being fairly small)?

tpu_commons/models/jax/common/attention/attention.py
gpolovets1, line 1: You may want to move this code to deepseek_v3_attention.py - similar to what you did with MoE.

tests/models/jax/common/attention/mla.py
wenxindongwork: DeepSeek v3 uses qk_nope_head_dim=128 and qk_rope_head_dim=64

tpu_commons/models/jax/common/attention/attention.py
wenxindongwork: I think it should be `kernel_kv_down_proj_DA` because the latent vector resulting from this kernel is used to produce `k_rope`, `k_nope` and `v` later on.

tpu_commons/models/jax/common/attention/attention.py
wenxindongwork: sure!

tpu_commons/models/jax/common/sharding.py
wenxindongwork: I'm not sure, query lora rank is 1536 in v3. and num_heads is 128, and head dim is 192. 

this is (1536, 128, 192) = 72mb in bf16. Across 61 layers it will be 4392mb. 

should we shard just in case? if perf is not good we can get rid of this line.

tpu_commons/models/jax/common/sharding.py
wenxindongwork: sure, I also see the comment
```
#TODO split this into block unique sharding config, i.e. attentionShardingConfig, MoEShardingConfig
``` 
that makes sense!

wenxindongwork: attempted to make this change by creating `DeepSeekV3ShardingRulesConfig`, `DeepSeekV3PrefillShardingRulesConfig` and `DeepSeekV3GenerateShardingRulesConfig`, intended usage is

```
ShardingConfig(
            default_rules_cls=DeepSeekV3ShardingRulesConfig,
            prefill_rules=DeepSeekV3PrefillShardingRulesConfig,
            generate_rules=DeepSeekV3GenerateShardingRulesConfig,
        )
```

please lmk if this approach makes sense to you!

tests/models/jax/common/test_rope.py
wenxindongwork, line 1: I think Github is broken. It attaches files from the latest commits from main to this PR. I did not create this file.

tpu_commons/models/jax/common/rope.py
wenxindongwork, line 1: I think Github is broken. It attaches files from the latest commits from main to this PR. I did not create this file.

I did make some changes to this file though. I added a `dtype` arg to the RoPE classes.

tpu_commons/models/jax/recipes/deepseek_v3.py
gpolovets1, line 1: Thanks WenXin for adding this! Typically Prefill/Generate ShardingRules are instance variables of ShardingConfig, so we may update this in the future to set generate/prefill rules in the ShardingConfig or Sharding post_init but we can play around with this a bit more.

tpu_commons/models/jax/common/sharding.py
gpolovets1: I see, thanks for walking through the calculations! 4GB seems pretty large so we can shard for now and revisit based on the xprof

tpu_commons/models/jax/common/base.py
gpolovets1: I'm a bit concerned about defaulting this to None. This file defines our interface so we should try not to compromise on this. This config is also what's necessary to override the recipe file settings.
So far passing vllm_config has worked for all of the previous layer implementations

wenxindongwork: culprit was KVCacheConfig which is no longer in use. Fixed unit test and this patch is no longer needed.

PR #194, opened 2025-07-15
Clean up platforms/tpu_jax

xiangxu-google:
no-op

### DIFF
--- a/tpu_commons/platforms/tpu_jax.py
+++ b/tpu_commons/platforms/tpu_jax.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
-from typing import TYPE_CHECKING, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Optional, Tuple, Union, cast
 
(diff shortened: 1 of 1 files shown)


tpu_commons/platforms/tpu_jax.py
```diff
@@ -96,15 +96,17 @@ def inference_mode(cls):
...
     @classmethod
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
+        if not envs.VLLM_USE_V1:
+            raise RuntimeError("VLLM_USE_V1=1 must be set for JAX backend.")
+
         from vllm.config import CompilationLevel
 
         cache_config = vllm_config.cache_config
-        # For TPU, the default block size is 32
+        # For v0, the default block size is 16.
```
wwl2755-google, line 105: nit: Maybe we don't need this part as vLLM V1 is required already?

xiangxu-google: this is copied from vllm and I added an assert above for V1

PR #205, opened 2025-07-16
Add Qwix quantization + per-tensor KV cache quantization

jrplatin:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,43 +226,37 @@ being sure to pass the proper args for your machine.
 You might need to run the benchmark client *twice* to make sure all compilations are cached server-side.
 
 ## Quantization
-Currently, we support both per-tensor KV cache quantization and overall model/activation quantization through the [Qwix](https://github.com/google/qwix?tab=readme-ov-file#quantization-config) framework.  NOTE: these are only supported using the `NEW_MODEL_DESIGN`!
+### Overview
+Currently, we support both per-tensor KV cache quantization (int8, int4, and fp8 (`float8_e4m3fn)) and overall model weight/activation quantization through the [Qwix](https://github.com/google/qwix?tab=readme-ov-file#quantization-config) framework.
 
-### KV cache quantization
-In order to enable KV cache quantization, simply add the following to your base command:
-
-```
-... --additional_config='{"quantization": {"kv_cache_quant_dtype": "int8"}}'
-```
-
-We currently support int8, int4, and FP8 (`float8_e4m3fn`) quantization dtypes.
-
-### Model/activation quantization
-In order to enable overall model quantization, simply add the following to your base command:
+To enable quantization, you can specify a quantization config filename found inside the quantization config directory (`tpu_commons/models/jax/utils/quantization/configs/`), for example:
 
 ```
-.... --additional_config='{"quantization": {"dtype": "int8"}}'
+... --additional_config='{"quantization": "int8_default.yaml"}'
 ```
 
-By default, we will use the following Qwix rules (with the given `dtype`), which will quantize attention weights-only and MLP with weights and activations:
+### Creating your own quantization config
+To create your own quantization:
+
+1. Add a new file to the quantization config irectory (`tpu_commons/models/jax/utils/quantization/configs/`)
+2. For Qwix quantization, add a new entry to the file as follows:
 
 ```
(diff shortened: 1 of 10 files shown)


tpu_commons/kernels/ragged_paged_attention/kernel.py
```diff
@@ -602,14 +615,13 @@ def prefetch_next_kv_blk():
...
-                        # That's why it is converted to float32 first.
-                        k = k.astype(jnp.float32) * k_scale
-                        k = k.astype(q_ref.dtype)
-                    if v_scale is not None:
-                        v = v.astype(jnp.float32) * v_scale
-                        v = v.astype(q_ref.dtype)
+
+                    k_scale, v_scale = None, None
+                    if k_scale_ref is not None:
+                        k_scale = k_scale_ref[0]
```
jrplatin, line 621: TODO @jrplatin: do we want to keep this `astype`?

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -230,35 +231,23 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
-        for _ in layer_names:
-            self.kv_caches.append(sharded_allocate())
+        maybe_kv_cache_quant_dtype = self.vllm_config.additional_config.get(
+            "quantization", {}).get("kv_cache_quant_dtype", None)
+
+        self.kv_caches = create_kv_caches(
+            num_blocks=kv_cache_config.num_blocks,
+            block_size=kv_cache_spec.block_size,
+            num_kv_heads=kv_cache_spec.
+            num_kv_heads,  # NOTE: we'll multiply by 2 in the function
```
vipannalla, line 248: nit: can you make this into single like? The current split is confusing and if the comment is the reason for split, we you add in a separate like above?

README.md
```diff
@@ -224,3 +224,45 @@ While this will run the code in a Docker image, you can also run the bare `tests
...
+    qwix.QuantizationRule(
+        module_path='.*mlp.*',
+        weight_qtype=quant_dtype,
+        act_qtype=quant_dtype,
+        tile_size=None,
+    ),
+]
+```
+
+You may also create a file that defines your own rules (e.g. `tpu_commons/models/jax/utils/quantization/quantize_all_modules_int8_wa.yaml`), where each entry under `rules` corresponds to a `qwix.QuantizationRule`.  To pass this file (which is mutually exclusive with `quantization.dtype`), you can something similar to:
```
mitalisi: Why not make a standard dir for quantization configs and ask user to add config to that dir. Can just pass the file name. We can store kv cache quantization config in the same file.

--additional_config='{"quantization": "int8_default.yaml"}'

And for a custom file
--additional_config='{"quantization": "int8_default_int8_kv.yaml"}'


Files can be in tpu_commons/models/jax/utils/quantization/configs/

We will likely have different configs checked in to this dir for different models/datasets.

README.md
```diff
@@ -224,3 +224,45 @@ While this will run the code in a Docker image, you can also run the bare `tests
...
+### Model/activation quantization
+In order to enable overall model quantization, simply add the following to your base command:
+
+```
+.... --additional_config='{"quantization": {"dtype": "int8"}}'
+```
+
+By default, we will use the following Qwix rules (with the given `dtype`), which will quantize attention weights-only and MLP with weights and activations:
+
+```
```
mitalisi: Better to add path to file here so that it remains up to date

README.md
jrplatin: Agreed on making a standardized dir, but I think keeping KV cache / model quant separate makes more sense since we really only need the YAML for the Qwix rules, so I think it keeps the code / UX much cleaner if separate the two quants out (since the KV quant is only a dtype specification) -- can iterate more on this in the KV cache quant PR

mitalisi: Kv cache config also had multiple options - dtype, per_tensor/per_token/dimension_to_quantize_on, diff conf for key/value. We will only support a simple per_tensor int8 initially but keeping it in file keeps it flexible. Also one config for all quant is easier to read.

PR #216, opened 2025-07-17
Skipping weight initialization when loading weights.

gpolovets1:
# Description
Avoiding random weight initialization when loading saved weights to prevent OOM issues.

Build kite successfully passed: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/676

### DIFF
--- a/tpu_commons/models/jax/recipes/llama3.py
+++ b/tpu_commons/models/jax/recipes/llama3.py
@@ -108,7 +108,8 @@ class Llama3_8B(Model):
                                  mesh=self.mesh,
                                  default_rules_cls=Llama8BShardingRulesConfig,
                                  vllm_config=self.vllm_config)
-
+        self.use_random_init = self.vllm_config.additional_config.get(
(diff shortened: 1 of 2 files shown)


tpu_commons/models/jax/recipes/llama3.py
```diff
@@ -108,7 +108,8 @@ def __init__(self, vllm_config: VllmConfig, rng: PRNGKey, mesh: Mesh):
                                  mesh=self.mesh,
                                  default_rules_cls=Llama8BShardingRulesConfig,
                                  vllm_config=self.vllm_config)
-
+        self.use_random_init = self.vllm_config.additional_config.get(
```
jrplatin, line 111: Is there a way to push this logic into the `ParamFactory` so you don't need to duplicate this code for every model?

gpolovets1: This is just passing the runtime param to ParamFactory (so mostly for readability).

PR #219, opened 2025-07-17
Increase the timeout to get from queue in test.

QiliangCui:
# Description

The [test](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/679#01981a2c-9c5d-4288-b8b5-69639dae0a68) still shows failure but it is a timeout when getting from queue. 

I can repro this problem if I insert some sleep in orchestrator code.   
Otherrwise, it is hard to repro. 

Increase the timeout to 30. If it still happens, a deeper investigation is needed to see if there is some race condition that causes deadlock. 

# Tests

See CI. 

# Checklist

### DIFF
--- a/tests/core/test_orchestrator.py
+++ b/tests/core/test_orchestrator.py
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
-import queue
 import threading
 import time
 import unittest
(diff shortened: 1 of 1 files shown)


tests/core/test_orchestrator.py
```diff
@@ -125,26 +131,26 @@ def test_disaggregated_mode_lifecycle(self):
             driver.place_request_on_prefill_queue(mock_req)
 
             # Get prefill output
-            prefill_output = driver._vllm_output_backlogs.get(timeout=5)
+            prefill_output = driver._vllm_output_backlogs.get(timeout=30)
```
QiliangCui, line 134: The only changes I made are the timeout. Others were made by "pre-commit" automatically.

PR #288, opened 2025-07-22
Make HostInterface optional until full DI is in.

yarongmu-google:
# Tests

```
pytest
```

# Checklist

- [X ] I have performed a self-review of my code.
- [X ] I have necessary comments in my code, particularly in hard-to-understand areas.
- [X ] I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/tpu_commons/backend.py
+++ b/tpu_commons/backend.py
@@ -13,6 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from typing import Optional
+
 from tpu_commons.adapters.vllm_adapters import (VllmLoRARequestAdapter,
                                                 VllmSchedulerOutputAdapter)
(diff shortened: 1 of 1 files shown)


tpu_commons/backend.py
```diff
@@ -28,13 +30,15 @@ class TPUBackend(BackendInterface):
   and managing the concrete TPU worker instance and delegating calls to it.
   """
 
-    def __init__(self, host_interface: HostInterface, **worker_kwargs):
+    def __init__(self,
+                 host_interface: Optional[HostInterface] = None,
```
yaochengji, line 34: Do we need to add a unit test when host_interface is None?

yarongmu-google: I wouldn't worry about it ... it is there because a vllm PR needs to be submitted soon. Once that one is on, this PR will be reverted :)

But I can if you insist - WDYT?

yaochengji: I don't think the test is needed if it will be reverted soon. Thanks!

PR #310, opened 2025-07-24
Add input/output length options to MLPerf / MMLU datasets + add JAX server-based profiling

jrplatin:
# Description

* Add `input` / `output` lengths arguments to the MLPerf/MMLU datasets
* Add ability to start a JAX profiling server for the JAX path.  To use it, simply set the following env var: `USE_JAX_PROFILER_SERVER=True` .  You can also optionally set the `JAX_PROFILER_SERVER_PORT`.
    * This is useful in the case where you'd like to profile a cross section of a full E2E benchmark run with many prompts (instead of using the old method of setting `VLLM_TORCH_PROFILER_DIR` and running benchmarking with a single prompt).

In order to capture (cross-sectional) profiles:

1. Run your typical `vllm serve` or `offline_inference` cmd (making sure to set `USE_JAX_PROFILER_SERVER=True)
2. Run your benchmarking command (`python benchmark_serving.py...`)
3. Once the warmup has completed and your benchmark is running, start a new tensorboard instance with your `logdir` set to the desired output location of your logs (e.g. `tensorboard --logdir=profiles/llama3-mmlu/`
4. Open the tensorboard instance and navigate to the `profile` page (e.g. `http://localhost:6006/#profile`)
5. Click `Capture Profile` and, in the `Profile Service URL(s) or TPU name` box, enter `localhost:XXXX` where `XXXX` is your `JAX_PROFILER_SERVER_PORT` (default is `9999`):
image

6. Enter the desired amount of time you'd like to capture the profile for and then click `Capture`.  If everything goes smoothly, you should see the following, with your `logdir` being populated:
image

# Checklist

### DIFF
--- a/scripts/vllm/benchmarking/benchmark_dataset.py
+++ b/scripts/vllm/benchmarking/benchmark_dataset.py
@@ -288,7 +288,7 @@ class MMLUDataset(BenchmarkDataset):
             #     skip_min_output_len_check=output_len is not None,
             # ):
             #     continue
-            if input_len <= prompt_len:
+            if input_len is not None and input_len <= prompt_len:
                 raise ValueError(
                     f"prompt is too short: prompt_len is {prompt_len} but input_len is {input_len}"
                 )
@@ -359,7 +359,7 @@ class MLPerfDataset(BenchmarkDataset):
                 completion_ids) if output_len is None else output_len
             # NOTE (jacobplatin): I don't believe that we filter the MLPerf dataset
             # at all, but it could be done here
-            if input_len <= prompt_len:
+            if input_len is not None and input_len <= prompt_len:
                 raise ValueError(
                     f"prompt is too short: prompt_len is {prompt_len} but input_len is {input_len}"
                 )
--- a/README.md
+++ b/README.md
@@ -144,12 +144,6 @@ MODEL_IMPL_TYPE=flax_nnx
 MODEL_IMPL_TYPE=vllm
 ```
 
-To enable profiling:
-
-```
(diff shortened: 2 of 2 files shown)


scripts/vllm/benchmarking/benchmark_dataset.py
```diff
@@ -287,11 +288,15 @@ def sample(
             #     skip_min_output_len_check=output_len is not None,
             # ):
             #     continue
+            if input_len <= prompt_len:
```
bzgoogle: may i know what does input_len refer to?  prompt_len is the # tokens in a prompt, what about input_len?

tpu_commons/worker/tpu_worker_jax.py
```diff
@@ -72,6 +72,15 @@ def __init__(self,
 
         logger.info(f"Using devices: {self.devices}")
 
+        use_jax_profiler_server = os.getenv("USE_JAX_PROFILER_SERVER", False)
```
bzgoogle, line 75: is the new change still compatible with the old way of profiling? i.e. add --profile and it will dump to the pointer given to vllm serve

scripts/vllm/benchmarking/benchmark_dataset.py
jrplatin: It's the `prompt_len` yep, I just kept it as a `input_len` since that's the vLLM convention

tpu_commons/worker/tpu_worker_jax.py
jrplatin: Yep, the old way will still work!

PR #318, opened 2025-07-25
[vllm model] Process MergedParallelLinear during model partitioning

lsy323:
# Description

We implemented a SPMD friendly version of MergedParallelLinear and do module swap after model init and weight loading. However, the MergedParallelLinear is missing from the replacement list. This PR adds MergedParallelLinear to the list.

From benchmarking, the throughput of llama 70B on 1800 : 128 in/out ratio improved from ~5.3 req/s to ~6.6 req/s

# Tests

Added a test.

# Checklist

### DIFF
--- a/tests/models/vllm/test_torchax_tpu_loader.py
+++ b/tests/models/vllm/test_torchax_tpu_loader.py
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
-import gc
 import tempfile
 
 import jax
@@ -88,9 +87,6 @@ def test_vllm_model_loader(model):
     for v in params.values():
         v.delete()
 
-    del model
-    gc.collect()
-
 
 @pytest.mark.parametrize("model", [
     "Qwen/Qwen2-1.5B-Instruct",
@@ -122,6 +118,3 @@ def test_tpu_model_loader(model):
     for v in params_buffers.values():
         if isinstance(v, torchax.tensor.Tensor):
             v._elem.delete()
-
-    del model
-    gc.collect()


lsy323:
A weird observation is we will hit TPU HBM OOM in the test, where we load 2 Qwen 1.5B and 2 Llama 8B models on TPU. `del model` and `gc.collect` are called after each test. But still hit OOM.

Later I figured we need to manually call `jax.Array.delete()` on the loaded weight tensors to free the TPU memory.

Triggered a CI run here
https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1109

tests/models/vllm/test_torchax_tpu_loader.py
```diff
@@ -44,6 +52,46 @@ def _setup_environment(model):
...
+            assert fqn not in JAX_REPLACED_MODULE_NAMES, \
+                f"Module {fqn} should be replaced by JAX version, " \
+                "please check the TPU backend configuration."
+
+    # Need to manually delete the JAX arrays otherwise they will stay on TPU.
+    for v in params.values():
+        v.delete()
+
+    del model
+    gc.collect()
```
hfan: is gc.collect() still needed?

lsy323: Not needed now, I kept it there since it's suppose to work without the workaround. I can delete it

tests/models/vllm/test_torchax_tpu_loader.py
```diff
@@ -62,8 +110,18 @@ def test_tpu_model_loader(model):
...
+
+    # Need to manually delete the JAX arrays otherwise they will stay on TPU.
+    params, buffers = extract_all_buffers(model)
+    params_buffers = {**params, **buffers}
+    for v in params_buffers.values():
+        if isinstance(v, torchax.tensor.Tensor):
+            v._elem.delete()
+
+    del model
+    gc.collect()
```
hfan: ditto

tests/models/vllm/test_torchax_tpu_loader.py
hfan: If we don't need it, cleaning it up to the minimum would be nice

lsy323: Done

tests/models/vllm/test_torchax_tpu_loader.py
lsy323: Done

lsy323:
Trigger another CI run before merging https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1129

PR #319, opened 2025-07-25
[JAX] Disable sanity check for scheduler output.

py4:
# Description

There is no guarantee that vLLM is not sending empty schedule. One example is due to timing effect (FSM used for structured decoding is getting compiled so request cannot be scheduled) [here](https://github.com/vllm-project/vllm/blob/7cfea0df390c154c1026f77d3682e2733ca4aca8/vllm/v1/core/sched/scheduler.py#L352).
[Here](https://github.com/vllm-project/vllm/blob/7cfea0df390c154c1026f77d3682e2733ca4aca8/vllm/v1/engine/core.py#L275) vLLM doesn't care if there is output in scheduler or not. 
So we cannot have that sanity check in our code. I don't know why vLLM is not preventing call to worker when there is no schedule but If we want the guarantee, this should be done in vLLM core.

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -10,20 +10,10 @@ import jax
 import jax.numpy as jnp
 import jaxtyping
 import numpy as np
-import vllm.envs as envs
 from flax import nnx
 from jax.sharding import NamedSharding, PartitionSpec
-from vllm.config import VllmConfig
-from vllm.distributed.kv_transfer import (get_kv_transfer_group,
-                                          has_kv_transfer_group)
-from vllm.sequence import IntermediateTensors
-from vllm.utils import cdiv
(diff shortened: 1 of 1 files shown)


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -599,9 +599,11 @@ def _execute_model(
             # Return empty ModelRunnerOutput if there's no work to do.
             # TODO(fhzhang): We rely on empty cycles to remove requests in input batch. Fix it to reduce overhead.
             logger.debug(f"Nothing scheduled: {scheduler_output}!")
-            if len(scheduler_output.finished_req_ids) == 0:
-                raise Exception(
```
fenghuizhang, line 603: Can you add a warning here instead of commenting out the whole block? Thanks.

py4: Done. tnx

PR #321, opened 2025-07-25
[Kernel] support dynamic_validate_inputs for kv_cache_update

yaochengji:
# Description

Added the `dynamic_validate_inputs` argument in the kv_cache_update kernel.

# Tests
```
pytest -s -v tests/ragged_kv_cache_update_test.py -k test_basic
```

# Checklist

### DIFF
--- a/tests/ragged_kv_cache_update_test.py
+++ b/tests/ragged_kv_cache_update_test.py
@@ -121,3 +121,121 @@ class KVCacheUpdateTest(jtu.JaxTestCase):
                             updated_kv_cache_ref,
                             atol=1e-4,
                             rtol=1e-4)
+
+    def test_invalid_inputs(self):
+        # Test all the cases when the inputs are invalid in the `_dynamic_validate_inputs` method
+        page_size = 32
+        combined_kv_head_num = 2
+        head_dim = 128
+        num_slices_per_block = 4
+
+        new_kv, slot_mapping, kv_cache, num_slices = self._generate_data(
+            page_size, combined_kv_head_num, head_dim, num_slices_per_block)
+
+        with jax.disable_jit():
+            # Case 1: new_kv_start < 0
+            invalid_slot_mapping = slot_mapping.at[1, 0].set(-1)
+            with self.assertRaisesRegex(
+                    ValueError, "new_kv_start=-1 must be greater than"):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 2: kv_cache_start < 0
+            invalid_slot_mapping = slot_mapping.at[0, 0].set(-1)
+            with self.assertRaisesRegex(
+                    ValueError, "kv_cache_start=-1 must be greater than"):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 3: slice_len <= 0
+            invalid_slot_mapping = slot_mapping.at[2, 0].set(0)
+            with self.assertRaisesRegex(
+                    ValueError, "slice_len=0 must be less or equal to"):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 4: slice_len > page_size
+            invalid_slot_mapping = slot_mapping.at[2, 0].set(page_size + 1)
+            with self.assertRaisesRegex(
+                    ValueError,
+                    f"slice_len={page_size + 1} must be less or equal to"):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 5: new_kv_start + slice_len > new_token_num
+            invalid_slot_mapping = slot_mapping.at[1, 0].set(new_kv.shape[0])
+            with self.assertRaisesRegex(
+                    ValueError,
+                    "new_kv_start=128 \+ slice_len=7 must be less or equal to new_token_num=128"
+            ):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 6: kv_cache_start + slice_len > kv_cache_token_num
+            invalid_slot_mapping = slot_mapping.at[0, 0].set(kv_cache.shape[0])
+            with self.assertRaisesRegex(
+                    ValueError,
+                    "kv_cache_start=640 \+ slice_len=7 must be less or equal to kv_cache_token_num=640"
+            ):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
+                                dynamic_validate_inputs=True)
+
+            # Case 7: Each slice must reside in the same page
+            invalid_slot_mapping = slot_mapping.at[0, 0].set(page_size - 1)
+            invalid_slot_mapping = invalid_slot_mapping.at[2, 0].set(page_size)
+            with self.assertRaisesRegex(
+                    ValueError, "Each slice must reside in the same page"):
+                kv_cache_update(new_kv,
+                                invalid_slot_mapping,
+                                kv_cache,
+                                num_slices,
+                                page_size=page_size,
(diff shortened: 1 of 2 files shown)


yaochengji:
@bythew3i could you take a look?

@sixiang-google , you might enable this check when debugging.

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -67,6 +67,23 @@ def _kv_cache_update_kernel(
...
+        kv_cache_start = slices[0, i]
+        new_kv_start = slices[1, i]
+        slice_len = slices[2, i]
+        if kv_cache_start >= kv_cache_token_num:
+            raise ValueError(
+                f"{kv_cache_start=} must be less than {kv_cache_token_num=}")
+        if new_kv_start >= new_token_num:
+            raise ValueError(
+                f"{new_kv_start=} must be less than {new_token_num=}")
+        if slice_len > page_size:
```
bythew3i: I would suggest using this pattern:

if not (0 < slice_len <= page_size)... so you can cover all the OOB check.

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -67,6 +67,23 @@ def _kv_cache_update_kernel(
         async_copy.wait()
 
 
+def _dynamic_validate_inputs(slices, new_token_num, kv_cache_token_num,
+                             page_size):
+    for i in range(slices.shape[1]):
+        kv_cache_start = slices[0, i]
+        new_kv_start = slices[1, i]
+        slice_len = slices[2, i]
+        if kv_cache_start >= kv_cache_token_num:
```
bythew3i: You may also want to check these **invalid** cases:

- `kv_cache_start < 0`
- `kv_cache_start  + slice_len >= kv_cache_token_num`
- `new_kv_start < 0`
- `new_kv_start + slice_len >= new_token_num`

Also I think you may need check this **invalid** case (to make sure the slice update is within same page):
- `new_kv_start // page_size != (new_kv_start + slice_len) // page_size`
- `kv_cache_start // page_size != (kv_cache_start + slice_len) // page_size`

yaochengji: Great suggestions, Jevin! Do you think we need to make sure the token indices to be updated needs to cover the front part of the new_kv_cache tensor with no bubbles and no overlap? Although I don't think it will trigger crash in the kernel.

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: Thanks Jevin! I think it should be `0 <= slice_len <= page_size` because the padding's slice_len is 0.

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -67,6 +67,23 @@ def _kv_cache_update_kernel(
         async_copy.wait()
 
 
+def _dynamic_validate_inputs(slices, new_token_num, kv_cache_token_num,
+                             page_size):
+    for i in range(slices.shape[1]):
```
bythew3i: If it is dynamic check we can just use `num_slices`?

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: IIUC, if we iterate through `num_slices` instead of `slices.shape[1]` we won't check padding.

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: Did you mean checking if there is any overlap between slices?

yaochengji:
> Did you mean checking if there is any overlap between slices?

Yeah, no overlap and only one bubble in the end.

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: > `new_kv_start // page_size != (new_kv_start + slice_len) // page_size`

I'm afraid this condition might not be satisfied even in a valid case. E.g. page_size = 16, 
slice0: len = 3
slice1: len = 16

yaochengji: > `kv_cache_start // page_size != (kv_cache_start + slice_len) // page_size`

should be `kv_cache_start // page_size != (kv_cache_start + slice_len - 1) // page_size`?

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: I think we'd better also check the padding part because it will also go into the kernel, WDYT?

bythew3i: IMO, we should not care about the padding. Because  I don't think we use the padding in the kernel.

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: > should be kv_cache_start // page_size != (kv_cache_start + slice_len - 1) // page_size?

Yes!

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: We will, even with dynamic grid.
E.g. num_slice_per_block=8, actual_slices_num=5, and we padded the slices_num to 16. There're 3 padding slices used in the kernel.

tests/ragged_kv_cache_update_test.py
```diff
@@ -60,20 +60,24 @@ def _generate_data(self, page_size, combined_kv_head_num, head_dim,
...
         old_kv_cache_copy = kv_cache.copy()
 
-        updated_kv_cache = kv_cache_update(
-            new_kv,
-            slot_mapping,
-            kv_cache,
-            num_slices,
-            page_size=page_size,
-            num_slices_per_block=num_slices_per_block)
+        with jax.disable_jit(disable=dynamic_validate_inputs):
```
vanbasten23, line 72: I wonder why in the service code we don't disable the JIT.

vanbasten23, line 72: I assume the use case is that by default we set dynamic_validate_inputs to False. When there is a bug, then we set it to True, right?

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: Ah I see, you did not guard that using either dynamic for loop or if-else branch. Then I think it is really dangerous, the kernel assumes padding are zero. If non-zero, it is actually going to emit DMA or some OOB access (eg, stupid large number in padding). I would suggest changing the python for to lax.for_i to make sure only effective DMA is issued.

yaochengji: I tried to use `lax.for_i ` before, but it will trigger a jax trace error.

Currently, all the padding slice has slice_len=0, do you still think it's dangerous? Or we can add the check that when `i >= slice_num`, slice_len[i] = 0.

tests/ragged_kv_cache_update_test.py
yaochengji: Yes

yaochengji: Because jax.jit is already annotated in the `kv_cache_update` method.

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: > Currently, all the padding slice has slice_len=0

You can only guarantee that on call site. But kernel just assume 0. You can do sth similar to https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/kernels/ragged_paged_attention/kernel.py#L53 while also make the size to 0. In this way the padded start index is guranteed to be in valid range instead of assuming user has padded to 0 already.

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: > kv_cache_start  + slice_len >= kv_cache_token_num
> new_kv_start + slice_len >= new_token_num

Should be `kv_cache_start  + slice_len > kv_cache_token_num` and `new_kv_start + slice_len > new_token_num`?

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: Thanks for the suggestion! I remembered during the implementation of kernel, I was afraid it will trigger some overhead and that's why I left the input validation to the kernel users.

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i: yes, right!

yaochengji:
@bythew3i @vanbasten23 

Thanks for your review. I think I already added all the checks @bythew3i you mentioned and additional checks to make sure the slices in new_kv is continuous, no overlap in kv cache slices.

A new test is also added to make sure the appropriate exceptions are raised.

Could you take another look?

PR #323, opened 2025-07-25
fix insert logic in decode

sixiang-google:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_commons/core/core_tpu.py
+++ b/tpu_commons/core/core_tpu.py
@@ -7,6 +7,7 @@ import threading
 import time
 import traceback
 import signal
+import math
 from typing import Any, Callable, Optional, TypeVar, Union
 
 import jax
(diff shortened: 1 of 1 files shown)


tpu_commons/core/core_tpu.py
```diff
@@ -423,12 +423,17 @@ def _decode(self, idx: int):
                 # Insert the request to the decoder.
                 req_id = prefill_output["req_id"]
                 vllm_request = self._requests[req_id]
+                # Caching num_computed_tokens. The tokens in kv manager allocate blocks 
+                # is computed as num_computed_tokens + num_new_tokens, so without caching
+                # the token number would double. 
+                prompt_tokens = vllm_request.num_computed_tokens
```
fenghuizhang, line 430: Tweak tests to reflect the change to avoid regression in the future?

sixiang-google: This is hard to check in the mock settings in ut. I added an assert in the main logic for checking this.

PR #334, opened 2025-07-29
Remove usage of torchax `jax_device`

lsy323:
# Description

We had `jax_device('tpu')` in the code, because at some point `.to('jax')` always transfer data to CPU devices. In `torchax==0.0.5` and HEAD, `.to('jax')` will transfer to TPU by default. Therefore we don't need it anymore

Also, `torchax.jax_device` is moved to another path at HEAD, the current code would fail in our next torchax pin update. Let's just remove the usage of this API then.

# Tests

CI, offline benchmarking

# Checklist

### DIFF
--- a/tpu_commons/models/torchax/tpu.py
+++ b/tpu_commons/models/torchax/tpu.py
@@ -6,7 +6,6 @@ import torch
 import torch.nn as nn
 import torchax
 from jax.sharding import Mesh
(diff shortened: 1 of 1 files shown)


lsy323:
kicks off a CI run at https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1155

PR #338, opened 2025-07-29
feat: Add JAX scopes for clearer Xprof traces

JiriesKaileh:
# Description

All details can be seen in this [buganizer comment](https://buganizer.corp.google.com/issues/433516648#comment7)

# Tests

To test my change I ran this command
```
python tpu_commons/examples/offline_inference.py \
    --model=meta-llama/Llama-3.1-8B \
    --tensor_parallel_size=4 \
    --task=generate \
    --max_model_len=1024
```
then visually compared the xprof to an xprof created prior to the changes. I also used the "Find Events..." feature in Trace Viewer to ensure that most of my named scopes appeared.

### DIFF
--- a/tpu_commons/models/jax/recipes/llama3.py
+++ b/tpu_commons/models/jax/recipes/llama3.py
@@ -250,7 +250,11 @@ class LlamaForCausalLM(Model):
             for (i, layer) in enumerate(self.layers):
                 kv_cache = kv_caches[i] 
 
-                if i == 0: #Leaving out layer_0 for clearer xprof reading
+                # The first layer is unscoped to avoid JAX tracing issues.
+                # JAX's profiler may incorrectly apply the scope name from the first
+                # layer's kernel compilation to all subsequent layers. Skipping the
+                # first layer ensures distinct scope names for the remaining layers.
+                if i == 0: 
                     new_kv_cache, x = layer(x, is_prefill, kv_cache,
                                             attention_metadata)
                 else:


tpu_commons/models/jax/common/layers.py
```diff
@@ -86,20 +86,24 @@ def __call__(self, x: Float, op_mode='generate') -> Float:
...
         return normed_x.astype(self.dtype)
 
     def generate_kernel(self, rngs: nnx.Rngs):
-        self.scale = self.param_factory.create_scale_param(
-            rngs,
-            shape=(self.dims, ),
-            sharding=self.scale_sharding,
-            dtype=self.dtype)
+        with jax.named_scope("rms_norm_kernel_init"):
+            self.scale = self.param_factory.create_scale_param(
```
bzgoogle: could we put the named_scope in side of self.param_factory.create_scale_param() or create_kernel_param(), so that when there's new module created, we don't need to explicitly add the name, but relying on  create_kernel_param to handle it.  i.e adding a new arg in  create_scale_param(name, **) and with jax.named_scope(name) in create_kernel_param()?

tpu_commons/models/jax/common/layers.py
```diff
@@ -202,21 +206,24 @@ def __call__(self, x, op_mode):
...
-            shape=(D, F),
-            dtype=self.cfg.dtype,
-            sharding=self.df_sharding)
-        self.kernel_down_proj_FD = self.param_factory.create_kernel_param(
-            rngs,
-            shape=(F, D),
-            dtype=self.cfg.dtype,
-            sharding=self.fd_sharding)
+        with jax.named_scope("ffw_gating_kernel_init"):
+            self.kernel_gating_DF = self.param_factory.create_kernel_param(
```
bzgoogle: ditto

tpu_commons/models/jax/common/layers.py
```diff
@@ -202,21 +206,24 @@ def __call__(self, x, op_mode):
...
+                rngs,
+                shape=(D, F),
+                dtype=self.cfg.dtype,
+                sharding=self.df_sharding)
+        with jax.named_scope("ffw_up_proj_kernel_init"):
```
bzgoogle: ditto

tpu_commons/models/jax/common/layers.py
```diff
@@ -202,21 +206,24 @@ def __call__(self, x, op_mode):
...
+            self.kernel_up_proj_DF = self.param_factory.create_kernel_param(
+                rngs,
+                shape=(D, F),
+                dtype=self.cfg.dtype,
+                sharding=self.df_sharding)
+        with jax.named_scope("ffw_down_proj_kernel_init"):
```
bzgoogle: ditto

tpu_commons/models/jax/common/layers.py
```diff
@@ -307,14 +314,14 @@ def __call__(self, x, decode=False):
...
     def generate_kernel(self, rngs: nnx.Rngs):
-
-        V = getattr(self.cfg, HuggingFaceArgNames.VOCAB_SIZE.value)
-        D = getattr(self.cfg, HuggingFaceArgNames.HIDDEN_SIZE.value)
-        self.input_embedding_table_VD = self.param_factory.create_kernel_param(
-            rngs,
-            shape=(V, D),
-            sharding=self.vd_sharding,
-            dtype=self.cfg.dtype)
+        with jax.named_scope("embedding_kernel_init"):
```
bzgoogle: ditto

tpu_commons/models/jax/common/layers.py
```diff
@@ -370,15 +380,15 @@ class LMhead(Embedder):
...
 
-        V = getattr(self.cfg, HuggingFaceArgNames.VOCAB_SIZE.value)
-        D = getattr(self.cfg, HuggingFaceArgNames.HIDDEN_SIZE.value)
-
-        self.input_embedding_table_DV = self.param_factory.create_kernel_param(
-            rngs,
-            shape=(D, V),
-            sharding=self.dv_sharding,
-            dtype=self.cfg.dtype)
+            self.input_embedding_table_DV = self.param_factory.create_kernel_param(
```
bzgoogle: ditto

tpu_commons/models/jax/recipes/llama3.py
```diff
@@ -242,21 +242,26 @@ def __call__(
...
-        x = self.embedder.encode(input_ids)
-        for (i, layer) in enumerate(self.layers):
-            kv_cache = kv_caches[i]
-            with jax.named_scope(f'layer_{i}'):
-                new_kv_cache, x = layer(x, is_prefill, kv_cache,
-                                        attention_metadata)
-            kv_caches[i] = new_kv_cache
-
-        final_activation = self.final_norm(x)
+        with jax.named_scope("llama_embed_input"): #Embedding
```
bzgoogle, line 246: we have name of  "embedder_encode_lookup" in the embedder.encode(). May i know why we need another hier of llama_embed_input on top of it?

bzgoogle: reading below. I guess you want to have the heroical name scope, even it fully overlaps the other. If so, it lgtm

tpu_commons/models/jax/recipes/llama3.py
```diff
@@ -242,21 +242,26 @@ def __call__(
...
+            x = self.embedder.encode(input_ids)
+        
+        with jax.named_scope("llama_model_transformer_blocks"):
+            for (i, layer) in enumerate(self.layers):
+                kv_cache = kv_caches[i]
+                with jax.named_scope(f'layer_{i}'):
```
bzgoogle: An unblocking comments: previously when i add the layer_{i} name, I notice that the ragged_atttention kernel call is always layer_0(https://screenshot.googleplex.com/3HfQfpu6fRAJD3Y), the possible reason could be due to the attention kernel is only compiled once(in layer 0) and shared by all the other layers. Do we have a better way to address this in the xprof?

tpu_commons/models/jax/common/layers.py
jrplatin: @bzgoogle  feel free to correct me here, but I don't think we really need to profile the kernel init in the first place since that's obviously not contributing to any serving time, so IMO, I think we should just get rid of this (and related changes) entirely

tpu_commons/models/jax/recipes/llama3.py
jrplatin: I agree -- @JiriesKaileh this is something that you should dive a bit into (Moma/gChat etc.) since it would be great to mitigate this

tpu_commons/models/jax/common/layers.py
bzgoogle: Agree that kernel init happens before the serving and even before the checkpoint loading. So it won't account for the serving timing in xprof

PR #341, opened 2025-07-30
[RPA] Add RPA kernel v3

bythew3i:
[RPA] Introduce new Ragged Paged Attention kernel

Highlights:

- New RPA kernel now supports any model spec (num_q_heads, num_kv_heads, head_dim, dtype, num_heads_sharding)
- Completely new design of the RPA by getting rid of the traditional GMM solution to handle the Ragged inputs and we now always only DMA effective data (no more padding will be fetched)
- Allow user to specify the distribution of decode-only / fixed chunked-prefill-only / mixed. This helps compiler generate more efficient code. In other words, new RPA supports pure decode-only or pure prefill-only or totally ragged use case.
- The new RPA requires special shapes for q, k, v. User can consider reshape their W-q, W-k, W-v in the first place.
- The interface make it very easy to fuse KV cache write into the RPA kernel (coming soon)

---

Please leave any comments and feedback to cl/784761837. This PR aims to check in to unblock the integration work for new RPA.


tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
```diff
@@ -0,0 +1,87 @@
...
+#     - kv_dtype_name
+#     - actual_num_q_heads
+#     - actual_num_kv_heads
+#     - head_dim
+#     - max_num_tokens
+#     - max_model_len = page_size * pages_per_seq
+# value:
+#     - num_kv_pages_per_block
+#     - num_queries_per_block
+TUNED_BLOCK_SIZES = {}
```
bythew3i, line 21: TODO: Fill in the block sizes

PR #348, opened 2025-07-30
Adds the log rate limitng.

wang2yn84:
The current run is overwhelmed by these warning messages. Adds the rate limiting filter so that we can control the freq. Defaults to 60 seconds. Ideally we probably should switch to another more sophisticated logging library. Stick to the current one for now.

# Tests

Manually tested.

# Checklist

### DIFF
--- a/tpu_commons/utils.py
+++ b/tpu_commons/utils.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
+import logging
 import os
+import time
 from collections import defaultdict
 from typing import Any, List, Tuple
 
@@ -99,3 +101,18 @@ def get_padded_num_heads(num_heads: int, sharding_size: int) -> int:
         assert sharding_size % num_heads == 0
         num_heads = sharding_size
     return num_heads
+
+
+class RateLimitFilter(logging.Filter):
+
+    def __init__(self, name='', interval=60):
+        super().__init__(name)
+        self.interval = interval
+        self.last_logged = 0
+
+    def filter(self, record):
+        now = time.monotonic()
+        if now - self.last_logged >= self.interval:
+            self.last_logged = now
+            return True
+        return False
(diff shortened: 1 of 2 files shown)


tpu_commons/utils.py
```diff
@@ -82,3 +84,18 @@ def get_padded_num_heads(num_heads: int, sharding_size: int) -> int:
         assert sharding_size % num_heads == 0
         num_heads = sharding_size
     return num_heads
+
+
+class RateLimitFilter(logging.Filter):
```
fenghuizhang, line 106: Instead of rate limit the logs, I prefer us to clean up the log messages. If we are spamming warnings, something is wrong. Thanks.

xiangxu-google: +1, @yarongmu-google to review if we can reduce or remove those warning logs in _temporary_vllm_compat.py

jrplatin:
What logs in particular is this targeting? Can you give an example?

tpu_commons/utils.py
yarongmu-google: +1 to minimize the logs. As for the ones in _temporary_vllm_compat.py, I will remove them later, because:
(1) these are not currently called in the stack yet so shouldn't yield any warnings.Not called because of (2) below
(2) these are the ones that caught my wrong entry when I was trying to DI the workers in vllm (instead, I should have done so at the platform level).
So I need them for now :)

wang2yn84: Hi Yarong, to provide more context, line 80 of _temporary_vllm_compat.py prints repeated (for each batch I believe).

yarongmu-google: You are right - removed already - try syncing?

wang2yn84:
Close this PR since the flooding comments are handled in [PR#355](https://github.com/vllm-project/tpu_commons/pull/355)

PR #349, opened 2025-07-30
Capture sampling graph during warm up

lsy323:
# Description

Capture sampling graph during warm up. Logits is partitioned along hsize dim

# Tests

CI, offline benchmarking

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -5,8 +5,6 @@ steps:
    - label: "E2E MLPerf tests for JAX models"
      key: test_0
      soft_fail: true
-     env:
-       VLLM_XLA_CHECK_RECOMPILATION: "1"
      agents:
        queue: tpu_v6e_queue
      commands:
@@ -17,7 +15,6 @@ steps:
      soft_fail: true
      env:
        MODEL_IMPL_TYPE: "vllm"
-       VLLM_XLA_CHECK_RECOMPILATION: "1"
      agents:
        queue: tpu_v6e_queue
      commands:


.buildkite/pipeline_jax.yml
```diff
@@ -15,6 +17,7 @@ steps:
      soft_fail: true
      env:
        MODEL_IMPL_TYPE: "vllm"
+       VLLM_XLA_CHECK_RECOMPILATION: "1"
```
xiangxu-google: we already enabled it in https://github.com/vllm-project/tpu_commons/blob/main/.buildkite/scripts/run_in_docker.sh#L56

lsy323: Thank you for the heads up, let me remove it. I guess ti's not expected that it's not caputrued in CI before?

lsy323:
> Thanks a lot @lsy323 for the catch!
> 
> I have two questions:
> 
> 1. `pipeline_jax.yaml` uses `run_in_docker.sh`. `run_in_docker.sh` has `VLLM_XLA_CHECK_RECOMPILATION=1`. Then do we still need the `VLLM_XLA_CHECK_RECOMPILATION=1` additions to the yaml in this PR?
> 2. Also, I'm a bit confused, how wasn't this triggered in CI but triggered in benchmark?

Like @xiangxu-google pointed out in offline chat, CI doesn't capture this because no multi-chip in CI yet

.buildkite/pipeline_jax.yml
py4: @xiangxu-google discovered that the reason it's not triggered in CI is that CI is one single chip but benchmark is on multi chip

lsy323:
CI Run triggered at https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1213

PR #354, opened 2025-07-31
minor fix in insert logic

sixiang-google:
# Description

block indices are used to compute offset after inserted into block table, so the indices cannot be set to a oob value. After fixing this the e2e is able to run. 

# Tests

commands to reproduce.

# Checklist


fenghuizhang:
Can you add more details to the description? Thanks.

PR #360, opened 2025-07-31
[Kernel] optimize kv cache update kernel block size

yaochengji:
# Description

- Automatically adjust kv cache update kernel's num_slices_per_block, porting from https://github.com/vllm-project/vllm/pull/20415 and https://github.com/vllm-project/vllm/pull/21007
- change RPA kernel's vmem_limit_bytes to 100MB to avoid vmem OOM issue

# Tests

1. pytest -s -v tests/kernels/ragged_kv_cache_update_test.py
2. server: TPU_BACKEND_TYPE=torchax VLLM_TORCHAX_ENABLED=1 vllm serve meta-llama/Llama-3.1-8B-Instruct --disable-log-requests --gpu-memory-utilization 0.98 --max-num-batched-tokens 2048 --max-num-seqs 128 --max-model-len 2048 --no-enable-prefix-caching --tensor_parallel_size=1 
client: python3 ./benchmarks/benchmark_serving.py --model meta-llama/Llama-3.1-8B-Instruct --dataset-name sonnet --dataset-path benchmarks/sonnet_4x.txt --sonnet-input-len 1800 --sonnet-output-len 128 --ignore_eos

We can observe the throughput increases from 8.11 req/s to 8.14 req/s.

# Checklist

### DIFF
--- a/tpu_commons/attention/backends/pallas_torchax.py
+++ b/tpu_commons/attention/backends/pallas_torchax.py
@@ -17,13 +17,9 @@ from tpu_commons.logger import init_logger
 # Register custom op dispatcher.
 from tpu_commons.models.torchax.torchax_wrapper import (kv_cache_update,
                                                         ragged_paged_attention)
+from tpu_commons.utils import TPU_HEAD_SIZE_ALIGNMENT
 
 logger = init_logger(__name__)
-
-# TPU requires the head size to be a multiple of 128.
-TPU_HEAD_SIZE_ALIGNMENT = 128
-# Block size used for kv cache updating kernel
-NUM_SLICES_PER_KV_CACHE_UPDATE_BLOCK = 8
 
 
 class PallasAttentionBackend(AttentionBackend):
@@ -233,7 +229,7 @@ class PallasAttentionBackendImpl(AttentionImpl):
             # these can be manually adjusted for debugging if necessary.
             num_kv_pages_per_block=None,
             num_queries_per_block=None,
-            vmem_limit_bytes=None,
+            vmem_limit_bytes=100 * 1024 * 1024,
             use_kernel=True,
             sm_scale=self.scale,
             sliding_window=self.sliding_window,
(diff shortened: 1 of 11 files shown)


yaochengji:
cc @lsy323 @xiangxu-google

tpu_commons/attention/backends/pallas_torchax.py
```diff
@@ -233,7 +229,7 @@ def forward(
             # these can be manually adjusted for debugging if necessary.
             num_kv_pages_per_block=None,
             num_queries_per_block=None,
-            vmem_limit_bytes=None,
+            vmem_limit_bytes=100 * 1024 * 1024,
```
bythew3i, line 232: Nit: better to test the throughput before and after this change. Kernel won't be affected but next op's prefetch will be affected.

tpu_commons/kernels/ragged_kv_cache_update.py
```diff
@@ -140,8 +143,8 @@ def _kv_cache_update(
     page_size: int,
     num_slices_per_block: int,
     dynamic_validate_inputs: bool,
+    vmem_limit_bytes: int = 40 * 1024 * 1024,
```
bythew3i, line 146: How is this calculated?

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i, line 1: I would suggest - have an autotune and benchmarking in google workspace internally first. Like we do for RPA and quantized_matmul

tpu_commons/attention/backends/pallas_torchax.py
yaochengji: It would crash due to vmem OOM before the change.

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: Basically I'd like to have 32MB vmem to be used for scratch buffer, and it's round up to 40MB.

tpu_commons/kernels/ragged_kv_cache_update.py
yaochengji: Thanks for the suggestion! Since I'd like to make the kv cache update kernel on par with vLLM torch/xla path, we can add the auto-tuning and benchmarking later in google internally later?

PR #364, opened 2025-08-01
Use numpy instead of jnp to calculate rope cache

Chenyaaang:
# Description

Use numpy instead of jnp to calculate rope cache. This is part of #293  following small pr practices.

The model_loader uses [nnx.eval_shape](https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/models/jax/model_loader.py#L136) to trace the shapes and dtypes of the model without actually allocating memory for them. If using jnp in compute_sin_cos(), the result will be a tracer which can't be assigned to an object attribute (self.sin_cos_cache) and will cause `jax.errors.UnexpectedTracerError: Encountered an unexpected tracer. A function transformed by JAX had a side effect, allowing for a reference to an intermediate value with type float32[163840,64] wrapped in a DynamicJaxprTracer to escape the scope of the transformation.` error. 

By switching to numpy, the cache is a numpy array, and will be converted to jnp when it needs to be indexed inside `apply_rope`.

# Tests

- test_rope.py passed
- Deepseek E2E run passed

# Checklist

### DIFF
--- a/tpu_commons/models/jax/common/rope.py
+++ b/tpu_commons/models/jax/common/rope.py
@@ -1,6 +1,7 @@
 import math
 
 import jax
+import numpy as np
 from jax import numpy as jnp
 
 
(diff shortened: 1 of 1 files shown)


wenxindongwork:
I'm a bit worried about performance if we switch to numpy. Can we do something like `self.sin_cos_cache = jax.jit(self._compute_sin_cos)()` to store a jnp array?

Chenyaaang:
> I'm a bit worried about performance if we switch to numpy. Can we do something like `self.sin_cos_cache = jax.jit(self._compute_sin_cos)()` to store a jnp array?

Actually I just tested without this change and it also passed, this one is probably no longer needed. Thanks!

PR #381, opened 2025-08-04
deepseek v3 model structure and weight loading

Chenyaaang:
# Description

Implement deepseek v3 model and load weights

This pr is separated from #293 following small pr practice, and is based on #361 and #362.
- Define model structure in deepseek_v3.py
- Register model in model_loader.
- Integrate ragged paged attention kernel V3 (there will be a followup pr once the kernel refactor finishes cc @bythew3i)

Some future todos:
- Update attention integration
- reorder batch on tpu_jax_runner, so that it we can benefit from kernel v3, as it needs to know which requests are prefill/decode.

# Tests

This pr contains 2 commits, the old commit (00c2df) reduces num_layers to 4 (with 3 ffw and 1 moe) so that it can be tested on v6e-8. The new commit(d9fb76) uses the original model shape which will be finally submitted. 

There are also some changes needed on vllm side to bypass the ModelConfig check, and there will be a formal pr to implement this. To run the model temporarily, apply my hacks in this [commit](https://github.com/vllm-project/vllm/commit/8b2ee0ee355012d894f205046841394829c5d096)

Command to run E2E:
`VLLM_MLA_DISABLE=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --task=generate --model=deepseek-ai/DeepSeek-V3  --max-model-len=1024 --tensor-parallel-size 8 --max-num-batched-tokens 1024 --max-num-seqs=1 --hf-config=deepseek-ai/DeepSeek-V3 --hf_overrides '{"architectures": ["DeepSeekV3"]}'`

If load dummy weight, add `JAX_RANDOM_WEIGHTS=True`.

I've locally tested 4 layers model worked with both standard weights and dummy weights. 

# Checklist

### DIFF
--- a/tpu_commons/models/jax/common/attention/attention.py
+++ b/tpu_commons/models/jax/common/attention/attention.py
@@ -12,9 +12,12 @@ from vllm.config import VllmConfig
 
 from tpu_commons.kernels.ragged_paged_attention.kernel import \
     ragged_paged_attention
-from tpu_commons.kernels.ragged_paged_attention.v3.kernel import ragged_paged_attention as ragged_paged_attention_v3
-from tpu_commons.kernels.ragged_paged_attention.v3.kernel import (prepare_inputs, prepare_outputs)
-from tpu_commons.kernels.ragged_paged_attention.v3.util import (align_to, get_dtype_packing)
+from tpu_commons.kernels.ragged_paged_attention.v3.kernel import (
+    prepare_inputs, prepare_outputs)
+from tpu_commons.kernels.ragged_paged_attention.v3.kernel import \
+    ragged_paged_attention as ragged_paged_attention_v3
+from tpu_commons.kernels.ragged_paged_attention.v3.util import (
+    get_dtype_packing)
 from tpu_commons.models.jax.attention_interface import update_kv_cache
 from tpu_commons.models.jax.attention_metadata import AttentionMetadata
 from tpu_commons.models.jax.common.base import Config, ParamFactory
@@ -31,8 +34,8 @@ AttentionConfig = make_dataclass(
      (HuggingFaceArgNames.NUM_KEY_VALUE_HEADS.value, int),
      (HuggingFaceArgNames.HEAD_DIM.value, int),
      (HuggingFaceArgNames.ROPE_SCALING.value, Dict[str, Any]),
-     (HuggingFaceArgNames.ROPE_THETA.value, float),
-     ("dtype", DTypeLike), ("rope_input_ordering", str, "split"),
+     (HuggingFaceArgNames.ROPE_THETA.value, float), ("dtype", DTypeLike),
+     ("rope_input_ordering", str, "split"),
      (HuggingFaceArgNames.ATTENTION_CHUNK_SIZE.value, int, None),
      ("vllm_config", VllmConfig, field(repr=False, default=None))],
     bases=(Config, ))
@@ -296,7 +299,7 @@ class Attention(nnx.Module):
             )
 
         return kv_cache, output_TNH
-    
+
     def attention_v3(
         self,
         is_prefill: bool,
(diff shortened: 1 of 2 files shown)


tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -293,3 +296,84 @@ def _ragged_paged_attention(*args):
...
+        q_transformed, actual_num_q_heads_per_kv_head, actual_head_dim = prepare_inputs(q_TNH, self.K)
+        kv_packing = get_dtype_packing(kv_cache.dtype)  # 2 for bf16, 4 for fp8.
+        L, S, K_2, H = kv_cache.shape
+        kv_cache_transformed = kv_cache.reshape(
+            L, S, K_2 // kv_packing, kv_packing, H)
+        page_indices_flat = md.block_tables.flatten()
+        
+        # TODO: update this once attention_metadata has distribution info.
+        # Currently put all requests into prefill mode.
+        distribution = jnp.array([0, md.num_seqs[0], md.num_seqs[0]], dtype=jnp.int32)
```
wenxindongwork: curious what does `distribution` do here?

tpu_commons/models/jax/common/sharding.py
```diff
@@ -63,10 +63,14 @@ class ShardingRulesConfig:
     activation_q_td: tuple = (None, None)
     # Attention Out activation after projection: (Batch * Sequence, NumHeads, HeadDim)
     attn_o_tnh: tuple = (None, None, None)
+    # Ragged attention v3 Out: (actual_num_kv_heads, max_num_tokens, num_q_heads_per_kv_head // q_packing, q_packing, head_dim)
+    attn_o_ktnph: tuple = (None, None, None, None, None)
```
wenxindongwork, line 67: do we also need to specify the prefill and generate sharding spec for these tensors?

tpu_commons/models/jax/recipes/deepseek_v3.py
```diff
@@ -15,6 +140,15 @@ class DeepSeekV3ShardingRulesConfig(ShardingRulesConfig):
     attn_mla_kva_weight_da: tuple = (None, None)
     # MLA KV up projection weight: (KVLoRA, NumHeads, QKNopeHeadDim + VHeadDim)
     attn_mla_kvb_weight_anh: tuple = (None, None, None)
+    # Attention V3 output: (actual_num_kv_heads, max_num_tokens, num_q_heads_per_kv_head // q_packing, q_packing, head_dim)
```
wenxindongwork, line 144: would these sharding rules be shared across models? if so, maybe they could be moved into `ShardingRulesConfig`

tpu_commons/models/jax/common/sharding.py
Chenyaaang: yes it's in deepseek_v3 -> DeepSeekV3ShardingRulesConfig.

tpu_commons/models/jax/recipes/deepseek_v3.py
Chenyaaang: Yes it will, but according to Jevin, he's going to move the prepare_input inside the kernel, so we no longer need this soon.

tpu_commons/models/jax/common/attention/attention.py
Chenyaaang: It divides the requests into 3 parts: 1 prefill only, 2 decode only, 3 mixed, and the kernel optimizes them separately. https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/kernels/ragged_paged_attention/v3/kernel.py#L948

tpu_commons/models/jax/recipes/deepseek_v3.py
wenxindongwork: I see! That makes sense. Maybe we could leave a comment/TODO so others also know?

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -293,3 +296,84 @@ def _ragged_paged_attention(*args):
...
+            ))(
+                q_transformed,
+                kv_cache_transformed,
+                md.seq_lens,
+                page_indices_flat,
+                md.query_start_loc,
+                distribution,
+            )
+        output_TNH = prepare_outputs(output_transformed, actual_num_q_heads_per_kv_head, actual_head_dim)
+        return kv_cache, output_TNH
```
wenxindongwork: nit: new line is missing

Chenyaaang: added, thanks

PR #385, opened 2025-08-04
Add torchax-compatible jax flash attention kernel

vlad-karp:
# Description

Multimodal inference requires standard flash attention kernel.
This PR adds torchax wrapper for a standard jax flash attention pallas kernel.

Also takes the 3-rd parameter of the jax wrapper for model forward inputs as multimodal input embeddings.
This change will be required for multimodal inference.

The kernel wrapper now tested and works for input shapes which are multiples of 128*n which reflects vllm's upstream logic to pre-compile multimodal encoders and run them with inputs which are powers of 2.
Basic padding logic is added but is not producing results comparable with reference implementation since this is not a goal of this PR. This will require further investigation if we'll actually need it.
There is still a remaining issue of MM inference crashing using this kernel along with the compiled backbone model.

# Tests

Added a unit test and was also tested with a downstream multimodal support PR.

### DIFF
--- a/tests/kernels/flash_attention_kernel_test.py
+++ b/tests/kernels/flash_attention_kernel_test.py
@@ -1,0 +1,126 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import jax
+import torch
+import torch.nn.functional as F
+import torchax
+from absl.testing import absltest, parameterized
+from jax._src import test_util as jtu
+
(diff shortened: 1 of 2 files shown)


yarongmu-google:
@hfan @yaochengji @xiangxu-google too - this supposedly makes the MM crash go away.

yarongmu-google:
@bythew3i too :)

PR #401, opened 2025-08-05
Add torchax path for multimodal models

vlad-karp:
# Description

torchax path for multimodal models is adopted from vLLM's vllm/v1/worker/tpu_model_runner.py class
It includes:
- calculating visual encoder tokens budget
- instantiating and pre-compiling a separate encoder tower for non-text inputs
- preparing multimodal input as a separated jax jitted function
- an e2e offline inference script for demonstration and testing: tested on llava model.

Current issues:
- This pass only works until the backbone is compiled. Somehow using visual encoder's flash attention kernel leaves runtime in a state that allows for a single(i.e. non compiled) backbone run and after there there is generic jax runtime error
- tested only on a single tpu device (v6e), no sharded model test was performed.

# Tests
Run a single sample example using  
examples/offline_inference_multimodal.py

### DIFF
--- a/tpu_commons/runner/tpu_torchax_runner.py
+++ b/tpu_commons/runner/tpu_torchax_runner.py
@@ -180,7 +180,6 @@ class TPUModelRunner(LoRAModelRunnerMixin):
         self.vocab_size = model_config.get_vocab_size()
 
         # Multi-modal data support
-        # NOTE(vlakarp): this mm budget computation ...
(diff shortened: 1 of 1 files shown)


yarongmu-google:
@hfan @yaochengji @xiangxu-google too - this supposedly makes the MM crash go away.

yarongmu-google:
@bythew3i too :)

PR #421, opened 2025-08-06
Improve throughput on some single chip models

hfan:
# Description

Bring back the code path to run matmuls in JaxMergedColumnParallelLinear and JaxQKVParallelLinear separately, and provide model specific overwrites for certain matmuls, as for large matmuls the output can spill to HBM and it's slower than running matmuls separately and having the output on VMEM such that later computes doesn't need to load from HBM.

# Tests

Unit test: `pytest -v tests/models/vllm/test_jax_*linear.py` and `pytest -v tests/models/jax/test_model_loader.py`

E2e benchmarking shows better performance for `Qwen2.5-7B-Instruct` and `Llama-3.1-8B-Instruct`.

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/jax_merged_column_parallel_linear_fusion_assignments.py
+++ b/tpu_commons/models/vllm/jax_merged_column_parallel_linear_fusion_assignments.py
@@ -35,7 +35,8 @@ def get_model_matmul_fusion_assignment(model_name: str, batch_size: int,
     key = (model_name, batch_size, tp_size, layer_name)
     fuse_matmuls = MODEL_MATMUL_FUSION_TRUTH_TABLE.get(key, None)
     if fuse_matmuls is None:
-        logger.info(f"No matmul fusion overwrite for {key=}, default to True.")
         return True
     else:
+        logger.info(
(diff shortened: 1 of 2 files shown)


hfan:
Also @yarongmu-google

tpu_commons/models/vllm/jax_merged_column_parallel_linear_fusion_assignments.py
```diff
@@ -0,0 +1,42 @@
+from tpu_commons.logger import init_logger
+
+logger = init_logger(__name__)
+
+MODEL_MATMUL_FUSION_TRUTH_TABLE = {
+    ("Qwen/Qwen2.5-7B-Instruct", 1024, 1, "QKVParallelLinear"):
```
lsy323, line 6: What about we exclude quant models in the table for now? maybe with q matmul it will fuse differently?

hfan: benchmarked that the w8a8 quantized llama3 8b is also faster with the overwrite

PR #423, opened 2025-08-06
Reorder requests in tpu jax runner

Chenyaaang:
# Description

Reorder the requests into (decode, prefill, mixed) to fit in RPA V3. 

# Tests

- Add tests in test_tpu_jax_runner.py
- Verified with test deepseek model. 

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -1234,56 +1234,55 @@ class TPUModelRunner():
         decode_index_set = set(decodes)
         prefill_index_set = set(non_decodes[most_freq_prefill_length]
                                 ) if max_prefill_cnt != 0 else set()
+        modified_batch = False
         for i in range(num_decode):
             if i in prefill_index_set:
                 prefill_in_decode.append(i)
+                modified_batch = True
             elif i in decode_index_set:
                 continue
             else:
                 mixed_in_decode.append(i)
+                modified_batch = True
         for i in range(num_decode, num_decode + num_prefill):
             if i in prefill_index_set:
                 continue
             elif i in decode_index_set:
                 decode_in_prefill.append(i)
+                modified_batch = True
             else:
                 mixed_in_prefill.append(i)
+                modified_batch = True
         for i in range(num_decode + num_prefill, num_reqs):
             if i in prefill_index_set:
                 prefill_in_mixed.append(i)
+                modified_batch = True
             elif i in decode_index_set:
                 decode_in_mixed.append(i)
+                modified_batch = True
             else:
                 continue
 
-        modified_batch = False
         # 2 way swap
-        while prefill_in_decode and decode_in_prefill:
-            i, j = prefill_in_decode.pop(), decode_in_prefill.pop()
-            self.input_batch.swap_states(i, j)
-            modified_batch = True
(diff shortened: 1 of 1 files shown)


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1161,7 +1167,10 @@ def _update_states(self, scheduler_output: "VllmSchedulerOutput") -> bool:
         if removed_req_indices:
             self.input_batch.condense(removed_req_indices)
 
-        return len(unscheduled_req_ids) > 0 or len(req_ids_to_add) > 0
+        batch_changed = len(unscheduled_req_ids) > 0 or len(req_ids_to_add) > 0
+        # batch_reordered = self._reorder_batch(scheduler_output)
```
wenxindongwork: Should this line be uncommented?

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1183,6 +1192,103 @@ def _sync_weights(
...
+            num_tokens = scheduler_output.num_scheduled_tokens[req_id]
+            if num_tokens == 1:
+                decodes.append(i)
+            else:
+                if num_tokens not in non_decodes:
+                    non_decodes[num_tokens] = []
+                non_decodes[num_tokens].append(i)
+        assert num_reqs == self.input_batch.num_reqs
+
+        # find the most frequent prefill length
```
wenxindongwork, line 1217: Maybe this could be simplified to something like the following:

if non_decodes:
    most_freq_prefill_length = max(non_decodes, key=lambda prefill_len: len(non_decodes[prefill_len]))
else:
    most_freq_prefill_length = 0

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1183,6 +1192,103 @@ def _sync_weights(
...
+        max_prefill_cnt, most_freq_prefill_length = 0, 0
+        for prefill_len in non_decodes.keys():
+            prefill_cnt = len(non_decodes[prefill_len])
+            if prefill_cnt > max_prefill_cnt:
+                max_prefill_cnt = prefill_cnt
+                most_freq_prefill_length = prefill_len
+
+        # Find out requests that need to be swapped.
```
wenxindongwork, line 1223: Wondering if this logic could be simplified by introducing a new function `reorder_states` in the InputBatch class. 

Here we simply call 
```
self.input_batch.reorder_states(new_order)
```
and in Input Batch, we create a new function 

```
  def reorder_states(self, new_order):

      new_req_ids = [self.req_ids[i] for i in new_order]      
      self.req_ids = new_req_ids
      ...
```

tpu_commons/runner/jax/tpu_jax_runner.py
Chenyaaang: Yes, thanks! I forgot it.

tpu_commons/runner/jax/tpu_jax_runner.py
Chenyaaang: I agree it will be cleaner but the current way we can take advantage of the existing swap_states function, and also the block table is using swap_rows to change the order, so I prefer not change (and it won't introduce new error).

PR #438, opened 2025-08-07
[RPA] Fuse kv cache update to RPA v3

bythew3i:
# Changes in RPA v3

* Fuse kv cache update to RPA
* Move reshape of q, k, v to RPA
* Move k, v merging to RPA
* Add debug mode

Follow up:
* Integrate to JAX path
* Retune block sizes
* Deprecate RPA v2 and change tpu_commons to use latest jax and libtpu

# Tests

Tests require the lastest JAX and LIBTPU. I tested the kernel on local already. Next PR will add the integration and pin the newer JAX and LIBTPU versions.

# Checklist


bythew3i:
@yarongmu-google @yaochengji @vanbasten23 @kyuyeunk Can you please start to review the kernel change in parallel while I am working the integration on vllm/ullm side. Thanks!

yarongmu-google:
LGTM

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -62,15 +88,27 @@ def ref_ragged_paged_attention(
...
         indices_start = i * pages_per_seq
         indices_end = indices_start + cdiv(kv_len, page_size)
         indices = page_indices[indices_start:indices_end]
         q = queries[q_start:q_end, :, :actual_head_dim]
-        kv = (kv_cache[indices].reshape(
+
+        # Update the kv cache.
+        assert kv_len - q_len >= 0
+        gathered_kv = kv_cache[indices]
+        gathered_shape = gathered_kv.shape
```
vanbasten23, line 98: Is it possible to add a comment about the shape of `gathered_kv`: [num_pages, page_size, num_kv_heads_x2 // kv_packing, kv_packing, head_dim], so that it could be easier to see what the -3rd dim is

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -535,15 +731,33 @@ def compute_with_bkv(bkv_idx, _):
...
                 # Wait for cur bkv
-                wait_bkv(seq_idx, bkv_idx, bkv_sem_idx)
+                offset, update_sz = wait_fetch_bkv(seq_idx, bkv_idx,
+                                                   bkv_sem_idx)
+
+                # Start updating bkv to kv cache if applicable.
+                # Only needed in first bq loop.
+                @pl.when(jnp.logical_and(update_sz > 0, bq_idx == 0))
+                def update_cur_bkv_to_cache():
+                    start_update_kv_cache(seq_idx, bkv_sem_idx, offset,
```
yaochengji, line 747: Is it possible to update next_seq_idx's kv cache when executing flash attention of cur_seq_idx?

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -62,15 +88,27 @@ def ref_ragged_paged_attention(
...
         indices = page_indices[indices_start:indices_end]
         q = queries[q_start:q_end, :, :actual_head_dim]
-        kv = (kv_cache[indices].reshape(
+
+        # Update the kv cache.
+        assert kv_len - q_len >= 0
+        gathered_kv = kv_cache[indices]
+        gathered_shape = gathered_kv.shape
+        gathered_kv = gathered_kv.reshape(-1, *gathered_shape[-3:])
+        gathered_kv = gathered_kv.at[kv_len - q_len:kv_len].set(
```
vanbasten23, line 100: Not sure if I follow the logic here, could you explain?

Also, could it be simplified as the the reverse of the previous impl? Previously, it is
```
kv = (kv_cache[indices].reshape(
           -1, num_kv_heads_x2,
            head_dim)[:, :actual_num_kv_heads * 2, :].reshape(
                -1, actual_num_kv_heads, head_dim * 2))
```
So, now you can revert it to be
```
kv_cache[indices].reshape().at[].set(...)
```

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
bythew3i: No it is not possible in this implementation. Because we need to wait next_seq_idx's new kv arrives in buffer first then start to update to kv cache. The new kv arrives in line 743: wait_fetch_bkv(...)....Anyhow the active buffer for bkv is idle when we doing flash attention, current implementation is sending the "fetched partial new kv" from active bkv buffer to HBM while flash attention is going on.

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
bythew3i: `gathered_kv` is the gathered all pages for current seq from cache and concatenated to  the shape [gathered_num_pages, page_size, num_kv_heads_x2 // kv_packing, kv_packing, head_dim]

And then we merge dim `gathered_num_pages` and `page_size` to make the merged dim become `gathered_tokens`. Then we update the part from the new kv (not from cache).

Then scatter these pages back to update kv cache.

While gathered_kv is already here, we just reuse it to continue to do attention.

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
bythew3i: That should be easy to infer from the kv_cache shape which is on line 31, right?

PR #450, opened 2025-08-09
Fix DeepSeek-V3 Model Loading

jrplatin:
# Description

Recent model consolidation work has caused a few errors in the DS-V3 model loading code -- this PR fixes those

# Tests

Verified that the model can be loaded via:

```
VLLM_MLA_DISABLE=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --task=generate --model=deepseek-ai/DeepSeek-V3  --max-model-len=1024 --tensor-parallel-size 8 --max-num-batched-tokens 1024 --max-num-seqs=1 --hf-config=deepseek-ai/DeepSeek-V3 --hf_overrides '{"architectures": ["DeepSeekV3"]}'
```


tpu_commons/models/jax/recipes/deepseek_v3.py
```diff
@@ -162,6 +162,24 @@ def _create_mla() -> MLA:
 
         for i in range(first_k_dense_replace):
             block = TransformerBlock(
+                RMSNorm(
```
fenghuizhang: use keyword to make it easier to read?

PR #464, opened 2025-08-12
Add input preparation and ngram proposer for spec decoding

Lumosis:
# Description

The PR introduces speculative decoding to the JAX/TPU runner. By proposing several "draft" tokens with a fast N-gram drafter and validating them in a single forward pass of the larger target model, this feature reduces the number of sequential model execution steps required for generation.

This initial implementation supports N-gram based drafting with greedy sampling.

# Tests

Unit tests added

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -29,6 +29,7 @@ from vllm.v1.kv_cache_interface import (FullAttentionSpec, KVCacheConfig,
                                         KVCacheSpec)
 from vllm.v1.outputs import EMPTY_MODEL_RUNNER_OUTPUT, ModelRunnerOutput
 from vllm.v1.request import Request
+from vllm.v1.spec_decode.ngram_proposer import NgramProposer
 from vllm.v1.worker.kv_connector_model_runner_mixin import \
     KVConnectorModelRunnerMixin
 from vllm.v1.worker.utils import (gather_mm_placeholders,
@@ -39,6 +40,8 @@ from tpu_commons.logger import init_logger
 from tpu_commons.models.jax.attention_metadata import AttentionMetadata
 from tpu_commons.models.jax.common.sharding import build_mesh
 from tpu_commons.models.jax.layers.misc import shard_put
+from tpu_commons.models.jax.layers.sample.rejection_sampler import \
+    RejectionSampler
 from tpu_commons.models.jax.layers.sample.sampling import (compute_logprobs,
                                                            gather_logprobs,
                                                            sample)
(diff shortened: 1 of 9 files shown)


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -998,36 +1040,93 @@ def _execute_model(
...
-            spec_token_ids=None,
+            spec_token_ids=spec_token_ids,
             logprobs=logprobs_lists,
             prompt_logprobs_dict=prompt_logprobs_dict,
             pooler_output=[],
             kv_connector_output=kv_connector_output,
         )
         return attn_metadata, model_runner_output
 
+    def propose_draft_token_ids(
```
py4, line 1078: As discussed offline, would be great if can add good tests for the added changes to keep test coverage above 0.75

tpu_commons/spec_decode/metadata.py
py4, line 1: Maybe we can move this to `tpu_commons/runner/jax` as this directly is gonna contain jax implementation

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -952,20 +967,47 @@ def _execute_model(
...
-                tpu_sampling_metadata,
-            )
+            if spec_decode_metadata is None:
+                next_tokens = sample(
+                    self.rng_params_for_sampling,
+                    self.mesh,
+                    logits,
+                    tpu_sampling_metadata,
+                )
+                next_tokens = np.asarray(jax.device_get(next_tokens))
```
py4: I think these 3 lines in this part of `if` and also its in `else`, to be moved to after the following block:

```
# Update the cache state concurrently. Code above will not block until
        # we use `selected_token_ids`. Add mark_step if post-processing changes
 ```
 
 Because these 3 lines will block, so we should postpone them as much as possible.

File: tpu_commons/runner/jax/tpu_jax_runner.py [hunk 3]
```diff
@@ -952,20 +967,47 @@ def _execute_model(
...
+                    draft_probs=None,
+                    target_probs=target_logits,
+                    bonus_token_ids=bonus_token_ids,
+                    sampling_metadata=tpu_sampling_metadata,
+                )
+                selected_token_ids = self.rejection_sampler.parse_output(
+                    next_tokens, self.input_batch.vocab_size,
+                    spec_decode_metadata.draft_lengths)
+                valid_sampled_token_ids = selected_token_ids
+                next_tokens = np.asarray(jax.device_get(next_tokens))
```
py4: This line appears to be superfluous and can be removed. The `valid_sampled_token_ids` variable already holds the necessary processed output, and the raw next_tokens array isn't used again after this point. Removing the line will prevent the redundant device-to-host data copy and improve performance.

Lumosis:
> Correctness checks before merging:
> 
> * We should also try with `--max_num_batched_tokens=8/16/.../A SMALL NUMBER` to make the logic is also ok with partial prefills and outputs look good
> * Trigger a manual CI run to make sure all tests pass

* Tried with partial prefill and passed.
* Manual CI run passed: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1876#_

PR #468, opened 2025-08-13
[RPA] Add support for RPA query quantization

kyuyeunk:
# Description

Adds following changes
- Add support for query quantization (w8a8)
- Optimize performance of kv cache quantization (similar approach in https://github.com/pytorch/xla/pull/9528)

Please note that a proper evaluation of w8a8 performance cannot be done for now due to the lack of fp8 hardware.

# Tests

Tests require the lastest JAX and LIBTPU.

```
pytest -vs tests/kernels/ragged_paged_attention_kernel_v3_test.py -k test_ragged_paged_attention_quantized
```

# Checklist

### DIFF
--- a/tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
+++ b/tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
@@ -38,6 +38,7 @@ def ref_ragged_paged_attention(
     sliding_window: int | None = None,
     soft_cap: float | None = None,
     mask_value: float | None = DEFAULT_MASK_VALUE,
+    q_scale: float | None = None,
     k_scale: float | None = None,
     v_scale: float | None = None,
 ):
@@ -57,6 +58,7 @@ def ref_ragged_paged_attention(
         sliding_window=sliding_window,
         soft_cap=soft_cap,
         mask_value=mask_value,
+        q_scale=q_scale,
         k_scale=k_scale,
         v_scale=v_scale,
     )
@@ -108,17 +110,19 @@ def ref_ragged_paged_attention(
                 -1, actual_num_kv_heads, head_dim * 2)
         k = kv[:kv_len, :, :head_dim][:, :, :actual_head_dim]
         v = kv[:kv_len, :, head_dim:][:, :, :actual_head_dim]
-        if k_scale is not None:
-            k = (k * k_scale).astype(q.dtype)
-        if v_scale is not None:
-            v = (v * v_scale).astype(q.dtype)
         k = jnp.repeat(k, actual_num_q_heads_per_kv_head, axis=1)
         v = jnp.repeat(v, actual_num_q_heads_per_kv_head, axis=1)
+
         attn = jnp.einsum("qhd,khd->hqk",
                           q,
                           k,
                           preferred_element_type=jnp.float32)
         attn *= sm_scale
+        if k_scale is not None:
+            attn *= k_scale
+        if q_scale is not None:
+            attn *= q_scale
+
         q_span = (kv_len - q_len) + jax.lax.broadcasted_iota(
             jnp.int32, attn.shape, 1)
         kv_span = jax.lax.broadcasted_iota(jnp.int32, attn.shape, 2)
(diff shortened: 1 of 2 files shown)


kyuyeunk:
Hi @bythew3i @yaochengji @vanbasten23, can you review this PR? Thanks!

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -108,17 +110,19 @@ def ref_ragged_paged_attention(
...
+
         attn = jnp.einsum("qhd,khd->hqk",
                           q,
                           k,
                           preferred_element_type=jnp.float32)
         attn *= sm_scale
+        if k_scale is not None:
+            attn *= k_scale
+        if q_scale is not None:
+            attn *= q_scale
```
yaochengji, line 124: When both k_scale and q_scale are not None, do we need to check qkv have the same dtype?

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -38,6 +38,7 @@ def ref_ragged_paged_attention(
     sliding_window: int | None = None,
     soft_cap: float | None = None,
     mask_value: float | None = DEFAULT_MASK_VALUE,
+    q_scale: float | None = None,
```
bythew3i, line 41: Does it make sense to make these 3 to dynamic? Static scales trigger recompilation. I wondering if the scales are fixed for whole model? Or only fixed for small batch. If the latter, we should change this to prefetch scalars.

CC: @yaochengji @jrplatin

yaochengji: Dynamic per-tensor kv cache quantization is actually a bug. Considering the original kv of bfloat16 is of value `k`, when they're written into cache, they were scaled as `k / k_scale_0`. In a later step, `k_scale_0` will be changed to `k_scale_1` if it's dynamic per-tensor quantization. The restored value will become `k / k_scale_0 * k_scale_1`, not equal to `k`.

But it makes sense to make query per-tensor quantization dynamic.

yaochengji: For kv,  only per-token quantization is dynamic

kyuyeunk: I assume @bythew3i 's comment isn't regarding dynamic quantization - which computes scale values during inference time - but on making something like `q_scale_jax = jnp.array(q_scale)` so it won't trigger recompilation for every different q_scale values?

In current approach, we are assuming that scale value is loaded from checkpoint - meaning it is fixed/same for the whole model.

kyuyeunk:
> Thanks @kyuyeunk ! One small ask - can you please help setup and extend RPA autotuning/benchmarking for quantized type in internal repo? The benefit we can get from it is every time we change kernel we can detect the perf degradation in quantized RPA in the first place.

@bythew3i Sure! I'll work on that as well!

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
kyuyeunk: No, for w4a8 scenario, kv dtype is 4-bits while q dtype is 8-bits. Meaning even when `k_scale` and `q_scale` are not None, q and kv can have a different dtypes.

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
yaochengji: Yeah, but if it's a static scalar scale, we don't need to handle it like `q_scale_jax = jnp.array(q_scale)`.

bythew3i: In a separated PR, can we make it to dynamic array?

kyuyeunk: got it. sgtm!

PR #477, opened 2025-08-14
[RPA][JAX] Fix JAX pipeline tests in CI

bythew3i:
# Description

#470 does not update the change in `tpu_commons/models/jax/common/attention/attention.py`, `experimental/` and the tests. Now this PR should fix all the JAX unit tests

# Tests

```
pytest -x tests/ --ignore=/workspace/tpu_commons/tests/models/vllm/test_pallas_torchax.py            --ignore=/workspace/tp
u_commons/tests/models/jax/common/attention/test_deepseek_v3_attention.py
```

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #482, opened 2025-08-15
[RPA][JAX] Continue fixing JAX CI tests

bythew3i:
# Description

Continue fixing JAX CI tests after #477 

# Checklist


.buildkite/pipeline_jax.yml
```diff
@@ -29,8 +29,8 @@ steps:
        - |
          .buildkite/scripts/run_in_docker.sh \
            python3 -m pytest -s -v /workspace/tpu_commons/tests/ \
-           --ignore=/workspace/tpu_commons/tests/models/vllm/test_pallas_torchax.py \
-           --ignore=/workspace/tpu_commons/tests/models/jax/common/attention/test_deepseek_v3_attention.py \
+           --deselect=/workspace/tpu_commons/tests/models/jax/test_model_loader.py::test_get_vllm_model \
```
bythew3i, line 32: temporarily skip the tests because of ongoing work of torchax integration CC: @QiliangCui

QiliangCui: ack.

tests/runner/test_utils.py
```diff
@@ -259,9 +260,11 @@ def test_create_kv_caches(mesh: Mesh):
     layer_names = ["decoder.0", "decoder.1", "decoder.2"]  # Test with 3 layers
     devices = jax.devices()
 
-    expected_shape = (num_blocks, block_size, num_kv_heads * 2, head_size)
-    expected_sharding = NamedSharding(mesh, PartitionSpec(None, None, "model"))
+    expected_sharding = NamedSharding(mesh, PartitionSpec())
     expected_dtype = jnp.bfloat16
+    expected_shape = get_kv_cache_shape_with_mesh(mesh, num_blocks, block_size,
+                                                  num_kv_heads, head_size,
```
QiliangCui, line 266: nit: 

It is better to write the shape directly in here instead of using another piece of code to calculate it get_kv_cache_shape_with_mesh.

bythew3i: I am trying to avoid keeping inserting kv_cache_shape calculation logic in other places other than RPA file. Because once we change sth in kernel, we need to update everywhere.

The actual shape will run this code: https://github.com/vllm-project/tpu_commons/blob/838c9f1e583deb0bdcff483d853a8eb44fe1d50e/tpu_commons/runner/utils.py#L227-L248 which is the part of real attention process.

The expected shape will run this code: https://github.com/vllm-project/tpu_commons/blob/838c9f1e583deb0bdcff483d853a8eb44fe1d50e/tpu_commons/models/jax/attention.py#L13-L21 which is more like a util function to get kv cache shape and is not part of real attention process.

PR #488, opened 2025-08-16
[RPA3][Torchax-pt] Delete RPAv2 code and torchax-pt code

QiliangCui:
# Description
After the PR 470, torchax-jax path already works. 
This change is to remove RPAv2 from codebase to make sure no path references RPAv2 and lower the maintenance cost. 
 
The torchax-pt path, which is not in use, still depends on RPAv2. Rather than fixing it to v3, deleting it makes more sense. 

FIXES: b/438779127

**Created for comparing files. Do not review.**

# Tests

Unit Test not verified yet.
https://buildkite.com/tpu-commons/tpu-commons-ci/builds/1884

Launches and run benchmark for both jax path and torchax-jax path. 
Detail will be pasted in bug. b/438779127#comment19

Manually test: 

```
# Jax
TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=meta-llama/Llama-3.1-8B-Instruct --tensor_parallel_size=1 --task=generate --max_model_len=1024 --download_dir /mnt/disks/persist

# Torchax Jax( Jax + vllm)
MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=meta-llama/Llama-3.1-8B-Instruct --tensor_parallel_size=1 --task=generate --max_model_len=1024 --download_dir /mnt/disks/persist
#
```

### DIFF
--- a/tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
+++ b/tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
@@ -79,7 +79,6 @@ def ref_ragged_paged_attention(
     actual_num_q_heads_per_kv_head = actual_num_q_heads // actual_num_kv_heads
     max_num_seqs = kv_lens.shape[0]
     num_page_indices = page_indices.shape[0]
-    assert len(page_indices.shape) == 1
     assert num_page_indices % max_num_seqs == 0
     pages_per_seq = num_page_indices // max_num_seqs
     outputs = []


tpu_commons/models/torchax/torchax_wrapper.py
```diff
@@ -92,88 +92,145 @@ def get_mesh():
...
+        values,
+        kv_cache,
+        kv_lens,
+        page_indices,
+        cu_q_lens,
+        distribution,
+    ):
         """Calls the ragged paged attention kernel."""
+        # TODO(cuiq): We should flatten page_indices in the caller.
+        page_indices = page_indices.flatten()
```
QiliangCui: See see [example in jax](https://github.com/vllm-project/tpu_commons/blob/a6f110e68958e4aa250e8a2a71f555867d5415ee/tpu_commons/runner/jax/tpu_jax_runner.py#L306)

QiliangCui:
@lsy323 @bythew3i , please let me know if this is good.

tpu_commons/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -1321,7 +1322,7 @@ def ragged_paged_attention(
                 bkv_sz,
                 q.dtype,
                 kv_cache.dtype,
-            ) * 2.1)  # TODO(jevinjiang): figure out why it is so inaccurate?
+            ) * 2.4)  # TODO(jevinjiang): figure out why it is so inaccurate?
```
QiliangCui: @bythew3i , increase the memory a little bit because some model failed to load because of OOM.

bythew3i: That sounds good to me. Thanks!!

tpu_commons/kernels/ragged_kv_cache_update.py
bythew3i, line 1: Could you please archive RPA v2 into `kernels/ragged_paged_attention/v2`? Instead of deleting them fully?

bythew3i: Also for `ragged_kv_cache_update.py`, I think it is better to keep these kernels in kernel folder. Just for future reference. WDYT?

PR #506, opened 2025-08-19
Split JAX unit tests

xiangxu-google:
# Description

Separate kernel tests and other tests 

# Tests

# Checklist


xiangxu-google:
> Should we rename the keys from test_2 to something more meaningful?

They key is only for internal ref, the label name is for reading and is already meaningful. I'll try to remove those key refs in the future.

PR #510, opened 2025-08-19
Enable kv cache hybrid allocator in TPU

Chenyaaang:
# Description

Enable kv cache hybrid allocator in TPU. 

This is used when a model has multiple attention types (for example: sliding window attention + full attention in gpt-oss, Gemma 2/3, Ministral, cohere), the kv cache manager needs to allocate different slots to different layers, and support layer-specific prefix-cache rules. Vllm has different types of [kv cache manager](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/single_type_kv_cache_manager.py#L245) and use different [kv cache coordinators](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_coordinator.py#L396) to handle this, this pr modifies `get_kv_cache_spec` and `initialize_kv_cache` logic in tpu_jax_runner to support hybrid kv cache allocator.

- Inside `get_kv_cache_spec`, create full attention spec or sliding window attention spec according to the layer's attention info. The attention info is propagated from vllm's model definition (`MODEL_IMPL_TYPE=vllm`), if model is defined in pure jax (`MODEL_IMPL_TYPE=flax_nnx`), we simply use full attention spec for each layer. We can also turn it on for pure jax, depending on whether we want to include multi-attn models in Jax.
- Inside `initialize_kv_cache`, create kv cache for each kv cache tensor, and create a map between layer_name and kv cache index to help each layer to get its corresponding kv cache. 
- Modify torchax wrapper to take in the layer_name_to_kvcache_index mapping and select the right kv cache for each layer.
- Add unit test in `test_tpu_jax_runner.py`
- The current implementation only supports sliding window attention. chunked local attention / mamba will be supported in future prs.

# Tests

- Unit tests in `test_tpu_jax_runner.py` passed.
- E2E run with one kv cache group succeeded. 
```
MODEL_IMPL_TYPE=vllm JAX_RANDOM_WEIGHTS=True  TPU_BACKEND_TYPE=jax  python examples/offline_inference.py --task=generate --model=meta-llama/Llama-3.1-8B-Instruct --max-model-len=1024 --tensor-parallel-size 1 --max-num-batched-tokens 1024 --max-num-seqs=1 --hf-config=meta-llama/Llama-3.1-8B-Instruct
```
- E2E run with multiple kv cache groups succeeded.
```
MODEL_IMPL_TYPE=vllm JAX_RANDOM_WEIGHTS=True  TPU_BACKEND_TYPE=jax  python examples/offline_inference.py --task=generate --model=google/gemma-3-27b-it --max-model-len=1024 --tensor-parallel-size 8 --max-num-batched-tokens 1024 --max-num-seqs=1 --hf-config=google/gemma-3-27b-it
```

# Perf
The baseline is from vllm commit 2a167b and tpu_commons commit 18eb3c.
- Test model (llama3-8b) with only 1 kv cache group, which isn't impacted from this change, there's no perf loss.
- Test model (gemma3-27b) with multiple kv cache groups, at long context and decode heavy use case, we can expect significant perf gain from A to B.
A: Use full attention for all layers
B: use both sliding window attention (52 layers) and full attention (10 layers)

image

# Checklist

### DIFF
--- a/tpu_commons/models/jax/model_loader.py
+++ b/tpu_commons/models/jax/model_loader.py
@@ -180,6 +180,7 @@ def get_flax_model(
             hidden_states_sharding,
         ),
         donate_argnums=2,  # 0 is graphdef, 1 is state, 2 is kv_cache
+        static_argnums=6,  #6 is layer_name_to_kvcache_index.
     )
     def run_model(graphdef, state, *args):
         model = nnx.merge(graphdef, state)
--- a/tpu_commons/models/vllm/sharding.py
+++ b/tpu_commons/models/vllm/sharding.py
@@ -1,4 +1,6 @@
 import functools
+from dataclasses import dataclass
+from typing import Optional
 
 import humanize
 import jax
@@ -36,8 +38,26 @@ P = PartitionSpec
 logger = init_logger(__name__)
 
 
+@dataclass
+class AttentionInfo:
+    layer_name: str
+    num_kv_heads: int
+    head_size: int
+    attn_type: str
+    sliding_window: Optional[int] = None
+    kv_sharing_target_layer_name: Optional[str] = None
+
+
 def shard_attention(layer: torch.nn.Module, mesh: Mesh,
                     vllm_config: VllmConfig):
+    vllm_config.compilation_config.static_forward_context[
+        layer.layer_name] = AttentionInfo(
+            layer_name=layer.layer_name,
+            num_kv_heads=layer.num_kv_heads,
+            head_size=layer.head_size,
+            attn_type=layer.attn_type,
+            sliding_window=layer.sliding_window,
+            kv_sharing_target_layer_name=layer.kv_sharing_target_layer_name)
     return JaxAttention(layer, mesh)
 
 
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -352,8 +352,8 @@ class TPUModelRunner(KVConnectorModelRunnerMixin):
                     input_ids,
                     attn_metadata,
                     inputs_embeds,
+                    tuple(self.layer_name_to_kvcache_index.items()),
                 )
-
             hidden_states = self._select_from_array_fn(hidden_states,
                                                        logits_indices)
             logits = self.compute_logits_fn(self.state, hidden_states)
--- a/tests/runner/jax/test_kv_cache_manager.py
+++ b/tests/runner/jax/test_kv_cache_manager.py
@@ -1,12 +1,19 @@
 from unittest.mock import MagicMock, patch
 
+import jax
 import jax.numpy as jnp
 import numpy as np
+import torch
+from vllm.attention.backends.abstract import AttentionType
 from vllm.config import (CacheConfig, ModelConfig, ParallelConfig,
                          SchedulerConfig, VllmConfig)
 from vllm.sampling_params import SamplingType
+from vllm.v1.kv_cache_interface import (FullAttentionSpec, KVCacheConfig,
+                                        KVCacheGroupSpec, KVCacheTensor,
+                                        SlidingWindowSpec)
 from vllm.v1.request import Request
 
+from tpu_commons.models.vllm.sharding import AttentionInfo
 from tpu_commons.runner.jax.input_batch_jax import CachedRequestState
 from tpu_commons.runner.jax.tpu_jax_runner import TPUModelRunner
 
@@ -16,8 +23,14 @@ class TestKVCacheManager:
     def setup_method(self):
         # Mock JAX dependencies
         self.mock_devices = [MagicMock()] * 4
-        self.mock_mesh = MagicMock()
         self.mock_rng_key = MagicMock()
+
+        # create 1x1 mesh
+        devices = np.asarray(jax.devices()[:1])
+        axis_names = ('data', 'model')
+        mesh_shape = (1, 1)
+        self.mock_mesh = jax.sharding.Mesh(devices.reshape(mesh_shape),
+                                           axis_names)
 
         with patch('jax.devices', return_value=self.mock_devices), \
              patch('jax.make_mesh', return_value=self.mock_mesh), \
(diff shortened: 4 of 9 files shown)


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -244,51 +253,128 @@ def get_kv_cache_spec(self):
...
+                block_sizes=block_sizes,
+            )
+
     def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
         self.kv_caches: List[jax.Array] = []
+        self.maybe_reinitialize_input_batch(kv_cache_config)
+
+        kv_cache_sizes = {}
+        for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
+            for layer_name in kv_cache_tensor.shared_by:
```
Chenyaaang: I'm not super sure about here, the tensor is shared by `shared_by` layers, do they each get num_blocks equally? 
context: https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_utils.py#L988-L999

Besides, there used to be an [assertion](https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/tpu_model_runner.py#L1645), here I think num_blocks doesn't gaurantee to be divided by len(shared_by) so I removed it.

yaochengji:
@Chenyaaang thanks for such an important feature!

Could you also benchmark Gemma-3 model to see if we have any perf gain?

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -118,6 +121,11 @@ def __init__(
 
         # Cached draft tokens.
         self._draft_token_ids: Optional[list[list[int]]] = None
+        # Layer pairings for cross-layer KV sharing.
+        # If an Attention layer `layer_name` is in the keys of this dict, it
+        # means this layer will perform attention using the keys and values
+        # from the KV cache of `shared_kv_cache_layers[layer_name]`.
```
vanbasten23: if key is the layer_name, what's the value? Could you add it to the comment?

Chenyaaang: key is the layer name, value is the target layer name which it shares kv cache with.

vanbasten23:
This is really cool! 

QQ, the test plan mentioned you used `google/gemma-3-27b-it` to test the  `multiple kv cache groups` case. I wonder how I can tell if a model requires "multiple kv cache groups". Or if it's not the model that determine that, who else determines it?

xiangxu-google:
Thanks for adding this Chenyang! Wondering if you hve the plan to add Gemma2 and Gemma3 models to JAX.

Chenyaaang:
> This is really cool!
> 
> QQ, the test plan mentioned you used `google/gemma-3-27b-it` to test the `multiple kv cache groups` case. I wonder how I can tell if a model requires "multiple kv cache groups". Or if it's not the model that determine that, who else determines it?

The model decides it, in this case, we use `MODEL_IMPL_TYPE=vllm`, so the model is defined in vllm/model_executor/models, and the attention info is propagated to be used in kv cache spec.

Chenyaaang:
> Thanks for adding this Chenyang! Wondering if you hve the plan to add Gemma2 and Gemma3 models to JAX.

The current implementation can work with torchax-jax path. Probably in the future we want to add gemma into Jax, then we can refactor the attention class to record the attention type (the existing models in Jax are all using full attention only).

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -346,13 +346,23 @@ def _execute_model(
...
-                    attn_metadata,
-                    inputs_embeds,
-                )
+                if os.getenv("MODEL_IMPL_TYPE", "flax_nnx").lower() == "vllm":
+                    self.kv_caches, hidden_states = self.model_fn(
+                        self.state,
+                        self.kv_caches,
+                        input_ids,
+                        attn_metadata,
+                        tuple(self.layer_name_to_kvcache_index.items()),
```
xiangxu-google: can you put this tuple after `input_embeds`, so that for the jax path we can skip this tuple using the `*args` wildcard.
```
    def __call__(
        self,
        kv_caches: List[jax.Array],
        input_ids: Optional[jax.Array],
        attention_metadata: AttentionMetadata,
        inputs_embeds: Optional[jax.Array] = None,
        *args,
    )
```

Then we can remove this `if else` check and run the same interface for both.

Chenyaaang: Just did, thanks

tpu_commons/models/jax/model_loader.py
```diff
@@ -180,6 +180,7 @@ def get_flax_model(
             hidden_states_sharding,
         ),
         donate_argnums=2,  # 0 is graphdef, 1 is state, 2 is kv_cache
+        static_argnums=6,  #6 is layer_name_to_kvcache_index.
```
yaochengji, line 183: I would prefer `static_argnames`, WDYT? donate_argnums could also be changed to `donate_argnames`.

tpu_commons/models/vllm/sharding.py
```diff
@@ -36,8 +38,26 @@
...
+class AttentionInfo:
+    layer_name: str
+    num_kv_heads: int
+    head_size: int
+    attn_type: str
+    sliding_window: Optional[int] = None
+    kv_sharing_target_layer_name: Optional[str] = None
+
+
 def shard_attention(layer: torch.nn.Module, mesh: Mesh,
```
yaochengji, line 51: I'm not sure `shard_attention` is still a good name now. It's basically wrapping attention module for tpu. cc @hfan

tpu_commons/models/jax/model_loader.py
Chenyaaang: It's using *args, so we can't use argnames here.

tpu_commons/models/vllm/sharding.py
Chenyaaang: I think it's fine, it takes in `mesh` and eventually being sharded here: https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/models/jax/attention.py#L24

PR #504, opened 2025-08-19
Archive RPAv2 and Delete Torchax-pt path.

QiliangCui:
# Description

1. Archive RPAvv2 
2. Delete Torchax-pt path but leave some Pallas attention files.

FIXES: b/438779127

# Tests

Waiting for 
https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2396

PR #512, opened 2025-08-19
[Torchax] Refactor torchax layers to use vLLM APIs

kyuyeunk:
# Description

Instead of overriding vLLM layers into a custom JAX layer, this PR aims to utilize existing vLLM layer but use their provided APIs and templates (such as `get_quantization_config` and `process_weights_after_loading`) to call our own JAX code.

This will loose some flexibility but has following benefits:
1. We don’t have to implement everything from scratch and able to leverage pre-existing vLLM APIs.
2. Plays more nicely with existing vLLM configs / features. Therefore, it helps with customer migration from vLLM GPU to vLLM TPU.

Using this change, I was able to implement FP8 model support for torchax with relative ease - which I'll create a follow-up PR for it.

# Tests

Ran following models and verified that performance has not changed

| model                                               | before | after |
| --------------------------------------------------- | ------ | ----- |
| Qwen/Qwen2.5-14B-Instruct                           | 18.36  | 18.37 |
| Qwen/Qwen2.5-32B                                    | 12.42  | 12.65 |
| Qwen/Qwen3-32B                                      | 12.07  | 12.14 |
| RedHatAI/Meta-Llama-3.1-70B-Instruct-quantized.w8a8 | 8.82   | 8.72  |
| google/gemma-3-27b-it                               | 19.22  | 19.19 |
| meta-llama/Llama-3.1-70B-Instruct                   | 7.55   | 7.55  |
| mistralai/Codestral-22B-v0.1                        | 14.94  | 14.98 |
| mistralai/Mistral-Small-24B-Instruct-2501           | 20.39  | 20.85 |
| mistralai/Mixtral-8x7B-Instruct-v0.1                | 18.51  | 18.61 |

# Checklist


kyuyeunk:
@hfan @yaochengji @vanbasten23 work in progress. Pinging to get some early feedback.

kyuyeunk:
> @hfan @yaochengji @vanbasten23 work in progress. Pinging to get some early feedback.

 \+ @yarongmu-google  @bythew3i @QiliangCui @lsy323

This PR is ready for review. This is a pretty big change so please feel free to request any additional test results or design discussion.

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
```diff
@@ -0,0 +1,370 @@
+import tempfile
+from typing import Optional
+
+from vllm.model_executor.layers.linear import (
+    ColumnParallelLinear, LinearBase, MergedColumnParallelLinear,
+    QKVParallelLinear, RowParallelLinear, UnquantizedLinearMethod)
+from vllm.model_executor.model_loader import get_model as vllm_get_model
```
vanbasten23: could you add the tests to the CI if they haven't?

vanbasten23:
Considering it's a large change, could you enable the torchax+jax_runner tests in the CI, make sure it pass, then merge? Afaik, the torchax+jax_runner tests are disabled in the CI today.

yarongmu-google:
Thanks @kyuyeunk. cc @bvrockwell this is the refactor that brings torchax much closer to vllm upstream to max reuse.

tpu_commons/attention/backends/pallas_torchax.py
```diff
@@ -168,110 +174,88 @@ def forward(
...
+    assert k.shape == v.shape
+    assert q_compute_dim == head_size * num_heads
+    assert k_compute_dim == head_size * num_kv_heads
+
+    # Convert the shapes from vLLM's convetion to what the attention function expects
+    # bs, num_heads, q_len, head_size
+    q = q.reshape(q_len, num_heads, head_size)
+    # bs, num_kv_heads, k_len, head_size
+    k = k.reshape(k_len, num_kv_heads, head_size)
+    v = v.reshape(k_len, num_kv_heads, head_size)
```
vanbasten23, line 145: Before the change, I see "# Skip this if sharing KV cache with an earlier attention layer.". Have you taken that into consideration in your change?

vanbasten23:
Just out of curiosity, do you know why we created our own layers such as JaxMergedColumnParallelLinear?

tpu_commons/models/vllm/jax_linear_common.py
```diff
@@ -124,29 +94,53 @@ def slice_sharded_tensor_for_concatenation(sharded_tensor: jax.Array,
...
+    tensor = t2j(tensor, use_dlpack=False)
+    if fused:
+        tensor = reorder_concatenated_tensor_for_sharding(
+            tensor, output_sizes, n_shards, 0)
+        tensor = jax.device_put(tensor, sharding)
+        param = torch.nn.Parameter(torch_view(tensor), requires_grad=False)
+    else:
+        tensors = []
+        start_offset = 0
+        for size in output_sizes:
```
vanbasten23, line 132: I wonder which test exercises the linear layer for the SPMD multi-chip case

tpu_commons/models/vllm/quantization/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    method_to_config: dict[str, str] = {
+        None: JaxUnquantizedConfig,
+        "compressed-tensors": JaxCompressedTensorsConfig,
+    }
+
+    if model_config.quantization not in method_to_config:
+        raise NotImplementedError
+    quant_config = method_to_config[model_config.quantization]
+    assert issubclass(quant_config, JaxCommonConfig)
+    JaxCommonConfig.vllm_config = vllm_config
```
vanbasten23: Not sure if I understand this part. JaxCommonConfig is a class?

tpu_commons/models/vllm/quantization/unquantized.py
```diff
@@ -0,0 +1,158 @@
...
+        if isinstance(layer, LinearBase):
+            return JaxUnquantizedLinearMethod(self, layer)
+        if isinstance(layer, FusedMoE):
+            return UnquantizedFusedMoEMethod
+        if isinstance(layer, Attention):
+            return None
+        return None
+
+
+class JaxUnquantizedLinearMethod(UnquantizedLinearMethod):
```
vanbasten23, line 63: could you please point me to the tests that exercise this part of code, both for single chip and multi-chip?

tpu_commons/models/vllm/sharding.py
```diff
@@ -97,11 +27,6 @@ def shard_fused_moe(layer: torch.nn.Module, mesh: Mesh,
 
 
 MODULE_TYPE_TO_WRAPPING_FUNC = {
-    VllmAttention: shard_attention,
-    QKVParallelLinear: shard_qkv_parallel_linear,
-    MergedColumnParallelLinear: shard_merged_column_parallel_linear,
-    ColumnParallelLinear: shard_column_parallel_linear,
-    RowParallelLinear: shard_row_parallel_linear,
     FusedMoE: shard_fused_moe,
```
vanbasten23, line 31: why is FusedMoE left here?

tpu_commons/models/vllm/sharding.py
```diff
@@ -154,33 +79,71 @@ def _move_to_tpu_replicated(x):
...
+            elif k in module._buffers:
+                buffers[qual_name] = v
+            elif isinstance(v, torch.Tensor):
+                variables[qual_name] = v
+
+        for name, child in module.named_children():
+            extract_one(child, prefix + name + '.')
+
+    extract_one(m, '')
+    return params, buffers, variables
```
vanbasten23, line 132: I wonder why you need the `variable`

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -179,7 +185,11 @@ def step_fun(
 
             return new_kv_caches, hidden_states
 
-        return step_fun
+        def wrapper(*args, **kwargs):
```
vanbasten23: Could you point me to the place where the linear weights are sharded? Before the change, I see
```
layer.weight = w.apply_jax_(jax.device_put,
                                NamedSharding(mesh, P('model', None)))
```
But I didn't find it after the change.

kyuyeunk: Sure! This happens inside of [process_weights_after_loading](https://github.com/vllm-project/tpu_commons/blob/abe3c957ee529659027f87ba3639fdd41531ad10/tpu_commons/models/vllm/quantization/unquantized.py#L94) of each quantization methods. 

This is the function that vLLM calls for each layer after creating their respective weights.

tpu_commons/models/vllm/sharding.py
kyuyeunk: So there are 3 types of `torch.Tensor`s that we need to keep track of: `torch.nn.parameter.Parameter`, `torch.nn.parameter.Buffer`, and just a regular member variable. (like `layer.scale = torch.empty...`).

First two needs to be passed as arguments when calling forward function, but the latter does not.

However, we still need to track it & convert variables into `torchax.tensor.Tensor`, and replace existing `torch.Tensor` so torchax can use it properly.

tpu_commons/models/vllm/sharding.py
kyuyeunk: This is simply because I'm not familiar with vLLM's MoE interface. I plan to revisit it in the future and apply similar treatment as other layers.

Left a todo.

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: Most (if not all) tests in `test_unquantized.py` calls this part of the code.

I've added a check like `isinstance(obj, JaxUnquantizedLinearMethod)` to make this more explicit: [test_unquantized.py](https://github.com/vllm-project/tpu_commons/blob/597b58a78adde9945f00c27406d096746942e661/tests/models/vllm/quantization/test_unquantized.py#L153-L154).

Also, how can I add a different test for single chip & multi-chip?

tpu_commons/models/vllm/quantization/__init__.py
kyuyeunk: It's a class that's inherited by every jax quantization configuration class. I've using class variables to sent some variables to jax quantization configuration.

It's a bit hacky way of doing this, but this was done so because the initialization of a config object happens outside of tpu_commons (inside vLLM code), I needed a way to send some extra arguments to jax quantization config.

tpu_commons/attention/backends/pallas_torchax.py
kyuyeunk: I haven't looked into this but this new code is mostly a copy-paste from JAX attention layer.

My guess is that previous code is for RPA v2 - where KV cache writing logic is separate from the kernel. With RPA v3 kernel, this logic is integrated into the kernel and we have not added a functionality to skip KV cache writing if a certain condition is met. cc: @bythew3i

tpu_commons/models/vllm/jax_linear_common.py
kyuyeunk: All the tests for calling quantized/unquantized linear layers are present in `tests/models/vllm/quantization`. Can you point me to an example of a test that uses SPMD multi-chip case so I can add it as well?

vanbasten23:
> > Considering it's a large change, could you enable the torchax+jax_runner tests in the CI, make sure it pass, then merge? Afaik, the torchax+jax_runner tests are disabled in the CI today.
> 
> @vanbasten23 sgtm. Dumb question: can you point me to which file I need to change to enable this?

IIUC it's here https://github.com/vllm-project/tpu_commons/blob/83f0e0a8e9b86efb3f0d400f65b93542c8e386d1/.buildkite/pipeline_jax.yml#L34-L42. Probably also check if the tests have already been added to the CI.

tpu_commons/models/vllm/jax_linear_common.py
vanbasten23: I'm not quite sure. Perhaps the linear layer tests in https://github.com/vllm-project/tpu_commons/blob/0d6f115d311dc3540c2deb7a0c1010aefb8b9abc/tests/models/vllm/ cc @hfan @lsy323 to confirm.

tpu_commons/models/vllm/quantization/unquantized.py
vanbasten23: it seems we are doing this way https://github.com/vllm-project/tpu_commons/blob/0d6f115d311dc3540c2deb7a0c1010aefb8b9abc/tests/models/vllm/test_jax_row_parallel_linear.py#L53

tpu_commons/models/vllm/vllm_model_wrapper.py
vanbasten23: then I guess at the time we shard the weight, the weight is still on cpu. When do we move it to tpu?

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: Done. Added a single chip and multi-chip unit tests.

tpu_commons/models/vllm/vllm_model_wrapper.py
kyuyeunk: `jax.device_put` also takes care of putting the weights to TPU. To add more detail, `mesh: jax.sharding.Mesh` also contains information about hardware (https://docs.jax.dev/en/latest/jax.sharding.html#jax.sharding.Mesh). Meaning, when we call `jax.device_put`, the weights are put into TPUs in a sharded manner.

kyuyeunk:
> Just out of curiosity, do you know why we created our own layers such as JaxMergedColumnParallelLinear?

I think it's just boils down to a flexibility that writing our own layer provides. Even for my own experience, making our features work well with existing vLLM infrastructure was quite tricky. But as I tried to add more and more features, I found that writing our own implementation was not scalable and it was better to utilize pre-existing vLLM infrastructure.

kyuyeunk:
> > > Considering it's a large change, could you enable the torchax+jax_runner tests in the CI, make sure it pass, then merge? Afaik, the torchax+jax_runner tests are disabled in the CI today.
> > 
> > 
> > @vanbasten23 sgtm. Dumb question: can you point me to which file I need to change to enable this?
> 
> IIUC it's here
> 
> https://github.com/vllm-project/tpu_commons/blob/83f0e0a8e9b86efb3f0d400f65b93542c8e386d1/.buildkite/pipeline_jax.yml#L34-L42
> 
> . Probably also check if the tests have already been added to the CI.

Thanks. Enabled it. here: https://github.com/vllm-project/tpu_commons/blob/758117fb488176bd108d2d9f9891a8c8d21b5b05/.buildkite/pipeline_jax.yml#L34-L42

Waiting for CI results: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2215

kyuyeunk:
Verified that CI is running correctly except for torchax-pt path. But given that there is an ongoing work to archive torchax-pt path (https://github.com/vllm-project/tpu_commons/pull/504), I think it can safely be ignored: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2280#0198e50f-f3ee-4606-aedf-2e4216e6894b

@hfan, @lsy323 can you review this PR? It's blocking many upcoming features (like fp8 support, kv cache quantization, attention quantization), so it's a high priority to merge these changes.

.buildkite/pipeline_jax.yml
```diff
@@ -31,15 +31,15 @@ steps:
...
-  #      - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/mlperf.sh
+   - label: "E2E MLPerf tests for JAX + vLLM models"
+     key: test_3
+     soft_fail: true
+     env:
+       MODEL_IMPL_TYPE: "vllm"
+     agents:
+       queue: tpu_v6e_queue
+     commands:
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/mlperf.sh
```
hfan: Thanks for trying to uncomment it, I also noticed it and fixed in: https://github.com/vllm-project/tpu_commons/pull/574

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
```diff
@@ -0,0 +1,407 @@
...
+        max_model_len=64,
+        max_num_batched_tokens=64,
+        max_num_seqs=4,
+    )
+    vllm_config = engine_args.create_engine_config()
+    vllm_config.model_config.dtype = torch.bfloat16
+    vllm_config.quant_config = get_tpu_quantization_config(vllm_config, mesh)
+    vllm_config.device_config.device = "cpu"
+
+    with patch("torch._sync", return_value=None):
```
hfan: OOC, why is the `patch()` needed? The model `RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w8a8"` can be loaded in the prod code without the `patch()`

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
```diff
@@ -0,0 +1,407 @@
...
+        assert isinstance(jax_row_linear.quant_method,
+                          CompressedTensorsLinearMethod)
+        assert isinstance(jax_row_linear.scheme, JaxCompressedTensorsW8A8Int8)
+        jax_row_linear.quant_method.process_weights_after_loading(
+            jax_row_linear)
+        jax_output = jax_row_linear(jax_input_tensor)
+        jax_output = j2t(jax_output.to(torch.float32)).to(dtype)
+
+    # Call reference w8a8 int8
+    output = ref_w8a8_int8(
```
hfan: Is there a way we can just directly call vllm's layer on CPU for the reference output, like in the original tests?

tests/models/vllm/quantization/utils.py
```diff
@@ -0,0 +1,19 @@
+import jax
```
hfan: How about merge the content of this file into `tests/models/vllm/utils.py`? There is nothing about quantization in this file.

tpu_commons/attention/backends/pallas_torchax.py
```diff
@@ -164,114 +79,79 @@ def forward(
...
-                        kv_cache,
-                        num_slices,
-                        page_size=block_size)
-    kv_cache = kv_cache.reshape(num_blocks, block_size, num_combined_kv_heads,
-                                head_size)
-    return kv_cache
+                "fused output quantization is not yet supported for"
+                " PallasAttentionBackendImpl")
+
+        del kv_cache  # Use kv_cache from vllm wrapper context values instead.
```
hfan, line 96: is vllm still going to allocate hbm for this unused `kv_cache`?

tests/models/vllm/test_torchax_tpu_loader.py
```diff
@@ -105,13 +68,6 @@ def test_tpu_model_loader(model):
                               model_config=vllm_config.model_config,
                               mesh=mesh)
 
-    if mesh is not None:
-        for _, module in model.named_modules():
-            fqn = module.__class__.__qualname__
-            assert fqn not in REPLACED_MODULE_NAMES, \
-                f"Module {fqn} should be replaced by JAX version, " \
-                "please check the TPU backend configuration."
-
```
hfan: this is the torchax_runner (for torchax pt), we might want to keep it, until torchax_pt is removed? Tagging @lsy323 for suggestion

tests/models/vllm/test_torchax_tpu_loader.py
```diff
@@ -51,43 +51,6 @@ def _setup_environment(model):
...
-    if mesh is not None:
-        for _, module in model.model.named_modules():
-            fqn = module.__class__.__qualname__
-            assert fqn not in JAX_REPLACED_MODULE_NAMES, \
-                f"Module {fqn} should be replaced by JAX version, " \
-                "please check the TPU backend configuration."
-
-    # Need to manually delete the JAX arrays otherwise they will stay on TPU.
-    for v in params.values():
-        v.delete()
```
hfan: Since the JaxFusedMoE layer is still left for now, could we still keep this test logic, until MoE is also unified with vLLM layer? We can just remove all the layers except the `FusedMoE` in `JAX_REPLACED_MODULE_NAMES`

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -192,7 +196,9 @@ def compute_logits_func(
             params_and_buffers: Any,
             hidden_states: jax.Array,
         ) -> jax.Array:
-            with torchax.default_env():
+            with torchax.default_env(), set_vllm_model_wrapper_context(
+                    kv_caches=None, mesh=self.mesh), set_forward_context(
+                        attn_metadata=None, vllm_config=self.vllm_config):
```
hfan: do you need add `set_vllm_model_wrapper_context` and `set_forward_context`?

tpu_commons/models/vllm/sharding.py
hfan: OOC, which vllm layers have these variables? I'm just curious how come we didn't notice it until now

tpu_commons/models/vllm/jax_linear_common.py
```diff
@@ -76,36 +57,25 @@ def reorder_concatenated_tensor_for_sharding(concatenated_tensor: jax.Array,
...
-                                     axis=dim)
-            split_tensors_for_this_shard.append(t)
-        tensor_shard = jnp.concatenate(split_tensors_for_this_shard, axis=dim)
-        reordered_tensor_shards.append(tensor_shard)
-
-    reordered_tensor = jnp.concatenate(reordered_tensor_shards, axis=dim)
-    return reordered_tensor
+    # While maintaining 0th dim as a shard dim, we concatenate along 1th dim to
+    # to create concatenated tnensor where 0th dim maps to shard dim.
+    reordered_tensor = jnp.concatenate(split_tensors, axis=dim + 1)
```
hfan: this depends on the size of the `dim` is same for all the split_tensors? For example, if the split tensors are of shape: [4, 2] and [6, 2], and the dim=1, this trick does not work?

hfan: Eitherway, this change probably can be in its own PR

tpu_commons/models/vllm/jax_linear_common.py
```diff
@@ -124,29 +94,53 @@ def slice_sharded_tensor_for_concatenation(sharded_tensor: jax.Array,
...
+        # Therefore, splitting happens locally.
+        split_tensor = sharded_tensor[..., start_offset:end_offset]
+        split_tensors.append(split_tensor.reshape(new_shape[:-2] + (-1, )))
         start_offset = end_offset
-        split_tensor = shard_map(_get_slice_on_shard_bound,
-                                 mesh=mesh,
-                                 in_specs=(sharding_spec),
-                                 out_specs=(sharding_spec),
-                                 check_rep=False)(sharded_tensor)
-        split_tensors.append(split_tensor)
```
hfan: Have you tested that sharding is expected, and the performance is the same for the models relies on the weight reordering? I wasn't sure if this was the case, and just opt to use `shard_map` in which I can manually specify the output sharding.

Similar to above, this change can be together put in their own PR. The "Coding Guidelines for uLLM" doc perfers small PRs

tpu_commons/models/vllm/quantization/unquantized.py
```diff
@@ -0,0 +1,158 @@
...
+                self.output_sizes = layer.output_sizes
+
+            self.fuse_matmuls = get_model_matmul_fusion_assignment(
+                vllm_config.model_config.model,
+                vllm_config.scheduler_config.max_num_batched_tokens,
+                vllm_config.parallel_config.tensor_parallel_size,
+                layer._get_name())
+
+        self.bias_sharding = P(self.weight_sharding[0])
+        self.n_shards = self.mesh.shape.get(self.weight_sharding[0], 1)
```
hfan: is it correct for `RowParallelLinear` where `self.weight_sharding[0]` is None?

tpu_commons/models/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
```diff
@@ -0,0 +1,119 @@
...
+            if self.fuse_matmuls:
+                out = self._apply_fused(layer, x, bias)
+            else:
+                out = self._apply_split(layer, x, bias)
+
+        if self.enable_sequence_parallelism:
+            token_num = x.shape[0]
+            # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
+            if token_num // self.mesh.shape['model'] >= TPU_SECOND_LAST_MINOR:
+                out.shard_(NamedSharding(self.mesh, P('model', None)))
```
hfan: for quantized matmul, this shouldn't be applied, because XLA don't know how to overlap quant matmul kernel and collectives

cc @yaochengji

tpu_commons/models/vllm/quantization/unquantized.py
```diff
@@ -0,0 +1,158 @@
...
+            if self.fuse_matmuls:
+                out = self._apply_fused(layer, x, bias)
+            else:
+                out = self._apply_split(layer, x, bias)
+
+        if self.enable_sequence_parallelism:
+            token_num = x.shape[0]
+            # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
+            if token_num // self.mesh.shape['model'] >= TPU_SECOND_LAST_MINOR:
+                out.shard_(NamedSharding(self.mesh, P('model', None)))
```
hfan: For qkv merged and col merged parallel linear layers, the sharding is applied on the input before the matmul is done.

cc @yaochengji

tpu_commons/models/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
```diff
@@ -0,0 +1,119 @@
...
+
+class JaxCompressedTensorsW8A8Int8(CompressedTensorsW8A8Int8):
+
+    def __init__(self, strategy: str, is_static_input_scheme: bool,
+                 input_symmetric: bool, quant_config: JaxCommonConfig,
+                 layer: LinearBase):
+        assert isinstance(quant_config, JaxCommonConfig)
+        assert isinstance(layer, LinearBase)
+        super().__init__(strategy, is_static_input_scheme, input_symmetric)
+        JaxUnquantizedLinearMethod.__init__(self, quant_config, layer)
```
hfan: nit: it might be better to refactor the logic in `JaxUnquantizedLinearMethod.__init__` to a common initializer function that assigns the common members for both `JaxUnquantizedLinearMethod` and `JaxCompressedTensorsW8A8Int8`.

This if for future bug proof, as `JaxCompressedTensorsW8A8Int8` does not inherit from `JaxUnquantizedLinearMethod`, changes in latter could break former.

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: Fixed it!

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: if `self.weight_sharding[0]=None`, we call `get(None, 1)`. Meaning, `self.n_shards=1`. Which is the correct value we want to use.

tpu_commons/models/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
kyuyeunk: Removed!

tpu_commons/models/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
kyuyeunk: Done. I've created separate config class `JaxCommonLinearConfig` that stores all configurations.

tpu_commons/models/vllm/jax_linear_common.py
kyuyeunk: This should work even if dim is different for all split tensors. But I do agree that it can be it's own PR. 

I've moved this logic to a separate PR and built this PR on top of it: https://github.com/vllm-project/tpu_commons/pull/590

tpu_commons/models/vllm/jax_linear_common.py
kyuyeunk: Done. As mentioned above, create separate PR. I've also benchmarked and verified that numerics and performance is correct.

tpu_commons/attention/backends/pallas_torchax.py
kyuyeunk: No, vLLM's kv cache is initialized as an empty tensor: https://github.com/vllm-project/vllm/blob/5c4b6e66fec0a59f702e32ec5c476cdd8a800566/vllm/attention/layer.py#L212-L218

Only when `bind_kv_cache` is called like here it gets allocated: https://github.com/vllm-project/tpu_commons/blob/d86751a4881c73fee4b70731aba67d68dc1db438/tpu_commons/runner/tpu_torchax_runner.py#L1074-L1079

However, since this function never gets called for torchax-jax backend, it is kv cache is never allocated.

tests/models/vllm/test_torchax_tpu_loader.py
kyuyeunk: I'm hoping https://github.com/vllm-project/tpu_commons/pull/504 - which deletes all torchax-pt code - to be submitted before this PR so I don't have to worry about this PR affecting torchax-pt. @QiliangCui, do you know when the PR will be submitted?

tests/models/vllm/quantization/utils.py
kyuyeunk: Slight complication was that it was difficult to import a file from a parent directory. (`import tpu_commons.tests.models.vllm.utils` did not work). I've refactored the directory so all files that uses `utils` are in the same directory.

tests/models/vllm/test_torchax_tpu_loader.py
kyuyeunk: Done.

tpu_commons/models/vllm/vllm_model_wrapper.py
kyuyeunk: If we don't add `set_forward_context`, this part of vLLM code complains that forward context is not set: https://github.com/vllm-project/vllm/blob/585e0bde36abdb2ab2967fd42005cbe62459020e/vllm/attention/layer.py#L268

I've considered wrapping `set_foward_context` into `set_vllm_model_wrapper_context` so that a single call can set both `vllm_model_wrapper_context` and `forward_context`. But I wasn't sure if it's a good UI design so I left it as-is.

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
kyuyeunk: This code was copy pasted from other unit tests, but I believe the `patch` is used to skip weight loading process to make the unit test lighter.

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
kyuyeunk: There is, but this was mostly for future proofing. There are some features that isn't even supported in vLLM CPU & only supported in vLLM GPU. For instance, when I was working on FP8, I've learned that vLLM's w8a8 FP8 is only available on GPU and vLLM falls back to w8a16 if no GPU is detected. So I wanted to create a unit test that does not rely on vLLM's layer.

kyuyeunk:
Referenced https://github.com/QiliangCui/bm-infra/blob/main/cases/hourly_torchax_jax.csv and benchmarked following models. Confirmed that the performance has not changed.

| model                                               | before | after |
| --------------------------------------------------- | ------ | ----- |
| Qwen/Qwen2.5-14B-Instruct                           | 18.36  | 18.37 |
| Qwen/Qwen2.5-32B                                    | 12.42  | 12.65 |
| Qwen/Qwen3-32B                                      | 12.07  | 12.14 |
| RedHatAI/Meta-Llama-3.1-70B-Instruct-quantized.w8a8 | 8.82   | 8.72  |
| google/gemma-3-27b-it                               | 19.22  | 19.19 |
| meta-llama/Llama-3.1-70B-Instruct                   | 7.55   | 7.55  |
| mistralai/Codestral-22B-v0.1                        | 14.94  | 14.98 |
| mistralai/Mistral-Small-24B-Instruct-2501           | 20.39  | 19.72 |
| mistralai/Mixtral-8x7B-Instruct-v0.1                | 18.51  | 18.61 |

cc: @hfan

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
hfan: Could you provide a pointer to the existing tests? I don't think it's needed. I know for loading the random weight, I added a `patch("torch._sync")` to get around PTXLA.

tpu_commons/attention/backends/pallas_torchax.py
hfan: Makes sense. Could you add an assert to make sure the kv_cache to be deleted is indeed empty?

tests/models/vllm/test_torchax_tpu_loader.py
hfan: SG, #504 is already merged.

tpu_commons/models/vllm/vllm_model_wrapper.py
hfan: what I'm trying to say is that here it only calls this function for `compute_logits`, it doesn't call Attention, so I thought both types of context are not needed.

tpu_commons/models/vllm/quantization/unquantized.py
hfan: but then for RowParallelLinear, the output is need to be sharded as you originally had. We need to handle both Row and Col cases.

hfan:
> Referenced https://github.com/QiliangCui/bm-infra/blob/main/cases/hourly_torchax_jax.csv and benchmarked following models. Confirmed that the performance has not changed.
> 
> model	before	after
> Qwen/Qwen2.5-14B-Instruct	18.36	18.37
> Qwen/Qwen2.5-32B	12.42	12.65
> Qwen/Qwen3-32B	12.07	12.14
> RedHatAI/Meta-Llama-3.1-70B-Instruct-quantized.w8a8	8.82	8.72
> google/gemma-3-27b-it	19.22	19.19
> meta-llama/Llama-3.1-70B-Instruct	7.55	7.55
> mistralai/Codestral-22B-v0.1	14.94	14.98
> mistralai/Mistral-Small-24B-Instruct-2501	20.39	19.72
> mistralai/Mixtral-8x7B-Instruct-v0.1	18.51	18.61
> cc: @hfan

Thanks! The performance diff for `mistralai/Mistral-Small-24B-Instruct-2501` seems bit larger than noise? Is it reproducible?

tpu_commons/models/vllm/vllm_model_wrapper.py
kyuyeunk: Oh yeah. You were right. Deleted `set_forward_context`.

tpu_commons/models/vllm/sharding.py
kyuyeunk: Things like QKV quantization scale is one of those variables: https://github.com/vllm-project/vllm/blob/5da4f5d857933329aaca779e3a81f1385c84e34a/vllm/attention/layer.py#L219-L221

If I understand correctly, this wasn't an issue when we replaced the entire Attention layer with our custom JAX attention layer. But because we are using vLLM's version, we have to be thorough with all the variables.

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: btw, I'm slightly confused what is the difference between sharding input vs. output. After sharding propagation, isn't the end result essentially the same?

Here is an example of `RowParallelLinear`:

```
# RowParallelLinear - Shard input

in # ('model', None) 
weight # (None, 'model') 
out  # not specified

After sharding propagation, out is sharded as ('model', None)

in # ('model', None) 
weight # (None, 'model')
weight = allgather(weight, 'model') # (None, None)
out = in * weight # ('model', None)


# RowParallelLinear - Shard output

in # not specified
weight # (None, 'model') 
out  # ('model', None) 

After sharding propagation, in is sharded as ('model', None)

in # ('model', None)
weight # (None, 'model')
weight = allgather(weight, 'model') # (None, None)
out = in * weight # ('model', None)
```

And here is the scenario for `ColumnParallelLinear`:

```
# ColumnParallelLinear - Shard input

in # ('model', None) 
weight # ('model', None) 
out  # not specified

After sharding propagation, out is sharded as ('model', None)

in # ('model', None) 
weight # ('model', None)
weight = allgather(weight, 'model') # (None, None)
out = in * weight # ('model', None)


# ColumnParallelLinear - Shard output

in # not specified
weight # ('model', None)
out  # ('model', None) 

After sharding propagation, in is sharded as ('model', None)

in # ('model', None)
weight # ('model', None)
weight = allgather(weight, 'model') # (None, None)
out = in * weight # ('model', None)
```

Things might differ based on how previous / next layers are sharded, but that's my understanding. @yaochengji, do you have any insights? Or some unit tests to verify sharding?

tpu_commons/attention/backends/pallas_torchax.py
kyuyeunk: Done.

tests/models/vllm/quantization/test_compressed_tensors_w8a8_int8.py
kyuyeunk: Ah okay, I think I was mistaken, I think I just copy pasted this code but copied assumed `use_random_weights` was set to `True`: https://github.com/vllm-project/tpu_commons/blob/49893d80302d4e6f9d14889ff736eead562619a0/tpu_commons/models/vllm/vllm_model_wrapper.py#L118-L125

I've deleted the code.

tpu_commons/models/vllm/jax_linear_common.py
kyuyeunk: Added.

tpu_commons/models/vllm/quantization/__init__.py
kyuyeunk: Updated the code.

kyuyeunk:
Confirmed that CI passes: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2429#0198ee7d-539b-4556-8db4-315e40888e7f

tpu_commons/models/vllm/vllm_model_wrapper.py
hfan: Can the `set_vllm_model_wrapper_context` be removed too? The code also didn't have `set_vllm_model_wrapper_context` before the change.

tpu_commons/models/vllm/quantization/unquantized.py
hfan: My understanding is that @yaochengji set the input and output sharding so that XLA can pattern match the collective and matmul, but for Row and Col parallel linear layers, they are different.

Yea, for this part of code, make sure Chengji approves.

tpu_commons/models/vllm/quantization/unquantized.py
```diff
@@ -0,0 +1,149 @@
...
+              x: torch.Tensor,
+              bias: Optional[torch.Tensor] = None) -> torch.Tensor:
+        with jax.named_scope(layer._get_name()):
+            if self.jax_config.enable_sequence_parallelism:
+                token_num = x.shape[0]
+                # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
+                if token_num // self.jax_config.mesh.shape[
+                        'model'] >= TPU_SECOND_LAST_MINOR:
+                    x.shard_(
+                        NamedSharding(self.jax_config.mesh, P('model', None)))
```
hfan: I can approve this PR except these XLA collective matmul handling.

Will leave it to @yaochengji

tpu_commons/models/vllm/quantization/unquantized.py
yaochengji: @kyuyeunk I'm afraid actually we don't apply all-gather on weight for RowParallelLinear or ColumnParallelLinear.

For `ColumnParallelLinear`, it should be 
```
in # ('model', None)
gathered_in # (None, None)
weight # (None, 'model')
output # (None, 'model')
```

For `RowParallelLinear`, it should be 

```
in # (None, 'model')
weight # (None, 'model')
output_partial
output_after_reduce_scatter # ('model', None)
```

yaochengji: That's why I added the sharding annotation of  ('model', None) on the input of `ColumnParallelLinear` and output of `RowParallelLinear`.

Also, pls note that the output of  `ColumnParallelLinear` might differ when it's a fused matmul, that's why I don't put annotation on it. Instead I let the sharding propagation in XLA compiler figure it out.

kyuyeunk: Thanks for the comment. I have modified the code to shard input for `ColumnParallelLinear` and output for `RowParallelLinear` when sequence parallelism is enabled. PTAL.

PR #514, opened 2025-08-19
[Spec Decoding] Use correct shape for logit positions.

py4:
# Description

The number of speculative tokens can be more than total number of requests. We cannot use `num_padded_reqs` to use for logit positions. Because logit positions can go up to `num_reqs * (num_speculative_tokens + 1)`.
But this is not the case for bonus tokens. Each request, has only one bonus tokens. 

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -10,9 +10,31 @@ import jax
 import jax.numpy as jnp
 import jaxtyping
 import numpy as np
-import vllm.envs as envs
 from flax import nnx
 from jax.sharding import NamedSharding, PartitionSpec
+
+import vllm.envs as envs
+from tpu_commons import utils as common_utils
+from tpu_commons.logger import init_logger
+from tpu_commons.models.jax.attention_metadata import AttentionMetadata
+from tpu_commons.models.jax.common.sharding import build_mesh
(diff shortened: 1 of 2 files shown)


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1244,20 +1258,25 @@ def _get_spec_decode_metadata(
...
         ])
+
+        assert bonus_logits_indices.shape[0] <= padded_num_reqs, (
+            f"bonus_logits_indices.shape[0]={bonus_logits_indices.shape[0]} "
+            f"padded_num_reqs={padded_num_reqs}")
+
         padded_bonus_logits_indices = np.concatenate([
             bonus_logits_indices,
-            np.zeros(padded_logits_length - bonus_logits_indices.shape[0],
+            np.zeros(padded_num_reqs - bonus_logits_indices.shape[0],
```
Lumosis, line 1274: Have you tested this part with spec decoding enabled? I vaguely remembered we need to keep the padding length consistent. Otherwise, there might be some mismatch during sampling. (I could be wrong)

py4: @Lumosis Yeah I tested it on 1000 examples. The reason is that we have 1 bonus token per request, so the correct padding for `bonus_logits_indices` is up to `padded_num_reqs`

PR #520, opened 2025-08-20
Support sequence parallelism and collective matmul

yaochengji:
# Description

- add sequence parallelism sharding annotation for torchax dense models
- fix the issue of v6e-4, v6e-8 logical device id order
- the feature can be enabled by `-O '{"pass_config": {"enable_sequence_parallelism": true}}'`

# Tests

```
MODEL_IMPL_TYPE=vllm vllm serve Qwen/Qwen2.5-32B  --seed 42  --disable-log-requests  --tensor-parallel-size 8 --max-model-len 2048 --gpu-memory-utilization 0.96 --no-enable-prefix-caching --max-num-seqs 256 --max-num-batched-tokens 4096 -O '{"pass_config": {"enable_sequence_parallelism": true}}' |& tee run.log
```
```
python3 ./benchmarks/benchmark_serving.py --model Qwen/Qwen2.5-32B --dataset-name sonnet --dataset-path benchmarks/sonnet_4x.txt --sonnet-input-len 1800 --sonnet-output-len 128 --ignore_eos
```

```
MODEL_IMPL_TYPE=vllm vllm serve meta-llama/Llama-3.1-70B-Instruct  --seed 42  --disable-log-requests  --tensor-parallel-size 8 --max-model-len 2048 --gpu-memory-utilization 0.96 --no-enable-prefix-caching --max-num-seqs 128 --max-num-batched-tokens 2048 -O '{"pass_config": {"enable_sequence_parallelism": true}}' |& tee run.log
```

```
python3 ./benchmarks/benchmark_serving.py --model meta-llama/Llama-3.1-70B-Instruct --dataset-name sonnet --dataset-path benchmarks/sonnet_4x.txt --sonnet-input-len 1800 --sonnet-output-len 128 --ignore_eos
```

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -71,14 +71,9 @@ steps:
        - |
          .buildkite/scripts/run_in_docker.sh \
            python3 -m pytest -s -v /workspace/tpu_commons/tests/ \
-           --ignore=/workspace/tpu_commons/tests/models/jax/test_phi3.py \
-           --ignore=/workspace/tpu_commons/tests/models/jax/test_qwen2.py \
-           --ignore=/workspace/tpu_commons/tests/models/jax/test_qwen3.py \
-           --ignore=/workspace/tpu_commons/tests/models/jax/test_llama3.py \
-           --ignore=/workspace/tpu_commons/tests/models/jax/test_qwen2_5_vl.py \
            --ignore=/workspace/tpu_commons/tests/kernels \
+           --ignore=/workspace/tpu_commons/tests/e2e \
            --ignore=/workspace/tpu_commons/tests/models/vllm \
-           --ignore=/workspace/tpu_commons/tests/e2e/test_multi_modal_inference.py \
            --cov-config=/workspace/tpu_commons/.coveragerc --cov tpu_commons --cov-report term-missing --cov-fail-under=69
 
    - label: "JAX unit tests - kernels"
--- a/tests/models/jax/test_model_loader.py
+++ b/tests/models/jax/test_model_loader.py
@@ -1,93 +1,17 @@
 import os
-import sys
-from types import ModuleType
 from unittest.mock import MagicMock, patch
 
 import jax
 import jax.numpy as jnp
 import pytest
 import torch
-from flax import nnx
 from jax.sharding import Mesh
 from transformers import PretrainedConfig
-from vllm.config import VllmConfig
+from vllm.config import ModelConfig, VllmConfig
 from vllm.engine.arg_utils import EngineArgs
 
-
-class MockCausalLM(nnx.Module):
-    """A mock nnx.Module that mimics the behavior of a causal language model."""
-
-    def __init__(self,
-                 vllm_config: VllmConfig,
-                 rng: jax.Array,
-                 mesh: Mesh,
-                 force_random_weight: bool = False):
-        """Initializes a dummy parameter."""
-        # Using the inputs to show they are passed correctly
-        self.config = vllm_config
-        self.mesh = mesh
-        # Create a dummy parameter to ensure weight loading works
-        self.w = nnx.Param(jax.random.normal(rng, (4, 4)))
-        self.hidden_dim = 128
-        self.vocab_size = 256
-
-    def load_weights(self, rng: jax.Array):
-        """Simulates loading weights by setting the parameter to a known value."""
-        self.w.value = jnp.ones_like(self.w.value)
-
-    def __call__(self, input_ids: jax.Array, positions: jax.Array,
-                 kv_cache: jax.Array) -> tuple[jax.Array, jax.Array]:
-        """Simulates a forward pass."""
-        # Simulate some work and return dummy outputs with expected shapes/types
-        batch_size, seq_len = input_ids.shape
-        new_kv_cache = kv_cache + jnp.sum(self.w.value)  # Dummy op
-        hidden_states = jnp.ones((batch_size, self.hidden_dim))
-        return new_kv_cache, hidden_states
-
-    def compute_logits(self, hidden_states: jax.Array) -> jax.Array:
-        """Simulates computing logits from hidden states."""
-        batch_size = hidden_states.shape[0]
-        logits = jnp.ones((batch_size, self.vocab_size))
-        return logits * jnp.mean(self.w.value)  # Dummy op
-
-
-@pytest.fixture(scope="session", autouse=True)
-def mock_dependencies(request):
-    """
-    This autouse fixture runs for the entire test session. It mocks modules
-    that are imported locally within the functions of model_loader.py.
-    This ensures that the module under test can be imported and used
-    without having the actual dependencies present.
-    """
-    # Create mock modules for the models
-    mock_llama_module = ModuleType("tpu_commons.models.jax.llama3")
-    setattr(mock_llama_module, "LlamaForCausalLM", MockCausalLM)
-
-    mock_qwen2_module = ModuleType("tpu_commons.models.jax.qwen2")
-    setattr(mock_qwen2_module, "Qwen2ForCausalLM", MockCausalLM)
-
-    # Create a mock for the logger
-    mock_logger_module = ModuleType("tpu_commons.logger")
-    mock_logger = MagicMock()
-    setattr(mock_logger_module, "init_logger",
-            MagicMock(return_value=mock_logger))
-
-    # Add the mock modules to sys.modules
-    original_modules = sys.modules.copy()
-    sys.modules["tpu_commons.models.jax.llama3"] = mock_llama_module
-    sys.modules["tpu_commons.models.jax.qwen2"] = mock_qwen2_module
-    sys.modules["tpu_commons.logger"] = mock_logger_module
-
-    # Allow the tests to import the module under test
-    global model_loader
-    from tpu_commons.models.jax import model_loader
-
-    # Teardown: restore original sys.modules
-    def fin():
-        sys.modules.clear()
-        sys.modules.update(original_modules)
-
-    request.addfinalizer(fin)
+from tpu_commons.models.jax import model_loader
+from tpu_commons.models.jax.qwen3 import Qwen3ForCausalLM
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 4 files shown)


yaochengji:
| Model    | Best_throuput_before | Throuput_new| Improvement|
| -------- | ----------------------|----------------|-------------|
|  Qwen/Qwen2.5-32B| 12.1req/s    |14.4req/s| 18.9%|
|  Llama-3.1-70B| 7.58req/s     |8.1req/s|6.9%|

yaochengji:
cc @hfan could you also take a look?

cc @xiangxu-google , you may also apply this on jax models.

tpu_commons/utils.py
```diff
@@ -106,3 +112,65 @@ def get_padded_num_heads(num_heads: int, sharding_size: int) -> int:
...
+                    devices[4],
+                    devices[6],
+                    devices[7],
+                    devices[5],
+                    devices[3],
+                    devices[1],
+                ])
+                ordered_devices = np.array(ordered_devices)
+                ordered_devices = ordered_devices.reshape(axis_shapes)
+                mesh = mesh_lib.Mesh(ordered_devices, axis_names)
```
hfan: The above 3 lines are repeated in the `if ordered_devices is not None` branch

yaochengji: Thanks for catching this!

hfan:
@kyuyeunk - this affects your refactor too?

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -138,6 +138,12 @@ def jit_step_func(self):
         @functools.partial(
             jax.jit,
             donate_argnums=(1, ),  # donate kv_cache
+            compiler_options={
+                "xla_tpu_all_gather_collective_matmul_mode":
+                "post_spmd_conservative",
+                "xla_tpu_reduce_scatter_collective_matmul_mode":
+                "post_spmd_conservative"
+            },
         )
```
hfan, line 147: OOC, if we don't fix the device order, do the collective matmul flags work? I don't recall if I've tried these flag, it could be because the device order issue that they didn't work.

tpu_commons/models/vllm/jax_row_parallel_linear.py
```diff
@@ -96,4 +99,10 @@ def forward(self, input: torch.Tensor):
             if not self.return_bias:
                 return output
             output_bias = self.bias if self.skip_bias_add else None
+            if self.enable_sequence_parallelism:
+                token_num = input.shape[0]
+                # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
+                if token_num // self.mesh.shape[
+                        'model'] >= TPU_SECOND_LAST_MINOR:
+                    output.shard_(NamedSharding(self.mesh, P('model', None)))
```
hfan, line 107: does it require the output to be reduced first? IOW, the `self.reduce_results` has to be true?

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
```diff
@@ -230,7 +232,14 @@ def forward_split(self, input):
...
+                if token_num // self.mesh.shape[
+                        'model'] >= TPU_SECOND_LAST_MINOR:
+                    input.shard_(NamedSharding(self.mesh, P('model', None)))
             if self.fuse_matmuls:
-                return self.forward_fused(input)
+                output, output_bias = self.forward_fused(input)
             else:
-                return self.forward_split(input)
+                output, output_bias = self.forward_split(input)
+            return output, output_bias
```
hfan, line 245: so the output is sharded by P('model', None)? Does it just work for the later layers if the output is changed to be sharded this way?

tpu_commons/models/vllm/vllm_model_wrapper.py
yaochengji: It works but the performance is much worse.

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
```diff
@@ -230,7 +232,14 @@ def forward_split(self, input):
 
     def forward(self, input: torch.Tensor):
         with jax.named_scope(self.name):
+            if self.enable_sequence_parallelism:
```
hfan, line 235: Unit test probably is needed

hfan:
Anything prevent it being turned on by default?

tpu_commons/models/vllm/jax_row_parallel_linear.py
yaochengji: From the definition of `reduce_results` is to add an `all_reduce` on the output of row_linear, but sequence_parallelism/collective_matmul requires it's a `reduce_scatter` instead.

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
yaochengji: I intended to add an e2e test, but cannot find a good example.

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
yaochengji: Emmm... what do you mean by `later layers`?

yaochengji:
> Anything prevent it being turned on by default?

@hfan I'm not sure if this can benefit all the models, maybe @QiliangCui can first try this in auto-tuning

tpu_commons/models/vllm/jax_row_parallel_linear.py
hfan: I see, makes sense

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
hfan: `tests/models/vllm/test_jax_XXX_linear.py` are the existing unit tests.

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
hfan: nvmd, resolved in the offline discussion, thanks!

kyuyeunk:
> @kyuyeunk - this affects your refactor too?

Yes, this does get affected. @yaochengji, I have created a draft PR so you can take an early look: https://github.com/vllm-project/tpu_commons/pull/512

hfan:
> > Anything prevent it being turned on by default?
> 
> @hfan I'm not sure if this can benefit all the models, maybe @QiliangCui can first try this in auto-tuning

As discussed offline, quantized matmuls that uses kernels whose collectives can't be automatically handled by XLA, so we probably have to wait for your collective-matmul kernel and better be safe (or only make it default for the non-quantized code path)

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
yaochengji: Thanks, @hfan ! The unit test is added, manually triggered in https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2043

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
```diff
@@ -230,7 +232,14 @@ def forward_split(self, input):
...
+                token_num = input.shape[0]
+                # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
```
vanbasten23, line 237: If "sharded token_num is larger than TPU_SECOND_LAST_MINOR", I guess the downside is waste of memory?

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
```diff
@@ -230,7 +232,14 @@ def forward_split(self, input):
...
+                if token_num // self.mesh.shape[
+                        'model'] >= TPU_SECOND_LAST_MINOR:
+                    input.shard_(NamedSharding(self.mesh, P('model', None)))
```
vanbasten23, line 240: very neat and convenient way of doing `self.apply_jax_(jax.lax.with_sharding_constraint, sharding)`

tpu_commons/utils.py
```diff
@@ -106,3 +112,62 @@ def get_padded_num_heads(num_heads: int, sharding_size: int) -> int:
...
+            # (0,1,0)
+            # (1,1,0)
+            # (0,2,0)
+            # (1,2,0)
+            # (0,3,0)
+            # (1,3,0)
+            if device_num == 8:
+                ordered_devices = np.array([
+                    devices[0],
+                    devices[2],
```
vanbasten23, line 145: I wonder why `(1,0,0)` maps to device `[2]`.

Also, does the order
```
            # (0,0,0)
            # (1,0,0)
            # (0,1,0)
            # (1,1,0)
            # (0,2,0)
            # (1,2,0)
            # (0,3,0)
            # (1,3,0)
```
matter?

yaochengji: Yeah, from the order, it matters

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
yaochengji: There will be more communication in the final result, you can have a try.

tpu_commons/utils.py
vanbasten23: why does `(1,0,0)` maps to device `[2]`?

tests/models/vllm/test_jax_merged_column_parallel_linear.py
```diff
@@ -78,7 +80,10 @@ def test_jax_merged_column_parallel_linear(bias, mesh, fuse_matmuls):
     # Set jax default device to workaround a layout bug in JAX 0.7.0 and earlier
     with torchax.default_env(), jax.default_device(jax.devices("tpu")[0]):
```
vanbasten23, line 81: is it using one TPU device? Do you need to test when there are multiple TPU device?

tpu_commons/models/vllm/jax_merged_column_parallel_linear_core.py
vanbasten23, line 240: It seems that SP is implemented by sharding the num_tokens dimension. Do you need to do an all-gather at the very end? I couldn't find it in your pr.

yaochengji: sharding propagation can handle this and make it correct.

vanbasten23: chatted offline, all-gather is not needed because some later ops (e.g. select which token to get logit (need to select num_reqs tokens)) may hint the compiler to get the global view. At that time, compiler will do a all-gather implicitly.

tpu_commons/models/vllm/jax_row_parallel_linear.py
kyuyeunk, line 107: Doesn't sequence parallelism usually use `'data'` to shard batch dim? Meaning, we can use both sequence and model parallelism and shard the inputs/outputs using `P('data', 'model')`

yaochengji: In this case, only the mesh "model" axis is enough. Here the sequence parallelism is applied on layer_norm, and model parallelism is applied on matmul. It is described in this paper: https://arxiv.org/abs/2205.05198

PR #521, opened 2025-08-20
fix scaling factor for DeepSeek

bzgoogle:
# Description

fix scaling factor for DeepSeek
We should use yarn based scaling factor for attention score

# Tests

tested with offline benchmarking with reasonable accuracy: https://paste.googleplex.com/4921594585284608

# Checklist

### DIFF
--- a/tpu_commons/models/jax/common/attention/deepseek_v3_attention.py
+++ b/tpu_commons/models/jax/common/attention/deepseek_v3_attention.py
@@ -75,6 +75,13 @@ class MLA(nnx.Module):
 
         assert self.N == self.K, "N and K must be equal for MLA"
 
+        if self.rope_scaling["factor"] <= 1.0:
+            yarn_mscale = 1.0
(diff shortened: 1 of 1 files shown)


tpu_commons/models/jax/common/attention/deepseek_v3_attention.py
```diff
@@ -289,11 +290,17 @@ def attention(
             P(),  # distribution: Replicated
         )
         out_specs = (self.attn_o_tnh.spec, P())
+        if self.rope_scaling["factor"] <= 1.0:
```
jrplatin: Are these constants? Does `rope_scaling["factor"]` ever change? If not, can we move this logic out to the `__init__` so we don't waste compute on this?

bzgoogle: Done. Moved to __post__init__

PR #540, opened 2025-08-22
Add phase-based profiler

jrplatin:
# Description

Currently, taking XProfs is a difficult process because of the stochastic nature of the vLLM scheduler, which can batch together (prefill/decode) requests arbitrarily.  For performance analysis, we'd like to obtain traces at three major batch types: prefill-heavy (the quotient of prefill / total scheduled tokens => 0.9), decode-heavy (the quotient of prefill / total scheduled tokens <= 0.2, mixed (the quotient is between 0.4 and 0.6)).  This PR enables the capture of these profiles automatically.

# Usage
In order to enable this feature, simply pass the `PHASED_PROFILING_DIR` environment variable.

The resulting profiling directory should look like this: 
image

# Tests

Ran CI here:  https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2181

# Checklist

### DIFF
--- a/tpu_commons/runner/jax/tpu_jax_runner.py
+++ b/tpu_commons/runner/jax/tpu_jax_runner.py
@@ -95,6 +95,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin):
         self.dtype = self.model_config.dtype
 
         self.phased_profiling_dir = os.getenv("PHASED_PROFILING_DIR", "")
+        self.phase_based_profiler = None
         if self.phased_profiling_dir:
             self.phase_based_profiler = runner_utils.PhasedBasedProfiler(
                 self.phased_profiling_dir)
--- a/README.md
+++ b/README.md
@@ -134,14 +134,19 @@ SKIP_JAX_PRECOMPILE=1
 
 There are two ways to profile your workload:
 
-#### Using `VLLM_TORCH_PROFILER_DIR`
+#### Using `PHASED_PROFILING_DIR`
 If you set the following environment variable:
 
 ```
-VLLM_TORCH_PROFILER_DIR=<DESIRED PROFILING OUTPUT DIR>
+PHASED_PROFILING_DIR=<DESIRED PROFILING OUTPUT DIR>
 ```
 
-vLLM will profile your entire workload, which can work well for toy workloads (like `examples/offline_inference.py`).
+we will automatically capture profiles during three phases of your workload (assuming they are encountered):
+1. Prefill-heavy (the quotient of prefill / total scheduled tokens for the given batch is => 0.9)
+2. Decode-heavy (the quotient of prefill / total scheduled tokens for the given batch is <= 0.2)
+3. Mixed (the quotient of prefill / total scheduled tokens for the given batch is between 0.4 and 0.6)
+
+To aid in your analysis, we will also log the batch composition for the profiled batches.
 
 #### Using `USE_JAX_PROFILER_SERVER`
 If you set the following environment variable:


tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -94,6 +95,21 @@ def __init__(
...
 
+        self.profiling_n_steps_left = 0
+        self.num_steps_to_profile_for = 15
+        self.profile_dir = None
+        self.inference_phase_seen = {
+            runner_utils.InferencePhase.PREFILL_HEAVY: False,
+            runner_utils.InferencePhase.DECODE_HEAVY: False,
+            runner_utils.InferencePhase.BALANCED: False
+        }
+        hacky_profile_dir = os.getenv("HACKY_PROFILE_DIR", "")
```
jrplatin: TODO: pick a better name here

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -468,12 +484,12 @@ def capture_model(self) -> None:
             logger.info("[TEMP] skip precompiling for multi-modal models")
             return
 
-        self._precompile_backbone()
-        self._precompile_select_hidden_states()
-        self._precompile_compute_logits()
-        self._precompile_sampling()
```
jrplatin: TODO: add back

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -94,6 +95,21 @@ def __init__(
...
+            runner_utils.InferencePhase.PREFILL_HEAVY: False,
+            runner_utils.InferencePhase.DECODE_HEAVY: False,
+            runner_utils.InferencePhase.BALANCED: False
+        }
+        hacky_profile_dir = os.getenv("HACKY_PROFILE_DIR", "")
+        if hacky_profile_dir:
+            self.profile_dir = hacky_profile_dir
+            logger.info("Hacky profiling enabled. Traces will be saved to: %s",
+                        self.profile_dir)
+        self.last_time = 0
```
hfan: nit: more descriptive name?

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1388,6 +1404,68 @@ def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
...
+                date_string_in_profiler_format = now.strftime(
+                    "%Y_%m_%d_%H_%M_%S/")
+                runner_utils.write_batch_composition_stats_to_file(
+                    os.path.join(
+                        self.profile_dir_with_phase_prefix,
+                        f"batch_composition_stats_{date_string_in_profiler_format}.json"
+                    ), batch_composition_stats)
+                if self.profiling_n_steps_left <= 0:
+                    jax.profiler.stop_trace()
+                    logger.info("Profiling for current phase finished")
```
hfan: possible to move it to a separate function or structure all the related members/functions into something like ProfileUtil class/module?

tpu_commons/runner/jax/tpu_jax_runner.py
jrplatin: Ah, I meant to remove this logging entirely (maybe I can just switch to `debug` though), thanks!

tpu_commons/runner/jax/tpu_jax_runner.py
jrplatin: Ah good suggestion, will do so, thanks!

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1388,6 +1404,68 @@ def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
         self.input_ids_cpu[
             total_num_scheduled_tokens:padded_total_num_scheduled_tokens] = 0
 
+        batch_composition_stats = runner_utils.get_batch_composition_stats(
```
xiangxu-google: Can we guard this whole new code section using an env? Usually profiling code needs to be zero overhead when profiling is not enabled. Otherwise there would be some CPU overhead here if we don't run profiling.

Currently we rely on the env `VLLM_TORCH_PROFILER_DIR` to determine if run profiling or not. You can reuse that or create a new one.

And +1 to @hfan 's suggestion that it's better to use a profiling manager to host the profiling related code.

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -94,6 +95,21 @@ def __init__(
         self.devices = devices
         self.dtype = self.model_config.dtype
 
+        self.profiling_n_steps_left = 0
```
xiangxu-google: Also these self.* code can be moved to the profiling manager class.

tpu_commons/runner/jax/tpu_jax_runner.py
py4: +1 to guarding with a flag

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -1388,6 +1405,72 @@ def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
...
+        # log batch composition stats every 1 minute
+        if time.time() - self.last_time > 60:
+            self.last_time = time.time()
+            logger.info(f"Batch composition stats: {batch_composition_stats}")
+
+        phase = runner_utils.determine_phase_from_batch_composition_stats(
+            batch_composition_stats)
+
+        # There is an initial profiling step that we need to skip, hence the latter condition
+        if self.profile_dir and batch_composition_stats[
```
py4: The nested `if/else` block for the profiling logic is hard to follow. It would be better to refactor it into smaller, more focused functions (e.g., `_start_profiling_if_needed`, `_handle_ongoing_profiling`).

tpu_commons/runner/utils.py
```diff
@@ -247,3 +257,86 @@ def _allocate() -> jax.Array:
...
+def determine_phase_from_batch_composition_stats(
+        batch_composition_stats: dict[str, Any]) -> InferencePhase:
+    """
+    TODO
+    """
+    num_prefill_tokens = batch_composition_stats["num_prefill_tokens"]
+    total_num_scheduled_tokens = batch_composition_stats[
+        "total_num_scheduled_tokens"]
+    prefill_ratio_for_batch = num_prefill_tokens / total_num_scheduled_tokens
+    if prefill_ratio_for_batch >= 0.9:
```
py4: Would be good to use a separate constant variable for these values, alongside a comment on why these are chosen this way

tpu_commons/runner/jax/tpu_jax_runner.py
hfan: Do it now?

tpu_commons/runner/utils.py
```diff
@@ -247,3 +269,206 @@ def _allocate() -> jax.Array:
...
+    num_prefill_tokens = batch_composition_stats["num_prefill_tokens"]
+    total_num_scheduled_tokens = batch_composition_stats[
+        "total_num_scheduled_tokens"]
+    prefill_ratio_for_batch = num_prefill_tokens / total_num_scheduled_tokens
+    if prefill_ratio_for_batch > PREFILL_HEAVY_RATIO_THRESHOLD:
+        return InferencePhase.PREFILL_HEAVY
+    elif prefill_ratio_for_batch <= DECODE_HEAVY_RATIO_THRESHOLD:
+        return InferencePhase.DECODE_HEAVY
+    else:
+        return InferencePhase.BALANCED
```
hfan, line 361: nit: this means BALANCED type is 0.2~0.9, it might be a little too wide

tpu_commons/runner/utils.py
```diff
@@ -247,3 +269,206 @@ def _allocate() -> jax.Array:
...
+        "num_decode_tokens": num_decode_tokens,
+        "padded_total_num_scheduled_tokens": padded_total_num_scheduled_tokens,
+        "num_reqs": num_reqs
+    }
+
+
+def determine_phase_from_batch_composition_stats(
+        batch_composition_stats: dict[str, Any]) -> InferencePhase:
+    """
+    TODO
```
hfan: do it now?

tpu_commons/runner/utils.py
```diff
@@ -247,3 +269,206 @@ def _allocate() -> jax.Array:
...
+        return InferencePhase.PREFILL_HEAVY
+    elif prefill_ratio_for_batch <= DECODE_HEAVY_RATIO_THRESHOLD:
+        return InferencePhase.DECODE_HEAVY
+    else:
+        return InferencePhase.BALANCED
+
+
+def write_batch_composition_stats_to_file(file_path, batch_composition_stats):
+    """
+    TODO
```
hfan: ditto

tpu_commons/runner/utils.py
```diff
@@ -247,3 +269,206 @@ def _allocate() -> jax.Array:
...
+        self.default_profiling_options = jax.profiler.ProfileOptions()
+        self.default_profiling_options.python_tracer_level = os.getenv(
+            "PYTHON_TRACER_LEVEL", 0)
+
+        self.current_phase = ""
+
+    def _write_batch_composition_stats_to_file_helper(
+            self, batch_composition_stats: dict):
+        """
+        TODO
```
hfan: ditto

tpu_commons/runner/utils.py
hfan: the function name and code is quite self-explaining though

tpu_commons/runner/utils.py
```diff
@@ -247,3 +269,206 @@ def _allocate() -> jax.Array:
...
+                batch_composition_stats)
+
+            jax.profiler.start_trace(
+                self.profile_dir_with_phase_suffix,
+                profiler_options=self.default_profiling_options)
+            break
+
+    def _step_or_stop_profiling(self, batch_composition_stats: dict):
+        """
+        TODO
```
hfan: ditto

tpu_commons/runner/utils.py
jrplatin: Updated to be 0.4 - 0.6 and categorize everything else as `AMBIGUOUS`

tpu_commons/runner/jax/tpu_jax_runner.py
```diff
@@ -94,6 +94,11 @@ def __init__(
         self.devices = devices
         self.dtype = self.model_config.dtype
 
+        self.phased_profiling_dir = os.getenv("PHASED_PROFILING_DIR", "")
```
py4, line 97: Maybe reuse `VLLM_TORCH_PROFILER_DIR`? Less ENV variable to deal with

jrplatin: The issue is that the worker is [already](https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/worker/tpu_worker_jax.py#L87) using that ENV var, so I think it makes sense to separate it out

py4: Yeah but it's a directory, we can write more files (the artifact of this PR to that directory) to it as well, no?

py4:
LGTM overall, but one question:

isn't it possible by profiling just two steps? I think this might be a simpler approach:

you can run offline example with 256 examples, set `--max_num_seqs=256` with high value for `--max_num_batched_tokens`. This way, the first step includes all prefills (prefill heavy) and the second step is all decode (decode heavy).

tpu_commons/runner/jax/tpu_jax_runner.py
xiangxu-google: What's the relationship with VLLM_TORCH_PROFILER_DIR now? 
I assume:
if we only want to run normal profiling, enabling `VLLM_TORCH_PROFILER_DIR` is enough.
if we want to run phased profiling, we have to enable both `VLLM_TORCH_PROFILER_DIR` and `PHASED_PROFILING_DIR`.
Am I right?

jrplatin: No, `VLLM_TORCH_PROFILER_DIR` will profile the entire workload if running `offline_inference` and vLLM serve + benchmarking with the `--profile` flag

PHASED_PROFILING_DIR is entirely separate and you only need to specify it to enable phased-based profiling

PR #542, opened 2025-08-22
Remove the logic of padding head dim to 128 before calling to kernel.

QiliangCui:
# Description

PRA3 kernel natively supports head dim being less than 128. So, we can remove the padding logic.

FIXES: b/440571948

# Tests
1. Ran on benchmark testing machines and got successful run.
2. Add a CIT to cover head-dim not equal to 128. (https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2176)

# Checklist


tests/kernels/ragged_paged_attention_kernel_v3_test.py
```diff
@@ -498,6 +499,24 @@ def test_ragged_paged_attention_soft_cap_cannot_be_zero(self):
                 soft_cap=0.0,
             )
 
+    @parameterized.product(dtype=[jnp.float32, jnp.bfloat16],
+                           head_dim=[64, 156, 256])
+    def test_ragged_paged_attention_padding(self, dtype, head_dim):
```
bythew3i: The `test_ragged_paged_attention_complex` should already cover unaligned head_dims. I would prefer not adding this test since the RPA tests are very slow due to the reference RPA. WDYT?

PR #556, opened 2025-08-25
[JAX][DeepSeek] Add support for FP8 weight loading for DeepSeek V3/R1

jrplatin:
# Description

In this PR, I add support for loading native FP8 DeepSeek checkpoints ([for example](https://huggingface.co/deepseek-ai/DeepSeek-V3/tree/main)) using Qwix.  This includes the following work:

* Ensure that the `model.layers.XX.mlp.gate.e_score_correction_bias` parameter is correctly updated to use FP32.
* Modify `tpu_commons/models/jax/deepseek_v3.py` to:
    * Cast model weights to their desired dtype (instead of BF16)
    * Load in scales (including adding the proper reshapes)
    * Load in the weights as QArrays if the model is quantized
    * Update the return value of `_load_individual_weight` to include the scale size (if using the quantized model)
    * I preserve BF16 backwards compatibility, which can be specified via ` --additional_config='{"skip_quantization": true}'`
* Add new logic to automatically detect if a given checkpoint (only DeepSeekV3/R1 for now) is quantized and use a default Qwix config if so (so that the user doesn't need to pass it in via the `additional-config`).  If a user does pass a config in, we'll use that instead.

Example command to run DeepSeek-V3 with quantization:

```
VLLM_MLA_DISABLE=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --task=generate --model=deepseek-ai/DeepSeek-V3  --max-model-len=1024 --tensor-parallel-size 8 --max-num-batched-tokens 1024 --max-num-seqs=1 --hf-config=deepseek-ai/DeepSeek-V3 --hf_overrides '{"architectures": ["DeepSeekV3"]}'
```

# Tests

Added / updated quantization-related testing

# Checklist


tpu_commons/kernels/ragged_paged_attention/v3/util.py
```diff
@@ -35,6 +35,7 @@ def next_power_of_2(x: int):
 
 def get_tpu_version() -> int:
     """Returns the numeric version of the TPU, or -1 if not on TPU."""
+    return 7
```
jrplatin: TODO: remove

examples/offline_inference.py
```diff
@@ -108,6 +108,7 @@ def main(args: dict):
 
     parser = create_parser()
     args: dict = vars(parser.parse_args())
+    args.pop("enable_prompt_adapter")
```
jrplatin: TODO: remove

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -443,21 +456,60 @@ def _process_moe_weights(self, loaded_name, loaded_weight, weights_dict):
...
 
         # Convert weights from torch into numpy
-        # TODO: set cast_type based on model weight's type.
-        cast_type = ml_dtypes.bfloat16
+        cast_type = model_weight.value.dtype
         weight_np = weight.to(torch.float32).numpy().astype(cast_type)
 
+        if scale is not None:
+            scale = scale.to(torch.float32).numpy().astype(self.scale_dtype)
+            # TODO: remove
```
jrplatin: TODO: remove

jrplatin:
Synced with @gpolovets@google.com today to go over the logic and will update the following in the PR:

* We don't need a new unquant/kernel dtype
* I should go back and double verify logits against GPU for the quantized matmuls
* We should `scale_dtype` as a Qwix config param (right now, it's hard-coded into the DS model)
* Add a quick comment the scale reshape maping to note that it's used / needed for quantized models
* Add an assertion [here](https://github.com/vllm-project/tpu_commons/pull/556/files#diff-ed9b2fb41f9f568f5883cff40c004ac2dc3855c56186eb43a8f27ac769d098fcL512) to ensure that the Qwix dtype and incoming model dtype are the same
* Clean-up the random weights PR

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -442,21 +472,56 @@ def _process_moe_weights(self, loaded_name, loaded_weight, weights_dict):
...
         weight_np = reshape_params(name, weight_np, self._weight_shape_map)
+        if scale is not None:
+            scale = reshape_params(name, scale, self._scale_shape_map)
         weight_np = self._transpose_params(name, weight_np)
+        if scale is not None:
+            scale = self._transpose_params(name, scale)
+            # This handles a weird case where the scale is not cleanly divisible by 128
+            # (the base weight shape is 576 and the block size is 128), so we'll repeat
+            # the scale to bypass this
+            if "attn.kernel_kv_down_proj_DA" in mapped_name:
```
jrplatin: Too manual?

tpu_commons/models/jax/utils/quantization/configs/fp8_deepseek.yaml
kyuyeunk, line 1: Having to specify `fp8_deekseek.yaml` just to load the model is a really bad user experience. Please modify the code to automatically parse the [config.json](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json), identify it's [quant_method](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L40), and apply quantization config accordingly (such as [quantization format](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L39), [dynamic or static quantization](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L38), [subchannel block size](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L41-L44), etc)

jrplatin: @kyuyeunk  Very good point, and I think it depends on whether we want the dequantized BF16 model to be the default or the FP8 model to be the default -- I will ask George for his thoughts, thanks

jrplatin: Synced with George, and we agree that quantization will be the default -- thanks for the insights Kyuyeun!

tpu_commons/models/jax/deepseek_v3.py
bzgoogle: coud we extract the shape(576, ) from  kernel_kv_down_proj_DA, in stead of hard code the number 576 here?

tpu_commons/models/jax/common/moe/deepseek_moe.py
```diff
@@ -106,6 +106,6 @@ def __post_init__(self, rngs: nnx.Rngs):
                                       random_init=self.random_init)
         self.bias_E = create_param(rngs,
                                    shape=(E, ),
-                                   dtype=self.dtype,
+                                   dtype=jnp.float32,
```
bzgoogle: curious why we have to set the dtype to float32 rather than dtype of the model? for accuracy purpose?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -339,7 +341,14 @@ def __init__(self, vllm_config: VllmConfig, num_layers, hidden_size,
             (attn_heads, qk_nope_head_dim + v_head_dim, kv_lora_rank),
             "o_proj": (hidden_size, attn_heads, v_head_dim)
         }
-
+        # NOTE: this is only needed for pre-quantized models
+        self._scale_shape_map = {
+            "q_b_proj":
+            (1, qk_nope_head_dim + qk_rope_head_dim, q_lora_rank // 128),
```
bzgoogle: is the tile size of 128 subject to tunable? If so can we get it from a config?

tpu_commons/models/jax/common/moe/deepseek_moe.py
jrplatin: In the quantized/original checkpoint, it's actually [FP32](https://screenshot.googleplex.com/8szXaoAUJZ8yKyk), so I figured we should do the same here -- not sure if there's a better way?

tpu_commons/models/jax/deepseek_v3.py
jrplatin: Good point, will do!

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -306,8 +308,8 @@ def __init__(self, vllm_config: VllmConfig, num_layers, hidden_size,
             model_name_or_path=vllm_config.model_config.model,
             framework="pt",
             download_dir=vllm_config.load_config.download_dir)
-        self.is_verbose = getattr(vllm_config.additional_config, "is_verbose",
-                                  False)
+        self.is_verbose = vllm_config.additional_config.get(
```
kyuyeunk, line 311: What is the reason for this change?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -397,6 +399,48 @@ def __init__(self, vllm_config: VllmConfig, num_layers, hidden_size,
...
+                    assert quant_dtype_str, "Quantization dtype not found in Qwix config! We currently expect your Qwix config to have a rule with module_path '.*' and a weight_qtype."
+                    self.quant_dtype = getattr(jnp, quant_dtype_str)
+                    logger.info(
+                        f"Quantizing DeepSeek with quantization dtype: {self.quant_dtype} and scale dtype: {self.scale_dtype}"
+                    )
+
+            quantization_block_sizes = vllm_config.model_config.hf_config.quantization_config[
+                "weight_block_size"]
+            # TODO (jacobplatin): we should handle the case where these aren't all the same
+            self.quantization_block_size = quantization_block_sizes[0]
```
kyuyeunk: please handle this properly. you can reference this code in vLLM: https://github.com/kyuyeunk/vllm/blob/4fdd6f5cbf877de7c4de33086fe41bb0ac1d3cf3/vllm/model_executor/layers/quantization/fp8.py#L278-L281

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -509,36 +644,79 @@ def load_weights(self, model_for_loading: nnx.Module):
...
+                            scale = loaded_weight
+                            loaded_weight = quantized_weights[weight_name]
+                            loaded_name = weight_name
+                            del quantized_weights[weight_name]
+                        else:
+                            quantized_scales[loaded_name] = loaded_weight
+                            continue
+                    # In the case that we don't want to use the quantized weights,
+                    # we'll dequantize the weights using the loaded scale on-the-fly
+                    else:
```
kyuyeunk, line 685: What is the use case for dequantizing weights? Meaning, isn't it always better to serve the model quantized? Or is it due to performance bug with xla/qwix w.r.t. subchannel qunantization when serving quantized model?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -397,6 +399,48 @@ def __init__(self, vllm_config: VllmConfig, num_layers, hidden_size,
...
+            "skip_quantization", False)
+        if self.is_model_quantized:
+            # NOTE: this will only be used for loading in quantized weights (via Qwix)
+            qwix_config = vllm_config.additional_config.get(
+                "quantization", {}).get("qwix", {})
+            self.scale_dtype = getattr(
+                jnp, qwix_config.get("scale_dtype", "bfloat16"))
+            # NOTE (jacobplatin): this is certainly a bit hacky, but I don't think there's any
+            # super clean way to get the desired quantization dtype
+            for rule in qwix_config.get("rules", []):
```
kyuyeunk, line 419: This logic isn't really model-specific and doesn't belong in the model implementation. Please move this code into utility function that can figure out a dtype from a single function call.

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -509,36 +644,79 @@ def load_weights(self, model_for_loading: nnx.Module):
...
                         continue
-                if loaded_weight.dtype == torch.float8_e4m3fn:
-                    fp8_weights[loaded_name] = loaded_weight
-                    continue
+                # NOTE: loaded_weight will be a Torch tensor, so we need to convert it to the
+                # equivalent jnp dtype
+                scale = None
+                if loaded_weight.dtype == j2t_dtype(self.quant_dtype.dtype):
+                    if self.is_model_quantized:
+                        scale_name = loaded_name.replace(
```
kyuyeunk, line 659: this is really hacky. please do it properly and make the changes to `model_weights_generator` & make it return different paths depending `quant_method` - such as [fp8](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L40), [compressed-tensors](https://huggingface.co/EmbeddedLLM/deepseek-r1-FP8-Dynamic/blob/main/config.json#L79), etc. Please be aware that each quant method will store checkpoints in a different paths.

https://github.com/vllm-project/tpu_commons/blob/34af638133f75cbf30a2979d859e404759eff9e7/tpu_commons/models/jax/utils/weight_utils.py#L73-L82

tpu_commons/models/jax/utils/quantization/quantization_utils.py
```diff
@@ -298,3 +318,76 @@ def apply_qwix_on_abstract_model(vllm_config: "VllmConfig") -> bool:
...
+        quant_method: the quantization method
+        skip_quantization: whether to skip quantization.  In this case, we'll return None
+
+    Returns:
+        a dictionary containing the default Qwix quantization rules
+    """
+    if skip_quantization:
+        return None
+    if model_type == "deepseek_v3" and quant_method == "fp8":
+        return DEFAULT_DEEPSEEK_FP8_CONFIG
```
kyuyeunk, line 345: Why is deepseek handled specially and has it's own separate default configuration?

Also, it's not a good idea to make assumptions like this & better to just refer to ground truth - which is the [config.json](https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/config.json#L42) file. What happens if a config has a different block size, uses static instead of dynamic, etc?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -442,21 +486,67 @@ def _process_moe_weights(self, loaded_name, loaded_weight, weights_dict):
...
+            scale = reshape_params(name, scale, self._scale_shape_map)
         weight_np = self._transpose_params(name, weight_np)
+        if scale is not None:
+            scale = self._transpose_params(name, scale)
+            weight_shape = weight_np.shape
+            scale_shape = scale.shape
+            assert len(weight_shape) == len(scale_shape)
+            for idx, (weight_dim,
+                      scale_dim) in enumerate(zip(weight_shape, scale_shape)):
+                if weight_dim // self.quantization_block_size != scale_dim and weight_dim // scale_dim != 1:
```
kyuyeunk: isn't it better to just throw an error?

tpu_commons/models/jax/deepseek_v3.py
jrplatin: `additional_config` is a `dict` and not an object, so this check will always fail

tpu_commons/models/jax/deepseek_v3.py
jrplatin: Ideally yes -- we always want to serve the quantized model, but some of our initial testing/numbers were with BF16 and we'd like to have an escape hatch to use that

cc @gpolovets1

tpu_commons/models/jax/deepseek_v3.py
jrplatin: Sorry, can you expand on what you mean by "do it properly" -- I don't follow

tpu_commons/models/jax/deepseek_v3.py
jrplatin: I don't believe so -- this handles an unusual case where the scale shape (5) multiplied by the block size (128) does not equal the weight shape (576), but we can still serve the model by simply repeating the scale

tpu_commons/models/jax/utils/quantization/quantization_utils.py
jrplatin: DeepSeek is the only case we handle for now because it's the only pre-quantized model that we support in the JAX path -- as I add support for AWQ/BnB checkpoints, this list will be expanded

As for the second point, I've only tested / plan to support (in the scope of this PR) the original DeepSeek models (e.g. [V3](https://huggingface.co/deepseek-ai/DeepSeek-V3/tree/main) and [R1](https://huggingface.co/deepseek-ai/DeepSeek-R1-0528/tree/main)), which use the same quantization scheme.  That said, I think it might be best to add a few checks to ensure the quantization config is setup as desired (e.g. `block_size` =  128, `quant_method` == `fp8`, etc.)

tpu_commons/models/jax/deepseek_v3.py
kyuyeunk: Can you elaborate the example of that unusual case & when that might be a valid case & not an error?

tpu_commons/models/jax/utils/quantization/quantization_utils.py
kyuyeunk: > as I add support for AWQ/BnB checkpoints, this list will be expanded

What do you mean by this? There is practically unlimited list of quantized models. Do you mean you will expand this list for every single model?

tpu_commons/models/jax/deepseek_v3.py
kyuyeunk: the last suffix of the checkpoint isn't model specific. It's rather quantization method specific. So two different models that was quantized with same method will have the same suffix & two same models that was quantized with different method will have the different suffix. 

Here are some examples:
- quantization method fp8: 
  - suffix
    - weight: `weight`
    - scale: `weight_scale_inv` (regular scale doesn't exist and stores inverse instead)
  - example
    - deekseek: https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/model-00001-of-000163.safetensors
    - qwen: https://huggingface.co/Qwen/Qwen3-8B-FP8/blob/main/model-00001-of-00002.safetensors
- compressed-tensors:
  - suffix
    - weight: `weight`
    - scale: `weight_scale`
  - example
    - deepseek: https://huggingface.co/EmbeddedLLM/deepseek-r1-FP8-Dynamic/blob/main/model-00001-of-00135.safetensors
    - qwen: https://huggingface.co/RedHatAI/Qwen3-4B-FP8-dynamic/blob/main/model-00001-of-00002.safetensors

So there is two problems with this implementation

1. you shouldn't just blindly do "check if it weight path ends with `.weight_scale_inv`, if it does do x". There is a proper specification how a weight is stored for a given quantization method (`fp8`, `compressed-tensors`, etc). So you should do this instead
   1. check the quantization method in the `config.json`
   1. create mapping based on that quantization method
   1. throw error if a weight does not exist despite what we have expected from quantization method
1. As I mentioned above, this isn't model specific. two same models with a different quantization method will have a different weight path. Because of this, this code does not belong in model implementation.

tpu_commons/models/jax/deepseek_v3.py
kyuyeunk: Got it. So this code path will be deprecated once evaluation with the unquantized fp8 weight is complete?

tpu_commons/models/jax/deepseek_v3.py
jrplatin: [Here's](https://screenshot.googleplex.com/3Q4FZ8KRj6muGmT) an example from the model -- this logic gets around the fact that 576 // 128 != 5

tpu_commons/models/jax/utils/quantization/quantization_utils.py
jrplatin: My current understanding is that, regardless of the AWQ/BnB model you want to load in, the Qwix config will be the same -- the only major difference between the two is that AWQ adds support for Zero Point, which BnB checkpoints don't have

Ideally, the list shouldn't be infinite and definitely understand your concern

tpu_commons/models/jax/deepseek_v3.py
jrplatin: Ah got it! Thank you for the detailed explanation and will update

jrplatin:
After offline discussions, I've decided to address Kyuyeun's excellent feedback in a fast follow-up PR (tracking bug [here](https://b.corp.google.com/issues/445467331))

PR #607, opened 2025-08-28
[Model] Shard embedding & lm_head

qihqi:
# Description

Upstream models usually uses VocabParallelEmbedding and ParallelLMHead to parallelize the weights of embedding and final out proj to many devices. In torchax path, we are not sharding those. 

This PR will shard those according same as GPU.

# Tests

CI

# Checklist


hfan:
I assume this is the reason why more HBM was used for torchax models?

hfan:
Do you think the title of the PR could start with "[Model]" instead of misc?

PR #616, opened 2025-08-30
enable hybrid kv cache allocator

Chenyaaang:
# Description

Fix #510 

head_size needs to be padded to 128 when creating kv_cache_spec to match the kv cache shape during `initialize_kv_cache`. There used to be error on Qwen2.5-0.5B model since its head_size is 64. 

# Tests

- E2E test on single attention type models, multi attention type models passed
- CI passed: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2585

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/sharding.py
+++ b/tpu_commons/models/vllm/sharding.py
@@ -3,15 +3,16 @@ import functools
 import humanize
 import jax
 import torch
-from torch.nn import Parameter
 import torchax
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
+from torch.nn import Parameter
 from torch.utils import _pytree as pytree
 from torchax.interop import torch_view
 from torchax.ops.mappings import t2j
 from vllm.config import VllmConfig
 from vllm.model_executor.layers.fused_moe import FusedMoE
-from vllm.model_executor.layers.vocab_parallel_embedding import VocabParallelEmbedding, ParallelLMHead
+from vllm.model_executor.layers.vocab_parallel_embedding import (
+    ParallelLMHead, VocabParallelEmbedding)
 
 from tpu_commons.logger import init_logger
 from tpu_commons.models.vllm.jax_fused_moe import JaxFusedMoE
@@ -28,11 +29,12 @@ def shard_fused_moe(layer: torch.nn.Module, mesh: Mesh,
     return jax_layer
 
 
-def shard_vocab_parallel_embedding(layer: torch.nn.Module, mesh: Mesh, vllm_config: VllmConfig):
+def shard_vocab_parallel_embedding(layer: torch.nn.Module, mesh: Mesh,
+                                   vllm_config: VllmConfig):
     assert isinstance(layer, VocabParallelEmbedding)
     w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)
     layer.weight = w.apply_jax_(jax.device_put,
-        NamedSharding(mesh, P('model', None)))
+                                NamedSharding(mesh, P('model', None)))
     return layer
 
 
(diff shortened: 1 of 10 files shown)


tpu_commons/models/vllm/sharding.py
kyuyeunk, line 1: Other than very JAX specific logic (like sharding) we are trying to minimize any other logic in this file (like changes in this PR which creates `AttentionInfo`) as it utilizes custom tpu_commons logic. Instead, we are trying to use pre-existing vLLM API as much as possible.

Is the feature trying to implement in this PR difficult to express in existing vLLM's API? For instance, is there any API in vLLM's Attention layer that allows calling custom code? What about in this line where it calls a code in tpu_commons's `pallas_torchax.py`?

- https://github.com/vllm-project/vllm/blob/4071c76cf3cff46e14630f1e66cbf006b6eb51d3/vllm/attention/layer.py#L181-L185
- https://github.com/vllm-project/tpu_commons/blob/8523c8ede281e9fa54e032288bacb1b1a4978ce2/tpu_commons/attention/backends/pallas_torchax.py#L35-L50

If some necessary argument is missing, can we utilize `extra_impl_args`?: https://github.com/vllm-project/vllm/blob/4071c76cf3cff46e14630f1e66cbf006b6eb51d3/vllm/attention/layer.py#L86

tpu_commons/models/vllm/sharding.py
```diff
@@ -110,6 +145,9 @@ def _move_to_tpu_replicated(x):
     with jax.default_device(jax.devices("cpu")[0]), torchax.default_env():
         shard_parallel_layers_to_tpu(model, mesh, vllm_config)
 
+        # extract attention info.
+        get_attention_info(model, vllm_config)
```
kyuyeunk: This logic is in `sharding.py/shard_model_to_tpu` but it's not related to sharding - which can make developers confused. Please consider moving this logic elsewhere or rename `shard_model_to_tpu` to be generic.

tpu_commons/models/vllm/sharding.py
Chenyaaang: thank you for your comment. The problem here is: when we convert a vllm pt model into jax, all we do is inside `shard_model_to_tpu`, which walks through the pt model, graps all the torch.Tensor weights, and converts them into jax array. One thing missing from the existing torch model is, it also binds its attention class into vllm_config.compilation_config.static_forward_context, and this will further be used to initialize kv cache. What I added here is to add the attention info back to the compilation config. I've moved this into vllm_model_wrapper, before `shard_model_to_tpu`.

kyuyeunk: hmm sorry but I still could not understand.

So why isn't it possible to just copy paste below code into the `__init__` function of `PallasAttentionBackendImpl`?

```
self.vllm_config.compilation_config.static_forward_context[
                    module.layer_name] = AttentionInfo(
                        layer_name=module.layer_name,
                        num_kv_heads=module.num_kv_heads,
                        head_size=module.head_size,
                        attn_type=module.attn_type,
                        sliding_window=module.sliding_window,
                        kv_sharing_target_layer_name=module.
                        kv_sharing_target_layer_name)
```

One blocker I see is that `vllm_config` is not accessible in there. But wouldn't it be possible use `get_current_vllm_config()` like this? https://github.com/vllm-project/vllm/blob/862f2ef893d9751db0a92bd2d4ae0e3d9677872f/vllm/attention/layer.py#L196-L199

Chenyaaang: This is a very good idea, PallasAttentionBackendImpl's `__init__` almost receives everything AttentionInfo needs except for `layer_name`, Attention class doesn't pass in `prefix` into impl_class. 
https://github.com/vllm-project/vllm/blob/862f2ef893d9751db0a92bd2d4ae0e3d9677872f/vllm/attention/layer.py#L182-L185

If we want to limit the changes inside tpu_commons, I believe we still have to adopt the current way.

Chenyaaang: As synced offline, we can simply update the compilation_config from `vllm_config_for_load`.

PR #619, opened 2025-09-02
[Torchax][Kernel] Support for KV cache quantization

kyuyeunk:
# Description

- Following PR enables KV cache quantization in Torchax 
- Using model [RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic](https://huggingface.co/RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic) in Ironwood, I have verified that it's running correctly.
	- Numeric validation
		- Command: `TPU_BACKEND_TYPE=jax MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8-dynamic --tensor_parallel_size=2 --task=generate --max_model_len=1024 --download_dir=/mnt/disks/persist --kv_cache_dtype=fp8`
		- Result: http://gpaste/6030890064674816
	- Performance validation
		- bf16 kv cache: 1107.23 toks/s
		- fp8 kv cache: 1809.04 toks/s (+8%)

# Tests

```
pytest -v tests/models/vllm/layers/test_pallas_torchax.py
```

# Checklist

### DIFF
--- a/tpu_commons/attention/backends/pallas_torchax.py
+++ b/tpu_commons/attention/backends/pallas_torchax.py
@@ -4,6 +4,7 @@ import functools
 from typing import Optional, Tuple
 
 import jax
+import jax.numpy as jnp
 import torch
 from jax.sharding import Mesh
 from torchax.interop import jax_view, torch_view
@@ -18,6 +19,14 @@ from tpu_commons.models.vllm.vllm_model_wrapper_context import \
     get_vllm_model_wrapper_context
 
 logger = init_logger(__name__)
+
+TPU_STR_DTYPE_TO_JAX_DTYPE = {
+    "bfloat16": jnp.bfloat16,
+    "fp8": jnp.float8_e4m3fn,
+    "fp8_e4m3": jnp.float8_e4m3,
+    "fp8_e5m2": jnp.float8_e5m2,
+    "int8": jnp.int8,
+}
 
 
 class PallasAttentionBackend(AttentionBackend):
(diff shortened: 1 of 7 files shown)


tpu_commons/attention/backends/pallas_torchax.py
```diff
@@ -100,20 +125,30 @@ def forward(
...
         return torch_view(outputs)
 
 
 @functools.partial(
     jax.jit,
-    static_argnums=(5, 6, 7, 8,
-                    9),  # mesh, scale, head_size, num_heads, num_kv_heads
+    static_argnums=(
+        5, 6, 7, 8, 9, 10, 11, 12
+    ),  # mesh, scale, head_size, num_heads, num_kv_heads, q_scale, k_scale, v_scale
```
hfan, line 152: how does layer._v_scale_float and layer._k_scale_float calculated? I'm wondering if we should use q_scale/k_scale/v_scale as part of static_argnums, if the scale values changes a lot, they can't be effectively used as compilation cache lookup keys.

kyuyeunk: It's either loaded from checkpoint or calculated once during inference

more information here:
* https://github.com/vllm-project/vllm/blob/862f2ef893d9751db0a92bd2d4ae0e3d9677872f/vllm/attention/layer.py#L118-L126
* https://github.com/vllm-project/vllm/blob/862f2ef893d9751db0a92bd2d4ae0e3d9677872f/vllm/attention/layer.py#L296

PR #625, opened 2025-09-02
[Phase-Based Profiler] Allow workloads with `max-num-seq=1` to be profiled + add environment variable option to configure number of steps to profile for

jrplatin:
# Description

This PR fixes a bug where runs with `max-num-seqs=1` could not be profiled with the phase-based profiler.  I also add an env var option to change the number of steps to profile for.

# Tests

Tested locally.

# Checklist


py4:
suggestion: maybe a less specific tag? like [Profiler]. [Phase-Based Profiler] is too specific and is beyond a tag/category

PR #630, opened 2025-09-03
[Bugfix] Fix _PP assertion error in torchax

kyuyeunk:
# Description

Continuation of the fix from https://github.com/vllm-project/tpu_commons/pull/627

# Tests

CI: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2598

# Checklist

### DIFF
--- a/tpu_commons/worker/tpu_worker_jax.py
+++ b/tpu_commons/worker/tpu_worker_jax.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import tempfile
 from typing import Callable, Dict, Optional, Tuple, Union
 
 import jax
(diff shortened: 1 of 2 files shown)


tpu_commons/worker/tpu_worker_jax.py
```diff
@@ -121,8 +125,20 @@ def init_device(self):
...
+            world_size=1,
+            rank=0,
+            local_rank=0,
+            distributed_init_method=f"file://{temp_file}",
+            backend="gloo",
+        )
+        ensure_model_parallel_initialized(
+            tensor_model_parallel_size=1,
+            pipeline_model_parallel_size=1,
+        )
```
xiangxu-google, line 141: Thanks for the PR!
I didn't realize that the vLLM's change also breaks our CI, I sent https://github.com/vllm-project/tpu_commons/pull/627 just for fixing the multi-host run.
I kind of feel calling `init_distributed_environment` is not the ideal way, it can fix the issue but it'll introduce side effect because JAX doesn't need those pytorch process groups. And the `init_distributed_environment` here is hard coded for single host. That's why I mocked PP in my last PR to avoid calling any pytorch dist methods.
Can we keep this worker file unchanged, I think removing `ensure_model_parallel_initialized` in `vllm_modek_wrapper` should fix our CI.

xiangxu-google: I just tested locally by simply removing those code in `vllm_modek_wrapper`, it doesn't work. Seems torchax requires those TP and PP groups.
Then let's merge this PR first to pass CI first, I'll think if there's a cleaner solution for both torchax and jax.

PR #631, opened 2025-09-03
Add single lora (single-chip) support to torchax+jax_runner path

vanbasten23:
# Description

This PR adds single lora (on single-chip) support to torchax+jax_runner path. Currently limitation is the precompiling is very slow so it will be address in a follow-up PR.

# Tests

`VLLM_XLA_CHECK_RECOMPILATION=1  SKIP_JAX_PRECOMPILE=1  VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  pytest -s -vv tests/lora/test_lora.py -k test_single_lora`

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/vllm_model_wrapper.py
+++ b/tpu_commons/models/vllm/vllm_model_wrapper.py
@@ -244,6 +244,7 @@ def load_lora_model(model: torch.nn.Module, model_config: ModelConfig,
     return lora_manager, lora_manager.create_lora_manager(model)
 
 
+# The reason why replace the method is that the set_lora and reset_lora need to run under torchax env.
 def replace_set_lora(model):
 
     def _tpu_set_lora(
--- a/tpu_commons/runner/lora_utils.py
+++ b/tpu_commons/runner/lora_utils.py
@@ -32,11 +32,10 @@ class LoraUtils:
         lora_requests: set[LoRARequest]
         prompt_lora_mapping, token_lora_mapping, lora_requests = \
                             self.runner.input_batch.make_lora_inputs(padded_num_scheduled_tokens_per_req)
+        # One should not put lora_manager.set_active_loras under torchax.default_env() because set_active_loras also load lora from disk and torchax currently does not support that.
+        # Here we load the lora and set the lora weight to the linear layers.
         self.runner._set_active_loras(prompt_lora_mapping, token_lora_mapping,
                                       lora_requests)
-        # One should not put lora_manager.set_active_loras under torchax.default_env() because set_active_loras also load lora from disk and torchax currently does not support that.
-        self.runner.set_active_loras(self.runner.input_batch,
-                                     padded_num_scheduled_tokens_per_req)
 
         shard_lora_weights_and_move_to_tpu(self.runner.model.model,
                                            self.runner.mesh)


tpu_commons/utils.py
kyuyeunk, line 1: can you explain why `shard_and_move_lora_to_tpu` is needed? it seems like there isn't any specialized sharding logic going on and it just uses `NamedSharding(mesh P()`) i.e., replicated across all layers? What is this different from pre-existing logic in here? https://github.com/vllm-project/tpu_commons/blob/9ed9953ef5b6561479730f688fa6ebbf65618a07/tpu_commons/models/vllm/sharding.py#L105-L110

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -109,6 +113,21 @@ def load_weights(self):
         # Load the vLLM model and wrap it into a new model whose forward function can calculate the hidden_state and logits.
         with load_context:
             vllm_model = vllm_get_model(vllm_config=vllm_config_for_load)
+        lora_manager = None
```
kyuyeunk, line 116: How much logic can we re-use from vLLM? https://github.com/vllm-project/vllm/blob/cb55ad86fe47759c22de06b3b3b93b3aa28690a2/vllm/worker/model_runner.py#L1068-L1079

tpu_commons/models/vllm/quantization/unquantized.py
```diff
@@ -65,7 +65,7 @@ class JaxUnquantizedLinearMethod(UnquantizedLinearMethod):
     def __init__(self, jax_config: JaxCommonLinearConfig):
         self.jax_config = jax_config
 
-    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+    def move_weights_to_torchax_tensor(self, layer: torch.nn.Module) -> None:
```
kyuyeunk: This function gets called by vLLM during model loading process. Meaning, changing the function name will cause pre-existing models/logics to break: https://github.com/vllm-project/vllm/blob/12e1e63cc5565c5926a2d3454d3b31a1ee03f564/vllm/model_executor/model_loader/base_loader.py#L50

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -109,6 +113,21 @@ def load_weights(self):
...
+            # replace the layer with LoRA layers.
+            lora_manager, vllm_model = load_lora_model(
+                vllm_model,
+                vllm_config_for_load.model_config,
+                vllm_config_for_load.scheduler_config,
+                vllm_config_for_load.lora_config,
+                device="cpu")
+
+        # why moved below function here whereas it used to be in vllm_get_model?
+        # vllm_get_model moves the vllm_model to jax device. But we expect to initialize the weight on cpu
```
kyuyeunk: Can you explain this part? why does weights being loaded into TPU by `process_weights_after_loading` breaks lora and we have to keep it in cpu until `load_lora_model` is called?

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -196,3 +216,38 @@ def compute_logits_func(
             return jax_view(logits)
 
         return compute_logits_func
+
+
+def move_weights_to_torchax_tensor(model: torch.nn.Module):
```
kyuyeunk: Related to the above comment. What is the rationale for calling `process_weights_after_loading` (not renamed to `move_weights_to_torchax_tensor`) in here?

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -109,6 +113,21 @@ def load_weights(self):
         # Load the vLLM model and wrap it into a new model whose forward function can calculate the hidden_state and logits.
         with load_context:
             vllm_model = vllm_get_model(vllm_config=vllm_config_for_load)
+        lora_manager = None
+        if vllm_config_for_load.lora_config is not None:
+            # replace the layer with LoRA layers.
+            lora_manager, vllm_model = load_lora_model(
```
lsy323: nit: seems this `lora_manager` is not used in this file. Probably we don't need to return it?

tpu_commons/runner/compilation_manager.py
```diff
@@ -67,15 +67,16 @@ def capture_model(self) -> None:
...
-        self._precompile_backbone()
-        self._precompile_select_from_array()
-        self._precompile_compute_logits()
-        self._precompile_disagg_utils()
-        self._precompile_sampling()
-        self._precompile_gather_logprobs()
-        self._precompile_structured_decoding()
-        if self.runner.speculative_config:
-            self._precompile_rejection_sampler()
+        with self.runner.maybe_setup_dummy_loras(self.runner.lora_config):
```
lsy323: Let's include the change in the pre compilation in a follow up PR?

lsy323:
cc @xiangxu-google @py4 on the change around JAX runner, ptal, thanks!

tpu_commons/utils.py
vanbasten23: > can you explain why shard_and_move_lora_to_tpu is needed?

The reason why `shard_and_move_lora_to_tpu ` is needed is because we need to shard the lora weight and move the weights to TPU. The way it works is: 
- during inference time, we get the lora_request, extract the lora_path from the request, load the lora weight from HF or local disk
- then we shard the lora weights and move them to TPU

> it seems like there isn't any specialized sharding logic going on and it just uses NamedSharding(mesh P()) i.e., replicated across all layers?

Yeah, this PR intends to only support single chip. When we add the multi-chip support in a follow-up pr, we should shard the weight using TP. At that time, we will change the sharding accordingly.

> What is this different from pre-existing logic in here?

Which pre-existing logic are you referring to? Like how vLLM upstream implements it for TP>1?

tpu_commons/models/vllm/vllm_model_wrapper.py
vanbasten23: Actually, the relevant logic in vLLM is here https://github.com/vllm-project/vllm/blob/cb55ad86fe47759c22de06b3b3b93b3aa28690a2/vllm/v1/worker/tpu_model_runner.py#L1170-L1186. I'm reusing most of it.

tpu_commons/models/vllm/quantization/unquantized.py
vanbasten23: This is related to your another comment https://github.com/vllm-project/tpu_commons/pull/631#discussion_r2321016496. 

I understand `process_weights_after_loading` is called during model loading process and it shards/move the weights to TPU. The reason why I cannot call `process_weights_after_loading ` during model loading process is that after `process_weights_after_loading` the base model's device becomes 'jax', then vLLM will move the lora weight to the same device as the base_model which is `jax`, and pytorch doesn't recognize this device.

So my solution is that I delay `process_weights_after_loading` after I load the lora model. After I load the lora model, I still call `move_weights_to_torchax_tensor ` to shard/move the model weights to TPU.

tpu_commons/models/vllm/vllm_model_wrapper.py
vanbasten23: replied in https://github.com/vllm-project/tpu_commons/pull/631#discussion_r2322718295

tpu_commons/models/vllm/vllm_model_wrapper.py
vanbasten23: Not sure I understand the question. I'm not calling `process_weights_after_loading ` anymore.

tpu_commons/models/vllm/vllm_model_wrapper.py
vanbasten23: actually it's used in `LoRAModelRunnerMixin` so I need the model runner to hold a reference to the lora_manager

tpu_commons/runner/compilation_manager.py
vanbasten23: sure

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,19 @@ steps:
            --ignore=/workspace/tpu_commons/tests/kernels/ragged_paged_attention_kernel_v2_test.py \
            --ignore=/workspace/tpu_commons/tests/kernels/ragged_kv_cache_update_v2_test.py
 
+   - label: "lora tests for JAX + vLLM models"
```
xiangxu-google, line 103: Hi, do we support running lora with JAX models? if not what's the gap there

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: > the base model's device becomes 'jax', then vLLM will move the lora weight to the same device as the base_model which is jax, and pytorch doesn't recognize this device.

Can you explain more on this part? Shouldn't torchax / torchview handle mapping between jax array / pytorch array so we don't have to worry about this kind of things?

tpu_commons/models/vllm/vllm_model_wrapper.py
kyuyeunk: sorry, i meant to write "now renamed to `move_weights_to_torchax_tensor`") but this has been answered in the different comment.

tpu_commons/runner/tpu_jax_runner.py
```diff
@@ -625,6 +631,20 @@ def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
              self.mesh, (input_ids, positions, block_tables, query_start_loc,
                          seq_lens, logits_indices, request_distribution))
 
+        if self.lora_config is not None:
```
xiangxu-google, line 665: Can you move lora related logic to a separate file `lora_manager.py`, this is for keeping the runner file short.
example: https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/runner/speculative_decoding_manager.py

tpu_commons/models/vllm/quantization/unquantized.py
vanbasten23: Sure. The process looks like this (before this change):
1. We load the base model (via `vllm_get_model`). It will first load the model , then shard the model weights and move the weight to TPU (layer.weight=='jax', via `process_weights_after_loading `)
2. We load the lora model (via `load_lora_model`). It will initialize the lora weight on the same device as the base model via `torch.zeros(..., device=base_model_device)`. Here, `base_model_device` will be `jax` and `torch.zeros(..., device='jax')` will fail because pytorch doesn't recognize this device. 

Oh do you mean in the step #2, I can put `load_lora_model` under `with torchax.default_env()` and let torchax handles it?

.buildkite/pipeline_jax.yml
vanbasten23: Hi Xiang, we don't support it yet. I thought about it and list the high level gap/AI [here](https://docs.google.com/document/d/1jAk72asfFya_yw_V9CwP5Z0AGJ6PelpMzFXGZhB_I4M/edit?pli=1&tab=t.0#bookmark=id.5cqajvpuyukc). Feel free to take a look.

tpu_commons/models/vllm/quantization/unquantized.py
kyuyeunk: Yes, I meant if we can use step 2 and let torchax handle computation.

tpu_commons/runner/tpu_jax_runner.py
vanbasten23: done. Created a lora_utils.py. I didn't create a lora_manager because in lora there are already 2 manager concepts: LoraWorkerManager and LoraModelManager. To avoid the confusion, I named it util. The runner file is shortened.

tpu_commons/models/vllm/quantization/unquantized.py
vanbasten23: Fixed it by reusing `process_weights_after_loading`.

I met some bumper because after I put `load_lora_model` in a torchax env, some tensor becomes torchax tensor, then I got a few cases where there is mixed of torchax tensors and torch.Tensor. But they're all resolved.

tpu_commons/models/vllm/quantization/common.py
```diff
@@ -2,14 +2,20 @@
 from jax.sharding import Mesh, PartitionSpec
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
-from vllm.model_executor.layers.linear import (
-    ColumnParallelLinear, LinearBase, MergedColumnParallelLinear,
-    QKVParallelLinear, ReplicatedLinear, RowParallelLinear)
+# yapf: disable
```
lsy323, line 5: nit: why do we disable yapf here

tpu_commons/runner/tpu_jax_runner.py
```diff
@@ -281,7 +287,7 @@ def _init_inputs(self) -> None:
                                             dtype=np.int64)
 
     def load_model(self):
-        self.model_fn, self.compute_logits_fn, self.get_multimodal_embeddings_fn, self.get_input_embeddings_fn, self.state = get_model(
+        self.model_fn, self.compute_logits_fn, self.get_multimodal_embeddings_fn, self.get_input_embeddings_fn, self.state, self.lora_manager, self.model = get_model(
```
lsy323, line 290: qq: where is self.model and self.lora_manager used?

tpu_commons/lora/torch_punica_tpu.py
kyuyeunk, line 1: Since this is a torch code, would it make sense to move this into `tpu_commons/models/vllm` directory? Or can it be expanded to support both torch and jax code?

tpu_commons/utils.py
kyuyeunk, line 1: I think it is better to consolidate logic for sharding torch tensors into `tpu_commons/models/vllm/sharding.py` file. wdyt?

tpu_commons/models/vllm/quantization/common.py
vanbasten23: isort and yapf have conflicting order so I need to disable yapf.

tpu_commons/runner/tpu_jax_runner.py
vanbasten23: `self.model` is used in `shard_lora_weights_and_move_to_tpu` and it shards/moves the lora_a/lora_b/lora_bias to TPU.

`self.lora_manager` is used in `LoRAModelRunnerMixin` for example https://github.com/vllm-project/vllm/blob/2e5d21378db8226b04bb76f3e7119a9ddd0239df/vllm/v1/worker/lora_model_runner_mixin.py#L75

tpu_commons/lora/torch_punica_tpu.py
vanbasten23: currently I put it to the `lora` dir since it's all lora specific logic. If later we support lora in jax, we can put the current one in `tpu_commons/lora/vllm` and the jax one in `tpu_commons/lora/jax`

tpu_commons/utils.py
vanbasten23: SG. Fixed it

tpu_commons/models/vllm/sharding.py
```diff
@@ -168,7 +169,8 @@ def set_one(module, prefix):
             qual_name = prefix + k
             if (potential_v := buffers.get(qual_name)) is not None or (
                     potential_v := variables.get(qual_name)) is not None:
-                setattr(module, k, potential_v)
+                if getattr(module, k, None) is None:
```
kyuyeunk, line 172: What are the purpose of this change? If the `k` is only replaced only if it's value is `None`, how is it supposed to replace existing `k` value?

vanbasten23: I remember I got an error at `setattr(module, k, potential_v)` saying module.k already exist so I check `if getattr(module, k, None) is None`

kyuyeunk: Maybe you need to do something like this here and delete the original variable first? https://github.com/vllm-project/tpu_commons/blob/39414cb7ae65c49b9d57520cc39cf4dbab4411db/tpu_commons/models/vllm/quantization/unquantized.py#L77-L78

lsy323:
torchax path lgtm! Thanks @vanbasten23 !

lsy323:
cc @xiangxu-google @py4 could you please help take another pass at the changes in jax runner? Thank you!

vanbasten23:
The CI looks good https://screenshot.googleplex.com/9yjQgUe2dDwVTGc. The failure is unrelated to this PR as it also appear on HEAD: https://screenshot.googleplex.com/9CFHEy8SnVjMXV9

vanbasten23:
Thanks for the review!

PR #641, opened 2025-09-04
feat: Add benchmark test for Buildkite CI procedure

meiyeh123:
# Description

Based on the Benchmark Test requirements, this commit introduces a new benchmark testing mechanism for our CI/CD
pipeline.

It runs a Docker container, starts the vLLM server, and executes a benchmark with varying inputs, outputs, and prefixes.

The changes also add the implement of Sonnet and Random Dataset in benchmark_serving.py. All of benchmark tests do the benchmark with Sonnet Dataset.

# Tests

[BuildKite](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/3973/steps/canvas)

# Checklist


.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+
+   - label: "Stress Test llama-3.1-70B Input:16000 Output:16000 Prefix:0"
+     key: stress_test_llama_3_1_70B_i16000_o16000_p0
+     soft_fail: true
+     depends_on:
+       - validation_unit_test_result
+     env:
+       USE_V6E8_QUEUE: "True"
+     agents:
+       queue: tpu_v6e_queue
```
jcyang43: Wrong queue here? You have `tpu_v6e_queue` with `USE_V6E8_QUEUE: "True"`

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-8B Input:2048 Output:8192 Prefix:0"
+     key: stress_test_llama_3_1_8B_i2048_o8192_p0
```
jcyang43: For the naming, let's use their original intent (Decode heavy - reasoning). It's more readable & meaningful than the input # - output # - prefix # combination

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-70B Input:2048 Output:8192 Prefix:0"
+     key: stress_test_llama_3_1_70B_i2048_o8192_p0
```
jcyang43: Ditto for the naming. Decode heavy - reasoning

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-8B Input:1024 Output:4096 Prefix:0"
+     key: stress_test_llama_3_1_8B_i1024_o4096_p0
```
jcyang43: Ditto for the naming. Decode heavy - text generation

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       PREFIX_LENGTH: 0
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-70B Input:1024 Output:4096 Prefix:0"
```
jcyang43: Ditto for the naming. Decode heavy - text generation

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-8B Input:4096 Output:1024 Prefix:3600"
+     key: stress_test_llama_3_1_8B_i4096_o1024_p3600
```
jcyang43: Ditto. Effective prefill heavy - chatbot

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-70B Input:4096 Output:1024 Prefix:3600"
+     key: stress_test_llama_3_1_70B_i4096_o1024_p3600
```
jcyang43: Ditto. Effective prefill heavy - chatbot

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-8B Input:8192 Output:2048 Prefix:7900"
+     key: stress_test_llama_3_1_8B_i8192_o2048_p7900
```
jcyang43: Ditto. Effective prefill heavy - code generation

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-70B Input:8192 Output:2048 Prefix:7900"
+     key: stress_test_llama_3_1_70B_i8192_o2048_p7900
```
jcyang43: Ditto. Effective prefill heavy - code generation

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+
+   - label: "Stress Test llama-3.1-8B Input:30 Output:30 Prefix:0"
+     key: stress_test_llama_3_1_8B_i30_o30_p0
```
jcyang43: Ditto. Short sequence - machine translation

.buildkite/pipeline_jax.yml
```diff
@@ -99,6 +99,343 @@ steps:
...
+       THROUGHPUT_BASE_PATH: "/workspace/"
+       BUCKET_NAME: "tpu-commons-ci"
+     commands:
+       - export THROUGHPUT_BASE_FILE="throughput_baseline_I$${INPUT_LENGTH}_O$${OUTPUT_LENGTH}_P$${PREFIX_LENGTH}.json"
+       - bash tests/e2e/benchmarking/download_throughput.sh gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE ./
+       - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_commons/tests/e2e/benchmarking/stress_test.sh -i $$INPUT_LENGTH -o $$OUTPUT_LENGTH -pl $$PREFIX_LENGTH -t $$THROUGHPUT_BASE_FILE
+       - gsutil cp $$THROUGHPUT_BASE_FILE gs://$$BUCKET_NAME/$$THROUGHPUT_BASE_FILE
+  
+   - label: "Stress Test llama-3.1-70B Input:30 Output:30 Prefix:0"
+     key: stress_test_llama_3_1_70B_i30_o30_p0
```
jcyang43: Ditto. Short sequence - machine translation

.buildkite/scripts/run_in_docker.sh
```diff
@@ -11,6 +11,20 @@ if [ "$#" -eq 0 ]; then
   exit 1
 fi
 
+# Touch the default file for volume
+touch "$(pwd)/$THROUGHPUT_BASE_FILE"
```
jcyang43: Please leave a comment explaining what `THROUGHPUT_BASE_FILE` is used for. (Or rename it such that the name is self-explanatory)

tests/e2e/benchmarking/download_throughput.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+GCS_SOURCE="$1"
+LOCAL_DEST="$2"
+
+if gsutil cp "$GCS_SOURCE" "$LOCAL_DEST"; then
+    echo "Success: File downloaded."
```
jcyang43: Can we improve the logging? When people see this debug log, it doesn't provide any information on what file is downloaded. For example, something like `"Successfully downloaded baseline throughput file from gcs bucket"` will be a lot more meaningful.

tests/e2e/benchmarking/stress_test.sh
```diff
@@ -0,0 +1,312 @@
+#!/bin/bash
+
+# Log the vLLM server output to a file
+LOG_FILE="server.log"
+BENCHMARK_LOG_FILE="benchmark.log"
+# The sentinel message that indicates the server is ready (in LOG_FILE)
+READY_MESSAGE="Application startup complete."
+# After how long we should timeout if the server doesn't start
+TIMEOUT_SECONDS=1800
```
jcyang43: May need to check on this timeout (1hr). Based on our previous experience with manual testing, some stress test cases exceed an hour

tests/e2e/benchmarking/stress_test.sh
```diff
@@ -0,0 +1,312 @@
...
+   echo -e "\t-r, --root-dir-path\tThe path to your root directory (default: /workspace/, which is used in the Dockerfile)"
+   echo -e "\t-d, --dataset-name\tThe name of the dataset to use (default: random)"
+   echo -e "\t-p, --dataset-path\tThe path to the processed dataset. This is required when using a custom model (default: None)."
+   echo -e "\t-m, --model-list\tA space-separated list of HuggingFace model IDs (default: "meta-llama/Llama-3.1-8B-Instruct" or "meta-llama/Llama-3.1-70B-Instruct")"
+   echo -e "\t-i, --input-len\tThe input sequence length for the benchmark (default: 1024)"
+   echo -e "\t-o, --output-len\tThe output sequence length for the benchmark (default: 1024)"
+   echo -e "\t-pl, --prefix-len\tThe prefix length for the benchmark (default: 0)"
+   echo -e "\t-v, --vllm-download-dir\tThe directory to download vLLM into (default: /tmp/hf_home)"
+   echo -e "\t-t, --throughput\tThe JSON file that stores expected throughput values for each model. (default: None, a default throughput of 1500 will be used for all models.)"
```
jcyang43: 1. Can we rename this from `throughput` to `throughput-file`?
2. The "default" part is a little misleading/confusing? The default will be the baseline throughput stored in gcs bucket instead of 1500 for all models

.buildkite/pipeline_jax.yml
meiyeh123: remove this step, since it's 16K
BTW, I remove all 16K tests

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/pipeline_jax.yml
meiyeh123: done

.buildkite/scripts/run_in_docker.sh
meiyeh123: done. add some comment for clear.
And also fix the bug that:
  If the step (ex. other unit steps) doesn't have the env $THROUGHPUT_BASE_FILE, the test will fail.

tests/e2e/benchmarking/download_throughput.sh
meiyeh123: done

tests/e2e/benchmarking/stress_test.sh
meiyeh123: done

tests/e2e/benchmarking/stress_test.sh
meiyeh123: 1. done
2. I change the default to `throughput_file="throughput_baseline.json"`, and add a comment to clarify that if we don't pass the throughput_file, it doesn't make sense for this part. the procedure won't have error, but not meaningful.

meiyeh123:
FYI, the latest version do:
1.  implement the RandomDataset in benchmark_serving.py.
2. solve the reviewer's comments.
3. remove the 16K steps.
4. fix the bug that the run_in_docker.sh would be failed when running Unit test steps (Due to no include the env THROUGHPUT_BASE_FILE in the step.)

the latest test [results](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2727/steps/canvas)

jcyang43:
> FYI, the latest version do:
> 
> 1. implement the RandomDataset in benchmark_serving.py.
> 2. solve the reviewer's comments.
> 3. remove the 16K steps.
> 4. fix the bug that the run_in_docker.sh would be failed when running Unit test steps (Due to no include the env THROUGHPUT_BASE_FILE in the step.)
> 
> the latest test [results](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2727/steps/canvas)

Hi Dennis, it looks like this PR also copies some files over from the vllm upstream (ex: `scripts/vllm/lora/` & `scripts/vllm/multimodal/`). Can you explain this part? Thanks

tests/e2e/benchmarking/stress_test.sh
```diff
@@ -0,0 +1,312 @@
...
+dataset_name=random
+dataset_path=""
+vllm_download_dir=/tmp/hf_home
+input_len=1024
+output_len=1024
+prefix_len=0
+exit_code=0
+
+# The minimum throughput scores we expect
+target_throughput=1500
```
jcyang43: As discussed offline, this may be less than ideal. Let's improve this

jcyang43:
Hi @yarongmu-google, some of the stress test cases/configurations can take a long time (> 3 hrs) to run from our experience with manual testing (specifically when tensor parallel size is only 1 and the input/output/token lengths are high). Not sure if the long runtime will disrupt our workflow given that we'll be adding more models in the future and the runtime of a CI job can add up quickly. Do we want to try one of the following? 
1. Add more Buildkite agents to handle the load
2. Try to reduce the runtime of each stress test (Currently each stress test is a performance benchmark based on 1000 requests; maybe cut that down to 500 requests for example?)
3. Some combination of the two above
4. Something else 

Thanks!

p.s. The runtime concern is also the reason the PR breaks down each stress test into its own in the pipeline yaml (as opposed to having a single script that loops through all the models/configurations) so that they can run in parallel on different buildkite agents to reduce the total runtime.

meiyeh123:
> > FYI, the latest version do:
> > 
> > 1. implement the RandomDataset in benchmark_serving.py.
> > 2. solve the reviewer's comments.
> > 3. remove the 16K steps.
> > 4. fix the bug that the run_in_docker.sh would be failed when running Unit test steps (Due to no include the env THROUGHPUT_BASE_FILE in the step.)
> > 
> > the latest test [results](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2727/steps/canvas)
> 
> Hi Dennis, it looks like this PR also copies some files over from the vllm upstream (ex: `scripts/vllm/lora/` & `scripts/vllm/multimodal/`). Can you explain this part? Thanks

Yes, we want to use benchmark_serving.py from tpu_common, but it doesn’t implement the RandomDataset component, which we need to execute. So, I copied some code and dependencies from the vLLM upstream.

yarongmu-google:
Hi @jcyang43:

Thanks! Please see my response inline :)

> Hi @yarongmu-google, some of the stress test cases/configurations can take a long time (> 3 hrs) to run from our experience with manual testing (specifically when tensor parallel size is only 1 and the input/output/token lengths are high). Not sure if the long runtime will disrupt our workflow given that we'll be adding more models in the future and the runtime of a CI job can add up quickly. Do we want to try one of the following?
> 
> 1. Add more Buildkite agents to handle the load

Where are we going to set up this test? I mean, did we add new TPU capacity, or did we share the existing capacity for CI/benchmarking/auto-tune? If new capacity, the disruption would be minimal, assuming we indeed:
(A). set up this job as incremental (ie, only models that have successfully passed the auto-tune/benchmakring stage will be run by this)
(B). set up this as post-merge (ie not blocking PR submissions).

I vote for more buildkite agents, but the underlying TPU capacity os the key to this.

> 2. Try to reduce the runtime of each stress test (Currently each stress test is a performance benchmark based on 1000 requests; maybe cut that down to 500 requests for example?)

This can be done, but we need to take caution. The thing is to make the tests meaningful, we will need to make sure in each test, the TPUs run in "steady state" for at least a few (>= 3) steps. Otherwise, if the TPU is only ramping up and ramping down, the test results are not meaningful. (This is why vLLM initially chose 1000, which is reasonable but do not always hold). During your current runs, have you happened to also noted whether the TPU entered a steady state for each test?

> 3. Some combination of the two above
> 4. Something else
> 
> Thanks!
> 
> p.s. The runtime concern is also the reason the PR breaks down each stress test into its own in the pipeline yaml (as opposed to having a single script that loops through all the models/configurations) so that they can run in parallel on different buildkite agents to reduce the total runtime.

Thanks - this is something we can improve in the future, right?

.buildkite/scripts/bootstrap.sh
```diff
@@ -2,25 +2,25 @@
...
+#     echo "No 'ready' label found on PR. Skipping main pipeline upload."
+#     exit 0 # Exit with 0 to indicate success (no error, just skipped)
+#   fi
+# else
+#   # If it's NOT a Pull Request (e.g., branch push, tag, manual build)
+#   echo "This is not a Pull Request build. Uploading main pipeline."
+#   buildkite-agent pipeline upload .buildkite/pipeline_jax.yml
+#   # buildkite-agent pipeline upload .buildkite/pipeline_torch.yml
+# fi
+buildkite-agent pipeline upload .buildkite/pipeline_jax.yml
```
yarongmu-google: The logic for checking pull request labels has been commented out, and the pipeline is now uploaded unconditionally. If the intention is to run the stress tests on all pull requests, this is fine. However, if you want to control when the tests are run (e.g., only when a PR has passed all previous testing stages), you will need to uncomment this block.

tests/e2e/benchmarking/stress_test.sh
```diff
@@ -0,0 +1,310 @@
...
+            break
+        fi
+    done
+
+
+    if $did_find_ready_message; then
+        echo "Starting the benchmark for $model_name..."
+        echo "Current working directory: $(pwd)"
+        pip install pandas
+        pip install datasets
```
yarongmu-google: The cleanUp function uses pkill and pgrep to stop the vLLM server. This can be unreliable if multiple vLLM processes are running on the same machine. A more robust approach would be to save the Process ID (PID) of the server when it's started and then use that PID to kill the specific process.

Maybe modify the server startup to save the PID, and then use that PID in the cleanUp function.

.buildkite/pipeline_jax.yml
```diff
@@ -2,121 +2,359 @@ steps:
...
+  #      - test_8
+  #    agents:
+  #      queue: tpu_v6e_queue
+  #    commands:
+  #      - echo "Validation Unit Test Result"
+
+  #  - label: "Decode heavy - reasoning 8B"
+  #    key: decode_heavy_reasoning_8B
+  #    soft_fail: true
+  #    depends_on:
```
yarongmu-google: The commands section for the short_sequence_machine_translation_8B step contains a fair amount of shell script logic. To improve readability and maintainability of the pipeline YAML, this logic could be moved into a separate script. This would also make it easier to reuse this logic for the other (currently commented-out) stress test jobs.

tests/e2e/benchmarking/stress_test.sh
meiyeh123: I believe this should not be an issue, as `run_in_docker.sh bash stress_test.sh` runs in a new Docker container, ensuring that the environment is fully isolated.

meiyeh123:
This PR is WIP, as we are still confirming the further requirements with google.

tests/e2e/benchmarking/stress_test.sh
```diff
@@ -0,0 +1,288 @@
...
+throughput_file_path=""
+
+helpFunction()
+{
+   echo ""
+   echo "Usage: $0 [-r root-dir-path] [-d dataset-name] [-p dataset-path] [-m model-list] [-i input-len] [-o output-len] [-pl prefix-len] [-v vllm-download-dir] [-t throughput] [-h]"
+   echo -e "\t-r, --root-dir-path\tThe path to your root directory (default: /workspace/, which is used in the Dockerfile)"
+   echo -e "\t-d, --dataset-name\tThe name of the dataset to use (default: sonnet)"
+   echo -e "\t-p, --dataset-path\tThe path to the processed dataset. This is required when using a custom model (default: benchmarks/sonnet.txt)."
+   echo -e "\t-m, --model-list\tA space-separated list of HuggingFace model IDs (default: "meta-llama/Llama-3.1-8B-Instruct" or "meta-llama/Llama-3.1-70B-Instruct")"
```
jcyang43: Please review all the arguments here. Ex: here you have `--model-list` in the usage but `--model` below

meiyeh123: Done

meiyeh123:
The current commit remove the model list and rename to  benchmark.sh
[(buildkite test)](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3498/steps/canvas)

tests/e2e/benchmarking/benchmark.sh
```diff
@@ -0,0 +1,251 @@
...
+    if [ "$dataset_name" = "sonnet" ]; then
+        dataset_args="--sonnet-input-len $input_len --sonnet-output-len $output_len --sonnet-prefix-len $prefix_len"
+    elif [ "$dataset_name" = "random" ]; then
+        dataset_args="--random-input-len $input_len --random-output-len $output_len --random-prefix-len $prefix_len" 
+    fi
+
+    echo "Starting the benchmark for $model..."
+    echo "Current working directory: $(pwd)"
+    pip install pandas
+    pip install datasets
```
jcyang43: Datasets is already specified in [requirements_benchmarking.txt](https://github.com/vllm-project/tpu_commons/blob/88e2c21f6c459a3e571fa74271615202690238bf/requirements_benchmarking.txt#L5). If you need pandas, please move that to that requirements file as well

meiyeh123: done

meiyeh123:
the new commit replace some args with env, and move the pandas installation to requirements_benchmarking.txt
([buildkite test](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3560/steps/canvas?sid=0199995f-fd4a-46c5-ad5f-0c58fc78e46e))

vipannalla:
@QiliangCui, can you also take a look?

.buildkite/scripts/run_in_docker.sh
```diff
@@ -11,6 +11,15 @@ if [ "$#" -eq 0 ]; then
   exit 1
 fi
 
+ENV_VARS=(
```
jcyang43, line 14: This will cause errors for places that use the `run_in_docker.sh` script without these environment variables (ex: `mlperf.sh`)

meiyeh123: done. thanks

meiyeh123:
Ensure the tests for both the original and new pipelines pass. The test is on a separate branch from this commit, and it adds the file [‎.buildkite/models/meta-llama_Llama-3_1-8B.yml](https://github.com/vllm-project/tpu-inference/commit/8f9134e0042e0184f0b784915ff2c04ff370b7e4#diff-e1ae40d66f274103fb94e3a874722a085fd501564f740ffed61b73195faf9b4b) for testing. ([buildkite](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/3973/steps/canvas))

FYI, this commit was rebased on the previous commit (56f4e99f) from the main branch (68fc6329) because the main branch couldn't pass the JAX unit tests.

PR #648, opened 2025-09-04
[JAX] vllm-compatible jax model registration

karan:
# Description

JAX models follow the current interface with forward pass in __call__. For custom out of tree models, we define a new tpu_commons model registration flow that:

1. Registers the model with tpu_commons.
2. Patches the class with a vLLM-compatible unimplemented forward method.
3. Registers the model [with](https://github.com/vllm-project/vllm/blob/78336a0c3ee4eb9dba6e37959d926160e91623fd/vllm/model_executor/models/registry.py#L459) vLLM.

# Tests

Tested with the following command:

```
TPU_BACKEND_TYPE=jax \
  python -m vllm.entrypoints.cli.main serve \
  /path/to/huggingface/cached/snapshot \
  --max-model-len=1024 \
  --tensor-parallel-size 8 \
  --max-num-batched-tokens 1024 \
  --max-num-seqs=1 \
  --disable-log-requests
```

```
curl -X POST http://127.0.0.1:8000/v1/completions -H "Content-Type: application/json" -d '{"prompt": "The colors of the rainbow are", "max_tokens": 10}' | jq -r
```

# Checklist


karan:
@kyuyeunk as well

tests/models/jax/test_model_loader.py
```diff
@@ -80,25 +81,62 @@ def clear_model_registry_after_test():
 
 def test_register_model_new_arch():
     """Tests registering a new model architecture."""
-    model_loader.register_model("NewArch", MockModelA)
-    config = PretrainedConfig(architectures=["NewArch"])
+    arch = "NewArch"
+    model_loader.register_model(arch, MockModelA)
```
kyuyeunk, line 85: Instead of a mock model, is it possible to test it with a proper model with a `__call__` function?

We want to see that it correctly calls `__call__` function of a model, create `forward` function that allows vllm model registration.

tpu_commons/models/jax/model_loader.py
```diff
@@ -240,14 +241,55 @@ def get_model(
...
-    # TODO: Support lazy loading (like vllm)
-    # TODO: Also call vllm's `ModelRegistry.register_model`.
+    # Register with tpu_commons registry for the JAX backend
     _MODEL_REGISTRY[arch] = model
+
+    # Create a vLLM-compatible wrapper for the JAX model class.
+    # This wrapper inherits from the JAX model and torch.nn.Module
+    # to pass vLLM's type checks. It is not meant to be instantiated
+    # or executed by vLLM's PyTorch backend.
+    def unimplemented_forward(
```
kyuyeunk, line 263: Does it mean that model registration will pass even if model's call function is completely random? Is it possible to add a vllm torch validation logic equivalent for jax/flax in here as well?

Specific details can differ, but something like checking if `__call__(self, input_ids: jax.Array, positions: jax.Array, intermediate_tensors: Optional[Any] = None,  inputs_embeds: Optional[jax.Array] = None) -> None` exists or not.

tests/models/jax/test_model_loader.py
karan: The code added here does not change how a model is called at all. And the model registry isn't responsible for calling the model either.

All these tests are doing is that validating that the registry is working as expected (updating the dictionary) - ensuring that the model runner is working would make the test too complex and brittle here IMO.

What you are describing I think is better suited as an e2e test (which I will run manually with the script you had shared).

tpu_commons/models/jax/model_loader.py
karan: I don't mind adding the validation here but I'd like the team to chime in since they know future plans for the interface better than I do.

@xiangxu-google Wdyt? 

Side note -  I'd prefer to add the validation as a follow-up to keep this PR very focussed on just model registration.

xiangxu-google: it's good to add the check in the later PRs

PR #652, opened 2025-09-05
[Test][Spec Decoding] Make e2e performance test for random workload less tight.

py4:
# Checklist


yaochengji:
@py4 , thanks for the quick fix. Do you know why the speedup is a bit lower?

py4:
@yaochengji The number was set based on offline benchmark but I noticed in offline benchmark I used `ignore_eos=False` but in the test I am using `ignore_eos=True`. When I changed the offline test to True as well, The speed up is ~3.8

PR #657, opened 2025-09-08
[Bugfix] Cleanup docker by IDs

patemotter:
# Description

Perform cleanup on Docker images using IDs instead of relying on image names. 

# Tests

Tested locally on setups that could have triggered the cases in a similar manner to what we are seeing in the test pipeline. 

# Checklist

### DIFF
--- a/.buildkite/scripts/run_in_docker.sh
+++ b/.buildkite/scripts/run_in_docker.sh
@@ -51,17 +51,29 @@ DOCKER_HF_HOME="/tmp/hf_home"
 
 # Cleanup of existing containers and images.
 echo "Starting cleanup for vllm-tpu..."
-leftover_containers=$(docker ps -a -q --filter "ancestor=vllm-tpu")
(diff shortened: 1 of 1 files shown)


py4:
@patemotter Tnx. isn't the issue that we don't do `docker stop CONTAINER_THAT_USE_IMAGE_X` before running `docker rmi IMAGE_X`?

patemotter:
> @patemotter Tnx. isn't the issue that we don't do `docker stop CONTAINER_THAT_USE_IMAGE_X` before running `docker rmi IMAGE_X`?

No that's done inherently as a part of the `docker rm -f` command because it sends a `SIGKILL` to terminate any associated container before removing it. 

So `docker rm -f "$leftover_containers"` from the original code stops the containers. 

I tested the original cleanup code locally and it works fine with different containers running or not. But it fails with the exact error we see in the tests when only using the names to search instead of by id.

PR #661, opened 2025-09-09
Fix structured decoding and sampler

Chenyaaang:
# Description

Fix TPU structured decoding and sampler

1. Same as https://github.com/vllm-project/vllm/pull/24458, when there are guided and non-guided requests in the same batch, `scheduler_output.grammar_bitmask` only contains the entries for guided requests, sort the guided requests and use cumulative index to avoid index out of bound error.
2. The sampler should be only applied to temperature > 0, for requests with temperature < 0, they should apply the all-greedy results, otherwise, 
`logits /= temperatures` will flip the result, causing error (e.g. result is !!!!!!!!, where ! refers to token 0). 

# Tests

- vllm/tests/v1/entrypoints/llm/test_struct_output_generate.py passed.
- CI: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2802

# Checklist

### DIFF
--- a/tpu_commons/runner/structured_decoding_manager.py
+++ b/tpu_commons/runner/structured_decoding_manager.py
@@ -61,27 +61,22 @@ class StructuredDecodingManager:
         self.runner.grammar_bitmask_cpu.fill(0)
         self.runner.require_structured_out_cpu.fill(0)
 
-        # We receive the structured output bitmask from the scheduler, but the
-        # indices of the requests in the batch may not match the indices of
-        # the bitmask since the scheduler doesn't know how the tpu runner is
-        # ordering the requests in the batch. We need to match the order of
-        # bitmask with the order of requests
-        struct_out_indices: list[int] = []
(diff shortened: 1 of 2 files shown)


tpu_commons/runner/structured_decoding_manager.py
```diff
@@ -61,27 +61,22 @@ def prepare_structured_decoding_input(
...
-        # the bitmask since the scheduler doesn't know how the tpu runner is
-        # ordering the requests in the batch. We need to match the order of
-        # bitmask with the order of requests
-        struct_out_indices: list[int] = []
-        mask_indices: list[int] = []
-        for req_id in self.runner.input_batch.req_ids:
-            mask_index = scheduler_output.structured_output_request_ids.get(
-                req_id)
-            if mask_index is None:
+        sorted_struct_requests = sorted(
```
py4, line 64: @Chenyaaang would be good to have a comment here what's happening here and below

PR #667, opened 2025-09-10
[Bugfix] fix the hash name util

yaochengji:
# Description

Fixed the issue caused by https://github.com/vllm-project/vllm/pull/23673/


vanbasten23:
Thanks for the fix Chengji!

PR #668, opened 2025-09-10
[Bugfix] fix model loader

yaochengji:
# Description

Fixed the CI test caused by https://github.com/vllm-project/tpu_commons/pull/631


vanbasten23:
Thanks Chengji for the fix!

PR #671, opened 2025-09-10
[Spec Decoding] Implemented Llama-based eagle-3 model.

Lumosis:
# Description

To enable eagle-3, add 
```
    args['speculative_config'] = {
        'model': "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B",
        "num_speculative_tokens": 3,
        "method": "eagle3",
        "draft_tensor_parallel_size": 1
    }
```
to offline_inference.py

# Tests

Unit tested

# Checklist


tpu_commons/models/jax/model_loader.py
```diff
@@ -233,17 +235,65 @@ def get_model(
...
+            return get_flax_draft_model(vllm_config, rng, mesh)
+        else:
+            return get_flax_model(vllm_config, rng, mesh)
     elif impl == "vllm":
         return get_vllm_model(vllm_config, rng, mesh)
     else:
         raise NotImplementedError("Unsupported MODEL_IMPL_TYPE")
 
 
+def get_flax_draft_model(
```
py4: Why do we need a seperate `get_flax_draft_model`?

tpu_commons/spec_decode/jax/eagle.py
```diff
@@ -2,10 +2,12 @@
...
 
+import jax
 import jax.numpy as jnp
 from vllm.config import VllmConfig
 
 from tpu_commons.models.jax.attention_metadata import AttentionMetadata
+from tpu_commons.models.jax.model_loader import get_model
 
 
 class EagleProposer:
```
py4, line 13: Let's rename this class and this file to Eagle3?

tpu_commons/models/jax/utils/weight_utils.py
```diff
@@ -275,6 +276,9 @@ def _load_hf_weights_on_thread(vllm_config,
         if hf_key.endswith(".weight"):
             hf_key = hf_key.removesuffix(".weight")
 
+        if 'midlayer' in hf_key:
```
py4: Can you explain this part?

tpu_commons/models/jax/utils/weight_utils.py
```diff
@@ -291,6 +295,9 @@ def _load_hf_weights_on_thread(vllm_config,
                 logger.warning(
                     f"Skip loading {hf_key} due to tie_word_embeddings")
                 continue
+            if hf_key not in name_map and hf_key == "t2d":
```
py4: Can you explain this part? what is "t2d"?

tpu_commons/runner/tpu_jax_runner.py
```diff
@@ -292,6 +295,11 @@ def load_model(self):
             self.rng_key,
             self.mesh,
         )
+
+        if hasattr(self, "drafter"):
```
py4: Maybe set `self.drafter = None` as default and just check if it's not None?

tpu_commons/models/jax/utils/weight_utils.py
Lumosis: t2d is target_id_to_draft_id map but I don't think it's used for eagle-3 for now. The corresponding part is github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/llama_eagle3.py#L271

tpu_commons/models/jax/utils/weight_utils.py
Lumosis: The eagle-3 model has only one layer which is named as `midlayer` in the checkpoint. The corresponding part in vllm is github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/llama_eagle3.py#L175

Lumosis:
> Thank you @Lumosis ! Can we also add a TODO note regarding sharing embeding and lm_head with target model?

Done :)

tpu_commons/models/jax/utils/weight_utils.py
py4: I think maybe this is better to go into `metadata_map` and get handled by this block:

```
        else:
            if hf_key not in name_map and hf_key == "lm_head":
                logger.warning(
                    f"Skip loading {hf_key} due to tie_word_embeddings")
                continue
            model_key = name_map.get(hf_key, hf_key)
```

The reason is that this current function is supposed to be generic and adding eagle-3 specific logic to it doesn't feel right

tpu_commons/models/jax/utils/weight_utils.py
py4: That code blocks check if "t2d" exists in key, is `hf_key == t2d` right comparison?

tpu_commons/models/jax/utils/file_utils.py
```diff
@@ -91,6 +91,7 @@ def download_model_weights_from_hf(model_path: str, cache_dir: Optional[str],
             allow_patterns=weights_format,
             tqdm_class=DisabledTqdm,
             local_files_only=huggingface_hub.constants.HF_HUB_OFFLINE,
+            revision="refs/pr/4",
```
py4: What does this line do?

tpu_commons/runner/tpu_jax_runner.py
```diff
@@ -186,9 +186,13 @@ def _init_mm(self) -> None:
         self.uses_mrope = self.model_config.uses_mrope
 
     def _init_speculative_decoding(self) -> None:
+        self.drafter = None
         if self.speculative_config:
             if self.speculative_config.method == "ngram":
                 self.drafter = NgramProposer(self.vllm_config)
+            elif self.speculative_config.method == "eagle3":
+                from tpu_commons.spec_decode.jax.eagle3 import EagleProposer
```
py4: Should this import be moved to top level?

tpu_commons/models/jax/utils/file_utils.py
Lumosis: Sorry, removed

PR #674, opened 2025-09-10
[Spec Decoding] Add EAGLE-3 integration.

py4:
# Checklist


tpu_commons/models/jax/llama3.py
```diff
@@ -237,14 +237,27 @@ def __init__(self, vllm_config: VllmConfig, rng: nnx.Rngs,
                 sharding=(None, "model"),
             )
 
+        # if spec config is enable and eagle3, set should should_return_aux_hidden_states to True
```
Lumosis: The comment looks outdated

tpu_commons/runner/speculative_decoding_manager.py
```diff
@@ -44,9 +46,86 @@ def take_draft_token_ids(self) -> Optional[DraftTokenIds]:
...
+        spec_decode_metadata: Optional[SpecDecodeMetadata],
     ) -> None:
-        self._draft_token_ids = self.propose_ngram_draft_token_ids(
-            sampled_token_ids)
+        if self.runner.speculative_config.method == "ngram":
+            self._draft_token_ids = self.propose_ngram_draft_token_ids(
+                sampled_token_ids)
+        elif self.runner.speculative_config.use_eagle():
+            assert self.runner.speculative_config.method == "eagle3"
+            if self.runner.speculative_config.method != "eagle3":
```
Lumosis: It seems redundant to have two checks

tpu_commons/models/jax/model_loader.py
```diff
@@ -162,6 +162,7 @@ def get_flax_model(
         out_shardings=(
             kv_cache_sharding,
             hidden_states_sharding,
+            hidden_states_sharding,  # aux hidden states
```
Lumosis: Out of curiosity, what will happen if the return value doesn't include aux hidden states?

tpu_commons/models/jax/llama3.py
py4: Fixed

tpu_commons/models/jax/model_loader.py
py4: That should not happen otherwise there will be an exception. Thanks for bringing this up. I need to change other models as well

tpu_commons/runner/speculative_decoding_manager.py
py4: Right tnx

PR #676, opened 2025-09-10
[Bugfix] fix TPU device name

yaochengji:
# Description

Fixed the bug when the name is "TPU v6e", kind[:-1] will be "TPU v6" not "TPU v".


bythew3i:
I have a fix in https://github.com/vllm-project/tpu_commons/pull/675

PR #702, opened 2025-09-16
Cast checkpoint to model dtype.

py4:
# Checklist

### DIFF
--- a/tpu_commons/models/jax/llama_eagle3.py
+++ b/tpu_commons/models/jax/llama_eagle3.py
@@ -301,6 +301,10 @@ class EagleLlama3ForCausalLM(nnx.Module):
             "d2t": "draft_id_to_target_id",
         }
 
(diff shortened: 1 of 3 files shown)


tpu_commons/models/jax/llama_eagle3.py
```diff
@@ -291,6 +291,11 @@ def load_weights(self, rng_key: jax.Array):
             "d2t": "draft_id_to_target_id",
         }
 
+        # Define keys to keep in original dtype (e.g., float32 for stability)
+        keep_original_dtype_keys = [
+            r".*draft_id_to_target_id.*",
```
Lumosis, line 305: What about other models?

py4: @Lumosis that's a good point. i don't know among all models that we supports, which ones have some weights that need to retain their original type. What if we run e2e tests to catch such issues?

PR #705, opened 2025-09-17
[JAX][KV Cache] Add initial support for static KV cache quantization

jrplatin:
# Description

In this PR, I add quantized KV cache / RPA kernel support to the JAX model path.  Please note the following:
* We would like to refactor the JAX modeling code so that changes would only need to be made in a few places rather than in all individual model files, but that is outside the scope of this PR
* Query (Q) quantization is currently disabled (i.e. we don't pass the `q_scale` to RPA) because of a minimal performance  gain as a result of VMEM register spilling
* KV scales can theoretically be loaded in a from a checkpoint (for [example](https://huggingface.co/RedHatAI/Phi-3.5-mini-instruct-FP8-KV)), but I will add that in a future PR -- right now they are set to `1.0` by default, but I have no observed any major degradation in performance (see below)

# Performance

| **Dataset** | **KV Quant** | **Model** | **Output Throughput (tokens/s)** | **% Gain Relative to Baseline** | **ROUGE**                                                                                         |   |
|-------------|--------------|-----------|-----------------------|---------------------------------|---------------------------------------------------------------------------------------------------|---|
| MLPerf      | N/A          | Qwen3-32B | 2976.00               | N/A                             | {'rouge1': 46.4217, 'rouge2': 22.8296, 'rougeL': 30.9906, 'rougeLsum': 43.7816, 'gen_num': 24576} |   |
| MLPerf      | fp8          | Qwen3-32B | 3388.25               | 13.8%                           | {'rouge1': 46.3846, 'rouge2': 22.83, 'rougeL': 30.9983, 'rougeLsum': 43.7322, 'gen_num': 24576}   |   |

# Tests

Add relevant unit testing + tested all models with and without quantized KV cache offline.

# Checklist

### DIFF
--- a/tpu_commons/attention/backends/pallas_torchax.py
+++ b/tpu_commons/attention/backends/pallas_torchax.py
@@ -109,7 +109,8 @@ class PallasAttentionBackendImpl(AttentionImpl):
         if self.kv_cache_quantized_dtype:
             key, value = utils.quantize_kv(key, value,
                                            self.kv_cache_quantized_dtype,
-                                           layer._k_scale, layer._v_scale)
+                                           layer._k_scale_float,
+                                           layer._v_scale_float)
             # TODO(kyuyeunk): Enable w8a8 when VREG spill issue is resolved.
             # q_scale = layer._q_scale_float
             k_scale = layer._k_scale_float
--- a/tests/experimental/test_llama3_jax_stashed.py
+++ b/tests/experimental/test_llama3_jax_stashed.py
@@ -50,6 +50,10 @@ class MockVllmConfig:
             }
         }
 
+        # NOTE (jacobplatin): we could add a quantized KV cache test, but
+        # we'll skip it for now.
+        self.cache_config = MagicMock(cache_dtype="auto")
+
 
 @pytest.fixture(scope="module")
 def mesh():
--- a/tests/models/jax/test_llama4.py
+++ b/tests/models/jax/test_llama4.py
@@ -54,6 +54,8 @@ class MockVllmConfig:
             }
         }
 
+        self.cache_config = MagicMock(cache_dtype="auto")
+
 
 @pytest.fixture(scope="module")
 def mesh():
--- a/tests/models/jax/test_llama_eagle3.py
+++ b/tests/models/jax/test_llama_eagle3.py
@@ -25,7 +25,7 @@ class MockSpeculativeConfig:
 
 class MockVllmConfig:
 
-    def __init__(self, model: str, draft_model: str):
+    def __init__(self, model: str, draft_model: str, kv_cache_dtype):
         self.model_config = ModelConfig(model)
         self.model_config.dtype = jnp.bfloat16
         self.load_config = MagicMock()
@@ -37,12 +37,14 @@ class MockVllmConfig:
             max_model_len=2048,
             skip_tokenizer_init=True,
             trust_remote_code=True)
+        self.cache_config = MagicMock(cache_dtype=kv_cache_dtype)
 
 
 @pytest.fixture
 def mock_vllm_config() -> MockVllmConfig:
     return MockVllmConfig(model="meta-llama/Meta-Llama-3-8B-Instruct",
-                          draft_model="yuhuili/EAGLE3-LLaMA3.1-Instruct-8B")
+                          draft_model="yuhuili/EAGLE3-LLaMA3.1-Instruct-8B",
+                          kv_cache_dtype="auto")
 
 
 @pytest.fixture(scope="module")
@@ -96,7 +98,12 @@ class TestEagleLlama3ForCausalLM:
         dtype = jnp.bfloat16
         rngs = nnx.Rngs(rng)
 
-        layer = Eagle3LlamaDecoderLayer(hf_config, dtype, rngs, mesh)
+        layer = Eagle3LlamaDecoderLayer(
+            hf_config,
+            dtype,
+            rngs,
+            mesh,
+            kv_cache_dtype=mock_vllm_config.cache_config.cache_dtype)
 
         # Check if projection layers are overridden with correct input size
         original_hidden_size = hf_config.hidden_size
@@ -110,6 +117,12 @@ class TestEagleLlama3ForCausalLM:
             0] == expected_input_size
         assert isinstance(layer.hidden_norm, nnx.RMSNorm)
 
+    @pytest.mark.parametrize("mock_vllm_config", [
+        MockVllmConfig("meta-llama/Meta-Llama-3-8B-Instruct",
+                       "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B", "auto"),
+        MockVllmConfig("meta-llama/Meta-Llama-3-8B-Instruct",
+                       "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B", "fp8"),
+    ])
     def test_forward_pass(self, mock_vllm_config: MockVllmConfig, rng: PRNGKey,
                           mesh: Mesh, mock_model_inputs):
         """Tests the forward pass of the EagleLlama3ForCausalLM model."""
@@ -126,7 +139,9 @@ class TestEagleLlama3ForCausalLM:
             head_size=hf_config.hidden_size // hf_config.num_attention_heads,
             mesh=mesh,
             layer_names=["layer"] * hf_config.num_hidden_layers,
-        )
+            cache_dtype=jnp.float8_e4m3fn
+            if mock_vllm_config.cache_config.cache_dtype == "fp8" else
+            jnp.bfloat16)
 
         _, output_hidden_states, aux_hidden_states = model(
             kv_caches, input_ids, hidden_states, attention_metadata)
--- a/tests/models/jax/test_model_loader.py
+++ b/tests/models/jax/test_model_loader.py
@@ -54,6 +54,7 @@ def vllm_config() -> MagicMock:
     mock_config.load_config = MagicMock()
     mock_config.load_config.download_dir = None
     mock_config.additional_config = {}
+    mock_config.cache_config = MagicMock(cache_dtype="auto")
     return mock_config
 
 
--- a/tests/test_quantization.py
+++ b/tests/test_quantization.py
@@ -152,7 +152,7 @@ class TestQwixQuantizeNnxModel(unittest.TestCase):
                 kv_cache_block_size=self.kv_cache_block_size,
                 kv_cache_num_kv_heads=self.kv_cache_num_kv_heads,
                 kv_cache_head_size=self.kv_cache_head_size,
-            )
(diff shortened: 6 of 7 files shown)


tpu_commons/attention/backends/pallas_torchax.py
```diff
@@ -129,9 +107,10 @@ def forward(
         query, key, value = jax_view(query), jax_view(key), jax_view(value)
         q_scale = k_scale = v_scale = None
         if self.kv_cache_quantized_dtype:
-            key, value = self.quantize_kv(layer, key, value)
-            # TODO(kyuyeunk): Enable w8a8 when VREG spill issue is resolved.
-            # q_scale = layer._q_scale_float
+            key, value = utils.quantize_kv(key, value,
+                                           self.kv_cache_quantized_dtype,
+                                           layer._k_scale, layer._v_scale)
+            q_scale = layer._q_scale_float
```
kyuyeunk: please revert the change in line 133/134

tpu_commons/models/jax/phi3.py
```diff
@@ -119,6 +127,13 @@ def __call__(
             k = apply_rope(k, md.input_positions, self.head_dim_original,
                            self.rope_theta, self.rope_scaling)
         # o: (T, N, H)
+        q_scale = k_scale = v_scale = None
+        if self.kv_cache_quantized_dtype:
```
kyuyeunk, line 131: why are these changes inside the model implementation? I feel like the much better approach is to modify attention class to accept kv cache config, and share the logic for kv cache quantization? https://github.com/vllm-project/tpu_commons/blob/ed76bdad79a26d12b51064669f511d34250c3149/tpu_commons/models/jax/common/attention/attention.py#L22-L56

tpu_commons/models/jax/phi3.py
```diff
@@ -56,7 +56,7 @@ def __call__(self, x: jax.Array) -> jax.Array:
 class Phi3Attention(nnx.Module):
```
kyuyeunk, line 56: Does every model define their own attention class? why? it should be much scalable to inherit from base attention class and tweak the code to write in a custom logic? if you have that, you don't need to repeat the same kv cache quantization code into every attention implementation https://github.com/vllm-project/tpu_commons/blob/ed76bdad79a26d12b51064669f511d34250c3149/tpu_commons/models/jax/common/attention/attention.py#L22-L56

tpu_commons/attention/backends/pallas_torchax.py
jrplatin: Oh good catch, sorry -- took the wrong merge conflict code

tpu_commons/models/jax/phi3.py
jrplatin: This is what I was getting at offline -- there isn't "one" attention class for the JAX models.  We effectively have one per model (Phi3, Llama3, Llama4, DeepSeek, Qwen3, etc.).  The class you linked is only used for Llama4 right now.

I do think it makes sense to move towards a single, configurable attention class, but that's not in scope for this PR, but happy to hear suggestions obviously!

tpu_commons/models/jax/phi3.py
jrplatin: Yep, see my comment above -- we aren't doing this (which I think is a mistake), and let me create a discussion in the engineering chat

tpu_commons/utils.py
```diff
@@ -198,3 +207,32 @@ def get_hash_fn_by_name(hash_fn_name: str) -> Callable[[Any], bytes]:
     if hash_fn_name == "builtin":
         return hash
     return utils.get_hash_fn_by_name(hash_fn_name)
+
+
+def quantize_kv(key: jax.Array, value: jax.Array,
```
BirdsOfAFthr, line 217: Is it possible to have one `quantize_kv` method to share between JAX and Torchax path?

jrplatin: Yep, it is now!

bythew3i:
Thanks Jacob. Attention and kv cache change LGTM.

is there any metrics to measure quality for the models? 

BTW, you may still need to override the KvCacheSpec in kv_cache_manager to fix wrong page_size_bytes.. I believe this is being resolved yet, @kyuyeunk can help confirm.

kyuyeunk:
> Thanks Jacob. Attention and kv cache change LGTM.
> 
> is there any metrics to measure quality for the models?
> 
> BTW, you may still need to override the KvCacheSpec in kv_cache_manager to fix wrong page_size_bytes.. I believe this is being resolved yet, @kyuyeunk can help confirm.

@bythew3i Correct. I'm currently working on creating `TPUKVCacheSpec` as we have discussed in the chat.

bythew3i:
> > Thanks Jacob. Attention and kv cache change LGTM.
> > is there any metrics to measure quality for the models?
> > BTW, you may still need to override the KvCacheSpec in kv_cache_manager to fix wrong page_size_bytes.. I believe this is being resolved yet, @kyuyeunk can help confirm.
> 
> @bythew3i Correct. I'm currently working on creating `TPUKVCacheSpec` as we have discussed in the chat.

Thanks @kyuyeunk

tpu_commons/utils.py
patemotter, line 217: Nit: I can see the logic for putting the `quantize_kv` function inside `utils.py`, but it might make more sense to split these into multiple files with more specific naming. It might be outside of the scope of this PR, but I fear having a giant `utils.py` file serving as a catch-all for all types of functions.

tpu_commons/utils.py
```diff
@@ -19,6 +20,14 @@
 TPU_HEAD_SIZE_ALIGNMENT = 128
 TPU_SECOND_LAST_MINOR = 8
 
+TPU_STR_DTYPE_TO_JAX_DTYPE = {
```
patemotter, line 26: Nit: I know this existed and was just moved, but can you add a small comment explaining the need for this?

tpu_commons/models/jax/qwen3.py
```diff
@@ -111,6 +119,14 @@ def __call__(
         # v: (T, K, H)
         v = self.v_proj(x)
         # o: (T, N, H)
+        q_scale = k_scale = v_scale = None
+        if self.kv_cache_quantized_dtype:
+            # TODO(kyuyeunk/jacobplatin): Enable w8a8 when VREG spill issue is resolved.
```
patemotter, line 124: Nit: add link to the corresponding bug (if it exists) for easier tracking

tpu_commons/models/jax/qwen3.py
```diff
@@ -89,6 +89,14 @@ def __init__(self, config: Qwen3Config, dtype: jnp.dtype, rng: nnx.Rngs,
             rngs=rng,
         )
 
+        self._q_scale = 1.0
+        self._k_scale = 1.0
+        self._v_scale = 1.0
+        self.kv_cache_quantized_dtype = None
+        if kv_cache_dtype != "auto":
+            self.kv_cache_quantized_dtype = utils.TPU_STR_DTYPE_TO_JAX_DTYPE.get(
+                kv_cache_dtype.lower().strip())
```
patemotter: Nit: This would be cleaner to have as a dedicated function in `utils.py` that doesn't require the caller to use `lower` and `strip`.

tpu_commons/models/jax/qwen2.py
```diff
@@ -155,7 +174,7 @@ def __call__(
 class Qwen2DecoderLayer(nnx.Module):
 
     def __init__(self, config: Qwen2Config, dtype: jnp.dtype, rng: nnx.Rngs,
-                 mesh: Mesh):
+                 mesh: Mesh, kv_cache_dtype: str):
```
patemotter, line 177: Nit: Not sure if there's a good way around this, but having `dtype` be of type `jnp.dtype` and then `kv_cache_dtype` as `str` seems a bit confusing. Is there a way to do any conversion at a higher-level and then pass them both in as `jnp.dtypes` or is that not feasible?

tpu_commons/models/jax/common/attention/deepseek_v3_attention.py
```diff
@@ -67,6 +69,10 @@ class MLA(nnx.Module):
 
     rngs: InitVar[nnx.Rngs]
 
+    _q_scale: float = 1
+    _k_scale: float = 1
+    _v_scale: float = 1
```
patemotter, line 74: Super-Nit: `= 1.0` for added clarity.

tpu_commons/models/jax/common/attention/attention.py
```diff
@@ -79,6 +85,11 @@ def __post_init__(self, rngs: nnx.Rngs):
                                               self.dtype,
                                               random_init=self.random_init)
 
+        self.kv_cache_quantized_dtype = None
```
patemotter: Nit: Would this make sense to assign as the default value as part of the class definition itself?

tpu_commons/utils.py
jrplatin: I think I'll keep it as is and we can split later, just since there isn't enough quantization-specific content IMO to justify a whole new file

tpu_commons/models/jax/qwen2.py
jrplatin: Synced offline -- will implement in a future PR and leave TODOs

tpu_commons/experimental/llama3_jax_stashed.py
```diff
@@ -97,6 +97,9 @@ def __init__(self,
                         rope_scaling={},
                         rngs=self.rng,
                         dtype=dtype,
+                        # TODO (jacobplatin): we should refactor this to pass a dtype (or config) directly
+                        kv_cache_dtype=self.vllm_config.cache_config.
+                        cache_dtype,
```
patemotter: Can we either move this up a line or indent the value on this line so it doesn't get mistaken as two separate parameters?

tpu_commons/models/jax/common/attention/deepseek_v3_attention.py
```diff
@@ -230,6 +241,15 @@ def __call__(self,
                                     (0, multiple_of_128 - self.qk_head_dim)))
             v_SNH = jnp.pad(v_SNH, ((0, 0), (0, 0),
                                     (0, multiple_of_128 - self.v_head_dim)))
+            q_scale = k_scale = v_scale = None
+            if self.kv_cache_quantized_dtype:
+                # TODO(kyuyeunk/jacobplatin): Enable w8a8 when VREG spill issue is resolved.
+                # q_scale = self._q_scale
```
patemotter, line 247: Should we just remove this line?

jrplatin: I personally would vote to keep it since it might be confusing to some why we aren't passing it in when the RPA kernel code supports it

PR #710, opened 2025-09-18
[Bug fix] Fix logging error on HEAD and sync with vLLM HEAD

wenxindongwork:
# Description

Currently there is an error when running tpu_commons due to JAX attempting to initialize the backend twice. The first attempt is within `__init__.py` (`detect_platform()`) and the second attempt is during model loading within `model.inspect_model_cls()` where `_run_in_subprocess()` is called. To work around this, we shouldn't call `jax.device()`, `jax.default_backend()` or any function that triggers JAX to initialize the backend in the `__init__.py` file.

Full error for context: https://paste.googleplex.com/4839904509034496#l=23 

This PR also fixes a param naming change error (guided_decoing -> structured_output) that sourced from vLLM upstream.  (Upstream PR: https://github.com/vllm-project/vllm/pull/22772)

# Tests

Manually tested on TPU vm, CPU, and Pathways

# Checklist

### DIFF
--- a/tpu_commons/__init__.py
+++ b/tpu_commons/__init__.py
@@ -31,7 +31,6 @@ if "proxy" in os.environ.get('JAX_PLATFORMS', '').lower():
 else:
     # Either running on TPU or CPU
     try:
-        logger.info("Running vLLM on TPU.")
         logger.info(f"TPU info: node_name={ti.get_node_name()} | "
                     f"tpu_type={ti.get_tpu_type()} | "
                     f"worker_id={ti.get_node_worker_id()} | "


tpu_commons/__init__.py
```diff
@@ -15,38 +15,29 @@
...
+        logger.info("Running vLLM on TPU.")
+        logger.info(f"TPU info: node_name={ti.get_node_name()} | "
+                    f"tpu_type={ti.get_tpu_type()} | "
+                    f"worker_id={ti.get_node_worker_id()} | "
+                    f"num_chips={ti.get_num_chips()} | "
+                    f"num_cores_per_chip={ti.get_num_cores_per_chip()}")
+    except Exception as e:
+        logger.error(
+            f"Error occurred while logging TPU info: {e}. Are you running on CPU?"
+        )
```
patemotter, line 42: So is CPU supported or not? That functionality seems to be removed. 

Also in the current state it will always log `Running vLLM on TPU.` even if you are running on CPU which seems like it could be a bit confusing when reading the log. If we don't want to support CPU then can we do a higher level check for this instead of a try/except block?

wenxindongwork: thanks Pate! 
I can remove the "running on TPU log" since the TPU info log may be self sufficient. 
I also think there can be a better way for detecting whether we are running on CPU or not. Currently tpu_commons errors out on CPU. 
One way is to use `jax.default_platform()` can check if it equals to `cpu` but this approach errors out due to the reasons mentioned in the PR description. One way to enforce not supporting CPU is to raise `ValueError` inside this Exception, so we give a clear message to the user that CPU is not supported. What do you think?

PR #711, opened 2025-09-18
Add multi-lora support on single chip in torchax path.

vanbasten23:
# Description

This PR adds multi-lora support on single chip in torchax+jax_runner path. To be specific, this PR

- extracts the punica_wrapper (LoRA metadata containing torch.Tensors and other python primitives, not torch.nn.Module) and pass it to the jitted function as parameter.
- since the punica_wrapper is not torch.nn.Module, we manually replace it instead of relying on torch.func.functional_call. 

# Tests

`VLLM_XLA_CHECK_RECOMPILATION=1   VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  pytest -s -vv tests/lora/test_lora.py -k test_multi_lora`

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ flax==0.11.1
 torchax==0.0.7
 qwix @ git+https://github.com/google/qwix@main
 pathwaysutils
+parameterized
--- a/tests/experimental/test_llama3_jax_stashed.py
+++ b/tests/experimental/test_llama3_jax_stashed.py
@@ -50,6 +50,10 @@ class MockVllmConfig:
             }
         }
 
+        # NOTE (jacobplatin): we could add a quantized KV cache test, but
+        # we'll skip it for now.
+        self.cache_config = MagicMock(cache_dtype="auto")
+
 
 @pytest.fixture(scope="module")
 def mesh():
--- a/tests/models/jax/common/attention/test_common_attention.py
+++ b/tests/models/jax/common/attention/test_common_attention.py
@@ -6,6 +6,7 @@ import jax.numpy as jnp
 import numpy as np
 from flax import nnx
 from jax.sharding import Mesh
+from parameterized import parameterized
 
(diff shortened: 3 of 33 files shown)


vanbasten23:
cc: @qihqi @lsy323 @kyuyeunk @hfan 

The current known issue is that running the test without precompiling takes 8 min, which is much longer than it should be (as a comparison, the same test runs 2 min in vLLM also without precompiling).

vanbasten23:
I've fixed the issue and the test run time drops from 7 min to 1.5 min.

tpu_commons/runner/lora_utils.py
```diff
@@ -75,3 +76,50 @@ def _move_to_tpu_replicated(x):
...
+                break
+        assert punica_wrapper is not None, "Should have been able to find a punica wrapper from the Lora wrapper."
+
+        # vars does not show inherited methods or class attributes but this is fine because we only care about instance attributes.
+        for k in vars(punica_wrapper):
+            v = getattr(punica_wrapper, k, None)
+            if k == "device":  # this is str and jax.jit doesn't like it and we don't need to replace it.
+                continue
+            metadata[k] = v
+        return jax_view(metadata)
```
qihqi, line 102: qq: so this line makes metadata into a dict of JAX arrays;
I don't see where it converts it back: i.e. there is no call to torch_view in replace_lora_metadata

vanbasten23: I convert it back right before we call the torch.func.functional_call, e.g. https://github.com/vllm-project/tpu_commons/pull/711/files#diff-fe4a959ef637c84fdc39b16e6fb34c3fb490a83e0b9a9cfb07f8249106992b2cR189

vanbasten23:
The CI is green: https://screenshot.googleplex.com/7d8TmbGiSgrm7Zk. The only failure is due to missing Torchvision and irrelevant to this PR.

Thanks for the review!

PR #713, opened 2025-09-19
extract docker build step

CienetStingLin:
# Description

In the run_in_docker.sh script, the same tpu_commons commit is used to build a Docker image, and this process is executed repeatedly across multiple agents. This approach consumes significant compute resources and can also lead to test inaccuracies (a low probability event, but more likely on Buildkite agents with fewer resources) due to different test steps potentially using different vllm commits. To resolve this, the Docker image build process has been moved to the top level. The image is now stored in GCP Artifact Registry and downloaded from there for testing in each test step.

This process now cleans up lingering Docker images more effectively. Using a unique tag, it removes only the images generated during this docker build process. The cleanup runs regardless of whether the build step succeeds or fails. In theory, no images should remain in buildkite agent after docker build step.

Taking the attached [Test](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4688) as an example, if there are enough agents to run all steps simultaneously, the overall **Build Time** will not decrease. However, the usage time for **each agent** will be reduced by approximately **8 minutes** (as the Docker image build takes about 8 mins). This prevents agents from being tied up for too long. If there are not enough agents and not all steps can run concurrently, this change can significantly reduce the **total CI Build time** (the amount saved depends on the number of currently free agents; fewer available agents result in greater time savings).

# Tests
Tested on the [Buildkite](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4688).

You can verify on the CI page that the deletion of the Docker image is executed regardless of the test result.

# Checklist


.buildkite/pipeline_jax.yml
```diff
@@ -140,6 +158,8 @@ steps:
      agents:
        queue: tpu_v6e_queue
      commands:
+       - echo "Using VLLM_COMMIT_HASH=$(buildkite-agent meta-data get 'VLLM_COMMIT_HASH')"
+       - echo "Using TPU_COMMONS_COMMIT_HASH=$(buildkite-agent meta-data get 'TPU_COMMONS_COMMIT_HASH')"
```
meiyeh123: could we move them to here:
   - label: "Build Docker Image"
     key: build_docker
     agents:
       queue: tpu_v6e_queue
     commands:
       - .buildkite/scripts/build_docker.sh
       - move to here

.buildkite/scripts/run_in_docker.sh
```diff
@@ -49,34 +49,21 @@ DOCKER_HF_HOME="/tmp/hf_home"
...
-        echo "Removing leftover containers using vllm-tpu image(s)..."
-        echo "$total_containers" | xargs -n1 | sort -u | xargs -r docker rm -f
-    fi
-
-    echo "Removing old vllm-tpu image(s)..."
-    docker rmi -f "$old_images"
-else
-    echo "No vllm-tpu images found to clean up."
+if [ -z "$IMAGE_TAG" ]; then
+    echo "Error：Can't get Image Tag"
```
meiyeh123: echo "Error：Can't get Image Tag"
 "Error： -> this is chinese '：' intead of English ':'

CienetStingLin: Thanks, will fix it

.buildkite/scripts/bootstrap.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 upload_pipeline() {
+    buildkite-agent pipeline upload .buildkite/build.yml
```
jcyang43: How about we execute `build_docker.sh` in `bootstrap.sh` before uploading any pipeline yml? 

This way, we don't have to worry if all yml files have included the code to wait for the build docker dependency

.buildkite/scripts/run_in_docker.sh
```diff
@@ -58,41 +89,26 @@ DOCKER_HF_HOME="/tmp/hf_home"
 
 # (TODO): Consider creating a remote registry to cache and share between agents.
 # Subsequent builds on the same host should be cached.
```
jcyang43: I think this TODO comment can be removed? This PR is basically trying to address this

.buildkite/scripts/run_in_docker.sh
```diff
@@ -58,41 +89,26 @@ DOCKER_HF_HOME="/tmp/hf_home"
...
-    fi
-
-    echo "Removing old vllm-tpu image(s)..."
-    docker rmi -f "$old_images"
-else
-    echo "No vllm-tpu images found to clean up."
+export LOCATION="asia-south1"
+gcloud auth configure-docker ${LOCATION}-docker.pkg.dev -q
+TAG_FILE_NAME="vllm_image_tag"
+buildkite-agent artifact download $TAG_FILE_NAME .
```
jcyang43, line 92: I think we should also remove this downloaded file during the cleanup. Or did you do this somewhere else?

.buildkite/scripts/bootstrap.sh
CienetStingLin: The step to build the Docker image has been moved to be executed within `bootstrap.sh`.

jcyang43:
LGTM. @vipannalla, do you mind taking a look when you get a chance? Thanks

CienetStingLin:
Completed by another PR, close this one

PR #721, opened 2025-09-20
[Torchax] Merge regular and lora weight sharding

kyuyeunk:
# Description

Merge regular and lora weight sharding logic for better readability.

# Tests

https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3354

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/sharding.py
+++ b/tpu_commons/models/vllm/sharding.py
@@ -1,16 +1,15 @@
-import functools
-from collections import OrderedDict
-
-import humanize
 import jax
 import torch
 import torchax
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torch.nn import Parameter
 from torch.utils import _pytree as pytree
-from torchax.interop import torch_view
+from torchax.interop import jax_view, torch_view
 from torchax.ops.mappings import t2j
-from vllm.config import VllmConfig
+from vllm.lora.layers import (MergedColumnParallelLinearWithLoRA,
+                              MergedQKVParallelLinearWithLoRA,
+                              RowParallelLinearWithLoRA)
+from vllm.lora.layers.base_linear import BaseLinearLayerWithLoRA
 from vllm.model_executor.layers.vocab_parallel_embedding import (
     ParallelLMHead, VocabParallelEmbedding)
 
@@ -21,60 +20,8 @@ P = PartitionSpec
 logger = init_logger(__name__)
 
 
-def shard_vocab_parallel_embedding(layer: torch.nn.Module, mesh: Mesh,
-                                   vllm_config: VllmConfig):
-    assert isinstance(layer, VocabParallelEmbedding)
-    w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)
-    layer.weight = w.apply_jax_(jax.device_put,
-                                NamedSharding(mesh, P('model', None)))
-    return layer
-
-
-def shard_lm_head(layer: torch.nn.Module, mesh: Mesh, vllm_config: VllmConfig):
-    # TODO(qihqi): currently this is not handling case of tie_word_weights=True.
-    # if that config is set, then we should not create new weights but reuse the
-    # weight from VocabParallelEmbedding
-    assert isinstance(layer, ParallelLMHead)
-    w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)
-    layer.weight = w.apply_jax_(jax.device_put,
-                                NamedSharding(mesh, P('model', None)))
-    if hasattr(layer, 'bias'):
-        bias = Parameter(torch_view(t2j(layer.bias)), requires_grad=False)
-        layer.bias = bias.apply_jax_(jax.device_put,
-                                     NamedSharding(mesh, P('model')))
-    return layer
-
-
-MODULE_TYPE_TO_WRAPPING_FUNC = {
-    VocabParallelEmbedding: shard_vocab_parallel_embedding,
-    ParallelLMHead: shard_lm_head,
-}
-
-
-def shard_parallel_layers_to_tpu(model: torch.nn.Module, mesh: Mesh,
-                                 vllm_config: VllmConfig) -> None:
-
-    def _shard_layer(module, name=None, parent=None):
-        for module_type, wrapping_func in MODULE_TYPE_TO_WRAPPING_FUNC.items():
-            if isinstance(module, module_type):
-                wrapped_module = wrapping_func(module, mesh, vllm_config)
-
-                assert parent is not None and name is not None, (
-                    "Top Level module is not expected to be wrapped.")
-                logger.debug("replace %s with %s", module, wrapped_module)
-                setattr(parent, name, wrapped_module)
-
-                module = wrapped_module
-                break
-
-        for child_name, child_module in list(module.named_children()):
-            _shard_layer(child_module, child_name, module)
-
-    _shard_layer(model)
-
-
-def shard_model_to_tpu(model: torch.nn.Module, mesh: Mesh,
-                       vllm_config: VllmConfig):
+def shard_model_to_tpu(model: torch.nn.Module,
+                       mesh: Mesh) -> dict[str, torchax.torch.Tensor]:
     """
     Shard the model weights and move them to TPU.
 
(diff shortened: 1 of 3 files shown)


tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -106,81 +105,54 @@ def load_weights(self):
...
-            donate_argnums=(1, ),  # donate kv_cache
-            compiler_options={
-                "xla_tpu_all_gather_collective_matmul_mode":
-                "post_spmd_conservative",
-                "xla_tpu_reduce_scatter_collective_matmul_mode":
-                "post_spmd_conservative"
-            },
-            static_argnames=('layer_name_to_kvcache_index', ))
+        @functools.partial(jax.jit,
+                           donate_argnames=("kv_caches", ),
```
vanbasten23: Kudos for changing `argnums` to `donate_argnames`!

tpu_commons/models/vllm/sharding.py
```diff
@@ -21,213 +20,123 @@
...
     assert isinstance(layer, MergedColumnParallelLinearWithLoRA)
-    shard_and_move_lora_to_tpu(layer, mesh)
-    return layer
+    shard_base_linear_lora(layer, mesh)
 
 
-def partition_qkv_parallel_linear_lora(layer: torch.nn.Module,
-                                       mesh: Mesh) -> torch.nn.Module:
-    from vllm.lora.layers import MergedQKVParallelLinearWithLoRA
+def shard_qkv_parallel_linear_lora(layer: torch.nn.Module, mesh: Mesh) -> None:
```
vanbasten23: hi @kyuyeunk , qq for the SPMD case, since we are reusing the linear layer from vLLM instead of creating our own, could you point to me how we current shard the qkv linear layer? Do we need special treatment for it?

tpu_commons/models/vllm/sharding.py
```diff
@@ -21,213 +20,123 @@
...
-    return layer
 
 
-def shard_lm_head(layer: torch.nn.Module, mesh: Mesh, vllm_config: VllmConfig):
+def shard_lm_head(layer: torch.nn.Module, mesh: Mesh):
     # TODO(qihqi): currently this is not handling case of tie_word_weights=True.
     # if that config is set, then we should not create new weights but reuse the
     # weight from VocabParallelEmbedding
     assert isinstance(layer, ParallelLMHead)
     w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)
```
vanbasten23: Do you need `with jax.default_device(jax.devices("cpu")[0]), torchax.default_env():` as before? The reason we had  it is we want to make the full shape of weight `w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)` on CPU and then the following line `layer.weight = w.apply_jax_(jax.device_put,` moves it to TPU. Without that line, you need to make sure `w = Parameter(torch_view(t2j(layer.weight)), requires_grad=False)` is still on CPU otherwise you may run into OOM

tpu_commons/models/vllm/sharding.py
```diff
@@ -21,213 +20,123 @@
...
+def shard_tensor_to_tpu_replicated(tensor, mesh) -> torchax.tensor.Tensor:
+    if isinstance(tensor, torch.Tensor) and not isinstance(
+            tensor, torchax.tensor.Tensor):
+        tensor = t2j(tensor, use_dlpack=False)
+        tensor = torch_view(tensor).apply_jax_(jax.device_put,
+                                               NamedSharding(mesh, P()))
+    return tensor
+
+
+def shard_model_to_tpu(model: torch.nn.Module,
```
vanbasten23: Somehow I didn't see `_register_lora_weights_as_param` after the change. Does the CI pass? We need that change because by default the lora weight are not torch.nn.Parameter. For vllm, it's not an issue. But for torchax, since it's functional and we need to extract the nn parameter/buffers, we need to register them as nn.Parameter.

tpu_commons/models/vllm/sharding.py
```diff
@@ -21,213 +20,123 @@
...
+        for module_type, sharding_func in MODULE_TYPE_TO_SHARDING_FUNC:
+            if isinstance(module, module_type):
+                logger.debug("shard %s with %s", path, sharding_func)
+                sharding_func(module, mesh)
+                break
 
 
-def shard_lora_weights_and_move_to_tpu(model: torch.nn.Module,
-                                       mesh: Mesh) -> None:
+def shard_tensor_to_tpu_replicated(tensor, mesh) -> torchax.tensor.Tensor:
```
vanbasten23: nit: suggest to make shard_module_to_tpu and shard_tensor_to_tpu_replicated private, e.g. _shard_module_to_tpu, _shard_tensor_to_tpu_replicated

vanbasten23:
Thanks Kyuyeun for the improvements. Left some comments.

tpu_commons/models/vllm/sharding.py
kyuyeunk: I must have deleted it by mistake. Thanks for pointing it out! Added it back!

tpu_commons/models/vllm/sharding.py
kyuyeunk: Right now, the sharding happens in this function. But I am considering whether it is better to only do a weight manipulation logic (like reshape) in `process_weights_after_loading` and consolidate all the sharding / device_put logic in this file. Any inputs would be appreciated @hfan @yaochengji @qihqi  : https://github.com/vllm-project/tpu_commons/blob/f327efed474bb0c37aa1d9396afff13f7782f456/tpu_commons/models/vllm/quantization/unquantized.py#L75-L100

tpu_commons/models/vllm/sharding.py
kyuyeunk: I've added line `(BaseLinearLayerWithLoRA, shard_base_linear_lora),` to the list `MODULE_TYPE_TO_SHARDING_FUNC`. In theory, that logic should also take care of creating `torch.nn.Parameter`. I'm waiting for CI results to confirm this but it's broken due to unrelated error & I'm waiting for it to be fixed.

kyuyeunk:
@vanbasten23 Verified that CI passes: https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3308

tpu_commons/models/vllm/sharding.py
```diff
@@ -85,149 +32,127 @@ def shard_model_to_tpu(model: torch.nn.Module, mesh: Mesh,
...
+# NOTE: Ordering is important as it calls first matched type of a given module
+MODULE_TYPE_TO_SHARDING_FUNC = [
+    # Shard embedding layers
+    (ParallelLMHead, _shard_lm_head),
+    (VocabParallelEmbedding, _shard_vocab_parallel_embedding),
+    # Shard LoRA layers
+    (MergedColumnParallelLinearWithLoRA, _shard_column_parallel_linear_lora),
+    (MergedQKVParallelLinearWithLoRA, _shard_qkv_parallel_linear_lora),
+    (RowParallelLinearWithLoRA, _shard_row_parallel_linear_lora),
+    (BaseLinearLayerWithLoRA, _shard_base_linear_lora),
```
vanbasten23, line 148: nit: BaseLinearLayerWithLoRA may never be executed because MergedColumnParallelLinearWithLoRA/MergedQKVParallelLinearWithLoRA/RowParallelLinearWithLoRA all inherit from BaseLinearLayerWithLoRA.

tpu_commons/models/vllm/sharding.py
```diff
@@ -85,149 +32,127 @@ def shard_model_to_tpu(model: torch.nn.Module, mesh: Mesh,
...
-        return torch_view(x).apply_jax_(jax.device_put,
-                                        NamedSharding(mesh, P()))
-
-    with jax.default_device(jax.devices("cpu")[0]), torchax.default_env():
-        shard_parallel_layers_to_tpu(model, mesh, vllm_config)
+        params, buffers = _extract_all_params_buffers(model)
 
         # For other weight tensors, repliate them on all the TPU chips.
-        params, buffers = extract_all_params_buffers(model)
+        params, buffers = pytree.tree_map_only(
```
vanbasten23, line 46: Do you know why we need this tree_map here? I thought before this line, all the weights (the linear layer defined in MODULE_TYPE_TO_SHARDING_FUNC and the linear layers we shard in `process_weights_after_loading `) are already on TPU.

tpu_commons/models/vllm/sharding.py
```diff
@@ -85,149 +32,127 @@ def shard_model_to_tpu(model: torch.nn.Module, mesh: Mesh,
...
-    ("MergedColumnParallelLinearWithLoRA",
-     partition_column_parallel_linear_lora),
-    ("MergedQKVParallelLinearWithLoRA", partition_qkv_parallel_linear_lora),
-    ("RowParallelLinearWithLoRA", partition_row_parallel_linear_lora),
-])
 
+# NOTE: Ordering is important as it calls first matched type of a given module
+MODULE_TYPE_TO_SHARDING_FUNC = [
+    # Shard embedding layers
+    (ParallelLMHead, _shard_lm_head),
```
vanbasten23, line 142: Do you remember the reason why we cannot use `process_weights_after_loading ` to shard ParallelLMHead and VocabParallelEmbedding like other linear layers?

tpu_commons/models/vllm/sharding.py
kyuyeunk: Yes, but I like to think of this as a safety net: There might be some random/minor layers that wasn't covered in `MODULE_TYPE_TO_SHARDING_FUNC` and will trigger an error if that isn't converted from torch tensor to tornchax tensor.

This logic ensures that never happens.

tpu_commons/models/vllm/sharding.py
kyuyeunk: I think that should be possible. Reason we haven't tried it before is that we are essentially piggybacking `process_weights_after_loading` that is invoked during quantization process. And because embedding is often unquantized, there wasn't a good reference to follow on how to utilize `process_weights_after_loading` to manipulate embedding weights.

tpu_commons/models/vllm/sharding.py
kyuyeunk: Thanks! I'll leave that as-is in case we want to use it as a fallback.

PR #729, opened 2025-09-23
[Kernel][RPAv3] Simplify Kernel Table Searching Logic

kyuyeunk:
# Description

Simplify kernel table searching logic for better code readability.

Also, print a warning message if key is not found in `TUNED_BLOCK_SIZES`.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
+++ b/tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
@@ -5,6 +5,9 @@ import jax.numpy as jnp
 from tpu_commons.kernels.ragged_paged_attention.v3.util import (
     align_to, get_device_name, get_dtype_packing, get_tpu_version,
     next_power_of_2)
+from tpu_commons.logger import init_logger
+
+logger = init_logger(__name__)
 
 # key
 #   - device_name
@@ -3728,10 +3731,21 @@ def get_tuned_block_sizes(
     max_num_tokens,
     pages_per_seq,
 ) -> tuple[int, int]:
-    """Look up for the best (num_kv_pages_per_blk, num_queries_per_blk) from auto-tuned table."""
+    """Search tuned values for (num_kv_pages_per_blk, num_queries_per_blk)."""
+
+    # Set default block sizes for each tpu_version.
     tpu_version = get_tpu_version()
     if tpu_version < 4:
         raise NotImplementedError('TPU version must be 4 or higher.')
+    match tpu_version:
+        case 4:
+            # TPUv4 has much smaller VMEM size so we pick fixed block sizes.
+            bkv_p, bq = (512 // page_size, 32)
+        case 7:
+            bkv_p, bq = (4096 // page_size, 32)
+        case _:
+            bkv_p, bq = (2048 // page_size, 32)
+
     keys = get_lookup_keys(
         page_size,
         q_dtype,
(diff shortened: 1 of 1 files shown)


kyuyeunk:
@bythew3i unrelated to this PR, but for quantized matmul kernel, I used `NamedTuple` as I think it gives a better readability - also less error-prone since linters can tell you if you are doing something wrong:  https://github.com/vllm-project/tpu_commons/blob/b2bbac105333a714549029a2c36d6dcb32e719cc/tpu_commons/kernels/quantized_matmul/tuned_block_sizes.py#L14-L26

jrplatin:
LGTM, thanks for this!

tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
```diff
@@ -3399,23 +3413,12 @@ def get_tuned_block_sizes(
...
-            if dtypes in tb:
-                tb = tb[dtypes]
-                if head_dims in tb:
-                    tb = tb[head_dims]
-                    if max_model_len in tb:
-                        bkv_p, bq = tb[max_model_len]
+    try:
+        bkv_p, bq = TUNED_BLOCK_SIZES[device][page_size][dtypes][head_dims][
+            max_model_len]
+    except KeyError:
```
bythew3i, line 3763: The reason I did not go for this in the first place is raising KeyError is slow compared to if-else

tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
```diff
@@ -3384,10 +3387,21 @@ def get_tuned_block_sizes(
...
     pages_per_seq,
 ) -> tuple[int, int]:
-    """Look up for the best (num_kv_pages_per_blk, num_queries_per_blk) from auto-tuned table."""
+    """Search tuned values for (num_kv_pages_per_blk, num_queries_per_blk)."""
+
+    # Set default block sizes for each tpu_version.
     tpu_version = get_tpu_version()
     if tpu_version < 4:
         raise NotImplementedError('TPU version must be 4 or higher.')
+    match tpu_version:
```
bythew3i, line 3740: Nice clean up!

bythew3i:
> @bythew3i unrelated to this PR, but for quantized matmul kernel, I used `NamedTuple` as I think it gives a better readability - also less error-prone since linters can tell you if you are doing something wrong:
> 
> https://github.com/vllm-project/tpu_commons/blob/b2bbac105333a714549029a2c36d6dcb32e719cc/tpu_commons/kernels/quantized_matmul/tuned_block_sizes.py#L14-L26

You will have many collisions when table gets too big and make looking up slow

bythew3i:
Can you please also update in g3?

tpu_commons/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
kyuyeunk: Huh, didn't knew about that. And you seem to be right: https://stackoverflow.com/questions/1835756/using-try-vs-if-in-python

But I was thinking few microseconds of overhead during compile time (and it's only invoked once per input shape thanks to jit) is worth it for better readability. Easier to catch future potential bugs w.r.t. tuning table like https://github.com/vllm-project/tpu_commons/pull/726. wdyt?

bythew3i: Yeah, LGTMed. Please update the G3 as well! Thanks!

kyuyeunk:
> Can you please also update in g3?

Sure! Will do that!

PR #731, opened 2025-09-23
Add docker build cache cleanup before new docker run

kwang3939:
# Description

Add docker build cache cleanup before new docker run

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/scripts/run_in_docker.sh
+++ b/.buildkite/scripts/run_in_docker.sh
@@ -74,6 +74,9 @@ else
     echo "No vllm-tpu images found to clean up."
 fi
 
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/run_in_docker.sh
```diff
@@ -74,6 +74,9 @@ else
     echo "No vllm-tpu images found to clean up."
 fi
 
+echo "Pruning old Docker build cache..."
+docker builder prune -f
```
QiliangCui, line 78: agressively cleaning up may cause other issues... like increasing the error chance when pulling docker image. But I think it is OK to do it and solve that issue when we meet it.

PR #734, opened 2025-09-24
[Misc] Fix logic that determines available memory

kyuyeunk:
# Description

According to [vllm/vllm/config/cache.py](https://github.com/vllm-project/vllm/blob/164299500be5ae1e8c71e06d4b90ac82d9a0b31b/vllm/config/cache.py#L42-L49), `gpu_memory_utilization` determines the ratio of total GPU memory (or TPU memory in our example) that vLLM can use. Meaning `model weight + kv cache = total device memory * utilization`

However, current logic is used to determine the ratio of free memory that vLLM can use for kv cache. Meaning `kv cache = (total device memory - model weight) * utilization`.

This PR fixes the issue.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_commons/worker/tpu_worker_jax.py
+++ b/tpu_commons/worker/tpu_worker_jax.py
@@ -143,14 +143,37 @@ class TPUWorker(AbstractTpuWorker):
                     f"rank={self.rank} | "
                     f"node_id={get_node_id()} | "
                     f"is_driver_worker={self.is_driver_worker} | "
-                    f"hbm={utils.hbm_usage_gb(self.devices)}Gb")
+                    f"hbm={utils.hbm_usage_gb(self.devices)}GiB")
 
     def determine_available_memory(self) -> int:
+        gpu_memory_utilization = self.cache_config.gpu_memory_utilization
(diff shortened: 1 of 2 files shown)


tpu_commons/worker/tpu_worker_jax.py
```diff
@@ -147,10 +147,20 @@ def init_device(self):
...
+            total_hbm_limit += limit
+        total_hbm_limit *= self.cache_config.gpu_memory_utilization
+        total_hbm_free = int(total_hbm_limit - total_hbm_used)
+
+        if total_hbm_free <= 0:
+            raise ValueError(f"{total_hbm_used=} exceeds {total_hbm_limit=} by"
+                             f"{total_hbm_free}. Please consider increasing "
+                             f"--gpu-memory-utilization from "
+                             f"{self.cache_config.gpu_memory_utilization} to "
+                             f"a larger value.")
```
patemotter: Nit: Should we log these values for informational purposes even when the value is > 0?

kyuyeunk: Done.

PR #739, opened 2025-09-24
[JAX] Adding fallback mechanism for models which are not supported in jax-native to torchax implementation in vllm.

aman2930:
# Description

Adding fallback mechanism for models which are not supported in jax-native to torchax implementation in vllm.
Without this change, if the model is not supported by jax-native then `vllm serve` fail to start with error "Model architectures <architecture_name> are not supported for now". 
With this change it fallbacks on the torch implementation in vllm.
# Tests
Running the below command is successful with this fix, which was failing earlier:
HF_TOKEN=<hf_token> vllm serve mistralai/Mistral-7B-Instruct-v0.2 --max-model-len=10240 --max-num-seqs=1024  --disable-log-requests --tensor-parallel-size 2 --max-num-batched-tokens 4096 --no-enable-prefix-caching

# Checklist

Signed-off-by: Aman Gupta <amangu@google.com>

### DIFF
--- a/.buildkite/scripts/run_in_docker.sh
+++ b/.buildkite/scripts/run_in_docker.sh
@@ -74,6 +74,9 @@ else
     echo "No vllm-tpu images found to clean up."
 fi
 
+echo "Pruning old Docker build cache..."
+docker builder prune -f
+
 echo "Cleanup complete."
 
 docker build --no-cache -f docker/Dockerfile -t "vllm-tpu:${BUILDKITE_COMMIT}" .
--- a/tests/e2e/benchmarking/mm_bench.sh
+++ b/tests/e2e/benchmarking/mm_bench.sh
@@ -90,7 +90,7 @@ checkThroughputAndRouge() {
 }
 
 echo "Spinning up the vLLM server..."
-(vllm serve "$model_name" --max-model-len "$max_model_len" --disable-log-requests --max-num-batched-tokens "$max_batched_tokens" 2>&1 | tee -a "$LOG_FILE") &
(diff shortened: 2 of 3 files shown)


tpu_commons/models/jax/model_loader.py
```diff
@@ -298,8 +299,29 @@ def get_model(
...
+        try:
+            # Try to load the flax model first
+            return get_flax_model(vllm_config, rng, mesh, is_draft_model)
+        except ValueError as e:
+            # Convert the error message to a string to check its contents
+            error_msg = str(e)
+
+            # Check if this is the specific "unsupported architecture" error.
+            # We check if it starts with the expected phrase and contains the key error message.
+            if error_msg.startswith("Model architectures ") and " are not supported for now" in error_msg:
```
hfan: Is it possible to change the exception type of "Model architectures ... are not supported for now" to some subclass of ValueError (or RuntimeError), then catch it by type?

Matching the error message tends to be brittle.

hfan:
Also, is it possible to add a test for it? Maybe in the `tests/models/jax/test_model_loader.py` file

tpu_commons/models/jax/model_loader.py
karan: +1. This seems quite brittle.

Wouldn't it just be easier to check that the model doesn't exist in `_MODEL_REGISTRY` or directly call `_get_model_architecture`?

aman2930: Done

PR #740, opened 2025-09-24
[JAX][Quantization] Add Qwix support for SparseMatul

jrplatin:
# Description

In this PR, I add manual Qwix weight quantization and activation quantization for `SparseMatmul` in `DeepSeekMoE`.  This is needed because `qwix.quantize_model` does not work nicely with the `shard_map` that we are using (which hides Qwix metadata).

# Tests

Added relevant unit testing + tested with and without `sparse_matmul` and `skip_quantization`.

# Checklist

### DIFF
--- a/tests/models/jax/common/moe/test_deepseek_moe.py
+++ b/tests/models/jax/common/moe/test_deepseek_moe.py
@@ -1,7 +1,4 @@
-import os
 import unittest
-
-os.environ['XLA_FLAGS'] = '--xla_force_host_platform_device_count=8'
 
 import jax
 import jax.numpy as jnp
--- a/tpu_commons/models/jax/deepseek_v3.py
+++ b/tpu_commons/models/jax/deepseek_v3.py
@@ -21,14 +21,7 @@ from tpu_commons.models.jax.common.layers import (DenseFFW, Embedder, LMhead,
                                                   RMSNorm)
 from tpu_commons.models.jax.common.moe.deepseek_moe import (DeepSeekV3Router,
                                                             SparseMoE)
-<<<<<<< HEAD
-<<<<<<< HEAD
 from tpu_commons.models.jax.common.moe.moe import MoE
-=======
->>>>>>> a1aaa359 (Implement the SparseMatmul for DeepSeek)
-=======
-from tpu_commons.models.jax.common.moe.moe import MoE
->>>>>>> 8cbe9352 (add unit test; add flag to support switching between dense/sparse matmul)
 from tpu_commons.models.jax.common.transformer_block import (
     SharedExpertsTransformerBlock, TransformerBlock)
 from tpu_commons.models.jax.utils.quantization.quantization_utils import \
@@ -589,15 +582,16 @@ class DeepSeekV3WeightLoader:
         # Convert weights from torch into numpy
         cast_type = model_weight.value.dtype
 
-
         torch_view_type = DTYPE_VIEW_MAP.get(jnp.dtype(cast_type))
 
         if torch_view_type:
             # Avoid unnecessary upcasting and mem copy by viewing the tensor's
             # raw data as integers before converting to a JAX array.
-            weight_np = jnp.array(weight.view(torch_view_type).numpy()).view(cast_type)
+            weight_np = jnp.array(
+                weight.view(torch_view_type).numpy()).view(cast_type)
         else:
-            raise ValueError(f"Unsupported dtype for tensor conversion: {cast_type}")
+            raise ValueError(
+                f"Unsupported dtype for tensor conversion: {cast_type}")
 
         if scale is not None:
             scale = scale.to(torch.float32).numpy().astype(self.scale_dtype)


tpu_commons/models/jax/common/moe/deepseek_moe.py
```diff
@@ -141,6 +144,8 @@ class SparseMoE(MoE):
     tile_size: tuple[int, int, int] = (128, 64, 128)
     use_megablox: bool = False
     mesh: jax.sharding.Mesh
+    # This should be set if and only if you have quantized your model (via Qwix)
+    quantized_dtype: Optional[jnp.dtype] = None
```
liudangyi, line 149: You shouldn't need any extra config: everything should be obtained from `qpl.get_current_rule`.

tpu_commons/models/jax/common/moe/deepseek_moe.py
```diff
@@ -583,12 +593,19 @@ def __call__(self, x_TD: Float):
...
-        return mapped_moe_fwd(
-            self,
-            x_TD,
-            router_weights_TX,
-            selected_experts_TX,
-            self.kernel_gating_EDF.value,
-            self.kernel_up_proj_EDF.value,
-            self.kernel_down_proj_EFD.value,
-        )
+        quantized_kernel_gating_EDF = manually_quantize_qwix_weight(
```
liudangyi:
```
kernel_gating_EDF = self.kernel_gating_EDF.value
kernel_up_proj_EDF = self.kernel_up_proj_EDF.value
kernel_down_proj_EFD = self.kernel_down_proj_EFD.value

rule = qpl.get_current_rule('ragged_dot')
if rule and rule.weight_qtype:
  weight_how = ptq.qarray.HowToQuantize(qtype=rule.weight_qtype, channelwise_axes=[2])
  if not isinstance(kernel_gating_EDF, WithAux):
    kernel_gating_EDF = ptq.create_quantized_param(kernel_gating_EDF, weight_how)
  if not isinstance(kernel_up_proj_EDF, WithAux):
    kernel_up_proj_EDF = ptq.create_quantized_param(kernel_up_proj_EDF, weight_how)
  if not isinstance(kernel_down_proj_EFD, WithAux):
    kernel_down_proj_EFD = ptq.create_quantized_param(kernel_down_proj_EFD, weight_how)
  kernel_gating_EDF = kernel_gating_EDF.array
  kernel_up_proj_EDF = kernel_up_proj_EDF.array
  kernel_down_proj_EFD = kernel_down_proj_EFD.array

return mapped_moe_fwd(self, x_TD, router_weights_TX,
                      selected_experts_TX, kernel_gating_EDF,
                      kernel_up_proj_EDF,
                      kernel_down_proj_EFD,
                      # pass quantize_act_as into _gmm and use this to quantize.
                      quantize_act_as=rule.act_qtype if rule else None)
```

tpu_commons/models/jax/common/moe/deepseek_moe.py
jrplatin: This isn't necessarily Qwix-related -- this is just to determine whether to go down the quantization route or not for SparseMatmul

tpu_commons/models/jax/utils/quantization/quantization_utils.py
```diff
@@ -507,3 +509,79 @@ def load_random_weights_into_qwix_abstract_model(rng: PRNGKey,
...
+
+    Returns:
+        A tuple of the scale dtype and quant dtype.
+    """
+    qwix_config = vllm_config.additional_config.get("quantization",
+                                                    {}).get("qwix", {})
+    scale_dtype = getattr(jnp, qwix_config.get("scale_dtype", "bfloat16"))
+    quant_dtype = None
+    # TODO (jacobplatin): this needs to be much more robust
+    for rule in qwix_config.get("rules", []):
```
liudangyi, line 583: A more idiomatic approach is to call `qwix._src.qconfig.get_current_rule` inside the model and use the rule returned to determine how to quantize weights and activations.

It's safe to call `get_current_rule` even when no quantization is enabled, in which case it just returns None. By using method, theoretically you shouldn't need `DeepSeekV3.scale_dtype`, `DeepSeekV3.quant_dtype`, or `SparseMoE.quantized_dtype` attribute.

tpu_commons/models/jax/utils/quantization/quantization_utils.py
```diff
@@ -507,3 +509,79 @@ def load_random_weights_into_qwix_abstract_model(rng: PRNGKey,
...
+                                       channelwise_axes=channelwise_axes,
+                                       tiled_axes=tiled_axes,
+                                       calibration_method=calibration_method)
+    # This is needed because we aren't passing `act_name` right now
+    assert not rule.act_static_scale, "Static scale not supported right now"
+
+    # channelwise_axes should be set to (a subset of) non-contraction axes. e.g.
+    # for ragged_dot [m, k] x [g, k, n], they are [0] and [0, 2]
+    # TODO (jacobplatin): add support for `act_name`
+    return ptq._quantize_act(inputs, lhs_how, rule, "")
```
liudangyi: Please update the Qwix pin to use the non-underscored version.

tpu_commons/models/jax/utils/quantization/quantization_utils.py
jrplatin: This code actually isn't related to the logic in this PR -- that code is used to determine what the scale/quant dtype should be for the model weights themselves

PR #746, opened 2025-09-26
[Bugfix] Fix error where expert parallelism isn't being enabled

kyuyeunk:
# Description

Fix a bug where option `--enable-expert-parallel` was not being passed correctly

# Tests

Ran following command and logged `FusedMoEConfig`

```
python vllm/examples/offline_inference/basic/generate.py \
    --model=Qwen/Qwen3-30B-A3B \
    --tensor_parallel_size=4 \
    --max_model_len=1024 \
    --enable-expert-parallel
```

Before

```
FusedMoEConfig(num_experts=128, experts_per_token=8, hidden_dim=2048, num_local_experts=128, moe_parallel_config=FusedMoEParallelConfig(tp_size=1, dp_size=1, ep_size=1, tp_rank=0, dp_rank=0, ep_rank=0, use_ep=False), in_dtype=torch.bfloat16, max_num_tokens=256, has_bias=False)
```

After

```
FusedMoEConfig(num_experts=128, experts_per_token=8, hidden_dim=2048, num_local_experts=128, moe_parallel_config=FusedMoEParallelConfig(tp_size=1, dp_size=1, ep_size=1, tp_rank=0, dp_rank=0, ep_rank=0, use_ep=True), in_dtype=torch.bfloat16, max_num_tokens=256, has_bias=False)
```

# Checklist

### DIFF
--- a/tpu_commons/models/vllm/quantization/common.py
+++ b/tpu_commons/models/vllm/quantization/common.py
@@ -2,6 +2,7 @@ import torchax
 from jax.sharding import Mesh, PartitionSpec
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
+from vllm.model_executor.layers.fused_moe.layer import FusedMoE, FusedMoEConfig
 # yapf: disable
 from vllm.model_executor.layers.linear import (ColumnParallelLinear,
                                                LinearBase,
@@ -95,3 +96,10 @@ class JaxCommonConfig:
     def get_linear_config(self, layer: LinearBase) -> JaxCommonLinearConfig:
         assert isinstance(layer, LinearBase)
         return JaxCommonLinearConfig(self.vllm_config, self.mesh, layer)
+
+    def get_moe_config(self, layer: FusedMoE) -> FusedMoEConfig:
+        assert isinstance(layer, FusedMoE)
+        moe_config = layer.moe_config
+        use_ep = self.vllm_config.parallel_config.enable_expert_parallel
+        moe_config.moe_parallel_config.use_ep = use_ep
+        return moe_config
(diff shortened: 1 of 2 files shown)


tpu_commons/models/vllm/quantization/common.py
```diff
@@ -95,3 +96,10 @@ def set_configs(cls, vllm_config: VllmConfig, mesh: Mesh):
     def get_linear_config(self, layer: LinearBase) -> JaxCommonLinearConfig:
         assert isinstance(layer, LinearBase)
         return JaxCommonLinearConfig(self.vllm_config, self.mesh, layer)
+
+    def get_moe_config(self, layer: FusedMoE) -> FusedMoEConfig:
+        assert isinstance(layer, FusedMoE)
+        moe_config = layer.moe_config
+        use_ep = self.vllm_config.parallel_config.enable_expert_parallel
```
wenxindongwork, line 103: just curious, why are we using `enable_expert_parallel` here instead of `additional_config.sharding.sharding_strategy.expert_parallelism` (similar to TP)?

kyuyeunk: I assume you are referring to this class? https://github.com/vllm-project/tpu_commons/blob/2fc1cf6946122d54af38c7945b870601b88b4fe1/tpu_commons/models/jax/common/sharding.py#L20-L38

Can you point me to where it is used? I have seen that class but I haven't seen it being referenced anywhere.

torchax path relies on vllm's parallel configuration: https://github.com/vllm-project/vllm/blob/392edee34a008af2453d936cb3cdbd97842984a7/vllm/config/parallel.py#L64

And used it like this: https://github.com/vllm-project/tpu_commons/blob/2fc1cf6946122d54af38c7945b870601b88b4fe1/tpu_commons/models/vllm/quantization/common.py#L50-L54

kyuyeunk: My bad, it is using it. But it's really weird. It seems like JAX path uses completely different way to set sharding strategy? Will need a discussion around this: https://github.com/vllm-project/tpu_commons/blob/2fc1cf6946122d54af38c7945b870601b88b4fe1/tpu_commons/runner/tpu_jax_runner.py#L150-L176

PR #753, opened 2025-09-26
[Pathways] Fix hbm limit calculation and remove device info logging to support AsyncEngine

wenxindongwork:
# Description

Previously we hardcoded the HBM limit to 32GB when running tpu_commons on Pathways. This is incorrect and this PR fixes this hbm limit calculation to account for the device type. 

Additionally this PR removes the Pathways device info logging during the `__init__.py` file because `jax.devices()` initiates a Pathways client. When using `vllm serve`, 2 processes will both run the `__init__.py` file and result in an error where Pathways fails to initialize the second JAX runtime. Thus, removing the log resolves this issue.  

# Tests

Manually tested, and will test with BuildKite. 

# Checklist

### DIFF
--- a/tpu_commons/utils.py
+++ b/tpu_commons/utils.py
@@ -85,11 +85,45 @@ def hbm_usage_bytes(devices: Any) -> List[Tuple[int, int]]:
     return usage
 
 
+def get_device_name(num_devices: int | None = None):
+    kind = jax.devices()[0].device_kind
+    if 'TPU' not in kind:
+        raise RuntimeError('Expected TPU devices')
+    suffix = ''
+    if kind.endswith(' lite'):
+        kind = kind[:-len(' lite')]
+        suffix = 'e'
+    if kind.endswith('p'):
+        kind = kind[:-1]
+        suffix = 'p'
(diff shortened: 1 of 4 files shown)


tpu_commons/utils.py
```diff
@@ -85,11 +85,45 @@ def hbm_usage_bytes(devices: Any) -> List[Tuple[int, int]]:
...
+
+    device_kind = get_device_name()
+    if device_kind == "TPU v5p" or device_kind == "TPU v5":
+        return 95 * GBYTES
+    elif device_kind == "TPU v5e":
+        return 16 * GBYTES
+    elif device_kind == "TPU v6e" or device_kind == "TPU v4":
+        return 32 * GBYTES
+    elif device_kind == "TPU v7":
+        return 192 * GBYTES
```
samos123, line 118: I think this should be 192 * GBYTES / 2 because each jax device (v7x core) has 1/2 of the total chip HBM.

@wenxindongwork

wenxindongwork: thanks for letting me know!

PR #756, opened 2025-09-27
[Disagg] Enable disagg initialization over certain slice shape

sixiang-google:
# Description

Also:
- Pathways does not allow multiple client accessing IFRT. Remove some jax.devices() call to prevent this.
- 

# Tests

unit tests will be added later

# Checklist

### DIFF
--- a/tests/core/test_core_tpu.py
+++ b/tests/core/test_core_tpu.py
@@ -10,9 +10,108 @@ from vllm.v1.request import Request
 
 from tpu_commons.core.adapters import (VllmConfigAdapter, VllmEngineAdapter,
                                        VllmRequestAdapter)
-from tpu_commons.core.core_tpu import DisaggEngineCoreProc, _DisaggOrchestrator
+from tpu_commons.core.core_tpu import (DisaggEngineCore, DisaggEngineCoreProc,
+                                       _DisaggOrchestrator)
 from tpu_commons.interfaces.config import IConfig
 from tpu_commons.interfaces.engine import IEngineCore
+
+
+class TestDisaggEngineCore(unittest.TestCase):
+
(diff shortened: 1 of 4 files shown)


tpu_commons/platforms/tpu_jax.py
```diff
@@ -70,7 +69,8 @@ def get_attn_backend_cls(cls, selected_backend: _Backend, head_size: int,
     def get_device_name(cls, device_id: int = 0) -> str:
         try:
             if envs.VLLM_TPU_USING_PATHWAYS:
-                return jax.local_devices()[0].device_kind
+                # Causes mutliprocess accessing IFRT when calling jax.devices()
+                return "TPU v6 lite"
```
mrjunwan-lang, line 73: nit: will it cause confusion that we hardcode the device_kind here? as we may upgrade to v7 in the future?

tpu_commons/core/disagg_utils.py
```diff
@@ -31,7 +31,7 @@ def _parse_slices(slices_str: str) -> Tuple[int, ...]:
             if len(dims) == 1:
                 slice_sizes.append(int(dims[0]))
             elif len(dims) == 2:
-                slice_sizes.append(int(dims[0]) * int(dims[1]))
+                slice_sizes.append((int(dims[0]), int(dims[1])))
```
mrjunwan-lang, line 34: Love this solution, it use 2d dimension to present the slice shape

tpu_commons/platforms/tpu_jax.py
sixiang-google: Yes. We need to find out a way to obtain or pass in the device type

PR #757, opened 2025-09-27
[KV Cache] Override Number of KV cache blocks

kyuyeunk:
# Description

Use [num_gpu_blocks_override](https://github.com/vllm-project/vllm/blob/cd87bfbf37f2300b7076b496366cd69048819777/vllm/config/cache.py#L59) to override number of KV cache size if padding is applied.

# Tests

Verified that running following config which requires KV cache padding works without error

```
vllm serve RedHatAI/Meta-Llama-3.1-70B-Instruct-quantized.w8a8 --tensor-parallel-size 8 --no-enable-prefix-caching --kv_cache_dtype=fp8
```

It returns following log message

```
(EngineCore_DP0 pid=699134) INFO 09-28 01:43:24 [tpu_worker_jax.py:262] KV cache page size calculated by vLLM (262144 Bytes) does not match with actual page size used by RPA kernel (524288 Bytes). Recalculating number of KV blocks using actual page size.
(EngineCore_DP0 pid=699134) INFO 09-28 01:43:24 [kv_cache_utils.py:805] Overriding num_gpu_blocks=8676 with num_gpu_blocks_override=4338
```

# Checklist

### DIFF
--- a/tpu_commons/worker/tpu_worker_jax.py
+++ b/tpu_commons/worker/tpu_worker_jax.py
@@ -15,6 +15,7 @@ from vllm.distributed.parallel_state import (ensure_model_parallel_initialized,
                                              init_distributed_environment)
 from vllm.lora.request import LoRARequest
 from vllm.tasks import SupportedTask
+from vllm.v1.core.kv_cache_utils import get_num_blocks, get_uniform_page_size
 from vllm.v1.core.sched.output import SchedulerOutput
 from vllm.v1.kv_cache_interface import KVCacheConfig, KVCacheSpec
 from vllm.v1.outputs import DraftTokenIds, ModelRunnerOutput
@@ -27,6 +28,7 @@ from tpu_commons.di.interfaces import HostInterface
 from tpu_commons.distributed.utils import (get_host_ip, get_kv_transfer_port,
                                            get_node_id)
 from tpu_commons.logger import init_logger
+from tpu_commons.runner.kv_cache import get_rpa_page_size_bytes
 from tpu_commons.runner.tpu_jax_runner import TPUModelRunner
 from tpu_commons.worker._temporary_vllm_compat import (
     adapt_kv_cache_config_if_needed, adapt_lora_request_if_needed,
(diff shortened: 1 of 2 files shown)


kyuyeunk:
> > We can get the real page_size_byte by:
> >
> > ``` 
> > from tpu_commons.runner.kv_cache import get_kv_cache_shape_with_mesh
> > 
> > page_size_byte = kv_dtype_byte * np.prod(get_kv_cache_shape_with_mesh(
> >   mesh=mesh,
> >   total_num_pages=1,
> >   ...
> > ))
> > 
> > cache_config.num_gpu_blocks_override = get_num_blocks(self.vllm_config, len(kv_cache_specs),
> >                                         available_memory, page_size_bytes)
> > ```
> Please consider the sugguestion I made below so that we do not need to assume RPA is using a fixed way to calculate kv_Cahce_shape. In case RPA changes, we should rely on RPA to recalculate kv_Cahce_shape without changing anything upstream like here.

@bythew3i thanks for the comment. I have modified the code as you have suggested. PTAL!

tpu_commons/worker/tpu_worker_jax.py
```diff
@@ -251,7 +253,26 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
+                f"Recalculating number of KV blocks using actual page size.")
+
+            available_memory = self.determine_available_memory()
+            num_blocks = get_num_blocks(self.vllm_config, len(kv_cache_specs),
+                                        available_memory, rpa_page_size_bytes)
+
+            cache_config = self.vllm_config.cache_config
+            cache_config.num_gpu_blocks_override = num_blocks
+
+        return kv_cache_specs
```
bythew3i, line 278: I think we still need to override page_size_byte in each layer of kv_cache_specs

kyuyeunk: As of now, considering a case where `page_size_bytes` is different for each `kv_cache_specs` isn't needed as vllm also makes assumption that all `page_size_bytes` will be the same:
- when calculating number of kv cache blocks, vllm calls `get_uniform_page_size`: [link](https://github.com/vllm-project/vllm/blob/0efd540dbc5405ada2f57f09d2a376aecad576dc/vllm/v1/core/kv_cache_utils.py#L1046-L1056)
- and `get_uniform_page_size` assumes all kv cache specs have same `page_size_bytes`: [link](https://github.com/vllm-project/vllm/blob/0efd540dbc5405ada2f57f09d2a376aecad576dc/vllm/v1/core/kv_cache_utils.py#L834-L836)

Also, while I definitely agree that making changes to [KVCacheSpec](https://github.com/vllm-project/vllm/blob/0efd540dbc5405ada2f57f09d2a376aecad576dc/vllm/v1/kv_cache_interface.py#L20-L36) is more proper fix, this is a stopgap solution to unblock ourselves in tight deadline while we figure out a proper fix to propose to vllm community. I've added a todo to reflect that.

bythew3i: Not sure if I like this fix but I can approve it. I just feel if the `page_size_bytes` remains wrong in AttentionSpec, it can be a footgun that come back to bite us one day.

Why do you think adding the change as we discussed to vllm is harder? 

```
@dataclass(frozen=True)
class AttentionSpec(KVCacheSpec):
    num_kv_heads: int
    head_size: int
    dtype: torch.dtype
    use_mla: bool
    page_size_bytes: int = None

    def __post_init__(self):
        if self.page_size_bytes is None:
          # For MLA we only store a single latent vector
          coef = 1 if self.use_mla else 2
          self.page_size_bytes = coef * self.block_size * self.num_kv_heads * self.head_size \
                * get_dtype_size(self.dtype)
```

We have @yaochengji can approve in vllm upstream.

kyuyeunk: Thanks. I think that's a fair criticism. I think I wanted to avoid making changes to upstream for some unknown reason?? I'll make a PR for vLLM upstream first, see how vllm community reacts, and if I hit some roadblocks, than I'll push this PR.

yaochengji: I think overriding KVCacheSpec to align with our needs might be a better way. I can help if  @kyuyeunk you want to make changes to the upstream.

I also have one dumb question: could we just implement and use our own KVCacheSpec here? https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/runner/kv_cache_manager.py#L11
like TPUFullAttentionSpec and TPUSlidingWindowSpec?

kyuyeunk: > could we just implement and use our own KVCacheSpec here?

That's not possible because vllm has a hard coded mapping between kv cache type and its manager: https://github.com/vllm-project/vllm/blob/14c1432789c9c1b66308481b2c37439d3ee6661a/vllm/v1/core/single_type_kv_cache_manager.py#L657-L663

kyuyeunk:
Submitting this change as we this blocks running qwen3 235b at all and we require the model to be runnable in main branch asap. will create a proper fix later that apply fix to vllm upstream.

PR #758, opened 2025-09-28
Avoid unnecessary upcasting and memory copy to reduce weight loading …

Lumosis:
Avoid unnecessary upcasting and memory copy to reduce weight loading time for Deepseek

# Description

After benchmarking and analysis, I found [this line](https://github.com/vllm-project/tpu_commons/blob/main/tpu_commons/models/jax/deepseek_v3.py#L540) introduces unnecessary upcasting and memory copy which significantly slow down the weight loading process.

With this fix, I was able to reduce the weight loading time on single-host for deepseek:

From 143 mins to 90 mins (Cold run)
From 63 mins to 3 mins (With cache in memory)

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_commons/models/jax/deepseek_v3.py
+++ b/tpu_commons/models/jax/deepseek_v3.py
@@ -30,6 +30,13 @@ from tpu_commons.models.jax.utils.weight_utils import (get_param,
 
 logger = init_logger(__name__)
 
+# A map from JAX dtype to the corresponding PyTorch integer dtype for raw memory viewing.
+DTYPE_VIEW_MAP = {
+    jnp.dtype(jnp.float8_e4m3fn): torch.uint8,
+    jnp.dtype(jnp.bfloat16): torch.uint16,
+    jnp.dtype(jnp.float32): torch.uint32,
+}
+
 
 @dataclass
 class DeepSeekV3(nnx.Module):
@@ -538,16 +545,15 @@ class DeepSeekV3WeightLoader:
         # Convert weights from torch into numpy
         cast_type = model_weight.value.dtype
 
-        if cast_type == jnp.float8_e4m3fn:
-            # Avoid unnecessary upcasting and mem copy.
-            weight_np = jnp.array(weight.view(
-                torch.uint8).numpy()).view(cast_type)
-        elif cast_type == jnp.bfloat16:
-            # Avoid unnecessary upcasting and mem copy.
-            weight_np = jnp.array(weight.view(torch.uint16).numpy()).view(
-                jnp.bfloat16)
+
+        torch_view_type = DTYPE_VIEW_MAP.get(jnp.dtype(cast_type))
+
+        if torch_view_type:
+            # Avoid unnecessary upcasting and mem copy by viewing the tensor's
+            # raw data as integers before converting to a JAX array.
+            weight_np = jnp.array(weight.view(torch_view_type).numpy()).view(cast_type)
         else:
-            weight_np = weight.to(torch.float32).numpy().astype(cast_type)
+            raise ValueError(f"Unsupported dtype for tensor conversion: {cast_type}")
 
         if scale is not None:
             scale = scale.to(torch.float32).numpy().astype(self.scale_dtype)


tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -537,7 +537,17 @@ def _load_individual_weight(self,
...
+        if cast_type == jnp.float8_e4m3fn:
+            # Avoid unnecessary upcasting and mem copy.
+            weight_np = jnp.array(weight.view(
+                torch.uint8).numpy()).view(cast_type)
+        elif cast_type == jnp.bfloat16:
+            # Avoid unnecessary upcasting and mem copy.
+            weight_np = jnp.array(weight.view(torch.uint16).numpy()).view(
+                jnp.bfloat16)
+        else:
+            weight_np = weight.to(torch.float32).numpy().astype(cast_type)
```
gpolovets1: Why not use weight.view for consistency?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -537,7 +537,17 @@ def _load_individual_weight(self,
...
+
+        if cast_type == jnp.float8_e4m3fn:
+            # Avoid unnecessary upcasting and mem copy.
+            weight_np = jnp.array(weight.view(
+                torch.uint8).numpy()).view(cast_type)
+        elif cast_type == jnp.bfloat16:
+            # Avoid unnecessary upcasting and mem copy.
+            weight_np = jnp.array(weight.view(torch.uint16).numpy()).view(
+                jnp.bfloat16)
+        else:
```
gpolovets1, line 555: Can we add a check for whether `cast_type == jnp.float32`? We may want to throw an error otherwise (e.g. a different unsupported fp8 format is requested).

jrplatin:
Can you update the PR description, o.w. LGTM!

tpu_commons/models/jax/deepseek_v3.py
Lumosis: Done

tpu_commons/models/jax/deepseek_v3.py
Lumosis: Done

Lumosis:
> Can you update the PR description, o.w. LGTM!

Done

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -537,7 +544,16 @@ def _load_individual_weight(self,
...
         cast_type = model_weight.value.dtype
-        weight_np = weight.to(torch.float32).numpy().astype(cast_type)
+
+
+        torch_view_type = DTYPE_VIEW_MAP.get(jnp.dtype(cast_type))
+
+        if torch_view_type:
+            # Avoid unnecessary upcasting and mem copy by viewing the tensor's
+            # raw data as integers before converting to a JAX array.
+            weight_np = jnp.array(weight.view(torch_view_type).numpy()).view(cast_type)
```
vanbasten23, line 554: Thanks for the improvement. It seems the impact of this line is huge `From 143 mins to 90 mins`. Just to confirm my understanding, by "unnecessary upcasting" you are referring to "weight.to(torch.float32)" and by "unnecessary memory copy" you are referring to "weight.to(torch.float32).numpy()". And the reason why this fix can greatly reduce weight loading time is that you cast `weight` to the correct dtype (`torch_view_type`). Is my understanding correct?

Lumosis: IIUC, in the original code, there are two memory copies. 1. weight.to(torch.float32) performs a numerical conversion, upcasting the data to float32 which introduces the first mem copy. 2. .astype(cast_type) performs a second numerical conversion which introduces the second mem copy. 

`.numpy()` just converts a pytorch tensor to a numpy array and is a zero-copy operation.

`And the reason why this fix can greatly reduce weight loading time is that you cast weight to the correct dtype (torch_view_type)` I didn't perform the `cast` operation. I just changed the interpretation of the tensor. Since numpy doesn't have native support for bf16 and bf8, we cannot convert a torch tensor (bf16 or bf8) to a numpy array directly. That's why we first interpret it to uint8, uint16 or uint32 without changing the underlying data, do the conversion, and then change the interpretation back to bf8 and bf16.

vanbasten23: Thanks a lot @Lumosis for the explanation!

PR #770, opened 2025-10-01
[DeepSeek] Add E2E functionality test for DeepSeek-R1 (12 layers only)

jrplatin:
# Description

This PR does the following:

* Changes the DeepSeek model to take in the number of hidden layers from the HF config (default is 61)
* Adds an option to the E2E MLPerf test to skip accuracy verification, use random weights, and disable the MLA cache (all of which are needed for the DeepSeek run, in which we won't check accuracy)
* Fixes the failing [CI](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3601/canvas) caused by [this](https://github.com/vllm-project/tpu_commons/commit/dc21dff9f097d31b5f56a5f14c8cf6b582862d61) PR

# Tests

Passing CL for relevant tests [here](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3617#01999e24-0a71-4524-bf8b-075035f53e48)

# Checklist

### DIFF
--- a/tests/e2e/benchmarking/mlperf.sh
+++ b/tests/e2e/benchmarking/mlperf.sh
@@ -251,7 +251,7 @@ for model_name in $model_list; do
         max_batched_tokens=1024
         if [ "$model_name" == "meta-llama/Llama-4-Scout-17B-16E-Instruct" ]; then
             current_serve_args+=(--hf-overrides '{"architectures": ["Llama4ForCausalLM"]}')
-        elif [ "$model_name" == "deepseek-ai/DeepSeek-V3" ]; then
+        elif [ "$model_name" == "deepseek-ai/DeepSeek-R1-0528" ]; then
             current_serve_args+=(--hf_overrides '{"num_hidden_layers": 12}')
         fi
     fi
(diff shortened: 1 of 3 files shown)


tests/e2e/benchmarking/mlperf.sh
```diff
@@ -248,6 +251,8 @@ for model_name in $model_list; do
         max_batched_tokens=1024
         if [ "$model_name" == "meta-llama/Llama-4-Scout-17B-16E-Instruct" ]; then
             current_serve_args+=(--hf-overrides '{"architectures": ["Llama4ForCausalLM"]}')
+        elif [ "$model_name" == "deepseek-ai/DeepSeek-V3" ]; then
```
gpolovets1: Can we change this to DeepSeek-R1?

tpu_commons/models/jax/deepseek_v3.py
```diff
@@ -54,7 +54,7 @@ def __init__(self,
         self.vllm_config = vllm_config
         self.rng = nnx.Rngs(rng)
 
-        num_layers: int = 61
+        num_layers: int = vllm_config.model_config.hf_config.num_hidden_layers
```
gpolovets1, line 58: Maybe just add a small comment that it is 61 by default?

jrplatin: Yep good point!

PR #774, opened 2025-10-01
[Kernel] Add flash attention kernel into tpu_commons

hfan:
# Description

Maintain an own copy of the flash attention kernel. It is copied from the OSS JAX repo with the backward pass related code removed.

# Tests

Qwen2.5 VL 7B model generates the same results: `SKIP_JAX_PRECOMPILE=1 VLLM_XLA_CHECK_RECOMPILATION=0 python examples/multi_modal_inference.py    --model=Qwen/Qwen2.5-VL-7B-Instruct      --tensor_parallel_size=1         --max_model_len=16384     --num-prompts=1 --gpu_memory_utilization=0.8`

# Checklist

### DIFF
--- a/tpu_commons/kernels/flash_attention/kernel.py
+++ b/tpu_commons/kernels/flash_attention/kernel.py
@@ -47,15 +47,6 @@ class BlockSizes:
     block_k: int
     block_b: int
 
-    block_q_major_dkv: int | None = None
-    block_k_major_dkv: int | None = None
-    block_k_dkv: int | None = None
-    block_q_dkv: int | None = None
-
-    block_k_major_dq: int | None = None
-    block_k_dq: int | None = None
-    block_q_dq: int | None = None
-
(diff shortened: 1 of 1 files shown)


vanbasten23:
OOC, I wonder why we couldn't import from JAX https://github.com/jax-ml/jax/blob/main/jax/experimental/pallas/ops/tpu/flash_attention.py since tpu_commons already takes dependency on JAX.

hfan:
> OOC, I wonder why we couldn't import from JAX https://github.com/jax-ml/jax/blob/main/jax/experimental/pallas/ops/tpu/flash_attention.py since tpu_commons already takes dependency on JAX.

Because we're introducing tuned block sizes for it and maybe add more optimizations, it is more process to add into the jax repo, much the same as the decision that we put RPA and quantized matmul kernels in tpu_commons

PR #775, opened 2025-10-01
add support matrices functionality to CI

jcyang43:
# Description

Set up CI such that support matrices can be generated automatically. Provide scripts for model owners and feature owners to add their models/features to CI. 

# Tests

- [poc success case](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3631#_)
- [poc failure case](https://buildkite.com/tpu-commons/tpu-commons-ci/builds/3632#_)

# Checklist

### DIFF
--- a/.buildkite/README.md
+++ b/.buildkite/README.md
@@ -11,3 +11,54 @@ steps:
       queue: tpu_v6e_queue
     command: "bash .buildkite/scripts/bootstrap.sh"
 ```
+
+# Support Matrices
+Besides continuous integration and continuous delivery, a major goal of our pipeline is to generate support matrices for our users for each release:
+- model support matrix (intended to replace [this](https://github.com/vllm-project/vllm/blob/f552d5e578077574276aa9d83139b91e1d5ae163/docs/models/hardware_supported_models/tpu.md) from the vllm upstream)
+- feature support matrix (intended to replace [this](https://github.com/vllm-project/vllm/blob/f552d5e578077574276aa9d83139b91e1d5ae163/docs/features/README.md) from the vllm upstream)
+
+To support this requirement, each model and feature will go through a series of stages of testing, and the test results will be used to generate the support matrices automatically.
+
+# Adding a new model to CI
+## Adding a TPU-optimized model
+TPU-optimized models are models we rewrite the model definition as opposed to using the model definition from the vLLM upstream. These models will go through benchmark on top of unit and integration (accuracy) tests. To add a TPU-optimized model to CI, model owners can use the prepared [add_model_to_ci.py](pipeline_generation/add_model_to_ci.py) script. The script will populate a buildkite yaml config file in the `.buildkite/models` directory; config files under this directory will be integrated to our pipeline automatically. The python script takes 2 arguments:
+- **model_name**: this is the **full name** of your model on Hugging Face. Please ensure to use the **full name** (ex: `meta-llama/Llama-3.1-8B` instead of `Llama-3.1-8B`) or else we won't be able to find your model.
+- **queue**: this is the queue you want to run on (ex: `tpu_v6e_queue`)
+
+```bash
+python add_model_to_ci.py --model-name <MODEL_NAME> --queue <QUEUE_NAME>
+```
+
+In the generated yml file, there are three TODOs that will need your input:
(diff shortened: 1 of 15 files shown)


.buildkite/README.md
```diff
@@ -11,3 +11,54 @@ steps:
...
+- **queue**: this is the queue you want to run on (ex: `tpu_v6e_queue`)
+
+```bash
+python add_model_to_ci.py --model-name <MODEL_NAME> --queue <QUEUE_NAME>
+```
+
+In the generated yml file, there are three TODOs that will need your input:
+1. The test command for the unit tests of your model
+2. The accuracy target for your model
+3. The performance benchmark target for your model
```
vipannalla, line 35: We need to pick a model and go through these steps to verify if we already have this info for current models (unless you have already done it with `Llama-3.1-8B` I see you have referred above).

.buildkite/README.md
```diff
@@ -11,3 +11,54 @@ steps:
       queue: tpu_v6e_queue
     command: "bash .buildkite/scripts/bootstrap.sh"
 ```
+
+# Support Matrices
+Besides continuous integration and continuous delivery, a major goal of our pipeline is to generate support matrices for our users for each release:
+- model support matrix (intended to replace [this](https://github.com/vllm-project/vllm/blob/f552d5e578077574276aa9d83139b91e1d5ae163/docs/models/hardware_supported_models/tpu.md) from the vllm upstream)
+- feature support matrix (intended to replace [this](https://github.com/vllm-project/vllm/blob/f552d5e578077574276aa9d83139b91e1d5ae163/docs/features/README.md) from the vllm upstream)
```
vipannalla, line 18: Unless Brittany already has a one, we need to come up with a list of all the features that tpu_commons will support at launch.

.buildkite/scripts/bootstrap.sh
```diff
@@ -10,17 +16,15 @@ if [ "$BUILDKITE_PULL_REQUEST" != "false" ]; then
   # If it's a PR, check for the specific label
   if [[ $PR_LABELS == *"ready"* ]]; then
     echo "Found 'ready' label on PR. Uploading main pipeline..."
-    buildkite-agent pipeline upload .buildkite/pipeline_jax.yml
-    # buildkite-agent pipeline upload .buildkite/pipeline_torch.yml
+    upload_pipeline
```
vipannalla, line 19: I thought the support matrix generation was for new builds which initially will be generated on a nightly basis.  This change will trigger the logic every PR.

vipannalla:
@QiliangCui, can you please take a look as well?

.buildkite/scripts/bootstrap.sh
jcyang43: So for regular PRs, support matrices aren't committed; they're just printed in the console for developers' information. The matrices are only committed to main when we're doing releases. 

I see. My previous understanding from Yarong is that PRs are supposed to go through the stages of testing (unit test, integration test, benchmark, etc). Do we only want to go through this for new builds now?

.buildkite/README.md
jcyang43: Yea there're existing unit tests under `tests/models/jax` for current models ([link](https://github.com/vllm-project/tpu_commons/tree/main/tests/models/jax))

.buildkite/scripts/bootstrap.sh
vipannalla: That makes sense, I missed the fact that support matrices only get committed for releases. 
Agree, we do need to do unittest/integration test/benchmark for each PR.

.buildkite/main.yml
```diff
@@ -0,0 +1,54 @@
+steps:
+   - label: "Upload Tests for Models & Features"
+     agents:
+       queue: tpu_v6e_queue
```
QiliangCui, line 5: nit: maybe we don't need tpu machines to do this. 

We can setup some normal non-tpu machines for work like this. 

If you agree, could you please create a TODO and a bug tracking it? (just keep a record not to forget. :) )

jcyang43: Thanks Qiliang! I agree. Added TODO comment and bug (b/449203039)

PR #779, opened 2025-10-01
[Llama4] Implement Config-Driven MoE/Dense Layer Interleaving for Maverick Support

sierraisland:
# Description

This PR refactors the Llama-4 model implementation (llama4.py) to support Llama-4-Maverick. This is a necessary change to correctly instantiate and load weights for Llama-4-Maverick, which alternate between MoE and Dense layers, a configuration not previously supported.

The original implementation of Llama4 hardcoded several architectural decisions, primarily assuming a structure where all layers were MoE layers (as is the case for Llama-4-Scout). This assumption is invalid for other variants, notably Llama-4-Maverick, which uses an alternating pattern where MoE and dense layers are interleaved (e.g., MoE layer, then Dense layer, then MoE, etc., typically controlled by an interleave_moe_layer_step of 2).

This change solves two main problems:

1. Architecture Mismatch: The model can now read interleave_moe_layer_step and other variant-specific parameters (like intermediate_size_mlp) from the configuration, allowing the single Llama4 class to correctly build the topology for both Scout and Maverick.

2. Weight Loading Error: The Llama4WeightLoader now dynamically checks the layer index to ensure weights are loaded into the correct module (MoE's shared experts or DenseFFW's kernels), preventing errors during initialization.

## Command for Llama-4-Maverick
HF_TOKEN="your-huggingface-token" VLLM_ENABLE_V1_MULTIPROCESSING=0 NEW_MODEL_DESIGN=True \python  tpu_commons/examples/offline_inference.py \
    --model=meta-llama/Llama-4-Maverick-17B-128E-Instruct \
    --tensor_parallel_size=8 \
    --max_model_len=1024 \
    --hf-overrides '{"architectures": ["Llama4ForCausalLM"]}'

## Implement Details

1. Architecture Configuration: Read all relevant parameters (num_layers, num_local_experts, intermediate_size_mlp, interleave_moe_layer_step, etc.) from the vllm_config.model_config.hf_config.text_config. Use the interleave_moe_layer_step to conditionally instantiate either an MoE module or a DenseFFW module inside the layer loop.

2. Weight Loader Logic: The Llama4WeightLoader's load_weights method is updated to dynamically determine the layer type based on the layer number and self.interleave_moe_layer_step before attempting to map and load weights, correctly handling the shared_expert prefix only when necessary.

## Related Bug
FIXES: b/445185679
FIXES: b/445221326
FIXES: b/446983033
FIXES: b/446984293
FIXES: b/447480960

# Tests

## Correctness check
Correctness check for first four layers as current resource only support up to four layers
## Unite Test
A new unit test was added to validate the private utility method _get_layer_num.

- Boundary Cases: Correctly extracts index 0 ("layers.0...").
- Layered Parameters: Correctly extracts common and high two-digit indices (e.g., 15 and 99). The inclusion of 99 tests the function's ability to handle two-digit numbers generally, ensuring future model variants with more layers are supported.
- Global Parameters: Correctly returns None for non-layered weights like final_norm ("...model.norm.weight") and the embedder ("...embed_tokens.weight").

# Checklist


tpu_commons/models/jax/llama4.py
```diff
@@ -54,27 +55,36 @@ def __init__(self,
...
-        rms_norm_eps = 1e-5
-        self.num_attention_heads = 40
-        self.num_key_value_heads = 8
-        self.head_dim = 128
+        self.intermediate_size_moe: int = getattr(text_config, "intermediate_size", 8192)
+        self.intermediate_size_mlp = getattr(text_config, "intermediate_size_mlp", 16384)
 
-        intermediate_size = 16384
+        # num_local_experts uses 128 experts specifically for Llama-4-Maverick-17B-128E-Instruct.
+        # while Llama-4-Scout-17B-16E-Instruct uses 16 experts.
```
wenxindongwork: is this comment suggesting that the num_local_experts applies only to Maverick?

tpu_commons/models/jax/llama4.py
```diff
@@ -54,27 +55,36 @@ def __init__(self,
...
+        self.num_local_experts: int = getattr(text_config, "num_local_experts", 16)
+        self.hidden_act: str = getattr(text_config, "hidden_act", "silu")
 
-        self.embedder = Embedder(vocab_size=vocab_size,
+        # Note: The original config files show "no_rope_layers", not "no_rope_layers_interval"
```
wenxindongwork: thanks for the note, I think we can remove this for succinctness

tpu_commons/models/jax/llama4.py
sierraisland: We get the num_local_experts directly from config file, which means it will be 16 when running Scout and be 128 when running Maverick.

tpu_commons/models/jax/llama4.py
sierraisland: Done

tpu_commons/models/jax/llama4.py
```diff
@@ -84,46 +101,51 @@ def __init__(self,
...
         self.layers = []
 
-        for i in range(num_layers):
+        for i in range(self.num_layers):
             # For Llama4-Scout, all layers are MoE layers.
             # This can be adjusted for other variants.
             is_moe_layer = (i + 1) % \
                             self.interleave_moe_layer_step == 0
-            use_attention_rope = (i + 1) % self.no_rope_layer_interval != 0
+            # use_attention_rope = (i + 1) % self.no_rope_layer_interval != 0
```
wenxindongwork: is this comment intended to be deleted?

tpu_commons/models/jax/llama4.py
```diff
@@ -84,46 +101,51 @@ def __init__(self,
...
+                activation_ffw_ted=('data', 'expert', None),
+                edf_sharding=('expert', None, 'model'),
+                efd_sharding=('expert', 'model', None),
+                random_init=force_random_weights
+            ) if is_moe_layer else DenseFFW(
+                dtype=dtype,
+                hidden_act=self.hidden_act,
+                hidden_size=self.hidden_size,
+                intermediate_size=self.
+                intermediate_size_mlp,  # Use the intermediate size for the MLP
```
wenxindongwork: we can remove this comment for succinctness as well

tpu_inference/models/jax/llama4.py
```diff
@@ -161,29 +183,30 @@ def __init__(self,
...
+                dtype=dtype,
+                hidden_act=self.hidden_act,
+                hidden_size=self.hidden_size,
+                intermediate_size=self.num_shared_experts *
+                self.intermediate_size_moe,
+                rngs=self.rng,
+                random_init=force_random_weights,
+                df_sharding=(None, 'model'),
+                fd_sharding=('model', None),
+                activation_ffw_td=('data', None)) if is_moe_layer else None
```
wenxindongwork, line 182: here why are we setting `activation_ffw_td=('data', None)) if is_moe_layer else None`?

tpu_commons/models/jax/llama4.py
```diff
@@ -247,6 +270,16 @@ def load_weights(self, rng: jax.Array, cache_dir: Optional[str] = None):
...
         weight_loader.load_weights(self)
 
+        logger.info("Starting final verification of model parameters...")
+        state = nnx.state(self)
+
+        # Let's inspect the names of the parameters in state[0]
+        logger.info(
+            "Inspecting names and types of parameters in the first state container:"
+        )
+        for i, (name, param) in enumerate(state.items()):
```
wenxindongwork: do we want to remove these debugging related logic?

tpu_commons/models/jax/llama4.py
```diff
@@ -284,17 +318,26 @@ def __init__(self, vllm_config: VllmConfig, hidden_size, attn_heads,
             download_dir=vllm_config.load_config.download_dir)
         self.is_verbose = getattr(vllm_config.additional_config, "is_verbose",
                                   False)
+
+        self.interleave_moe_layer_step = getattr(
```
wenxindongwork: can we get this attribute in the __init__ function similar to other model config attributes?

tpu_inference/models/jax/llama4.py
```diff
@@ -284,17 +318,26 @@ def __init__(self, vllm_config: VllmConfig, hidden_size, attn_heads,
             download_dir=vllm_config.load_config.download_dir)
         self.is_verbose = getattr(vllm_config.additional_config, "is_verbose",
                                   False)
+
+        self.interleave_moe_layer_step = getattr(
+            vllm_config.model_config.hf_config.text_config,
+            "interleave_moe_layer_step", 1)
+        self.expert_prefix = "shared_expert."
```
wenxindongwork, line 298: What if you have both "down_proj" and "shared_expert.down_proj" in `self._transpose_map` -- does it throw an error?

tpu_inference/models/jax/llama4.py
```diff
@@ -372,11 +421,7 @@ def _map_llama4_gate_up_proj(self, model_for_loading: nnx.Module,
             mapped_name = re.sub(r"layers\.\*", f"layers.{layer_num}",
                                  mapped_name)
             mapped_model_weight = get_param(model_params, mapped_name)
-            if not loaded_name.endswith(".bias"):
```
wenxindongwork, line 373: why did you have to remove this logic? (did it impact correctness?)

tpu_commons/models/jax/llama4.py
```diff
@@ -393,8 +438,17 @@ def _map_llama4_gate_up_proj(self, model_for_loading: nnx.Module,
 
     def load_weights(self, model_for_loading: nnx.Module):
         model_params = nnx.state(model_for_loading)
+
         with jax.default_device(jax.devices("cpu")[0]):
             for loaded_name, loaded_weight in self.names_and_weights_generator:
+                is_moe_layer = False
+                layer_num_match = re.search(r"layers\.(\d+)", loaded_name)
```
wenxindongwork: could you make a util function for getting layer number? 

and we should add unit tests for this function as well

tpu_inference/models/jax/llama4.py
```diff
@@ -256,6 +289,7 @@ def __call__(
     ) -> Tuple[List[KVCacheType], jax.Array, List[jax.Array]]:
         is_prefill = False
```
chandrasekhard2, line 266: This param is hardcoded, is this intentional?

tpu_commons/models/jax/llama4.py
chandrasekhard2, line 140: Instead of hardcoding these values, can we move into a configuration file?

tpu_inference/models/jax/llama4.py
```diff
@@ -284,17 +318,26 @@ def __init__(self, vllm_config: VllmConfig, hidden_size, attn_heads,
...
+        self.expert_prefix = "shared_expert."
         self._transpose_map = {
             "q_proj": (2, 0, 1),
             "k_proj": (2, 0, 1),
             "v_proj": (2, 0, 1),
             "router": (1, 0),
-            "shared_expert.down_proj": (1, 0),
-            "shared_expert.gate_proj": (1, 0),
-            "shared_expert.up_proj": (1, 0),
+            f"{self.expert_prefix}down_proj": (1, 0),
```
chandrasekhard2, line 304: Any reason for adding f-string instead of hardcoding like other layers?

tpu_inference/models/jax/llama4.py
sierraisland: It's a standard sharding choice for MOE layer. For MLP layer, it do not need transpose for shared_experts, which means should set as None.

tpu_commons/models/jax/llama4.py
sierraisland: This one is already under __init__ in Llama4WeightLoader.

tpu_inference/models/jax/llama4.py
sierraisland: I have tried, and it will get into error, that's why I choose these way to control when there should have shared_expert in self._transpose_map.

tpu_inference/models/jax/llama4.py
sierraisland: It impact the Maverick loading weights, with these lines, Maverick cannot load weights correctly and will get into error. Remove these lines, Scout still works correctly. So I feel we can safely remove them.

tpu_inference/models/jax/llama4.py
sierraisland: We need a control when self._transpose_map should have shared_expert. MOE layers should have it, while MLP layers should have it. Otherwise, Maverick will get into error when loading weights.

tpu_inference/models/jax/llama4.py
sierraisland: I have no idea on this one, I do not make any change on it.

tpu_commons/models/jax/llama4.py
sierraisland: Yes, we can do this in next PR maybe. How about keep the TODO here as this is not in our highest priority.

tpu_commons/models/jax/llama4.py
sierraisland: Done

tpu_commons/models/jax/llama4.py
sierraisland: Done

tpu_commons/models/jax/llama4.py
sierraisland: Done

tpu_commons/models/jax/llama4.py
sierraisland: Done

tpu_inference/models/jax/llama4.py
```diff
@@ -54,76 +55,85 @@ def __init__(self,
...
-        num_shared_experts = 1
-        rms_norm_eps = 1e-5
-        self.num_attention_heads = 40
-        self.num_key_value_heads = 8
-        self.head_dim = 128
+        self.intermediate_size_moe: int = getattr(text_config, "intermediate_size", 8192)
+        self.intermediate_size_mlp = getattr(text_config, "intermediate_size_mlp", 16384)
 
-        intermediate_size = 16384
+        # num_local_experts: uses 16 experts for Llama-4-Scout-17B-16E-Instruct and uses 128 experts Llama-4-Maverick-17B-128E-Instruct.
```
wenxindongwork, line 66: Instead of this comment on num_local_experts, can you make a comment saying that the default values are from Scout?

sierraisland: Done

tpu_inference/models/jax/llama4.py
wenxindongwork: sorry I missed the brackets. This looks good!

tpu_inference/models/jax/llama4.py
wenxindongwork: I believe this is intentional and deserves some clean up. George and Beinuo should have more context. For now, I think we can leave it here.

tpu_inference/models/jax/llama4.py
wenxindongwork: thanks for the verification!

PR #789, opened 2025-10-06
[Torchax] Optimize weight loading time

kyuyeunk:
# Description

Optimize weight loading time by placing weights into CPU first.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -114,8 +114,7 @@ class VllmModelWrapper:
 
         # Load the vLLM model and wrap it into a new model whose forward
         # function can calculate the hidden_state and logits.
-        with load_context:
-            # vllm_get_model will move the model weight to TPU.
+        with load_context, jax.default_device(jax.devices('cpu')[0]):
             vllm_model = vllm_get_model(vllm_config=vllm_config_for_load)
         lora_manager = None
         if vllm_config_for_load.lora_config is not None:


hfan:
Interesting, I thought it was default to CPU prevoiusly. Maybe after we update the torchax version, the default behavior changed.

Does it mean it goes to TPU device[0] first? If so, I'd expect it to get HBM OOM, but how come the weight load speed is faster?

tpu_commons/models/vllm/vllm_model_wrapper.py
```diff
@@ -114,7 +114,7 @@ def load_weights(self):
 
         # Load the vLLM model and wrap it into a new model whose forward
         # function can calculate the hidden_state and logits.
-        with load_context:
+        with load_context, jax.default_device(jax.devices('cpu')[0]):
             # vllm_get_model will move the model weight to TPU.
```
vanbasten23: nit: update this comment at line117?

kyuyeunk:
> Interesting, I thought it was default to CPU prevoiusly. Maybe after we update the torchax version, the default behavior changed.
> 
> Does it mean it goes to TPU device[0] first? If so, I'd expect it to get HBM OOM, but how come the weight load speed is faster?


You can see details in http://b/446339233

* Single CPU is connected to multiple TPU chips. So sending it to CPU 0 doesn't equal to sending it to TPU 0 and thus no OOM. 
* Assuming we have `N` bytes of weight with `M` number of chips, it is faster for CPU to send `N/M` bytes of data to each TPU chips in parallel vs. CPU sending `N` bytes of data to a single TPU first and that single TPU scatter the data across `M` TPUs.

tpu_commons/models/vllm/vllm_model_wrapper.py
kyuyeunk: Done.

hfan:
> > Interesting, I thought it was default to CPU prevoiusly. Maybe after we update the torchax version, the default behavior changed.
> > Does it mean it goes to TPU device[0] first? If so, I'd expect it to get HBM OOM, but how come the weight load speed is faster?
> 
> You can see details in http://b/446339233
> 
> * Single CPU is connected to multiple TPU chips. So sending it to CPU 0 doesn't equal to sending it to TPU 0 and thus no OOM.
> * Assuming we have `N` bytes of weight with `M` number of chips, it is faster for CPU to send `N/M` bytes of data to each TPU chips in parallel vs. CPU sending `N` bytes of data to a single TPU first and that single TPU scatter the data across `M` TPUs.

Got it. I meant how come it didn't HBM OOM before the change, since all the weights are going to the TPU device[0] first.

kyuyeunk:
> > > Interesting, I thought it was default to CPU prevoiusly. Maybe after we update the torchax version, the default behavior changed.
> > > Does it mean it goes to TPU device[0] first? If so, I'd expect it to get HBM OOM, but how come the weight load speed is faster?
> > 
> > 
> > You can see details in http://b/446339233
> > 
> > * Single CPU is connected to multiple TPU chips. So sending it to CPU 0 doesn't equal to sending it to TPU 0 and thus no OOM.
> > * Assuming we have `N` bytes of weight with `M` number of chips, it is faster for CPU to send `N/M` bytes of data to each TPU chips in parallel vs. CPU sending `N` bytes of data to a single TPU first and that single TPU scatter the data across `M` TPUs.
> 
> Got it. I meant how come it didn't HBM OOM before the change, since all the weights are going to the TPU device[0] first.

It's because we were processing weights sequentially.

Assume we have `X` number of `N` bytes of weight with `M` number of TPUs.

1. Move weight[0] to TPU[0]: TPU[0] HBM usage: `N`
2. Shard weight[0] from TPU[0] across `M` TPUs: TPU[0] HBM usage: `N/M`
3. Move weight[1] to TPU[0]: TPU[0] HBM usage: `N/M + N` 
4. Shard weight[1] from TPU[0] across `M` TPUs: TPU[0] HBM usage: `N/M + N/M`
5. repeat `X` times.

So the peak HBM usage of TPU[0] is `(X-1) * N/M + N`

PR #792, opened 2025-10-06
Jax model readme

gpolovets1:
# Description
Readme to instruct users of tpu_commons how to onboard new Jax NNX models. Instructions try to stay grounded on model implementation requirements, leaving broader discussion for inference features to our main README.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/jax/README.md
+++ b/tpu_inference/models/jax/README.md
@@ -39,11 +39,11 @@ tpu_inference
    │   ├── qwen3.py
    │   └── utils
 ```
-- registration of new Jax model types should be performed in tpu_inference/models/common/model_loader.py
-- new Jax model definitions should be added to tpu_inference/models/jax.
-- commonly used layers (e.g. embedding, feed-forward) can be imported from tpu_inference/layers/jax.
-- model-specific layer implementations should be added to tpu_inference/layers/<layer_type>/<model_type>_<layer_type>.py (e.g. attention/deepseek_v3_attention.py, moe/deepseek_v3_moe.py).
-- custom (Qwix) quantization configs (yaml files) should be stored in tpu_inference/models/jax/utils/quantization/configs.
+- registration of new Jax model types should be performed in `tpu_inference/models/common/model_loader.py`
+- new Jax model definitions should be added to `tpu_inference/models/jax`.
+- commonly used layers (e.g. embedding, feed-forward) can be imported from `tpu_inference/layers/jax`.
+- model-specific layer implementations should be added to `tpu_inference/layers/<layer_type>/<model_type>_<layer_type>`.py (e.g. `attention/deepseek_v3_attention.py`, `moe/deepseek_v3_moe.py`).
+- custom (Qwix) quantization configs (yaml files) should be stored in `tpu_inference/models/jax/utils/quantization/configs`.
 
 
 # Model Implementation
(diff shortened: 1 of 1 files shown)


tpu_commons/models/jax/README.md
```diff
@@ -0,0 +1,134 @@
...
+The ingredients for integrating a new model type consist of:
+- defining the model architecture and implementing any custom layers
+- implementing weight loading logic
+- (optional) adding quantization support
+- registering the new model into tpu_commons
+
+
+# Code Organization
+It is helpful to familiarize with the code organization before beginning model  development:
+(TODO: confirm this is the organization we will use)
```
jrplatin: Is this to be addressed now?

tpu_inference/models/jax/README.md
```diff
@@ -0,0 +1,134 @@
...
+# Quantization Support
+Many large LLMs like DeepSeek-V3 employ quantization to reduce hardware requirements and improve perf. tpu_commons supports quantized models via the Qwix library to load pre-quantized models and/or apply additional quantization settings to loaded model weights.
+
+Please note that you may need to update the list of supported quantization types on TPU [here](https://github.com/vllm-project/tpu_commons/pull/680/files).
+
+
+For the sake of demonstration, we will be referencing [deepseek_v3.py](https://github.com/vllm-project/tpu_commons/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu_commons/models/jax/deepseek_v3.py) for implementation details in this section.
+
+
+## Loading Prequantized Checkpoints and Applying Quantization Rules
```
jrplatin: Thanks for streamlining this! It may be good to link the full doc I made (and make it public), since I do think that level of depth is helpful/needed for the community

tpu_commons/models/jax/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+Conversely, if the checkpoint is not prequantized then no custom model loading code is needed and one should set `use_abstract_model: False` in the Qwix config.
+
+**Please note** that the Qwix quantization settings are the source of truth and will override the data types used for the loaded weights (even if prequantized weights were supplied).
+
+
+# Model Registration
+Once a new model type is implemented, it must be added to the model registry in [model_loader.py](https://github.com/vllm-project/tpu_commons/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu_commons/models/jax/model_loader.py#L27).\
+WARNING: per vLLM’s validation process, a model must be registered under a supported HuggingFace model name (see [here](TODO jacobplatin@) for more detail).
```
jrplatin: Seems like we are exploring this offline still

tpu_inference/models/jax/README.md
gpolovets1: Thanks Jacob! We may want to have that be a separate readme that we can place in the utils/quantization directory. I expect we will continue to update the documentation over time as well.

tpu_inference/models/jax/README.md
```diff
@@ -0,0 +1,136 @@
...
+Many large LLMs like DeepSeek-V3 employ quantization to reduce hardware requirements and improve performance. tpu-inference utilizes the [Qwix library](https://github.com/google/qwix) to support loading pre-quantized models and/or apply additional quantization settings to loaded model weights.
+
+Please note that you may need to update the list of supported quantization types on TPU [here](https://github.com/vllm-project/tpu-inference/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu-inference/platforms/tpu_jax.py#L45).
+
+
+For the sake of demonstration, we will be referencing [deepseek_v3.py](https://github.com/vllm-project/tpu-inference/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu-inference/models/jax/deepseek_v3.py) for implementation details in this section.
+
+
+## Loading Prequantized Checkpoints and Applying Quantization Rules
+To correctly load a prequantized checkpoint, the following steps need to be run:
```
kyuyeunk: There is ongoing work to refactor this so we don't have to do this at a model definition level: See `JaxEinsumLayer` in this doc: http://go/quantized-matmul-kernel-jax-ullm

So maybe add a disclaimer that this might be subject to change.

tpu_commons/models/jax/README.md
```diff
@@ -0,0 +1,136 @@
...
+   │   └── model_loader.py
+   └── jax  # Contains model files for each type of model family.
+   │   ├── deepseek_v3.py 
+   │   ├── llama3.py 
+   │   ├── qwen3.py 
+   │   └── utils 
+```
+- registration of new Jax model types should be performed in tpu-inference/models/common/model_loader.py
+- new Jax model definitions should be added to tpu-inference/models/jax.
+- commonly used layers (e.g. embedding, feed-forward) can be imported from tpu-inference/layers/common/.
```
kyuyeunk: Idea of `layers/common` isn't just commonly used layer, but its implementation is framework agnostic that it can be shared with jax and torch layer: which essentially means it's purely "stateless functional code" that has no dependency with `flax.nnx.Module` or `torch.nn.Module`.

So better way to frame it is "commonly used **logic**".

tpu_inference/models/jax/README.md
```diff
@@ -0,0 +1,136 @@
...
+- Defining a mapping of tensor transformations to apply on the loaded parameters. (these transformations can include transposing or reshaping loaded tensors).
+- Performing model-specific loading logic (e.g. splitting a loaded weight tensor and loading into multiple parameters).
+- (optional) Support for loading pre-quantized models.
+
+
+Please refer to [deepseek_v3.py](https://github.com/vllm-project/tpu-inference/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu-inference/models/jax/deepseek_v3.py#L355) or [llama4.py](https://github.com/vllm-project/tpu-inference/blob/ecfd349b7b1ce976fad8ff9cc0d60706d39dabb9/tpu-inference/models/jax/llama4.py#L276) for some examples on how to implement weight loading.
+
+
+# Quantization Support
+Many large LLMs like DeepSeek-V3 employ quantization to reduce hardware requirements and improve performance. tpu-inference utilizes the [Qwix library](https://github.com/google/qwix) to support loading pre-quantized models and/or apply additional quantization settings to loaded model weights.
```
kyuyeunk: External developers are more familiar with workflow of using offline quantization to convert checkpoints into list of popular quantization tools (like [llm-compressor](https://github.com/vllm-project/llm-compressor) or [llm-awq](https://github.com/mit-han-lab/llm-awq)) and not familiar with Qwix.

So better way to frame it is

"To extend model to support quantized checkpoints, do x. tpu-inference also uses Qwix to support online quantization to allow quantizing model even without a quantized checkpoint."

tpu_inference/models/jax/README.md
gpolovets1: I think we should just update the documentation when we have a better solution? This is meant to help folks develop their models with the code that is currently available.

tpu_inference/models/jax/README.md
gpolovets1: In the current code, Qwix needs to be used to also load the prequantized models. So it is used for both online quantization and loading quantized checkpoints (which could have been generated by the popular tools you mentioned).
However, if it would be helpful I can mention that we support loading quantized checkpoints that were generated with popular tools, as long as they are in safetensor format.

tpu_commons/models/jax/README.md
gpolovets1: Good catch, this needed to be updated after the folder restructuring. I have changed this to refer to `tpu_inference/layers/jax/` instead of `tpu_inference/layers/common`

tpu_inference/models/jax/README.md
kyuyeunk: > However, if it would be helpful I can mention that we support loading quantized checkpoints that were generated with popular tools, as long as they are in safetensor format. 

Yes, that's a good idea. For end-user who are using pre-quantized models, qwix is just an implementation detail. For them, what's more important is whether quantization format of their choice is supported or not.

But if I understand correctly, is this true? In vLLM upstream, the framework handles all the quantized weight loading & necessary post processing for target device. But in JAX path, that isn't currently implemented and the model developers has to implement pre-quantized model & weight support into model implementation right?

tpu_commons/models/jax/README.md
kyuyeunk: nit: please update this line as well. Probably something like `Functionalities shared between torchax and jax implementations.`?: https://github.com/vllm-project/tpu-inference/pull/792/files#diff-789e11213737f8cb393c169af3722db31d3d687309b385e8abce2c46cfc4818aR30

tpu_inference/models/jax/README.md
gpolovets1: Yes that's correct, at the moment Jax path needs to implement pre-quantized model weight loading support  in order to store and perform any necessary transforms on the scales before loading them into the model definition. Jacob is planning on simplifying the process in the future.

PR #795, opened 2025-10-07
feat(ci): Record vLLM and tpu-inference commit hashes

meiyeh123:
# Description

This commit adds steps to the CI pipeline to automatically log the commit hashes for the `vllm` and `tpu-inference` submodules upon successful completion of the nightly steps.
This enables tracking the exact versions used in each verified build, improving reproducibility and debugging.

# Tests

Since only the main branch allows pushing the commit, I couldn't test the push action on my branch.
Therefore, I tested the pipeline logic and replaced most of the procedures with a simple echo command.

[BuildKite failed procedure](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5561/steps/canvas)
[BuildKite succeeded procedure](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5559/steps/canvas)

# Checklist

### DIFF
--- a/.buildkite/main.yml
+++ b/.buildkite/main.yml
@@ -27,28 +27,9 @@ steps:
      commands:
        - bash .buildkite/scripts/generate_support_matrices.sh
 
-#   - label: "Notify test results"
-##     if: build.env("NIGHTLY") == "1"
-#     key: notify_test_results
-#     depends_on: generate_support_matrices
-#     agents:
-#       queue: cpu
-#     commands:
-#       - bash .buildkite/scripts/notify_test_results.sh
-
-#   - label: "Record verified commit hashes"
-#     key: record_verified_commit_hashes
-#     depends_on: notify_test_results
-#     branches: "main"
-#     agents:
-#       queue: cpu
-#     commands:
-#       # TODO : record verified vllm-tpu_inference commit hash pair if test successful (to be handled in separate PR)
-#       - echo "committing verified commit hashes"
-#
-   - label: "Commit support matrices"
+   - label: "Commit support matrices if is nightly build or release tag specified"
      <<: *run-on-nightly-or-tag
-     key: commit_nightly_support_matrices
+     key: commit_support_matrices
      depends_on: generate_support_matrices
      agents:
        queue: cpu
@@ -56,3 +37,25 @@ steps:
       - GITHUB_PAT
      command:
        - bash .buildkite/scripts/commit_support_matrices.sh
+
+   - label: "Notify test results"
+     <<: *run-on-nightly-or-tag
+     key: notify_test_results
+     depends_on: commit_support_matrices
+     agents:
+       queue: cpu
+     commands:
+       - bash .buildkite/scripts/notify_test_results.sh
+
+   - label: "Record verified commit hashes"
+     if: build.env("NIGHTLY") == "1"
+     key: record_verified_commit_hashes
+     depends_on:
+       - notify_test_results
+       - tpu_test_notification
+     agents:
+       queue: cpu
+     secrets:
+      - GITHUB_PAT
+     command:
+       - bash .buildkite/scripts/commit_verified_commit_hashes.sh
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -265,6 +265,7 @@ steps:
   # NOTIFICATION STEP
   # -----------------------------------------------------------------
    - label: "TPU Test Notification"
+     key: tpu_test_notification
      depends_on:
        - test_0
        - test_1
--- a/.buildkite/scripts/commit_verified_commit_hashes.sh
+++ b/.buildkite/scripts/commit_verified_commit_hashes.sh
@@ -5,17 +5,8 @@ set -e
 REPO_URL="https://github.com/vllm-project/tpu-inference.git"
 TARGET_BRANCH="main"
 
-# Conditional Configuration for Release vs. Nightly
-if [ "${NIGHTLY}" = "1" ]; then
-  # Set path and commit message for nightly builds.
-  ARTIFACT_DOWNLOAD_PATH="support_matrices/nightly"
-  COMMIT_MESSAGE="[skip ci] Update nightly support matrices"
-else
-  # Set path and commit message for release tag builds.
-  COMMIT_TAG="${BUILDKITE_TAG:-unknown-tag}"
-  ARTIFACT_DOWNLOAD_PATH="support_matrices"
-  COMMIT_MESSAGE="[skip ci] Update support matrices for ${COMMIT_TAG}"
-fi
+COMMIT_MESSAGE="Update verified commit hashes"
+
 # Construct the repository URL with the access token for authentication.
 AUTHENTICATED_REPO_URL="https://x-access-token:${GITHUB_PAT}@${REPO_URL#https://}"
 
@@ -34,12 +25,24 @@ git fetch origin "${TARGET_BRANCH}"
 git checkout "${TARGET_BRANCH}"
 git reset --hard origin/"${TARGET_BRANCH}"
 
-echo "--- Downloading CSV artifacts"
-mkdir -p "${ARTIFACT_DOWNLOAD_PATH}"
-buildkite-agent artifact download "*.csv" "${ARTIFACT_DOWNLOAD_PATH}/" --flat
+VLLM_COMMIT_HASH=$(buildkite-agent meta-data get "VLLM_COMMIT_HASH" --default "")
 
-echo "--- Staging downloaded artifacts"
-git add "${ARTIFACT_DOWNLOAD_PATH}"/*.csv
+if [ -z "${VLLM_COMMIT_HASH}" ]; then
+    echo "VLLM_COMMIT_HASH not found in buildkite meta-data"
+    exit 1
+fi
+
+if [ -z "${BUILDKITE_COMMIT:-}" ]; then
+    echo "BUILDKITE_COMMIT not found"
+    exit 1
+fi
+
+if [ ! -f verified_commit_hashes.csv ]; then
+    echo "timestamp,vllm_commit_hash,tpu_inference_commit_hash" > verified_commit_hashes.csv
+fi
+echo "$(date '+%Y-%m-%d %H:%M:%S'),${VLLM_COMMIT_HASH},${BUILDKITE_COMMIT}" >> verified_commit_hashes.csv
+
+git add verified_commit_hashes.csv
 
 # --- Check for changes before committing ---
 if git diff --quiet --cached; then
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,7 @@
 ARG NIGHTLY_DATE="20250714"
 ARG BASE_IMAGE="us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.12_tpuvm_$NIGHTLY_DATE"
+# The latest main will be used if arg unspecified
+ARG VLLM_COMMIT_HASH=""
 
 FROM $BASE_IMAGE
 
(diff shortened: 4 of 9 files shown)


.buildkite/main.yml
```diff
@@ -1,6 +1,18 @@
 steps:
+   - label: "Get current commit hashes"
```
jcyang43: Why not get the commit hashes at the start of the CI run (i.e. in bootstrap.sh)? This way, the information is available to all pipelines (ex: pipeline_jax.yml, main.yml & whatever the project adds in the future) and all the jobs within the same run can be run using the same commits.

.buildkite/main.yml
```diff
@@ -19,32 +31,45 @@ steps:
...
+       - |
+         # Conditional Configuration for Release vs. Nightly
+         if [ "$${NIGHTLY}" = "1" ]; then
+           # Set path and commit message for nightly builds.
+           export COMMIT_MESSAGE="Update nightly support matrices"
+         else
+           # Set path and commit message for release tag builds.
+           COMMIT_TAG="$${BUILDKITE_TAG:-unknown-tag}"
+           export COMMIT_MESSAGE="Update support matrices for $${COMMIT_TAG}"
+         fi
```
jcyang43: I think these can be done in the shell script for improved readability

.buildkite/scripts/commit_push.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -e
+
+# Please do `chmod +x <path/to/pre_commit_script.sh>`
+if test ! -x "$1"; then
+    echo "ERROR: The first argument must be a path to an executable pre-commit script."
+    echo "Usage: ./commit_push.sh <path/to/pre_commit_script.sh>"
+    exit 1
+fi
```
jcyang43: The `pre_commit_script.sh` naming and the usage are a little confusing. 

"pre_commit" naturally gets associated to the [pre-commit hook](https://github.com/vllm-project/tpu-inference/blob/main/.github/workflows/pre-commit.yml), but that's not what you're doing here 

https://pre-commit.com/

.buildkite/main.yml
```diff
@@ -19,32 +31,45 @@ steps:
...
      command:
-       - bash .buildkite/scripts/commit_support_matrices.sh
+       - |
+         # Conditional Configuration for Release vs. Nightly
+         if [ "$${NIGHTLY}" = "1" ]; then
+           # Set path and commit message for nightly builds.
+           export COMMIT_MESSAGE="Update nightly support matrices"
+         else
+           # Set path and commit message for release tag builds.
+           COMMIT_TAG="$${BUILDKITE_TAG:-unknown-tag}"
```
jcyang43: We should never go ahead with an `unknown-tag`

.buildkite/scripts/commit_push.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# --- Check for changes before committing ---
+if git diff --quiet --cached; then
+  echo "No changes to commit. Exiting successfully."
+  exit 0
+else
+  echo "--- Committing changes"
+  git commit -s -m "[skip ci] ${COMMIT_MESSAGE:-unknown commit message}"
```
jcyang43: We should never go ahead with a `unknown commit message`

.buildkite/main.yml
```diff
@@ -19,32 +31,45 @@ steps:
...
+     if: build.env("NIGHTLY") == "1" || build.tag =~ /^v?[0-9]+(\.[a-zA-Z0-9]+)*([a-zA-Z]+[0-9]*)?$/
+     key: notify_test_results
+     depends_on: commit_support_matrices
+     agents:
+       queue: tpu_v6e_queue
+     commands:
+       - bash .buildkite/scripts/notify_test_results.sh
+
+   - label: "Record verified commit hashes"
+     if: build.env("NIGHTLY") == "1" || build.tag =~ /^v?[0-9]+(\.[a-zA-Z0-9]+)*([a-zA-Z]+[0-9]*)?$/
```
jcyang43: I think this needs to be updated. 
- release tag: We don't need to be recording commit hashes for release tags. The reason why we want to record commit hashes in the first place is that we know which commits have been verified so we can select from the verified commits when cutting a release. In other words, release tags will only come from already verified commits and hence there's no need to re-record them

.buildkite/main.yml
meiyeh123: The use of unknown-tag is only intended for manual test builds; it acts as a default value to prevent going ahead without a proper tag or nightly build

meiyeh123: or just remove :-unknown-tag
change to?
COMMIT_TAG="$${BUILDKITE_TAG}"

.buildkite/scripts/commit_push.sh
meiyeh123: I'll let it fail if ${COMMIT_MESSAGE} is not set.

.buildkite/main.yml
meiyeh123: I'll remove :-unknown-tag and let it fail if ${BUILDKITE_TAG} is not set.

.buildkite/main.yml
meiyeh123: Done.
Since $BUILDKITE_COMMIT is currently "HEAD", I removed the step that stores it. ([buildkite](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4755/steps/canvas?sid=019a2dfb-0381-4fb2-887d-030455659441) line 29)
I will just use $BUILDKITE_COMMIT directly in all subsequent procedures.

.buildkite/main.yml
meiyeh123: done. coved in .buildkite/scripts/export_commit_message.sh

.buildkite/scripts/commit_push.sh
meiyeh123: done
use git_add_script.sh instead

.buildkite/main.yml
meiyeh123: done

.buildkite/scripts/commit_push.sh
meiyeh123: done

.buildkite/main.yml
meiyeh123: I move them to pipeline_jax.yml. PTAL

meiyeh123:
the latest commit will commit the verified hashes upon successful completion of the nightly steps. (Both pipeline_jax.yml and main.yml)

.buildkite/main.yml
```diff
@@ -25,32 +25,35 @@ steps:
...
+     <<: *run-on-nightly-or-tag
+     key: notify_test_results
+     depends_on: commit_support_matrices
+     agents:
+       queue: cpu
+     commands:
+       - bash .buildkite/scripts/notify_test_results.sh
+
+   - label: "Record verified commit hashes"
+     <<: *run-on-nightly-or-tag
```
jcyang43: Per [previous discussion](https://github.com/vllm-project/tpu-inference/pull/795#discussion_r2471209228), we don't need to run this on release tag

.buildkite/pipeline_jax.yml
```diff
@@ -287,3 +287,25 @@ steps:
...
+       - test_10
+       - test_11
+       - test_12
+       - test_13
+     agents:
+       queue: cpu
+     commands:
+       - |
+         .buildkite/scripts/check_results.sh \
+           "TPU JAX Nightly Tests Failed" test_1 test_2 test_4 test_5 test_6 test_8 test_9 test_10 test_11 test_12 test_13
```
jcyang43: 1. Is there a reason why we can't use the `TPU Test Notification` above directly? 
2. Why did you skip `test_14`, `test_15` & `test_16`?

.buildkite/scripts/commit_support_matrices.sh
```diff
@@ -2,8 +2,9 @@
 set -e
 
 # --- Configuration ---
-REPO_URL="https://github.com/vllm-project/tpu-inference.git"
-TARGET_BRANCH="main"
+# $BUILDKITE_PULL_REQUEST_REPO would be "" if not a pull request. (ex. manual trigger)
+REPO_URL="${BUILDKITE_PULL_REQUEST_REPO:-https://github.com/vllm-project/tpu-inference.git}"
+TARGET_BRANCH="$BUILDKITE_BRANCH"
```
jcyang43: Are the changes in this file still needed? Only the main branch has permission to access the github token for safety reasons. We're not supporting committing to any other branch by design. 

We should add a condition in the yml config such that we don't bother running `commit_support_matrices.sh` or `commit_verified_commit_hashes.sh` if it's not on main branch

.buildkite/scripts/commit_verified_commit_hashes.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
+set -e
+
+# --- Configuration ---
+# $BUILDKITE_PULL_REQUEST_REPO would be "" if not a pull request. (ex. manual trigger)
+REPO_URL="${BUILDKITE_PULL_REQUEST_REPO:-https://github.com/vllm-project/tpu-inference.git}"
+TARGET_BRANCH="$BUILDKITE_BRANCH"
```
jcyang43: Ditto. Please see comment in `commit_support_matrices.sh`

.buildkite/scripts/commit_verified_commit_hashes.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "VLLM_COMMIT_HASH not found in buildkite meta-data"
+    exit 1
+fi
+
+if [ -z "${BUILDKITE_COMMIT:-}" ]; then
+    echo "BUILDKITE_COMMIT not found"
+    exit 1
+fi
+
+if [ ! -f commit_hashes.csv ]; then
```
jcyang43: Let's rename the file to `verified_commit_hashes.csv`, thanks

.buildkite/pipeline_jax.yml
meiyeh123: I think our goal is that only checking the nightly tests (which have `if [[ "$$NIGHTLY" == "1" ]]`).
So it doesn't matter if non-nightly test failed, right? (test_0, test_3, test_7, test_14, test_15, test_16)
1. If we just depend on the `TPU Test Notification`:
   when the nightly tests all passes, but non-nightly tests failed, this will not record the hashes.
   So I create a new step which only check the nightly tests. And recording hashes step depend on it
2. test_0, test_3, test_7, test_14, test_15, test_16 are non-nightly tests. so I skip them.

.buildkite/scripts/commit_support_matrices.sh
meiyeh123: I think use variable instead hardcoding is a better style.
the script won't that specific, and let the yml condition decide what branch can run this.
And we do have the buildkite policy to ensure only the main branch can push the code.
But I think it's okay we just keep it to `main` for more readability.

.buildkite/main.yml
meiyeh123: done. thanks

.buildkite/scripts/commit_verified_commit_hashes.sh
meiyeh123: done

.buildkite/scripts/commit_support_matrices.sh
meiyeh123: I remove the changes.

.buildkite/scripts/commit_verified_commit_hashes.sh
meiyeh123: done. thanks

.buildkite/pipeline_jax.yml
jcyang43: That is incorrect... 
- tests with `if NIGHTLY == 1` condition: these are only run during the nightly build 
- tests without `if NIGHTLY == 1` condition : these are run in both regular build AND nightly build. In other words, these tests are part of the nightly run as well

meiyeh123: Sorry, I misunderstood.
I made record_verified_commit_hashes directly depend on `TPU Test Notification`.

meiyeh123:
The step record_verified_commit_hashes directly depend on `TPU Test Notification` in the latest commit.
And the Buildkite tests in the PR description have been updated.

docker/Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG NIGHTLY_DATE="20250714"
 ARG BASE_IMAGE="us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.12_tpuvm_$NIGHTLY_DATE"
+ARG VLLM_COMMIT_HASH=""
```
jcyang43, line 4: Please add a comment `the latest main will be used if arg unspecified` to avoid potential confusion, thanks

jcyang43 (approved):

PR #829, opened 2025-10-10
Enable spmd on lora

vanbasten23:
# Description

This PR enables SPMD for LoRA.

# Tests

- USE_V6E8_QUEUE=1  VLLM_LOG_LEVEL=INFO  VLLM_XLA_CHECK_RECOMPILATION=1   VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  pytest -s -vv  tests/lora/test_lora.py  (USE_V6E8_QUEUE=1  enables the multi-chip test)
- VLLM_LOG_LEVEL=INFO  VLLM_XLA_CHECK_RECOMPILATION=1   VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  pytest -s -vv  tests/lora/test_lora.py (Without `USE_V6E8_QUEUE=1 `, it uses single chip.) 
- SKIP_JAX_PRECOMPILE=1   VLLM_LOG_LEVEL=INFO  MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  pytest -s -vv -x  tests/lora/test_lora.py (SKIP_JAX_PRECOMPILE=1 for faster testing)
Profile: VLLM_TORCH_PROFILER_DIR=/mnt/disks/persist/myprofiles MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax  python examples/offline_lora_inference.py --tensor-parallel-size 2  (offline inference with profiling.)

# Checklist

### DIFF
--- a/tests/lora/test_lora.py
+++ b/tests/lora/test_lora.py
@@ -1,6 +1,5 @@
 # https://github.com/vllm-project/vllm/blob/ed10f3cea199a7a1f3532fbe367f5c5479a6cae9/tests/tpu/lora/test_lora.py
 import os
-import subprocess
 
 import pytest
 import vllm
@@ -37,20 +36,6 @@ def setup_vllm(num_loras: int, tp: int = 1) -> vllm.LLM:
                     enable_lora=True,
                     max_loras=num_loras,
                     max_lora_rank=8)
-
-
-@pytest.fixture(autouse=True)
-def run_after_each_test():
-    # --- Setup code (runs before each test) ---
-    # print("\nSetting up...")
-    yield  # This is where the test runs
-    # --- Teardown code (runs after each test) ---
-    command = "lsof -t /dev/vfio/* | xargs kill"
-    results = subprocess.run(command,
-                             shell=True,
-                             capture_output=True,
-                             text=True)
-    print(f"Killing TPU resources: {results.stdout}, {results.stderr}")
 
 
 # For multi-chip test, we only use TP=2 because the base model Qwen/Qwen2.5-3B-Instruct has 2 kv heads and the current attention kernel requires it to be divisible by tp_size.
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -156,7 +156,7 @@ steps:
      commands:
        - |
          .buildkite/scripts/run_in_docker.sh \
-           bash -c 'MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python3 -m pytest -s -v -x /workspace/tpu_inference/tests/lora/test_lora.py'
+           bash -c 'MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python3 -m pytest -s -v -x /workspace/tpu_inference/tests/lora/test_lora.py -k multi_lora'
 
 
   # -----------------------------------------------------------------


tests/lora/test_lora.py
```diff
@@ -90,22 +100,24 @@ def test_lora_hotswapping():
...
+@pytest.mark.parametrize("tp", TP)
+def test_multi_lora(tp):
     """
     This test ensures we can run multiple LoRA adapters on the TPU backend, when
     we have enough space to store all of them.
 
     We run "Username6568/Qwen2.5-3B-Instruct-1_plus_1_equals_x_adapter" which
     will force Qwen2.5-3B-Instruct to claim 1+1=x, for a range of x.
     """
+    print(f"xw32 TP: {tp}")
```
vanbasten23: remove

vanbasten23:
cc: @hfan @kyuyeunk @lsy323 @qihqi Could you take a look at the PR? Thanks!

tpu_inference/layers/vllm/sharding.py
```diff
@@ -121,7 +121,24 @@ def _shard_column_parallel_linear_lora(
 
 def _shard_qkv_parallel_linear_lora(layer: MergedQKVParallelLinearWithLoRA,
                                     mesh: Mesh) -> None:
-    _shard_base_linear_lora(layer, mesh)
```
kyuyeunk, line 124: is there a reason why this change is only applied to `MergedQKVParallelLinearWithLoRA` and not other layers like `MergedColumnParallelLinearWithLoRA`?

tpu_inference/layers/vllm/sharding.py
```diff
@@ -145,7 +162,7 @@ def _shard_row_parallel_linear_lora(layer: RowParallelLinearWithLoRA,
 def _shard_module_to_tpu(model: torch.nn.Module, mesh: Mesh) -> None:
     for path, module in model.named_modules():
         for module_type, sharding_func in MODULE_TYPE_TO_SHARDING_FUNC:
-            if isinstance(module, module_type):
```
kyuyeunk, line 148: Can you explain what is the reason for this change?

vanbasten23: Sure. `isinstance` checks inheritance. So if 
```
MODULE_TYPE_TO_SHARDING_FUNC = [
    ...
    (MergedColumnParallelLinearWithLoRA, _shard_column_parallel_linear_lora),
    (MergedQKVParallelLinearWithLoRA, _shard_qkv_parallel_linear_lora),
...
]
```
 and because in vLLM MergedQKVParallelLinearWithLoRA inherits MergedQKVParallelLinearWithLoRA, that means that if the module has type `MergedQKVParallelLinearWithLoRA`, then it will use the wrong sharding function: it will use `_shard_column_parallel_linear_lora` and never uses `_shard_qkv_parallel_linear_lora`

tpu_inference/layers/vllm/sharding.py
vanbasten23: Yeah good call. I sharded both now. For RowParallelLinear, I need to create a lora adapter with lora on RowParallelLinear to test it out so it will be in a follow up PR.

tpu_inference/layers/vllm/sharding.py
kyuyeunk: that's actually by design so that it can be flexible enough that it can be applied to any other child classes. And if we want a different behavior (like this case), the suggestion was to make `MergedQKVParallelLinearWithLoRA` come before `MergedColumnParallelLinearWithLoRA` - thats why I added this comment in the file: `# NOTE: Ordering is important as it calls first matched type of a given module`.

It's mostly nit comment but can you verify if changing the ordering resolves the issue?

kyuyeunk (approved):
Approved because remaining comment is very minor. But can you still verify the suggestion I've left?

tpu_inference/layers/vllm/sharding.py
vanbasten23: Yeah, I verified that reordering resolve the issue. But it feels like it's more implicit and error prone with `isinstance`: we depends on the inheritance order of `MODULE_TYPE_TO_SHARDING_FUNC ` which is not obvious and there is multi-inheritance in python.

vanbasten23:
Thanks for the review!

vanbasten23:
The CI is green: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4474

PR #838, opened 2025-10-13
[Spec Decoding] Precompile jitted helpers for eagle3

Lumosis:
# Description

Jit all the JAX-related operations and precompile the jitted helpers for eagle3.

# Tests

UT tested.

# Checklist

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -5,9 +5,9 @@ from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple
 import jax
 import jax.numpy as jnp
 import numpy as np
-import vllm.envs as envs
 from jax.sharding import NamedSharding, PartitionSpec
 
+import vllm.envs as envs
 from tpu_inference.core.disagg_utils import is_disagg_enabled
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
 from tpu_inference.layers.jax.sample.sampling import sample
@@ -15,6 +15,7 @@ from tpu_inference.layers.jax.sample.sampling_metadata import \
     TPUSupportedSamplingMetadata
 from tpu_inference.logger import init_logger
 from tpu_inference.utils import device_array
+from vllm.utils import cdiv
 
 if TYPE_CHECKING:
     from tpu_inference.runner.tpu_jax_runner import TPUModelRunner
@@ -81,7 +82,7 @@ class CompilationManager:
             self._precompile_gather_logprobs()
             self._precompile_structured_decoding()
             if self.runner.speculative_config:
-                self._precompile_rejection_sampler()
+                self._precompile_speculative_decoding()
 
     def _precompile_backbone_helper(self, name, *, input_ids, positions,
                                     inputs_embeds) -> None:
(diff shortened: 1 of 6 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -82,6 +83,8 @@ def capture_model(self) -> None:
             self._precompile_structured_decoding()
             if self.runner.speculative_config:
                 self._precompile_rejection_sampler()
+                if self.runner.speculative_config.method == "eagle3":
```
jrplatin: Not sure if this is cleaner, but maybe it'd be better to have a `_precompile_speculative_decoding` method or something and we handle which (if any) speculative decoding helpers should be called.

Otherwise, we'll need to have the if-else statement here for each speculative decoding model, correct, which might get unwiedly?

tpu_inference/runner/compilation_manager.py
```diff
@@ -181,6 +184,7 @@ def _precompile_select_from_array_helper(
         hidden_dim: int,
         sharding: Optional[NamedSharding] = None,
         only_equal_paddings: bool = False,
+        check_should_skip: bool = True,
```
jrplatin: Not clear to me what this variable does (from the name)

tpu_inference/runner/compilation_manager.py
```diff
@@ -279,6 +293,15 @@ def _precompile_compute_logits(self) -> None:
                     lora_metadata,
                     num_reqs=num_reqs,
                 )
+            if self.runner.speculative_config and self.runner.speculative_config.method == "eagle3":
```
jrplatin: Same comment as above here -- maybe make this more generic so we handle better handle other spec decoding models?

tpu_inference/runner/compilation_manager.py
```diff
@@ -408,6 +431,211 @@ def _precompile_rejection_sampler(self) -> None:
                         do_sampling=do_sampling,
                     )
 
+    def _precompile_eagle3_helpers(self) -> None:
```
jrplatin, line 431: Is all of this logic specific to eagle3, or is some of it relevant to all spec decoding models? If so, same sentiments as above where maybe it makes sense to abstract the core functionaltiy out to common methods

jrplatin: Similarly, maybe this method should live in a utils/helper file?

tpu_inference/runner/compilation_manager.py
Lumosis: Makes sense. Done

tpu_inference/runner/compilation_manager.py
Lumosis: Done.

tpu_inference/runner/compilation_manager.py
Lumosis: Done

tpu_inference/runner/compilation_manager.py
Lumosis: Currently, all the logic is specific to eagle-3 implementation. We can evolve if other spec decoding strategies are introduced.

PR #843, opened 2025-10-13
Cpu offload v1

saikat-royc:
# Description

the past.

- why is this change being made
Early tests to verify the functional correctness of the tpu_connector_local implementation

- the problem being solved and any relevant context
- why this is a good solution
This solution emulates a similar baseline vs connector comparision done in https://github.com/vllm-project/vllm/blame/main/tests/v1/kv_connector/nixl_integration/run_tpu_disagg_accuracy_test.sh and https://github.com/vllm-project/vllm/blob/main/tests/v1/kv_connector/nixl_integration/test_disagg_accuracy.py

- some information about the specific implementation

    1. offline_inference_kv_cache_verification.py which performs 2 runs of a
       given prompt, one with tpu_connector_local disabled, followed by a
       run with the tpu_connector_local enabled, and compare the output.
    2. pod_tpu_commons_cpu_offload_verification.yaml pod spec to trigger the
       verification py script

1. the place holder for the tests would be re-arranged later.
2. Further the verification script needs to run the test case with the tpu connector local enabled twice, so that we can exercise the load spec statemachine on a prefix cache hit. this will be done in a followup PR

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

I have run the script with the pod spec manually and verified the output match in this current iteration of the code

# Checklist

### DIFF
--- a/examples/gke/pod_tpu_commons_cpu_offload_verification.yaml
+++ b/examples/gke/pod_tpu_commons_cpu_offload_verification.yaml
@@ -1,0 +1,39 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: tpu-job-offline-inference-verification
+  # This pod verifies the correctness of the TPUConnector implementation.
+  # It runs a script that internally performs two text generations:
+  # 1. A baseline run with a standard vLLM engine.
+  # 2. A test run with the TPUConnector enabled.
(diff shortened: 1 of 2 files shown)


saikat-royc:
PR sent out by mistake to main branch

PR #844, opened 2025-10-13
[Feature] [TPU host offload] verification script of baseline vs tpu connector

saikat-royc:
the past.

verification script of baseline vs tpu connector implementation

why is this change being made
Early tests to verify the functional correctness of the tpu_connector_local implementation

the problem being solved and any relevant context

why this is a good solution

```
This solution emulates a similar baseline vs connector comparision done in https://github.com/vllm-project/vllm/blame/main/tests/v1/kv_connector/nixl_integration/run_tpu_disagg_accuracy_test.sh and https://github.com/vllm-project/vllm/blob/main/tests/v1/kv_connector/nixl_integration/test_disagg_accuracy.py
```

some information about the specific implementation

```
offline_inference_kv_cache_verification.py which performs 2 runs of a
given prompt, one with tpu_connector_local disabled, followed by a
run with the tpu_connector_local enabled, and compare the output.
pod_tpu_commons_cpu_offload_verification.yaml pod spec to trigger the
verification py script
```

shortcomings of the solution and possible future improvements.

```
the place holder for the tests would be re-arranged later.
Further the verification script needs to run the test case with the tpu connector local enabled twice, so that we can exercise the load spec statemachine on a prefix cache hit. this will be done in a followup PR
```

FIXES: b/123456
FIXES: #123456

Tests
commands to reproduce.

```
I have run the script with the pod spec manually and verified the output match in this current iteration of the code
```
Checklist

I have performed a self-review of my code.
I have necessary comments in my code, particularly in hard-to-understand areas.
I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/examples/offline_inference_kv_cache_verification.py
+++ b/examples/offline_inference_kv_cache_verification.py
@@ -1,0 +1,139 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+This script performs an automated correctness verification for the TPUConnector.
+
+The verification works by performing a two-stage experiment:
+1.  Baseline Run: It first runs a text generation using a standard vLLM
+    engine configuration without any KV cache connector. The output from this
+    run is considered the "source of truth".
+
+2.  Test Run: It then runs the exact same text generation, but this time
+    with the TPUConnector enabled via the `--kv-transfer-config` argument.
+    It runs the generation multiple times to verify prefix caching.
+
+3.  Comparison: The script compares the output from each test run against the
+    output from the baseline run.
+
+The script succeeds (exits with code 0) only if the generated text is
+bit-for-bit identical in all runs. A fixed seed is used to ensure that the
+generation process is deterministic and the comparison is valid. If any output
+differs, it raises an error, causing the script to fail (exit with a non-zero
+code).
+"""
+
+import copy
+import os
+import time
+from typing import List
+
+import vllm.envs as envs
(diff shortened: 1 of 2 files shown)


examples/offline_inference_kv_cache_verification.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # 2. Run baseline and store the output
+    print("--- Running Baseline (Standard vLLM) ---")
+    baseline_output = run_single_generation(baseline_args)
+    print(f"Baseline Generated Text: {baseline_output!r}")
+    time.sleep(
+        10
+    )  # adding this sleep fixes device busy errors for the next test case run with the connector enabled
+
+    # 3. Run the test with the local tpu kv connector enabled
```
juncgu-google, line 111: should we test the prompt / request twice to verify the correctness of  saving + loading KV cache from CPU mem?

Or, the correctness test will be included in the future PRs?

saikat-royc: it will be added to future PRs as called out in description. I am hitting an assertion error 

```
(EngineCore_DP0 pid=2546) ERROR 10-14 04:45:37 [core.py:784]   File "/workspace/vllm/vllm/v1/core/sched/scheduler.py", line 468, in schedule
(EngineCore_DP0 pid=2546) ERROR 10-14 04:45:37 [core.py:784]     assert num_new_tokens > 0
```

when the exact same prompt is invoked twice for the tpu connector, when the connector reports all tokens have a match in cpu backend. So this needds to be handled in a separate PR

Added a TODO

PR #865, opened 2025-10-14
Data Parallelism support 

wenxindongwork:
This PR introduces support for Data Parallelism in the vLLM TPU backend. 

Data Parallelism is a sharding strategy that is intended to be applied to the following scanerios: 

1. Model replication. Replicating the model across a large slice increases the overall throughput of the system. 
2. KV cache de-duplication for large models with small number of kv heads, e.g. DeepSeekV3, Qwen3 235b, and models that use fp8 kv cache quantization. By default we shard the kv heads by TP, and if there is not enough heads to shard, we end up replicating the heads and kv cache, which results in wasted memory. Using attention DP will eliminate this waste by replicate the attention layer and sending different data to each attention replica.  

Note that the vLLM TPU DP design (SPMD) is very different from vLLM GPU DP design (MPMD). vLLM GPU DP launches multiple vLLM EngineCore instances (one for DP rank) and communicate between processes, whereas vLLM TPU DP launches a single vLLM EngineCore and does the data sharding within one instance. SPMD is a more TPU and JAX native approach to DP. 

**DP scheduler** 
We introduce a new DP scheduler class that extends the base vLLM Scheduler to support data-parallel (DP) execution. It manages request distribution and KV cache allocation across multiple DP ranks, where each rank has its own logical KV cache shard and processes a subset of the total requests.

**Input preparation**
See changes in `tpu_jax_runner._prepare_inputs()` . When DP is enabled, we assign each request to a DP rank.  Input tokens should be sorted by same DP ranks (e.g. inputs tokens from DP rank 0 should come before input tokens from DP rank 1, and so on). We do this sorting and padding in the `_prepare_inputs_dp` function. The input tokens will then be sharded by the DP axis, such that each attention layer replica processes one shard of the global input. 

**Sharding and Attention Axis Updates:**

We introduced a new mesh axis: `attn_dp` to enable attention-only DP. Additionally, we modified sharding annotations for Llama3 to enable both model-wise and attention-only DP. All models should follow the same rough idea: attention weights (e.g. qkv projs, out_proj)  should be replicated across the `data` and `attn_dp` dimension, and data should be sharded across the `data` and `attn_dp` dimension. MLP and MoE layers should be replicated across the `data` dimension but not the `attn_dp` dimension. 

**Usage:** 
It is recommended to use DP with async scheduling 

Model wise DP: `python examples/offline_inference.py   --tensor_parallel_size=4 --data_parallel_size=2 --async-scheduling`
Attention DP: `python examples/offline_inference.py --tensor_parallel_size=8 --kv-cache-dtype=fp8 --additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}' --async-scheduling`

Attention DP will be automatically triggered when the `enable_dp_attention` flag is passed, and the exact dp_size will be automatically determined based on the number of kv heads and TP size. 

**Complementary vLLM PR**

We have to make some minor changes to vLLM upstream. 
https://github.com/vllm-project/vllm/pull/27365/files

# Tests

* New e2e BuildKite test on v6e-8(`test_data_parallel.py`) to verify 1. model parallelism, 2. attention data parallelism, 3. output correctness check.  Performance tests not added in this PR. 
* Unit tests for the `DPScheduler` class and `_prepare_input_dp` function. 

Buildkite: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4993

# Performance

This PR introduces the functional DP implementation. 

Model wise DP 
- Achieves ~72-80% of expected throughput.

Attention DP
- 1.5x gain in effective KV cache size. 

# Future PRs
- Support DP for speculative decoding, sequence paralellism, LoRA, and structured decoding. 
- Support DP for Torchax backend. 

# Checklist

### DIFF
--- a/tests/runner/test_tpu_jax_runner_dp.py
+++ b/tests/runner/test_tpu_jax_runner_dp.py
@@ -48,10 +48,17 @@ class TestTPUJaxRunnerDPInputsLightweight:
         self.runner.block_table_cpu = np.zeros((8, 8), dtype=np.int32)
         self.runner.arange_cpu = np.arange(64, dtype=np.int64)
 
+        # Mock scheduler config for async scheduling
+        self.runner.scheduler_config = MagicMock()
+        self.runner.scheduler_config.async_scheduling = False  # Default to False for most tests
+        self.runner._pre_async_results = None  # Default to None for most tests
+
         # Bind the actual methods to our mock
         self.runner._prepare_inputs_dp = TPUModelRunner._prepare_inputs_dp.__get__(
             self.runner)
         self.runner._prepare_dp_input_metadata = TPUModelRunner._prepare_dp_input_metadata.__get__(
+            self.runner)
+        self.runner._prepare_async_token_substitution_indices_dp = TPUModelRunner._prepare_async_token_substitution_indices_dp.__get__(
             self.runner)
 
     def _create_mock_scheduler_output(self,


tpu_inference/layers/jax/sharding.py
```diff
@@ -6,14 +6,15 @@
 from vllm.config import VllmConfig
 
 BATCH_AXIS_NAME = 'data'
-SEQUENCE_AXIS_NAME = 'data'
-DATA_AXIS_NAME = 'data'
+SEQUENCE_AXIS_NAME = ('data', 'attn_dp')
```
kyuyeunk: Can you use enum to group these variables?

So something like this


```
class ShardingAxisName(Enum):
  BATCH = 'data'

  def __getattr__(self, item):
        return getattr(self, item).value
```


And other codes can use something like this
```
P(ShardingAxisName.BATCH, None)
```

tpu_inference/models/vllm/jax_fused_moe.py
```diff
@@ -322,6 +329,8 @@ def jax_fused_moe_func(
     x = hidden_states[token_indices_sorted]
 
     if use_ep:
+        x = jax.lax.with_sharding_constraint(
```
kyuyeunk: Any rationale for changing this to be explicit?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -0,0 +1,232 @@
...
+        def _get_computed_blocks_dp(self, request: Request):
+            """Get computed blocks with DP logic - check all managers for cache hits."""
+            best_blocks = None
+            best_tokens = 0
+            best_rank = None
+            
+            # Check all DP managers for the best cache hit
+            for rank in range(self.dp_size):
+                blocks, tokens = self.dp_kv_cache_managers[rank].get_computed_blocks(request)
+                if len(blocks.blocks[0]) > best_tokens:
```
yaochengji: We are comparing num of blocks with number of tokens?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -0,0 +1,232 @@
...
+        def _init_dp_state(self):
+            """Initialize DP-specific state."""
+            self.assigned_dp_rank: dict[str, int] = {}
+            self.round_robin_counter = itertools.cycle(range(self.dp_size))
+        
+        def _replace_kv_cache_manager(self):
+            """Replace single KV cache manager with multiple DP managers."""
+            # Create DP-specific KV cache config
+            dp_kv_cache_config = self.kv_cache_config
+            dp_kv_cache_config.num_blocks = self.kv_cache_config.num_blocks // self.dp_size
```
yaochengji: Is it safe to modify the original `self.kv_cache_config`?

tpu_inference/core/sched/utils.py
```diff
@@ -0,0 +1,28 @@
+from vllm.config import VllmConfig
+
+
+def get_dp_size(vllm_config: VllmConfig) -> int:
```
kyuyeunk: Not really a fan of this code pattern:

```
try:
  val = vllm_config.additional_config["some_key]
except (KeyError, TypeError, AttributeError):
  val= 1
```

Probably not a scope of this PR but is it possible to add this feature?
1. During vLLM initialization, create a logic that parses `vllm_config.additional_config["sharding"]`
2. Add logic that creates `ShardingStrategy` config object based on the parsed value: https://github.com/vllm-project/tpu-inference/blob/1b8010d8bfbf8cb7dcdfb010bf79bcb84c01b144/tpu_inference/layers/jax/sharding.py#L20-L38
3. For any subsequent reference to sharding strategy, modify the code to reference the created shrading config object.

tpu_inference/core/sched/utils.py
kyuyeunk, line 1: Please use different name (http://go/no-utils)

kyuyeunk, line 1: Any reason for creating a new file?

I feel like functionality overlaps with these files
- https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/sharding.py
- https://github.com/vllm-project/tpu-inference/blob/1b8010d8bfbf8cb7dcdfb010bf79bcb84c01b144/tpu_inference/utils.py#L166-L242

tpu_inference/core/sched/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Returns:
+        dp_size: number of model replicas
+        attn_dp: number of attention replicas within each model replica
+        total_dp: total data parallel size = dp_size * attn_dp
+    """
+    try:
+        dp_size = vllm_config.additional_config["sharding"]["sharding_strategy"]["data_parallelism"]
+    except (KeyError, TypeError, AttributeError):
+        dp_size = 1
```
kyuyeunk: Why not use `ParallelConfig.data_parallel_size`?: https://github.com/vllm-project/vllm/blob/8c851f6d044bf7922122a1735e57aea727e30d45/vllm/config/parallel.py#L71

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -0,0 +1,232 @@
...
+                scheduled_requests.extend(base_output.scheduled_cached_reqs.req_ids)
+            
+            assigned_dp_rank = {
+                req_id: self.assigned_dp_rank[req_id]
+                for req_id in scheduled_requests
+                if req_id in self.assigned_dp_rank
+            }
+            logger.debug(f"Scheduled requests DP ranks: {assigned_dp_rank}")
+
+            return DPSchedulerOutput(
```
yaochengji: I think we can put these assignment logic in `DPSchedulerOutput`'s __init__ method, WDYT?

yaochengji: Another choice is to make `SchedulerOutput` a data member of `DPSchedulerOutput`, not base class.

tpu_inference/core/sched/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    
+    try:
+        tp = vllm_config.parallel_config.tensor_parallel_size
+    except (KeyError, TypeError, AttributeError):
+        tp = 1
+
+    if hasattr(vllm_config, 'model_config') and hasattr(vllm_config.model_config, 'hf_config'):
+        num_kv_heads = vllm_config.model_config.hf_config.num_key_value_heads
```
kyuyeunk: Please use `vllm_config.model_config.get_total_num_kv_heads()` instead: https://github.com/vllm-project/vllm/blob/8c851f6d044bf7922122a1735e57aea727e30d45/vllm/config/model.py#L1287

tpu_inference/core/sched/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        dp_size = vllm_config.additional_config["sharding"]["sharding_strategy"]["data_parallelism"]
+    except (KeyError, TypeError, AttributeError):
+        dp_size = 1
+    
+    try:
+        tp = vllm_config.parallel_config.tensor_parallel_size
+    except (KeyError, TypeError, AttributeError):
+        tp = 1
+
+    if hasattr(vllm_config, 'model_config') and hasattr(vllm_config.model_config, 'hf_config'):
```
kyuyeunk: Is there a case when this `hasattr` will return false?

tpu_inference/models/vllm/jax_fused_moe.py
wenxindongwork: No, if you prefer removing it, it will also work. Let me remove it then for succinctness.

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: Thank you for catching it! I don't know why Claude did this during refactoring

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: that's a great idea, I went with the first method so I don't have to change typing annotations and attribute getter syntax.

tpu_inference/core/sched/utils.py
wenxindongwork: Didn't know this existed, super helpful!

tpu_inference/core/sched/utils.py
wenxindongwork: sg, moved to sharding.py

tpu_inference/layers/jax/sharding.py
wenxindongwork: cool, let me change this. It also makes import cleaner this way. I left it was a normal python class instead of enum. 

This is because if I make an Enum, we will have to access the values using `ShardingAxisName.BATCH.value`

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: `kv_cache_config` is also used to initialize the kv cache: 
https://github.com/vllm-project/tpu-inference/blob/ceb27a8b2053b880d726024543654b066389db11/tpu_inference/runner/kv_cache_manager.py#L186-L189

I can modify `kv_cache_config` in-place here: https://github.com/vllm-project/tpu-inference/blob/ceb27a8b2053b880d726024543654b066389db11/tpu_inference/platforms/tpu_jax.py#L115-L116

But I'm not sure why the logic used to initialize kv cachei is `num_blocks = kv_cache_tensor.size // page_size_bytes` instead of `kv_cache_config.num_blocks`.  Can we use the latter?

tpu_inference/core/sched/utils.py
wenxindongwork: removed this check

tpu_inference/core/sched/utils.py
wenxindongwork: Sounds good, I made this change, can you please check if it looks as you've expected?

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -339,11 +333,15 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
         self.compilation_manager.capture_model()
 
+    @time_function
     def execute_model(
         self,
         scheduler_output: "VllmSchedulerOutput",
         intermediate_tensors: Optional[IntermediateTensors] = None,
     ) -> ModelRunnerOutput:
+        if os.environ.get("DISABLE_JIT", "0") == "1":
+            with jax.disable_jit():
```
vanbasten23: this is very handy!

tests/runner/test_tpu_jax_runner_dp.py
```diff
@@ -0,0 +1,555 @@
...
+        # Verify utility functions were called
+        mock_runner_utils.get_padded_token_len.assert_called()
+
+    def test_prepare_inputs_dp_error_conditions(self):
+        """Test error handling in DP input preparation."""
+        # Test with zero scheduled tokens
+        scheduler_output = self._create_mock_scheduler_output({}, {})
+        scheduler_output.total_num_scheduled_tokens = 0
+
+        with pytest.raises(AssertionError):
```
vanbasten23, line 116: could you add a comment or check which exception are we expecting?

tpu_inference/worker/tpu_worker_jax.py
```diff
@@ -101,6 +102,9 @@ def __init__(self,
                 # For TPU, we can only have 1 active profiler session for 1 profiler
                 # server. So we only profile on rank0.
                 self.profile_dir = envs.VLLM_TORCH_PROFILER_DIR
+                # DO_NOT_SUBMIT: tmp hack to upload profile to gs bucket. will revert.
```
vanbasten23: revert?

tpu_inference/utils.py
```diff
@@ -294,3 +296,22 @@ def get_jax_dtype_from_str_dtype(str_dtype: str) -> jnp.dtype:
...
+
+def time_function(func):
+    """
+    A decorator to measure the execution time of a function.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        start_time = time.perf_counter()
+        result = func(*args, **kwargs)
```
vanbasten23, line 309: do you need something like "jax.block_until_ready"?

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -300,12 +285,18 @@ def _init_inputs(self) -> None:
...
         # to avoid dynamic shapes. Also, avoid suboptimal alignment.
-        self.max_num_reqs = max(scheduler_config.max_num_seqs, MIN_NUM_SEQS)
+        # The total number of requests is dp_size * max_num_seqs
+        self.max_num_reqs = max(self.dp_size * scheduler_config.max_num_seqs,
+                                MIN_NUM_SEQS)
         # [16, 32, 64, 128, 256, 512, 1024, 2048]
         self.num_tokens_paddings = runner_utils.get_token_paddings(
-            min_token_size=16,
-            max_token_size=scheduler_config.max_num_batched_tokens,
+            min_token_size=max(16, self.dp_size),
```
vanbasten23, line 329: why is min_token_size max(16, self.dp_size)?

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -770,15 +782,324 @@ def _execute_model(
...
+        logits_indices_selector = all_positions[sorted_indices]
+
+        return (req_ids_dp, req_indices_dp, num_scheduled_tokens_per_dp_rank,
+                scheduled_tokens_per_dp_rank, num_req_per_dp_rank,
+                padded_num_scheduled_tokens_per_dp_rank, padded_num_reqs,
+                padded_total_num_scheduled_tokens, padded_num_reqs_per_dp_rank,
+                logits_indices_selector, max_num_reqs_per_dp_rank)
+
     def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
+        if self.dp_size > 1:
```
vanbasten23, line 1056: So for attn dp case, self.dp_size here should be 1, right?

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -130,17 +131,19 @@ def tensor_sharded_gmm_row_parallel(
...
         mesh=mesh,
-        in_specs=(P(None, "model"), P(None, None, "model"), P()),
+        in_specs=(P(None, ShardingAxisName.MLP_TENSOR),
+                  P(None, None, ShardingAxisName.MLP_TENSOR), P()),
         out_specs=(P()),
         check_rep=False,
     )(lhs, rhs, group_sizes)
 
 
+# TODO (wenxindongwork): support model-wise DP.
```
vanbasten23: is this todo still relevant? In the pr description, it seems to mention that this pr already support model-wise dp.

tpu_inference/layers/vllm/sharding.py
```diff
@@ -99,11 +101,12 @@ def _shard_lm_head(layer: ParallelLMHead, mesh: Mesh):
...
     # weight from VocabParallelEmbedding
     weight = _convert_to_torchax_and_shard(
-        layer.weight, NamedSharding(mesh, P('model', None)))
+        layer.weight, NamedSharding(mesh, P(ShardingAxisName.MLP_TENSOR,
+                                            None)))
     layer.weight = Parameter(weight, requires_grad=False)
     if layer.bias is not None:
-        bias = _convert_to_torchax_and_shard(layer.bias,
-                                             NamedSharding(mesh, P('model')))
+        bias = _convert_to_torchax_and_shard(
```
vanbasten23: also change the sharding for lora. e.g. at line 138 "ora_b_partition_spec = P(None, None, 'model', None)"?

tpu_inference/models/jax/qwen2.py
```diff
@@ -32,23 +33,26 @@ def __init__(self, config: Qwen2Config, dtype: jnp.dtype, rng: nnx.Rngs):
             intermediate_size,
             use_bias=False,
             param_dtype=dtype,
-            kernel_init=nnx.with_partitioning(init_fn, (None, "model")),
+            kernel_init=nnx.with_partitioning(
```
vanbasten23: What's the test command that can exercise the change in this file?

tpu_inference/layers/jax/attention/attention.py
```diff
@@ -42,15 +43,15 @@ class Attention(nnx.Module):
     mesh: Mesh
     kv_cache_dtype: str
 
-    dnh_sharding: Sharding = ()
```
gpolovets1, line 45: Does this trigger an AG on the weights? Is the tradeoff still worth it?

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -27,7 +28,7 @@ def sample(
     if tpu_sampling_metadata.do_sampling:
         # Unshard the logits explicity to avoid latency increase.
         logits = jax.lax.with_sharding_constraint(
-            logits, NamedSharding(mesh, P(None, None)))
```
gpolovets1, line 30: Do you know if sampling is compute bound? If not, then I wonder if this increases the latency due to AG at the end.
Maybe sparsecore offloading would be a good strategy to consider as well?

tpu_inference/utils.py
gpolovets1, line 1: Just wanted to confirm if we need to update the get_max_concurrency_for_kv_cache_config method: https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_utils.py#L773
Or any of the other memory statistics reported by the vLLM log.

tpu_inference/utils.py
wenxindongwork: we could block ready on all the args, but that might be a bit constraining, perhaps we can let folks add block until readys inside the function they want to time?

tpu_inference/runner/tpu_jax_runner.py
wenxindongwork: we want the min_token_size to be divisible by self.dp_size, since every dp rank should get at least one token

tpu_inference/runner/tpu_jax_runner.py
wenxindongwork: self.dp_size is the product of dp and attn_dp, so in the case of atten_np, here the self.dp_size should be greater than 1.

tpu_inference/layers/vllm/sharding.py
wenxindongwork: yes! I haven't made any changes to the LoRA path, I want to do it in another PR. Do you think I can do that? right now the changes should not break LoRA

tpu_inference/models/jax/qwen2.py
wenxindongwork: It should be the same as the JAX ones but with `MODEL_IMPL_TYPE=vllm`. 

E.g. `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1 MODEL_IMPL_TYPE= vllm python examples/offline_inference.py  --model=Qwen/Qwen3-30B-A3B --tensor_parallel_size=8   --max_model_len=1024 --enable-expert-parallel --additional_config='{"sharding":{"sharding_strateg y": {"enable_dp_attention":1}}}'`

tpu_inference/layers/vllm/fused_moe.py
wenxindongwork: model-wise support for moe is not supported in the torchax backend. I will need to add this support in another PR.

tpu_inference/layers/jax/attention/attention.py
wenxindongwork: Thank you for catching this! this shouldn't be sharded. Though this change did not impact performance weirdly.

tpu_inference/layers/jax/sample/sampling.py
wenxindongwork: I got slightly better performance when using this sharding, but I haven't thoroughly tested this out. There will be an all-gather but async scheduling might have helped with hiding this latency.

kyuyeunk:
are you planning to create a separate PR for sharding annotation related changes? it seems like that change makes the PR significantly larger than just DP related changes and make it hard to review.

tpu_inference/utils.py
wenxindongwork: currently the number returns by `get_max_concurrency_for_kv_cache_config` is correct because it is calculated based off the global kv cache shape

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -242,45 +257,21 @@ def _init_random(self):
...
-        except KeyError:
-            enforce_device_order = False
-
-        logger.info(f"Device sequence enforced: {enforce_device_order}")
-
-        if os.getenv("NEW_MODEL_DESIGN", False):
-            self.mesh = build_mesh(self.devices, sharding_strategy)
+        sharding_strategy: ShardingConfigManager = self.vllm_config.sharding_config
+        axis_names = ("data", "attn_dp", "expert", "model")
+        mesh_shape = (sharding_strategy.model_dp_size,
```
kyuyeunk: I believe the new MoE kernel that was recently introduced always expect mesh shape rank of 2 and changing this to rank of 4 will break the MoE kernel: https://github.com/vllm-project/tpu-inference/blob/60c14f53e1a0e5a7348e09bc60d51d79461d0f71/tpu_inference/kernels/fused_moe/v1/kernel.py#L861

@bythew3i, @lsy323 is my understanding correct?

bythew3i: yes, you need to make some changes accordingly to support this

kyuyeunk: @bythew3i thanks for confirming it. @wenxindongwork, can you modify the code so that the default behavior is to create mesh shape rank 2 and only use rank 4 when needed? Given the current bandwidth, I wouldn't expect the moe kernel support for arbitrary mesh shape to land anytime soon.

PR #867, opened 2025-10-15
Pin-Aware LRU Cache Policy for Local CPU Backend

saikat-royc:
# Description

the past.

This commit introduces a caching layer for the `LocalCPUBackend` to manage offloaded KV cache data efficiently. The backend is enhanced with a size-based LRU (Least Recently Used) eviction policy and a pinning mechanism to protect in-flight cache entries from being prematurely evicted. The core improvements include:

1.  LRU Eviction Policy: `collections.OrderedDict` to maintain the access order of cache entries. `max_cpu_cache_size_bytes` limit is introduced, ensuring the cache size is capped based on the memory footprint of the stored tensors, not just the number of entries `add` method contains eviction logic that removes the least recently used items when the cache exceeds its capacity.

2.  Cache Entry Pinning: A pinning mechanism is implemented to prevent essential cache entries from being evicted. Pinned entries are ignored by the LRU eviction logic. The `contains` method now accepts a `pin_on_hit` parameter. When set, a successful cache lookup will automatically pin the corresponding key. A corresponding `unpin_keys` method allows for the release of these pins, making the entries eligible for eviction again.

3. The `TPUConnector` is updated to integrate with this new caching logic, ensuring that the pinning and unpinning operations are synchronized with the request lifecycle: Pinning on Lookup: When the scheduler checks for a cache hit in `get_num_new_matched_tokens`, it now calls `contains(key, pin_on_hit=True)`, immediately protecting any found entries. Delayed Unpinning: To ensure data integrity, unpinning is deferred until a request is fully complete. The worker's `get_finished` method is responsible for regenerating the keys associated with a finished request and calling `unpin_keys` just before the request's resources are released.

4. This implementation gracefully handles below scenarios: Full Prefix Hit (`N-1` Workaround): When a prompt of N tokens is a full cache hit, the connector reports N-1 tokens to the vLLM scheduler to avoid an assertion. The logic correctly handles this by:
   *   Pinning all N keys during the initial lookup.
   *   Storing the true N-token list in the internal `RequestTracker`.
   *   Using this true list in `get_finished` to regenerate and unpin all N original keys, ensuring no pins are leaked.
   Newly Generated Tokens: For a request with N prompt tokens and M generated tokens, the system correctly identifies that only the N prompt keys were ever pinned.
   The final call to `unpin_keys` for the full N+M sequence results in a correct unpinning of the N keys and a harmless no-op for the M keys, demonstrating robust and precise state management.

FIXES: b/123456
FIXES: #123456

# Tests
Tested with existing verification test cases in offline_inference_kv_cache_verification.py.  **These test would not exercise eviction. more tests to be added in followup PRs

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/distributed/local_cpu_backend.py
+++ b/tpu_inference/distributed/local_cpu_backend.py
@@ -1,13 +1,19 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 
-from typing import Any, Optional
+import os
+import sys
+from collections import OrderedDict
+from typing import Any, List, Optional
 
 from tpu_inference.logger import init_logger
 
 from .util import CacheKey
 
 logger = init_logger(__name__)
+
+GB = 1024**3
+DEFAULT_CPU_CACHE_SIZE_BYTES = 1 * GB
 
 
 class LocalCPUBackend:
@@ -18,6 +24,9 @@ class LocalCPUBackend:
     worker, running in the same process, can share the same cache.
     The scheduler reads from this to find cache hits, and the worker writes
     to it after saving KV blocks from the TPU.
+
+    It implements an LRU (Least Recently Used) eviction policy with a maximum
+    size limit and support for pinning cache entries to prevent eviction.
     """
     _instance: Optional["LocalCPUBackend"] = None
     _initialized: bool = False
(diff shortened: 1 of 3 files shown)


tpu_inference/distributed/local_cpu_backend.py
```diff
@@ -27,15 +33,33 @@ def __new__(cls, *args, **kwargs):
             cls._instance = super(LocalCPUBackend, cls).__new__(cls)
         return cls._instance
 
-    def __init__(self):
+    def __init__(self, max_size_bytes: int = 1 * 1024**3):  # Default to 1 GB
```
juncgu-google: how about add a macro variable `DEFAULT_MAX_BUFFER_SIZE`

tpu_inference/distributed/local_cpu_backend.py
```diff
@@ -27,15 +33,33 @@ def __new__(cls, *args, **kwargs):
...
         if self._initialized:
             return
-        # The cache is now a dictionary mapping CacheKey -> KV cache data
-        self.cache: dict[CacheKey, Any] = {}
+
+        env_cache_size_gb = os.getenv("TPU_OFFLOAD_CPU_CACHE_SIZE_GB")
```
juncgu-google: How about `env_cache_size_gb = int(os.getenv("TPU_OFFLOAD_CPU_CACHE_SIZE_GB", str(default_max_size_gb)))` to remove the following if-else statement.

tpu_inference/distributed/local_cpu_backend.py
saikat-royc: good point. done. using the name `DEFAULT_CPU_CACHE_SIZE_BYTES` to keep it consistent with the env variable name

tpu_inference/distributed/local_cpu_backend.py
saikat-royc: done

PR #906, opened 2025-10-20
[Feature][TPU host offload] Improvements to the save workflow for tpu connector local

saikat-royc:
# Description

the past.

Improvements to the save path to identify the delta tokens that actually require to be saved.

    1. Add logic to identify precedeeing tokens in a prefix cache hit that
       does not need to be saved again in wait_for_save().
    2. During save operation, calculate the absolute token positions so that the key that is
       created for the token chunks is is consistent. This ensures that in
       the get_finished() call, the keys derived to unpin from the cpu backend
       cache is correct.
    3. Introduce an env variable TPU_OFFLOAD_DECODE_SAVE (default '0')
       to skip save tokens during decode phase and finished requests during decode. The finished request needed a bit of handling to ensure we report the reqs as finished in get_finished() correctly.
    5. Minor variable renames for readability
    6. minor fixes in load reporting time
    7. add seed to vllm serve pods used for benchmark runs

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

1. existing verification tests
2. sglang benchmark tool run

# Checklist

### DIFF
--- a/tpu_inference/distributed/tpu_connector_local.py
+++ b/tpu_inference/distributed/tpu_connector_local.py
@@ -85,6 +85,7 @@ Worker Side Execution:
     KV data from TPU to CPU and update the CPU backend. It then waits for all
     submitted save tasks for the current step to complete.
 """
+import os
 import time
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass, field
@@ -120,6 +121,9 @@ class SaveSpec:
     """A confirmed work order for the worker to save KV data."""
     skip_leading_tokens: int
     is_final_save: bool = False
+    # A direct signal to the worker to skip the data transfer but still
+    # process the completion signal if is_final_save is True.
+    skip_save: bool = False
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


tpu_inference/distributed/tpu_connector_local.py
```diff
@@ -450,21 +458,25 @@ def _prepare_req_meta(
...
-            f"Preparing req meta for req: {tracker.req_id}, num_total_tokens={num_total_tokens}, num_saved_tokens={tracker.num_saved_tokens}, has_new_tokens={has_new_tokens}, is_finished={is_finished}"
-        )
-
+        has_new_tokens = num_total_tokens > tracker.save_watermark
         # Determine if a save is needed for this step
         should_save = False
         if is_finished:
-            should_save = True
+            # If the request is finished, we should save unless
+            # TPU_OFFLOAD_FINISHED_REQ_SAVE is enabled.
```
juncgu-google: why do we need `TPU_OFFLOAD_FINISHED_REQ_SAVE` for saving finished req or not?
A finished request is already in the decode phase, `decode_save` determines its save behaviors. Line 475 ~ 482 should be sufficient to cover the cases.

saikat-royc: Ack. I have simplified the logic to reuse the same TPU_OFFLOAD_DECODE_SAVE to handle the scenario where a request reports finished, when it is also in decode phase.

PR #915, opened 2025-10-21
Fix the jax device ordering.

wang2yn84:
# Description

When user create the mesh outside, jax devices has certain order. We need to maintain such order to make sure tpu inference sees the same mesh as upstream modules. Previously we assumes jax devices  index are the same as device id. However, there are cases that such assumption are invalid. This CL orders devices by devices id instead. Run the test locally and it's working.

# Tests

Manually tested in Tunix.

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_jax_runner.py
+++ b/tpu_inference/runner/tpu_jax_runner.py
@@ -154,6 +154,15 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         except KeyError:
             sharding_strategy = {"tensor_parallelism": len(self.devices)}
 
+        try:
+            enforce_device_order = self.vllm_config.additional_config[
+                "sharding"]["sharding_strategy"]["device_indexes"] is not None
+
+        except KeyError:
+            enforce_device_order = False
+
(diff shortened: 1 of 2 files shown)


tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -154,7 +154,14 @@ def _init_mesh(self) -> None:
...
 
-        if os.getenv("NEW_MODEL_DESIGN", False):
+        try:
+            device_indexes = self.vllm_config.additional_config["sharding"][
+                "sharding_strategy"]["device_indexes"]
+        except KeyError:
+            device_indexes = None
+            print("Device sequence not enforced.")
+
+        if os.getenv("NEW_MODEL_DESIGN", False) or device_indexes:
```
hfan: This means if `device_indexes` is specified in the additional_config, it always use the "old model design"? This limits what models tunix can use; and the code branch for not using NEW_MODEL_DESIGN might be on the way for deprecation.

@bzgoogle @gpolovets1 to keep me honest.

wang2yn84: Chatted offline, decided to when device_indexes is provided by upstream, bypass the make optimal mesh and use the default mesh creation. Updated!

PR #920, opened 2025-10-22
[Spec Decoding] Merge jitted helpers for eagle3

Lumosis:
# Description

With this PR, the gap between decoding is reduced from 8.5ms to 6ms while the acceptance rate remains the same.
Next step: avoid waiting for TPU->CPU transfer,

# Tests

UTs tested.

# Checklist

### DIFF
--- a/tpu_inference/spec_decode/jax/eagle3.py
+++ b/tpu_inference/spec_decode/jax/eagle3.py
@@ -55,13 +55,6 @@ class Eagle3Proposer:
         self.state.model.embed_tokens = target_model.model.embed
 
     @functools.partial(jax.jit, static_argnums=(0, ))
-    def _concate_hidden_states(
-            self, aux_hidden_states: tuple[jax.Array, ...]) -> jax.Array:
-        """JIT-compiled helper for concatenating auxiliary hidden states."""
-        # Concat aux hidden states along feature dim.
-        return jnp.concatenate(aux_hidden_states, axis=-1)
-
-    @functools.partial(jax.jit, static_argnums=(0, ))
     def _prepare_input_ids(
             self, query_start_loc: jax.Array, target_token_ids: jax.Array,
             next_token_ids: jax.Array,
@@ -88,7 +81,7 @@ class Eagle3Proposer:
         return input_ids, last_token_indices
 
     @functools.partial(jax.jit, static_argnums=(0, ))
-    def _prepare_input_loop(
+    def _update_inputs_for_loop_speculation(
         self, positions: jax.Array, seq_lens: jax.Array,
         block_tables: jax.Array
     ) -> tuple[jax.Array, jax.Array, jax.Array, jax.Array, jax.Array]:
@@ -120,10 +113,29 @@ class Eagle3Proposer:
         return positions, clamped_positions, new_seq_lens, query_start_loc, new_block_tables
 
     @functools.partial(jax.jit, static_argnums=(0, ))
-    def _stack_draft_token_ids_in_jit(
+    def _stack_draft_token_ids(
             self, draft_token_ids_list: list[jax.Array]) -> jnp.ndarray:
         """JIT-compiled helper for stacking draft token IDs."""
         return jnp.stack(draft_token_ids_list, axis=1)
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _prepare_hidden_states_and_input_ids(
+        self,
+        aux_hidden_states: tuple[jax.Array, ...],
+        query_start_loc: jax.Array,
+        target_token_ids: jax.Array,
+        next_token_ids: jax.Array,
+        num_reqs: jax.Array,
+    ) -> tuple[jax.Array, jax.Array, jax.Array]:
+        target_hidden_states = jnp.concatenate(aux_hidden_states, axis=-1)
+        target_hidden_states = self.combine_hidden_states_fn(
+            self.state, target_hidden_states)
+
+        input_ids, last_token_indices = self._prepare_input_ids(
+            query_start_loc, target_token_ids, next_token_ids, num_reqs)
+        # NOTE(pooyam): For now, we don't support multimodal.
+
+        return target_hidden_states, input_ids, last_token_indices
 
     def prepare_inputs(
         self,
@@ -152,16 +164,20 @@ class Eagle3Proposer:
         num_kv_cache_groups = len(self.runner.kv_cache_config.kv_cache_groups)
         draft_kv_cache_group_id = num_kv_cache_groups - 1
         block_tables = self.runner.input_batch.block_table[
-            draft_kv_cache_group_id].get_device_tensor()
+            draft_kv_cache_group_id].get_cpu_tensor().reshape(-1)
         # Number of active requests in this step (un-padded count).
         num_reqs = self.runner.input_batch.num_reqs
 
         if num_rejected_tokens is None:
-            target_hidden_states, input_ids, last_token_indices, block_tables = self._prepare_draft_inputs_in_jit(
-                self._concate_hidden_states(aux_hidden_states),
-                attn_metadata.query_start_loc, input_ids, next_token_ids,
-                block_tables, jnp.asarray([num_reqs], dtype=jnp.int32))
-            attn_metadata = replace(attn_metadata, block_tables=block_tables)
+            num_reqs = device_array(self.mesh,
+                                    np.asarray([num_reqs], dtype=jnp.int32))
+            # block_tables = device_array(self.mesh, block_tables)
+            attn_metadata = replace(attn_metadata,
+                                    block_tables=device_array(
+                                        self.mesh, block_tables))
+            target_hidden_states, input_ids, last_token_indices = self._prepare_hidden_states_and_input_ids(
+                aux_hidden_states, attn_metadata.query_start_loc, input_ids,
+                next_token_ids, num_reqs)
             return target_hidden_states, input_ids, last_token_indices, attn_metadata
 
         # Host copies from the metadata prepared by the runner.
@@ -218,17 +234,18 @@ class Eagle3Proposer:
         # Update seq_lens for active requests only: new_seq_lens = s - n.
         new_seq_lens_cpu = seq_lens_cpu - nrt_cpu
 
-        query_start_loc, seq_lens, token_indices = device_array(
+        query_start_loc, seq_lens, token_indices, num_reqs, block_tables = device_array(
             self.mesh,
-            (new_query_start_loc_cpu, new_seq_lens_cpu, token_indices_cpu))
+            (new_query_start_loc_cpu, new_seq_lens_cpu, token_indices_cpu,
+             np.asarray([num_reqs], dtype=jnp.int32), block_tables))
 
-        return self._prepare_inputs_in_jit(
+        attn_metadata = replace(attn_metadata, block_tables=block_tables)
+        return self._filter_token_and_prepare_initial_inputs(
             token_indices, query_start_loc, seq_lens, input_ids,
-            aux_hidden_states, attn_metadata, next_token_ids, block_tables,
-            jnp.asarray([num_reqs], dtype=jnp.int32))
+            aux_hidden_states, attn_metadata, next_token_ids, num_reqs)
 
     @functools.partial(jax.jit, static_argnums=(0, ))
-    def _prepare_inputs_in_jit(
+    def _filter_token_and_prepare_initial_inputs(
         self,
         token_indices: jax.Array,
         query_start_loc: jax.Array,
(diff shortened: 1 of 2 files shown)


tests/e2e/test_speculative_decoding.py
```diff
@@ -240,38 +240,38 @@ def test_ngram_performance_random(
         }, 3.0)
 
 
-def test_eagle3_correctness(
```
py4, line 243: I think it's better not to disable tests and handle precompilation. This also make sures EAGLE functionality won't be broken due to upstream changes without us knowing

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -275,32 +275,86 @@ def propose(
...
-        block_tables = self._reshape_block_tables(block_tables)
+        block_tables = block_tables.reshape(-1)
         attn_metadata = replace(attn_metadata, block_tables=block_tables)
 
+        return target_hidden_states, input_ids, last_token_indices, attn_metadata
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _select_draft_token_ids_in_jit(
+        self,
+        hidden_states,
```
py4: Type for these args and other args in other functions

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -275,32 +275,86 @@ def propose(
...
+
+    def _select_draft_token_ids(
+        self,
+        hidden_states,
+        last_token_indices,
+    ):
+        sample_hidden_states = hidden_states[last_token_indices]
+        return self._get_draft_token_ids(sample_hidden_states)
+
+    def _get_draft_token_ids(self, hidden_states):
```
py4: If this is not used standlone, maybe remove it and just keep the jitted function

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,56 +203,70 @@ def prepare_inputs(
...
-        Returns:
-            A tuple containing the updated KV caches and a tensor of proposed
-            draft token IDs.
-        """
+        return self._prepare_draft_inputs(target_hidden_states,
+                                          query_start_loc, target_token_ids,
+                                          next_token_ids, attn_metadata)
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _prepare_draft_inputs_in_jit(self, target_hidden_states,
```
py4: Seems to be pointless as all usages of `_prepare_draft_inputs` are already jitted function. i mean this jitted function is not having any benefit compared to just using the non jitted function

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -275,32 +275,86 @@ def propose(
...
+        last_token_indices,
+        target_hidden_states,
+    ) -> tuple[list[jax.Array], jnp.ndarray]:
+        """Proposes draft tokens using the draft model.
+        Returns:
+            A tuple containing the updated KV caches and a tensor of proposed
+            draft token IDs.
+        """
+
         kv_caches, hidden_states, residual = self.model_fn(
```
py4, line 322: The `input_ids` and `target_hidden_states` for first speculation are handled through `prepare_inputs` called by spec decoding manager? If yeah, let's add a comment here. Previously we used to do input prep for each speculation here, now we have merged the first one with `prepare_inputs` call to improve perf but it's not explicit now and can be misleading

py4 (commented):
if a jitted function is only used within another jitted function, where is no point in marking that function as jitted as things get inlined anyway

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: _get_draft_token_ids is also used in _select_draft_token_ids(). I derived it to avoid nested jitted function.

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: _prepare_draft_inputs_in_jit is also used in prepare_inputs which is not jitted.

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tests/e2e/test_speculative_decoding.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +214,135 @@ def prepare_inputs(
...
         new_seq_lens_cpu = seq_lens_cpu - nrt_cpu
 
+        return self._prepare_inputs_in_jit(token_indices_cpu,
+                                           new_query_start_loc_cpu,
+                                           new_seq_lens_cpu, input_ids,
+                                           aux_hidden_states, attn_metadata,
+                                           next_token_ids, block_tables)
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _prepare_inputs_in_jit(
```
py4: The jitted function should work with device arrays.
I think we should do the `device_array` outside of `_prepare_inputs_in_jit`. Basically this function should get some arrays on TPU and produce the inputs on TPU again.
there should not be any cpu argument as it's input

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +214,135 @@ def prepare_inputs(
...
+
+        target_hidden_states, input_ids, last_token_indices, block_tables = self._prepare_draft_inputs(
+            target_hidden_states, query_start_loc, target_token_ids,
+            next_token_ids, block_tables)
+
+        attn_metadata = replace(attn_metadata, block_tables=block_tables)
+        return target_hidden_states, input_ids, last_token_indices, attn_metadata
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _prepare_draft_inputs_in_jit(
```
py4: There is no need for this seperate jit function. We can just do @jax.jit above `_prepare_draft_inputs`

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +214,135 @@ def prepare_inputs(
...
+            query_start_loc, target_token_ids, next_token_ids,
+            self.runner.input_batch.num_reqs)
+        # NOTE(pooyam): For now, we don't support multimodal.
+
+        block_tables = block_tables.reshape(-1)
+
+        return target_hidden_states, input_ids, last_token_indices, block_tables
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _select_draft_token_ids_in_jit(
```
py4: ditto. We can just @jax.jit _select_draft_token_ids instead of creating a new one

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +214,135 @@ def prepare_inputs(
...
+
+    @functools.partial(jax.jit, static_argnums=(0, ))
+    def _select_and_stack_draft_token_ids_in_jit(
+        self,
+        hidden_states: jax.Array,
+        last_token_indices: jax.Array,
+    ) -> jax.Array:
+        draft_token_ids = self._select_draft_token_ids(hidden_states,
+                                                       last_token_indices)
+        return jnp.stack([draft_token_ids], axis=1)
```
py4: Isn't this no-op? If yeah, the entire `_select_and_stack_draft_token_ids_in_jit` is redundant

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -262,45 +351,26 @@ def propose(
...
         draft_token_ids_list = [draft_token_ids]
-        # Early exit if there is only one draft token to be generated.
-        if self.num_speculative_tokens == 1:
-            return kv_caches, self._stack_draft_token_ids(draft_token_ids_list)
 
-        positions = self.runner._select_from_array_fn(
-            attn_metadata.input_positions, last_token_indices)
-        hidden_states = self.runner._select_from_array_fn(
-            residual[0], last_token_indices)
+        positions, hidden_states = self._select_positions_and_hidden_states_in_jit(
```
py4: We can do these in a single jit call:

```
        draft_token_ids = self._select_draft_token_ids_in_jit(
            hidden_states, last_token_indices)

        draft_token_ids_list = [draft_token_ids]

        positions, hidden_states = self._select_positions_and_hidden_states_in_jit(
            attn_metadata.input_positions, residual[0], last_token_indices)
```

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: No, the stack operation will change the shape.

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Done

tpu_inference/spec_decode/jax/eagle3.py
py4: it seems like there is only one jax.array in the list. in that case what does the stack do ?

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +214,117 @@ def prepare_inputs(
...
+        block_tables = block_tables.reshape(-1)
+
+        return target_hidden_states, input_ids, last_token_indices, block_tables
+
+    def _select_draft_token_ids(
+        self,
+        hidden_states: jax.Array,
+        last_token_indices: jax.Array,
+    ) -> jax.Array:
+        sample_hidden_states = hidden_states[last_token_indices]
```
py4, line 288: This is working with two TPU arrays and should be merged with another jitted boundary

py4 (commented):
Let's fix the acceptance rate issue to make sure we are not introducing regression

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: E.g. draft_token_ids with shape (3, ). After the stack operation, the shape becomes (3, 1)

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -159,13 +148,22 @@ def prepare_inputs(
...
+        block_tables = self.runner.input_batch.block_table[
+            draft_kv_cache_group_id].get_device_tensor()
         # Number of active requests in this step (un-padded count).
         num_reqs = self.runner.input_batch.num_reqs
 
+        if num_rejected_tokens is None:
+            target_hidden_states, input_ids, last_token_indices, block_tables = self._prepare_draft_inputs_in_jit(
+                self._concate_hidden_states(aux_hidden_states),
+                attn_metadata.query_start_loc, input_ids, next_token_ids,
+                block_tables, jnp.asarray([num_reqs], dtype=jnp.int32))
```
py4: I think the `jnp.asarray([num_reqs], dtype=jnp.int32)` should not be passed to the jitted function because it's not on TPU. We should first put it on TPU and then pass it

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -217,43 +215,110 @@ def prepare_inputs(
...
+        new_seq_lens_cpu = seq_lens_cpu - nrt_cpu
+
+        query_start_loc, seq_lens, token_indices = device_array(
+            self.mesh,
+            (new_query_start_loc_cpu, new_seq_lens_cpu, token_indices_cpu))
+
+        return self._prepare_inputs_in_jit(
+            token_indices, query_start_loc, seq_lens, input_ids,
+            aux_hidden_states, attn_metadata, next_token_ids, block_tables,
+            jnp.asarray([num_reqs], dtype=jnp.int32))
```
py4: ditto. We can put on TPU and reuse

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -55,22 +55,17 @@ def load_model(self, target_model: Any) -> None:
         self.state.model.embed_tokens = target_model.model.embed
 
     @functools.partial(jax.jit, static_argnums=(0, ))
-    def _concate_hidden_states(self, aux_hidden_states):
+    def _concate_hidden_states(
```
py4: 1. `concate` -> `concat`
2. This function is used just in `_prepare_draft_inputs_in_jit`. Can we move this one line logic into that function and get rid of this function?

py4:
Can we also rename some of these functions? We have so many `prepare_inputs_" names. It's a bit difficult to understand what's the difference between `prepare_inputs` and `prepare_inputs_in_jit` for instance. Can we come up with more concrete names? if that makes sense

Lumosis:
> Can we also rename some of these functions? We have so many `prepare_inputs_" names. It's a bit difficult to understand what's the difference between `prepare_inputs`and`prepare_inputs_in_jit` for instance. Can we come up with more concrete names? if that makes sense

Done

PR #924, opened 2025-10-23
[Feature] Code implementation of Async Scheduler

cychiuak:
feat(scheduler): Implement asynchronous scheduler to reduce CPU wait overhead

Problem:
The existing synchronous scheduler blocks the CPU while waiting for TPU
results. This wait time creates significant overhead, especially for
small models or decode-heavy scenarios where the scheduler is invoked
frequently.

Solution:
This commit introduces an asynchronous scheduler that decouples the CPU
scheduler from the TPU execution. The CPU no longer blocks and can
continue processing, effectively eliminating the wait-time overhead.

Impact:
Observed a ~33% throughput increase on Llama3.2-1B. This change is
most significant for workloads with high scheduler call frequency.

Benchmark (Llama3.2-1B):

Before: 6.38 req/s (Sync @ https://github.com/cychiuak/tpu-inference/commit/8c7e7bbaff700a2aea1ccae3f034994e33f89b75)
After: 8.89 req/s (Async)
How to Reproduce:
```
HF_TOKEN=<YOUR_HF_TOKEN> VLLM_SERVER_DEV_MODE=1 VLLM_USE_V1=1 \
VLLM_TORCH_PROFILER_DIR=/dev/shm/vllm-prof \
python -m vllm.entrypoints.cli.main serve meta-llama/Llama-3.2-1B \
--max-model-len 800 --max-num-seqs 256 --max-num-batched-tokens 512 \
--no-enable-prefix-caching --tensor-parallel-size 1 --async-scheduling
```
```
python -m vllm.entrypoints.cli.main bench serve --backend vllm \
--model meta-llama/Llama-3.2-1B --dataset-name random --random-input-len 230 \
--random-output-len 120 --num-prompts 200 --max-concurrency 3 \
--percentile-metrics "ttft,tpot,itl,e2el" --ignore-eos
```

### DIFF
--- a/tpu_inference/runner/tpu_jax_runner.py
+++ b/tpu_inference/runner/tpu_jax_runner.py
@@ -146,25 +146,12 @@ def _substitute_placeholder_token(
         f"Shape mismatch: input_ids and index arrays must have identical shapes due to precompilation assumptions. " \
         f"Got: {input_ids.shape=}, {token_in_tpu_cur_input_indices.shape=}, {token_in_tpu_pre_next_tokens_indices.shape=}"
 
-    def updated_input_ids_array(i: int,
-                                current_input_ids: jax.Array) -> jax.Array:
-        """
-        Iteratively updates the input_ids for all placeholders.
-å
-        Args:
-            i: The current loop index.
-            current_input_ids: The loop carry state (the input_ids being modified).
-
-        Returns:
-            The updated input_ids array.
-        """
-        update_idx = token_in_tpu_cur_input_indices[i]
-        value_idx = token_in_tpu_pre_next_tokens_indices[i]
-        new_token_value = next_tokens[value_idx]
-        return current_input_ids.at[update_idx].set(new_token_value)
-
-    return jax.lax.fori_loop(0, placeholder_num, updated_input_ids_array,
-                             input_ids)
+    # updates the input_ids for all placeholders.
+    mask = jnp.arange(input_ids.shape[0]) < placeholder_num
+    new_token_values = next_tokens[token_in_tpu_pre_next_tokens_indices]
+    original_values = input_ids[token_in_tpu_cur_input_indices]
+    update_values = jnp.where(mask, new_token_values, original_values)
+    return input_ids.at[token_in_tpu_cur_input_indices].set(update_values)
 
 
 class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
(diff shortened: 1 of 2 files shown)


py4:
Tnx for the PR. Please add e2e CI test for this feature to prevent feature from breaking

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -80,6 +82,91 @@
 }
 
 
+class AsyncTPUModelRunnerOutput(AsyncModelRunnerOutput):
```
py4, line 85: Please add e2e tests that makes sure:
- correctness is not changed
- performance improvement is observed when this feature is enabled

you can take a look at spec decoding e2e tests that tests for both correctness and performance:

https://github.com/vllm-project/tpu-inference/blob/main/tests/e2e/test_speculative_decoding.py

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -80,6 +82,91 @@
...
+
+        Returns:
+            The updated input_ids array.
+        """
+        update_idx = token_in_tpu_cur_input_indices[i]
+        value_idx = token_in_tpu_pre_next_tokens_indices[i]
+        new_token_value = next_tokens[value_idx]
+        return current_input_ids.at[update_idx].set(new_token_value)
+
+    return jax.lax.fori_loop(0, placeholder_num, updated_input_ids_array,
```
jrplatin: Any way this can be parallelized? Like:

```
input_ids.at[update_indices].set(new_token_values)
```

yuyanpeng-google: It's because of the input shape. To prevent update_indices be any length from 1 to len(current_input_ids), we pad the update_idx to len(current_input_ids) and use fori_loop a dynamic shape.

py4: @yuyanpeng-google Can we use a simple mask to mask items that don't need to be updated and then do a single scatter updated like what @jrplatin said?
the index and update size has length of input_ids. So we can mask the last `len - num_placeholder` easily.
A single scatter update should be more performant than the sequential loop

bvrockwell:
Great contribution @yuyanpeng-google !! Excited to see it in action.

tpu_inference/runner/tpu_jax_runner.py
py4: I see that now e2e tests are added. Thanks! Can you please add them in https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/pipeline_jax.yml so they are actually called in CI?

PR #930, opened 2025-10-23
Reduce the host overhead for LoRA

vanbasten23:
# Description

This PR optimizes the host perf of LoRA.

On TPU-inference, the current step time is [130 ms](https://screenshot.googleplex.com/5JPABWpvnJKRxvK) and the execution time is [5.59 ms](https://screenshot.googleplex.com/65xzDf8fffggfWY). So why does it take [so long between  each execution](https://screenshot.googleplex.com/QFTZ99TKNJjCuhA)? I discovered that `shard_model_to_tpu` takes  94% of the step time.

This PR optimize `shard_model_to_tpu` and only do the what is necessary for lora:

During inference, before each fwd, LoRA needs to call `shard_model_to_tpu` because the linear.lora matrices may be updated. `shard_model_to_tpu` does 3 things essentially: shard the module and move to TPU (via _shard_module_to_tpu), extract params and buffers, and move the remaining params/buffers to TPU (via PyTree.tree_map_only). In the [profile](https://screenshot.googleplex.com/6VqsrSFXvgKGog7), both `_shard_module_to_tpu` and `PyTree.tree_map_only` take equally excessive time. This is what this PR proposes: 

1. I don't need to re-shard every time. When vllm [updates the lora](https://github.com/vllm-project/vllm/blob/0552cfb195b24a490f253d485de11f2b94b0c6b8/vllm/lora/layers/column_parallel_linear.py#L259-L267), the sharding still remains the same and the new lora is moved from CPU to TPU so the device also remains the same  (both via `_copy`). That means we only need to shard once at first when we load the model and don't need to re-shard later. (I also wrote a torchax [repro](https://gist.github.com/vanbasten23/64be65a8331dae95e95b4a9d90214691) that demonstrates `copy_` won't change the tensor's sharding)

2. 'PyTree.tree_map_only' is only used to move the remaining params/buffers to TPU. But for lora, the only possibly updated params are `lora` related matrices and we know the lora matrices are already torchax tensor on TPU. So there is really no need to do this tree_map_only step.

3. That said, the only necessary thing for the LoRA case is to re-extract the parameter for lora and update the original `params_buffers`.

With the change, I profiled again and the step time has reduced from 130ms to [13ms](https://screenshot.googleplex.com/9xF9vk9jDkHeqSg)

# Tests

CI

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/sharding.py
+++ b/tpu_inference/layers/vllm/sharding.py
@@ -66,15 +66,11 @@ def _extract_all_params_buffers(model: torch.nn.Module):
 
 
 def _tensor_is_in_cpu(tensor: torch.tensor) -> bool:
-    # xw32: There are many torch.nn.parameter.Parameter. e.g. https://paste.googleplex.com/4571608492670976
     # Check if a tensor haven't been converted to torchax tensor.
     if not isinstance(tensor, torchax.tensor.Tensor):
         return True
     # Check if torchax tensor is still in CPU.
-    if tensor.jax_device == jax.devices('cpu')[0]:
-        return True
-    else:
-        return False
+    return tensor.jax_device == jax.devices('cpu')[0]
 
 
 def _convert_to_torchax_and_shard(tensor: torch.Tensor,
--- a/examples/offline_inference.py
+++ b/examples/offline_inference.py
@@ -14,9 +14,8 @@ def create_parser():
     parser = FlexibleArgumentParser()
     # Add engine args
     EngineArgs.add_cli_args(parser)
-    parser.set_defaults(model="Qwen/Qwen2.5-3B-Instruct")
-    parser.set_defaults(max_model_len=256)
-    parser.set_defaults(max_num_seqs=8)
+    parser.set_defaults(model="meta-llama/Llama-3.2-1B-Instruct")
+    parser.set_defaults(max_model_len=1024)
 
     # Add sampling params
     sampling_group = parser.add_argument_group("Sampling parameters")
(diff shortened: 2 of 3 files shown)


vanbasten23:
cc: @qihqi @kyuyeunk

tpu_inference/layers/vllm/sharding.py
```diff
@@ -51,6 +49,18 @@ def shard_model_to_tpu(model: torch.nn.Module,
         return {**params, **buffers}
 
 
+def update_lora(model: torch.nn.Module,
+                initial_params_buffers) -> dict[str, torchax.torch.Tensor]:
+    params, buffers = _extract_all_params_buffers(model)
+    params_buffers = {**params, **buffers}
+    for k, v in params_buffers.items():
+        if 'lora_a_stacked' in k or 'lora_b_stacked' in k:
```
kyuyeunk, line 57: Instead of string based approach, can you opt for layer type checking approach instead?

Meaning, you iterate through layers and if a layer matches a specific type (i.e., lora layer), you'll check for `lora_a_stacked` and `lora_b_stacked` and if that doesn't exists, throw error because that's should not happen.

I think it will have added benefit of ensuring all lora layers are accounted for.

vanbasten23: If lora is enabled, all the linear layer will be replaced by the corresponding lora layer.

Actually, all the lora layers in vllm are guaranteed to have the `lora_a_stacked` and `lora_b_stacked`. That is where each lora layer keeps the lora matrices. So `_extract_all_params_buffers ` already guarantees that we account for all lora layers and we get all the updated lora matrices. 

Then here, I need to make sure I get all the possibly updated lora matrices and update the `params_buffers`. Then I can skip the expensive `tree_map_only`.

vanbasten23:
Thanks for the review.

The CI is green https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4711#_. The failure is irrelevant to this PR as they also fail in head: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4711#_

tpu_inference/layers/vllm/sharding.py
vanbasten23: Feel free to let me know if you have other concerns, Kyuyeun.

PR #932, opened 2025-10-23
Enable multi-host P/D and adopt the vllm distributed executor changes

mrjunwan-lang:
# Description

Enable the multi-host P/D in ray setup.
Adopt the vllm distributed executor changes.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/executors/ray_distributed_executor.py
+++ b/tpu_inference/executors/ray_distributed_executor.py
@@ -160,7 +160,7 @@ class RayDistributedExecutor(RayDistributedExecutorV1):
         current_placement_group = ray.util.placement_group(
             placement_group_specs, strategy="PACK")
         _wait_until_pg_ready(current_placement_group)
-        
+
         assert current_placement_group is not None
         # Set the placement group in the parallel config
         self.parallel_config.placement_group = current_placement_group


tpu_inference/distributed/tpu_connector.py
```diff
@@ -570,7 +576,12 @@ def _prepare_kv_and_wait(self, req_id: str, req_meta: SendMeta):
         self.kv_transfer_server.await_pull(req_meta.uuid, kv)
 
     def _maybe_build_kv_connection(self, req_meta: LoadMeta) -> Any:
-        remote_addr = f"{req_meta.remote_host}:{req_meta.remote_port}"
+        if isinstance(req_meta.remote_host, list):
+            assert len(req_meta.remote_host) == len(req_meta.remote_port)
+            remote_addr = f"{req_meta.remote_host[self.node_id]}:{req_meta.remote_port[self.node_id]}"
```
vanbasten23, line 581: What's this self.node_id?

vanbasten23 (approved):

tpu_inference/distributed/tpu_connector.py
mrjunwan-lang: that's the identity of each node. in the prefill and decode we has the same setup so the node_id will be 1:1 match

PR #939, opened 2025-10-24
[do not review yet] Get rid of a2a

vanbasten23:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -132,12 +132,12 @@ class VllmUnquantizedLinearMethod(UnquantizedLinearMethod):
 
         outs = slice_sharded_tensor_for_concatenation(
             outs, self.jax_config.output_sizes, self.jax_config.n_shards)
-        with jax.named_scope(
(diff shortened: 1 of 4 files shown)


vanbasten23:
`tests/lora/test_layers.py::test_column_parallel_packed[False-1-1]` test (aka the case `QKVParallelLinearWithLoRA(QKVParallelLinear)`) fails due to
```
TypeError: unsupported operand type(s) for +: 'Tensor' and 'Tensor'
```
The error is confusing because the real error is not about `+` is not support. Instead, both operands have different shape. The reason is for this case, QKVParallelLinear generate output (tuple of length 3) and QKVParallelLinearWithLoRA.stacked_lora_b (tuple of length 1).

NB: if you get torch.concatenate is not support in torchax, do something like
```
final_outputs = [(yi + outsi).jax() for yi, outsi in zip(y, lora_outputs)]
final_output = torch_view(jnp.concatenate(final_outputs, axis=-1))
```
by converting to jax and then use jnp.concatenate.

Issue #944, opened 2025-10-27
[Bug]: `ImportError: cannot import name 'cdiv' from 'vllm.utils'` when installing from source


RobMulla:
### Your current environment

<details>
<summary>The output of commands above</summary>

```text
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version                : Could not collect
CMake version                : version 4.1.2
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0] (64-bit runtime)
Python platform              : Linux-6.8.0-1015-gcp-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : False
CUDA runtime version         : No CUDA
CUDA_MODULE_LOADING set to   : N/A
GPU models and configuration : No CUDA
Nvidia driver version        : No CUDA
cuDNN version                : No CUDA
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                         x86_64
CPU op-mode(s):                       32-bit, 64-bit
Address sizes:                        52 bits physical, 57 bits virtual
Byte Order:                           Little Endian
CPU(s):                               180
On-line CPU(s) list:                  0-179
Vendor ID:                            AuthenticAMD
Model name:                           AMD EPYC 9B14
CPU family:                           25
Model:                                17
Thread(s) per core:                   2
Core(s) per socket:                   90
Socket(s):                            1
Stepping:                             1
BogoMIPS:                             5199.99
[...]

PR #946, opened 2025-10-27
fix: Update cdiv import location for vllm compatibility

RobMulla:
**Description:**

This PR updates the import location of the `cdiv` utility function to align with recent changes in the `vllm` library.

Previously, `tpu-inference` imported `cdiv` from `vllm.utils`. This function was moved to `vllm.utils.math_utils` in `vllm`, causing an `ImportError` when attempting to build `tpu-inference` from source against recent versions of `vllm`. This change updates the import path in the affected files, resolving the error and restoring the ability to build from source.

FIXES: #944 

**Tests:**

Verified that with this change, the `vllm serve` command starts successfully without the `ImportError: cannot import name 'cdiv' from 'vllm.utils'` error when using a version of `vllm` where `cdiv` has been moved.

**Checklist:**

- [ x] I have performed a self-review of my code.
- [ x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ x] I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -7,7 +7,6 @@ import jax.numpy as jnp
 import numpy as np
 import vllm.envs as envs
 from jax.sharding import NamedSharding, PartitionSpec
-from vllm.utils.math_utils import cdiv
 
 from tpu_inference.core.disagg_utils import is_disagg_enabled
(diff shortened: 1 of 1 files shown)


jrplatin:
Fixed [here](https://github.com/vllm-project/tpu-inference/pull/947), thanks!

PR #952, opened 2025-10-27
[MMLU] Add chat-template support for MMLU

bzgoogle:
# Description

Enable support of chat-template for tuned models, i.e. Deepseek, GPT-OSS etc

# Tests

On 1000 prompts from MMLU on GPT-OSS-120B: 
Without chat-template, the accuracy is [0.26](https://paste.googleplex.com/5632136262582272#l=38) for GPT-OSS model;
With chat-temaplte, the accuracy is [0.87](https://paste.googleplex.com/6178518731980800#l=38) for GPT-OSS model;

It won't change the legacy MMLU benchmarking results on other models with chat-template disabled(default)
Qwen3-32B has accuracy of [0.813](https://paste.googleplex.com/6139516050079744#l=40)

# Checklist

### DIFF
--- a/scripts/vllm/benchmarking/benchmark_dataset.py
+++ b/scripts/vllm/benchmarking/benchmark_dataset.py
@@ -201,7 +201,8 @@ class MMLUDataset(BenchmarkDataset):
     https://github.com/AI-Hypercomputer/JetStream/blob/bbfb5bd/benchmarks/benchmark_serving.py#L327.
     """
 
-    def __init__(self, num_shots: int, mmlu_method: str, use_chat_template: bool, **kwargs) -> None:
+    def __init__(self, num_shots: int, mmlu_method: str,
+                 use_chat_template: bool, **kwargs) -> None:
         super().__init__(**kwargs)
         self.mmlu_method = mmlu_method
         self.num_shots = num_shots
@@ -304,27 +305,21 @@ class MMLUDataset(BenchmarkDataset):
                 break
 
             if self.use_chat_template:
-                messages = [
-                    {
-                        "role": "system",
-                        "content": "Reasoning effort: high"
-                    },
-                    {
-                        "role": "user", 
-                        "content": prompt
-                    }
-                ]
+                messages = [{
+                    "role": "system",
+                    "content": "Reasoning effort: high"
+                }, {
+                    "role": "user",
+                    "content": prompt
+                }]
 
                 try:
                     prompt = tokenizer.apply_chat_template(
-                        messages,
-                        tokenize=False,
-                        add_generation_prompt=True
-                    )
+                        messages, tokenize=False, add_generation_prompt=True)
                 except Exception as e:
                     logger.error(f"Could not apply chat template: {e}. "
-                                "Falling back to raw prompt. "
-                                "This will likely fail for fine-tuned models")
+                                 "Falling back to raw prompt. "
+                                 "This will likely fail for fine-tuned models")
 
             prompt_ids = tokenizer(prompt).input_ids
             completion_ids = tokenizer(completion).input_ids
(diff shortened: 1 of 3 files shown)


vipannalla:
LGTM

vipannalla (approved):

scripts/vllm/benchmarking/benchmark_utils.py
```diff
@@ -102,11 +102,49 @@ def postprocess_text_mmlu(preds: List[str],
...
 
     def _parse_answer(output):
-        # ? marks the close parenthesis as optional
-        re_str = r"\s*\(([A-D])\)?\s*\w*"
-        match = re.search(re_str, output, re.IGNORECASE)
-        predicted_answer = match.group(1).upper() if match else None
-        return predicted_answer
+        """
+        Parses the MMLU answer from a gpt-oss model's raw output.
+        Prioritizes the 'assistantfinal' block.
```
jrplatin: Can you add a some quick context that this for GPT-OSS specifically? Otherwise, most folks don't know what `assistantfinal` is

scripts/vllm/benchmarking/benchmark_dataset.py
```diff
@@ -201,10 +201,11 @@ class MMLUDataset(BenchmarkDataset):
     https://github.com/AI-Hypercomputer/JetStream/blob/bbfb5bd/benchmarks/benchmark_serving.py#L327.
     """
 
-    def __init__(self, num_shots: int, mmlu_method: str, **kwargs) -> None:
+    def __init__(self, num_shots: int, mmlu_method: str, use_chat_template: bool, **kwargs) -> None:
```
jrplatin: Can you rename this arg / the argparser arg to something like `use-gpt-oss-chat-template` or something? To me, it doesn't seem like something that'll necessarily generalize to all models

scripts/vllm/benchmarking/benchmark_serving.py
```diff
@@ -868,6 +869,11 @@ def main(args: argparse.Namespace):
         choices=["HELM", "Harness", ""],
         help="mmlu method/format to generate shots",
     )
+    mmlu_group.add_argument(    
+        "--mmlu-use-chat-template",
+        action="store_true",
+        help="Whether to format MMLU prompts using the tokenizer's chat template.",
```
jrplatin, line 875: If you decide to rename this to `gpt-oss-chat-template` or something, please update this description str

scripts/vllm/benchmarking/benchmark_dataset.py
bzgoogle: I think we may need chat-template for other tuned-models, i.e. DeepSeek,  not only for gpt-oss purpose

vipannalla: Chat templates are generic and can be applied to many models, this shouldn't be gpt-oss specific.

jrplatin: Sure, but in this particular case, we're using a chat-template that's specific to OSS, for example:


-  `                output += "\nPlease think carefully and give the answer. "`
- "content": "Reasoning effort: high"

bzgoogle: We could fine-tune this prompt over models, but I think it's not to gpt-oss specific, it's part of prompt-enginerring to trigger reasoning

scripts/vllm/benchmarking/benchmark_serving.py
bzgoogle: similar as above, the chat-template is aimed for all fine-tuned model, not gpt-oss only

scripts/vllm/benchmarking/benchmark_utils.py
bzgoogle: Updated the docstring with todos

PR #981, opened 2025-10-30
Add lora layer tests

vanbasten23:
# Description

This PR adds unit tests for the `MergedColumnParallelLinearWithLoRA` lora wrapper.

# Tests

pytest -s -vv  tests/lora/test_layers.py

# Checklist

### DIFF
--- a/tests/lora/test_layers.py
+++ b/tests/lora/test_layers.py
@@ -202,6 +202,8 @@ def create_random_inputs(
 @pytest.mark.parametrize("repeats", [2])
 @pytest.mark.parametrize("stage", [True, False])
 def test_column_parallel_packed(dist_init, num_loras, repeats, stage) -> None:
+    set_random_seed(6)
+
     max_loras = 9
     max_lora_rank = 8
     lora_config = LoRAConfig(
@@ -213,24 +215,12 @@ def test_column_parallel_packed(dist_init, num_loras, repeats, stage) -> None:
     vllm_config = dist_init
     vllm_config.lora_config = lora_config
 
-    axis_names = ("data", "model")
-    devices = jax.devices()
-    mesh_shape = (1, len(devices))
-    mesh = jax.make_mesh(mesh_shape, axis_names, devices=devices)
-
-    set_random_seed(6)
-
+    mesh = _create_mesh()
     linear, lora_linear = _create_column_parallel_packed_layer(
         repeats, vllm_config, mesh)
-    with torchax.default_env():
-        # lora_linear.weight has type torchax.tensor.Tensor
-        # BaseLinearLayerWithLoRA.weight property guarantees this.
-        # if len(devices) != 1, `reorder_concatenated_tensor_for_sharding` function may reorder the out_features dimension of the weight matrix.
-        # So the below check will fail.
-        if len(devices) == 1:
-            assert torch.equal(linear.weight.data,
-                               lora_linear.weight.to('cpu'))
+    _verify_lora_linear_layer(linear, lora_linear)
 
+    # Create a punica wrapper and associate it with the lora linear layer.
     max_num_batched_tokens = 8192
     max_batches = 256
     with torchax.default_env():
@@ -251,6 +241,7 @@ def test_column_parallel_packed(dist_init, num_loras, repeats, stage) -> None:
         repeats=repeats,
     )
 
+    # Create inputs and lora mappings.
     # inputs: list[torch.Tensor] of size num_inputs. inputs[i] corresponds to a request which has several token of shape=[num_tokens, 64].
     # index_mapping: list[int]
     # prompt_mapping: list[int]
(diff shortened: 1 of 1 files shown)


vanbasten23:
cc: @qihqi @kyuyeunk @hfan @aman2930

tests/lora/utils.py
kyuyeunk, line 1: Is there are way to utilize existing implementation of `DummyLoRAManager` instead of creating a new one here?

tpu_inference/lora/torch_punica_tpu.py
```diff
@@ -281,7 +283,8 @@ def _update_prefill_metadata(self,
             self.batch_size = 1
             self._lora_indices_per_batch[:self.
                                          batch_size] = token_lora_tensor[:self.
-                                                                         batch_size]
+                                                                         batch_size].torch(
```
kyuyeunk, line 286: Is the formatting for this line correct? Looks really weird?

tests/lora/test_layers.py
kyuyeunk, line 1: There a lot of logics here which makes me wonder: Is there a similar test / logics on vLLM repo? If so, can you share a link so I can reference it?

vanbasten23: Sure, it's here https://github.com/vllm-project/vllm/blob/main/tests/lora/test_layers.py.

tests/lora/utils.py
vanbasten23: Thanks for the review. I sought to directly import the existing [implementation](https://github.com/vllm-project/vllm/blob/c765f0b443c2de886a5771efcbc154ee84c2c19d/tests/lora/utils.py#L14) as `from vllm.tests.lora.utils import DummyLoRAManager` but it didn't work.

tpu_inference/lora/torch_punica_tpu.py
vanbasten23: haha yeah. iirc, I ran the formatter on the latest commit. I can run again to double check.

vanbasten23: I ran the formatter again "pre-commit run --all-files" and the format remains the same.

tests/lora/utils.py
kyuyeunk: I just took a look at it and it seems almost identical? Can you elaborate why it didn't work & what changes you needed to make?

tests/lora/test_layers.py
```diff
@@ -0,0 +1,421 @@
...
+        fully_sharded_loras=False,
+        lora_dtype=torch.float16,
+    )
+
+    axis_names = ("data", "model")
+    devices = jax.devices()
+    mesh_shape = (1, len(devices))
+    mesh = jax.make_mesh(mesh_shape, axis_names, devices=devices)
+
+    def create_column_parallel_packed_layer():
```
kyuyeunk: can you move this function outside of `test_column_parallel_packed`? It hurts readability on what this unit test is supposed to run.

Please reference this unit test implementation where majority of initialization functions are implemented outside of the unit test function: https://github.com/vllm-project/tpu-inference/blob/330cb1be949d4d20188427a6faa3440453e11e3a/tests/layers/vllm/test_awq.py#L238-L266

tests/lora/test_layers.py
```diff
@@ -0,0 +1,421 @@
...
+        prompt_mapping += [lora_id]
+
+    return inputs, index_mapping, prompt_mapping
+
+
+@torch.inference_mode()
+@pytest.mark.parametrize("num_loras", [1, 2, 4, 9])
+@pytest.mark.parametrize("repeats", [2])
+@pytest.mark.parametrize("stage", [True, False])
+def test_column_parallel_packed(dist_init, num_loras, repeats, stage) -> None:
```
kyuyeunk, line 204: This unit test is really large and hard to understand what this unit test function is doing. Can you separate out all the initialization, inference, assertion, related logics into a separate function and maybe create a separate unit test function for it as well?

tests/lora/utils.py
vanbasten23: Yes, they are identical. It didn't work because the DummyLoRAManager is defined in the vllm/tests and direct importing it via `from vllm.tests.lora.utils import DummyLoRAManager` fails due to `ModuleNotFoundError: No module named 'vllm.tests'`.

I agree with you that we'd better utilize the existing implementation. I think the way to go is to modify vllm and move DummyLoRAManager from vllm/test to vllm/vllm so that we can directly import in TPU-inference. Wdyt?

vanbasten23: Made a upstream change https://github.com/vllm-project/vllm/pull/28313. Once it's in, I can remove this file.

tests/lora/test_layers.py
vanbasten23: done.

tests/lora/test_layers.py
vanbasten23: Yeah I get your point. I spent some time and separated out lots of logic. Now the test is much thinner and hope it's easier to understand.

tests/lora/test_layers.py
```diff
@@ -0,0 +1,430 @@
...
+    expected_result = torch.cat(expected_results)
+
+    rtol, atol = TOLERANCES[actual_result.dtype]
+    with torchax.default_env():
+        actual_result_cpu = actual_result.to('cpu')
+        torch.testing.assert_close(actual_result_cpu,
+                                   expected_result,
+                                   rtol=rtol,
+                                   atol=atol)
+        # print(
```
kyuyeunk, line 283: remove?

tests/lora/utils.py
kyuyeunk: got it. that makes sense. thanks for confirming it!

tests/lora/test_layers.py
kyuyeunk: it's still quite large but it's definitely better than before. thanks for making the change!

vanbasten23:
The CI https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5099#_ is green. The faililng tests are irrelevant to this PR.

Thanks for the review!

tests/lora/test_layers.py
vanbasten23: mind if I leave it as I find it helpful to debug the test. Sometimes, the tests fails because atol/rtol is not set properly.

PR #985, opened 2025-10-31
[Feature] Add automated PyPI publishing workflow

ylangtsou:
# Description

This commit introduces a complete GitHub Actions workflow for automatically building and publishing the Python package to PyPI.

- Adds the `release.yml` workflow, supporting multiple triggers: Tag push and Schedule.
- Adds the `determine_release_vars.sh` script, which contains the core versioning logic.

# Tests

[github](https://github.com/vllm-project/tpu-inference/actions/runs/18956645722)

# Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,63 @@
+name: Publish Package to PyPI
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+  schedule:
+      - cron: '0 8 * * *'
+
+jobs:
+  pypi_publish:
+    name: Build and Publish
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
+name: Publish Package to PyPI
```
jcyang43, line 1: Your current setup will lead to triggering the release workflow even if the tag is created on a non-main branch. 

We only want the release tag to go off if it's created on the protected (main) branch. This is required to ensure only admins or selected people can trigger a stable release via release tag

ylangtsou: add Check if tag is on main branch

PR #992, opened 2025-10-31
[GPT-OSS] Load MXFP4 and BF16 weights directly and enable online requantization

amishacorns:
# Description

Adds MXFP4 load support  to enable loading original OpenAI checkpoints, e.g. openai/gpt-oss-120b. Includes another swap axes since the original openai checkpoints come with [output, input] dimensions, while previous BF16 checkpoints have [input, output]. Requantization can be performed on the weights after load with an addition-config passed for both the MXFP4 and BF16 models.

The sharding had to be passed to RMSNorm to avoid Qwix-related issues, as well as wrapping with PartitionSpecs. The expert combine op was also factored out of the the MoE forward directly to allow Qwix to ignore it. The alternative is not using an einsums and filtering the quantized modules to be einsum, but currently if the op is the same type and within the same module, it doesn't seem like Qwix can target that specific op.

# Tests

I evaluated the accuracy using scripts/vllm/benchmarking/benchmark_serving.py for MMLU across the four variants, MXFP4, BF16, BF16 requantized (64 size subchannel), and MXFP4 requantized. All give similar accuracy on 10% of the MMLU benchmark.

### DIFF
--- a/tpu_inference/layers/jax/moe/gpt_oss_moe.py
+++ b/tpu_inference/layers/jax/moe/gpt_oss_moe.py
@@ -68,6 +68,34 @@ def _swiglu(x: Float, alpha: Float, limit: Float) -> Float:
 
 
 @dataclass(kw_only=True)
+class CombineExperts(nnx.Module):
+    """Module for combining expert outputs with weighted sum."""
+    dtype: jnp.dtype
+
+    def __call__(self, down_proj_TED: Float, weights_TX: Float,
+                 indices_TX: jax.Array) -> Float:
+        """Combines expert outputs using weighted sum.
+
+        Args:
+            down_proj_TED: Expert outputs, shape (tokens, experts, hidden_dim)
+            weights_TX: Router weights, shape (tokens, experts_per_token)
+            indices_TX: Selected expert indices, shape (tokens, experts_per_token)
+
+        Returns:
+            Combined output, shape (tokens, hidden_dim)
+        """
+        with jax.named_scope("combine_experts"):
+            indices_for_gather = indices_TX[..., None]
+            gathered_down_proj_TED = jnp.take_along_axis(down_proj_TED,
+                                                         indices_for_gather,
+                                                         axis=1)
+            output_TD = jnp.einsum('TXD,TX -> TD', gathered_down_proj_TED,
+                                   weights_TX)
+
+        return output_TD.astype(self.dtype)
+
+
+@dataclass(kw_only=True)
 class GptOssMoE(nnx.Module):
     """
     JAX implementation of the GPT-OSS Mixture-of-Experts MLP block.
@@ -114,19 +142,15 @@ class GptOssMoE(nnx.Module):
             down_proj_TED += self.mlp2_bias_ED.value
 
         # Weighted sum of expert outputs
-        with jax.named_scope("sum"):
-            indices_for_gather = indices_TX[..., None]
-            gathered_down_proj_TED = jnp.take_along_axis(down_proj_TED,
-                                                         indices_for_gather,
-                                                         axis=1)
-            output_TD = jnp.einsum('TXD,TX -> TD', gathered_down_proj_TED,
-                                   weights_TX)
+        output_TD = self.combine_experts(down_proj_TED, weights_TX, indices_TX)
 
-        return output_TD.astype(self.dtype)
+        return output_TD
 
     def __post_init__(self, rngs: nnx.Rngs):
         """Initializes all weights and biases for the MoE block."""
         D, F, E = self.hidden_size, self.intermediate_size_moe, self.num_local_experts
+
+        self.combine_experts = CombineExperts(dtype=self.dtype)
 
         # MLP #1 Weights (Combined Gate and Up-projection) and Bias
         self.mlp1_weight_EDF2 = create_param(


kyuyeunk:
@bzgoogle can you take a look?

tpu_inference/models/jax/gpt_oss.py
```diff
@@ -246,11 +250,11 @@ def load_weights(self, rng: PRNGKey, cache_dir: Optional[str] = None):
             "model.layers.*.mlp.router.bias":
             ("layers.*.custom_module.router.bias_E", None, None),
             "model.layers.*.mlp.experts.gate_up_proj":
-            ("layers.*.custom_module.mlp1_weight_EDF2", None, None),
+            ("layers.*.custom_module.mlp1_weight_EDF2", transforms["swap_last2"], None),
```
jrplatin: Do we always want to do this? I guess, why wasn't this change here before / what changed? Same comment below

tpu_inference/models/jax/gpt_oss.py
```diff
@@ -264,8 +268,44 @@ def load_weights(self, rng: PRNGKey, cache_dir: Optional[str] = None):
             framework="pt",
             download_dir=self.vllm_config.load_config.download_dir)
 
+        # Single pass: build a unified pool. Combine MXFP4 expert blocks/scales
+        # into a dequantized bf16 tensor as soon as both are seen.
+        pool: dict[str, torch.Tensor] = {}
+        pending_experts: dict[str, dict[str, torch.Tensor]] = {}
+        for loaded_name, loaded_weight in names_and_weights_generator:
```
jrplatin, line 358: Similar to above, but are we assuming that, for GPT-OSS, we are always loading in the MXFP4 checkpoint?

tpu_inference/models/jax/gpt_oss.py
```diff
@@ -264,8 +268,44 @@ def load_weights(self, rng: PRNGKey, cache_dir: Optional[str] = None):
...
+            if loaded_name.endswith("_blocks") or loaded_name.endswith("_scales"):
+                base = loaded_name[:-7]
```
jrplatin, line 361: Is this maybe too brittle?

tpu_inference/models/jax/gpt_oss.py
amishacorns: Previously this was hardcoded to the unsloth checkpoints, which for some reason had their experts transposed compared to the original gpt-oss models, and only the experts

amishacorns: This is hard to catch dynamically too since some of the expert weights are symmetric, but I assume only supported the original checkpoints weight layout

tpu_inference/models/jax/gpt_oss.py
amishacorns: Yeah good question. I assume only the original gpt-oss checkpoint would be supported, since it seems difficult to potentially accept arbitrary variants

But I can try to accept BF16 checkpoints too if helpful! It could be a little difficult if the weights are transposed like with unsloth

tpu_inference/models/jax/gpt_oss.py
amishacorns: Good point, and I agree completely if other gpt-oss checkpoints are accepted besides the original

tpu_inference/models/jax/gpt_oss.py
jrplatin: Gotcha, so basically, does this change break loading the unquantized checkpoints and is that ok?

tpu_inference/models/jax/gpt_oss.py
jrplatin: Ah just was curious -- I doubt anyone needs to load the BF16 checkpoint, but if it's the case that we only support one specific checkpoint, it might be good to add some defensive logic to ensure only that checkpoint can be loaded

tpu_inference/models/jax/gpt_oss.py
amishacorns: Yeah, to me that seems fine. This breaks loading the "dequantized" checkpoints. Currently people were using this code to load unsloth/gpt-oss-BF16, but to me that is just an arbitrary community variant on the original MXFP4 checkpoint. To my knowledge, there are no official unquantized BF16 checkpoints for gpt-oss.

kyuyeunk: hard disagree that it's okay to break existing logic. it shouldn't really matter whether a checkpoint is official or not. the code should be able to be flexible / modular enough that it can support any arbitrary checkpoint. 

Please modify the logic like below
- check "quant_method" field in config.json
  - `mxfp4`: https://huggingface.co/openai/gpt-oss-120b/blob/main/config.json#L69
  - `None` (since the field does not exist): https://huggingface.co/unsloth/gpt-oss-120b-BF16/blob/main/config.json 
- Based on the config, create branching logic that invoke different checkpoint loading code.
  -  `mxfp4`: use logic in this PR
  - `None`: use existing unquantized logic

amishacorns: Thanks. It's a slippery slope to support arbitrary and unstated community variants of popular models that contain weight transposes compared to the original. But in this case, the transpose logic seems like it can be hardcoded based on datatype, since it seems lmsys and unsloth are the most popular and share the same structure.

But I now agree existing support shouldn't be changed. This PR may want to wait then for a better support for pre-quantized models generally

amishacorns: And I'll update this tomorrow afternoon PT, and agree with all the suggestions. The weight layout is actually more standard on the BF16 checkpoints and the original MXFP4 likely just transposed to make the bit packed dimension the most minor. BF16 support will let anyone run their own models.

I'll condition on the quant_method and update here. Thanks again.

kyuyeunk: > It's a slippery slope to support arbitrary and unstated community variants of popular models

Not really. Checkpoint formats are pretty much standardized. different models with the same `quant_method` will have same checkpoint shapes.

So as long as we write things in modular way, having a function that can load popular unquantized model == having a function that can load every other unquantized model.

so it's actually extremely scalable and completely opposite of slippery slope.

PR #995, opened 2025-10-31
[Misc] Only run JAX unit tests and few e2e tests for each PR in CI.

py4:


vanbasten23:
Hi @py4 , do you know the link where we can check the nightly runs?

PR #1003, opened 2025-11-03
[Runner] Separate execute_model and sample_tokens to adapt upstream change.

py4:
Adapting to: https://github.com/vllm-project/vllm/pull/26866
The main changes are we have to separate `execute_model` and `sample_tokens`. 

BuildKite passing: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4882

### DIFF
--- a/tpu_inference/runner/tpu_jax_runner.py
+++ b/tpu_inference/runner/tpu_jax_runner.py
@@ -21,10 +21,12 @@ from vllm.forward_context import set_forward_context
 from vllm.sequence import IntermediateTensors
 from vllm.tasks import SupportedTask
 from vllm.utils.math_utils import cdiv
+from vllm.v1.core.sched.output import GrammarOutput
 from vllm.v1.core.sched.output import SchedulerOutput as VllmSchedulerOutput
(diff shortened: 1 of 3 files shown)


tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -430,9 +448,49 @@ def execute_model(
         self,
         scheduler_output: "VllmSchedulerOutput",
         intermediate_tensors: Optional[IntermediateTensors] = None,
+    ) -> ModelRunnerOutput | None:
+        if self.execute_model_state is not None:
+            raise RuntimeError("State error: sample_tokens() must be called "
+                               "after execute_model() returns None.")
```
patemotter, line 454: Nit: It's not immediately clear how sample_tokens is involved with this.

py4: You mean in terms of explanation? I copy pasted this from upstream. We can think about a better wording if you think it's not clear

patemotter: Correct, it's definitely not clear to me how it relates, but that doesn't necessarily mean a lot.

PR #1011, opened 2025-11-04
initial commit on compressed-tensors quantization support for fp8

qihqi:
# Description

This PR is to enable models with FusedMoE module (such as Qwen3-coder) to run with compressed-tensors quantization scheme.

NOTE: 
currently because of old jax version, we need to change
this line /home/hanq_google_com/uv_venv/lib/python3.12/site-packages/jax/experimental/pallas/ops/tpu/megablox/[common.py](http://common.py/): 44
which is:
```
_TPU_KIND_PATTERN = re.compile(r"TPU v(\d+)")
```
To:
```
_TPU_KIND_PATTERN = re.compile(r"TPU.*(\d+)")
```
# Tests

```
MODEL_IMPL_TYPE=vllm VLLM_DISABLE_SHARED_EXPERTS_STREAM=1 python examples/offline_inference.py --model=BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic --tensor_parallel_size=8 --task=generate --max_model_len=128 --max_num_seqs=1
```

runs and produces good result.

### DIFF
--- a/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
+++ b/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
@@ -4,7 +4,6 @@ import jax
 import jax.numpy as jnp
 import torch
 import torch.nn.functional as F
-from compressed_tensors.quantization import QuantizationStrategy
 from jax.experimental.layout import Format, Layout
 from jax.sharding import Mesh, NamedSharding
 from jax.sharding import PartitionSpec as P
@@ -31,46 +30,12 @@ class VllmCompressedTensorsW8A8Fp8MoEMethod(CompressedTensorsW8A8Fp8MoEMethod,
     def __init__(self, quant_config: "CompressedTensorsConfig",
                  moe: FusedMoEConfig, mesh: Mesh):
         super().__init__(quant_config, moe)
-
-        self.use_marlin = False
-        self.use_cutlass = False
-        self.is_fp8_w8a8_sm100 = False
-
         self.mesh = mesh
         self.quant_config = quant_config
-        # import sys
-        # sys.stdin = open(0)
-        # breakpoint()
-        self.weight_quant = self.quant_config.target_scheme_map["Linear"].get(
-            "weights")
-        self.input_quant = self.quant_config.target_scheme_map["Linear"].get(
-            "input_activations")
-        per_tensor = (self.weight_quant.strategy == QuantizationStrategy.TENSOR
-                      and self.input_quant.strategy
-                      == QuantizationStrategy.TENSOR)
-        per_channel = (
-            self.weight_quant.strategy == QuantizationStrategy.CHANNEL
-            and self.input_quant.strategy == QuantizationStrategy.TOKEN)
-        if not (per_tensor or per_channel):
-            assert self.weight_quant.strategy == QuantizationStrategy.BLOCK
-            self.weight_block_size = self.weight_quant.block_structure
-            assert self.weight_quant.dynamic is not None
-        else:
-            self.weight_block_size = None
-        self.block_quant = self.weight_block_size is not None
 
-        self.static_input_scales = not self.input_quant.dynamic
-        if self.static_input_scales and per_channel:
-            raise ValueError(
-                "For FP8 Fused MoE layer, we require either per tensor or "
-                "channelwise, dynamic per token quantization.")
-
-        # For GPUs that lack FP8 hardware support, we can leverage the Marlin
-        # kernel for fast weight-only FP8 quantization
+        # disable GPU paths
         self.use_marlin = False
         self.rocm_aiter_moe_enabled = False  # is_rocm_aiter_moe_enabled()
-
-        # cutlass path
         self.is_fp8_w8a8_sm100 = False
         self.use_cutlass = False
         self.disable_expert_map = False


tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -74,10 +75,10 @@ def get_scheme(self,
             return None
 
         # TODO(kyuyeunk): Add support for different act_quant_format
-        act_quant_format = is_activation_quantization_format(  # noqa: F841
-            format
-        ) if format is not None else is_activation_quantization_format(
-            self.quant_format)
+        # act_quant_format = (
+        #     is_activation_quantization_format(  # noqa: F841
```
kyuyeunk: can you delete `# noqa: F841` as well?

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,348 @@
...
+from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors import \
+    CompressedTensorsConfig
+from vllm.model_executor.layers.quantization.compressed_tensors.schemes.compressed_tensors_wNa16 import (  # noqa
+    WNA16_SUPPORTED_BITS, WNA16_SUPPORTED_TYPES_MAP)
+from vllm.model_executor.utils import set_weight_attrs
+
+logger = init_logger(__name__)
+
+
+class CompressedTensorsW8A8Fp8MoEMethod(vllm_ct_moe.CompressedTensorsMoEMethod
```
kyuyeunk: nit: follow convention from other code and import `vllm_ct_moe.CompressedTensorsMoEMethod` as just `CompressedTensorsMoEMethod` and use `VllmCompressedTensorsW8A8Fp8MoEMethod` for the new child class: https://github.com/vllm-project/tpu-inference/blob/330cb1be949d4d20188427a6faa3440453e11e3a/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors.py#L33

kyuyeunk: Instead of inheriting `CompressedTensorsMoEMethod`, is it possible to inherit `CompressedTensorsW8A8Fp8MoEMethod` instead?: https://github.com/vllm-project/vllm/blob/4bf56c79cc252d285d0cb4f5edf323f02af735ca/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L539C7-L539C40 

I'm hoping to minimize redundant code as much as possible.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,348 @@
...
+        # out = torch.einsum("tai,ta -> ti", expert_outs, expert_weights)
+        out = call_jax(jax.lax.dot,
+                       expert_outs,
+                       expert_weights,
+                       dimension_numbers=(((1, ), (1, )), ((0, ), (0, ))),
+                       preferred_element_type=jnp.bfloat16.dtype)
+
+        return out
+
+    def create_weights(
```
kyuyeunk: Similar to above code. Is it simply a copy-paste from `create_weights` of `CompressedTensorsW8A8Fp8MoEMethod`?: https://github.com/vllm-project/vllm/blob/4bf56c79cc252d285d0cb4f5edf323f02af735ca/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe.py#L601

If so, can we remove this logic by using inheritance?

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
qihqi: done.

qihqi: done.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
qihqi: done

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,231 @@
...
+                 moe: FusedMoEConfig, mesh: Mesh):
+        super().__init__(quant_config, moe)
+
+        self.use_marlin = False
+        self.use_cutlass = False
+        self.is_fp8_w8a8_sm100 = False
+
+        self.mesh = mesh
+        self.quant_config = quant_config
+        self.weight_quant = self.quant_config.target_scheme_map["Linear"].get(
```
kyuyeunk: I don't think many of codes aren't needed anymore since many of these are taken care of during `CompressedTensorsW8A8Fp8MoEMethod` initialization which happens when `super().__init__(quant_config, moe)` is called.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,231 @@
...
+                                    requires_grad=False)
+        w2_weight = Parameter(torch_view(w2_weight), requires_grad=False)
+        w2_weight_scale = Parameter(torch_view(w2_weight_scale),
+                                    requires_grad=False)
+        w3_weight = Parameter(torch_view(w3_weight), requires_grad=False)
+        w3_weight_scale = Parameter(torch_view(w3_weight_scale),
+                                    requires_grad=False)
+
+        # TODO dont reuse variable
+        layer.w13_weight = w1_weight
```
kyuyeunk, line 110: shouldn't it be `layer.w1_weight`?

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,231 @@
...
+        if renormalize:
+            expert_weights /= expert_weights.sum(dim=-1, keepdim=True)
+
+        # cond ffn
+        # e = total num of exp = 160
+        # t = seqlen
+        # o = config.imtermediate size
+        # i = config.dim
+        #torch.einsum("ti, eoi -> teo", x, layer.w13_weight) * self.w13_weight_scale)
+        ux1 = call_jax(jax.lax.dot,
```
kyuyeunk, line 169: instead of jax.lax.dot, just using einsum might be better.

so call `jnp.einsum('ti,eoi->teo', x, layer.w13_weight)`

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -0,0 +1,231 @@
...
+        # cutlass path
+        self.is_fp8_w8a8_sm100 = False
+        self.use_cutlass = False
+        self.disable_expert_map = False
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+        assert isinstance(layer, FusedMoE)
+
+        intermediate_size = layer.w13_weight.shape[1] // 2
+        w1_weight = layer.w13_weight[:, :intermediate_size]
```
kyuyeunk, line 47: any reason for splitting this w13 into w1 & w3?

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
qihqi: It can, however IDK if deleting a Parameter on nn.Module at this point will cause issues when other part of the system expects it. I can try to see what happens.

Given that this is temporary and the MoE kernel actually want fused (they call it w1 and w2 though) so I am inclined to live it as-is.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
qihqi: because the vanilla math in `apply` expects it in that format. I can also rewrite that math to use fused. However that would involve some trickeries with shard_map.

It is meant to be undone once the MoE kernel supports fp8. Therefore, not worth it.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
qihqi: torch.einsum I had before actually forward to jnp.einsum verbatim.

The issue is that jnp.einsum does not do mixed precision between fp8 and bf16.

qihqi:
> I've noticed that the code doesn't have any unit test? Can you add one?

added unit test, PTAL. Thanks!

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
kyuyeunk: ping?

PR #1013, opened 2025-11-05
Revert "Bump jax version to 0.8.0"

py4:
Reverts vllm-project/tpu-inference#1006


kyuyeunk:
Is there a reason why this was reverted? The new moe kernel requires jax version 0.8.0 to work, so the code breaks due to this revert.

PR #1019, opened 2025-11-05
[TPU host offload] Setup precompile functions to TPU host offload

saikat-royc:
# Description

the past.

This PR introduces a bucketing strategy of the swap operations. This decomposes operations involving arbitrary numbers of KV cache blocks into smaller, predefined block size buckets. This approach leverages JAX's JIT compilation cache, avoiding costly recompilations for varying input shapes during the hot path of serving. 

1. prepare precompile functions which will cycle through the load and save jitted functions
2. decompose the load and save util functions to be block buckets aligned. This would ensure that compilation is not triggered in the hot path of the serving

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

- existing unit tests
- sglang tool benchmark. some preliminary test cases show x% improvement in the tokens/s and TTFT improvements, because we are avoiding any recompilation during the actual load and save operations 

# Checklist

### DIFF
--- a/tpu_inference/distributed/tpu_connector_local.py
+++ b/tpu_inference/distributed/tpu_connector_local.py
@@ -898,6 +898,12 @@ class TPUConnectorWorker:
         logger.info(
             f"(cpu offloading) swap operation type is {self.swap_op_type}")
 
+        self.use_bucketed_swap_ops = os.getenv(
+            "TPU_OFFLOAD_SKIP_JAX_PRECOMPILE", "0") == "0"
+        logger.info(
+            f"(cpu offloading) use_bucketed_swap_ops={self.use_bucketed_swap_ops}"
+        )
+
         self.swap_in_fn: KVCacheSwapFn = None
         self.swap_out_fn: KVCacheSwapFn = None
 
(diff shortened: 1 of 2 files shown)


tpu_inference/distributed/tpu_connector_local.py
```diff
@@ -1032,56 +1278,40 @@ def _save_blocks_to_cpu(self, req_id: ReqId, full_block_ids: list[int],
 
         try:
             start_time = time.time()
-            blocks_to_save = jnp.array(blocks_to_save)
```
juncgu-google, line 1035: can we keep the original logic when pre-compile is off (the same of the following replacements)?
1. Bucketize may not be needed when pre-compile is off.
2. also, we may need to compare the performance of with and without pre-compile.

tpu_inference/distributed/tpu_connector_local.py
```diff
@@ -967,6 +969,250 @@ def register_runner(self, runner: TPUModelRunner):
...
+                                             token_offset,
+                                             chunk_size_in_tokens,
+                                             axis=0)
+                for layer_cache in flat_kv_caches_tpu
+            ]
+
+            # Swap the bucket to CPU, result is a flat tensor for this bucket. We are doing the chunking inside this function to avoid returning any jnp.concatenate
+            # of kv cache for the the bucketed blocks
+            cpu_chunk_flat_per_layer = self.swap_out_fn(tpu_chunk)
+            jax.block_until_ready(cpu_chunk_flat_per_layer)
```
saikat-royc: we can remove this, because we have a block after the function returns

tpu_inference/distributed/tpu_connector_local.py
saikat-royc: done

tpu_inference/distributed/tpu_connector_local.py
saikat-royc: done

PR #1024, opened 2025-11-06
Support float8_e4m3fn dtype weight loading for jax

inho9606:
This commit is to support float8 dtype weight loading for Jax. Pre-quantized models should have 8bit dtypes, we have to be able to load those dtype tensors.


bvrockwell:
Thanks so much @inho9606 for the contribution! Can you resolve your conflicts please so we can run CI on this and get it reviewed? cc @vkantamani-cienet @Lumosis

bvrockwell:
@jrplatin @kyuyeunk good to close?

PR #1052, opened 2025-11-08
Fix Torchax backend on Pathways

richardsliu:
# Description

This fixes running the Torchax backend on Pathways. In Pathways, the following line would cause the weights to be loaded onto CPU devices on the controller: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/vllm/vllm_model_wrapper.py#L108C9-L109C74

This would later raise a failure when the weights are transferred to the TPU device with `jax.device_put()`, since the target device is non-PJRT.

The fix is to detach the original pytorch/jax tensor wrapper and just use a numpy array for `jax.device_put()`.

FIXES: b/444030476

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/sharding.py
+++ b/tpu_inference/layers/vllm/sharding.py
@@ -5,8 +5,7 @@ import torchax
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torch.nn import Parameter
 from torch.utils import _pytree as pytree
-from torchax.interop import jax_view, torch_view
-from torchax.ops.mappings import t2j
+from torchax.interop import torch_view
 from vllm.lora.layers import (MergedColumnParallelLinearWithLoRA,
                               MergedQKVParallelLinearWithLoRA,
                               RowParallelLinearWithLoRA)


Chenyaaang (approved):

richardsliu:
CI: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5157

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -105,7 +105,7 @@ def load_weights(self):
 
         # Load the vLLM model and wrap it into a new model whose forward
         # function can calculate the hidden_state and logits.
-        with load_context, jax.default_device(jax.devices('cpu')[0]):
+        with load_context:
```
kyuyeunk, line 108: Hi @richardsliu, the line `jax.default_device(jax.devices('cpu')[0])` was introduced in https://github.com/vllm-project/tpu-inference/pull/789 for a specific performance optimization reasons.

And we've found that it causes an issue in https://github.com/vllm-project/tpu-inference/pull/1098. Can you investigate if there's a different way to achieve your desired outcome?

cc: @sixiang-google

PR #1054, opened 2025-11-08
Enable Pipeline Parallelism to use mp as distributed backend on Jax TPU platform

Chenyaaang:
# Description

The implementation for Pipeline Parallelism are splitted into the following small PRs.

- [ ] Util functions for PP on Jax (https://github.com/vllm-project/tpu-inference/pull/1042)
- [ ] worker changes (https://github.com/vllm-project/tpu-inference/pull/1043)
- [ ] runner changes (https://github.com/vllm-project/tpu-inference/pull/1053)
- [x] platform changes (The current PR)
- [ ] torchax changes
- [ ] Jax changes
- [ ] Single host changes (this will be in vLLM)
- [ ] Multi host changes (Ray)

This PR is to modify Jax TPU platform to support PP
- For single host, use `uni` when pp = 1, use `mp` when pp > 1

# Tests

 E2E test has verified the whole PP implementation works properly.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #1053, opened 2025-11-08
Enable Pipeline Parallelism on Jax runner

Chenyaaang:
# Description

The implementation for Pipeline Parallelism are splitted into the following small PRs.

- [ ] Util functions for PP on Jax (https://github.com/vllm-project/tpu-inference/pull/1042)
- [ ] worker changes (https://github.com/vllm-project/tpu-inference/pull/1043)
- [x] runner changes (The current PR)
- [ ] platform changes
- [ ] torchax changes
- [ ] Jax changes
- [ ] Single host changes (this will be in vLLM)
- [ ] Multi host changes (Ray)

This PR is to modify Jax runner to support PP.
- `tpu_jax_runner.py` pass in `intermediate_tensors` to `_execute_model`
- `compilation_manager.py` add PP to the precompile backbone
-  `persistent_batch_manager` update the internal status for all non-last rank, based on the scheduler output's last selected tokens.

# Tests

 E2E test has verified the whole PP implementation works properly.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #1043, opened 2025-11-07
Enable Pipeline Parallelism on jax worker

Chenyaaang:
# Description

The implementation for Pipeline Parallelism are splitted into the following small PRs.

- [ ] Util functions for PP on Jax (https://github.com/vllm-project/tpu-inference/pull/1042)
- [x] worker changes (The current PR)
- [ ] runner changes
- [ ] platform changes
- [ ] torchax changes
- [ ] Jax changes
- [ ] Single host changes (this will be in vLLM)
- [ ] Multi host changes (Ray)

This PR is to modify Jax worker to support PP.
- The worker `__init__` takes in the current worker's IP and its previous worker's IP, to start transfer server and connection later.
- During `execute_model`, for the PP workers who not in the first rank, they need to receive intermediate tensor from the previous worker. For the PP workers who is not the last rank, they need to send intermediate tensor to their next worker.
- The profiler should profile every PP worker, so add subfolders under the parent profile_dir to save profiles for each worker. 

# Tests

 E2E test has verified the whole PP implementation works properly.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #1042, opened 2025-11-07
Create Utils functions for PP

Chenyaaang:
# Description

The implementation for Pipeline Parallelism are splitted into the following small PRs.

- [x] Util functions for PP on Jax (The current PR)
- [ ] worker changes
- [ ] runner changes
- [ ] platform changes
- [ ] torchax changes
- [ ] Jax changes
- [ ] Single host changes (this will be in vLLM)
- [ ] Multi host changes (Ray)

This PR is to create Util functions/classes for Jax's PP.
- `jax_parallel_state.py` is a simplified, Jax version of vLLM's [parallel_state.py](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/parallel_state.py), to transfer intermediate tensors between each process, and also provide rank-control APIs for model loading, etc. 
- `jax_intermediate_tensor.py` is a Jax version of vLLM's [IntermediateTensor](https://github.com/vllm-project/vllm/blob/main/vllm/sequence.py#L52) class. It contains the tensor dictionary that needs to be transfer from one PP rank to another. 

# Tests

 E2E test has verified the whole PP implementation works properly.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #1055, opened 2025-11-08
Enable Pipeline Parallelism on torchax path

Chenyaaang:
# Description

The implementation for Pipeline Parallelism are splitted into the following small PRs.

- [ ] Util functions for PP on Jax (https://github.com/vllm-project/tpu-inference/pull/1042)
- [ ] worker changes (https://github.com/vllm-project/tpu-inference/pull/1043)
- [ ] runner changes (https://github.com/vllm-project/tpu-inference/pull/1053)
- [ ] platform changes (https://github.com/vllm-project/tpu-inference/pull/1054)
- [x] torchax changes (The current PR)
- [ ] Jax changes
- [ ] Single host changes (this will be in vLLM)
- [ ] Multi host changes (Ray)

This PR is to modify vllm_model_wrapper to support PP on torchax path.
- pass `intermediate_tensors` to `step_fun`
- also pass in static boolean variable to adapt jax.jit static property
- convert the intermediate tensor between torch and jax.

# Tests

 E2E test has verified the whole PP implementation works properly.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -150,9 +150,9 @@ class VllmModelWrapper:
             input_embeds: jax.Array,
             layer_name_to_kvcache_index: Sequence[Tuple[str, int]],
             lora_metadata,
-            intermediate_tensors: JaxIntermediateTensors,
-            is_first_rank: bool,
-            is_last_rank: bool,
+            intermediate_tensors: JaxIntermediateTensors = None,
+            is_first_rank: bool = True,
+            is_last_rank: bool = True,
             *args,
         ) -> Tuple[List[jax.Array], jax.Array]:
             layer_name_to_kvcache_index = dict(layer_name_to_kvcache_index)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -161,13 +167,15 @@ def step_fun(
                 # torch_view in order to call the Torch function.
                 original_lora_metadata = replace_lora_metadata(
                     self.model, lora_metadata, self.vllm_config.lora_config)
-                hidden_states = torch.func.functional_call(
+                if not is_first_rank:
```
yixinshi, line 170: Should this if intermediate_tensors and not is_first_rank: ?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -137,7 +139,8 @@ def jit_step_func(self):
                 "xla_tpu_reduce_scatter_collective_matmul_mode":
                 "post_spmd_conservative"
             },
-            static_argnames=("layer_name_to_kvcache_index", ),
+            static_argnames=("layer_name_to_kvcache_index", "is_first_rank",
```
yixinshi, line 142: nit: either fit them into one line, or one item per line.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -147,6 +150,9 @@ def step_fun(
             input_embeds: jax.Array,
             layer_name_to_kvcache_index: Sequence[Tuple[str, int]],
             lora_metadata,
+            intermediate_tensors: JaxIntermediateTensors,
```
mrjunwan-lang: to be compatible with none pp call, we may set the default for these 3 new parameters?

Chenyaaang: I justed added default to this func, actually it's already compatible with none pp call (set in runner)

tpu_inference/models/vllm/vllm_model_wrapper.py
Chenyaaang: if it's not the first rank, it already has intermediate_tensors, so we can skip the first statement.

tpu_inference/models/vllm/vllm_model_wrapper.py
Chenyaaang: pre-commit makes to this format...

PR #1060, opened 2025-11-10
 Multi-slice mesh creation support

wenxindongwork:
# Description

Add support for creating multi-slice mesh. This allows for running DP across multiple slices. Main use case for now is RL scaling. 

Modified and refactored the `_init_mesh` method in `tpu_jax_runner.py`.

# Tests

Tested on XPK cluster with 2 v6e-256 slices. 
Added unit tests for mesh creation. 
https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5139

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_jax_runner.py
+++ b/tpu_inference/runner/tpu_jax_runner.py
@@ -265,8 +265,9 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         if os.getenv("NEW_MODEL_DESIGN", False):
             self.mesh = self._create_new_model_mesh()
         else:
-            # NOTE(wenxindongwork): The new MoE kernel expects a 2D mesh, so we default
-            # to a 2D mesh for now, and we may change this in the future.
+            # NOTE(wenxindongwork): The new MoE kernel expects a 2D mesh, so we need
+            # to create a 2D mesh for now. We should make the new_model_mesh as the default
+            # in the future.
             self.mesh = self._create_2d_mesh()
 
         logger.info(f"Init mesh | mesh={self.mesh}")


tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -259,32 +260,82 @@ def _init_random(self):
...
+            self.mesh = self._create_new_model_mesh()
+        else:
+            # NOTE(wenxindongwork): The new MoE kernel expects a 2D mesh, so we default
+            # to a 2D mesh for now, and we may change this in the future.
+            self.mesh = self._create_2d_mesh()
+
+        logger.info(f"Init mesh | mesh={self.mesh}")
+
+    def _create_new_model_mesh(self) -> jax.sharding.Mesh:
+        axis_names = ("data", "attn_dp", "expert", "model")
```
gpolovets1: We should either use tpu_inference/layers/jax/sharding.py or if we are no longer using it, at least make this a constant.

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -259,32 +260,82 @@ def _init_random(self):
...
+
+        return mesh_utils.create_hybrid_device_mesh(
+            mesh_shape=ici_mesh_shape,
+            dcn_mesh_shape=dcn_mesh_shape,
+            devices=self.devices,
+            allow_split_physical_axes=True,
+        )
+
+    def _create_2d_mesh(self) -> jax.sharding.Mesh:
+        axis_names = ("data", "model")
```
gpolovets1: Do we need to support the 'expert' axis here?

wenxindongwork: we can't because the new MoE kernel only supports the 2D mesh ("data", "model"). To use the "expert" axis, we can set `NEW_MODEL_DESIGN=True`. That was the usage prior to this PR as well.

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -259,32 +262,83 @@ def _init_random(self):
...
-            axis_names = ("data", "attn_dp", "expert", "model")
-            mesh_shape = (sharding_strategy.model_dp_size,
-                          sharding_strategy.attn_dp_size,
-                          sharding_strategy.expert_size,
-                          sharding_strategy.tp_size)
+            self.mesh = self._create_new_model_mesh()
+        else:
+            # NOTE(wenxindongwork): The new MoE kernel expects a 2D mesh, so we default
+            # to a 2D mesh for now, and we may change this in the future.
+            self.mesh = self._create_2d_mesh()
```
gpolovets1, line 271: Given the kernel is still somewhat experimental, shouldn't we use 1D mesh as default and set 2D with a flag?

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -259,32 +262,83 @@ def _init_random(self):
...
+
+        return mesh_utils.create_device_mesh(
+            mesh_shape,
+            self.devices,
+            allow_split_physical_axes=True,
+        )
+
+    def _create_multi_slice_mesh(self, num_slices: int) -> jax.Array:
+        sharding_strategy: ShardingConfigManager = self.vllm_config.sharding_config
+        dp_inner = sharding_strategy.model_dp_size // num_slices
```
gpolovets1, line 305: Why do we only need to do this for DP?

tpu_inference/runner/tpu_jax_runner.py
wenxindongwork: I would agree, but I've been following the existing pattern -- which uses the 2D mesh as default. 

We can swap out the default behaviour in another PR?

tpu_inference/runner/tpu_jax_runner.py
wenxindongwork: so far only DP uses multi-slice, we don't need to use multi-slice for other sharding strategies

tpu_inference/runner/tpu_jax_runner.py
```diff
@@ -259,32 +262,83 @@ def _init_random(self):
...
+            mesh_shape=ici_mesh_shape,
+            dcn_mesh_shape=dcn_mesh_shape,
+            devices=self.devices,
+            allow_split_physical_axes=True,
+        )
+
+    def _create_2d_mesh(self) -> jax.sharding.Mesh:
+
+        sharding_strategy: ShardingConfigManager = self.vllm_config.sharding_config
+        mesh_shape = (
```
gpolovets1, line 326: Why is the mesh shape different from single and multi slice mesh?

wenxindongwork: for multi-slice mesh, there are two mesh shapes -- one for ICI and one for DCN, the product of the two is the global mesh shape.

gpolovets1: The mesh in _create_2d_mesh doesn't appear to use DCN and it also has different ICI shape from the others though.

tpu_inference/runner/tpu_jax_runner.py
gpolovets1: Sg, could you add a TODO for this in the comments?

tpu_inference/runner/tpu_jax_runner.py
wenxindongwork: yes, the 2d mesh only has ("data", "model"), since this is what's expected by the MoE kernel. And this was the previous behavior as well.

PR #1061, opened 2025-11-10
Disable E2E DP test unless nightly

wenxindongwork:
# Description

Currently we are running the DP e2e test for every build, this PR sets this test to trigger only in nightly runs. 

# Tests

commands to reproduce.

# Checklist


QiliangCui:
I would suggest don't disable it for now.

1. this is new feature and we should keep close monitor. 
2. the latest run shows 1 min waiting and 9 minutes of running, which is not bad. 
3. b/457722504 works on increasing v6e-8 capacity. 

I think no need to disable before seeing any blocking issue.

PR #1066, opened 2025-11-11
[Disagg] Fixes for vllm model impl disagg support

sixiang-google:
# Description

Also: removed additional hbm reservation for ironwood. HBM on ironwood is large enough for kv cache transferring buffer so not needed. 

# Tests

pytest

# Checklist

### DIFF
--- a/tpu_inference/core/core_tpu.py
+++ b/tpu_inference/core/core_tpu.py
@@ -446,13 +446,15 @@ class DisaggEngineCore(vLLMEngineCore):
         executor_fail_callback: Optional[Callable] = None,
     ):
         self.vllm_config = vllm_config
-        self.vllm_config.cache_config.gpu_memory_utilization = (
-            self.vllm_config.cache_config.gpu_memory_utilization - 0.1)
 
         self.output_queue = queue.Queue[Union[tuple[int, EngineCoreOutputs],
                                               bytes]]()
 
         self.devices = jax.devices()
+        device_kind = self.devices[0].device_kind
+        if device_kind != 'TPU7x':
+            self.vllm_config.cache_config.gpu_memory_utilization = (
+                self.vllm_config.cache_config.gpu_memory_utilization - 0.1)
         prefill_slice_sizes, decode_slice_sizes, slice_sizes = _get_slice_sizes(
             self.devices)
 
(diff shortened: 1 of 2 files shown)


tpu_inference/core/core_tpu.py
```diff
@@ -446,13 +446,15 @@ def __init__(
...
-            self.vllm_config.cache_config.gpu_memory_utilization - 0.1)
 
         self.output_queue = queue.Queue[Union[tuple[int, EngineCoreOutputs],
                                               bytes]]()
 
         self.devices = jax.devices()
+        device_kind = self.devices[0].device_kind
+        if device_kind != 'TPU7x':
+            self.vllm_config.cache_config.gpu_memory_utilization = (
+                self.vllm_config.cache_config.gpu_memory_utilization - 0.1)
```
hfan, line 457: for large models, would it still be necessary for v7x as well?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -82,9 +82,21 @@ def __init__(self, vllm_config: VllmConfig, rng: PRNGKey, mesh: Mesh):
...
+        modified_slice_config = False
+        if hasattr(
+                self.vllm_config.device_config,
+                'slice') and self.vllm_config.device_config.slice is not None:
+            slice_config = self.vllm_config.device_config.slice
+            modified_slice_config = True
+            self.vllm_config.device_config.slice = None
         vllm_config_for_load = copy.deepcopy(self.vllm_config)
+        if modified_slice_config:
+            self.vllm_config.device_config.slice = slice_config
```
hfan, line 95: why not just do the deepcopy first, then overwrite the device_config.slice directly on the `vllm_config_for_load` object?

tpu_inference/core/core_tpu.py
sixiang-google: Normally we have this threshold at 0.9. For v6e. this leaves ~3GB per device to hold the to-be-transferred kv cache, which is not sufficient so we reserved additional 3GB. For v7 this threshold leaves ~9GB which is generally sufficient.

tpu_inference/models/vllm/vllm_model_wrapper.py
sixiang-google: the slice object is a jax.device list which cannot be pickled

PR #1080, opened 2025-11-12
[Torchax] Add initial support for loading mxfp4

kyuyeunk:
# Description

Add support for loading mxfp4 weights in torchax path but dequantize into bf16 for now.

Verified that it returns correct numeric outputs with [openai/gpt-oss-120b](https://huggingface.co/openai/gpt-oss-120b)

Also verified that performance is the same with bf16 version [unsloth/gpt-oss-120b-BF16](https://huggingface.co/unsloth/gpt-oss-120b-BF16)

Referenced mxfp4 pr in JAX path for some of functionalities: https://github.com/vllm-project/tpu-inference/pull/992

# Tests

```
pytest -v -s tests/layers/vllm/test_mxfp4.py
```

```
MODEL_IMPL_TYPE=vllm examples/offline_inference.py --task=generate --model=openai/gpt-oss-120b --max-model-len=1024 --max-num-batched-tokens=1024 --max-num-seqs=128 --no-enable-prefix-caching --tensor-parallel-size=4 --gpu-memory-utilization=0.8
```

https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5302

# Checklist

### DIFF
--- a/tests/layers/vllm/test_mxfp4.py
+++ b/tests/layers/vllm/test_mxfp4.py
@@ -1,0 +1,192 @@
+import tempfile
+
+import jax
+import jax.numpy as jnp
+import pytest
+import torch
+import torchax
+import utils as test_utils
+from jax.sharding import NamedSharding, PartitionSpec
+from torchax.interop import torch_view
+from torchax.ops.mappings import j2t, t2j
+from vllm.config import set_current_vllm_config
+from vllm.distributed.parallel_state import (ensure_model_parallel_initialized,
+                                             init_distributed_environment)
+from vllm.engine.arg_utils import EngineArgs
+from vllm.forward_context import set_forward_context
+from vllm.model_executor.layers.fused_moe.layer import FusedMoE
+
+from tpu_inference.layers.vllm.quantization import get_tpu_quantization_config
+from tpu_inference.layers.vllm.quantization.mxfp4 import (VllmMxfp4Config,
+                                                          VllmMxfp4MoEMethod)
+
+P = PartitionSpec
+MODELS = ["openai/gpt-oss-20b"]
+MXFP4_BLOCK_SIZE = 32
+
+
+def quantize_to_mxfp4(weight: torch.tensor):
+    # Utilize JAX because native support for e2m1 makes it easier to work with.
+    weight = t2j(weight)
+    e2m1_finfo = jnp.finfo(jnp.float4_e2m1fn)
+    dtype_min = float(e2m1_finfo.min)
+    dtype_max = float(e2m1_finfo.max)
+
+    # Do a subchannel quantization where block size is 32.
+    weight_shape = weight.shape
+    weight_block = weight.reshape(weight_shape[:-1] + (-1, MXFP4_BLOCK_SIZE))
+    abs_max = jnp.max(jnp.abs(weight_block), axis=-1, keepdims=True)
+    scale = abs_max / dtype_max
+
+    weight_q = jnp.clip(weight_block / scale, dtype_min, dtype_max)
+    weight_q = weight_q.astype(jnp.float4_e2m1fn).reshape(weight_shape[:-1] +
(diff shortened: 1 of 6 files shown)


kyuyeunk:
cc @amishacorns @bzgoogle

tests/layers/vllm/test_mxfp4.py
```diff
@@ -0,0 +1,191 @@
...
+    weight_packed = jax.lax.bitcast_convert_type(weight_q, jnp.uint8)
+
+    # We convert scale into e8m0 manually because there is no hardware support.
+    e8m0_finfo = jnp.finfo(jnp.float8_e8m0fnu)
+    _, scale_exp = jnp.frexp(scale.squeeze(axis=-1))
+    # Subtract by one sinced e8m0 has no decimal
+    scale_exp -= 1
+    scale_exp = (scale_exp - e8m0_finfo.minexp).astype(jnp.uint8)
+
+    return j2t(weight_packed), j2t(scale_exp)
```
qihqi, line 53: qq: here you want to return torch.Tensor on CPU or torchax tensor? if latter should use torch_view

tests/layers/vllm/test_mxfp4.py
```diff
@@ -0,0 +1,191 @@
...
+            has_bias=True,
+        )
+    vllm_fused_moe.w13_weight.data = w1_weight
+    vllm_fused_moe.w2_weight.data = w2_weight
+    vllm_fused_moe.w13_weight_scale.data = w1_weight_scale
+    vllm_fused_moe.w2_weight_scale.data = w2_weight_scale
+    vllm_fused_moe.w13_bias.data = w1_bias
+    vllm_fused_moe.w2_bias.data = w2_bias
+
+    jax_a = torch_view(t2j(a, use_dlpack=False))
```
qihqi: can probably do a.to('jax') here.

you can disable dlpack globally here: https://github.com/google/torchax/blob/main/torchax/config.py#L36 default already disabled

tpu_inference/layers/vllm/quantization/mxfp4.py
```diff
@@ -0,0 +1,260 @@
...
+        w13_bias = t2j(layer.w13_bias, use_dlpack=False)
+
+        w2_weight = u8_unpack_e2m1(t2j(layer.w2_weight, use_dlpack=False))
+        w2_weight_scale = e8m0_to_fp32(
+            t2j(layer.w2_weight_scale, use_dlpack=False))
+        w2_bias = t2j(layer.w2_bias, use_dlpack=False)
+
+        # We dequantize fp4 weights into bf16.
+        # TODO(kyuyeunk): Add native support for MXFP4.
+        w13_weight = dequantize_block_weight(w13_weight, w13_weight_scale,
```
qihqi, line 132: here you dequantize and store the dequantized bf16.

What if you store the weights as-is, but do the dequantiization inside of `apply` below? Would that save some memory (in expense of compute) given that the dequantized tensor is now temporary?

tests/layers/vllm/test_mxfp4.py
kyuyeunk: done.

tests/layers/vllm/test_mxfp4.py
kyuyeunk: I found that using torch_view gave me following error when I tried to do `layer.w2_weight.data = w2_weight`.

```
RuntimeError: Attempted to call `variable.set_data(tensor)`, but `variable` and `tensor` have incompatible tensor type.
```

tpu_inference/layers/vllm/quantization/mxfp4.py
kyuyeunk: It's a bit hard to explain in text but the weight dequantization process is actually very memory intensive operation. And if that is done during inference time, it often leads to OOM.

PR #1098, opened 2025-11-13
[Disagg] Misc fix for vllm model loading in disagg

sixiang-google:
# Description

-Stop inserting when no new blocks are available
-Use only TPU cores in current mesh to load weights. In Disagg setup, default device (TPU_0) will have no HBM after prefill engine is spinned up.
-vllm_config.compilation_config.static_forward_context contains torchax tensors and cannot be deepcopied. 
-AsyncTPUModelRunnerOutput needs a get_output process
-Small fix in prefill engine eviction

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -189,120 +189,131 @@ class VllmUnquantizedFusedMoEMethod(UnquantizedFusedMoEMethod):
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
         assert isinstance(layer, FusedMoE)
-
-        w13_weight = t2j(layer.w13_weight, use_dlpack=False)
-        w2_weight = t2j(layer.w2_weight, use_dlpack=False)
-
-        if self.moe.has_bias:
-            w13_bias = t2j(layer.w13_bias, use_dlpack=False)
-            w2_bias = t2j(layer.w2_bias, use_dlpack=False)
-
-        if layer.activation == "swigluoai":
-            # When using swigluoai, vLLM splits gmm output in a interleaved way.
-            # However, interleaved split is not performant on TPU. Therefore,
-            # we preprocess the weight so that splitting gmm output by middle
-            # can still get the same result.
-            w1_weight = w13_weight[:, ::2, :]
-            w3_weight = w13_weight[:, 1::2, :]
-            w13_weight = jnp.concat([w1_weight, w3_weight], axis=1)
+        available_devices = self.mesh.devices.flatten()
+        with jax.default_device(available_devices[0]):
+            w13_weight = t2j(layer.w13_weight, use_dlpack=False)
+            w2_weight = t2j(layer.w2_weight, use_dlpack=False)
 
             if self.moe.has_bias:
-                w1_bias = w13_bias[:, ::2]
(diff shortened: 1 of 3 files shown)


kyuyeunk:
> Use only TPU cores in current mesh to load weights. In Disagg setup, default device (TPU_0) will have no HBM after prefill engine is spinned up.

I thought I have change the behavior so that weights are initialized in cpu? Isn't it not the case anymore? https://github.com/vllm-project/tpu-inference/pull/789

sixiang-google:
> > Use only TPU cores in current mesh to load weights. In Disagg setup, default device (TPU_0) will have no HBM after prefill engine is spinned up.
> 
> I thought I have change the behavior so that weights are initialized in cpu? Isn't it not the case anymore? #789

After calling t2j they are somehow moved to TPU:0. I printed out the residing devices of these tensors to verify.

kyuyeunk:
> > > Use only TPU cores in current mesh to load weights. In Disagg setup, default device (TPU_0) will have no HBM after prefill engine is spinned up.
> > 
> > 
> > I thought I have change the behavior so that weights are initialized in cpu? Isn't it not the case anymore? #789
> 
> After calling t2j they are somehow moved to TPU:0. I printed out the residing devices of these tensors to verify.

Weird. I've noticed that my code was removed in https://github.com/vllm-project/tpu-inference/pull/1052. Can you reapply the change in https://github.com/vllm-project/tpu-inference/pull/789 and verify if it resolves your issue?

sixiang-google:
> > > > Use only TPU cores in current mesh to load weights. In Disagg setup, default device (TPU_0) will have no HBM after prefill engine is spinned up.
> > > 
> > > 
> > > I thought I have change the behavior so that weights are initialized in cpu? Isn't it not the case anymore? #789
> > 
> > 
> > After calling t2j they are somehow moved to TPU:0. I printed out the residing devices of these tensors to verify.
> 
> Weird. I've noticed that my code was removed in #1052. Can you reapply the change in #789 and verify if it resolves your issue?

Just tested this out, loading on CPU is slower than on tpu:0.

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -191,120 +189,131 @@ def select_gemm_impl(
...
-
-        if layer.activation == "swigluoai":
-            # When using swigluoai, vLLM splits gmm output in a interleaved way.
-            # However, interleaved split is not performant on TPU. Therefore,
-            # we preprocess the weight so that splitting gmm output by middle
-            # can still get the same result.
-            w1_weight = w13_weight[:, ::2, :]
-            w3_weight = w13_weight[:, 1::2, :]
-            w13_weight = jnp.concat([w1_weight, w3_weight], axis=1)
+        available_devices = self.mesh.devices.flatten()
```
kyuyeunk, line 192: Can you move this logic into the context instead?: https://github.com/vllm-project/tpu-inference/blob/3810422081b4b7afe389c407adec2d46ace81ac1/tpu_inference/models/vllm/vllm_model_wrapper.py#L123

As you tried to expand the coverage, you probably need to apply a similar for all the other different configurations - for quantized layer, fp8 layer, mxfp4 - and it's not really scalable.

A scalable approach is to just add this logic into context of the model loading function.

PR #590, opened 2025-08-27
Simplify tensor slicing logic by removing shard_map

kyuyeunk:
# Description

Simplify tensor slicing logic by removing shard_map. Use reshape, slice, and concatenation to achieve the same result.

This logic was separated from https://github.com/vllm-project/tpu_commons/pull/512

# Tests

Ran following models and verified that performance and numerics results has not changed

- Unquantized
  - meta-llama/Llama-3.1-8B-Instruct
  - meta-llama/Llama-3.1-70B-Instruct
- Quantized
  - RedHatAI/Meta-Llama-3.1-8B-Instruct-quantized.w8a8
  - RedHatAI/Meta-Llama-3.1-70B-Instruct-quantized.w8a8

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #1137, opened 2025-11-20
[DP] Functional DP for GPT-OSS

wenxindongwork:
# Description

Add functional DP support for GPT-OSS Torchax backend. 

Verified baseline throughput unchanged (5037.82) , DP=2 throughput is 1.54x (7781.92). 

Validated numerical correctness with offline_inference.py

Full details: https://paste.googleplex.com/5240826907197440

# Tests

https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5851

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/fused_moe.py
+++ b/tpu_inference/layers/vllm/fused_moe.py
@@ -143,7 +143,6 @@ def tensor_sharded_gmm_merged_column_parallel(
             mesh=mesh,
             in_specs=(P("data", "model"), P(None, "model"), P("data")),
             out_specs=(P("data", "model")),
-            check_rep=False,
         )(gmm_result, rhs_bias, group_sizes)
 
     n_shards = mesh.shape["model"]
@@ -200,7 +199,6 @@ def tensor_sharded_gmm_row_parallel(
             mesh=mesh,
             in_specs=(P("data"), P(), P("data")),
             out_specs=(P("data")),
-            check_rep=False,
         )(gmm_result, rhs_bias, group_sizes)
 
     return gmm_result


tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -167,14 +182,26 @@ def _gmm_all_reduce(lhs, rhs, group_sizes):
...
-        gmm_result = (gmm_result + rhs_bias).astype(gmm_result.dtype)
+
+        def _add_bias(gmm_result_local, rhs_bias_local, group_sizes_global):
+            rhs_bis = jnp.repeat(rhs_bias_local,
+                                 group_sizes_global,
+                                 0,
+                                 total_repeat_length=m // mesh.shape["data"])
+            return (gmm_result_local + rhs_bis).astype(gmm_result_local.dtype)
+
+        gmm_result = shard_map(
```
kyuyeunk, line 197: personally, I prefer not using shard_map if possible as it is really prone to numeric error when not properly used. When using `check_rep=False`, unlike other ops, there isn't any safety feature that guarantees that all the numerics of a tensor adheres to a proper SPMD / sharding annotation.

I prefer using it only when it's really necessary (like using kernel).

Please modify this code not use shard map and you can refer to this PR where I replaced existing using of shard_map to a regular jax function: https://github.com/vllm-project/tpu-inference/pull/590

tpu_inference/layers/vllm/quantization/common.py
```diff
@@ -61,7 +61,12 @@ def __init__(self, vllm_config: VllmConfig, mesh: Mesh, layer: LinearBase):
                 " bad performance.", type(layer))
 
         self.bias_sharding = P(self.weight_sharding[0])
-        self.n_shards = self.mesh.shape.get(self.weight_sharding[0], 1)
+        if isinstance(self.weight_sharding[0], tuple):
```
kyuyeunk, line 64: is it an delibrate choice to not use `ShardingAxisName`?: https://github.com/vllm-project/tpu-inference/blob/3a9e2d4d34c5205ebd2c2ea913b8d82bc928db76/tpu_inference/layers/common/sharding.py#L50-L57

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: For more context, when using shard map with check_rep=False, when using something like 'out_spec=P(None)', it only annotates the tensor as having that sharding but shard map does not introduce any collective to ensure it.

Meaning, it is possible that the output tenor numeric is not replicated along multiple devices and all  devices have different numeric because shard_map does not provide any guarantees - which makes it really painful to debug when there's a numeric issue.

wenxindongwork: I tried to not use a shard_map but did not figure out a way that doesn't involve a for loop. 

The complexity here is that `jnp.repeat(bias, group_size, 0)` expects bias and group_size to share the same size on dimension 0, but group_size[0]= `DP*num_experts` whereas `bias[0] = num_experts`.

tpu_inference/layers/vllm/quantization/common.py
wenxindongwork: I wanted to make this PR minimal and debuggable. I plan to change to ShardingAxisName when I add attn_dp support, does that sounds like a good plan to you?

kyuyeunk: sgtm!

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: if that's the case, can you do something like this?

```
# convert (experts, model_dim) to (experts * dp_size, model_dim)
bias = jnp.repeat(bias, dp_size, 0) 

# (optional. may or may not needed) match bias's sharding with group_size's sharding
bias = jax.lax.with_sharding_constraint(bias, P("data", "model"))

# Now the bias.shape[0] and group_size.shape[0] matches
rhs_bias = jnp.repeat(bias, group_size, 0)
```

kyuyeunk (commented):
> Please add torchax dp related unit tests.

Also, please address this comment.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: Note that if complier optimization works correctly, the first 2 code (jnp.repeat & sharding constraint) will be a no-op. because the data is already present in each dp rank & we are just telling it to treat them differently starting from now.

wenxindongwork: Thanks Kyuyeun for this suggestio, For correctness, I have to use `jnp.tile` instead of `jnp.repeat`. however, I am noticing performance drop (7575.50 vs 7781.92) if I do this instead of shard_map. Maybe due to `jnp.tile` not being a no-op? 

```
        rhs_bias = jnp.tile(rhs_bias, (mesh.shape["data"], 1)) 
        # adding the sharding constraint does not make a difference 
        rhs_bias = jnp.repeat(rhs_bias, group_sizes, 0, total_repeat_length=m)
        gmm_result = (gmm_result + rhs_bias).astype(gmm_result.dtype)
```

wenxindongwork:
added e2e model parallelism test for Llama3.1 1b for torchax.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: Hmm in-theory it should be no-op.

because bias is already replicated along TPUs in dp axis, and combining tile/repeat with sharding constraint just tells TPU to treat them like a separate non-replicated tensor.

I'll do some test locally and get back to you asap.

kyuyeunk: okay, i've done various things and was honestly surprised that tile/repeat method didn't work??

so let's leave it as using shard_map for now and figure out a if we can do something differently later.

but can you remove `check_rep=False` arguments? Because that's what really scares be because it means compiler will not do any checks to ensure numerics are correct.

wenxindongwork: done!

kyuyeunk:
i think it all looks good. but can you update the branch, run the ci & confirmed that it all passes?

PR #1139, opened 2025-11-20
Fix numerical issue on hybrid kv cache allocation

Chenyaaang:
# Description

Fix numerical issue on hybrid kv cache allocation. When we enable hybrid kv cache, at each kv cache allocation round, the block_id is different between each kv cache group, which means different layers are writing to different block_ids, so we need to create individual attention metadata for each layer, instead of using the same attention metadata for every layer. 

# Tests

- unit tests in tpu_worker, tpu_runner passed
- The results w/ vs w/o hybrid kv cache are the same when I run offline_inference.py with Gemma model. `python examples/offline_inference.py --model google/gemma-3-27b-it --tensor-parallel-size 8`
- CI: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5787 all tasks are green except for lora, which I believe is an upstream change, not related to my pr. 

# Checklist

### DIFF
--- a/tests/runner/test_tpu_runner_dp.py
+++ b/tests/runner/test_tpu_runner_dp.py
@@ -46,8 +46,15 @@ class TestTPUJaxRunnerDPInputsLightweight:
         self.runner.query_start_loc_cpu = np.zeros(10, dtype=np.int32)
         self.runner.seq_lens_cpu = np.zeros(8, dtype=np.int32)
         self.runner.logits_indices_cpu = np.zeros(8, dtype=np.int32)
-        self.runner.block_table_cpu = np.zeros((8, 8), dtype=np.int32)
+        self.runner.block_tables_cpu = [np.zeros((8, 8), dtype=np.int32)]
         self.runner.arange_cpu = np.arange(64, dtype=np.int64)
+
+        # mock kv cache group
+        mock_kv_cache_config = MagicMock()
+        mock_kv_cache_group = MagicMock()
+        mock_kv_cache_config.kv_cache_groups = [mock_kv_cache_group]
+        self.runner.kv_cache_config = mock_kv_cache_config
+        self.runner.use_hybrid_kvcache = False
 
         # Mock scheduler config for async scheduling
         self.runner.scheduler_config = MagicMock()
@@ -102,8 +109,8 @@ class TestTPUJaxRunnerDPInputsLightweight:
         result = self.runner._prepare_inputs_dp(scheduler_output)
 
         # Basic assertions
-        assert len(result) == 7
-        input_ids, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
+        assert len(result) == 8
+        input_ids, positions, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
 
         # Verify utility functions were called
         mock_runner_utils.get_padded_token_len.assert_called()
@@ -380,7 +387,7 @@ class TestTPUJaxRunnerDPInputsLightweight:
 
         # Execute the method
         result = self.runner._prepare_inputs_dp(scheduler_output)
-        input_ids, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
+        input_ids, positions, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
         # 1. Verify input_ids content
         expected_input_ids = np.zeros(16, dtype=np.int32)
         expected_input_ids[:2] = [1006, 1007]
@@ -494,7 +501,7 @@ class TestTPUJaxRunnerDPInputsLightweight:
 
         # Execute the method
         result = self.runner._prepare_inputs_dp(scheduler_output)
-        input_ids, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
+        input_ids, positions, attention_metadata, sampling_metadata, logits_indices, spec_decode_metadata, logits_indices_selector, padded_num_reqs = result
 
         # 1. Verify input_ids
         expected_input_ids = np.zeros(16, dtype=np.int32)


tpu_inference/runner/tpu_runner.py
```diff
@@ -438,8 +438,11 @@ def _init_inputs(self) -> None:
 
         self.input_ids_cpu = np.zeros(self.max_num_tokens, dtype=np.int32)
         self.positions_cpu = np.zeros(self.max_num_tokens, dtype=np.int32)
-        self.block_table_cpu = np.zeros(
-            (self.max_num_reqs, self.max_num_blocks_per_req), dtype=np.int32)
+        self.block_tables_cpu = [
```
py4, line 444: We are initiliazing `block_tables_cpu` twice. Once in kv_cache_manager once here. Maybe do once? Also the length of this variable differ across these two

tpu_inference/runner/tpu_runner.py
```diff
@@ -701,6 +705,7 @@ def _execute_model(
         # TODO(pooyam): I guess we can remove returning sampling_metadata in `_prepare_inputs` after https://github.com/njhill/vllm/commit/b7433ca1a47732394b1bdea4099d98389515954b
         (
             input_ids,
+            input_positions,
```
py4, line 711: Why are we returning `input_positions` here? Shouldn't it be in `attn_metadata`?

tpu_inference/runner/tpu_runner.py
Chenyaaang: I prefer aligning with the existing implementation. In kv_cache_manager/maybe_reinitialize_input_batch, it's collecting the actual block_size from each kv cache group, and update input_batch if needed. As you can see, `input_batch` is also initialized in the runner at first.

tpu_inference/runner/tpu_runner.py
Chenyaaang: "Shouldn't it be in attn_metadata"- Yes, it is in attn_metadata. 

But if we use hybrid kv cache, attn_metadata is a dict instead of a single metadata obj, which means we need to get it by `attn_metadata[any_layer_name].input_positions`. Considering either pass in layer name or `input_positions` directly, I chose the later way.

tpu_inference/runner/tpu_runner.py
py4: Hmm, the issue when reading this line at tpu_runner.py, it gives this impression that the length of this array is always 1, which is not the case. If you want to keep it, I recommend adding a comment at least

tpu_inference/runner/tpu_runner.py
py4: I mean in `self.model_fn`, what's the difference between `input_positions` inside `attn_metadata` and the `input_positions` that you are passing directly? It doesn't seem clean to me that now there are two different fields for `input_positions`

Chenyaaang:
> Does this also work for JAX path? if no, can we also make JAX path work?

It should be backend agnostic, but to enable in Jax, we need to modify the individual jax model. Previously all jax models don't need hybrid kv cache, so it's not enabled. The numerical issue is also reported using vLLM model instead of flax nnx.

Chenyaaang:
> I mean in self.model_fn, what's the difference between input_positions inside attn_metadata and the input_positions that you are passing directly? It doesn't seem clean to me that now there are two different fields for input_positions

There's no difference between those 2 input_positions, it's just the attention_metadata becomes a dict[layer_name, attn_metadata_for_that layer] instead of a single attn_metadata shared by every layer. So inside vllm_model_wrapper's [step_fun](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/vllm/vllm_model_wrapper.py#L190), when we need to pass input_positions to the model, we used to get it from attn_metadata.input_positions, now we need to get it in this way: `attn_metadata[layer0_name].input_positions`, which requires us to know the layer name, so I chose to pass `input_positions` directly.

tpu_inference/runner/tpu_runner.py
Chenyaaang: [input_batch](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/runner/tpu_runner.py#L435) is also passing in [self.block_size] when it's first created in the runner, and later being updated in kv cache mgr. I've added some comments here.

tpu_inference/runner/tpu_runner.py
Chenyaaang: There's no difference between those 2 input_positions, it's just the attention_metadata becomes a dict[layer_name, attn_metadata_for_that layer] instead of a single attn_metadata shared by every layer. So inside vllm_model_wrapper's [step_fun](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/vllm/vllm_model_wrapper.py#L190), when we need to pass input_positions to the model, we used to get it from attn_metadata.input_positions, now we need to get it in this way: `attn_metadata[layer0_name].input_positions`, which requires us to know the layer name, so I chose to pass `input_positions` directly.

kyuyeunk:
with this PR, Ion gpt-oss, 've verified that numeric issue has been solved & also a performance issue that stemmed from numeric issues has been resolved.

tpu_inference/runner/tpu_runner.py
py4: i think it makes the code messy due to having two redundant fields. It's won't be clear one should read from `input_positions` or `attn_metadata.input_positions`.
Also isn't `input_positions` same for all layers? So at each later we can do something like `attn_metadata.values()[0].input_positions`. If not, i think it's better to get using layer name. 
Overall having two fields for the same thing doesn't look good i think. wdyt?

Chenyaaang: Jax.jit doesn't allow you to do `attn_metadata.values()[0]`, with error "TypeError: 'dict_values' object is not subscriptable", that's my original way, so I turned to the current way. 

Using layer name requires us to pass in layer names, which is even more complicated and needs more changes. I still believe my current solution is the easiest way. 

Actually I think this is a quite open-ended discussion, we both have our own opinion, I suggest we merge the PR first as our initial goal is to solve the numerical issue so that people waiting for this fix can use it asap, having back and forth debate can only lead to more conflicts to resolve and delay our primary goal.

Chenyaaang: Resolve this first, we can sync offline and have a separate PR to finalize this.

PR #1150, opened 2025-11-21
Eager resolution for vllm.current_platform when running Pathways

richardsliu:
# Description

Under Pathways, the platform resolution needs to be done eagerly in order to properly initialize the runtime environment for vLLM. This is because Pathways initialization happens in the controller node only and serves to bootstrap the global device mesh and topology. Lazy resolution could lead to either incorrect environment setup or other python loading errors.

FIXES: b/123456
FIXES: #123456

# Tests

https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5816

### DIFF
--- a/tpu_inference/__init__.py
+++ b/tpu_inference/__init__.py
@@ -1,6 +1,11 @@
 # The environment variables override should be imported before any other
 # modules to ensure that the environment variables are set before any
 # other modules are imported.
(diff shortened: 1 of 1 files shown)


wenxindongwork (approved):

QiliangCui (approved):
Let‘s wait for a the test: 
https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5816

PR #1153, opened 2025-11-21
[Draft] Change `sampled_tokens` to be a list of list[int] instead of list[np.array]

jrplatin:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -123,9 +123,11 @@ class AsyncTPUModelRunnerOutput(AsyncModelRunnerOutput):
         selected_token_ids = np.expand_dims(next_tokens_cpu[:self._num_reqs],
                                             1)
 
-        valid_sampled_token_ids = [token_id for token_id in selected_token_ids]
+        valid_sampled_token_ids = [
+            token_id.tolist() for token_id in selected_token_ids
+        ]
         for i in self._discard_sampled_tokens_req_indices:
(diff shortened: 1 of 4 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -941,21 +945,22 @@ def _sample_from_logits(
                 next_tokens = next_tokens[logits_indices_selector]
             selected_token_ids = np.expand_dims(next_tokens[:num_reqs], 1)
             valid_sampled_token_ids = [
-                token_id for token_id in selected_token_ids
+                token_id.tolist() for token_id in selected_token_ids
             ]
         else:
+            # TODO?
```
jrplatin, line 951: @py4 Any thoughts here? I think maybe best to update `parse_output` to return `list[list[int]]` instead of `list[np.array]`?

Lumosis: I think we should revert https://github.com/vllm-project/tpu-inference/commit/83249b216fad9444311be6f9d82dbed19db22e02 as well.

Lumosis: The upstream API keeps changing on us, so we have to keep going back and forth too. It's a pain.

py4: I think reverting the two commits in the chat should be enough. we can double check

PR #1158, opened 2025-11-22
[FusedMoE] Support sub-channel quantization: FP4, FP8, INT8, ...

bythew3i:
TODO (will resolve in following up PRs):
* Add benchmark test for GPT-OSS
* Tune GPT-OSS FP4 MoE block size
* Debug race condition

### DIFF
--- a/tpu_inference/kernels/fused_moe/v1/kernel.py
+++ b/tpu_inference/kernels/fused_moe/v1/kernel.py
@@ -535,7 +535,7 @@ def _fused_ep_moe_kernel(
                         0,
                         pl.ds(offset // subc_quant_wsz, bd1_per_t_packing //
                               subc_quant_wsz),
-                        0:1,
+                        pl.ds(0, 1),
                         pl.ds(bf_id * bf, bf),
                     ],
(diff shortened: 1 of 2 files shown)


tpu_inference/kernels/fused_moe/v1/kernel.py
```diff
@@ -435,44 +516,85 @@ def wait_a2a_gather_recv_all():
...
                 sem=local_sems.at[bw2_sem_id, 2],
             ).start()
+            if w2_scale_hbm is not None:
+                assert subc_quant_wsz is not None
+                pltpu.make_async_copy(
+                    src_ref=w2_scale_hbm.at[
+                        local_e_id,
+                        pl.ds(bf_id * bf // subc_quant_wsz, bf //
+                              subc_quant_wsz),
+                        0:1,
```
kyuyeunk: nit: use `pl.ds(0, 1)` for consistency with w2_scale_slices

bythew3i: updated

PR #1165, opened 2025-11-24
[Spec Decoding][Eagle3] Fix bug of eagle-3 not being compataible with non-8b models.

py4:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/spec_decode/jax/eagle3.py
+++ b/tpu_inference/spec_decode/jax/eagle3.py
@@ -9,9 +9,12 @@ import numpy as np
 from vllm.config import VllmConfig
 
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
+from tpu_inference.logger import init_logger
 from tpu_inference.models.common.model_loader import get_model
 from tpu_inference.runner import utils as runner_utils
 from tpu_inference.utils import device_array
+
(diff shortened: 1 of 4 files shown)


tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -51,9 +54,22 @@ def load_model(self, target_model: Any) -> None:
...
+            logger.info(
+                "Draft model does not have embedding. Setting draft model's embed_tokens to target model's embed"
+            )
+            self.state.model.embed_tokens = target_model.model.embed
+        elif jnp.array_equal(draft_embed_tokens.embedding,
+                             target_model.model.embed.embedding):
+            logger.info(
+                "Draft model's embed_tokens is identical to target model's embed. Sharing the embedding."
+            )
+            self.state.model.embed_tokens = target_model.model.embed
```
Lumosis, line 70: Do we need to free draft model's embed_tokens to save some memory?

py4: I think this is essentially freeing because there is no pointer to the old `embed_tokens` and it will be garbage collected

PR #1167, opened 2025-11-25
[Kernel][FusedMoE] Add support for bias

kyuyeunk:
# Description

Add bias support for fused moe kernel

# Tests

https://buildkite.com/tpu-commons/tpu-inference-ci/builds/5911

```
pytest tests/kernels/fused_moe_v1_test.py -k test_bias
pytest tests/layers/vllm/test_mxfp4.py -k test_mxfp4_fused_moe_use_kernel
pytest tests/layers/vllm/test_unquantized.py -k test_fused_moe_use_kerne
```

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/mxfp4.py
+++ b/tpu_inference/layers/vllm/quantization/mxfp4.py
@@ -24,6 +24,8 @@ from vllm.model_executor.layers.quantization.mxfp4 import (Mxfp4Backend,
 from vllm.model_executor.layers.quantization.utils.quant_utils import \
     is_layer_skipped
 
+from tpu_inference import envs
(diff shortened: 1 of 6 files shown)


tpu_inference/layers/vllm/quantization/mxfp4.py
bythew3i, line 1: why do we call the file mxfp4 but uses sub-channel quantization?

kyuyeunk: we can consider mxfp4 as a specific type of subchannel quantization where subchannel size is 32, weight dtype is fp4, and scale dtype is e8m0.

from code reusability point, it's better to just write a generic subchannel quantization code with block size &  weight dtype and scale dtype as arguments to support mxfp4.

PR #1171, opened 2025-11-25
Add e2e test for model registration (out of tree plugins)

karan:
Signed-off-by: Karan Goel <karangoel@google.com>

# Description

Adds tests to catch cases when vllm's model interface validation changes and it breaks tpu-inference model registration (for out of tree plugins).

There's also a perf test that validates that no torchax and jax performance are similar (this catches cases where for eg some global config/env var added to the jax path causes performance regressions - this would affect in-tree and out-of-tree perf equally).

# Tests

From vllm HEAD:

```
$ pytest tests/e2e/test_model_loader.py
...
...
3 passed, 6 warnings in 255.94s (0:04:15)
```

Change vllm's `is_vllm_model` to also validate the existence of a `foobar` method and see the expected test failure:

```
$ pytest tests/e2e/test_model_loader.py
...
...
FAILED tests/e2e/test_model_loader.py::test_register_model_success - AttributeError: type object 'VllmCompatibleDummyGoodModel' has no attribute 'foobar. Did you mean: 'format'?
FAILED tests/e2e/test_model_loader.py::test_registered_model_passes_vllm_interface_check - AttributeError: type object 'VllmCompatibleDummyGoodModel' has no attribute 'foobar'. Did you mean: 'format'?
```

NOTE that for the perf test, I had to revert vllm to e50209864 (due to issue caused by https://github.com/vllm-project/vllm/pull/29143).

# Checklist


.buildkite/features/Out_Of_Tree_Model_Support.yml
```diff
@@ -30,7 +33,10 @@ steps:
       queue: tpu_v6e_queue
     commands:
       - |
-        buildkite-agent meta-data set "out_of_tree_model_support_PerformanceTest" "to be added"
+      - |
```
jcyang43, line 35: We need to remove this duplicate line (w/ line 35) to avoid syntax error

jcyang43 (approved):

PR #1172, opened 2025-11-25
Fix load_context to use nullcontext() if running under Pathways

richardsliu:
# Description

`jax.default_device(jax.devices("cpu")[0])` was added in https://github.com/vllm-project/tpu-inference/pull/789 as a performance optimization. However if vllm is running under Pathways, this would cause the model weights to be loaded on a CPU-only node, typically the Pathways controller. This would cause subsequent `jax.put_device()` calls to fail.

This fix would check for Pathways and use a nullcontext if that is the case.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -9,6 +9,7 @@ import jax
 import torch
 import torch.nn
 import torchax
+import vllm.envs as vllm_envs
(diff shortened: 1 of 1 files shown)


kyuyeunk (approved):
This should resolve both OOM issue on non-pathways issue & unblock pathways. Thanks!

PR #1180, opened 2025-11-26
[RPA][Kernel] Update hd64 variant sliding window code

kyuyeunk:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist


kyuyeunk:
> Could you please explain why do we need `strict_sliding_window` for hd64 variant?

`strict_sliding_window` is used for making sliding window numeric output 100% correct by only calculating tokens that are strictly within sliding window. This accuracy comes with a significant performance degradation.

but thing is, in real workload, tokens outside of sliding window are still all valid tokens. So incorporating few tokens outside of sliding window does not affect final result in any meaningful way.

So I've added this option so we can enable & disable it based on our use cases.

PR #1185, opened 2025-11-26
fix(rpa-v3): add sliding window mask to h64 kernel and attention_sink to h128

### LINKED ISSUE #1169: [Bug]: Sliding mask in v3 rpa might be wrong
### Your current environment

i was testing models in EasyDel with VLLM-RPA3, and I noticed a few sliding mask issues, so I fixed them. If u confirm that the sliding mask logic is kinda wrong, I can push the corrected and debugged version

### 🐛 Describe the bug

.

- [x] Make sure you already searched for relevant issues and checked the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

erfanzar:
## Summary
Fixes #1169

This PR fixes two issues in the ragged-paged attention v3 kernels:

- **kernel_hd64.py (h64)**: Added missing sliding window mask in the kernel. The 
original code only skipped fetching KV blocks outside the window, but didn't apply 
token-level masking within partially-covered blocks.

- **kernel.py (h128)**: Added attention_sink support following the same pattern as the 
h64 kernel. Attention sinks allow the model to "dump" attention to a virtual token that 
doesn't contribute to the output.

## Changes
| File | Change |
|------|--------|
| kernel_hd64.py | Added sliding window mask in flash_attention |
| kernel.py | Added attention_sink parameter to all functions |

### DIFF
--- a/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
+++ b/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
@@ -35,6 +35,7 @@ def ref_ragged_paged_attention(
     page_indices: jax.Array,  # i32[max_num_seqs * pages_per_seq]
     cu_q_lens: jax.Array,  # i32[max_num_seqs + 1]
     distribution: jax.Array,  # i32[3]
+    attention_sink: jax.Array | None = None,  # f32[actual_num_q_heads]
     *,
     sm_scale: float = 1.0,
     sliding_window: int | None = None,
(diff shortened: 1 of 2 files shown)


kyuyeunk:
Hi @erfanzar, thanks for the quick PR. Here are few comments:

- Sliding mask related issue in hd64 variant is being taken care of in  this PR: https://github.com/vllm-project/tpu-inference/pull/1180
- Not adding attention sink feature into kernel.py and only kernel_hd64.py was a deliberate choice to streamline kernel.py codebase since the feature only seems to be used for gpt-oss - which has head dim size of 64. We will reevaluate this if there is non head dim 64 model that also requires attention sink.

PR #1192, opened 2025-11-27
[Spec][Eagle3] Improve perf and compilation time

py4:
Pass state through jit boundary to prevent long compilation time.
We were not passing model state through jit boundary. As a result, concrete values taking gigabytes were not being traced and had become part of the jaxpr as constant


Lumosis (approved):

Lumosis:
Thank you for fixing this. Curious how fast can it be after the fix.

PR #1199, opened 2025-11-29
Refactor moe codebase

kyuyeunk:
# Description

Cleanup fused moe file. No change in functionality.

- Add type annotations
- Add better function documentation
- Remove `transpose_rhs` since it's always `True`
- Remove unnecessary sharding constraints which were already enforced via `shard_map` 
- Replace `jax.experimental.shard_map.shard_map` into `jax.shard_map`
- Remove `fused_moe_func_padded` since padding is now handled by jax runner
- Remove unnecessary arguments that can be inferred from tensor or mesh shapes

# Tests

https://buildkite.com/tpu-commons/tpu-inference-ci/builds/6118

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/fused_moe.py
+++ b/tpu_inference/layers/vllm/fused_moe.py
@@ -2,17 +2,16 @@ import functools
 
 import jax
 from jax import numpy as jnp
+from jax import shard_map
 from jax.experimental.pallas.ops.tpu.megablox.gmm import gmm
-from jax.experimental.shard_map import shard_map
-from jax.sharding import Mesh, NamedSharding, PartitionSpec
+from jax.sharding import Mesh
+from jax.sharding import PartitionSpec as P
 
 from tpu_inference.layers.vllm.linear_common import \
     slice_sharded_tensor_for_concatenation
 
-P = PartitionSpec
 
-
-def activation_fn(activation: str, x1, x2):
+def activation_fn(activation: str, x1: jax.Array, x2: jax.Array) -> jax.Array:
     match activation:
         case "silu":
             return jax.nn.silu(x1) * x2
@@ -23,7 +22,10 @@ def activation_fn(activation: str, x1, x2):
                 f"FusedMoE does not support {activation} activation")
 
 
-def _swigluoai(x1, x2, alpha=1.702, limit=7.0):
+def _swigluoai(x1: jax.Array,
+               x2: jax.Array,
+               alpha=1.702,
+               limit=7.0) -> jax.Array:
     x1 = jnp.clip(x1, a_max=limit)
     x2 = jnp.clip(x2, a_min=-limit, a_max=limit)
 
@@ -103,40 +105,39 @@ def tensor_sharded_gmm_merged_column_parallel(
     rhs: jax.Array,
     rhs_bias: jax.Array | None,
     group_sizes: jax.Array,
-    transpose_rhs: bool,
     mesh: Mesh,
-    intermediate_size: int,
-) -> jax.Array:
-    # adapted from https://github.com/pytorch/xla/blob/1d409399474197c484894be90b75d9855393dda5/torch_xla/experimental/custom_kernel.py#L1401
-    m, k, g = lhs.shape[0], lhs.shape[1], rhs.shape[0]
-    n = rhs.shape[1] if transpose_rhs else rhs.shape[2]
-    tm, tk, tn = _get_tiling_size_for_gmm_kernel(m // mesh.shape["data"], k, n,
-                                                 g)
+) -> tuple[jax.Array, jax.Array]:
 
-    _gmm = functools.partial(
-        gmm,
-        preferred_element_type=lhs.dtype,
-        tiling=(tm, tk, tn),
-        transpose_rhs=transpose_rhs,
-        group_offset=jnp.array(0),
-    )
+    def _gmm(lhs, rhs, group_sizes):
+        m, g, n, k = lhs.shape[0], *rhs.shape
+        tm, tk, tn = _get_tiling_size_for_gmm_kernel(m, k, n, g)
+        return gmm(
+            lhs,
+            rhs,
+            group_sizes,
+            preferred_element_type=lhs.dtype,
+            tiling=(tm, tk, tn),
+            transpose_rhs=True,
+            group_offset=jnp.array(0),
+        )
 
     gmm_result = shard_map(
         _gmm,
         mesh=mesh,
         in_specs=(P("data", None), P(None, "model", None), P("data")),
         out_specs=(P("data", "model")),
-        check_rep=False,
+        check_vma=False,
     )(lhs, rhs, group_sizes)
 
     if rhs_bias is not None:
 
         def _add_bias(gmm_result_local, rhs_bias_local, group_sizes_global):
-            rhs_bis = jnp.repeat(rhs_bias_local,
-                                 group_sizes_global,
-                                 0,
-                                 total_repeat_length=m // mesh.shape["data"])
-            return (gmm_result_local + rhs_bis).astype(gmm_result_local.dtype)
+            rhs_bias = jnp.repeat(
+                rhs_bias_local,
+                group_sizes_global,
+                0,
+                total_repeat_length=gmm_result_local.shape[0])
+            return gmm_result_local + rhs_bias
 
         gmm_result = shard_map(
             _add_bias,
@@ -144,12 +145,13 @@ def tensor_sharded_gmm_merged_column_parallel(
             in_specs=(P("data", "model"), P(None, "model"), P("data")),
             out_specs=(P("data", "model")),
         )(gmm_result, rhs_bias, group_sizes)
+    gmm_result = gmm_result.astype(lhs.dtype)
 
-    n_shards = mesh.shape["model"]
+    tp_size = mesh.shape["model"]
+    intermediate_size = gmm_result.shape[-1] // 2
     output_sizes = [intermediate_size, intermediate_size]
-
     return slice_sharded_tensor_for_concatenation(gmm_result, output_sizes,
-                                                  n_shards)
+                                                  tp_size)
 
 
 def tensor_sharded_gmm_row_parallel(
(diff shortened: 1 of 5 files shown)


tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -23,7 +23,10 @@ def activation_fn(activation: str, x1, x2):
                 f"FusedMoE does not support {activation} activation")
 
 
-def _swigluoai(x1, x2, alpha=1.702, limit=7.0):
+def _swigluoai(x1: jax.Array,
```
vanbasten23, line 25: could you paste the test commands for the 3 files in this PR?

kyuyeunk: Sure! Here they are

```
pytest -v -s tests/layers/vllm/test_unquantized.py
pytest -v -s tests/layers/vllm/test_mxfp4.py
```

Also, as I have linked in the PR description, I have verified that CI all passes: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/6118

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -103,96 +105,93 @@ def tensor_sharded_gmm_merged_column_parallel(
...
+            group_offset=jnp.array(0),
+        )
 
     gmm_result = shard_map(
         _gmm,
         mesh=mesh,
         in_specs=(P("data", None), P(None, "model", None), P("data")),
         out_specs=(P("data", "model")),
-        check_rep=False,
+        check_vma=False,
```
vanbasten23, line 129: why set it to False? Can it be true?

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -103,96 +105,93 @@ def tensor_sharded_gmm_merged_column_parallel(
...
             in_specs=(P("data", "model"), P(None, "model"), P("data")),
             out_specs=(P("data", "model")),
         )(gmm_result, rhs_bias, group_sizes)
+    gmm_result = gmm_result.astype(lhs.dtype)
 
-    n_shards = mesh.shape["model"]
-    output_sizes = [intermediate_size, intermediate_size]
-
+    tp_size = mesh.shape["model"]
+    output_sizes = [gmm_result.shape[-1] // 2] * 2
```
vanbasten23: i'd prefer the previous version which is much clearer `[intermediate_size, intermediate_size]`

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -201,42 +200,33 @@ def _add_bias(gmm_result_local, rhs_bias_local, group_sizes_global):
...
 
-    return gmm_result
+    return gmm_result.astype(lhs.dtype)
 
 
 def expert_sharded_gmm(
     lhs: jax.Array,
     rhs: jax.Array,
     group_sizes: jax.Array,
-    transpose_rhs: bool,
```
vanbasten23, line 211: thanks for simplifying it!

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -263,30 +253,20 @@ def _gmm(lhs, rhs, group_sizes, group_offset):
...
     # For i-th shard, it is responsible groups (AKA experts) from
     # i*num_experts_per_shard to (i+1)*num_experts_per_shard We sum them up to
     # get total rows in that shard, and that is the size for shard to send to
     # its peers. This is also the number of non-zero rows from the gmm results.
     # In the working example, send_sizes would be [3, 2, 5, 4]
-    send_sizes = jnp.array([
-        group_sizes[i * num_experts_per_shard:(i + 1) *
-                    num_experts_per_shard].sum() for i in range(ep_size)
-    ])
+    send_sizes = group_sizes.reshape(-1, num_experts_per_shard).sum()
```
vanbasten23: what's the shape of group_sizes before this line?

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: so for context, when using `shard_map`, it is really easy to mess up sharding.

For example, if we have mesh `{"data": 4, "model" 1}` and `shard_map`'s `out_specs=P(None, "model")`. Correct sharding means the data is replicated along `"data"` dimension. but with `shard_map`, programmer might make a mistake where data is different along `"data"` dimension. When using `check_vma=True`, compiler runs a check during compile time and ensure this doesn't happen.

But when using kernel, this check does not work because compiler does not have an inside view of the kernel. So we disable it.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: it's `[num_total_tokens]` where `num_total_tokens=num_tokens_per_shard * num_experts_per_shard`. So this reshape transforms it into `[num_tokens_per_shard, num_experts_per_shard]`.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: done. as long as we dont have to add unnecessary argument and infer from existing input / mesh shapes, it's all fine by me.

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -410,51 +409,41 @@ def _process_tokens_locally(hidden_states_local, topk_indices_local):
         mesh=mesh,
         in_specs=(P("data", None), P("data", None)),
         out_specs=(P("data", None), P("data"), P("data")),
-        check_rep=False,
     )(hidden_states, topk_indices)
+
     if use_ep:
         x = expert_sharded_gmm(
```
vanbasten23, line 419: Somehow I couldn't find a test that triggers this case (use_ep==True, line415). I tried "pytest tests/layers/vllm/test_unquantized.py -k test_fused_moe" but it didn't. Do you know which test exercise this case?

kyuyeunk: Here it is: https://github.com/vllm-project/tpu-inference/blob/77cb15682a4955f5ee5a90213b40e8885f930dc5/tests/layers/vllm/test_unquantized.py#L406

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -214,76 +216,56 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
 
-            # Reshape and transpose w13_weight to (num_experts, 2, hidden_size, intermediate_size)
             w13_reshaped = w13_weight.reshape(num_experts, 2,
                                               intermediate_size, hidden_size)
-            w13_weight_transposed = jnp.transpose(w13_reshaped, (0, 1, 3, 2))
 
-            # Transpose w2_weight to (num_experts, intermediate_size, hidden_size)
-            w2_weight_transposed = jnp.transpose(w2_weight, (0, 2, 1))
+            # Transpose non-constracting dim to right most dim
+            w13_weight_transposed = jnp.swapaxes(w13_reshaped, 2, 3)
```
vanbasten23, line 231: previously it's transpose. Why is it changed to swapaxes? They should to be different

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -214,76 +216,56 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
-                w2_bias = jax.device_put(
-                    w2_bias,
-                    Format(Layout((0, 1)),
-                           NamedSharding(self.mesh, P("model", None))))
-
+                    w13_bias, Format(Layout((0, 1, 2)), ep_sharding))
+                w2_bias = jax.device_put(w2_bias,
+                                         Format(Layout((0, 1)), ep_sharding))
         else:
-            # Original logic for non-kernel path
```
vanbasten23, line 260: could you leave the comment?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -168,7 +168,7 @@ def __init__(self,
                  ep_axis_name: str = 'model'):
         super().__init__(moe)
         self.mesh = mesh
-        self.use_kernel = envs.USE_MOE_EP_KERNEL
+        self.use_kernel = envs.USE_MOE_EP_KERNEL and moe.use_ep
```
vanbasten23, line 171: hm, if we don't use ep, why can't we use the fused moe kernel?

tpu_inference/layers/vllm/fused_moe.py
vanbasten23: hmm, I ran `pytest -s -v tests/layers/vllm/test_unquantized.py -k test_fused_moe` but I couldn't trigger that branch, even with use_ep==True. In the code, use_ep is always false so that branch is never triggered. Am I missing something?

You can stick a print and try yourself.

tpu_inference/layers/vllm/fused_moe.py
vanbasten23: Thanks for the context.

tpu_inference/layers/vllm/quantization/unquantized.py
kyuyeunk: `jnp.transpose(w13_reshaped, (0, 1, 3, 2))` and `jnp.swapaxes(w13_reshaped, 2, 3)` are the same things. The latter is design specifically for only transposing two axes - which is what `jnp.transpose` does by going from `(0, 1, 2, 3)` to `(0, 1, 3, 2)`. So using `swapaxes` is simpler.

tpu_inference/layers/vllm/quantization/unquantized.py
kyuyeunk: So i've deleted that because it was misleading. That path does utilize kernel - it's just a different kernel.

tpu_inference/layers/vllm/quantization/unquantized.py
kyuyeunk: The fused moe kernel is designed specifically for EP use case. If you look at the kernel code, there is no code path to support TP.

tpu_inference/layers/vllm/fused_moe.py
vanbasten23: Could you add a comment please?

Also, here it seems that it changes the previous approach? Previously, send_sizes is a 1d array of size `ep_size`. But right now, since you don't specify the axis, the send_sizes will be a scalar.

tpu_inference/layers/vllm/quantization/unquantized.py
yaochengji: the fused moe kernel only supports ep at the moment.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: You were right! And I was also able to catch easy to miss bug after I fixed the unit test. Thanks! And I've modified the unit test code.

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: Done. Also, your comment is correct and this was fixed after the unit test fix I've mentioned in another thread.

vanbasten23 (approved):

PR #1201, opened 2025-12-01
Save size in scalar scratch for bo and bq

rupengliu-meta:
# Description
In bo and bq, we could save size in smem to avoid calculation. This will  reduce unnecessary computation.
seems only having pretty minimal throughput improvement, but the improvement is consistently around 1%-2% before tuning!
Tests have passed for both kernels
截屏2025-12-01 上午10 52 38
截屏2025-12-01 上午11 11 47

FIXES: b/123456
FIXES: #123456

# Tests
Ran unit tests and done local e2e testing
commands to reproduce.

# Checklist


rupengliu-meta:
> Thanks for the contribution! I think the trade-off is between scalar computation and scalar load/store, do you have any performance number after the modification?

yes, I will update the perf numbers later

rupengliu-meta:
seems only having pretty minimal throughput improvement, but the improvement is consistently around 1%-2%. tested through the kernel benchmarking script (not e2e)

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -1487,8 +1529,8 @@ def ragged_paged_attention(
                                      dtype=kv_cache.dtype),
             ],
             input_output_aliases={
-                7: 0,
-                9: 1
+                8: 0,
```
vanbasten23, line 1532: do you know why `queries` is not in the donate_argnames in jax.jit?

rupeng-liu: it depends on if it will be used again after the attention. I took a quick look and didn't find where it is used again. So if no reuse of the queries, then we could donate it. @bythew3i thought?

rupeng-liu:
@kyuyeunk yep, good idea. I just checked the bkv sz, the sz is offset + bkv_sz_frm_new, which during wait is False, there is no existing value for this, we need to still do the extra calculation if added in the wait=false. So this might not be applicable for bkv?

kyuyeunk:
> > Thanks for the contribution! I think the trade-off is between scalar computation and scalar load/store, do you have any performance number after the modification?
> 
> 
> 
> yes, I will update the perf numbers later

Ping on updating perf numbers on the pr description.

rupengliu-meta:
> > > Thanks for the contribution! I think the trade-off is between scalar computation and scalar load/store, do you have any performance number after the modification?
> > 
> > 
> > yes, I will update the perf numbers later
> 
> Ping on updating perf numbers on the pr description.

Updated, thanks!

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -230,8 +230,9 @@ def _ragged_paged_attention_kernel(
     # TODO(jevinjiang): merge these into one so we can save SMEM.
     distribution_ref,  # [3] (decode_end, prefill_end, mixed_end)
     sem_ids_ref,  # [3] (bq_sem_idx, bkv_sem_idx, bo_sem_idx)
-    bo_ids_ref,  # [4] (bo_sem_0_seq_idx, bo_sem_1_seq_idx, bo_sem_0_bo_idx, bo_sem_1_bo_idx)
+    bo_ids_ref,  # [6] (bo_sem_0_seq_idx, bo_sem_1_seq_idx, bo_sem_0_bo_idx, bo_sem_1_bo_idx, bo_sem_0_sz, bo_sem_1_sz)
     bkv_update_ids_ref,  # [6] (bkv_sem_0_seq_idx, bkv_sem_1_seq_idx, bkv_sem_0_offset, bkv_sem_1_offset, bkv_sem_0_sz, bkv_sem_1_sz)
+    bq_fetch_ids_ref,  # [2] (bq_sem_0_sz, bq_sem_1_sz)
```
bythew3i, line 235: nit: just call `bq_ids_ref`

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -562,50 +563,89 @@ def loop_body(i, states):
...
+
+            _async_copy(
+                q_hbm_ref.at[:, pl.ds(q_len_start, sz)],
+                vmem_ref.at[:, pl.ds(0, sz)],
+                sem,
+                wait,
+            )
+        else:
+            # Retrieve sz from scratch instead of recalculating
+            sz = bq_fetch_ids_ref[bq_sem_idx]
```
bythew3i, line 595: Definitely need to retune and update the tuned block sizes.  I understand you may not have autotuen script. But please write a benchmarking script even with same block size, we want to see perf on different block sizes and different models. I am very strict with this in Google internal kernel development as well. We don't want to just check in the code without really understanding how much it can bring in different model(shapes) and block sizes.

bythew3i: Even appending throughput change on different models is acceptable. Thanks

kyuyeunk: @rupengliu-meta, is this still being worked on? If tuning is the issue, I can help you run the tuning using our internal tools.

PR #1209, opened 2025-12-02
Add a SP e2e test.

vanbasten23:
# Description

This PR adds a SP e2e test and adds it to the CI.

# Tests

pytest -s -vv tests/e2e/test_sequence_parallelism.py

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/common.py
+++ b/tpu_inference/layers/vllm/quantization/common.py
@@ -73,12 +73,12 @@ class JaxCommonLinearConfig:
             token_num = x.shape[0]
             # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
             if token_num // self.mesh.shape["model"] >= TPU_SECOND_LAST_MINOR:
-                logger.info(
+                logger.debug(
                     f"SP is enabled, returning non-None input sharding {self.input_sharding}. token_num // self.mesh.shape[model] >= TPU_SECOND_LAST_MINOR. {token_num=}, {self.mesh.shape['model']=}, {TPU_SECOND_LAST_MINOR=}."
                 )
                 return self.input_sharding
             else:
-                logger.info(
+                logger.debug(
                     f"SP is enabled, returning input sharding None. token_num // self.mesh.shape[model] < TPU_SECOND_LAST_MINOR. {token_num=}, {self.mesh.shape['model']=}, {TPU_SECOND_LAST_MINOR=}."
                 )
                 return None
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/vllm/quantization/common.py
```diff
@@ -73,8 +73,14 @@ def get_input_sharding(self, x: torchax.tensor.Tensor):
             token_num = x.shape[0]
             # NOTE(chengjiyao): make sure the sharded token_num is larger than TPU_SECOND_LAST_MINOR
             if token_num // self.mesh.shape["model"] >= TPU_SECOND_LAST_MINOR:
+                logger.info(
```
yaochengji: In our e2e test, should we check the log to see sp is actually enabled? Another way is to check the final optimized graph, but that's more difficult.

vanbasten23: I checked manually that this line(76) is executed. But I think you meant if we can check it in the test. I'm not sure how we can do that in the test. Let me know if you have some ideas.

Though, I added a long prompt (line24 ""Three Rings...") to ensure `token_num//8 >= 8` is triggered. Also the precompilation phase use very large num_tokens  so this case is triggered.

yaochengji: Can we output the log in a file and checked the file's content later?

tpu_inference/layers/vllm/quantization/common.py
```diff
@@ -83,8 +89,14 @@ def get_output_sharding(self, x: torchax.tensor.Tensor):
...
```
kyuyeunk: `SP is enabled, but token_num // self.mesh.shape[\"model\"] >= TPU_SECOND_LAST_MINOR` the inclusion of the word "but" make it sounds like something negative has happened - i.e., sp was disabled due to some limitation - but that isn't the case right?

Also, I don't think we need to print log for this behavior - only print logs for `else` case - where SP is disabled - which is something that user might not have expected to happen.

tests/e2e/test_sequence_parallelism.py
```diff
@@ -0,0 +1,116 @@
...
+        enable_prefix_caching=enable_prefix_caching,
+        additional_config=additional_config,
+        kv_cache_dtype=kv_cache_dtype,
+        async_scheduling=async_scheduling,
+    )
+
+    engine_args_dict = asdict(engine_args)
+    llm = LLM(**engine_args_dict)
+
+    try:
```
kyuyeunk, line 81: any reason for wrapping this function call with try / finally?

vanbasten23: good question. It's needed because when there are multiple cases, there is a possibility that TPU resources cannot be released between tests. The stuff in the finally clause mitigates it.

tpu_inference/layers/vllm/quantization/common.py
vanbasten23: Agreed with the wording. I improved it. Though we do need both cases for now because we want to ensure the sp sharding case is triggered.

.buildkite/pipeline_jax.yml
```diff
@@ -262,6 +262,19 @@ steps:
        - |
          .buildkite/scripts/run_in_docker.sh \
            bash -c 'python3 -m pytest -s -v -x /workspace/tpu_inference/tests/lora/test_layers.py'
+
+   - label: "E2E sequence parallelism test"
```
py4: Can you please not add it here but add it here: https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/parallelism/SP.yml
so it runs in nightly but not per commit?

vanbasten23: Was not aware of it. Thanks for pointing it out.

tpu_inference/layers/vllm/quantization/common.py
yaochengji: Could it be `logger.debug` instead of `logger.info`?

vanbasten23: No luck for now. Somehow I couldn't capture the logger. I agree that this may be the easiest way to test. But if someone write a similar logging string, then this test may not work as intended. Other parallelism seem to have the same issue: it's hard to examine how each layer is sharded in the test.

I've been thinking if there is better way to test. Since SP's main benefits is to reduce memory, we can check with SP if the mem usage is indeed reduced. But there is no jax api that let me check the mem usage.

How about let's merge this pr so that it verifies with "enable_sequence_parallelism=True" and "tensor_parallelism=8" the test runs to completion, since that is the intended way to enable SP. Then when we do integration, we improve the test. Wdyt?

PR #1212, opened 2025-12-02
Fix dp sharding for compute_logits_func

kyuyeunk:
# Description

the past.

FIXES: b/123456
FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -221,7 +221,7 @@ class VllmModelWrapper:
         @functools.partial(
             jax.jit,
             out_shardings=(NamedSharding(self.mesh,
-                                         PartitionSpec(None, "model"))),
+                                         PartitionSpec("data", "model"))),
         )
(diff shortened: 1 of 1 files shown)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -221,7 +221,7 @@ def jit_compute_logits_func(self):
         @functools.partial(
             jax.jit,
             out_shardings=(NamedSharding(self.mesh,
-                                         PartitionSpec(None, "model"))),
+                                         PartitionSpec("data", "model"))),
```
py4, line 224: DP does not work with jax path? we should not do the same in jax path?

wenxindongwork: we are already using this sharding in the jax path: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/common/model_loader.py#L247-L258

wenxindongwork (approved):

PR #1235, opened 2025-12-03
Fix FP8 dtype type mismatch issue

helloworld1:
# Description

j2t_dtype expects jnp.dtype, not _ScalarMeta. This will lead to dtype mismatch when FP8 is used.

# Tests

Tested 
```
vllm serve unsloth/gpt-oss-120b-BF16 --max-model-len=16384 --max-num-batched-tokens=8192 --max-num-seqs=2048 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=4 --download-dir=/mnt/pd --async-scheduling --port=8000 --kv-cache-dtype=fp8
```

# Checklist


kyuyeunk:
Hi @helloworld1, can you update the branch, and verify all checks passes before we can merge it?

PR #1236, opened 2025-12-03
[multihost] Add NEW_MODEL_DESIGN to additional_env_vars

Lumosis:
# Description

Add NEW_MODEL_DESIGN to additional_env_vars since NEW_MODEL_DESIGN is not deprecated yet

# Tests

commands to reproduce.

# Checklist


Lumosis:
> Might be dumb question, but we also have this file which keeps track of all the environment variables used in our code base: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/envs.py
> 
> Is it possible to use that instead?

I think this is orthogonal. This is for propagating the env variables for multi-host

PR #1237, opened 2025-12-03
[Disagg] local disagg e2e test

sixiang-google:
# Description

Disagg e2e test. 

FIXES: b/123456
FIXES: #123456

# Tests

 pytest --capture=no tests/e2e/test_local_disagg.py

# Checklist

### DIFF
--- a/.buildkite/features/Single-Host-P-D-disaggregation.yml
+++ b/.buildkite/features/Single-Host-P-D-disaggregation.yml
@@ -1,0 +1,25 @@
+# Single-Host-P-D-disaggregation
+# features support matrix
+steps:
+  - label: "Correctness tests for Single-Host P-D disaggregation"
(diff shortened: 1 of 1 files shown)


tests/e2e/test_local_disagg.py
```diff
@@ -0,0 +1,119 @@
+# SPDX-License-Identifier: Apache-2.0
```
vipannalla, line 1: This is great! Can you also add a correctness test comparing output of disagg setup with that of non-disagg?

vipannalla: This is similar to what DP test does -- https://github.com/vllm-project/tpu-inference/blob/main/tests/e2e/test_data_parallel.py#L171

sixiang-google: Added

PR #1243, opened 2025-12-04
update run_in_docker script for running on local env

ernie-chang:
# Description

The run_in_docker script only run successfully on buildkite env before, so we update run_in_docker script, let users can  run the script on their own local env too.

# Tests

test on local v6e env [log](https://paste.googleplex.com/4653076386545664)

# Checklist

### DIFF
--- a/support_matrices/nightly/feature_support_matrix.csv
+++ b/support_matrices/nightly/feature_support_matrix.csv
@@ -5,13 +5,13 @@ Feature,CorrectnessTest,PerformanceTest
 "Llama 4 Maverick",to be added,to be added
 "LoRA_Torch",❌,N/A
 "Multimodal Inputs",✅,❌
-"Out-of-tree model support",✅,✅
+"Out-of-tree model support",❌,N/A
 "Prefix Caching",✅,✅
 "Single Program Multi Data",✅,✅
 "Single-Host-P-D-disaggregation",N/A,N/A
 "Speculative Decoding: Eagle3",❌,N/A
-"Speculative Decoding: Ngram",✅,✅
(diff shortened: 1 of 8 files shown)


.buildkite/scripts/setup_docker_env.sh
```diff
@@ -58,8 +52,12 @@ setup_environment() {
   echo "Installing Python dependencies"
   python3 -m pip install --progress-bar off buildkite-test-collector==0.1.9
   echo "Python dependencies installed"
-
-  VLLM_COMMIT_HASH=$(buildkite-agent meta-data get "VLLM_COMMIT_HASH" --default "")
+  if [ -z "${BUILDKITE:-}" ]; then
+      VLLM_COMMIT_HASH=""
+      BUILDKITE_COMMIT=$(git log -n 1 --pretty="%H")
```
QiliangCui: nit: here use TPU_INFERENCE_HASH.

then the logic of the if else block should be:

```
if "not BUILDKITE":
   VLLM_COMMIT_HASH=""
   TPU_INFERENCE_HASH = $(git log -n 1 --pretty="%H")
else:
    VLLM_COMMIT_HASH=$(buildkite-agent meta-data get "VLLM_COMMIT_HASH" --default "")
   TPU_INFERENCE_HASH=BUILDKITE_COMMIT

```

then, in line 64 where BUILDKITE_COMMIT is used, change it to use TPU_INFERENCE_HASH.

This makes the if-else clear and not confusing the name of "buildkite commit".

QiliangCui (changes requested):

PR #1252, opened 2025-12-05
[Kernel][FusedMoE] Fix MoE crash and hang issues

bythew3i:
... also move weight padding to weight load instead of inside kernel.

# Root Cause:

1. TopK on padded gating scores could return out of range topK index
2. Missing sync_barrier to make sure all devices finish metadata propagation before a2a scatter.
3. Large `bt` (block num tokens) caused OOB read and write

# TODO:
- Tune the block sizes  in MoE for GPT-OSS

### DIFF
--- a/tpu_inference/kernels/fused_moe/v1/kernel.py
+++ b/tpu_inference/kernels/fused_moe/v1/kernel.py
@@ -19,7 +19,7 @@ def align_to(x, a):
 
 
 def get_dtype_packing(dtype):
-    bits = dtypes.itemsize_bits(dtype)
+    bits = dtypes.bit_width(dtype)
     return 32 // bits
 
 


tpu_inference/runner/kv_cache.py
kyuyeunk, line 1: we need a different PR for this. I found out it's not simple replacement because this function doesn't exist in order jax versions and causes crash.

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -336,6 +377,11 @@ def apply(
         gating_output = jax_view(router_logits)
 
         if self.use_kernel and layer.use_ep:
```
kyuyeunk: I think this branch is outdated - because i see old code. can you update again?

tpu_inference/runner/kv_cache.py
bythew3i: Yes, agree, looks like we need to bump jax version in tpu-inference. Google internal dev env is always using latest jax

tpu_inference/layers/vllm/quantization/unquantized.py
bythew3i: Synced to main. I do not see any difference or conflict here with main branch

tpu_inference/runner/kv_cache.py
bythew3i: Reverts the item_width change

tpu_inference/kernels/fused_moe/v1/kernel.py
```diff
@@ -19,7 +19,7 @@ def align_to(x, a):
 
 
 def get_dtype_packing(dtype):
-    bits = dtypes.bit_width(dtype)
+    bits = dtypes.itemsize_bits(dtype)
```
kyuyeunk: This change should also be removed

kyuyeunk:
@bythew3i Can you update the branch & resolve the conflict?

PR #1255, opened 2025-12-05
Add  default 'auto' MODEL_IMPL_TYPE that resolves based on architecture

xingliu14:
# Description

- Add `auto` as default value for MODEL_IMPL_TYPE env var
- For GptOssForCausalLM, `auto` resolves to `vllm` for better performance
- For all other architectures, 'auto' resolves to `flax_nnx` for better performance
- Add tests for 'auto' resolution behavior

# Tests

pytest

```
tests/test_envs.py
tests/models/common/test_model_loader.py
```

# Checklist

### DIFF
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -27,7 +27,8 @@ _MODEL_REGISTRY = {}
 # Architectures that prefer "vllm" implementation type when MODEL_IMPL_TYPE is "auto".
 # These architectures are listed here because they have better performance with the
 # vLLM PyTorch backend compared to the flax_nnx JAX backend for now.
-_VLLM_REQUIRED_ARCHITECTURES: frozenset[str] = frozenset({"GptOssForCausalLM"})
+_VLLM_PREFERRED_ARCHITECTURES: frozenset[str] = frozenset(
+    {"GptOssForCausalLM"})
 
 
 class UnsupportedArchitectureError(ValueError):
@@ -355,7 +356,7 @@ def get_model(
             f"Expected exactly one architecture, got {len(architectures)}: "
             f"{architectures}")
         arch = architectures[0]
-        impl = "vllm" if arch in _VLLM_REQUIRED_ARCHITECTURES else "flax_nnx"
+        impl = "vllm" if arch in _VLLM_PREFERRED_ARCHITECTURES else "flax_nnx"
         logger.info(f"Resolved MODEL_IMPL_TYPE 'auto' to '{impl}'")
 
     match impl:


xingliu14:
@kyuyeunk Please review.

kyuyeunk (commented):
Wouldn't it be possible move 'auto' to 'match/case' as well?

xingliu14:
It is possible to move it in to match-case, but in that case it will have duplicated codes, including: get_vllm_model, get_flax_model and the fall back check. I think resolve first then use the same code will be more clean.

tpu_inference/models/common/model_loader.py
```diff
@@ -333,6 +333,12 @@ def get_vllm_model(
     return jit_model, compute_logits_fn, combine_hidden_states_fn, None, params, lora_manager, model
 
 
+# Architectures that require "vllm" implementation type when MODEL_IMPL_TYPE is "auto".
```
kyuyeunk: "require" might be too strong word. replace it with "prefer"

tpu_inference/models/common/model_loader.py
```diff
@@ -333,6 +333,12 @@ def get_vllm_model(
...
+# These architectures are listed here because they have better performance with the
+# vLLM PyTorch backend compared to the flax_nnx JAX backend for now.
+_VLLM_REQUIRED_ARCHITECTURES: frozenset[str] = frozenset({"GptOssForCausalLM"})
```
kyuyeunk: In general, this kind of constants should be placed at the start of the file. Please move it.

tpu_inference/models/common/model_loader.py
```diff
@@ -342,24 +348,37 @@ def get_model(
...
-            error_msg = str(e)
-
-            logger.warning(error_msg)
-
-            # Fall back to the vLLM model and updating the dtype accordingly
-            vllm_config.model_config.dtype = j2t_dtype(
-                vllm_config.model_config.dtype.dtype)
+    if impl == "auto":
+        # Resolve "auto" based on architecture
+        architectures = getattr(vllm_config.model_config.hf_config,
```
kyuyeunk, line 353: dumb question: is there a cases where there's a multiple "architectures" for a single model?

tpu_inference/models/common/model_loader.py
```diff
@@ -342,24 +348,37 @@ def get_model(
...
+                                "architectures", [])
+        for arch in architectures:
```
kyuyeunk: similar to above comment. can we just to an assert to check if `len(architectures)==1` and do a simple hash map fetch instead of iterating for loop?

tpu_inference/models/common/model_loader.py
```diff
@@ -342,24 +348,37 @@ def get_model(
...
+        else:
+            impl = "flax_nnx"
+        logger.info(f"Resolved MODEL_IMPL_TYPE 'auto' to '{impl}'")
+
+    match impl:
+        case "flax_nnx":
+            try:
+                # Try to load the flax model first
+                return get_flax_model(vllm_config, rng, mesh, is_draft_model)
+            except UnsupportedArchitectureError as e:
```
kyuyeunk, line 367: probably a nit question: in c's switch statements, if we don't put `break;`, it will automatically invoke next case. Is it not the case for python's match/case? I.e., if `UnsupportedArchitectureError` is thrown, we skip `break;` statement and automatically let the next case (which is `case "vllm"`) to be invoke.

tpu_inference/models/common/model_loader.py
xingliu14: Done.

tpu_inference/models/common/model_loader.py
xingliu14: Done.

tpu_inference/models/common/model_loader.py
xingliu14: Done.

kyuyeunk:
I don't see any update? Can you verify if the changes were pushed?

tpu_inference/models/common/model_loader.py
xingliu14: For python, per https://docs.python.org/3/tutorial/controlflow.html#match-statements: "Only the first pattern that matches gets executed".

tpu_inference/models/common/model_loader.py
xingliu14: Theorically it can be since it's a list, but I don't find any example in the codebase. The above code treat it as a list as well: https://github.com/vllm-project/tpu-inference/blob/774cdbe2e993dd43b02947f5021d3f46d1437347/tpu_inference/models/common/model_loader.py#L56-L59. But I'll change it to assert it len=1.

kyuyeunk (approved):
LGTM. Thank you for working on it!

kyuyeunk:
seems like CI is failing. Have you rebase the branch to HEAD?

xingliu14:
I believe so, let me look into it.

xingliu14:
Looks like tests/layers/vllm/test_awq.py is failing from the main branch as well.

kyuyeunk:
seems like it's due to an upstream change. let me create a quick fix for this.

kyuyeunk:
Please wait until this is merged: https://github.com/vllm-project/tpu-inference/pull/1284

kyuyeunk:
the PR has been merged. please update the branch and try again.

tpu_inference/models/common/model_loader.py
```diff
@@ -24,6 +24,11 @@
 
 _MODEL_REGISTRY = {}
 
+# Architectures that prefer "vllm" implementation type when MODEL_IMPL_TYPE is "auto".
+# These architectures are listed here because they have better performance with the
+# vLLM PyTorch backend compared to the flax_nnx JAX backend for now.
+_VLLM_REQUIRED_ARCHITECTURES: frozenset[str] = frozenset({"GptOssForCausalLM"})
```
kyuyeunk: please also update this `REQUIRED` to `PREFERRED` as well.

xingliu14: Done.

kyuyeunk:
https://github.com/vllm-project/tpu-inference/actions/runs/20125295806/job/57753569146?pr=1255

```
pre-commit hook(s) made changes.
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/tpu_inference/models/common/model_loader.py b/tpu_inference/models/common/model_loader.py
index f260e5f..fb035bd 100644
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -27,7 +27,8 @@ _MODEL_REGISTRY = {}
 # Architectures that prefer "vllm" implementation type when MODEL_IMPL_TYPE is "auto".
 # These architectures are listed here because they have better performance with the
 # vLLM PyTorch backend compared to the flax_nnx JAX backend for now.
-_VLLM_PREFERRED_ARCHITECTURES: frozenset[str] = frozenset({"GptOssForCausalLM"})
+_VLLM_PREFERRED_ARCHITECTURES: frozenset[str] = frozenset(
+    {"GptOssForCausalLM"})
 
 
 class UnsupportedArchitectureError(ValueError):
Error: Process completed with exit code 1.
```

Please fix pre-commit failure.

kyuyeunk:
Thank you so much for making this feature!

PR #1264, opened 2025-12-08
Replacing bit_width() with itemized_bits(). 

aman2930:
# Description
In latest version of Jax bit_width() is renamed to itemized_bits().
This change was not compatible with older versions. Currently tests using latest JAX are failing.

# Checklist

### DIFF
--- a/tests/kernels/quantized_matmul_kernel_test.py
+++ b/tests/kernels/quantized_matmul_kernel_test.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
-import functools
-
 import jax
 import jax.numpy as jnp
 from absl.testing import absltest, parameterized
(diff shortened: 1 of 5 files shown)


tests/kernels/ragged_paged_attention_kernel_v3_hd64_test.py
```diff
@@ -176,7 +176,9 @@ def gen_random(shape, dtype):
         )
         output = output[:cu_q_lens[distribution[-1]]]
 
-        dtype_bits = dtypes.bit_width(jnp.dtype(kv_dtype))
+        dtype_bits = (dtypes.bit_width(jnp.dtype(kv_dtype)) if hasattr(
```
kyuyeunk, line 179: It seems like this change is quite a lot? Maybe you can move these functionality into some helper function?

PR #1266, opened 2025-12-08
Fix TPU7x chip counting to account for chiplet architecture

burbajr:
**Description**

For TPU7x devices, each physical chip contains 2 chiplets that are exposed to the host as separate devices. The previous implementation counted these devices directly, resulting in **2x the actual chip count** being reported in logs.

**Problem Being Solved**
When running on tpu7x-8 hardware (which has 4 physical chips with 2 chiplets each), the system would log:
TPU info: ... | num_chips=8 | num_cores_per_chip=2

This was misleading because it reported 8 chips when there are only 4 physical chips. Each chip has 2 chiplets exposed as separate devices to the host.

**Solution**
This PR adds chiplet-aware logic to the chip counting mechanism:

1. **New helper function** `get_num_chiplets_per_chip()` that returns 2 for tpu7x devices and 1 for all other TPU types
2. **Modified** `get_num_chips()` to divide the device count by chiplets per chip
3. **Enhanced logging** to show `num_chiplets_per_chip` for tpu7x devices only (to avoid confusing non-tpu7x users)

**Why This is a Good Solution**
- **Follows existing patterns**: Uses the same pattern as `get_num_cores_per_chip()` for consistency
- **Backward compatible**: All non-tpu7x devices get `chiplets_per_chip=1`, making the division a no-op
- **Minimal code changes**: Only touches the necessary functions
- **Well tested**: Comprehensive test coverage for all scenarios

**Implementation Details**
The fix uses integer division (`//`) which is safe because:
- tpu7x-8: 8 devices // 2 chiplets = 4 chips ✓
- tpu7x-4: 4 devices // 2 chiplets = 2 chips ✓
- v5e/v6e: N devices // 1 chiplet = N chips ✓ (unchanged)

Logging is conditional - chiplet info only appears for tpu7x:
tpu7x-8 (after fix):
TPU info: ... | num_chips=4 | num_chiplets_per_chip=2 | num_cores_per_chip=2
v6e-8 (unchanged, no chiplet info to avoid confusion):
TPU info: ... | num_chips=8 | num_cores_per_chip=1

**Test Coverage Added**
- `test_get_num_chiplets_per_chip`: Tests all TPU types including tpu7x-8, tpu7x-4, v6e-8, v5litepod, and edge cases (None, empty string)
- `test_get_num_chips_tpu7x_from_accel`: Verifies tpu7x-8 with 8 `/dev/accel*` devices returns 4 chips
- `test_get_num_chips_tpu7x_4_from_accel`: Verifies tpu7x-4 with 4 `/dev/accel*` devices returns 2 chips  
- `test_get_num_chips_tpu7x_from_vfio`: Verifies tpu7x-8 with `/dev/vfio` path returns 4 chips
- `test_get_num_chips_non_tpu7x_unchanged`: Verifies v6e-8 still returns 8 chips (backward compatibility)

**How to Test**
pytest tests/test_tpu_info.py -v

**Run specific new tests**
pytest tests/test_tpu_info.py::test_get_num_chiplets_per_chip -v
pytest tests/test_tpu_info.py::test_get_num_chips_tpu7x_from_accel -v
pytest tests/test_tpu_info.py::test_get_num_chips_non_tpu7x_unchanged -v
All tests pass with the changes.

### DIFF
--- a/tests/test_tpu_info.py
+++ b/tests/test_tpu_info.py
@@ -99,22 +99,38 @@ def test_get_num_cores_per_chip(mock_get_tpu_type, tpu_type, expected):
 
 
 # Test get_num_chips
+@patch("tpu_inference.tpu_info.get_tpu_type", return_value="v6e-8")
 @patch("tpu_inference.tpu_info.glob.glob",
        return_value=["/dev/accel0", "/dev/accel1"])
-def test_get_num_chips_from_accel(mock_glob):
+def test_get_num_chips_from_accel(mock_glob, mock_get_tpu_type):
     """Test get_num_chips when /dev/accel* files exist."""
     assert get_num_chips() == 2
 
 
+@patch("tpu_inference.tpu_info.get_tpu_type", return_value="v6e-8")
 @patch("tpu_inference.tpu_info.glob.glob", return_value=[])
 @patch("tpu_inference.tpu_info.os.listdir", return_value=["0", "1", "2"])
-def test_get_num_chips_from_vfio(mock_listdir, mock_glob):
+def test_get_num_chips_from_vfio(mock_listdir, mock_glob, mock_get_tpu_type):
     """Test get_num_chips when /dev/accel* files don't exist but /dev/vfio entries do."""
     assert get_num_chips() == 3
 
 
+@patch("tpu_inference.tpu_info.get_tpu_type", return_value="v6e-8")
 @patch("tpu_inference.tpu_info.glob.glob", return_value=[])
 @patch("tpu_inference.tpu_info.os.listdir", side_effect=FileNotFoundError)
-def test_get_num_chips_not_found(mock_listdir, mock_glob, caplog):
+def test_get_num_chips_not_found(mock_listdir, mock_glob, mock_get_tpu_type,
+                                 caplog):
     """Test get_num_chips when neither files nor directory are found."""
     assert get_num_chips() == 0
+
+
+# Test get_num_chips with tpu7x
+@patch("tpu_inference.tpu_info.get_tpu_type", return_value="tpu7x-8")
+@patch("tpu_inference.tpu_info.glob.glob",
+       return_value=[
+           "/dev/accel0", "/dev/accel1", "/dev/accel2", "/dev/accel3",
+           "/dev/accel4", "/dev/accel5", "/dev/accel6", "/dev/accel7"
+       ])
+def test_get_num_chips_tpu7x(mock_glob, mock_get_tpu_type):
+    """Test get_num_chips for tpu7x divides by 2 (8 devices = 4 chips)."""
+    assert get_num_chips() == 4
--- a/tpu_inference/tpu_info.py
+++ b/tpu_inference/tpu_info.py
@@ -66,13 +66,20 @@ def get_num_cores_per_chip() -> int:
 def get_num_chips() -> int:
     accel_files = glob.glob("/dev/accel*")
     if accel_files:
-        return len(accel_files)
-    try:
-        vfio_entries = os.listdir("/dev/vfio")
-        numeric_entries = [
-            int(entry) for entry in vfio_entries if entry.isdigit()
-        ]
-        return len(numeric_entries)
-    except FileNotFoundError as e:
-        logger.error("Failed to detect number of TPUs: %s", e)
-        return 0
+        num_devices = len(accel_files)
+    else:
+        try:
+            vfio_entries = os.listdir("/dev/vfio")
+            numeric_entries = [
+                int(entry) for entry in vfio_entries if entry.isdigit()
+            ]
+            num_devices = len(numeric_entries)
+        except FileNotFoundError as e:
+            logger.error("Failed to detect number of TPUs: %s", e)
+            return 0
+
+    # For tpu7x, each chip has 2 chiplets exposed as separate devices
+    tpu_type = get_tpu_type()
+    if tpu_type and "tpu7x" in tpu_type.lower():
+        return num_devices // 2
+    return num_devices


tests/test_tpu_info.py
```diff
@@ -98,23 +99,105 @@ def test_get_num_cores_per_chip(mock_get_tpu_type, tpu_type, expected):
...
 @patch("tpu_inference.tpu_info.glob.glob", return_value=[])
 @patch("tpu_inference.tpu_info.os.listdir", side_effect=FileNotFoundError)
-def test_get_num_chips_not_found(mock_listdir, mock_glob, caplog):
+def test_get_num_chips_not_found(mock_listdir, mock_glob, mock_get_tpu_type,
+                                 caplog):
     """Test get_num_chips when neither files nor directory are found."""
     assert get_num_chips() == 0
+
+
+# Test get_num_chips with tpu7x
```
kyuyeunk, line 127: I feel like these newly created tests can be merged into a single test?

tpu_inference/__init__.py
```diff
@@ -42,10 +42,15 @@
...
+        # For tpu7x, show chiplets_per_chip since each chip has 2 chiplets.
+        # Other TPU types don't have chiplets, so we don't show this info.
+        chiplets = f" | num_chiplets_per_chip={ti.get_num_chiplets_per_chip()}" if (
+            tpu_type and "tpu7x" in tpu_type.lower()) else ""
         logger.info(f"TPU info: node_name={ti.get_node_name()} | "
-                    f"tpu_type={ti.get_tpu_type()} | "
+                    f"tpu_type={tpu_type} | "
                     f"worker_id={ti.get_node_worker_id()} | "
-                    f"num_chips={ti.get_num_chips()} | "
+                    f"num_chips={ti.get_num_chips()}{chiplets} | "
```
kyuyeunk: I think the better option is to just maintain logic as-is and just rename `num_chips` and `get_num_chips()` into `num_cores` and `get_num_cores()`?

burbajr: Thank you for the feedback.  I would propose that we remove introducing chiplets from this PR and just focus on reporting the correct number of chips for v7 TPUs.  This would only require a few changes to tpu_info.py, add one test to test_tpu_info.py, and no changes to __init__.py.  If that sounds reasonable I can update this PR with those changes for review. Thank you!

kyuyeunk: that sounds good to me! please update the PR as you've mentioned!

tpu_inference/tpu_info.py
```diff
@@ -66,13 +66,20 @@ def get_num_cores_per_chip() -> int:
...
+            ]
+            num_devices = len(numeric_entries)
+        except FileNotFoundError as e:
+            logger.error("Failed to detect number of TPUs: %s", e)
+            return 0
+
+    # For tpu7x, each chip has 2 chiplets exposed as separate devices
+    tpu_type = get_tpu_type()
+    if tpu_type and "tpu7x" in tpu_type.lower():
+        return num_devices // 2
```
kyuyeunk, line 84: not really a fan of hard coding the value "2" or having "tpu7x" specific logic.

Is there any logic within tpu_info that programatically fetches num_devices?

Or wouldn't a simple `len(jax.devices())` return correct number of cores - regardless of having 1 chiplet or 1 chiplets?

burbajr:
Updated implementation based on feedback to use a simpler approach:
- Removed get_num_chiplets_per_chip() helper function
- Simplified get_num_chips() with inline tpu7x check
- Reverted logging changes to original simple format
- Kept single focused test for tpu7x chip counting

tpu_inference/tpu_info.py
burbajr: Hi, and thank you for taking a closer look at this.

The original logic (below) for determining the number of cores per TPU family remains correct for v7, which has 2 cores per chip, so no changes are needed there.

_def get_num_cores_per_chip() -> int:
    tpu_type = get_tpu_type()
    if tpu_type.startswith(("v5litepod", "v6e")):
        return 1
    return 2_

The issue is only with chip count inference. Prior to v7, each physical chip (regardless of core counts) exposed a single logical device, so using jax.devices() or the OS device list was a reliable way to infer the number of chips, since there was a 1:1 mapping between physical chips and logical devices. The original logic in tpu_info simply counts the number of host devices to report the number of chips.

With v7, each physical chip now exposes two logical devices, which means len(jax.devices()) and the OS device list report twice as many devices as there are actual chips, leading to the current reporting bug. 

I was not able to identify a reliable programmatic way to retrieve the physical TPU topology on TPU/GKE nodes, so dividing the device count by two is the most accurate approach I have found for correcting the reported chip count.

burbajr: Hi, curious if anything is still outstanding on this PR or if you have any questions. Thank you!

tests/test_tpu_info.py
```diff
@@ -99,22 +99,38 @@ def test_get_num_cores_per_chip(mock_get_tpu_type, tpu_type, expected):
...
+@patch("tpu_inference.tpu_info.get_tpu_type", return_value="tpu7x-8")
+@patch("tpu_inference.tpu_info.glob.glob",
+       return_value=[
+           "/dev/accel0", "/dev/accel1", "/dev/accel2", "/dev/accel3",
+           "/dev/accel4", "/dev/accel5", "/dev/accel6", "/dev/accel7"
+       ])
+def test_get_num_chips_tpu7x(mock_glob, mock_get_tpu_type):
```
kyuyeunk, line 134: instead of creating a separate unit test for tpu7x, can you use existing unit test but add parameter for tpu7x?

tpu_inference/tpu_info.py
kyuyeunk: I feel this change feels unnecessary complicated? I think just changing our wording from "num chips" to "num cores" would be sufficient.

burbajr: Hi, thanks for taking another look at this. Let me try to explain the motivation a bit more clearly.

The existing / original logic already hard codes the number of cores per chip based on the TPU family prefix, and that part is correct and unchanged. v5litepod and v6e have 1 core per chip, and other families have 2 cores per chip.

_def get_num_cores_per_chip() -> int:
    tpu_type = get_tpu_type()
    if tpu_type.startswith(("v5litepod", "v6e")):
        return 1
    return 2_
    
Where things get tricky is inferring the total number of chips or total number of cores from the number of JAX or OS visible devices. That mapping is no longer consistent across TPU families.

For example, using the same 2x2x1 topology:

v5p
4 chips, 2 cores per chip, 8 cores total
This exposes **4** JAX or OS devices

v7
4 chips, 2 cores per chip, 8 cores total
This exposes **8** JAX or OS devices

So simply counting len(jax.devices()) gives different results depending on the TPU generation, even though the physical topology and total core count are the same. Because of that, changing the field to represent cores without additional logic would still be inaccurate in some cases.

I do not have a strong opinion on the final approach. Whether we preserve chip semantics, rename fields, or restructure how this is reported, I mainly wanted to point out that the current logic cannot reliably calculate either total chips or total cores if it relies only on device count.

Happy to align on whatever direction makes the most sense here, and thanks for taking a look.

kyuyeunk: what you are describing would be megacore. so if a megacore is enabled, 2 cores act like a 1 core. If it's disabled or not used at all, 2 cores act like 2 cores.

Reference: https://docs.jax.dev/en/latest/pallas/tpu/pipelining.html#tpus-in-megacore-configuration

Instead of hardcoding this logic, can you figure out a way to programatically determined if megacore is used or not and if it's used, treat 2 cores as a 1 core and if it's not used, treat 2 cores as 2 cores.

kyuyeunk: note that you can just use tpu_info's `devices_per_chip` logic: https://github.com/AI-Hypercomputer/cloud-accelerator-diagnostics/blob/c94f463dd3bf99eaecf4f5af46f87cc2a5ff1b22/tpu_info/tpu_info/device.py#L56-L62

PR #1273, opened 2025-12-09
[Quantization] Add option to bypass quantized matmul kernel for W8A8-FP8 Compressed Tensors

jrplatin:
# Description

In this PR, I introduce a `DISALBE_QUANTIZED_MATMUL_KERNEL` flag which will bypass the quantized matmul kernel in the TorchAX path's W8A8-FP8 Compressed Tensors module.  This is needed because the sharded matmul was observed to introduce NaNs when using a quantized KV cache with Qwen3-Coder-480B

Previously, with the matmul kernel, we saw the following result when running the following command:

```
python examples/offline_inference/llm_engine_example.py  --tensor-parallel-size=8 --model BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic --kv-cache-dtype=fp8
```

Result:
```
Prompt: 'A robot may not injure a human being'
Generated text: '!!!!!!!!!!!!!!!!'
Prompt: 'To be or not to be,'
Generated text: '!!!!!!!!!!!!!!!!'
Prompt: 'What is the meaning of life?'
Generated text: '!!!!!!!!!!!!!!!!'
```

# Tests

## Accuracy
Tested locally + added relevant env var tests.

Also tested accuracy with the following command:

```
lm_eval \
    --model vllm \
    --model_args '{"pretrained": "BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic", "tensor_parallel_size": 8, "max_model_len": 2048, "max_num_batched_tokens": 2048, "max_gen_toks": 256, "kv_cache_dtype": "fp8"}' \
    --tasks gsm8k_cot \
    --batch_size auto \
    --apply_chat_template \
    --num_fewshot 8
```

v7x-8 result:
```
|  Tasks  |Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|---------|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k_cot|      3|flexible-extract|     8|exact_match|↑  |0.8825|±  |0.0089|
|         |       |strict-match    |     8|exact_match|↑  |0.7597|±  |0.0118|
```

H100x8 result:
```
|  Tasks  |Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|---------|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k_cot|      3|flexible-extract|     8|exact_match|↑  |0.884|±  |0.0088|
|         |       |strict-match    |     8|exact_match|↑  |0.765|±  |0.0117|
```

## Perf
Comparing kernel vs reference (JAX) implementation to ensure no major regression

Base CMD:
```
MODEL_IMPL_TYPE=vllm vllm serve RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8-dynamic  --max-model-len=5120 --disable-log-requests --max-num-batched-tokens 4096 --max-num-seqs=256 --tensor-parallel-size=2  --no-enable-prefix-caching 
```

### Reference
```
Successful requests:                     256       
Benchmark duration (s):                  190.38    
Total input tokens:                      261892    
Total generated tokens:                  1048576   
Request throughput (req/s):              1.34      
Output token throughput (tok/s):         5507.69   
Peak output token throughput (tok/s):    7168.00   
Peak concurrent requests:                256.00    
Total Token throughput (tok/s):          6883.28   
Mean TTFT (ms):                          3385.70   
Median TTFT (ms):                        3279.12   
P99 TTFT (ms):                           6826.10   
Mean TPOT (ms):                          45.45     
Median TPOT (ms):                        45.52     
P99 TPOT (ms):                           45.86     
Mean ITL (ms):                           45.45     
Median ITL (ms):                         38.46     
P99 ITL (ms):                            69.49     
```

### With ENABLE_QUANTIZED_MATMUL_KERNEL=1
```
Successful requests:                     256       
Benchmark duration (s):                  189.18    
Total input tokens:                      261892    
Total generated tokens:                  1048576   
Request throughput (req/s):              1.35      
Output token throughput (tok/s):         5542.82   
Peak output token throughput (tok/s):    7168.00   
Peak concurrent requests:                256.00    
Total Token throughput (tok/s):          6927.20   
Mean TTFT (ms):                          3065.95   
Median TTFT (ms):                        2961.63   
P99 TTFT (ms):                           6182.34   
Mean TPOT (ms):                          45.23     
Median TPOT (ms):                        45.29     
P99 TPOT (ms):                           45.56     
Mean ITL (ms):                           45.23     
Median ITL (ms):                         38.38     
P99 ITL (ms):                            69.20     
```

# Checklist

### DIFF
--- a/tests/kernels/quantized_matmul_kernel_test.py
+++ b/tests/kernels/quantized_matmul_kernel_test.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
-import functools
-
 import jax
 import jax.numpy as jnp
 from absl.testing import absltest, parameterized
@@ -10,42 +8,12 @@ from jax._src import test_util as jtu
 from tpu_inference.kernels.quantized_matmul import (kernel, tuned_block_sizes,
                                                     util)
 
+xla_quantized_matmul = kernel.xla_quantized_matmul
 quantized_matmul_kernel = kernel.quantized_matmul_kernel
 quantize_tensor = util.quantize_tensor
 get_tuned_block_sizes = tuned_block_sizes.get_tuned_block_sizes
 
 jax.config.parse_flags_with_absl()
-
-
-@functools.partial(jax.jit, static_argnames=["quantize_activation"])
-def reference_quantized_matmul(
-    x: jax.Array,
-    w_q: jax.Array,
-    w_scale: jax.Array,
-    quantize_activation=True,
-):
-    if quantize_activation:
-        acc_dtype = jnp.float32
-        if quantize_activation and jnp.issubdtype(w_q.dtype, jnp.integer):
-            acc_dtype = jnp.int32
-
-        x_q, x_scale = quantize_tensor(x, w_q.dtype)
-        out = jax.lax.dot_general(
-            x_q,
-            w_q,
-            dimension_numbers=(((1, ), (1, )), ((), ())),
-            preferred_element_type=acc_dtype,
-        ).astype(jnp.float32)
-        out *= x_scale
-    else:
-        out = jax.lax.dot_general(
-            x,
-            w_q,
-            dimension_numbers=(((1, ), (1, )), ((), ())),
-            preferred_element_type=jnp.float32,
-        )
-    out *= jnp.expand_dims(w_scale, 0)
-    return out.astype(x.dtype)
 
 
 @jtu.with_config(jax_numpy_dtype_promotion="standard")
@@ -94,7 +62,7 @@ class QuantizedMatmulKernelTest(jtu.JaxTestCase):
             x_q_dtype=x_q_dtype,
             tuned_value=tuned_value,
         )
-        expected = reference_quantized_matmul(
+        expected = xla_quantized_matmul(
             x, w_q, w_scale, quantize_activation=quantize_activation)
 
         self.assertAllClose(output,
--- a/tpu_inference/layers/vllm/linear_common.py
+++ b/tpu_inference/layers/vllm/linear_common.py
@@ -9,30 +9,52 @@ from jax.sharding import PartitionSpec as P
 from torchax.interop import torch_view
 from torchax.ops.mappings import t2j
 
-from tpu_inference.kernels.quantized_matmul.kernel import \
-    quantized_matmul_kernel
+from tpu_inference import envs
+from tpu_inference.kernels.quantized_matmul.kernel import (
+    quantized_matmul_kernel, xla_quantized_matmul)
 
 
 def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
-                             mesh: Mesh, weight_sharding: P):
-    out_axis, in_axis = weight_sharding
-    x_sharding = P(None, in_axis)
-    scale_sharding = P(out_axis, )
-    out_sharding = P(None, out_axis)
+                             mesh: Mesh, weight_sharding: P) -> jax.Array:
+    """
+    Wrapper around the quantized matmul kernel.
 
-    x = jax.lax.with_sharding_constraint(x, NamedSharding(mesh, x_sharding))
+    Args:
+        x:  Activation.
+        w_q: Weight quantized array. [n_output_features, n_input_features]
+        w_s: Weight quantization scale. [n_output_features]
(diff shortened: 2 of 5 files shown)


tpu_inference/layers/vllm/linear_common.py
```diff
@@ -9,12 +9,74 @@
...
+    return out.astype(x.dtype)
+
 
 def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
-                             mesh: Mesh, weight_sharding: P):
+                             mesh: Mesh, weight_sharding: P) -> jax.Array:
+    """
+    Wrapper around the quantized matmul kernel.
+
+    NOTE (jacobplatin/kyuyeunk) there have been numeric issues (concerning) NaNs
```
yaochengji: As there's numerical issues on the quantized-matmul kernel, should we disable it by default? cc @kyuyeunk

jrplatin: If that's the case, I can also update the w8a8_int8 compressed tensors implementation as well (which also calls the quantized matmul)

kyuyeunk: Let's disable it by default. I think the quantized matmul kernel has served it's purpose - especially for per-channel Ironwood. But let's revisit this when we move onto subchannel quantization @jrplatin

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
```diff
@@ -155,10 +155,13 @@ def _apply_fused(self, layer: torch.nn.Module, x: torch.Tensor,
             outs *= weight_scale_jax
             outs = outs.astype(x_jax.dtype)
         else:
-            outs = sharded_quantized_matmul(x_jax, weight_jax,
-                                            weight_scale_jax,
-                                            self.jax_config.mesh,
-                                            self.jax_config.weight_sharding)
+            if envs.DISALBE_QUANTIZED_MATMUL_KERNEL:
```
kyuyeunk: I think moving this logic into `sharded_quantized_matmul` will make the logic cleaner. wdyt?

jrplatin: Yep makes sense, will change!

kyuyeunk:
Since it will be disabled by default, can you do a full performance evaluation and verify if performance doesn't change much?

For instance, quantized models like these?
- RedHatAl/Meta-Llama-3.1-8B-Instruct-quantized.w8a8
- RedHatAl/Meta-Llama-3.1-8B-Instruct-FP8-dynamic

jrplatin:
@kyuyeunk I updated the FP8 numbers on Ironwood -- I can get int8 if you think they're necessary but was running into VMEM OOM + untuned block sizes on Trillium

tpu_inference/layers/vllm/linear_common.py
```diff
@@ -9,30 +9,98 @@
...
 
+from tpu_inference import envs
+from tpu_inference.kernels.quantized_matmul import util
 from tpu_inference.kernels.quantized_matmul.kernel import \
     quantized_matmul_kernel
 
+quantize_tensor = util.quantize_tensor
+
+
+def reference_quantized_matmul(
```
kyuyeunk: Nit: Rename to something like 'xla_quantized_matmul'

kyuyeunk:
Can you do a rebase instead of merge to update this branch? I think our pre-commit ci doesn't correctly if we do merge for updating branch.

tests/kernels/quantized_matmul_kernel_test.py
```diff
@@ -17,37 +16,6 @@
 jax.config.parse_flags_with_absl()
 
 
-@functools.partial(jax.jit, static_argnames=["quantize_activation"])
-def reference_quantized_matmul(
```
vanbasten23, line 21: imo this ref impl should be in the kernel file, like other kernels. Then in your vllm code, you can import the ref impl from the kernel module.

vanbasten23:
lgtm with one comment

PR #1274, opened 2025-12-10
Fix moe layer from upstream change

kyuyeunk:
# Description

Fix due to change in https://github.com/vllm-project/vllm/pull/29066

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Optional, Union
+from typing import Any, Optional, Union
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Optional, Union
+from typing import Any, Optional, Union
 
 import jax
 import jax.numpy as jnp
```
vanbasten23, line 4: no need to change the file compressed_tensors_moe.py?

kyuyeunk: ohhh good point. I'm surprised that the ci still passed despite this? maybe there's a missing unit test? Anyway, i've added the change.

vanbasten23: wait, do we have a unit test for that file? I'm surprise the ci still passed too.

kyuyeunk: Okay, so in my other tests, i was calling vllm layer - which eventually invokes our own fused moe implmenetation: https://github.com/vllm-project/tpu-inference/blob/5f7dc4eda1706591983c617daa559ee11c2d7267/tests/layers/vllm/test_unquantized.py#L475

But in `tests/layers/vllm/test_compressed_tensors_moe.py`, it was only calling our own layer implementation and that's why it passed: https://github.com/vllm-project/tpu-inference/blob/5f7dc4eda1706591983c617daa559ee11c2d7267/tests/layers/vllm/test_compressed_tensors_moe.py#L173-L177

kyuyeunk: oh wait, the bigger problem is that this unit test was ignored completely: https://github.com/vllm-project/tpu-inference/blob/5f7dc4eda1706591983c617daa559ee11c2d7267/.buildkite/pipeline_jax.yml#L124

I have no idea why??

kyuyeunk: I suggest merging this pr for now - because the upstream issue is completing blocking our CI - and figure out a fix for this later. does that sounds good?

vanbasten23: because of https://github.com/vllm-project/tpu-inference/pull/1065#issue-3610239500

kyuyeunk: okay, let's merge this PR for now because this upstream issue is completely blocking our CI. And let's figure out how to re-enable that unit test as a next step.

PR #1279, opened 2025-12-10
Enable All Tests on TPUv7

QiliangCui:
# Description
Enable test on TPUv7.

1. duplicate the v6e jax yaml.
2. change the queue name, test key
3. fix a code in `tpu_inference/kernels/collectives/all_gather_matmul_tuned_block_sizes.py` to allow "tpuv7x" as tpu kind.

# Tests:
https://buildkite.com/tpu-commons/tpu-inference-ci/builds/6764

### DIFF
--- a/.buildkite/pipeline_jax_tpu7x.yml
+++ b/.buildkite/pipeline_jax_tpu7x.yml
@@ -8,21 +8,24 @@ steps:
   # -----------------------------------------------------------------
   # TEST STEPS - Calling wrapper
   # -----------------------------------------------------------------
-   - label: "E2E MLPerf tests for JAX models"
-     key: test_0
+   - label: "TPU7x E2E MLPerf tests for JAX models"
+     key: tpu7x_test_0
      soft_fail: true
+     env:
+      IS_FOR_V7X: "true"
      agents:
-       queue: tpu_v6e_queue
+       queue: tpu_v7x_2_queue
      commands:
        - .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_inference/tests/e2e/benchmarking/mlperf.sh
 
-   - label: "E2E MLPerf tests for JAX models with quantization"
-     key: test_1
+   - label: "TPU7x E2E MLPerf tests for JAX models with quantization"
+     key: tpu7x_test_1
      soft_fail: true
      env:
        QUANTIZATION: "True"
+       IS_FOR_V7X: "true"
      agents:
-       queue: tpu_v6e_queue
+       queue: tpu_v7x_2_queue
      commands:
        - |
          if [[ "$$NIGHTLY" == "1" ]]; then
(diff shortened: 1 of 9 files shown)


tests/lora/test_layers.py
```diff
@@ -205,6 +205,8 @@ def create_random_inputs(
 @pytest.mark.parametrize("repeats", [1, 2, 3])
 @pytest.mark.parametrize("stage", [True, False])
 def test_column_parallel_packed(dist_init, num_loras, repeats, stage) -> None:
+    if 'TPU7x' in jax.devices()[0].device_kind:
```
jrplatin, line 209: Why is this being skipped?

jrplatin:
Just curious -- do we want to run all tests on all platforms? E.g. all tests that run on v6e should also be run on v7x?

.buildkite/pipeline_jax_tpu7x.yml
jrplatin, line 1: Is this just the `.buildkite/pipeline_jax.yml` copied and only updating the `queue` / test name for v7?

QiliangCui: yes.

changes:

1. change the queue.
2. add a `IS_FOR_V7X: "true"` to make it build v7 image
3. change the test key to `tpu7x_test_[N]`

tests/lora/test_layers.py
QiliangCui: my bad. yes, because this test doesn't pass on v7. I want to enable all first then find owner of not passed. b/467379668 is tracking it

QiliangCui: this is the failure: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/6741#019b0784-99a5-48c5-945a-7e6cd158f503

QiliangCui: it is some accuracy error that i don't think i can fix in short time 

```

with torchax.default_env():
--
actual_result_cpu = actual_result.to('cpu')
>           torch.testing.assert_close(actual_result_cpu,
expected_result,
rtol=rtol,
atol=atol)
E           AssertionError: Tensor-likes are not close!
E
E           Mismatched elements: 1024 / 4096 (25.0%)
E           Greatest absolute difference: 13.25 at index (25, 77) (up to 0.02 allowed)
E           Greatest relative difference: 0.48046875 at index (9, 77) (up to 0.03 allowed)


```

.buildkite/pipeline_jax_tpu7x.yml
jrplatin: Discussed offline and Qiliang will track, but would be nice to not have this entirely replicated (but doesn't seem like BuildKite has a good solution for this right now)

.buildkite/pipeline_jax_tpu7x.yml
```diff
@@ -0,0 +1,311 @@
...
+       queue: tpu_v7x_2_queue
+     commands:
+       - |
+         IS_FOR_V7X=true .buildkite/scripts/run_in_docker.sh \
+           python3 -m pytest -s -v -x /workspace/tpu_inference/tests/ \
+           --ignore=/workspace/tpu_inference/tests/kernels \
+           --ignore=/workspace/tpu_inference/tests/lora \
+           --ignore=/workspace/tpu_inference/tests/e2e \
+           --ignore=/workspace/tpu_inference/tpu_inference/mock \
+           --ignore=/workspace/tpu_inference/tests/layers/vllm/test_compressed_tensors_moe.py \
```
vanbasten23, line 135: @QiliangCui Could you enable this /workspace/tpu_inference/tests/layers/vllm/test_compressed_tensors_moe.py  test? It was disabled previously because it can only be run on v7.

vanbasten23:
Thanks for enabling it!

PR #1287, opened 2025-12-11
[JAX][MoE] Integrate multiple MoE kernels in MoE modules

bzgoogle:
# Description

It is integrating multiple MoE kernel in MoE modules. 
*NOTE: some of the kernels are not supporting quantization, we will add it in the follow-up PR when they're ready
**NOTE: this PR supports random weight. For checkpoint loading, the weight loading logic of particular models may need to be updated to address different weight shapes. Those will be updated in another follow-up PR for related models

The supported kernel will now include:

- Fused MoE kernel
- torchax MoE kernel
- JAX sparse-matmul with ragged-dot
- JAX sparse-matmul with megablox
- Dense-matmul 

# Tests
- torchax MoE kernel: [xprof](https://screenshot.googleplex.com/5cHFZEiiM5TZhyy)
- JAX sparse-matmul with ragged-dot [xprof](https://screenshot.googleplex.com/AuCBUDfGhPrwYwx)
- JAX sparse-matmul with megablox [xprof](https://screenshot.googleplex.com/3VfQZdxAfSgCUmt)

Accuracy on DeepSeek:
- [offline results](https://paste.googleplex.com/6658328069537792) from this PR
- [offline results](https://paste.googleplex.com/5978455239950336) from head of main

Validated MMLU accuracy on multi-host:

```
Namespace(backend='vllm', base_url=None, host='127.0.0.1', port=8000, endpoint='/v1/completions', dataset_name='mmlu', dataset_path='/workspace/mmlu/data/test/', max_concurrency=None, model='/mnt/disks/tpu_vm/', tokenizer=None, use_beam_search=False, num_prompts=1000, logprobs=None, request_rate=inf, burstiness=1.0, seed=0, trust_remote_code=False, disable_tqdm=False, profile=False, ignore_eos=False, percentile_metrics='ttft,tpot,itl', metric_percentiles='99', goodput=None, request_id_prefix='benchmark-serving', mmlu_input_len=None, mmlu_output_len=None, mmlu_num_shots=1, mmlu_method='HELM', mmlu_use_chat_template=False, mlperf_input_len=None, mlperf_output_len=None, gpqa_output_len=2048, gpqa_use_chat_template=False, sonnet_input_len=550, sonnet_output_len=150, sonnet_prefix_len=200, random_input_len=1024, random_output_len=128, random_range_ratio=0.0, random_prefix_len=0, top_p=None, top_k=None, min_p=None, temperature=None, tokenizer_mode='auto', served_model_name=None, run_eval=True, warmup_mode='sampled')
Loaded 14042 data from mmlu dataset
Using sampling parameters: {'temperature': 0.0}
Warmup (mode: sampled) is starting.
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:01<00:00,  3.26it/s]
Warmup (mode: sampled) has completed.
Traffic request rate: inf
Burstiness factor: 1.0 (Poisson process)
Maximum request concurrency: None
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1000/1000 [02:10<00:00,  7.67it/s]
Benchmark duration (s):                  130.33    
Total input tokens:                      173929    
Request throughput (req/s):              7.67      
Output token throughput (tok/s):         15.35     
Total token throughput (tok/s):          1349.92   
Mean TTFT (ms):                          63322.74  
Median TTFT (ms):                        58248.17  
P99 TTFT (ms):                           128608.55 
Mean TPOT (ms):                          95.06     
Median TPOT (ms):                        95.15     
P99 TPOT (ms):                           98.14     
Mean ITL (ms):                           95.06     
Median ITL (ms):                         95.15     
P99 ITL (ms):                            98.14     
Mean E2EL (ms):                          63417.81  
Median E2EL (ms):                        58344.82  
P99 E2EL (ms):                           128701.75 
Evaluating MMLU...
Downloading builder script: 4.20kB [00:00, 21.0MB/s]
[nltk_data] Downloading package punkt to /root/nltk_data...
[nltk_data]   Unzipping tokenizers/punkt.zip.
[nltk_data] Downloading package punkt_tab to /root/nltk_data...
[nltk_data]   Unzipping tokenizers/punkt_tab.zip.

Results

{'accuracy': 0.865, 'gen_num': 1000}
```

Also validated MMLU for existing TPU-Friendly FP8/FP4 + native FP8 (e.g. `deepseek-ai/DeepSeek-R1`

# Checklist


tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -343,6 +315,31 @@ def _print_model_architecture(self):
         logger.debug("\n### LM Head ###")
         nnx.display(self.lm_head)
 
+    def _process_moe_kernel_flag(self):
+        self.use_fused_moe_kernel = bool(int(os.getenv("USE_MOE_EP_KERNEL", "0")))
+        self.use_vllm_moe_kernel = bool(int(os.getenv("USE_VLLM_MOE_KERNEL", "0")))
+        self.sparse_matmul = bool(int(os.getenv("USE_SPARSE_MATMUL", "0")))
```
kyuyeunk: please use `tpu_inference.envs` instead of manually calling `os.getenv`: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/envs.py

example: https://github.com/vllm-project/tpu-inference/blob/ac26352871e9d0ab759a2af5097bc419802a98dd/tpu_inference/layers/vllm/quantization/unquantized.py#L175

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -343,6 +315,31 @@ def _print_model_architecture(self):
...
+        self.use_megablox = bool(int(os.getenv("USE_MEGABLOCKS", "0")))
+        self.use_ragged_dot = bool(int(os.getenv("USE_RAGGED_DOT", "1")))
+
+        assert sum([self.use_fused_moe_kernel, self.use_vllm_moe_kernel]) <= 1, "You can enable at most one MoE kernels." 
+
+        if self.use_fused_moe_kernel:
```
kyuyeunk: match/case is more suitable for this kind of code.

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -46,14 +62,17 @@ def __call__(self, x_TD: Float):
...
-            router_logits_TE, self.num_experts_per_tok)
-        if self.router_act != "sigmoid":  # sigmoid does not accept axis argument.
-            normalized_weights_TX = router_act(weights_TX.astype(self.dtype),
-                                               axis=-1)
+        if self.use_moe_kernel:
+            return router_logits_TE
         else:
-            normalized_weights_TX = router_act(weights_TX.astype(self.dtype))
-        return normalized_weights_TX, selected_experts_TX
+            weights_TX, selected_experts_TX = jax.lax.top_k(
```
kyuyeunk, line 105: even when `self.use_moe_kernel == False`, recommend changing the code so that `Router` will always only return `router_logits_TE` and let the subsequent layers figure out how to deal with other parts - such as topk.

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -1,16 +1,31 @@
...
 from tpu_inference.layers.jax.base import create_param
 from tpu_inference.layers.jax.layers import FlaxUtils
+from tpu_inference.layers.jax.misc import round_up_to_multiple_of_128_within_limit
+from tpu_inference.layers.vllm.fused_moe import fused_moe_func
+from tpu_inference.models.jax.utils.quantization.quantization_utils import (
+    manually_quantize_qwix_activation, manually_quantize_qwix_weight)
 
 modeling_flax_utils = FlaxUtils()
-
+set_xla_metadata = xla_metadata.set_xla_metadata
```
kyuyeunk: why is this needed?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -92,20 +91,10 @@ def __init__(self,
 
         self.random_init = force_random_weights or self.vllm_config.additional_config.get(
```
kyuyeunk: please avoid using `additional_config`. for this use case, please use `vllm_config.load_config.load_format == "dummy"` instead.

example: https://github.com/vllm-project/tpu-inference/blob/ac26352871e9d0ab759a2af5097bc419802a98dd/tpu_inference/models/vllm/vllm_model_wrapper.py#L109-L110

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -99,46 +136,161 @@ def __call__(self, x_TD: Float):
...
+                self.num_expert_parallelism =self.mesh.shape[self.expert_axis_name]
+            else:
+                self.num_expert_parallelism = math.prod(self.mesh.shape[axis] for axis in self.expert_axis_name)
 
+        # Derive if data is sharded by expert
+        self.data_axis_name = self.activation_ffw_td[0]
+        self.is_batch_sharded_by_expert = (
+            self.expert_axis_name is not None) and (self.expert_axis_name == self.data_axis_name)
+
+    ## ======= Methods used for Dense Matmul ======= ##
```
kyuyeunk: please separate out a methods that is only used for `Dense Matmul` into a separate class / file.

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -207,3 +359,434 @@ def _moe_fwd(self, x_TD: Float, weights):
         with jax.named_scope("sum"):
             output_TD = jnp.einsum('TED,TE -> TD', down_proj_TED, weights)
         return output_TD.astype(self.dtype)
+
+    ## ======= END Methods used for Dense Matmul ======= ##
+
+    ## ======= Methods used for Sparse Matmul ======= ##
```
kyuyeunk: similar comment as above. separate out methods that is only used for sparse matmul into a separate class/file.

tpu_inference/layers/jax/moe/moe.py
bzgoogle: it's suggested by Piotr to optimize the ragged-dot with fp8 support, which is used in line 570

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: Done

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: Done

tpu_inference/layers/jax/moe/moe.py
bzgoogle: Added a TODO. Given the PR is for MoE, i will refactor the router in a follow-up PR

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: Done. remove the random init from additional_config, but relying on the load_format=dummy from vllm config

tpu_inference/layers/jax/moe/moe.py
bzgoogle: Done

tpu_inference/layers/jax/moe/moe.py
bzgoogle: Done

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -895,6 +896,32 @@ def _print_model_architecture(self):
...
+    def _process_moe_kernel_flag(self):
+        self.use_fused_moe_kernel = envs.USE_MOE_EP_KERNEL
+        self.use_vllm_moe_kernel = envs.USE_VLLM_MOE_KERNEL
+        self.sparse_matmul = envs.USE_SPARSE_MATMUL
+        self.use_megablox = envs.USE_MEGABLOCKS
+        self.use_ragged_dot = envs.USE_RAGGED_DOT
+
+        assert sum([self.use_fused_moe_kernel, self.use_vllm_moe_kernel]) <= 1, "You can enable at most one MoE kernels." 
+
+        match (self.use_fused_moe_kernel, self.use_vllm_moe_kernel, self.sparse_matmul):
```
kyuyeunk: Please use enum instead. example
- https://github.com/vllm-project/tpu-inference/blob/475a3352e4d24002771cabe136663a5cb970e1a5/tpu_inference/layers/vllm/fused_moe.py#L38-L56
- https://github.com/vllm-project/tpu-inference/blob/475a3352e4d24002771cabe136663a5cb970e1a5/tpu_inference/layers/vllm/process_weights/fused_moe_weights.py#L296-L340

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -101,7 +122,405 @@ def __post_init__(self, rngs: nnx.Rngs):
                                       sharding=self.ed_sharding,
                                       random_init=self.random_init)
 
+# --- Helper Class for Sparse MoE ---
+
+class SparseMoEEngine(nnx.Module):
```
kyuyeunk: Please move this function out to a different file. also, please use functional programming style. meaning, separate out the purely stateless computational logic to the outside of the class

example: primary job of `VllmMxfp4MoeMethod` is to act as a interface between vllm / pytorch world & jax world: https://github.com/vllm-project/tpu-inference/blob/343bbc71a626a8416a48d5b6773af5e4551b67cd/tpu_inference/layers/vllm/quantization/mxfp4.py#L95-L235

Meaning
1. it handles converting torch parameters into jax compatible torchax tensor during model initialization: https://github.com/vllm-project/tpu-inference/blob/343bbc71a626a8416a48d5b6773af5e4551b67cd/tpu_inference/layers/vllm/quantization/mxfp4.py#L143-L149
2. and it handles unpacking the weights during forward function: https://github.com/vllm-project/tpu-inference/blob/343bbc71a626a8416a48d5b6773af5e4551b67cd/tpu_inference/layers/vllm/quantization/mxfp4.py#L218-L225

but when it comes to actual computation, the functionalities are offloaded to pure jax functions
1. when pre-processing weights into tpu-friendly format & shard them to tpus, `VllmMxfp4MoeMethod` calls these functions that are pure jax: https://github.com/vllm-project/tpu-inference/blob/343bbc71a626a8416a48d5b6773af5e4551b67cd/tpu_inference/layers/vllm/process_weights/fused_moe_weights.py#L111-L369
2. during forward `VllmMxfp4MoeMethod` invokes these functions which are also pure jax: https://github.com/vllm-project/tpu-inference/blob/343bbc71a626a8416a48d5b6773af5e4551b67cd/tpu_inference/layers/vllm/fused_moe.py#L75-L113

tpu_inference/layers/jax/moe/moe.py
kyuyeunk: Why does it optimizes ragged dot?

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: Thanks Kyuyeun. Updated with MoEBackend Enum in util.py

tpu_inference/layers/jax/moe/moe.py
bzgoogle: Thanks. Moved those into vllm/tpu-inference/tpu_inference/layers/jax/moe/utils.py and refatored them.

tpu_inference/layers/jax/moe/moe.py
bzgoogle: It is used to set the fusion group and tiling so that the ragged-dot could fuse dequantization and matmul. i.e. https://buganizer.corp.google.com/issues/444760716#comment21

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,484 @@
...
+                preferred_element_type=dtype)
+
+    if pad_amount > 0:
+        output = output[:num_rows, :]
+    return output
+
+
+# --- Helper Class for Sparse MoE ---
+
+class SparseMoEEngine(nnx.Module):
```
kyuyeunk: this class is `nnx.Module` and it doesn't belong in `utils.py` file. please create another file (maybe `sparse_moe.py`?) that contains `SparseMoE` implementation.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,484 @@
...
+                                      global_sort_indices, router_weights_TX,
+                                      self.m.num_experts_per_tok, self.m.hidden_size,
+                                      self.m.dtype)
+
+        return output_TD
+
+
+# --- Helper Class for Dense MoE ---
+
+class DenseMoEEngine(nnx.Module):
```
kyuyeunk: Same as above comment. tanother file (maybe `dense_moe.py`?) that contains `DenseMoE` implementation.

tpu_inference/layers/jax/moe/utils.py
bzgoogle: Done

tpu_inference/layers/jax/moe/utils.py
bzgoogle: Done

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -895,6 +894,32 @@ def _print_model_architecture(self):
...
         nnx.display(self.lm_head)
 
+    def _process_moe_kernel_flag(self):
+        self.use_fused_moe_kernel = envs.USE_MOE_EP_KERNEL
+        self.use_vllm_moe_kernel = envs.USE_VLLM_MOE_KERNEL
+        self.sparse_matmul = envs.USE_SPARSE_MATMUL
+        self.use_megablox = envs.USE_MEGABLOCKS
+        self.use_ragged_dot = envs.USE_RAGGED_DOT
+
+        assert sum([self.use_fused_moe_kernel, self.use_vllm_moe_kernel]) <= 1, "You can enable at most one MoE kernels."
```
gpolovets1: I think you can replace this section wtih utils.select_moe_backend()?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -895,6 +894,32 @@ def _print_model_architecture(self):
         logger.debug("\n### LM Head ###")
         nnx.display(self.lm_head)
 
+    def _process_moe_kernel_flag(self):
+        self.use_fused_moe_kernel = envs.USE_MOE_EP_KERNEL
```
gpolovets1: nit: Should we just store self.moe_backend and perform isinstance(self.mode_backend, ..) when needed?

tpu_inference/models/jax/deepseek_v3.py
gpolovets1, line 1: main branch now only supports FP4 so did you confirm that the different moe backends (or at least Megablox kernel) work as expected with the FP4 weights?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -496,13 +501,53 @@ def load_weights(self, model_for_loading: nnx.Module):
...
+                                w_bytes_up, w_shards_up = self._load_individual_weight(
+                                    loaded_name, up_w, model_params, model_for_loading.mesh, scale=up_s)
+                                w_bytes += w_bytes_up
+                                w_shards += w_shards_up
+                                
+                                w_bytes_down, w_shards_down = self._load_individual_weight(
+                                    loaded_name, down_w, model_params, model_for_loading.mesh, scale=down_s)
+                                w_bytes += w_bytes_down
+                                w_shards += w_shards_down
                         weight_bytes, weight_shards = self._load_individual_weight(
```
gpolovets1: It seems _load_individual_weight is being called twice on stacked_weights. Could you explain why this is needed?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -496,13 +501,53 @@ def load_weights(self, model_for_loading: nnx.Module):
...
+                                
+                                fused_name = f"model.layers.{layer_num}.mlp.experts.0.gate_upproj_fused.weight"
+                                w_bytes, w_shards = self._load_individual_weight(
+                                    fused_name, fused_w, model_params, model_for_loading.mesh, scale=fused_s)
+                                
+                                w_bytes_down, w_shards_down = self._load_individual_weight(
+                                    loaded_name, down_w, model_params, model_for_loading.mesh, scale=down_s)
+                                
+                                # Update cumulative memory
+                                w_bytes += w_bytes_down
```
gpolovets1: These don't seem to be used anywhere?

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: Could you also confirm that the accuracy on MMLU looks reasonable with these changes?

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: It already been called. This part is not needed. Removed it.

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: same as above. this is not needed. removed

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: nice spot. yes, i should not be called twice. fixed

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: updated the variable names

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: this PR does not have quantization support for the MoE kernels. It will be addressed in another follow-up PR and we can test the MMLU accuracy there. https://github.com/vllm-project/tpu-inference/pull/1469

tpu_inference/layers/jax/moe/sparse_moe.py
```diff
@@ -0,0 +1,178 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jax
+import jax.numpy as jnp
+
+from flax import nnx
+
+from tpu_inference.layers.jax.moe.utils import *
```
kyuyeunk: using wildcard import isn't a good coding convention. Please do a explicit import instead.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,285 @@
...
+
+def select_moe_backend():
+
+    assert sum([envs.USE_MOE_EP_KERNEL, envs.USE_VLLM_MOE_KERNEL]) <= 1, "You can enable at most one MoE kernels." 
+
+    match (envs.USE_MOE_EP_KERNEL, envs.USE_VLLM_MOE_KERNEL, envs.USE_MEGABLOCKS, envs.USE_RAGGED_DOT):
+        case (True, _, _, _):
+            logger.info("[MoE]: Fused MoE kernel is enabled")
+            return MoEBackend.FUSED_MOE
+        case (_, True, _, _):
```
kyuyeunk: this is really confusing to read. it's hard to know what first element is, second element should be, etc. please find a better way.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,285 @@
...
+from tpu_inference.models.jax.utils.qwix.qwix_utils import (
+    manually_quantize_qwix_activation)
+
+logger = init_logger(__name__)
+modeling_flax_utils = FlaxUtils()
+set_xla_metadata = xla_metadata.set_xla_metadata
+
+class MoEBackend(enum.Enum):
+    FUSED_MOE = "fused_moe"
+    VLLM_MOE = "vllm_moe"
```
kyuyeunk, line 38: `VLLM_MOE` you are referring to here is just `MEGABLX_GMM`. Please only keep one option. Also, `VLLM_MOE` is not really a good name. please think of a different name.

tpu_inference/layers/jax/misc.py
```diff
@@ -28,3 +28,36 @@ def shard_put(x: jax.Array, sharding_names: Tuple[str, ...] | P,
     if math.prod(mesh.axis_sizes) == 1:
         return jax.device_put(x, mesh.devices.flatten()[0])
     return jax.device_put(x, NamedSharding(mesh, P(*sharding_names)))
+
+def round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
```
kyuyeunk: please avoid creating a duplicate function: https://github.com/vllm-project/tpu-inference/blob/c93db21ab07dfc0ab0c83c6b405aefe5665fa31d/tpu_inference/layers/common/fused_moe_gmm.py#L52

tpu_inference/layers/jax/moe/utils.py
bzgoogle: I think a better readability is to use if-elif-else. If you think it's better, i can switch back to if-elif-else clause.

tpu_inference/layers/jax/moe/utils.py
bzgoogle: No, VLLM_MOE is calling the tpu_inference.layers.vllm.fused_moe directly.
USE_MEGABLOCKS is for just using megablx_gmm . Please see more details in: https://screenshot.googleplex.com/98WZNYrdwgtR7Tg

kyuyeunk: `VLLM_MOE` uses a fork of `megablx_gmm`. it's practically a same thing.

tpu_inference/models/jax/deepseek_v3.py
jrplatin: Resolved offline and due to urgency, we will just support native FP8 for now

tpu_inference/layers/jax/moe/utils.py
jrplatin: Updated to use if-else

tpu_inference/layers/jax/moe/sparse_moe.py
jrplatin: Update to use explicit import

tpu_inference/layers/jax/misc.py
jrplatin: Good catch, thanks!

tests/layers/jax/moe/test_deepseek_moe.py
```diff
@@ -92,73 +90,6 @@ def test_router_e2e(self):
             self.assertEqual(weights.shape, (2, 2))
             self.assertEqual(indices.shape, (2, 2))
 
-
-class TestSparseMoE(unittest.TestCase):
```
gpolovets1, line 96: Could you create a new test file and it there to cover sparse_moe.py?

nohup.out
gpolovets1, line 1: Remove this?

jrplatin: Ah good catch...

tests/layers/jax/moe/test_deepseek_moe.py
jrplatin: Yep, I was going to mark as TODO just because we wanted to move fast here, but let me add

tpu_inference/layers/jax/moe/dense_moe.py
```diff
@@ -20,48 +20,45 @@
 from tpu_inference.layers.jax.moe.utils import modeling_flax_utils
 
 
-class DenseMoEEngine(nnx.Module):
-    """Encapsulates logic for Dense/Standard MoE forward passes."""
+def dense_moe_fwd(moe_instance, x_TD: Float, weights):
```
gpolovets1, line 23: Just a few follow up questions:
1. Qwix cannot handle class methods?
2. Did you just move the code from the class to a global method? Wondering what should be scanned over.

jrplatin: 1. I was running into an issue because where we were initting the params and where we called them were in different classes, so moving them to global functions solved the issue
2. Yep!

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -49,7 +49,7 @@ def _swigluoai(x1: jax.Array,
     return gated_activation * (x2 + 1)
 
 
-def _round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
+def round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
```
patemotter, line 53: Can we add a todo to move this generic method to a util file or similar? I imagine there's a few instances like this in the codebase that could be done as part of a future cleanup.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,277 @@
...
+                     local_expert_size,
+                     shard_index,
+                     is_offset,
+                     global_sorted_experts=None):
+    """Stateless local permutation logic."""
+    # global_group_sizes: (tokens parallelism, num_total_experts)
+    # all_shard_local_sizes: (tokens parallelism, num local experts in the shard)
+    all_shard_local_sizes = jax.lax.dynamic_slice_in_dim(global_group_sizes,
+                                                         shard_index *
+                                                         local_expert_size,
```
patemotter: Nit: I don't know if the formatter requires this to be split across two lines, but it would be a lot cooler if it was on one.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -0,0 +1,277 @@
...
+def gmm_fn(inputs, kernel, group_sizes, tile_size, moe_backend, dtype,
+           quantized_dtype):
+    """Stateless Grouped Matrix Multiply."""
+    num_rows = inputs.shape[0]
+    pad_amount = (tile_size[0] - num_rows % tile_size[0]) % tile_size[0]
+    if pad_amount > 0:
+        inputs = jnp.pad(inputs, ((0, pad_amount), (0, 0)))
+
+    if moe_backend == MoEBackend.MEGABLX_GMM:
+        m, _, k, n = inputs.shape[0], *kernel.shape
```
patemotter: Nit: This seems needlessly confusing to combine as one line since it's not obviously clear what variables are being assigned from where. I know you can intuit it to some degree, but there's no guarantee.

tests/layers/jax/moe/test_deepseek_moe.py
jrplatin: Added!

tpu_inference/layers/common/fused_moe_gmm.py
jrplatin: Done!

tpu_inference/layers/jax/moe/utils.py
jrplatin: Fixed

tpu_inference/layers/jax/moe/utils.py
jrplatin: Made a new variable on the line above for readability

PR #1295, opened 2025-12-12
Update multihost disagg sh to prepare integrate with buildkite

mrjunwan-lang:
# Description

Update multihost disagg sh to prepare integrate with buildkite
1. Make proxy server and benchmark running inside docker
2. Pass docker image in to leverage existing image code
3. build local mount switch to easy debug in local dev env

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/features/LoRA_Torch.yml
+++ b/.buildkite/features/LoRA_Torch.yml
@@ -31,7 +31,8 @@ steps:
       queue: tpu_v6e_queue
     commands:
       - |
-        buildkite-agent meta-data set "LoRA_Torch_PerformanceTest" "to be added"
+        .buildkite/scripts/run_in_docker.sh \
+          bash -c 'MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax python3 -m pytest -s -v -x /workspace/tpu_inference/tests/lora/test_lora_perf.py'
   - label: "Record performance test result for LoRA_Torch"
     key: "record_LoRA_Torch_PerformanceTest"
     depends_on: "LoRA_Torch_PerformanceTest"
--- a/.buildkite/features/Single-Host-P-D-disaggregation.yml
+++ b/.buildkite/features/Single-Host-P-D-disaggregation.yml
@@ -1,0 +1,25 @@
+# Single-Host-P-D-disaggregation
+# features support matrix
+steps:
+  - label: "Correctness tests for Single-Host P-D disaggregation"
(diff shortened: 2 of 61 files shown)


examples/disagg/run_disagg_multi_host.sh
```diff
@@ -4,27 +4,66 @@
...
 
-echo "--- DEBUG: The HOME variable is set to: $HOME ---"
-
-CONTAINERS=$(docker ps -a --filter "name=node*" -q)
+# Parameters may come from external
+# docker related
+CONTAINER_PREFIX=${CONTAINER_PREFIX:="disagg-node"}
+RUN_IN_BUILDKITE=${RUN_IN_BUILDKITE:=false}
+MODEL=${MODEL:="Qwen/Qwen3-0.6B"}
+DOCKER_IMAGE=${DOCKER_IMAGE:="us-central1-docker.pkg.dev/cloud-tpu-inference-test/jun-vllm-tpu/tpu-inference:jun-12-05-v7x"}
```
richardsliu: Why do we need this custom image?

examples/disagg/run_disagg_multi_host.sh
```diff
@@ -56,7 +95,7 @@ for ((i=0; i<NUM_HOSTS_PER_INSTANCE; i++)); do
         --privileged \
         --network host \
         --shm-size 16G \
-        --name "node-${i}" \
+        --name "${CONTAINER_PREFIX}-${i}" \
```
QiliangCui, line 98: here, does the `docker run` start docker locally or remotely on other machines? 

I mean... I imagine you need 4 docker image started and they are on different VM's, right?

examples/disagg/run_disagg_multi_host.sh
mrjunwan-lang: it is the well-known working image I already published, so if you choose this, you do not need build locally.

mrjunwan-lang: After thinking it twice, Updated to a common one to avoid confusion.

examples/disagg/run_disagg_multi_host.sh
mrjunwan-lang: No, 4 dockers are in the same v6e8 VM, each docker use 1 tpu chip. so 4 for prefill and 4 for decode.

examples/disagg/run_disagg_multi_host.sh
richardsliu: Thanks!

PR #1305, opened 2025-12-12
Update libtpu version for tpuv7.

QiliangCui:
# Description
```
jax==0.8.1
jaxlib==0.8.1
jaxtyping==0.3.2
libtpu==0.0.31
```

# Tests
unit test

# Checklist


requirements_v7x.txt
jrplatin, line 1: I think you can just delete this file and update `requirements.txt`?

jrplatin:
Ah so the skipped tests were failing because JAX/libtpu was old for v7x?

requirements_v7x.txt
QiliangCui: Let's do it in another bug/pr and have this focus on v7. For v7, this version issue is more critical because it broke some tests(see the removal of pytest.skip).

QiliangCui:
> Ah so the skipped tests were failing because JAX/libtpu was old for v7x?

yes. 

even with new jax, there are still some left over that needs to debug one by one..

tests/e2e/benchmarking/mlperf.sh
```diff
@@ -134,6 +134,18 @@ if [ "$use_dummy_weights" = true ]; then
     extra_serve_args+=("--load-format=dummy")
 fi
 
+if [ "$USE_V6E8_QUEUE" == "True" ]; then
+    # Set to 8 if job is in 8 chips queue.
+    # TODO (Qiliang Cui) Rename USE_V6E8_QUEUE to USE_8_CHIPS_QUEUE
+    extra_serve_args+=(--tensor-parallel-size 8)
+elif [ "$IS_FOR_V7X" == "true" ]; then
+    # Set the default value to 2 for tpu v7x
```
kyuyeunk, line 142: Can you mark a `TODO` for figuring out why it fails for tp=1 in v7x?

.buildkite/scripts/run_in_docker.sh
```diff
@@ -63,6 +63,7 @@ exec docker run \
   ${QUANTIZATION:+-e QUANTIZATION="$QUANTIZATION"} \
   ${NEW_MODEL_DESIGN:+-e NEW_MODEL_DESIGN="$NEW_MODEL_DESIGN"} \
   ${USE_V6E8_QUEUE:+-e USE_V6E8_QUEUE="$USE_V6E8_QUEUE"} \
+  ${IS_FOR_V7X:+-e IS_FOR_V7X="$IS_FOR_V7X"} \
```
kyuyeunk, line 66: This is purely to use `tp=2` when using v7x right? Similar to my other comment, can you leave a `TODO` here as well?

QiliangCui: yes. this is for the tp=2. i will create a bug to follow up. let me add the todo later avoiding triggering test again..

tests/e2e/benchmarking/mlperf.sh
QiliangCui: yes, i will create a bug and follow up. and add the todo in next PR to avoid triggering another round of tests.

PR #1308, opened 2025-12-14
Refactor tuning for RPA HD64 kernel tuning to improve RPA kernel throughput

helloworld1:
# Description

Improve block size tuning for RPA HD64 kernel

The following are changed
- Added tuning configuration that is tested on GPT OSS 120B, TP2, 1K prefill 8K decode / 8K prefill 1K decode. 
- Updated the tuning config table logic so it prioritize the tabled tuned values and use heuristic instead of combining the two
- Separately tune sliding window and full attention configuration

Compared with the default configuration, GPT OSS 120B TP2, 1024 req benchmark shows e2e throughput increase
1K prefill, 8K decode, +15% throughput
8K decode 1K prefill, +10% throughput

# Tests
E2E benchmark vllm bench serve
pytest tests/kernels/ragged_paged_attention_kernel_v3_hd64_test.py

# Checklist

### DIFF
--- a/tpu_inference/kernels/ragged_paged_attention/v3/kernel_hd64.py
+++ b/tpu_inference/kernels/ragged_paged_attention/v3/kernel_hd64.py
@@ -1428,12 +1428,20 @@ def ragged_paged_attention_hd64(
     page_size = kv_cache.shape[1]
     max_num_seqs = kv_lens.shape[0]
     num_page_indices = page_indices.shape[0]
-    assert num_page_indices % max_num_seqs == 0
-    pages_per_seq = num_page_indices // max_num_seqs
+
+    if sliding_window:
+        # When sliding window is only, it only need to access the tokens in the sliding window
+        # so kernel can use a different tuning of block sizes based on # of pages per seq
+        pages_per_seq = max(1, sliding_window // page_size)
+    else:
+        assert num_page_indices % max_num_seqs == 0
+        pages_per_seq = num_page_indices // max_num_seqs
+
     num_q_heads_per_kv_head = num_q_heads_per_kv_head_per_q_packing * q_packing
 
     bkv_p = num_kv_pages_per_block
     bq_sz = num_queries_per_block
+
     if bq_sz is None or bkv_p is None:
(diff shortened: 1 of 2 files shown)


tpu_inference/kernels/ragged_paged_attention/v3/kernel_hd64.py
```diff
@@ -1428,12 +1428,20 @@ def ragged_paged_attention_hd64(
     page_size = kv_cache.shape[1]
     max_num_seqs = kv_lens.shape[0]
     num_page_indices = page_indices.shape[0]
-    assert num_page_indices % max_num_seqs == 0
-    pages_per_seq = num_page_indices // max_num_seqs
+
+    if sliding_window:
+        # When sliding window is only, it only need to access the tokens in the sliding window
+        # so kernel can use a different tuning of block sizes based on # of pages per seq
+        pages_per_seq = max(1, sliding_window // page_size)
```
yuyanpeng-google, line 1435: Is this true for prefill? Although each query token only need sliding_window tokens, the query length affect the required kv length.

helloworld1: Good question. The tuning is true to prefill also. The kernel has an outer loop 
```
lax.fori_loop(0, num_bq, compute_with_bq, None, unroll=False)
```
And inner loop

```
            lax.fori_loop(bkv_idx_start,
                          num_bkv,
                          compute_with_bkv,
                          None,
                          unroll=False)
```

The hyperparameter `bq` depends on the query length. The `bkv_p` depends on the context length. 
I think it is the tuning key "max_model_len" is confusing here. It should be the `max_context_len`  to directly correspond to the `bkv_p` kernel parameter. I can change the naming here.

yuyanpeng-google: Is max_context_len = bq + sliding window? Should I say the `bq` is usually small, and we just ignore it and use sliding window as max_context_len?

PR #1310, opened 2025-12-14
Use 50 bit uuid for KV transfer key to avoid GO trunk the int in GKE

mrjunwan-lang:
# Description
GKE use GO to pass int, which will trunk bits more than 50, that cause missing KV id match in pull for a GKE env

Use 50 bit uuid for KV transfer key to avoid GO trunk the int in GKE

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax_tpu7x.yml
+++ b/.buildkite/pipeline_jax_tpu7x.yml
@@ -267,6 +267,7 @@ steps:
      env:
        USE_V6E8_QUEUE: "True"
        VLLM_LOG_LEVEL: "INFO"
+       IS_FOR_V7X: "true"
      agents:
        queue: tpu_v7x_8_queue
      commands:
--- a/.buildkite/scripts/run_in_docker.sh
+++ b/.buildkite/scripts/run_in_docker.sh
@@ -48,6 +48,9 @@ else
   exit 1
 fi
 
+# Some test scripts set tp=2 on IS_FOR_V7X=true to mitigate test failures.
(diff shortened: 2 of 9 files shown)


tpu_inference/distributed/tpu_connector.py
```diff
@@ -694,9 +694,9 @@ def get_finished(self) -> tuple[set[str], set[str]]:
 
 def get_uuid() -> int:
     int128 = uuid4().int
-    # Must be 64-bit int, otherwise vllm output encoder would raise error.
-    int64 = int128 >> 64
-    return int64
+    # Must be less than 64-bit int, otherwise vllm output encoder would raise error.
+    # use 50 bit to avoid GO trunk the int when doing JSon serialization
```
richardsliu, line 698: trunk -> truncate?

tpu_inference/distributed/tpu_connector.py
```diff
@@ -694,9 +694,9 @@ def get_finished(self) -> tuple[set[str], set[str]]:
 
 def get_uuid() -> int:
     int128 = uuid4().int
-    # Must be 64-bit int, otherwise vllm output encoder would raise error.
-    int64 = int128 >> 64
-    return int64
+    # Must be less than 64-bit int, otherwise vllm output encoder would raise error.
```
richardsliu, line 697: I think this comment is no longer relevant given the one below.

mrjunwan-lang: To make it more informative, we mention the limitation for both vllm and GKE llmd here.

PR #1322, opened 2025-12-16
Attention DP for Torchax backend

wenxindongwork:
# Description

This PR adds support for attention dp in the torchax backend. Attention DP is a sharding scheme where the attention layer is duplicated across the chips multiple times, while the MoE/MLP layer is sharded across all chips using pure TP/EP. It is particularly useful for increasing throughput for large MoE models where the effective number of attention heads is fewer than the number of chips. Effective means that we take into account kv cache and model quantization. Targeted use cases include DeepSeek, GPT-OSS 120b, and Qwen3 235b. 

### Usage 

Set environment variable `NEW_MODEL_DESIGN=True`
Pass argument `--additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'`

This UI is subject to change. The `NEW_MODEL_DESIGN=True` will likely be deprecated in the future but currently it is needed to support MoE kernel with 2D mesh as the default. 

### Example usage

**Attention DP + EP:** 

NEW_MODEL_DESIGN=True python examples/offline_inference.py --model=unsloth/gpt-oss-120b-BF16 --tensor_parallel_size=total_number_of_chips --async_scheduling **--enable_expert_parallel --additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'**

**Attention DP + TP:** 

NEW_MODEL_DESIGN=True python examples/offline_inference.py --model=unsloth/gpt-oss-120b-BF16 --tensor_parallel_size=total_number_of_chips --async_scheduling **--additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'**

Pure model DP is introduced in https://github.com/vllm-project/tpu-inference/pull/865, which is different from attention DP. Model DP replicates the entire model, whereas attention DP only replicates the attention layer. 

### Regression testing 

Verified no performance difference between HEAD (27da6f) and this branch (based on 27da6f) on the following benchmarks 

1. DP + TP (bf16)

`vllm serve unsloth/gpt-oss-120b-BF16 --max-model-len=8192 --max-num-batched-tokens=16384 --max-num-seqs=500 --no-enable-prefix-caching --gpu-memory-utilization=0.98 --tensor-parallel-size=4 --data-parallel-size=2  --download_dir=/mnt/disks/persist --async-scheduling`

`vllm bench serve --model unsloth/gpt-oss-120b-BF16 --dataset-name random --random-input-len 2048 --random-output-len 6144 --num-prompts 1000 --ignore-eos`

Branch: 10075.69
HEAD: 10122.32

4. EP (mxfp4)

`vllm serve openai/gpt-oss-120b --max-model-len=9216 --max-num-batched-tokens=16384 --max-num-seqs=1024 --no-enable-prefix-caching --gpu-memory-utilization=0.97 --tensor-parallel-size=2 --download_dir=/mnt/disks/persist --kv-cache-dtype=fp8 --async-scheduling --enable-expert-parallel`

`vllm bench serve --model=openai/gpt-oss-120b --dataset-name=random --random-input-len=1024 --random-output-len=8192 --num-prompts=1024 --ignore-eos`

Branch: 3751.50
HEAD: 3750.99

### MMLU

Verified accuracy with MMLU 

# Tests

Added E2E test for attention DP in the torchax path:

- Functional test for attention DP in the torchax path 
- Correctness test for attention DP in the torchax path

# Checklist

### DIFF
--- a/.buildkite/features/data_parallelism.yml
+++ b/.buildkite/features/data_parallelism.yml
@@ -39,7 +39,6 @@ steps:
 
   - label: "Performance tests for data_parallelism"
     key: "data_parallelism_PerformanceTest"
-    depends_on: "record_data_parallelism_CorrectnessTest"
     soft_fail: true
     agents:
       queue: tpu_v6e_8_queue
--- a/tests/e2e/test_data_parallel.py
+++ b/tests/e2e/test_data_parallel.py
@@ -229,7 +229,7 @@ def test_attention_data_parallelism(
     _check_correctness("Attention data parallelism", baseline_outputs,
                        dp_outputs)
 
-    # Different hardware gives different performance. This test runs on v6e_8    
+    # Different hardware gives different performance. This test runs on v6e_8
     _check_performance("Attention data parallelism",
                        baseline_time,
                        dp_time,


kyuyeunk:
nit: suggest renaming the pr to "support attention dp in torchax path" since it has much wider impact than just a single gpt-oss model & potentially provide performance improvement to wide arrange of torchax models.

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -28,7 +28,7 @@ def sample(
     if tpu_sampling_metadata.do_sampling:
         # Unshard the logits explicity to avoid latency increase.
         logits = jax.lax.with_sharding_constraint(
-            logits, NamedSharding(mesh, P(ShardingAxisName.ATTN_DATA, None)))
+            logits, NamedSharding(mesh, P(ShardingAxisName.MLP_DATA, None)))
```
wenxindongwork, line 45: @gpolovets1 Hi George, I made this sharding change which will impact DeepSeek I believe, so wanted to run through you to make sure I'm doing this correctly 

Before logits are sharded with `ATTN_DATA = ('data', 'attn_dp')`, but now they are sharded with `MLP_DATA = 'data'`. I think this makes more sense because the vocab dimension should sharded with `MLP_TENSOR = ('attn_dp', 'model', 'expert')` which includes the `attn_dp` axis. I think the vocab embedding must be sharded with the `attn_dp` axis otherwise it will replicate in that dimension and consume a lot HBM. Do you think my understanding is correct and that this change makes sense?

kyuyeunk (commented):
can you add tests to classes in these files as well?

- https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/vllm/test_unquantized.py
- https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/vllm/test_mxfp4.py
- https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/vllm/test_compressed_tensors_w8a8_int8.py
- https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/vllm/test_compressed_tensors_w8a8_fp8.py
- https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/vllm/test_compressed_tensors_moe.py

specifically, can you add one more parameter - something like `enable_attn_dp`? - in existing parameters like this? https://github.com/vllm-project/tpu-inference/blob/0a2fbe96a17aae565c333458f429011f9be3209e/tests/layers/vllm/test_unquantized.py#L110

i want to verify this feature doesn't break our code at per-layer level.

kyuyeunk:
also, is my understanding correct that we need to add following arguments to use this feature? Can you make this more explicit in the PR description?
- set environment variable `NEW_MODEL_DESIGN=True`
- pass argument `--additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'`

this doesn't seem like a good ux design. i'm okay if there is no other way, but do you know if there's a simpler approach? for instance, does vllm upstream have similar feature? if so, how can user enable it? and is it something that we can piggyback on?

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -124,25 +125,23 @@ def _gmm(lhs, rhs, rhs_scale, rhs_bias, group_sizes):
...
+        in_specs=(P(ShardingAxisName.MLP_DATA,
+                    None), P(None, ShardingAxisName.MLP_TENSOR,
+                             None), rhs_scale_spec, rhs_bias_spec,
+                  P(ShardingAxisName.MLP_DATA)),
+        out_specs=(P(ShardingAxisName.MLP_DATA, ShardingAxisName.MLP_TENSOR)),
         check_vma=False,
     )(lhs, rhs, rhs_scale, rhs_bias, group_sizes)
 
-    tp_size = mesh.shape["model"]
+    tp_size = mesh.shape['model'] * mesh.shape.get('attn_dp', 1)
```
kyuyeunk: just to double check, I assume you couldn't use `ShardingAxisName` here because that could be a tuple of strings and not just a silage string right?

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -162,7 +161,10 @@ def _gmm_all_reduce(lhs, rhs, rhs_scale, rhs_bias, group_sizes):
         m, g, n, k = lhs.shape[0], *rhs.shape
         tm, tk, tn = _get_tiling_size_for_gmm_kernel(m, k, n, g)
         if rhs_bias is not None:
-            shard_id = jax.lax.axis_index("model")
+            model_shard_id = jax.lax.axis_index("model")
```
kyuyeunk: > if "attn_dp" in mesh.axis_names

this looks too messy.

Given that `jax.lax.axis_index` can take in tuple as an argument, wouldn't it be possible to do something like this? `jax.lax.axis_index(ShardingAxisName.MLP_TENSOR).sum()`?

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -206,7 +205,7 @@ def expert_sharded_gmm(
     is_last_expert: bool,
     mesh: Mesh,
 ) -> jax.Array:
-    ep_size = mesh.shape["model"]
+    ep_size = mesh.shape["model"] * mesh.shape.get("attn_dp", 1)
```
kyuyeunk: nit: i see this kind of patter frequently but it looks quite messy. do you know if python provides a cleaner way to handle it? or maybe we can provide a helper function where
- Take shape dictionary as an one argument
- Either string or list of string as another argument.
- If the argument is string perform regular key/value fetch
- If the argument is list of string, iterate of the strings and perform key/value fetch.

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -247,21 +246,22 @@ def _gmm(lhs, rhs, rhs_scale, rhs_bias, group_sizes, group_offset):
     #       0, 0, 0, 0     0, 0, 0, 0     0, 0, 0, 0     D, D, D, D
     #        shard-0        shard-1        shard-2        shard-3
     # Each shards has 3 (row A), 2 (row B), 5 (row C) and 4 (row D).
-    lhs_spec = P("model") if is_last_expert else P()
-    rhs_scale_spec = None if rhs_scale is None else P("model")
-    rhs_bias_spec = None if rhs_bias is None else P("model")
+    lhs_spec = P(ShardingAxisName.EXPERT) if is_last_expert else P()
```
kyuyeunk: nit: declare a local variable `ep_p_spec = P(ShadingAxisName.EXPERT)` and use that throughout this function.

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -399,6 +398,9 @@ def fused_moe_func(
     assert gating_output.shape == (num_tokens, global_num_experts)
 
     topk_weights = jax.nn.softmax(gating_output.astype(jnp.float32), axis=-1)
+    # All-gather topk weights for attention dp
+    topk_weights = jax.lax.with_sharding_constraint(
```
kyuyeunk, line 412: mixing parallelism methods of `jax.lax.with_sharding_constraint` and `shard_map` (which is used in few lines below) doesn't look particularly nice. is it possible to choose either of two options and stick with it?

wenxindongwork: Just using the `shard_map` was not enough because `topk_weights` will be all-gathered twice, and if we add this sharding constraint here it will be all-gathered just once.

wenxindongwork:
re ux design: 

I believe currently vLLM upstream uses DP + EP/TP as the default for selected MoE models when `--data_parallel_size` is specified: https://docs.vllm.ai/en/latest/serving/data_parallel_deployment/ 

I don't see any vllm flags for differentiating model dp vs attention dp, so I introduced a new flag under `additional_config`. 

once the attention dp feature stabilizes in tpu_inference, we can also consider make attention dp the default behaviour for selected setups.

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -261,21 +244,22 @@ def _gmm(lhs, rhs, rhs_scale, rhs_bias, group_sizes, group_offset):
     #       0, 0, 0, 0     0, 0, 0, 0     0, 0, 0, 0     D, D, D, D
     #        shard-0        shard-1        shard-2        shard-3
     # Each shards has 3 (row A), 2 (row B), 5 (row C) and 4 (row D).
-    lhs_spec = P("model") if is_last_expert else P()
-    rhs_scale_spec = None if rhs_scale is None else P("model")
-    rhs_bias_spec = None if rhs_bias is None else P("model")
+    lhs_spec = ep_p_spec if is_last_expert else P()
+    rhs_spec = P(ShardingAxisName.EXPERT, None, None)
```
kyuyeunk: This can also be replaced with ep_p_spec

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -261,21 +244,22 @@ def _gmm(lhs, rhs, rhs_scale, rhs_bias, group_sizes, group_offset):
...
-            P("model", None, None),
+            rhs_spec,
             rhs_scale_spec,
             rhs_bias_spec,
             P(),
-            P("model"),
+            ep_p_spec,
         ),
-        out_specs=(P("model", None)),
+        out_specs=(P(ShardingAxisName.EXPERT, None)),
```
kyuyeunk: This can also be replaced with ep_p_spec

kyuyeunk (approved):
Made mostly nit comments. Otherwise, LGTM!

tpu_inference/models/jax/llama3.py
```diff
@@ -99,7 +100,8 @@ def __init__(self, config: LlamaConfig, dtype: jnp.dtype, rng: nnx.Rngs,
                                          self.hidden_size // self.num_heads)
         self.head_dim = utils.get_padded_head_dim(self.head_dim_original)
 
-        sharding_size = mesh.shape["model"] * mesh.shape.get("attn_dp", 1)
+        sharding_size = get_mesh_shape_product(mesh,
+                                               ShardingAxisName.MLP_TENSOR)
```
samos123, line 104: Why did `sharding_size` change from `tp * dp` to just `tp`?

Was there a bug previously?

samos123:
You mentioned this as an example for Attention DP + EP:

```
NEW_MODEL_DESIGN=True python examples/offline_inference.py --model=unsloth/gpt-oss-120b-BF16 --tensor_parallel_size=total_number_of_chips --async_scheduling --enable_expert_parallel --additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'
```

However, doesn't this set expert parallelism to 1 and tp=16? but the description this is DP + EP?

tpu_inference/models/jax/llama3.py
wenxindongwork: Since `ShardingAxisName.MLP_TENSOR = ('attn_dp', 'model', 'expert')`, we preserve the original behavior.

ref: https://github.com/vllm-project/tpu-inference/blob/a8da40d1aecb540107b2c37d4543b165ef0387ea/tpu_inference/layers/common/sharding.py#L40C5-L40C48

wenxindongwork:
> You mentioned this as an example for Attention DP + EP:
> 
> ```
> NEW_MODEL_DESIGN=True python examples/offline_inference.py --model=unsloth/gpt-oss-120b-BF16 --tensor_parallel_size=total_number_of_chips --async_scheduling --enable_expert_parallel --additional_config='{"sharding":{"sharding_strategy": {"enable_dp_attention":1}}}'
> ```
> 
> However, doesn't this set expert parallelism to 1 and tp=16? but the description this is DP + EP?

When using attention DP, the user needs to provide the total number of chips via the `--tensor_parallel_size=total_number_of_chips` flag, and the engine will default to DP + TP.  In order to activate expert parallelism and use DP + EP, please add the `--enable_expert_parallel` flag.  

The engine will calculate for the user how many attention replicas to create based on the model configuration (by dividing the total number of chips by the number of kv heads). Ref: https://github.com/vllm-project/tpu-inference/blob/a8da40d1aecb540107b2c37d4543b165ef0387ea/tpu_inference/layers/common/sharding.py#L134-L160

PR #1331, opened 2025-12-18
[Kernel] Simplify the MLA bkv loading logic

yaochengji:
# Description

The packed tokens are already in the right order, so we don't need to bit-shift, concat or reshape the tensor.

# Tests

pytest -s -v tests/kernels/mla_v1_test.py

# Checklist


kyuyeunk:
> The packed tokens are already in the right order, so we don't need to bit-shift, concat or reshape the tensor.

slightly confused by this. does it mean that the existing logic did all those reordering? wouldn't that have caused numeric error?

yaochengji:
> slightly confused by this. does it mean that the existing logic did all those reordering? wouldn't that have caused numeric error?

@kyuyeunk the existing logic doesn't reorder either, but more complex.

PR #1332, opened 2025-12-18
[DeepSeek] Support TPU-Friendly Checkpoints + Add DeepSeek Testing

jrplatin:
# Description

In this PR, I do the following

* Build upon excellent work by @amishacorns to load in FP4-MLP/FP8 attention TPU-friendly checkpoints, namely:
    * Change most of the attention weights to be 2D (except for the `k/v_up_proj` for MLA, which had issues with Qwix not quantizing it properly) + update shardings accordingly
    * Move the `combine_weights/experts` into its own module so that we don't quantize it (helping with accuracy)
    * Update weight loading logic to handle the new TPU-friendly checkpoints, including unpacking properly and removing the need for complex reshape scale/weight maps
    * Update the default DeepSeek quantization Qwix rules to default to the TPU-friendly checkpoint format (FP4 MLP, FP8 attention)

# Side Effects
Because there is current little demand for the pure-FP8 checkpoint, I did **not** add reverse compatibility for the existing default FP8 checkpoints (on HuggingFace).  **Thus, it's expected that all DeepSeek checkpoints will be using the mixed FP4/FP8 format.**.  We can add the reverse compatibility in the future as needed.

**NOTE:** Currently, there is a major regression in accuracy when using `max-num-seqs > ` for the FP4 checkpoints.  We are currently investigating this.

# Accuracy and Performance Verification
Accuracy and performance were validated on MMLU with/without MLA.  Performance was additionally verified via XProf.

# Tests

Updated relevant testing for `test_quantization` + DeepSeek attention + added a new suite of tests for DeepSeek as a whole.  Also ran BF16 + FP4 locally (with/without MLA) and confirmed offline inference results are as expected.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -136,6 +136,7 @@ steps:
            --ignore=/workspace/tpu_inference/tests/e2e \
            --ignore=/workspace/tpu_inference/tpu_inference/mock \
            --ignore=/workspace/tpu_inference/tests/layers/vllm/test_compressed_tensors_moe.py \
+           --ignore=/workspace/tpu_inference/tests/models/jax/test_deepseek_v3.py \
            --cov-config=/workspace/tpu_inference/.coveragerc --cov tpu_inference --cov-report term-missing --cov-fail-under=69
 
    - label: "JAX unit tests - kernels"
--- a/tests/layers/jax/test_qwix.py
+++ b/tests/layers/jax/test_qwix.py
@@ -480,6 +480,10 @@ class TestLoadRandomWeightsIntoQwixAbstractModel(unittest.TestCase):
         mock_random_array = jax.numpy.ones(1)
         mock_get_random_array.return_value = mock_random_array
 
+        self.model.weight_loader.scale_shape_map_for_random_weight_loading = {
+            'attention.wq': (1, 1)
+        }
+
         mock_iter_graph.return_value = [
             (('layers', '0', 'attention', 'wq', 'kernel'), mock_weight_param),
             (('layers', '0', 'attention', 'wq', 'array', 'scale'),
@@ -521,22 +525,9 @@ class TestLoadRandomWeightsIntoQwixAbstractModel(unittest.TestCase):
              mock_scale_var),
         ]
 
-        quantize_qwix.load_random_weights_into_qwix_abstract_model(
-            self.rng, self.model, self.mesh, self.quantization_config)
-
-        new_weight_param_val = mock_weight_param.value
-        new_scale_var_val = mock_scale_var.value
-
-        expected_scale_shape = (128 // 64, 64 // 64)
-        actual_scale_shape = new_scale_var_val.shape
-
-        expected_weight_shape = (128, 64)
-        actual_weight_shape = new_weight_param_val.shape
-
-        self.assertEqual(expected_scale_shape, actual_scale_shape)
-        self.assertEqual(expected_weight_shape, actual_weight_shape)
-        self.assertNotEqual(old_scale_var_val.shape, new_scale_var_val.shape)
-        assert jnp.not_equal(old_weight_param_val, new_weight_param_val).all()
+        with self.assertRaises(ValueError):
+            quantize_qwix.load_random_weights_into_qwix_abstract_model(
+                self.rng, self.model, self.mesh, self.quantization_config)
 
     @patch('tpu_inference.models.jax.utils.qwix.qwix_utils.nnx.iter_graph')
     def test_param_shape_setting_with_scale_map(self, mock_iter_graph):
@@ -549,7 +540,7 @@ class TestLoadRandomWeightsIntoQwixAbstractModel(unittest.TestCase):
         expected_scale_shape = (55, 34)
 
         self.model.weight_loader.scale_shape_map_for_random_weight_loading = {
-            'wq': expected_scale_shape
+            'attention.wq': expected_scale_shape
         }
 
         mock_iter_graph.return_value = [
@@ -841,17 +832,16 @@ class TestGetDefaultQwixQuantizationConfig(unittest.TestCase):
         # Patch the constants in the module where the function resides
         self.patchers = [
             patch(
-                "tpu_inference.models.jax.utils.quantization.quantization_utils.DEFAULT_DEEPSEEK_FP8_CONFIG",
+                "tpu_inference.models.jax.utils.qwix.qwix_utils.DEFAULT_DEEPSEEK_FP8_CONFIG",
                 self.mock_deepseek_config),
             patch(
-                "tpu_inference.models.jax.utils.quantization.quantization_utils.DEFAULT_LLAMA4_FP8_CONFIG",
+                "tpu_inference.models.jax.utils.qwix.qwix_utils.DEFAULT_LLAMA4_FP8_CONFIG",
                 self.mock_llama_config),
             patch(
-                "tpu_inference.models.jax.utils.quantization.quantization_utils.DEFAULT_GPT_OSS_FP4_CONFIG",
+                "tpu_inference.models.jax.utils.qwix.qwix_utils.DEFAULT_GPT_OSS_FP4_CONFIG",
                 self.mock_gpt_oss_config),
-            patch(
-                "tpu_inference.models.jax.utils.quantization.quantization_utils.logger",
-                MagicMock())
+            patch("tpu_inference.models.jax.utils.qwix.qwix_utils.logger",
+                  MagicMock())
         ]
         for p in self.patchers:
             p.start()
--- a/tests/models/jax/test_deepseek_v3.py
+++ b/tests/models/jax/test_deepseek_v3.py
@@ -59,7 +59,7 @@ class MockVllmConfig:
 
         # DeepSeek V3 specific config
         hf_config = MagicMock()
-        hf_config.num_hidden_layers = 4  # Small for testing
+        hf_config.num_hidden_layers = 1  # Small for testing
         hf_config.num_nextn_predict_layers = 1
         self.model_config.hf_config = hf_config
 
@@ -104,9 +104,9 @@ class TestDeepSeekV3:
     def test_init(self, mock_config, rng, mesh):
         """Tests if the model initializes with the correct hierarchy."""
         model = DeepSeekV3(mock_config, rng, mesh)
-        assert len(model.layers) == 4  # num_layers from mock
+        assert len(model.layers) == 3  # num_layers from mock
         assert isinstance(model.embedder, nnx.Module)
-        assert model.vllm_config.model_config.hf_config.num_hidden_layers == 4
+        assert model.vllm_config.model_config.hf_config.num_hidden_layers == 1
 
     def test_random_weights(self, mock_config, rng, mesh):
         """Tests that force_random_weights initializes non-zero weights."""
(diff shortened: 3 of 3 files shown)


tpu_inference/layers/jax/attention/deepseek_v3_attention.py
```diff
@@ -209,9 +214,10 @@ def __call__(self,
             q_TA = jnp.einsum("TD,DA -> TA", x_q_TD,
                               self.kernel_q_down_proj_DA.value)
             q_TA = self.q_rms_norm(q_TA)
-            # Query up projection.
-            q_TNH = jnp.einsum("TA,ANH -> TNH", q_TA,
-                               self.kernel_q_up_proj_ANH.value)
+            # Query up projection, then reshape to TNH.
+            q_TC = jnp.einsum("TA,AC -> TC", q_TA,
+                              self.kernel_q_up_proj_AC.value)
+            q_TNH = q_TC.reshape(q_TA.shape[0], self.N, self.qk_head_dim)
```
gpolovets1: Could we add info to the bug confirming whether this adds much latency overhead?

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
```diff
@@ -52,8 +52,9 @@ class MLA(nnx.Module):
     rms_norm_eps: float
 
     # Sharding attributes
-    nhd_sharding: Sharding = ()
+    cd_sharding: Sharding = ()
```
gpolovets1: Should we update the [glossary](https://github.com/vllm-project/tpu-inference/blob/a085cc6b980fd21d8b49c7bedfbe09b1f29957c4/tpu_inference/layers/jax/glossary.md#L4)? It actually appears C was previously reserved for 'expert capacity'.

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
```diff
@@ -127,6 +128,10 @@ def __post_init__(self, rngs: nnx.Rngs):
             self.dtype,
             random_init=self.random_init,
         )
+        # NOTE (jacobplatin): we are keeping these variables as 3D because
```
gpolovets1, line 144: Why not just keep all of them in 3D and avoid reshaping if the TPU-friendly checkpoints are all cleanly divisible?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -12,6 +12,29 @@
 modeling_flax_utils = FlaxUtils()
 
 
+@dataclass(kw_only=True)
+class CombineExperts(nnx.Module):
```
gpolovets1, line 30: Why do we need this new class?

tpu_inference/models/jax/utils/quantization/quantization_utils.py
gpolovets1, line 38: Should we add a new config for FP4 and allow the user to load FP4 or FP8?

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -417,9 +443,42 @@ def get_default_qwix_quantization_config(
...
+    # more flexible
+    # NOTE (jacobplatin): we'll default to mixed FP8 (attention) + FP4 (MoE experts)
+    # for DeepSeek
     if model_type == "deepseek_v3" and quant_method == "fp8":
-        return DEFAULT_DEEPSEEK_FP8_CONFIG
+        config = copy.deepcopy(DEFAULT_DEEPSEEK_FP8_CONFIG)
+
+        # Dynamically fetch block size from HF config if available
+        # Config fmt: 'weight_block_size': [1, 512] -> we want the 2nd dim for tile_size
+        # NOTE: if the checkpoint is not 1D subchannel, we will throw an error
```
gpolovets1, line 459: What should we do if we need to troubleshoot accuracy issues if the only checkpoints we can load are TPU-friendly checkpoints?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -512,66 +506,43 @@ def __init__(self,
...
             # TODO (jacobplatin): remove or clean this up
             self.scale_shap_map_for_random_weight_loading = {
-                "kernel_kv_down_proj_DA": (56, 576),
-                "kernel_kv_up_proj_ANH": (4, 128, 2),
-                "kernel_q_up_proj_ANH": (12, 1, 192),
-                "kernel_o_proj_NHD": (128, 1, 56),
-                "kernel_down_proj_EFD": (256, 16, 56),
-                "kernel_up_proj_EDF": (256, 56, 16),
-                "kernel_gating_EDF": (256, 56, 16),
+                # MoE experts (3D)
```
gpolovets1, line 534: Thanks for organizing by section and adding comments!

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -512,66 +506,43 @@ def __init__(self,
...
+                "attn.kernel_q_up_proj_AC": (6, 24576),
+                "attn.kernel_kv_down_proj_DA": (28, 576),
+                "attn.kernel_kv_up_proj_AC": (2, 32768),
+                "attn.kernel_o_proj_CD": (64, 7168),
+                "attn.kernel_k_up_proj_ANH": (2, 128, 128),  # MLA
+                "attn.kernel_v_up_proj_ANH": (2, 128, 128),  # MLA
             }
 
+            # TODO (jacobplatin): remove this check eventually!
+            assert self.quant_dtype == jnp.float8_e4m3fn, f"Expected quant_dtype to be float8_e4m3fn for DeepSeek but got {self.quant_dtype}"
```
gpolovets1, line 557: I thought you were still supporting FP8 in the rest of the code as well?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -649,45 +620,60 @@ def _load_individual_weight(self,
...
-                    scale = scale.repeat(self.quantization_block_size_n,
-                                         axis=idx)[:, :weight_dim]
+            if len(weight_shape) == len(scale_shape):
+                new_scale = scale
+                for idx, (wdim,
+                          sdim) in enumerate(zip(weight_shape, scale_shape)):
+                    # If divisible, leave scale as-is for this axis
+                    if (wdim % sdim == 0):
+                        continue
+                    # Otherwise repeat and then truncate to match weight dim
```
gpolovets1: Just confirming if this will retain the scale mapping? In previous conversations we talked about repeating the scales before reshaping into the final dimensions. Will repeating the scales correctly preserve the mapping though?

tpu_inference/layers/jax/moe/moe.py
jrplatin: This is mostly get around an issue where the Qwix rules weren't being picked up properly, so we separated this out for ease of targeting (so the combine op isn't quantized)

tpu_inference/models/jax/utils/quantization/quantization_utils.py
jrplatin: Oh good catch -- I should probably update the name of this to "FP4_MLP_FP8_ATTN" or something -- I think we agreed offline though that we weren't going to support the existing FP8 checkpoints for now?

tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin: I think we synced offline on this, but, if that's the case, we can just add that code back fairly easily

tpu_inference/models/jax/deepseek_v3.py
jrplatin: See above

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
jrplatin: Yes, @JiriesKaileh is working on this

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
jrplatin: At very least, o-proj isn't and (as will be discussed above), I don't believe there will be much overhead + it will make the code cleaner, so I think it's a fair trade-off

tpu_inference/models/jax/deepseek_v3.py
jrplatin: Double confirming but I actually don't think this logic is used so will remove

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
jrplatin: Ah didn't know this existed -- let me disambiguate this

gpolovets1: Thanks!

tpu_inference/layers/jax/moe/moe.py
gpolovets1: Ah interesting, I thought specifying the weight regex would allow Qwix to pick it up.

tpu_inference/models/jax/utils/quantization/quantization_utils.py
gpolovets1: I see, I think I got a bit confused because it seemed like the weight loading code was still checking for FP8 and supporting it.

tpu_inference/models/jax/utils/qwix/qwix_utils.py
gpolovets1: Thanks for reminding! Could you add a few of these decisions to the bug for posterity/avoid me re-asking in the future? :D

jrplatin: Of course!

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
```diff
@@ -247,9 +253,12 @@ def __call__(self,
                 k_rope_SNH = jnp.broadcast_to(
                     k_rope_SNH,
                     (k_rope_SNH.shape[0], self.N, self.qk_rope_head_dim))
-                # KV up projection.
-                kv_nope_SNH = jnp.einsum("SA,ANH -> SNH", kv_SA,
-                                         self.kernel_kv_up_proj_ANH.value)
+                # KV up projection, then reshape to SN(Hk+Hv).
+                kv_SNC = jnp.einsum("SA,AC -> SC", kv_SA,
```
jrplatin: Update the kv suffix here

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
jrplatin: Jiries and I looked into 3d random weights more, and, for some reason, the XProf (as shared in comment#8) was consistently produced with completely wonky MoE times (even after trying rebasing on `main`, trying to isolate the problematic changes, etc.), so I decided to try real weights instead on branch `jacobplatin/ds-tpu-fp4-load-3d-weights`, which produced [this](https://xprof.corp.google.com/trace_viewer/jacobplatin-3883812444327104446?hosts=t1v-n-c9c02cfd-w-0&host_index=0&trace_filter_config={}&view_start=111.121&view_end=114.962) XProf with a latency of around 3ms (very similar to the [2.94ms](https://screenshot.googleplex.com/BfttKFuFJxMiTP8) obtained with 2D reshape), so confident the 2D w/ reshape vs. 3D w/o reshape isn't an issue

tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin: Added [here](https://b.corp.google.com/issues/469158045#comment17)!

PR #1334, opened 2025-12-18
[Refactoring] use `jax.shard_map` instead of experimental one

lk-chen:
# Description

jax.experimental.shard_map is deprecated in v0.8.0 [source](https://github.com/jax-ml/jax/blob/c617efcdbb40281a260488e0e3a00f12651da021/jax/experimental/shard_map.py#L25-L31). This PR replace the experimental usage with `jax.shard_map`, also use `check_vma` instead of `check_rep` to follow the new argument naming

# Tests

Rely on CI/CD

# Checklist


kyuyeunk:
nit: instead of `jax.shard_map`, to be consistent with other code, can you do this instead?: 

```
from jax import shard_map
```  

example: https://github.com/vllm-project/tpu-inference/blob/0a2fbe96a17aae565c333458f429011f9be3209e/tpu_inference/layers/vllm/fused_moe.py#L131-L143

lk-chen:
Hi @kyuyeunk thanks for the suggestion, however I think "from * import shard_map" is the minority here (10 files v.s. 6files)
and `jax.shard_map` aligns with `jax.jit`

image
image

kyuyeunk:
Sgtm. If so, can you update other code to use Jax.shard_map as well?

lk-chen:
> Sgtm. If so, can you update other code to use Jax.shard_map as well?

Done, the checks all passed. PTAL, thx!

PR #1335, opened 2025-12-19
[Misc] Fix tpu platform init failure when vllm_config is not fully initialized

sixiang-google:
# Description

Caused by https://github.com/vllm-project/vllm/pull/30531/files#diff-6b7df8c9b2e791e958973a94a2027ffa4e6acafc3fbcb275d48b6e22e34fc94eR16-R19

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/platforms/tpu_platform.py
+++ b/tpu_inference/platforms/tpu_platform.py
@@ -145,17 +145,20 @@ class TpuPlatform(Platform):
             compilation_config.backend = "openxla"
 
         # TODO(cuiq): remove this dependency.
-        from vllm.v1.attention.backends.pallas import PallasAttentionBackend
-        cache_config.block_size = PallasAttentionBackend.get_page_size(
-            vllm_config)  # type: ignore[assignment]
-        min_page_size = PallasAttentionBackend.get_min_page_size(vllm_config)
-        if min_page_size > cache_config.block_size:
-            logger.warning(
-                "Increase the page size from %s to %s to avoid SMEM OOM",
-                cache_config.block_size,
-                min_page_size,
-            )
-            cache_config.block_size = min_page_size  # type: ignore[assignment]
+        if vllm_config.model_config:
+            from vllm.v1.attention.backends.pallas import \
+                PallasAttentionBackend
+            cache_config.block_size = PallasAttentionBackend.get_page_size(
+                vllm_config)  # type: ignore[assignment]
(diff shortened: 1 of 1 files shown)


tpu_inference/platforms/tpu_platform.py
```diff
@@ -145,17 +145,20 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                vllm_config)  # type: ignore[assignment]
+            min_page_size = PallasAttentionBackend.get_min_page_size(
+                vllm_config)
+            if min_page_size > cache_config.block_size:
+                logger.warning(
+                    "Increase the page size from %s to %s to avoid SMEM OOM",
+                    cache_config.block_size,
+                    min_page_size,
+                )
+                cache_config.block_size = min_page_size  # type: ignore[assignment]
```
QiliangCui, line 161: if so, when this value will be set?

tpu_inference/platforms/tpu_platform.py
```diff
@@ -197,8 +200,8 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         # Late initialization to avoid circular import
         from tpu_inference.models.jax.utils.quantization.quantization_utils import \
             update_vllm_config_for_qwix_quantization
-
-        update_vllm_config_for_qwix_quantization(vllm_config)
+        if vllm_config.model_config:
+            update_vllm_config_for_qwix_quantization(vllm_config)
```
QiliangCui, line 204: in this case, when will `update_vllm_config_for_qwix_quantization` be called? if never.. is it safe to launch without initializing?

tpu_inference/platforms/tpu_platform.py
sixiang-google: Full stack trace of this bug: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7147#019b3079-5883-449a-bc3f-7e0be9839e05/L262-L299. 

This problem happens when some submodules in vllm model impl (for example `SiluAndMul()`) initializes creation of empty `Vllmconfig` in vllm/vllm/config/vllm.py line 1407, and trigger the hook in `tpu_platform.py`. This is why we can safely ignore when the mentioned fields are yet to be set. The later initialization of VllmConfig with full model config will overwrite this.

QiliangCui: thank you!

PR #1340, opened 2025-12-19
[DP] Reduce DP scheduling overhead via multiprocessing

wenxindongwork:
# Description

Scheduling is a O(num_request) operation. When there are too many requests, this overhead shows.
E.g. Currently to schedule 4096 requests, it takes ~80ms to schedule, DP scheduler and the default vLLM scheduler both take this long, which this is very slow. 

With this PR, scheduling 4096 requests with DP=4 takes ~25m.

For more context: b/468694564#comment6

# Tests

Updated DP scheduler tests 
Correctness verified by e2e DP correctness tests. 

# Checklist

### DIFF
--- a/tpu_inference/core/sched/dp_scheduler.py
+++ b/tpu_inference/core/sched/dp_scheduler.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import copy
+import multiprocessing.reduction
 from collections import defaultdict, deque
 from dataclasses import dataclass
 from enum import Enum
@@ -20,9 +21,7 @@ from multiprocessing import Process, Queue
 from typing import Any, Dict, List, Optional, Tuple
 
 import cloudpickle
-import multiprocessing.reduction
 import torch
-from tpu_inference.utils import time_function
 from vllm.config import VllmConfig
 from vllm.multimodal import MULTIMODAL_REGISTRY, MultiModalRegistry
 from vllm.v1.core.sched.async_scheduler import AsyncScheduler
@@ -38,6 +37,7 @@ from vllm.v1.request import Request
 from vllm.v1.structured_output import StructuredOutputManager
 
 from tpu_inference.logger import init_logger
+from tpu_inference.utils import time_function
 
 logger = init_logger(__name__)
 
@@ -62,6 +62,7 @@ class SchedulerCommand(Enum):
 
 class SchedulerWorkerError(Exception):
     """Exception raised when a scheduler worker process encounters an error."""
+
     def __init__(self, rank: int, message: str):
         self.rank = rank
         self.message = message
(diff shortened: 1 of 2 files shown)


kyuyeunk:
I know this is a still wip, but want to mention that this is a really awesome fix! Thank you for coming up with the idea to use multiprocessing to fix the issue! Really clever idea that is also scalable even if parallelism size gets bigger!

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -37,6 +42,171 @@
...
+    )
+    
+    logger.debug(f"Scheduler worker process {rank} started")
+    
+    # Process commands from the input queue
+    while True:
+        try:
+            command, data = input_queue.get()
+            
+            if command == SchedulerCommand.ADD_REQUEST:
```
kyuyeunk: Please change this into match/case for better readability.

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -37,6 +42,171 @@
...
+
+class SchedulerWorkerError(Exception):
+    """Exception raised when a scheduler worker process encounters an error."""
+    def __init__(self, rank: int, message: str):
+        self.rank = rank
+        self.message = message
+        super().__init__(f"Scheduler worker {rank} error: {message}")
+
+
+# Monkey-patch multiprocessing to use cloudpickle
```
kyuyeunk, line 72: Can you elaborate more on why monkey patching is needed?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -37,6 +42,171 @@
...
+        block_size=block_size,
+        mm_registry=mm_registry,
+        include_finished_set=include_finished_set,
+        log_stats=log_stats,
+    )
+    
+    logger.debug(f"Scheduler worker process {rank} started")
+    
+    # Process commands from the input queue
+    while True:
```
kyuyeunk, line 130: Slightly concerned about potential infinite loop bug. Is it possible to change the logic to check whether some variable is true or not?

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: standard pickling can't serialize the vLLM `Request` object, so we need to use cloudpickle instead of the default pickle.

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: good point! added a fallback case, if there is no match for `command` the loop will break and the process will exist.

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -37,6 +42,177 @@
...
+                    output_queue.put(error)
+                    raise error
+                
+        except Exception as e:
+            logger.error(f"Error in scheduler worker {rank}: {e}", exc_info=True)
+            # Put error on output queue
+            error = SchedulerWorkerError(rank, str(e))
+            try:
+                output_queue.put(error)
+            except:
```
kyuyeunk: So we are catching any error and simply do a pass on them? This doesn't seem safe? It might be better to catch a specific error that can be safe to pass.

wenxindongwork: the main process will raise that error in `_get_result_from_queue`.

kyuyeunk: Okay go it. But why wrap 'output_queue.put' with try except?

wenxindongwork: ohh thanks for catching! removed the try catch

PR #1341, opened 2025-12-19
Precompile functions with large vocab_size tensors before allocating KV cache to avoid OOM

wenxindongwork:
# Description

Precompile functions with large vocab_size tensors before allocating KV cache to avoid OOM

For some functions, the input tensors are too large and there is not enough HBM to precompile them after KV cache initialization. Precompiling these functions before kv cache initialization helps to avoid HBM OOM. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -524,8 +524,7 @@ class CompilationManager:
                     num_reqs=num_reqs,
                     do_sampling=do_sampling,
                 )
-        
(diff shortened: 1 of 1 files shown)


kyuyeunk:
can you verify recompilation doesn't get triggered during runtime? the change looks fine & as soon as i have that validation, I can give approval.

wenxindongwork:
confirmed with VLLM_XLA_CHECK_RECOMPILATION=1, no recompilation triggered during runtime.

PR #1342, opened 2025-12-19
[Kernel][FusedMoe] Fix sync-barrier caused crash

bythew3i:
Root cause

- sync_barrier needs to sync all devices instead of only right neighbor.
- Insert more sync barrier to avoid race condition

Other changes:

- Fuse retiling in load and store to make them more efficient
- Refactor some code to make it more readable

### DIFF
--- a/tpu_inference/kernels/fused_moe/v1/kernel.py
+++ b/tpu_inference/kernels/fused_moe/v1/kernel.py
@@ -48,24 +48,6 @@ def broadcast_minor(src, shape):
                            axis=-1)[..., :shape[-1]]
 
 
-# TODO(jevinjiang): Apply the same trick for load.
-def store_with_retiling(ref, val, is_acc=False):
-    """Retiling (8, 128) to (1, 128) in store."""
-    assert ref.shape == val.shape
-    assert ref.dtype == val.dtype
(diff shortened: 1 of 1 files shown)


tpu_inference/kernels/fused_moe/v1/kernel.py
```diff
@@ -19,8 +19,7 @@ def align_to(x, a):
 
 
 def get_dtype_packing(dtype):
-    bits = (dtypes.bit_width(dtype)
-            if hasattr(dtypes, "bit_width") else dtypes.itemsize_bits(dtype))
+    bits = dtypes.itemsize_bits(dtype)
```
kyuyeunk: can you revert this change? this is for backward compatibility with older jax version. note that we are currently working on defaulting to 0.8.1 and we can remove this if/else once that lands.

QiliangCui:
manually triggered a test: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7191

tpu_inference/kernels/fused_moe/v1/kernel.py
bythew3i: Done.

PR #1349, opened 2025-12-20
Refactor MoE MLP to use split gate and up projections.

abhinavclemson:
# Description

Refactor MoE MLP to use split gate and up projections.

This change updates the MoE layer to align with MaxText's structure, using separate kernels and biases for the gate and up projections in the SwiGLU MLP. The checkpoint loading is also updated to map the new parameter names, and Sharding annotations were adjusted.

# Tests

Tested with Maxtext + Tunix + tpu-inference + vllm GRPO script.

# Checklist


tpu_inference/models/jax/gpt_oss.py
```diff
@@ -276,11 +277,16 @@ def load_weights(self, rng: PRNGKey, cache_dir: Optional[str] = None):
              transforms["transpose"], None),
             "model.layers.*.mlp.router.bias":
             ("layers.*.custom_module.router.bias_E", None, None),
-            "model.layers.*.mlp.experts.gate_up_proj":
-            ("layers.*.custom_module.mlp1_weight_EDF2", swap_mlp_transform,
+            "model.layers.*.mlp.experts.gate_proj":
```
bzgoogle: may i know which checkpoint it will read for the splited tensor? is it Maxtext generated checkpoint?  From the HF checkpoint, it is combined tensors: as *gate_up_proj_blocks  https://screenshot.googleplex.com/XUZkNDkHJB7TS2r

kyuyeunk (commented):
lgtm. this approach also aligns with how things are done in torchax path
- https://github.com/vllm-project/tpu-inference/blob/d232467eea6fce036a4eb412409aaec92a0ff083/tpu_inference/layers/vllm/process_weights/fused_moe_weights.py#L143-L157
- https://github.com/vllm-project/tpu-inference/blob/d232467eea6fce036a4eb412409aaec92a0ff083/tpu_inference/layers/common/fused_moe_gmm.py#L40-L49

but will require approval from people who are actively working on jax path (like @bzgoogle) to be merged.

tpu_inference/models/jax/gpt_oss.py
abhinavclemson: Addressed!

PR #1366, opened 2025-12-23
[DeepSeek] Use FP4 checkpoint instead of FP8 for E2E tests + add more clear warnings

jrplatin:
# Description

Because https://github.com/vllm-project/tpu-inference/pull/1332 removed support for FP8 DeepSeek checkpoints, we need to update the tests accordingly to use the TPU-friendly FP4 [checkpoints](https://huggingface.co/collections/jrplatin/deepseek-r1-1d-subchannel).

# Tests

Tested that BuildKite Nightly passes.

# Caveat
Because the current `requirements.txt` file uses `libtpu==0.0.24`, which doesn't support FP4, I am disabling the V6E test for DeepSeek here, but we should re-enable it when [this](https://b.corp.google.com/issues/467105149#comment11) bug is fixed.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -198,24 +198,25 @@ steps:
            exit 0
          fi
 
-   - label: "E2E MLperf tests for DeepSeek-R1 (no accuracy, 12-decoder layers only)"
-     key: test_12
-     soft_fail: true
-     env:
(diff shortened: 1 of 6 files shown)


bzgoogle:
A qq: for sparse Moe used in deepseek, we used to have the manual fp8 quantization on moe tensors. Will those function need to be updated to support fp4?

PR #1373, opened 2025-12-27
[CI] Dynamic generate notification for different pipelines

weiyu0824:
# Description

Dynamically generate notification.yml inside bootstrap.sh. For integration pipeline, we will notify oncaller, otherwise just notify the build creator. We don't have release/nightly pipeline right now, so just notify build creator for now.

Pipeline Cleanup: Removed redundant notify blocks from:
- `.buildkite/pipeline_integration.yml` (renamed to `integration_promote.yml` cause this yml only used in "integration" pipeline)
- `.buildkite/nightly_releases.yml`
- `.buildkite/main.yml` (renamed to `nightly_verify.yml`)
- `.buildkite/pipeline_jax.yml`

# Tests

# Checklist

### DIFF
--- a/.buildkite/scripts/bootstrap.sh
+++ b/.buildkite/scripts/bootstrap.sh
@@ -100,7 +100,7 @@ NOTIFY_FILE="generated_notification.yml"
 #      it won't alert the oncall team.
 
 if [[ "$BUILDKITE_PIPELINE_SLUG" == "tpu-vllm-integration" && "$BUILDKITE_SOURCE" == "schedule" ]] || \
-   [[ "$NIGHTLY" == "1" && "$BUILDKITE_SOURCE" == "schedule" ]]; then
+   [[ "${NIGHTLY:-0}" == "1" && "$BUILDKITE_SOURCE" == "schedule" ]]; then
     echo "Context: Scheduled Integration/Nightly. Notifying Oncall."
     cat <<EOF > "$NOTIFY_FILE"
 notify:


.buildkite/scripts/bootstrap.sh
```diff
@@ -78,21 +78,55 @@ upload_pipeline() {
     unset LABEL_PREFIX KEY_PREFIX TPU_QUEUE_SINGLE TPU_QUEUE_MULTI IS_FOR_V7X COV_FAIL_UNDER
 
     # buildkite-agent pipeline upload .buildkite/pipeline_torch.yml
-    buildkite-agent pipeline upload .buildkite/main.yml
```
mrjunwan-lang, line 85: Why we remove the main.yml? the exiting CI/CD will broken, right?

.buildkite/scripts/bootstrap.sh
```diff
@@ -78,21 +78,55 @@ upload_pipeline() {
...
+    buildkite-agent pipeline upload .buildkite/nightly_verify.yml
```
mrjunwan-lang, line 85: the nightly_verify should be after the nightly_release?

weiyu0824: I renamed main.yml to nightly_verify.yml. I think main.yml is very confusing

.buildkite/scripts/bootstrap.sh
weiyu0824: renamed main.yml to nightly_verify.yml

.buildkite/scripts/bootstrap.sh
weiyu0824: Upload are inserted in reverse order, so this seems correct to me.

jcyang43: Yea, the nightly release is responsible for the nightly docker image upload and does not depend on our nightly tests. 

For the main.yml naming, the plan was to combine multiple ymls into one, but I think having multiple ymls will be easier to maintain going forward given the complexity of this project.

jcyang43: I agree that we should rename main.yml to a more descriptive name. Personally I'd recommend renaming it to supoprt_matrices.yml as its main job is generating the support matrices, but nightly_verify.yml is fine too

weiyu0824: @jcyang43  why we didn't enforce nightly_releases to depend on nightly tests?

jcyang43: Because the requirement is that we'll release nightly docker images even if not all tests pass

PR #1385, opened 2026-01-01
Apply temperature scaling before top-p

oliverdutton:
Move temperature scaling to before top-p.

vLLM implements logit temperature scaling **before** top-p’ing. tpu-inference has logit temperature scaling applied **after** top-p’ing.  As temperature scaling affects the top-p this is a mismatch to be fixed.

See [here](https://github.com/vllm-project/vllm/blob/4bf6c2366818a1eeae257e06ec337039e6895f13/vllm/v1/sample/sampler.py#L177) and [here](https://github.com/vllm-project/vllm/blob/4bf6c2366818a1eeae257e06ec337039e6895f13/vllm/v1/sample/tpu/sampler.py#L63) in vLLM

(Duplicate of #1327 but with sign off on commit, apologies but I’m travelling and so only have access to a phone, git UI is a bit awkward on iOS so I just made a new branch, new commit and new PR rather in the UI rather than force push after editing the other PR)

PR #1391, opened 2026-01-05
Pallas Sampling Path

oliverdutton:
# Description
Make the top-k top-p sample faster. High capability models are getting more and more efficient and smaller (better training, RL, quantization) and the hardware is improving at matmul throughput much more quickly than memory bandwidth or vector throughput. A v7x paired with MoE of just 10B params activated like MiniMax-M2.1 could spend a lot of time in sampling, imagine if everything was w4a8 or even more quantized and even fewer params activated in frontier open/closed models in a few months time. Already a bf16 Gemma3 4B on v6e with 1024 token prefill can spend up to 5% runtime on sampling (there’s no such thing as a standard size for benchmarking but you can see where the trends going).
 
There’s room for improvement as most open and closed source frontier models recommend top-k of only 20-64 in sampling and users rarely(?) alter those settings (The Gemini 3 Pro is simply fixed at 64, Qwen3 and MiniMax-M2 models are k=20/40 recommended) so by specialising on these low-k regimes we can greatly speed up sampling from the current vLLM-TPU implementation.

# Implementation
A purely opt-in fast kernel which only compiles and runs if PALLAS_SAMPLING_TOPK_THRESHOLD is set. It does not appear to add any overheads and still supports users who make requests for a topk above the threshold.

Control for fast path is done via the env variable PALLAS_SAMPLING_TOPK_THRESHOLD.  tpu_sampling_metadata has been altered to add use_pallas_kernel which is a static bool metadata computed at creation of tpu_sampling_metadata on CPU, similarly to do_sampling. The sample function now has a path which uses that tpu_sampling_metadata.use_pallas_kernel (which is static metadata so determined at compile time) to use the fast path. This should mean near zero overhead from having the fast path, and if some batches have top-k requests higher than the threshold the current path is maintained.

If the env var isn't set it defaults to 0 and no extra precompiling is done and a tiny check against that env var at metadata creation. If set, extra specializations of sample are compiled for the fast path.

Note: this also implements #1385 which fixes the ordering of top-p and temperature scaling of logits.

# Sharding
The code path is 15-75x faster than the current vLLM sample function on a single v5e. It may be faster still on multi-TPU as it is sharding friendly, requiring a single all-gather of two (batch, k) arrays along the vocab dimension, avoids any resharding involving the full vocab size. The current vLLM sample function is not sharding friendly, requiring 66 reductions on vocab dimension and so a reshard replicating vocab dim is required.

The sharding path is **untested** (author is working on just a phone and setting up multi-TPU testing is hence quite hard and interpret=True compilation hangs indefinitely so can’t be tested on CPU).

# Comparison to current sampling
The pallas fast path gives **exact match** of sampled tokens for the same rng key, as long as top-k is uniquely defined. The current vLLM implementation inclaudes all tied values at the k'th largest value  boundary while this implementation uses an arbitrary subset such that only k values are selected (similar to jax.lax.top_k). Tests are included for this, as well as tests for all the subfunctions (bitonic top-k, the faster divide and filter top-k built on top of the bitonic version, cumsum, take_along_axis, jax.random.categorical with sparse indices, top_p). These tests have all been run on a single v5e and pass.

# Tests

`pytest tests/kernels/sampling` and `pytest tests/layers/jax/sample`
on single v5e

Done in colab with v5e:
```
!pip install -q vllm-tpu==0.13.3
branch = 'pallas_sampling'
!git clone -q -b {branch} --single-branch https://github.com/oliverdutton/tpu-inference.git && cd tpu-inference && pip install -q .
!cd tpu-inference && pytest -v -s tests/kernels/sampling
!cd tpu-inference && pytest -v -s tests/layers/jax/sample
```

verified on v5e and v6e that it compiles and runs, only compiling if the env var is set and supports requests with k>threshold by falling back to the current sample implementation. This also verified runtimes in the xprof that correct functions are called and the speedups in jit_sample.
```
!pip install -q -U vllm-tpu==0.13.3
branch = 'pallas_sampling'
!git clone -q -b {branch} --single-branch https://github.com/oliverdutton/tpu-inference.git && cd tpu-inference && pip install -U -q .
for k, threshold in [
  #WAI
  (64, 64),
  # Check compiles, but not used
  (65,64),
  # leave env unset if 0, which should be 0 to check that works and doesnt compile pallas kernels
  (64, 0)]:
  env = f'export PALLAS_SAMPLING_TOPK_THRESHOLD={threshold}' if threshold!=0 else ''
  cmd = f'''
export HF_TOKEN=your_hf_token
{env}
python tpu-inference/examples/tpu_profiling.py \
  --model google/gemma-3-4b-it \
  --input-len 1024 \
  --output-len 5 \
  --batch-size 128 \
  --profile-result-dir=/content/profiles/ \
  --top-k {k} \
  --top-p 0.95
  '''
  !{cmd}
  print(f'{k=} {threshold=} profiled')

```

Again:
The sharding path is **untested** (author is working on just a phone and setting up multi-TPU testing is hence quite hard and interpret=True compilation hangs indefinitely so can’t be tested on CPU). Help would be greatly appreciated.

# Checklist

Yes
Yes
Yes, but unsure about how/where to advertise this env variable to users

### DIFF
--- a/tpu_inference/kernels/sampling/sampling.py
+++ b/tpu_inference/kernels/sampling/sampling.py
@@ -93,7 +93,7 @@ def topk_topp_and_sample_shmap(
   Returns:
     Sampled token indices.
   """
-  batch_axis, vocab_axis = (ShardingAxisName.MLP_DATA, ShardingAxisName.VOCAB)
+  batch_axis, vocab_axis = (ShardingAxisName.MLP_DATA, ShardingAxisName.MLP_TENSOR)
   def shmap_topk(logits, k):
     topk_logits, topk_idxs = top_bounded_k(
       logits,


kyuyeunk:
Hi @oliverdutton this is a really (I mean really really) large pr. And it takes a significant bandwidth from reviewers to go though this pr. Can you please create rfc first that goes through each design decisions in detail and gather feedback first?

oliverdutton:
I've tried to make an rfc based on the vLLM repo template. Any feedback on bits missing welcome. I've focused on how I propose this standalone kernels/sampling folder integrates into tpu-inference and how the tests can show exact matches for sampling and so the kernel can be trusted (except I haven't yet got the multi-TPU setup to test the sharding rules as I'm only on a phone, if in principle the code was accepted I can invest in properly testing multi-TPU).

Anyone else I should cc in the rfc?

PR #1395, opened 2026-01-05
Override kv cache page size for padding

Lumosis:
# Description

Override kv cache's padded page size directly.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,6 +32,7 @@ RUN apt-get update && apt-get install -y \
     procps \
     curl \
     rclone \
+    wget \
     ca-certificates \
     && rm -rf /var/lib/apt/lists/*
 


tpu_inference/worker/tpu_worker.py
```diff
@@ -409,20 +406,11 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
-            group_size = max(
-                len(group.layer_names) for group in kv_cache_groups)
-            available_memory = self.determine_available_memory()
-            num_blocks = get_num_blocks(self.vllm_config, group_size,
-                                        available_memory,
-                                        attention_page_size_bytes)
-            cache_config = self.vllm_config.cache_config
-            cache_config.num_gpu_blocks_override = num_blocks
-
+                f"({attention_page_size_bytes} Bytes). Overridding the page size to "
```
kyuyeunk: reason why the page overriding logic was here was because it needed to access `self.determine_available_memory()`. but I wasn't really fond of having this logic in `tpu_worker.py` since i believe all kv cache related logic should be consolidated to the same place. given that we can now override page size, this logic can be placed inside https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/runner/kv_cache_manager.py right?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -177,6 +180,24 @@ def get_kv_cache_spec(self):
...
                         f"Unknown attention type: {attn_module.attn_type}")
+
+        if len(kv_cache_spec) == 0:
+            return kv_cache_spec
+
+        vllm_page_size_bytes = get_uniform_page_size(
+            list(kv_cache_spec.values()))
+        attention_page_size_bytes = get_attention_page_size_bytes(
+            self.runner.mesh, kv_cache_spec)
+        if vllm_page_size_bytes != attention_page_size_bytes:
```
kyuyeunk: Instead of overriding, isn't possible to just pass kv cache page size during initialization of attention spec?

Lumosis:
> i think it looks good, but i do notice a lot of duplicate code with only minor changes to argument for calling `get_attention_page_size_bytes`. is there a way to remove some of the duplicate code?
> 
> but if you absolutely cannot think of a way, i can give approval as-is.

Done

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -55,6 +56,29 @@ def __init__(self, runner: "TPUModelRunner"):
         self.shared_kv_cache_layers: dict[str, str] = {}
         self.use_mla = self.runner.model_config.use_mla
 
+    def _create_attention_spec(self, block_size: int, num_kv_heads: int,
```
kyuyeunk: This is much better! Thanks @Lumosis! But it seems like sliding window attention is handled separately? is there a rationale for that?

Lumosis: Done. I thought sliding window is a special case and need to be handled differently but it seems it can fit in the helper function.

vipannalla (approved):
Looks good!

PR #1412, opened 2026-01-07
[CI] Change base image to python:3.12-slim-bookworm

ernie-chang:
# Description

Change Dockerfile.pypi arg base image to python:3.12-slim-bookworm

# Checklist

### DIFF
--- a/tests/runner/test_kv_cache_manager.py
+++ b/tests/runner/test_kv_cache_manager.py
@@ -31,6 +31,7 @@ from vllm.v1.request import Request
 
 from tpu_inference import utils as common_utils
 from tpu_inference.runner.input_batch import CachedRequestState
+from tpu_inference.runner.kv_cache import get_attention_page_size_bytes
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.pypi
jcyang43: This shouldn't be needed

docker/Dockerfile.pypi
```diff
@@ -1,5 +1,5 @@
-ARG NIGHTLY_DATE="20250714"
-ARG BASE_IMAGE="us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.12_tpuvm_$NIGHTLY_DATE"
+
```
jcyang43, line 12: Please also leave a comment on the top of the file explaining the purpose of this dockerfile. Currently, it may not be immediately clear to users which dockerfile they need to use (`docker/Dockerfile` vs `docker/Dockerfile.pypi`). Thanks

PR #1419, opened 2026-01-08
Use local devices when computing hbm usage stats

richardsliu:
# Description

This fixes an issue with running multihost TPUs on Ironwood (v7x-16), where the computed HBM appears to be 0:
```
ValueError: total_hbm_used_gb=0.0GiB exceeds total_hbm_limit_cap_gb=0.0GiB by -0.0GiB. Please consider increasing --gpu-memory-utilization from 0.9 to a larger value.
```

The reasons for this fix are as follows:

1. JAX can only successfully run `memory_stats()` on the local cores (8 cores on each host VM). Attempting to run `memory_stats()` on non-local cores in Ironwood results in a value of 0, causing the error.

2. This only happens on multi-host because on a single-host TPU, all the devices are by definition local.

3. This only happens on Ironwood, because in previous TPU generations, calling `memory_stats()` on a remote TPU host returns cached metadata instead of real-time data, thus the value returned is non-0.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/worker/tpu_worker.py
+++ b/tpu_inference/worker/tpu_worker.py
@@ -266,7 +266,7 @@ class TPUWorker:
 
     def determine_available_memory(self) -> int:
         gpu_memory_utilization = self.cache_config.gpu_memory_utilization
-        hbm_usage = utils.hbm_usage_bytes(self.devices)
+        hbm_usage = utils.hbm_usage_bytes(jax.local_devices())
         total_hbm_limit = total_hbm_used = 0
         for used, limit in hbm_usage:
             total_hbm_used += used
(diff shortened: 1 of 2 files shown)


mrjunwan-lang (approved):

mrjunwan-lang:
Thanks for the fix!

tpu_inference/worker/tpu_worker.py
```diff
@@ -266,7 +266,7 @@ def initialize_pp_transfer_connect(self):
 
     def determine_available_memory(self) -> int:
         gpu_memory_utilization = self.cache_config.gpu_memory_utilization
-        hbm_usage = utils.hbm_usage_bytes(self.devices)
+        hbm_usage = utils.hbm_usage_bytes(jax.local_devices())
```
kyuyeunk, line 269: this code triggers error if tp size is smaller than number of devices. I will revert this change for now.

PR #1423, opened 2026-01-08
Address compatibility issues arising from the removal of the XLA dependency 

weiyu0824:
# Description
This PR addresses compatibility issues arising from the removal of the XLA dependency in upstream vLLM ([reference]( https://github.com/vllm-project/vllm/pull/30808).

Changes: 
- Moved PallasAttention static function from vllm to tpu-inference. 
- Update TpuPlatform in tpu_platform.py to use local PallasAttentionBackend. 
- Add create_weights in compressed_tensors_w8a8_int8.py because vllm's implementation would introduce another error. 
- Register PallasAttention using decorator (see line:32 and line:37 in tpu_inference/layers/vllm/attention.py) because we already remove Pallas enum in upstream.

# Tests
- [buildkite result](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7876/summary?sid=019b9d30-3db1-4b38-86a5-21bf6d64de63)

# Checklist

### DIFF
--- a/tests/e2e/test_speculative_decoding.py
+++ b/tests/e2e/test_speculative_decoding.py
@@ -207,7 +207,7 @@ def _test_performance_helper(
         del ref_llm
 
         # Waiting for TPUs to be released
-        time.sleep(10)
+        time.sleep(30)
 
         # Test speculative LLM timing with max_num_seqs=1
         spec_llm = LLM(model=model_name,
@@ -223,7 +223,7 @@ def _test_performance_helper(
 
         del spec_llm
         # Waiting for TPUs to be released
-        time.sleep(10)
+        time.sleep(30)
 
         speedup = ref_time / spec_time
         print(f"Reference LLM time: {ref_time:.2f}s")


tpu_inference/layers/vllm/attention.py
```diff
@@ -22,17 +26,79 @@
...
+# TPU requires the head size to be a multiple of 128.
+TPU_HEAD_SIZE_ALIGNMENT = 128
 
+@register_backend(AttentionBackendEnum.CUSTOM)
 class PallasAttentionBackend(AttentionBackend):
 
     @staticmethod
     def get_name() -> str:
-        return "PALLAS"
+        return "CUSTOM"
```
QiliangCui: why can't use "PALLAS"?

QiliangCui: I see. the "PALLAS" is registered as `PALLAS = "vllm.v1.attention.backends.pallas.PallasAttentionBackend"` in vllm.main.

If we use CUSTOM, in the future, we should consider remove the register in vllm.main.

tpu_inference/platforms/tpu_platform.py
```diff
@@ -54,10 +54,6 @@ class TpuPlatform(Platform):
     def get_attn_backend_cls(cls, selected_backend: "AttentionBackendEnum",
                              attn_selector_config: "AttentionSelectorConfig",
                              **kwargs) -> str:
-        from vllm.attention.backends.registry import AttentionBackendEnum
-
-        if selected_backend != AttentionBackendEnum.PALLAS:
-            logger.info("Cannot use %s backend on TPU.", selected_backend)
 
         logger.info("Using Pallas V1 backend.")
```
QiliangCui: to test the change, you need to set `VLLM_COMMIT_HASH=""` in here: 

https://github.com/vllm-project/tpu-inference/blob/main/docker/Dockerfile#L18

otherwise, you change here is not tested.

tpu_inference/layers/vllm/attention.py
QiliangCui: hm.. here, i think we can call it "PALLAS", can we? i mean it is CUSTOM, yes, but the name is "PALLAS".

tpu_inference/platforms/tpu_platform.py
```diff
@@ -54,10 +54,6 @@ class TpuPlatform(Platform):
     def get_attn_backend_cls(cls, selected_backend: "AttentionBackendEnum",
                              attn_selector_config: "AttentionSelectorConfig",
                              **kwargs) -> str:
-        from vllm.attention.backends.registry import AttentionBackendEnum
-
-        if selected_backend != AttentionBackendEnum.PALLAS:
```
QiliangCui, line 59: if selected_backend != AttentionBackendEnum.CUSTOM: throw.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
```diff
@@ -50,6 +57,60 @@ def __init__(self, strategy: str, is_static_input_scheme: bool,
         self.linear_config = linear_config
         self.is_channelwise = (self.strategy == QuantizationStrategy.CHANNEL)
 
+    def create_weights(
```
QiliangCui, line 58: why is this necessary?

tpu_inference/platforms/tpu_platform.py
weiyu0824: New buildkite run: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7880/summary

tpu_inference/layers/vllm/attention.py
weiyu0824: We already removed the "PALLAS" enum in attention backend ([reference](https://github.com/vllm-project/vllm/blob/83e1c76dbe07e30b7f4e6dbe17ba580f4afc98f0/vllm/attention/backends/registry.py#L34)), to register a new enum, we have to overwrite the "CUSTOM" enum. I follow the description in [here](https://github.com/vllm-project/vllm/blob/83e1c76dbe07e30b7f4e6dbe17ba580f4afc98f0/vllm/attention/backends/registry.py#L232-L234)

docker/Dockerfile
```diff
@@ -15,7 +15,8 @@
 ARG BASE_IMAGE="python:3.12-slim-bookworm"
 # The latest main will be used if arg unspecified
 # TODO (ranlihao): unpin after the upstream is stable.
-ARG VLLM_COMMIT_HASH="0dd5dee9b9bc88453f5f3eacfde751e6b9ba4871"
+# ARG VLLM_COMMIT_HASH="0dd5dee9b9bc88453f5f3eacfde751e6b9ba4871"
```
vanbasten23: you can remove the change and comments from https://github.com/vllm-project/tpu-inference/pull/1416

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
kyuyeunk: I've had a discussion with @weiyu0824, and for most classes in vllm, `create_weights` is platform agnostic and it's responsible for purely creating & loading weight. But we found out `CompressedTensorsW8A8Int8` has some logic that checks which platform it is running on. And it fails if the current platform is not on the list of expected platforms. This was okay when xla version existed because the tpu was on the list of expected platforms, but when xla version is removed, this function triggers error. So that why in here, we are overriding the `create_weights` function.

reference: the dictionary with list of expected platform where tpu gets removed after xla deletion: https://github.com/vllm-project/vllm/pull/30808/changes#diff-862feda1d21fda8238cdaf974043cc36201ed24d73b963efb27aca11c50c1b9c

@weiyu0824, even though this approach might work, is it possible to modify the code so that it reuses base class's logic as much as possible? I want to avoid duplicate code if we can.

kyuyeunk: Alternatively, do you think it's possible to remove the platform check in the vLLM upstream? Because as I mentioned, `create_weight`function should be platform agnostic, and only during  `process_weights_after_loading`we do platform specific weight modification.

vanbasten23: Yeah it's a valid point and thanks for the explanation.

Actually, if it's the only remaining comment, can we get this pr in and create a follow-up pr to fix this comment? The longer we pin a vllm commit, the more likely we miss vllm breaking change and the harder to fix the breaking change. @weiyu0824 @kyuyeunk  Wdyt?

kyuyeunk: sgtm. but @weiyu0824, please make sure to work on a follow-up pr immediately. I want to eliminate duplicate code as much as possible for better maintainability.

docker/Dockerfile
weiyu0824: OK

tpu_inference/platforms/tpu_platform.py
weiyu0824: Add back this 3 lines

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_int8.py
weiyu0824: Ok

kyuyeunk:
seems like ci is failing due to upstream change: https://github.com/vllm-project/vllm/pull/30519

I think this should be a simple fix. @weiyu0824 can you add a fix commit to this branch as well?

tpu_inference/layers/vllm/attention.py
kyuyeunk: I don't think there's anything stopping us from using `get_name` that isn't in `AttetnionBackendEnum`. There's already bunch of attentions in vllm that doesn't match values within `AttentionBackendEnum`. For instance, this class returns [AITER_TRITON_MLA](https://github.com/vllm-project/vllm/blob/e2d49ec2a4818186b8bafda43f97c298ff91a35e/vllm/v1/attention/backends/mla/aiter_triton_mla.py#L8-L9) but it's not in any of the `AttentionBackendEnum` list: https://github.com/vllm-project/vllm/blob/e2d49ec2a4818186b8bafda43f97c298ff91a35e/vllm/attention/backends/registry.py#L34-L83

tpu_inference/layers/vllm/attention.py
```diff
@@ -22,17 +26,79 @@
 
 logger = init_logger(__name__)
 
+# TPU requires the head size to be a multiple of 128.
+TPU_HEAD_SIZE_ALIGNMENT = 128
 
+
+@register_backend(AttentionBackendEnum.CUSTOM)
```
kyuyeunk: Is it possible to use `AttentionBackendEnum.FLASH_ATTN`? Because this pallas attention is still an implementation of flash attention. And it's not some completely new algorithm.

weiyu0824: I think we can override FLASH_ATTN but as I mentioned [here](https://github.com/vllm-project/vllm/blob/83e1c76dbe07e30b7f4e6dbe17ba580f4afc98f0/vllm/attention/backends/registry.py#L232-L234), I think CUSTOM is designed to be overridden. I could be wrong though.

kyuyeunk: my gut feeling is
- we have invented completely new attention algorithm: use custom
- we have re-implemented existing attention algorithm: use any of existing values in backend enum.

also, we'll likely want to support mla as well, which means, we cannot have two different attention implementations in a single `AttetnionBackendEnum.CUSTOM`. Instead, regular flash attetntion should be registered with `AttentionBackendEnum.FLASH_ATTN` while mla should be registered using `AttentionBackendendEnum.FLASH_ATTN_MLA`.

tpu_inference/layers/vllm/attention.py
weiyu0824: Do you suggest to override CUSTOM enum, but still using PALLAS as the name?

kyuyeunk: Yes, I think that's better.

tpu_inference/layers/vllm/attention.py
weiyu0824: good suggestion. I added another commit for this. 
https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7929/summary

tpu_inference/platforms/tpu_platform.py
```diff
@@ -55,10 +55,8 @@ def get_attn_backend_cls(cls, selected_backend: "AttentionBackendEnum",
                              attn_selector_config: "AttentionSelectorConfig",
                              **kwargs) -> str:
         from vllm.attention.backends.registry import AttentionBackendEnum
-
-        if selected_backend != AttentionBackendEnum.PALLAS:
+        if selected_backend != AttentionBackendEnum.FLASH_ATTN:
             logger.info("Cannot use %s backend on TPU.", selected_backend)
-
         logger.info("Using Pallas V1 backend.")
         return "tpu_inference.layers.vllm.attention.PallasAttentionBackend"
```
kyuyeunk: Probably not a good idea to hard code a string based path. Since we have registered the backend, is it possible to replace this with `get_path` function? : https://github.com/vllm-project/vllm/blob/7a05d2dc65065da3d6fd54c67a670ab08ffaf600/vllm/attention/backends/registry.py#L85

tpu_inference/layers/vllm/attention.py
kyuyeunk: ah okay it seems like the ci fails when the name doesn't match. can you replace `PALLAS_FLASH_ATTN` with `FLASH_ATTN`?

weiyu0824: Because of this [line](https://github.com/vllm-project/vllm/blob/707b240d7e7523c652b8a5b075ffc1ee9e1120ae/vllm/attention/layer.py#L279), we have to use the name defined in AttentionEnum. This is the corresponding [error](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/7929/summary?sid=019ba11c-065d-45af-b32c-695584eaf8a6)

tpu_inference/platforms/tpu_platform.py
weiyu0824: Not sure what do you mean

kyuyeunk: instead of returning a string, is it possible to return `selected_backend.get_path()` like you see in here? https://github.com/vllm-project/vllm/blob/707b240d7e7523c652b8a5b075ffc1ee9e1120ae/vllm/platforms/cuda.py#L316

kyuyeunk:
hmmm we have another error due to upstream change 1 hour ago: https://github.com/vllm-project/vllm/pull/32003

yeah, feel free to ignore all my other comments and let's just try to get this pr merged asap.

PR #1429, opened 2026-01-08
For GQA, replicate kv head when num_kv_heads < tp_size

Lumosis:
# Description
An example of the computation process (tp_size=16, kv_heads=8, q_heads=128):

16 Q heads attend to 1 kv head.
After sharding, the 16 Q heads are replicated on 2 chips. This PR will also replicate the kv head so that the kv head will be on the 2 chips as well.

The general data layout would be like:

Device ID | Logical Index | Actual Data Received
-- | -- | --
Chip 0 | 0 | H0
Chip 1 | 1 | H0 (Copy)
Chip 2 | 2 | H1
Chip 3 | 3 | H1 (Copy)
... | ... | ...
Chip 14 | 14 | H7
Chip 15 | 15 | H7 (Copy)

By replicating H0 onto both Chip 0 and Chip 1, we ensure that every chip has the local KV data required to calculate attention for the specific Query heads assigned to it.

# Tests

commands to reproduce.

# Checklist


tpu_inference/layers/common/attention_interface.py
```diff
@@ -302,6 +302,15 @@ def sharded_ragged_paged_attention(
...
 ):
     """Shards along KV heads."""
+    # Handle GQA/MQA where num_kv_heads < tp_size
+    # We replicate KV heads to match tp_size so that we can shard them evenly.
+    if ShardingAxisName.ATTN_HEAD in mesh.shape:
+        tp_size = mesh.shape[ShardingAxisName.ATTN_HEAD]
+        num_kv_heads = k.shape[1]
+        if num_kv_heads < tp_size and tp_size % num_kv_heads == 0:
+            factor = tp_size // num_kv_heads
+            k = jnp.repeat(k, factor, axis=1)
```
kyuyeunk, line 317: this is executed during inference time. is it a no-op? or does it introduce a performance overhead? if it does introduce performance overhead, would we be able to achieve same outcome by changing how qkv weights are laid out during weight loading process?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -302,6 +302,15 @@ def sharded_ragged_paged_attention(
...
+            v = jnp.repeat(v, factor, axis=1)
```
vipannalla, line 318: Can up throw exception here if tp is not a multiple of kv_heads?

tpu_inference/layers/common/attention_interface.py
Lumosis: For non-GQA/MQA case, this is no-op. For GQA/MQA case, I think the overhead is negligible. It's ideal to move it to weight loading but I need to take a closer look. I'll leave a TODO here for now.

kyuyeunk: can you share a trace file so i can verify that it's indeed a no-op & how much overhead is added for gqa/mqa?

vipannalla (commented):
Also, can you add a unitest in https://github.com/vllm-project/tpu-inference/blob/a282fade963da76fd342c7ea4f6f92129d38389b/tests/layers/common/test_attention_interface.py?

tpu_inference/layers/common/attention_interface.py
Lumosis: Done

Lumosis:
> Also, can you add a unitest in https://github.com/vllm-project/tpu-inference/blob/a282fade963da76fd342c7ea4f6f92129d38389b/tests/layers/common/test_attention_interface.py?

Done

kyuyeunk (approved):

PR #1434, opened 2026-01-09
[DO NOT REVIEW YET] Move weight to TPU and then re-quantize.

vanbasten23:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/fp8.py
+++ b/tpu_inference/layers/vllm/quantization/fp8.py
@@ -55,7 +55,8 @@ from tpu_inference.layers.vllm.quantization.configs import (
 from tpu_inference.layers.vllm.quantization.unquantized import (
     VllmUnquantizedFusedMoEMethod, VllmUnquantizedLinearMethod)
 from tpu_inference.logger import init_logger
-from tpu_inference.utils import get_mesh_shape_product
+from tpu_inference.utils import (device_array, get_mesh_shape_product,
+                                 time_function)
 
 P = PartitionSpec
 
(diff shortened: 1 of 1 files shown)


vanbasten23:
Hi @kyuyeunk , when you said "the weight loading time" for qwen3 coder is very slow. I guess you are referring to this function https://github.com/vllm-project/tpu-inference/blob/67b208209b231000e0a9343e9cf42ad3499161e4/tpu_inference/layers/vllm/quantization/fp8.py#L110 (for one layer) or https://github.com/vllm-project/vllm/blob/aaf4b70aae5976115eb5f8caead3d4eeb3657136/vllm/model_executor/model_loader/base_loader.py#L59 (for all layers).

This PR has the change that move the weights to TPU, then re-quantize. Before the change, [process_weights_after_loading](https://github.com/vllm-project/vllm/blob/aaf4b70aae5976115eb5f8caead3d4eeb3657136/vllm/model_executor/model_loader/base_loader.py#L59) takes 263s (which looks not super bad lol). With the change, it takes 327s. I think it's because the host to device overhead. (fwiw, I used this cmd `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1 MODEL_IMPL_TYPE=vllm python examples/offline_inference.py --model Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8 --tensor-parallel-size 8 --enable-expert-parallel`)

kyuyeunk:
yes, the function where the current change happened is the one i was thinking of. 

Instead of using `device_array`, what happens if you just specify input sharding & output sharding on the argument of `@jax.jit` in here? https://github.com/vllm-project/tpu-inference/blob/475a3352e4d24002771cabe136663a5cb970e1a5/tpu_inference/layers/vllm/quantization/fp8.py#L277

Also, have you averaged out the number reported by `@time_function`? I'm wondering if tpu version just has a longer compilation time but after it is compiled, it is faster than cpu version.

vanbasten23:
> what happens if you just specify input sharding & output sharding on the argument of `@jax.jit` in here?

How is it different from the current approach (device_array which internally uses jax.device_put)?

> Also, have you averaged out the number reported by @time_function?

I measure [this function](https://github.com/vllm-project/vllm/blob/aaf4b70aae5976115eb5f8caead3d4eeb3657136/vllm/model_executor/model_loader/base_loader.py#L59) which only reports the aggregated time loading all layers.

PR #1437, opened 2026-01-10
[Misc] Simplify block quantization logic

kyuyeunk:
# Description

Remove padding option for quantization and perform padding outside of quantization logic.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/quantization.py
+++ b/tpu_inference/layers/common/quantization.py
@@ -170,7 +170,6 @@ def quantize_tensor(
     tensor: jax.Array,
     axis: int | tuple | None = -1,
     block_size: int | None = None,
-    pad_tensor: bool = False,
 ) -> tuple[jax.Array, jax.Array]:
     """Quantize tensor.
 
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/common/quantization.py
```diff
@@ -239,16 +226,15 @@ def quantize_tensor(
 
     abs_max = jnp.max(jnp.abs(tensor), axis=axis, keepdims=True)
     scale = abs_max / dtype_max
+
+    # If scale=0, scale_inv=1/scale=1/0=NaN. Since NaN will cause numeric error
```
vanbasten23, line 230: thanks for the comment!

tpu_inference/layers/common/quantization.py
```diff
@@ -191,35 +189,24 @@ def quantize_tensor(
         axis = [axis]
```
vanbasten23, line 190: Do you need to check the precondition that the padding has been done before this function is called?

kyuyeunk: We already have a check for htat. in line 200, we have a raise `ValueError`. so if a dim size doesn't properly align to block size - which potentially mean that a tensor wasn't properly padded.

PR #1438, opened 2026-01-10
[Torchax][MoE] Always transpose moe weights

kyuyeunk:
# Description

Transpose MoE weights during weight loading time for better performance 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
+++ b/tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
@@ -156,14 +156,28 @@ def process_moe_weights(
             w3_bias = w13_bias[:, 1::2]
             w13_bias = jnp.concat([w1_bias, w3_bias], axis=1)
 
+    # Transpose non-constracting dim to right most dim
+    w13_weight = jnp.swapaxes(w13_weight, 1, 2)
+    w2_weight = jnp.swapaxes(w2_weight, 1, 2)
+
+    # Workaround for JAX error "must have valid byte strides"
+    w13_weight = with_layout_constraint(w13_weight, Layout((0, 1, 2)))
+    w2_weight = with_layout_constraint(w2_weight, Layout((0, 1, 2)))
+
     if w13_weight_scale is not None:
         w13_weight_scale = w13_weight_scale.astype(jnp.float32)
+        w13_weight_scale = jnp.swapaxes(w13_weight_scale, 1, 2)
+        w13_weight_scale = jnp.expand_dims(w13_weight_scale, 2)
     if w2_weight_scale is not None:
         w2_weight_scale = w2_weight_scale.astype(jnp.float32)
+        w2_weight_scale = jnp.swapaxes(w2_weight_scale, 1, 2)
+        w2_weight_scale = jnp.expand_dims(w2_weight_scale, 2)
     if w13_bias is not None:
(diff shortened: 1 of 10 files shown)


vanbasten23:
One general question, for this change "Always transpose moe weights", should it only apply to the gmm kernel instead of the fused moe kernel?

tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
```diff
@@ -156,14 +156,28 @@ def process_moe_weights(
             w3_bias = w13_bias[:, 1::2]
             w13_bias = jnp.concat([w1_bias, w3_bias], axis=1)
 
+    # Transpose non-constracting dim to right most dim
+    w13_weight = jnp.swapaxes(w13_weight, 1, 2)
```
vanbasten23, line 160: this will affect the fused moe kernel. Do we need to do this for the fused moe kernel as well?

tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
```diff
@@ -209,17 +217,14 @@ def process_moe_weights(
 
             if w13_weight_scale is not None:
                 w13_weight_scale = w13_weight_scale.reshape(
-                    num_experts, 2, intermediate_size, 1, -1)
-                w13_weight_scale = jnp.swapaxes(w13_weight_scale, 2, 4)
+                    num_experts, -1, 2, 1, intermediate_size)
+                w13_weight_scale = jnp.swapaxes(w13_weight_scale, 1, 2)
```
vanbasten23, line 221: I wonder if there is a test exercising this branch.

tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
kyuyeunk: no, fusedmoe kernel already transposes the weight. the codebase has actually be simplified because previously, we  had a branching logic where fusedmoe kernel path transposed weight while non-fusedmoe kernel didn't transposed it. but now, it will always be transposed regardless of moe backends.

tpu_inference/layers/vllm/process_weights/fused_moe_weights.py
kyuyeunk: all the moe related unit tests like these ones all trigger that code
- https://github.com/vllm-project/tpu-inference/blob/7e0cbe9c5ffac8db3bed29f643162291e630a0f4/tests/layers/vllm/test_unquantized.py#L522-L524
- https://github.com/vllm-project/tpu-inference/blob/7e0cbe9c5ffac8db3bed29f643162291e630a0f4/tests/layers/vllm/test_mxfp4.py#L205-L207
- https://github.com/vllm-project/tpu-inference/blob/7e0cbe9c5ffac8db3bed29f643162291e630a0f4/tests/layers/vllm/test_fp8.py#L469-L471

kyuyeunk:
> One general question, for this change "Always transpose moe weights", should it only apply to the gmm kernel instead of the fused moe kernel?

No, as i mentioned in the other comment, fusedmoe kernel was already transposing and we had a branching logic to only transpose it for fusemoe kernel and don't transpose it for other code path. but now, the logic has been simplified so that all code paths will transpose the weights.

Issue #1121, opened 2025-11-18
[Feature]: Support "fp8" Quantization Method for Torchax


xingliu14:
### 🚀 The feature, motivation and pitch

For an example fp8 quantized model [Qwen3-32B-FP8](https://huggingface.co/Qwen/Qwen3-32B-FP8), its quantization config is:
```
"quantization_config": {
  "activation_scheme": "dynamic",
  "fmt": "e4m3",
  "quant_method": "fp8",
  "weight_block_size": [
    128,
    128
  ]
}
```
The config includes:
- activation_scheme (dynamic or static activation quantization)
- fmt (FP8 format - e4m3 or e5m2)
- quant_method ("fp8")
- optionally weight_block_size for block-wise quantization (e.g., [128, 128] creates 128x128 quantization blocks).

Add support for `fp8` model quantization, consistent with [vLLM's implementation](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/quantization/fp8.py).

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

kyuyeunk:
Additional context:

- A similar PR that adds "awq" quantization method: https://github.com/vllm-project/tpu-inference/pull/618
- Necessary steps
  -  Add "fp8" to the list of quantization method mapping: https://github.com/vllm-project/tpu-inference/blob/45edde6c0f79da1455a1418337af820cfe531fab/tpu_inference/layers/vllm/quantization/__init__.py#L22-L27
  - Similar to AWQ, create a child class that inherits from vllm class including
    - config class: https://github.com/vllm-project/tpu-inference/blob/45edde6c0f79da1455a1418337af820cfe531fab/tpu_inference/layers/vllm/quantization/awq.py#L34-L35
    - linear method class: https://github.com/vllm-project/tpu-inference/blob/45edde6c0f79da1455a1418337af820cfe531fab/tpu_inference/layers/vllm/quantization/awq.py#L60-L61
    - Note initial version does not have to support more complex layers like MoE layers and throw `NotImplementedError` for now: https://github.com/vllm-project/tpu-inference/blob/45edde6c0f79da1455a1418337af820cfe531fab/tpu_inference/layers/vllm/quantization/awq.py#L54-L56
[...]

PR #1440, opened 2026-01-10
Handle unaligned FP8 block quantization edge case

catswe:
# Description

Handled edge case in linked issue https://github.com/vllm-project/tpu-inference/issues/1121#issuecomment-3706449354.

# Tests

Verified correctness manually by testing against `gaunernst/DeepSeek-V2-Lite-Chat-FP8`, a test model for `DeepSeek-V3/R1` (getting this model to work required additional changes not included in this PR, see https://github.com/vllm-project/tpu-inference/pull/1593). Also verified that the existing code fails under the new test case, `test_unaligned_block_quantization`, while the current changes passes.

```
JAX_TRACEBACK_FILTERING=off VLLM_MLA_DISABLE=1 VLLM_LOGGING_LEVEL=DEBUG MODEL_IMPL_TYPE=vllm vllm serve gaunernst/DeepSeek-V2-Lite-Chat-FP8   --max-model-len=4096   --gpu-memory-utilization=0.98  --download-dir /dev/shm
```

Screenshot 2026-01-31 at 10 34 29 PM

# Checklist

### DIFF
--- a/tests/layers/vllm/test_fp8.py
+++ b/tests/layers/vllm/test_fp8.py
@@ -188,9 +188,6 @@ def initialize_layer_weights(layer: torch.nn.Module):
         torch.rand(layer.output_size, layer.input_size, dtype=torch.float32) /
         10)
 
-    for i in range(0, layer.output_size, block_m):
-        w_f32[i:i + block_m] *= 1 + i / 100  # simulate varying scales
-
     w_q, w_scale_blocks = ref_quantize_fp8_block_2d(w_f32, block_m, block_n,
                                                     torch.float8_e4m3fn)
 


tpu_inference/layers/common/quantization/__init__.py
```diff
@@ -165,6 +165,39 @@ def dequantize_tensor_from_mxfp4_packed(
     )
 
 
+def pad_and_dequantize_tensor(
```
kyuyeunk: awesome generic solution that is applicable for variety of situations!

tests/layers/vllm/test_fp8.py
```diff
@@ -535,3 +552,37 @@ def test_blockwise_quant(requant_block_size, requant_weight_dtype):
...
+            output_size=output_size,
+            bias=False,
+            params_dtype=dtype,
+            return_bias=False,
+            quant_config=quant_config,
+        )
+
+    initialize_layer_weights(linear_layer)
+    ref_output, layer_output = return_ref_and_layer_output(linear_layer)
+    torch.testing.assert_close(ref_output, layer_output)
```
kyuyeunk: add new line

tpu_inference/layers/common/quantization/__init__.py
kyuyeunk: how about we move this functionality into the main function  - `dequantize_tensor` - so that padding / slicing is applied automatically if alignment issue is detected? (i.e., dim size is not divisible by block size?) but let me know if it makes the main function too complex / spaghetti code.

kyuyeunk: please add a test for this function in https://github.com/vllm-project/tpu-inference/blob/main/tests/layers/common/test_quantization.py

catswe: Thank you!

tpu_inference/layers/common/quantization/__init__.py
```diff
@@ -106,18 +108,26 @@ def dequantize_tensor(
...
+
+    aligned_shape = tensor_q.shape
     if tensor_q.ndim == scale.ndim:
         # Indicates the tensor was block quantized.
-        blocked_shape = [[i] for i in orig_shape]
+        blocked_shape = [[i] for i in aligned_shape]
         for i in axis:
             num_blocks = scale.shape[i]
             if tensor_q.shape[i] % num_blocks:
                 raise ValueError(
```
kyuyeunk, line 125: just to double check: if `block_size` is not passed as an argument and the tensor shape does not align with `num_blocks` found in `scale.shape[i]`, it will throw an error right? is there an unit test that checks if error is thrown in this scenario?

catswe: There isn't any unit test for this. I commented out the `raise ValueError`, and all of them still passed, so I added another test. This path isn't actually very robust, though. If `tensor_q` has shape `(100, 100)` and scale has shape `(4, 4)`, the error will never get raised since 100 is perfectly divisible by 4. Let me know if you want to do something like

```python
if calc_block_size not in [16, 32, 64, 128, 256, 512]:
    raise ValueError("Block size must be a valid power of 2 from 16 to 512")
```

kyuyeunk: also. can you slightly tweak the error message that basically says: "if you want to perform unaligned dequantization, please pass block size as an argument."

kyuyeunk: >  If tensor_q has shape (100, 100) and scale has shape (4, 4), the error will never get raised since 100 is perfectly divisible by 4. Let me know if you want to do something like

I think that's okay! the default expectation is that tensor shapes should always aligned. so for the example you've described, i think we can just say it's an as: "intended behavior. if you want a proper dequantization with unaligned case, make sure to pass in block size as well"

kyuyeunk (approved):

kyuyeunk:
thank you for the awesome change! triggered ci. please ping me if i forget to press merge even when ci has been passed.

catswe:
I removed the "simulate varying scales" code block since it was just a nice extra check. It causes repeated failing fp8 quant tests on TPU7x (seems like TPU7x is a lot more sensitive numerically since the same one without TPU7x is fine?). I don't really want to adjust test tolerances away from default just to account for this either. In practice, new bugs would likely still be caught by other errors raised by the test

kyuyeunk:
> I removed the "simulate varying scales" code block since it was just a nice extra check. It causes repeated failing fp8 quant tests on TPU7x (seems like TPU7x is a lot more sensitive numerically since the same one without TPU7x is fine?). I don't really want to adjust test tolerances away from default just to account for this either. In practice, new bugs would likely still be caught by other errors raised by the test

got it. that makes sense to me. anyway, merged!

PR #1021, opened 2025-11-06
Set VLLM_DISABLE_SHARED_EXPERTS_STREAM=1 by default for TPU inference

xingliu14:
# Description

Set `VLLM_DISABLE_SHARED_EXPERTS_STREAM=True` for TPU inference by default.

FIXES: #1039.

# Tests
```
MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=Qwen/Qwen3-30b-a3b --tensor-parallel-size 4 --max-model-len 12288 --gpu-memory-utilization 0.95 --no-enable-prefix-caching --max-num-seqs 1024 --max-num-batched-tokens 4096 --enable-expert-parallel
```

Current output:
```
liuxing_google_com@t1v-n-cfbce629-w-0:~/tpu-inference$ MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=Qwen/Qwen3-30b-a3b --tensor-parallel-size 4 --max-model-len 12288 --gpu-memory-utilization 0.95 --no-enable-prefix-caching --max-num-seqs 1024 --max-num-batched-tokens 4096 --enable-expert-parallel
/home/liuxing_google_com/tpu-inference/vllm_env/lib/python3.12/site-packages/transformers/utils/hub.py:110: FutureWarning: Using `TRANSFORMERS_CACHE` is deprecated and will be removed in v5 of Transformers. Use `HF_HOME` instead.
  warnings.warn(
INFO 11-07 06:32:40 [__init__.py:27] TPU info: node_name=liuxing-v6e | tpu_type=v6e-8 | worker_id=0 | num_chips=8 | num_cores_per_chip=1
INFO 11-07 06:32:40 [importing.py:44] Triton is installed but 0 active driver(s) found (expected 1). Disabling Triton to prevent runtime errors.
INFO 11-07 06:32:40 [importing.py:68] Triton not installed or not compatible; certain GPU-related functions will not be available.
WARNING 11-07 06:32:40 [interface.py:171] Failed to import from vllm._C: ModuleNotFoundError("No module named 'vllm._C'")
[...]

PR #1455, opened 2026-01-14
Integrate fused moe kernel with qwen3 coder

vanbasten23:
…rd model.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/mxfp4.py
+++ b/tpu_inference/layers/vllm/quantization/mxfp4.py
@@ -114,11 +114,7 @@ class VllmMxfp4MoEMethod(Mxfp4MoEMethod):
         if self.moe_backend == FusedMoEBackend.FUSED_MOE:
             # When fused moe kernle is used, we pass extra arguments like
             # tuned block sizes to the kernel.
-            self.extra_backend_kwargs = dict(
-                subc_quant_w1_sz=REQUANTIZED_BLOCK_SIZE,
-                subc_quant_w2_sz=REQUANTIZED_BLOCK_SIZE,
-                ep_axis_name=ep_axis_name,
-            )
+            self.extra_backend_kwargs = dict(ep_axis_name=ep_axis_name, )
 
     def get_fused_moe_quant_config(
             self, layer: torch.nn.Module) -> FusedMoEQuantConfig | None:


tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -249,18 +249,7 @@ def __init__(self,
...
-                bt=256,
-                bf=1024,
-                bd1=1024,
-                bd2=1024,
-                btc=256,
-                bfc=1024,
-                bd1c=1024,
-                bd2c=1024,
-            )
+            self.extra_backend_kwargs = dict(ep_axis_name=ep_axis_name, )
```
kyuyeunk, line 252: can you make similar changes to all the files that calls fused moe kernel? example: https://github.com/vllm-project/tpu-inference/blob/c83ae8d2a7825cdef5aa6329544f9a4110725262/tpu_inference/layers/vllm/quantization/mxfp4.py#L121-L129

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -98,10 +97,6 @@ def __init__(
         self.moe_backend = select_moe_backend(self.moe)
 
         self.extra_backend_kwargs = {}
```
kyuyeunk, line 99: i don't think you can delete this? don't you need to replace `raise NotImplementedError` to something like `self.extra_backend_kwargs = dict(ep_axis_name=ep_axis_name, )`

.buildkite/pipeline_jax.yml
```diff
@@ -294,6 +294,22 @@ steps:
...
+     key: ${KEY_PREFIX}test_17
+     soft_fail: true
+     env:
+       IS_FOR_V7X: "${IS_FOR_V7X}"
+     agents:
+       queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_8_queue}
+     commands:
+       - |
+         if [[ "$$IS_FOR_V7X" == "true" ]]; then
+           .buildkite/scripts/run_in_docker.sh bash /workspace/tpu_inference/tests/e2e/check_lm_eval.sh "Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8"
```
kyuyeunk: just realized that we don't have similar test for gpt-oss. it can be done in other pr, but can you create a similar change that check accuracy for gpt-oss as well?

tests/e2e/check_lm_eval.sh
```diff
@@ -0,0 +1,72 @@
...
+set -e # Exit immediately if a command exits with a non-zero status.
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <model_name>. A model name is needed."
+    exit 1
+fi
+MODEL_NAME=$1
+
+model_args_json=$(printf '{"pretrained": "%s", "tensor_parallel_size": 8, "max_model_len": 2048, "max_num_batched_tokens": 2048, "max_gen_toks": 256, "enable_expert_parallel": 1}' "$MODEL_NAME")
+output=$(USE_MOE_EP_KERNEL=1 VLLM_DISABLE_SHARED_EXPERTS_STREAM=1 MODEL_IMPL_TYPE=vllm lm_eval \
```
kyuyeunk: `VLLM_DISABLE_SHARED_EXPERTS_STREAM=1` shouldn't be needed anymore: https://github.com/vllm-project/tpu-inference/pull/1021

tests/e2e/check_lm_eval.sh
```diff
@@ -0,0 +1,72 @@
...
+
+set -e # Exit immediately if a command exits with a non-zero status.
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <model_name>. A model name is needed."
+    exit 1
+fi
+MODEL_NAME=$1
+
+model_args_json=$(printf '{"pretrained": "%s", "tensor_parallel_size": 8, "max_model_len": 2048, "max_num_batched_tokens": 2048, "max_gen_toks": 256, "enable_expert_parallel": 1}' "$MODEL_NAME")
```
kyuyeunk: I think it might be a good idea to make this test more generic and not hard code to specific config (like `enable_expert_parallel`, `USE_MOE_EP_KERNEL`) or specific threshold.

can you modify this code so that those configs are passed via argument instead?

tests/e2e/check_lm_eval.sh
```diff
@@ -0,0 +1,72 @@
...
+    --tasks gsm8k_cot \
+    --batch_size auto \
+    --apply_chat_template \
+    --num_fewshot 8)
+
+echo "Evaluation output:"
+echo "$output"
+
+# Thresholds
+flex_threshold=0.87
```
kyuyeunk: what exactly is `flex_threshold` and `strict_threshold`?

tpu_inference/kernels/fused_moe/v1/tuned_block_sizes.py
```diff
@@ -215,4 +316,8 @@ def get_tuned_block_sizes(
         ep_size,
     )
 
+    if key not in TUNED_BLOCK_SIZES:
+        print(
```
kyuyeunk: please use `logger.warning_once` instead. example: https://github.com/vllm-project/tpu-inference/blob/c83ae8d2a7825cdef5aa6329544f9a4110725262/tpu_inference/kernels/ragged_paged_attention/v3/tuned_block_sizes.py#L4368

tpu_inference/layers/vllm/fused_moe.py
```diff
@@ -70,6 +70,14 @@ def fused_moe_apply(
     with jax.named_scope(layer._get_name()):
         match moe_backend:
             case FusedMoEBackend.FUSED_MOE:
+                subc_quant_w1_sz = None
+                subc_quant_w2_sz = None
+                if layer.moe_quant_config.per_out_ch_quant or layer.moe_quant_config.is_block_quantized:
```
kyuyeunk: shouldn't it be possible to figure out a `subc_quant_w1_sz` and `subc_quant_w2_sz` by utilizing information from weight shape and scale shape? example: https://github.com/vllm-project/tpu-inference/blob/c83ae8d2a7825cdef5aa6329544f9a4110725262/tpu_inference/kernels/megablox/gmm.py#L395-L398

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
vanbasten23: I tested this path and it doesn't necessarily need it. This is because the kernel set `ep_axis_name` with the default value "model" https://github.com/vllm-project/tpu-inference/blob/248ff5e1b2562b0100f3822b4ecb97ada76f8be5/tpu_inference/kernels/fused_moe/v1/kernel.py#L1258

But I think this is a good point and I agree with you that we'd better set it explicitly.

.buildkite/pipeline_jax.yml
vanbasten23: For sure. Will do in a follow-up pr.

~~One more thing, I just realized that the v7 machine in our CI only has 1 chip. Do you know if qwen3-coder or gpt-oss can run in v6e-8?~~

Edit: Nvm, I find v7-8 machine in the CI.

tests/e2e/check_lm_eval.sh
vanbasten23: The lm_eval output looks like
```
|  Tasks  |Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|---------|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k_cot|      3|flexible-extract|     8|exact_match|↑  |0.8802|±  |0.0089|
|         |       |strict-match    |     8|exact_match|↑  |0.7453|±  |0.0120
```
The 'Value' shows the accuracy. 'flex_threshold` is the threshold for the 'flexible-extract' score.

tpu_inference/layers/vllm/fused_moe.py
vanbasten23: Note, hidden_size=6144, intermediate_size=2560. Here
w13_weight: [160, 2, 6144, 2560]=[num_expert, 2, hidden_size, intermediate_size]
w2_weight: [160, 2560, 6144]=[num_expert, intermediate_size, hidden_size]
w13_weight_scale: [160, 2, 1, 1, 2560]=(num_experts, 2, hidden_size // subc_quant_w1_sz, 1, intermediate_size)
w2_weight_scale: [160, 1, 1, 6144]=(num_experts, intermediate_size // subc_quant_w2_sz, 1, hidden_size)

Can you elaborate the issue with the current implementation? In the fused moe, the spec wants to use block quant. But we decided that for now we will use per-channel quant in the fused moe kernel. Is my understanding correct?

Or do you mean something like: because w13_weight_scale.shape==[160, 2, 1, 1, 2560]==(num_experts, 2, hidden_size // subc_quant_w1_sz, 1, intermediate_size), so 1==hidden_size // subc_quant_w1_sz, so subc_quant_w1_sz should be hidden_size?

vanbasten23:
for some reason, the lm_eval works locally but fails in the CI. Need to look into it...

[Edit]: i figured it out

tpu_inference/layers/vllm/fused_moe.py
kyuyeunk: yes, so we can think of it like this
- `w2_weight=[160, 2560, 6144]=[num_expert, intermediate_size, hidden_size]`
- therefore, `intermediate_size=2560`
- `w2_weight_scale=[160, 1, 1, 6144]=(num_experts, intermediate_size // subc_quant_w2_sz, 1, hidden_size)`
- therefore, `intermediate_size // subc_quant_w2_sz=1`
- putting them `subc_quant_w2_sz=intermediate_size // (intermediate_size // subc_quant_w2_sz)=2560 // 1 = 2560`

also, i think the existing logic in here will break for subchannel case? because if a weight is subchannel quantized, `subc_quant_w1_sz` will not be `padded_hidden_size` and will be some different value like 512 or 256. Meanwhile,  the technique i've described above will be applicable regardless of subchannel or per-channel quantization.

reason why i prefer this approach is that i'm trying my best to utilize functional programming style where it minimizes an access to class member variable & everything can be figured out via argument's shapes, numerics, etc.

for instance, if vllm upstream changes their variable name from `layer.moe_quant_config.per_out_ch_quant` to `layer.moe_quant_config.per_out_channel_quant`, this logic will break.

But if we write logic in a way we determine `subc_quant_w2_sz` by purely shapes of weights & weight scale, it is extremely robust to upstream changes.

vanbasten23: Got it. Thanks for the explanation. I like the functional programming way.

tpu_inference/layers/vllm/quantization/mxfp4.py
```diff
@@ -118,15 +118,6 @@ def __init__(
                 subc_quant_w1_sz=REQUANTIZED_BLOCK_SIZE,
```
kyuyeunk: you need to remove these lines to resolve ci failure.

kyuyeunk (approved):

PR #1458, opened 2026-01-14
Add 1D blockwise quantization

BirdsOfAFthr:
Following PRs will continue plumbing this feature.

# Description

Support 1d blockwise subchannel quantization in dense matmuls.

# Tests

Added unit tests

# Checklist

X I have performed a self-review of my code.
X I have necessary comments in my code, particularly in hard-to-understand areas.
X I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/tests/kernels/quantized_matmul_kernel_test.py
+++ b/tests/kernels/quantized_matmul_kernel_test.py
@@ -254,6 +254,8 @@ class QuantizedMatmulKernelTest(jtu.JaxTestCase):
         n_input_features: int,
         n_output_features: int,
     ):
+        if not jtu.is_device_tpu_at_least(version=7):
+            self.skipTest("Expect TPUv7+")
         self._test_quantized_matmul(
             dtype,
             jnp.float4_e2m1fn,
--- a/tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
+++ b/tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
@@ -16,6 +16,7 @@ from tpu_inference.kernels.quantized_matmul.util import (get_kernel_name,
                                                          unfold_args)
 
 quantize_tensor = util.quantize_tensor
+MXU_SIZE = 256
 
 
 @functools.partial(
@@ -129,7 +130,8 @@ def quantized_matmul_kernel(
     steps_k = in_block_size // block_size
     # n_lane_multiplier > 1 could improve perf by reducing loop overhead and increasing instruction-level parallelism,
     # allowing the compiler to overlap output fusion and packing overhead with MXU computation
-    compute_tile_n = pltpu.get_tpu_info().mxu_column_size * n_lane_multiplier
+    # TODO(amandaliang): use pltpu.get_tpu_info().mxu_column_size when JAX version is newer
+    compute_tile_n = MXU_SIZE * n_lane_multiplier
     steps_n = out_block_size // compute_tile_n
 
     def kernel(lhs_ref, rhs_ref, w_scales_ref, out_ref, acc_scratch):


tests/kernels/quantized_matmul_kernel_test.py
inho9606, line 1: Could you test shapes, x=(16,12800) w_q=(5120,12800) w_scale=(25,1,5120)? It seems to occur an error, but the error message is not meaningful like 
[0;36m(EngineCore_DP0 pid=2244378)[0;0m ERROR 01-14 23:34:13 [core.py:938] KeyError: '0-814b6194'

jrplatin: FYI you can get around this by explicitly raising the exception [here](https://github.com/vllm-project/vllm/blob/main/vllm/v1/executor/uniproc_executor.py#L93)

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -0,0 +1,242 @@
...
+    def kernel(lhs_ref, rhs_ref, w_scales_ref, out_ref, acc_scratch):
+        pid_k = pl.program_id(2)
+        is_first_step = pid_k == 0
+        is_last_step = pid_k == (orig_n_in // in_block_size - 1)
+
+        @pl.when(is_first_step)
+        def _init():
+            acc_scratch[...] = jnp.zeros_like(acc_scratch)
+
+        # Pre-calculate all quantized blocks for this tile. Relieves register
```
kyuyeunk: why does it relieve register pressure? doesn't it make register pressure even worse because now we have to quantize all inputs all at once, store them in vreg (or vmem if there is vreg spill) and then perform quantized matmul one at a time right?

wouldn't it be better if this logic was performed inside the for loop in line 163 and only perform input quantization for a input for that loop iteration?

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -0,0 +1,242 @@
...
+
+                rhs_q_slice = (rhs_q_full[n_start:n_end, :])
+                rhs_scale_slice = rhs_scale_full[:, n_start:n_end]
+                if x_q_dtype == jnp.int4 or x_q_dtype == jnp.int8:
+                    preferred_element_type = jnp.int32
+                else:
+                    preferred_element_type = jnp.float32
+                dot_res = jax.lax.dot_general(
+                    lhs_q,
+                    rhs_q_slice.astype(lhs_q.dtype),
```
kyuyeunk: why is this needed? performing w16a16 is slower than w8a16.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -0,0 +1,242 @@
...
+            lhs_sub = lhs_ref[:, k_start:k_end].astype(jnp.float32)
+            l_q, l_s = util.quantize_block(lhs_sub, 1, x_q_dtype)
+            lhs_q_list.append(l_q)
+            lhs_scale_list.append(l_s.astype(acc_dtype))
+            rhs_q_list.append(rhs_ref[:, k_start:k_end])
+            rhs_scale_list.append(w_scales_ref[i, :, :].astype(acc_dtype))
+
+        accumulators = [
+            jnp.zeros((batch_block_size, compute_tile_n), dtype=acc_dtype)
+            for _ in range(steps_n)
```
kyuyeunk: I think creating all these zeros at the initialization step might cause peak vreg memory usage to be too high. only creating zeros in vreg for a given index n and immediately storing back into vmem after computation has been finished might have better performance.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -0,0 +1,242 @@
...
+                res = res * lhs_scale
+                res = res * rhs_scale_slice
+
+                accumulators[j] += res
+        acc_block = jnp.concatenate(accumulators, axis=1)
+        acc_scratch[...] += acc_block
+
+        @pl.when(is_last_step)
+        def _write():
+            out_ref[...] = acc_scratch[...].astype(out_ref.dtype)
```
kyuyeunk: feels redundant and create unnecessary store operation. maybe move this logic into line 190 so that a result is stored into `out_ref` when `is_last_step` is true and stored into `acc_scratch` if its false?

tpu_inference/kernels/quantized_matmul/util.py
```diff
@@ -32,7 +32,10 @@ def unfold_args(
         fn(*fn_conditions)
 
 
-def quantize_tensor(x: jax.Array, dtype: jnp.dtype, dim: int = -1):
+def quantize_tensor(x: jax.Array,
```
kyuyeunk, line 35: what is difference between this function and `quantize_block` function below?

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
BirdsOfAFthr: I could move it inside one loop which is cleaer. I split it out because it showed slightly better perf when done separately.

tpu_inference/kernels/quantized_matmul/util.py
BirdsOfAFthr: Refactored so quantize_tensor calls quantize_block

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
BirdsOfAFthr: Done

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
BirdsOfAFthr: Done

tests/kernels/quantized_matmul_kernel_test.py
```diff
@@ -5,17 +5,55 @@
...
+from tpu_inference.kernels.quantized_matmul import (blockwise_kernel, kernel,
+                                                    tuned_block_sizes, util)
 
 xla_quantized_matmul = util.xla_quantized_matmul
-quantized_matmul_kernel = kernel.quantized_matmul_kernel
+per_channel_kernel = kernel.quantized_matmul_kernel
+blockwise_kernel = blockwise_kernel.quantized_matmul_kernel
 quantize_tensor = util.quantize_tensor
 get_tuned_block_sizes = tuned_block_sizes.get_tuned_block_sizes
+TunedValue = tuned_block_sizes.TunedValue
```
vanbasten23, line 16: Conceptually, 1-d blockwise quantization is the same as subchannel quantization, right?

tests/kernels/quantized_matmul_kernel_test.py
```diff
@@ -35,6 +73,8 @@ def _test_quantized_matmul(
         tuned_value=None,
         atol=0.5,
         rtol=0.5,
+        block_size: int | None = None,
```
vanbasten23, line 76: could you rename `dtype` and `q_dtype` to w_dtype and w_q_dtype (line 67-68) since we have one more x_q_dtype now?

tpu_inference/kernels/quantized_matmul/util.py
```diff
@@ -226,3 +227,29 @@ def validate_inputs(
...
+        keepdims=True,
+    )
+    dtype_max, dtype_min = get_max_min(target_dtype)
+    scale = abs_max / dtype_max
+    scale = jnp.where(scale == 0, 1.0, scale)
+
+    if jnp.issubdtype(target_dtype, jnp.floating):
+        data_q = (data / scale).clip(dtype_min, dtype_max).astype(target_dtype)
+    else:
+        data_q = jnp.round(data / scale).astype(target_dtype)
```
vanbasten23, line 254: why don't you need to clip in this case? In get_max_min, you already calculate the max/min int values.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -0,0 +1,239 @@
...
+    n_out = padded_n_out // out_block_size
+    n_in = padded_n_in // in_block_size
+
+    save_acc = n_in > 1
+    # Remove redundant input quantization logic by caching quantized input. For
+    # best performance, only enable this behavior when single input block is
+    # used per batch.
+    save_x_q = quantize_activation and n_in == 1 and n_out > 1
+
+    acc_dtype = jnp.bfloat16
```
vanbasten23, line 108: shouldn't acc_dtype be jnp.float32?

tpu_inference/kernels/quantized_matmul/util.py
BirdsOfAFthr: For int quant, the way we compute scales guarantees scaled values lie mathematically within [-max_int, max_int], making explicit clipping redundant. For floats (e.g., FP8), clipping is strictly required to enforce the specific finite range (e.g., [-448, 448] for E4M3) and prevent overflows to Inf or NaN due to precision edge cases.

Strictly speaking, it does not hurt to clip for ints, but it just adds unnecessary ops.

tests/kernels/quantized_matmul_kernel_test.py
BirdsOfAFthr: Correct, it's just a name to differentiate from 2D subchannel quant.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
BirdsOfAFthr: On Ironwood we can make use of bf16 vpu ops when we use bf16 scales.

added a todo to make it configurable.

Issue #1386, opened 2026-01-02
[Bug]: Default `top_k` Forces Greedy Sampling, Ignoring temperature


AlienKevin:
### Your current environment

<details>
<summary>My environment</summary>

```text
============================================================
vLLM Environment Info:
============================================================
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 22.04.5 LTS (x86_64)
GCC version                  : (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version                : 14.0.0-1ubuntu1.1
CMake version                : version 4.2.0
Libc version                 : glibc-2.35

==============================
       PyTorch Info
==============================
PyTorch version              : 2.8.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.11.11 | packaged by conda-forge | (main, Mar  3 2025, 20:43:55) [GCC 13.3.0] (64-bit runtime)
Python platform              : Linux-5.19.0-1027-gcp-x86_64-with-glibc2.35

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : False
CUDA runtime version         : No CUDA
CUDA_MODULE_LOADING set to   : N/A
GPU models and configuration : No CUDA
Nvidia driver version        : No CUDA
cuDNN version                : No CUDA
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   52 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          208
On-line CPU(s) list:             0-207
Vendor ID:                       GenuineIntel
Model name:                      Intel(R) Xeon(R) Platinum 8481C CPU @ 2.70GHz
CPU family:                      6
Model:                           143
Thread(s) per core:              2
Core(s) per socket:              52
Socket(s):                       2
Stepping:                        8
BogoMIPS:                        5399.99
[...]

PR #1460, opened 2026-01-14
Fix bug in top-k/top-p sampling

jrplatin:
# Description

This PR fixes https://github.com/vllm-project/tpu-inference/issues/1386.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/sample/sampling.py
+++ b/tpu_inference/layers/jax/sample/sampling.py
@@ -48,8 +48,17 @@ def sample(
         return greedy_sampled
 
     logits = logits.astype(jnp.float32)
-    logits = topk_mask(logits, tpu_sampling_metadata.top_k, replace_val=-1e12)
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -48,8 +48,17 @@ def sample(
         return greedy_sampled
 
     logits = logits.astype(jnp.float32)
-    logits = topk_mask(logits, tpu_sampling_metadata.top_k, replace_val=-1e12)
-    logits = topp_mask(logits, tpu_sampling_metadata.top_p, replace_val=-1e12)
+    # Only apply top-k masking if k > 0 for each token
+    top_k = tpu_sampling_metadata.top_k
```
gpolovets1, line 52: Is this a jax array or a config setting? If it's the latter, can we not just use an if-else instead of jnp.where?

jrplatin: This is a JAX array, yep!

PR #1465, opened 2026-01-15
Add tuned block sizes for tpu v7 when running qwen-coder-480b & gptoss

rupengliu-meta:
# Description

When I ran qwen-coder-480b with --max-num-seqs as 512, I noticed that in the log it uses 64 as page size, while there is no tuned block sizes with page size 64 on tpuv7, so I used the tune script provided by Google with some modifications and tuned the block sizes with 64 and targeting the qwen-coder-480 model spec. It improves the perf by around 3% e2e.

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist


rupengliu-meta:
is it possible to get these new block sizes in? if not, then I will not spend time fixing the presubmit

rupengliu-meta:
can any one add a ready label to the PR? thanks

tpu_inference/kernels/ragged_paged_attention/v3/tuned_block_sizes.py
```diff
@@ -32,6 +32,17 @@
 #   - (num_kv_pages_per_block, num_queries_per_block)
 TUNED_BLOCK_SIZES = {
     'TPU v7': {
+        64: {
```
vanbasten23, line 35: Here is my concern: since we keep 2 copies of the kernel/tuned_table, one internal and one OSS, with the internal one being the source of truth, this change may risk of being overwritten in the future: when we migrate the internal tuned table to OSS. The ideal case is we tune for the missing keys internally and migrate the complete tuned table here.

As a compromise, to avoid being overwritten, I think we can have another OSS_TUNED_BLOCK_SIZE and add your block sizes there and the logic becomes:
if we find the block sizes in TUNED_BLOCK_SIZES:
    we'll use it.
elif we find it in OSS_TUNED_BLOCK_SIZE:
    we'll use it
else:
    we'll use the default one.

Then in the future when we migrate from internal to OSS, we only migrate `TUNED_BLOCK_SIZES` table and the `OSS_BLOCK_SIZE` will not be overwritten.

@bythew3i @kyuyeunk let me know what you think.

kyuyeunk: I think current approach is fine? Also, we want to move away from keeping two copies. Instead, better approach is to have an infra in place that allows automatically syncing internal and oss version.

kyuyeunk (commented):
Looks good to me. Triggered CI.

PR #1468, opened 2026-01-16
[DeepSeek]Support 2D TP for Deepseek

bzgoogle:
# Description

2D TP support for DeepSeek. 

# Tests

To test, we have to patch the ShardingAxisNameBase in layers/common/sharding.py with [this](https://paste.googleplex.com/6647958030385152)

# Checklist


tpu_inference/layers/jax/attention/deepseek_v3_attention.py
```diff
@@ -397,13 +397,13 @@ def attention(
             self.query_tnh,  # q
             self.keyvalue_skh,  # k
             self.keyvalue_skh,  # v
-            P(None, None, "model"),  # kv_cache
+            P(None, None, ShardingAxisName.ATTN_HEAD),  # kv_cache
```
gpolovets1, line 400: Should this be MLP_TENSOR if you are using 2D TP?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -769,75 +796,50 @@ def _create_mla() -> MLA:
...
+                dtype=dtype,
+                num_local_experts=num_local_experts,
+                apply_expert_weight_before_computation=False,
+                hidden_size=hidden_size,
+                intermediate_size_moe=moe_intermediate_size,
+                num_experts_per_tok=num_experts_per_token,
+                mesh=self.mesh,
+                hidden_act=hidden_act,
+                rngs=self.rng,
+                activation_ffw_td=(ShardingAxisName.MLP_DATA, ShardingAxisName.MODEL_1),
```
gpolovets1: I believe you forgot to update sharding.py with these new axes.

tpu_inference/runner/compilation_manager.py
```diff
@@ -451,7 +451,7 @@ def _precompile_select_from_array(self) -> None:
                 indices_paddings=self.runner.num_reqs_paddings,
                 hidden_dim=vocab_size,
                 input_sharding=NamedSharding(self.runner.mesh,
-                                             PartitionSpec(None, "model")),
+                                             PartitionSpec(None, ShardingAxisName.TENSOR)),
```
gpolovets1: Do you mean to use MLP_TENSOR?

tpu_inference/runner/kv_cache.py
```diff
@@ -41,7 +42,13 @@ def get_kv_cache_shape_with_mesh(mesh: Mesh,
                                  use_mla: bool = False):
     """Gets the KV cache shape based on the mesh configuration."""
 
-    model_cnt = mesh.shape["model"]
+    axis_name = ShardingAxisName.ATTN_HEAD
```
gpolovets1, line 45: ATTN_HEAD is mapped to "model" so wondering why we need this?

tpu_inference/layers/jax/attention/deepseek_v3_attention.py
bzgoogle: forgot to append. I think we have to patch the sharding config with those in https://paste.googleplex.com/6647958030385152, otherwise, we have to construct a big refactor of how the model-specific sharding could propagate across the whole stack

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: the same as above

tpu_inference/runner/compilation_manager.py
bzgoogle: I created a new variable of TENSOR in the local sharding file

tpu_inference/runner/kv_cache.py
bzgoogle: I overwritten it to ("model", "expert")

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: I made another update by introducing a new sharding class: ShardingAxisName2DTP

tpu_inference/layers/common/sharding.py
```diff
@@ -35,13 +35,28 @@ class ShardingAxisNameBase:
...
     MLP_DATA = 'data'
+    TENSOR = 'model'
     ATTN_HEAD = 'model'
     ATTN_TENSOR = None
     MLP_TENSOR = ('attn_dp', 'model', 'expert')
     MOE_TENSOR = ('attn_dp', 'model')
     EXPERT = ('attn_dp', 'expert', 'model')
     VOCAB = ('expert', 'attn_dp', 'model')
 
+class ShardingAxisName2DTP:
```
kyuyeunk: @wenxindongwork can you take a look? I feel like creating a completely separate class `ShardingAxisName2DTP` isn't a nice solution & i feel like there should be a better one.

tpu_inference/layers/common/sharding.py
```diff
@@ -35,13 +35,28 @@ class ShardingAxisNameBase:
...
+    """sharding class for sharding MoE tensors on 2D tensor parallelism."""
+    SEQUENCE = ('data', 'attn_dp')
+    ATTN_DATA = ('data', 'attn_dp')
+    MLP_DATA = 'data'
+    TENSOR = ('model', 'expert')
+    ATTN_HEAD = ('model', 'expert')
+    ATTN_TENSOR = None
+    MLP_TENSOR = ('attn_dp', 'model', 'expert')
+    MOE_TENSOR = ('attn_dp', 'model')
+    EXPERT = ('attn_dp', 'expert', 'expert')
```
kyuyeunk: having two `'expert'` isn't going to work?

tpu_inference/layers/common/sharding.py
wenxindongwork: I think we can safely replace `ShardingAxisNameBase` with `ShardingAxisName2DTP`. i.e. just change `ShardingAxisNameBase` instead of adding a new class. The only difference is the placement of expert axis. The placement of the expert axis in `ShardingAxisName2DTP` looks correct while the ones in `ShardingAxisNameBase` are not.

kyuyeunk:
can you share trace when using 2d tp, how much performance gain is observed, (if there is a gain) explanation on where the gain is coming from?

patemotter:
> can you share trace when using 2d tp, how much performance gain is observed, (if there is a gain) explanation on where the gain is coming from?

There is a significant improvement over mainline. But when comparing this PR against itself (1D vs 2D sharding) the performance results are nearly identical. More info can be found here b/478266533

tpu_inference/envs.py
```diff
@@ -133,6 +133,9 @@ def _get_bool_env() -> bool:
     "MODEL_IMPL_TYPE":
     env_with_choices("MODEL_IMPL_TYPE", "auto",
                      ["auto", "vllm", "flax_nnx", "jetpack"]),
+    # Enable 2D tensor parallelism
```
jrplatin: Not sure if this (or somehwere else) is the place to do this, but can we expand on what "2D TP" is? I think I have a solid idea but from the code, I'm not sure it's immediately obvious to someone without context

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -879,8 +879,8 @@ def _create_mla() -> MLA:
                 dtype=dtype,
                 moe_backend=self.moe_backend,
                 activation_ffw_td=(ShardingAxisName.MLP_DATA, None),
-                ed_sharding=(ShardingAxisName.MLP_TENSOR, None),
```
jrplatin, line 882: Why is this needed?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -96,6 +97,12 @@ def get_kv_cache_spec(self):
         block_size = self.runner.cache_config.block_size
         kv_cache_spec: dict[str, KVCacheSpec] = {}
 
+        tp_axis_name = ShardingAxisName.ATTN_HEAD
+        if isinstance(tp_axis_name, (tuple, list)):
```
jrplatin: nit: I see this logic in `tpu_inference/runner/kv_cache.py` as well -- maybe move to utility?

jrplatin (approved):

tpu_inference/models/jax/deepseek_v3.py
patemotter: I believe it's just opting for replicating the relatively small matrices instead of sharding them. @bzgoogle was this the intent or is there something else to this?

tpu_inference/runner/kv_cache_manager.py
patemotter: Sure, will revise

tpu_inference/envs.py
patemotter: I added a small blurb in the new commit that should hopefully help.

tpu_inference/models/jax/deepseek_v3.py
bzgoogle: Yes, it is to replicating the small router weights, to get rid of some collective operations.

tpu_inference/layers/common/sharding.py
```diff
@@ -35,13 +35,15 @@ class ShardingAxisNameBase:
     SEQUENCE = ('data', 'attn_dp')
     ATTN_DATA = ('data', 'attn_dp')
     MLP_DATA = 'data'
-    ATTN_HEAD = 'model'
+    TENSOR = ('model', 'expert')
```
gpolovets1, line 38: Why do we need this if it has the same mesh axes as ATTN_HEAD?

tpu_inference/layers/common/sharding.py
```diff
@@ -35,13 +35,15 @@ class ShardingAxisNameBase:
...
+    ATTN_HEAD = ('model', 'expert')
     ATTN_TENSOR = None
     MLP_TENSOR = ('attn_dp', 'model', 'expert')
     MOE_TENSOR = ('attn_dp', 'model')
     EXPERT = ('attn_dp', 'expert', 'model')
-    EXPERT_DATA = ('data', 'attn_dp', 'expert', 'model')
-    VOCAB = ('expert', 'attn_dp', 'model')
+    VOCAB = ('model', 'attn_dp', 'expert')
```
gpolovets1, line 44: Isn't this redundant with EXPERT and MLP_TENSOR as well?

gpolovets1 (approved):
Seem to be a few redundant ShardingAxisNames but other than that lgtm.

tpu_inference/layers/common/sharding.py
patemotter: Arguments could be made that having the various sharding tuples could be helpful for expressing things more clearly in the code. Instead of referring to multiple unrelated things with the same sharding name, it can be easier to read if the sharding name is in-line with what is happening locally. 

It also allows for more fine-grained control as needed. So to me it is acceptable even if right now more than one value has the same tuple.

gpolovets1: I see, makes sense! Although in that case we may want to make a few of the names more descriptive (e.g. TENSOR) in the future.

patemotter: Yes I think that would be a great benefit going forward.

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -21,8 +21,8 @@
 import vllm.envs as envs
 from jax.sharding import NamedSharding, PartitionSpec
 from torchax.ops.mappings import t2j_dtype
+from vllm.attention.layer import Attention
```
lk-chen, line 24: I guess you need to rebase and use `from vllm.v1.attention.backend import AttentionType`

patemotter: Checking it now.

patemotter: Rebased, and it seems to be good.

PR #1469, opened 2026-01-16
[DeepSeek] Enable fp8 quantization for the DeepSeek MoE

bzgoogle:
# Description

In this PR, we enable FP8 quantization for DeepSeek MoE.  We also add testing for the core JAX MoE utilities.

# Note
Integrating the kernel changes for Megablox (as described below) are out-of-scope for this PR, and we will tackle them later, but we expect to see 10-15% E2E improvement from those.

# Tests

Tested MMLU (with MLA + Megablocks + TPU-friendly checkpoit) and both acc/perf look as expected:

```
USE_MEGABLOCKS=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax nohup vllm serve --model=jrplatin/DeepSeek-R1-1D-Subchannel-256-FP8 --max-model-len=5120 --tensor-parallel-size 8 --max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching --disable-log-requests --gpu-memory-utilization 0.95 --additional_config='{"sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' &


python benchmarks/benchmark_serving.py --backend vllm --model=jrplatin/DeepSeek-R1-1D-Subchannel-256-FP8 --dataset-name mmlu --dataset-path /mnt/disks/jacobplatin/mmlu/data/test/ --num-prompts 1000  --run_eval



Benchmark duration (s):                  104.14    
Total input tokens:                      162129    
Request throughput (req/s):              9.60      
Output token throughput (tok/s):         19.20     
Total token throughput (tok/s):          1576.01   
Mean TTFT (ms):                          59719.42  
Median TTFT (ms):                        63025.58  
P99 TTFT (ms):                           102915.19 
Mean TPOT (ms):                          632.33    
Median TPOT (ms):                        634.57    
P99 TPOT (ms):                           637.73    
Mean ITL (ms):                           632.33    
Median ITL (ms):                         634.58    
P99 ITL (ms):                            637.73    
Mean E2EL (ms):                          60351.76  
Median E2EL (ms):                        63661.50  
P99 E2EL (ms):                           103578.83 

{'accuracy': 0.844, 'gen_num': 1000}
```

# Checklist


tpu_inference/kernels/megablox/gmm_ds.py
gpolovets1, line 1: It's a bit hard to parse what has been introduced in this file or why it was moved out of kernels/megablox. I think it may be clearer to make the kernel changes in a separate PR

bzgoogle: a temporary solution given current megablox force to use rhs_transpose, which may not be suitable for DeepSeek. We could merge the gmm_ds.py with gmm.py in a later PR.

gpolovets1: Any specific sections that need review in this file? Should I diff it against gmm.py?

jrplatin: I think we want to diff with the reference Pallas [GMM impl](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/gmm.py) -- added a link to this in the file

jrplatin: Made a diff [here](https://diff.googleplex.com/#key=pdfR85oS0Rhy) for easier viewing!

kyuyeunk: i don't get why this file is needed. please refrain from creating any model specific code and always attempt to create a generic solution.

kyuyeunk: also, if you really want to create a variant, don't create a model specific variant. and instead create a variant based on functionality. for instance, for rpa, we don't say `kernel_gpt_oss.py` we say `kernel_hd64.py` because we name it based on its functionality.

kyuyeunk: the last point really helps with reviewer. when i'm reviewing, i want to know its ***functionality***. but `gmm_ds.py` only tells me its ***usage*** and not what it does.

also,  name like `gmm_ds.py` will make people treat it like **"bespoke kernel that has bunch of deepseek specific features"**, but if you name it like `gmm_a2a_pipelined.py` or something like this, it sends a clear message on what it does, when it is useful, and most importantly - which part can be removed without hurting its core functionality - which is significant improvement to maintainability.

jrplatin: Fair points -- I was syncing offline with a few folks on this as well, so let me see if I can just consolidate this to one file

kyuyeunk: thanks @jrplatin! while doing so, if you can also provide summary of
1. what is the core issue when trying to use current kernel on deepseek
2. how does this new code fix the issue

it will greatly help me provide context and thus give much better feedbacks for code review.

jrplatin: Will address this in another PR!

tpu_inference/kernels/megablox/gmm.py
```diff
@@ -338,7 +338,7 @@ def gmm(
 
   Args:
     lhs: A 2d, jnp.ndarray with shape [m, k].
-    rhs: A 3d, jnp.ndarray with shape [num_groups, n, k].
+    rhs: A 3d, jnp.ndarray with shape [num_groups, k, n].
```
patemotter, line 341: I assume the code was already formatted in this way, but the docstring had a typo?

tests/layers/jax/moe/test_moe_utils.py
```diff
@@ -0,0 +1,284 @@
...
+
+        _, _, size, expert_ids = local_permute_fn(inputs,
+                                                  global_group_sizes,
+                                                  local_expert_size=2,
+                                                  shard_index=0,
+                                                  is_offset=False)
+
+        # We expect it to process the local experts
+        self.assertEqual(list(size), [1, 1])
+        print(expert_ids)
```
gpolovets1: Do we need this if it's already asserted?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -211,16 +208,27 @@ def __call__(self, x_TD: Float):
...
+                    gating_up_proj_spec = (PartitionSpec(*self.edf_sharding),
+                                           PartitionSpec(
+                                               self.edf_sharding[0], None,
+                                               self.edf_sharding[2]))
+                    down_proj_spec = (PartitionSpec(*self.efd_sharding),
+                                      PartitionSpec(self.efd_sharding[0], None,
+                                                    self.efd_sharding[2]))
+                else:
+                    gating_up_proj_spec = PartitionSpec(*self.edf_sharding)
+                    down_proj_spec = PartitionSpec(*self.efd_sharding)
```
patemotter: Nit: Can we add a comment here about why the sharding is done this way for quantized values? I think someone could reason their way through it, but a comment about the scales sharding could be beneficial.

tests/layers/jax/moe/test_moe_utils.py
jrplatin: Oh whoops, good catch

tests/layers/jax/moe/test_moe_utils.py
```diff
@@ -0,0 +1,284 @@
...
+        self.assertTrue(jnp.array_equal(expert_ids, jnp.array([0, 1, 1, 1])))
```
gpolovets1, line 178: If there are 4 experts wondering why this only includes the first 2?

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -248,23 +245,35 @@ def gmm_fn(inputs, kernel, group_sizes, tile_size, moe_backend, dtype,
...
 
     elif moe_backend == MoEBackend.RAGGED_DOT:
+        # TODO: support more than just fp8_e4m3fn
         inputs = manually_quantize_qwix_activation(
             inputs, "ragged_dot", jnp.float8_e4m3fn, [0], {},
             "absmax") if quantized_dtype else inputs
         ragged_dot_func = qwix_ragged_dot if quantized_dtype else jax.lax.ragged_dot
+        if quantized_dtype:
+            q_value, q_scale = kernel
+            kernel = QArray(q_value, q_scale, qtype=jnp.float8_e4m3fn)
```
patemotter: Is there a reason we are hardcoding the fp8 value instead of using `quantized_dtype` directly?

tpu_inference/kernels/megablox/gmm.py
jrplatin: Correct

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -248,23 +245,35 @@ def gmm_fn(inputs, kernel, group_sizes, tile_size, moe_backend, dtype,
...
                               preferred_element_type=dtype,
                               tiling=(tm, tk, tn))
 
     elif moe_backend == MoEBackend.RAGGED_DOT:
+        # TODO: support more than just fp8_e4m3fn
         inputs = manually_quantize_qwix_activation(
             inputs, "ragged_dot", jnp.float8_e4m3fn, [0], {},
             "absmax") if quantized_dtype else inputs
         ragged_dot_func = qwix_ragged_dot if quantized_dtype else jax.lax.ragged_dot
+        if quantized_dtype:
```
gpolovets1, line 274: nit: can this be placed outside of the backend if-else?

tpu_inference/layers/jax/moe/utils.py
jrplatin: Thanks for pointing this out -- I was actually able to track down a big oversight on my part where FP4 isn't working as expected, so need to refactor a bit, thanks!

tpu_inference/layers/jax/moe/utils.py
jrplatin: I think it's better to leave it as is because each kernel handles the quantized inputs differently

tests/layers/jax/moe/test_moe_utils.py
jrplatin: I believe this intended because `local_permute_fn` operates on a per-shard basis. (here, we set local_expert_size=2 and shard_index=0)

tpu_inference/layers/jax/moe/utils.py
jrplatin: Fixed -- I am going to hard-code the ragged dot path for now in terms of the activation, but weights will be dynamic based on the checkpoint being used -- megablocks should work fine

tpu_inference/layers/jax/moe/moe.py
jrplatin: Added a comment!

PR #1473, opened 2026-01-16
Fix local safetensors loading on TPU

ArkVex:
### Summary
This PR fixes a critical bug where loading a local HuggingFace checkpoint (`.safetensors` format) on TPU would fail with a `TypeError: ShapeDtypeStruct is not a valid JAX type`.

Previously, the standard `model.load_weights(rng)` call failed to populate weights when providing a local file path, leaving the model in an abstract state (containing only shape placeholders). This caused the JIT compilation step to crash because it attempted to compile abstract shapes instead of concrete tensors.

### Changes
* **Manual Weight Loading Fallback:** Modified `_get_nnx_model` (in `model_loader.py`) to detect when a local `model.safetensors` file is present.
* **Abstract State Detection:** Implemented a check to verify if the model is still in an abstract state after initialization. If so, it manually triggers a weight load using `safetensors.flax.load_file`.
* **Key Mapping Support:** Added support for both direct state updates and unflattened (nested) state updates to robustly handle variable key structures in checkpoint files.

### Fixes
Fixes https://github.com/vllm-project/vllm/issues/30375

### Tests
Tested locally on **TPU v5litepod** by loading a converted Qwen2/3 model from a local directory.

* **Before fix:** Crashed with `TypeError: Argument 'model.states...' of type ShapeDtypeStruct is not a valid JAX type`.
* **After fix:** Model loads successfully, weights are correctly populated, and JIT compilation proceeds without error.


ArkVex:
@vipannalla @BirdsOfAFthr  plz review

ArkVex:
@vipannalla plz review this pr

PR #1475, opened 2026-01-16
Integrate the 1d blockwise quantized matmul kernel to fp8 TorchAx

inho9606:
# Description

this PR is to enable blockwise quantized matmul kernel for fp8 dense.

Users can enable the feature by setting REQUANT_BLOCK_SIZE and REQUANT_WEIGHT_DTYPE global variables like:

$ export ENABLE_QUANTIZED_MATMUL_KERNEL=1
$ export REQUANT_BLOCK_SIZE=512
$ export REQUANTIZE_WEIGHT_DTYPE=float4_e2m1fn

Without those above settings, the default will run XLA quantized matmul for per-channel quantization.

# Tests

Updated/added some unit tests to test_env.py and test_fp8.py

Plus, this change was tested with the qwen3-32b-fp8 model with the command like:

MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=Qwen/Qwen3-32B-FP8 --tensor_parallel_size=2 --runner=auto --max_model_len=1024 --download_dir=/mnt/disks/persist

# Checklist

### DIFF
--- a/tests/layers/test_jax.py
+++ b/tests/layers/test_jax.py
@@ -59,6 +59,29 @@ class TestJaxModule(unittest.TestCase):
         self.assertCountEqual(["inner.quantization_scale", "weight"],
                               [k for k, _ in module.named_parameters()])
 
+    def test_named_children_in_nested_module(self):
+        """Tests named_children method in nested modules."""
+
+        class Depth3Module(JaxModule):
+
+            def __init__(self):
+                super().__init__()
+                self.my_module_1 = MyModule()
+                self.my_module_2 = MyModule()
+                self.nested_module_a = NestedModule()
+                self.nested_module_b = NestedModule()
+                self.layers = [MyModule() for _ in range(3)]
+
+        module = Depth3Module()
+
+        children = set(name for name, _ in module.named_children())
+        # Should only return immediate children modules, without "inner" etc.
+        self.assertCountEqual(
+            children, {
+                "my_module_1", "my_module_2", "nested_module_a",
+                "nested_module_b", "layers"
+            })
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/models/jax/utils/test_weight_utils.py
+++ b/tests/models/jax/utils/test_weight_utils.py
@@ -1,0 +1,110 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import tempfile
+from unittest.mock import MagicMock
+
+import jax
+import numpy as np
+import torch
+from flax import nnx
+from jax.sharding import Mesh
+from safetensors.torch import save_file
+from torch import nn
+from vllm.model_executor.model_loader import LoadConfig, get_model_loader
+
+from tpu_inference.layers.jax import JaxModule
+from tpu_inference.layers.jax.linear import JaxLinear
+from tpu_inference.models.jax.utils.weight_utils import LoadableWithIterator
+
+
+class TorchMLP(nn.Module):
+    """MLP implemented with PyTorch."""
+
+    def __init__(self):
+        super().__init__()
+        self.w1 = nn.Linear(2, 6)
+        self.act = nn.ReLU()
+        self.w2 = nn.Linear(6, 2, bias=True)
+
+    def forward(self, x):
+        x = self.w1(x)
+        x = self.act(x)
+        x = self.w2(x)
+        return x
+
+
+class JaxMLP(JaxModule, LoadableWithIterator):
+    """MLP implemented with JAX."""
+
+    def __init__(self, rngs):
+        super().__init__()
+        self.w1 = JaxLinear(2, 6, rngs)
+        self.act = nnx.relu
+        self.w2 = JaxLinear(6, 2, rngs, use_bias=True)
+
+    def __call__(self, x):
+        x = self.w1(x)
+        x = self.act(x)
+        x = self.w2(x)
+        return x
+
+
+class TestJaxAutoWeightsLoader:
+
+    def test_load_from_safetensors(self):
+        """Load weights from a safetensors file saved from a PyTorch model.
+        """
+        torch_model = TorchMLP()
+        with torch.no_grad():
+            torch_model.w1.weight.fill_(1.1)
+            torch_model.w2.weight.fill_(0.9)
+            torch_model.w2.bias.fill_(0.1)
+
+        # Save the PyTorch model weights to a safetensors file. Load them
+        # into the JAX model.
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_file_path = os.path.join(tmpdir, "model.safetensors")
+            save_file(torch_model.state_dict(), tmp_file_path)
+
+            devices = jax.local_devices()
+            mesh = Mesh(devices, axis_names=('p', ))
+            with jax.set_mesh(mesh):
+                jax_model = JaxMLP(rngs=nnx.Rngs(0))
+
+                model_config = MagicMock()
+                model_config.quantization = None
+                model_config.model = tmpdir
+                model_config.revision = None
+
+                loader = get_model_loader(
+                    LoadConfig(load_format="safetensors"))
+                loader.load_weights(jax_model, model_config)
+
+        np.testing.assert_allclose(torch_model.w1.weight.T.detach().numpy(),
+                                   jax_model.w1.weight.value)
+        np.testing.assert_allclose(torch_model.w2.weight.T.detach().numpy(),
+                                   jax_model.w2.weight.value)
+
+        # Forward pass to verify correctness.
+        input_values = [[0.1, 0.2], [0.3, 0.4]]
+        torch_input = torch.tensor(input_values)
+        jax_input = np.array(input_values)
+        torch_output = torch_model(torch_input).detach().numpy()
+        jax_output = jax_model(jax_input)
+        np.testing.assert_allclose(torch_output,
+                                   jax_output,
+                                   rtol=1e-3,
+                                   atol=1e-2)
--- a/tpu_inference/layers/jax/__init__.py
+++ b/tpu_inference/layers/jax/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterator
+from typing import Iterable, Iterator
 
 from flax import nnx
 
@@ -21,9 +21,28 @@ class JaxModule(nnx.Module):
     """Base module for JAX layers, extending flax.nnx.Module.
     """
 
+    def _get_name(self) -> str:
+        return self.__class__.__name__
+
     def named_parameters(self,
-                         prefix: str = "") -> Iterator[tuple[str, nnx.Param]]:
-        """Yields the named parameters of the module."""
+                         prefix: str = "",
+                         recurse=True) -> Iterator[tuple[str, nnx.Param]]:
+        """Yields the named parameters of the module.
+        
+        Arguments:
+            prefix: Prefix to add to the parameter names.
+            recurse: If True, then yields parameters of this module and all submodules.
+                Otherwise, yields only parameters that are direct members of this module.
+
+        Yields:
+            (string, Param): Tuple containing a name and parameter
+        """
+        if not recurse:
+            for name, param in self.__dict__.items():
+                if isinstance(param, nnx.Param):
+                    yield (f"{prefix}.{name}" if prefix else name), param
+            return
+
         params = nnx.state(self, nnx.Param)
 
         def _traverse_params(params, path=()):
@@ -34,3 +53,55 @@ class JaxModule(nnx.Module):
                 yield ".".join(path), params
 
         yield from _traverse_params(params, path=(prefix, ) if prefix else ())
+
+    def named_children(self) -> Iterator[tuple[str, nnx.Module | list]]:
+        """Returns an iterator over immediate children modules.
+        
+        Yields:
+            (string, Module | list): Tuple containing a name and child module
+        """
+        for name, value in self.__dict__.items():
+            if isinstance(value, JaxModule):
+                yield name, value
+            elif isinstance(value, list):
+                yield name, JaxModuleList(value)
+
+
+class JaxModuleList(list):
+    """A list container for JaxModule objects."""
+
+    def __init__(self, modules: Iterable[JaxModule]):
+        """Initializes the JaxModuleList.
+
+        Args:
+            modules: An optional list of JaxModule objects to initialize the list.
+        """
+        super().__init__()
+        for module in modules:
+            self.append(module)
+
+    def _get_name(self) -> str:
+        return self.__class__.__name__
+
+    def named_parameters(self,
+                         prefix: str = "",
+                         recurse=True) -> Iterator[tuple[str, nnx.Param]]:
+        """Yields the named parameters of all modules in the list."""
+
+        for idx, module in enumerate(self):
+            module_prefix = f"{prefix}.{idx}" if prefix else str(idx)
+            yield from module.named_parameters(prefix=module_prefix,
+                                               recurse=recurse)
+
+    def named_children(
+            self) -> Iterator[tuple[str, "JaxModule | JaxModuleList"]]:
+        """Returns an iterator over the modules in the list with their indices as names.
+
+        Yields:
+            (string, JaxModule): Tuple containing the index as a string and the module
+        """
+        for idx, item in enumerate(self):
+            if isinstance(item, JaxModule):
+                yield str(idx), item
+            elif isinstance(item, list):
+                yield str(idx), JaxModuleList(item)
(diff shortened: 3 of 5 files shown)


tpu_inference/layers/vllm/linear.py
```diff
@@ -43,24 +56,27 @@ def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
...
+            scale_sharding = P(out_axis,)
         out_sharding = P(None, out_axis)
 
+        x_q_dtype = _get_x_q_dtype(w_q.dtype)
         x = jax.lax.with_sharding_constraint(x,
                                              NamedSharding(mesh, x_sharding))
 
         def wrapper(x, w_q, w_s):
-            output = quantized_matmul_kernel(x, w_q, w_s, x_q_dtype=w_q.dtype)
+            output = quantized_matmul_kernel(x, w_q, w_s, x_q_dtype=x_q_dtype, block_size=block_size)
```
BirdsOfAFthr: after you rebase on [this](https://github.com/vllm-project/tpu-inference/pull/1458/changes) PR,
you will need to choose the proper `quantized_matmul_kernel` based on if `block_size` is None.

inho9606: thanks, I rebased your recent changes and updated my PR.

tpu_inference/layers/common/linear.py
```diff
@@ -13,17 +13,38 @@
...
 
 
-def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
-                             mesh: Mesh, weight_sharding: P) -> jax.Array:
+def _get_x_q_dtype(w_q_dtype: jnp.dtype) -> jnp.dtype:
+    """Return 8-bit float or integer dtype depending on w_q_dtype."""
+    if jnp.issubdtype(w_q_dtype, jnp.integer):
+        return jnp.int8
+    elif jnp.issubdtype(w_q_dtype, jnp.floating):
+        return jnp.float8_e4m3fn
```
BirdsOfAFthr, line 30: I know we are supporting A8 right now, but this just restricts activation to be 8 all the time.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -102,10 +103,15 @@ def get_quant_method(
...
-                 linear_config: VllmQuantLinearConfig):
+    def __init__(
+        self,
+        quant_config: VllmFp8Config,
+        linear_config: VllmQuantLinearConfig,
+    ):
         super().__init__(quant_config)
         self.linear_config = linear_config
+        self.requant_block_size = envs.REQUANTIZE_BLOCK_SIZE
+        self.requant_weight_dtype = getattr(jnp, envs.REQUANTIZE_WEIGHT_DTYPE)
```
BirdsOfAFthr: nit: would it make more sense to put these two in `VllmQuantLinearConfig`?

tests/layers/vllm/test_fp8.py
```diff
@@ -475,3 +479,43 @@ def test_fused_moe(use_ep, num_devices, num_tokens, intermediate_size,
                                    check_device=False,
                                    atol=2e-2,
                                    rtol=1e-1)
+
+
+@pytest.mark.parametrize("requant_block_size", (128, 512))
+@pytest.mark.parametrize("requant_weight_dtype",
+                         (jax.numpy.float8_e4m3fn, jax.numpy.float4_e2m1fn))
+def test_process_weights_after_loading_for_blockwise_quant(
```
BirdsOfAFthr: have you run this on v7x? It is failing during precheck.

tpu_inference/layers/common/linear.py
inho9606: For the 4bit activation, we may need a new additional flag. I'll add a comment mentioning we need a flag for 4bit activation like w4a4 in the future.

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: done.

tpu_inference/layers/common/linear.py
inho9606: Added a TODO comment.

tests/layers/vllm/test_fp8.py
inho9606: Oops yes I tested it on v7 but I've set my global variable ENABLE_QUANTIZED_MATMUL_KERNEL=true so my local test worked. I fixed my unit test so that it can turn on the enable_quantized_matmul_kernel inside the unit test and it works now. Thanks for the raising!

BirdsOfAFthr: You need to add
```
if not jtu.is_device_tpu_at_least(version=7):
            self.skipTest("Expect TPUv7+")
```
to skip tests on previous generation TPUs.

The tests are failing on other TPUs.

inho9606: done.

BirdsOfAFthr: We need to do the same for test:
```
tests/kernels/fused_moe_v1_test.py::MoEKernelTest::test_activation
```

inho9606: There is another working PR, https://github.com/vllm-project/tpu-inference/pull/1536, for that. After it is merged, I'll rebase it so that the error will go awaye .

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -75,7 +77,8 @@ def quantized_matmul_kernel(
     out_block_size = tuned_value.out_block_size
     in_block_size = tuned_value.in_block_size
     n_lane_multiplier = tuned_value.n_lane_multiplier
-    block_size = tuned_value.in_block_size if block_size is None else block_size
+    # The num_blocks should become 1 in case of channelwise.
+    block_size = tuned_value.in_block_size if block_size == orig_n_in else block_size
```
BirdsOfAFthr, line 81: so we enabled channelwise through our kernel now? 

We still need to validate it through benchmarks, right?

tpu_inference/layers/common/linear.py
BirdsOfAFthr: @kyuyeunk please confirm

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
inho9606: No we don't enable channelwise officialy in this PR. But I think users may put the same block size as the k_dimension by chance. In that case, it will crash so I covered that case.

inho9606:
This PR looks approved, could you confirm this comment so that I can go merge? @BirdsOfAFthr @kyuyeunk  Or if you don't mind, let me close this for now and discuss it later if needed.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -130,21 +133,29 @@ def process_fp8_linear_weights(
...
                     (0, 1),
                 )
+                if requant_block_size and not self.linear_config.enable_quantized_matmul_kernel:
+                    raise ValueError(
+                        "Blockwise quantization is supported by quantized matmul kernel. Please enable quantized_matmul_kernel or unset the quantize block size to trigger XLA per-channel quantization."
+                    )
                 weight_slice, weight_scale_slice = quantize_tensor(
-                    jnp.float8_e4m3fn, dequantzed_weight)
+                    self.linear_config.requant_weight_dtype,
+                    dequantzed_weight,
```
patemotter: Nit (spelling): dequantized_weight

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -130,21 +133,29 @@ def process_fp8_linear_weights(
...
+                                                  original_block_size]
                 dequantzed_weight = dequantize_tensor(
                     weight_slice,
                     weight_scale_slice,
                     (0, 1),
                 )
+                if requant_block_size and not self.linear_config.enable_quantized_matmul_kernel:
+                    raise ValueError(
+                        "Blockwise quantization is supported by quantized matmul kernel. Please enable quantized_matmul_kernel or unset the quantize block size to trigger XLA per-channel quantization."
+                    )
```
patemotter: This information should be known during the `init()` phase, can we move the check there instead of the middle of this jitted loop or is there a specific reason for it to live here?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -167,12 +178,36 @@ def process_fp8_linear_weights(
             )
 
         weights = process_fp8_linear_weights(weight, weight_scale, bias)
+        weight_sharding = self.linear_config.weight_sharding
+        bias_sharding = self.linear_config.bias_sharding
+        weight_scale_sharding = None
+        if self.linear_config.enable_quantized_matmul_kernel:
```
patemotter, line 185: Same comment as above re: knowing info during init phase.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -167,12 +178,36 @@ def process_fp8_linear_weights(
...
+            if self.linear_config.requant_block_size is None:
+                raise ValueError(
+                    "You should set REQUANTIZE_BLOCK_SIZE to enable quantized matmul kernel. Please set the value or disable the quantized matmul kernel."
+                )
+            # The quantized_matmul_kernel expects weight scales shaped (n_out_features, 1, n_blocks) for blockwisze quantization.
+            weights.weight_scale = jnp.expand_dims(jnp.transpose(
+                weights.weight_scale),
+                                                   axis=1)
```
patemotter: Nit: I'm sure this was done as part of our pre-commit process, but this formatting seems wild to me.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -167,12 +178,36 @@ def process_fp8_linear_weights(
...
+            # Weight scale should be resharded accordingly as well.
+            if len(weight_sharding) == 2:
+                # Cannot be sharded on the first dimension in case the number of blocks is 1.
+                in_axis = weight_sharding[1] if num_blocks > 1 else None
+                out_axis = weight_sharding[0]
+                weight_scale_sharding = P(in_axis, None, out_axis)
+            else:
+                raise ValueError(
+                    F"The weight sharding shape length should be 2, but given {len(weight_sharding)}."
+                )
```
patemotter: Nit: I would reformat to be slightly cleaner by keeping the core logic un-nested.

```python

if len(weight_sharding) != 2:
  raise ValueError(
                    F"The weight sharding shape length should be 2, but given {len(weight_sharding)}."
                )

# Cannot be sharded on the first dimension in case the number of blocks is 1.
in_axis = weight_sharding[1] if num_blocks > 1 else None
out_axis = weight_sharding[0]
weight_scale_sharding = P(in_axis, None, out_axis)
                
```

tests/layers/vllm/test_fp8.py
```diff
@@ -480,3 +485,46 @@ def test_fused_moe(use_ep, num_devices, num_tokens, intermediate_size,
...
```
kyuyeunk: can you modify this test to not just test process_weights_after_loading, but also the forward pass as well?

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -145,6 +145,7 @@ def shard_linear_weights(
     mesh: Mesh,
     weight_p_spec: PartitionSpec,
     bias_p_spec: PartitionSpec,
+    weight_scale_p_spec: PartitionSpec | None = None,
```
kyuyeunk: why do you pass `weight_scale_p_spec`? isn't it possible to derived this value from another sharding p specs?

tpu_inference/layers/common/linear.py
kyuyeunk: yes, only supporting a8 sounds good to me. there isn't any use case of a4 for now. it tweak the code to expand support for a4 should be trivial.

tpu_inference/layers/vllm/quantization/configs.py
```diff
@@ -82,6 +84,9 @@ def __init__(self, vllm_config: VllmConfig, mesh: Mesh, layer: LinearBase):
         self.bias_sharding = P(self.weight_sharding[0])
         self.n_shards = get_mesh_shape_product(self.mesh,
                                                self.weight_sharding[0])
+        self.enable_quantized_matmul_kernel = envs.ENABLE_QUANTIZED_MATMUL_KERNEL
+        self.requant_block_size = envs.REQUANTIZE_BLOCK_SIZE
+        self.requant_weight_dtype = getattr(jnp, envs.REQUANTIZE_WEIGHT_DTYPE)
```
kyuyeunk: please use `to_jax_dtype` instead: https://github.com/vllm-project/tpu-inference/blob/b53de59e507e573284432bd1e0c4ece890f4b21e/tpu_inference/utils.py#L37-L49

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -99,8 +99,11 @@ def get_quant_method(
 class VllmFp8LinearMethod(vllm_fp8.Fp8LinearMethod,
                           common_fp8.Fp8LinearMethod):
 
-    def __init__(self, quant_config: VllmFp8Config,
-                 linear_config: VllmQuantLinearConfig):
+    def __init__(
+        self,
+        quant_config: VllmFp8Config,
+        linear_config: VllmQuantLinearConfig,
+    ):
```
kyuyeunk, line 106: seems like unnecessary change. can you roll back?

tpu_inference/layers/vllm/quantization/fp8.py
kyuyeunk: agree. let's move this to `init()` phase.

tpu_inference/layers/vllm/quantization/fp8.py
kyuyeunk: you can get slightly better pre-commit formatting if you add comma after the last elemnt of the argument. so change `axis=1)` to `axis=1,)`

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: done.

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: done.

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: seems to be back by pre-commit.. ACK

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: Thanks! done.

tpu_inference/layers/common/process_weights/linear_weights.py
inho9606: As the weight_scale sharding depends on the number of its weight dimensions, I think it is not easy to guess its correct one from other Specs like weight or bias. Instead, I found the `shard_linear_weights` takes 'weights' as an argument, so the scale spec could be processed within the function. I'll try fix it.

tests/layers/vllm/test_fp8.py
inho9606: Do you mean if we could test its actual quantized matmul outputs? For that, I think we can simply call apply function and check its output. Is it correct you meant by forward pass?

tpu_inference/layers/vllm/quantization/configs.py
inho9606: done.

tpu_inference/layers/vllm/quantization/fp8.py
inho9606: done.

tests/layers/vllm/test_fp8.py
inho9606: I added the forward pass test for the blockwise quantization, but left it to compare unquantized outputs and quantized outputs as TODO since it complains the outputs are different for unknown reasons. I quickly tried to debug but it seems to require a bit time.

tests/layers/vllm/test_fp8.py
```diff
@@ -480,3 +485,53 @@ def test_fused_moe(use_ep, num_devices, num_tokens, intermediate_size,
...
+        )
+
+    initialize_layer_weights(linear_layer)
+    quant_method = linear_layer.quant_method
+    quant_method.linear_config.requant_block_size = requant_block_size
+    quant_method.linear_config.requant_weight_dtype = requant_weight_dtype
+    quant_method.linear_config.enable_quantized_matmul_kernel = True
+
+    quant_method.process_weights_after_loading(linear_layer)
+    weight_jax = jax_view(linear_layer.weight)
```
kyuyeunk, line 525: no need to use jax_view and just do a torch based assertion?

inho9606: For dtype checking, we still need jax_view for jnp comparison. On Torch, it is shown as float32 like:
E       AssertionError: assert torch.float32 == <class 'jax.numpy.float8_e4m3fn'>

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -157,10 +157,24 @@ def shard_linear_weights(
 
     weight_sharding = NamedSharding(mesh, weight_p_spec)
     bias_sharding = NamedSharding(mesh, bias_p_spec)
+    if len(weights.weight_scale.shape) == 3:
```
BirdsOfAFthr: Can you check if `weights` and `weights.weight_scale` is `None` first? The tests are failing..

inho9606: Fixed, thanks.

Issue #1479, opened 2026-01-17
[RFC] fix weight parameter naming for better dev. experience


lk-chen:
### 🚀 The feature, motivation and pitch

As of today, I notice 2 places for potential improvement
1. this [.weight suffice removal](https://github.com/vllm-project/tpu-inference/blob/d0c84b437e9797cefcc221248d85f8f9e60719e8/tpu_inference/models/jax/utils/weight_utils.py#L331) seems meaningless
2. most [mapping](https://github.com/vllm-project/tpu-inference/blob/d0c84b437e9797cefcc221248d85f8f9e60719e8/tpu_inference/models/jax/qwen3.py#L285-L286) is just replace "weight" with "kernel"

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

lk-chen:
Qwen2 Qwen3 DeepSeek has been updated

PR #1481, opened 2026-01-17
[WeightLoader] Skip ".weight" suffix removal conditionally

lk-chen:
# Description

A refactoring work in WIP to make jax model parameters have ".weight" suffix like HF models (torch). We
need to skip removing this suffix conditionally.

see #1479

# Tests

`vllm serve Qwen/Qwen3-4B-Instruct-2507`

`vllm serve Qwen/Qwen2.5-VL-3B-Instruct`

# Checklist

### DIFF
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -328,6 +328,12 @@ def _load_and_shard_weight(vllm_config,
         )
         hf_weight = hf_weight.astype(model_config.dtype)
 
+    # For tensors whose name matches any string in `keep_hf_weight_suffix_when_match`, the
+    # '.weight' suffix in HF keys will be kept.
+    # Context: some models are being refactored to have identical parameter names as HF
+    # models, so the suffix does not need to be removed for those parameters. Eventually
+    # we want to get rid of the ".weight" suffix removal logic altogether.
+    # TODO(#1479): remove this argument and related logic after the refactoring is done.
     if hf_key.endswith(".weight") and all(
             substr not in hf_key
             for substr in keep_hf_weight_suffix_when_match):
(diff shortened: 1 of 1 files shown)


jrplatin:
Thanks for this! Any chance you could add some testing?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -327,19 +328,21 @@ def _load_and_shard_weight(vllm_config,
...
+            substr not in hf_key
+            for substr in keep_hf_weight_suffix_when_match):
         hf_key = hf_key.removesuffix(".weight")
 
     # Find the corresponding model key using the HF key
     if "layers" in hf_key:
         layer_num = re.search(r"layers\.(\d+)", hf_key).group(1)
         layer_key = re.sub(r"layers\.\d+", "layers.*", hf_key)
-        model_key = name_map[layer_key]
+        model_key = name_map.get(layer_key, layer_key)
```
jrplatin, line 346: Why would we want to return the `layer_key` itself as the default here?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -327,19 +328,21 @@ def _load_and_shard_weight(vllm_config,
         )
         hf_weight = hf_weight.astype(model_config.dtype)
 
-    if hf_key.endswith(".weight"):
+    if hf_key.endswith(".weight") and all(
+            substr not in hf_key
```
jrplatin, line 338: Might be good to add a comment here describing what's going on with the second condition since it's not immediately obvious to me

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -647,18 +665,24 @@ def __init__(self, vllm_config: VllmConfig, mesh: Mesh):
...
-        metadata_map = get_default_maps(self.vllm_config.model_config,
-                                        self.mesh, mappings)
-
-        load_hf_weights(vllm_config=self.vllm_config,
-                        model=model,
-                        metadata_map=metadata_map,
-                        mesh=self.mesh,
-                        pp_missing_layers=getattr(model, 'pp_missing_layers',
-                                                  []))
+        if isinstance(mappings, MetadataMap):
```
jrplatin, line 693: Why is this change needed?

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: The plan is:
1. remove identical weight names from [mappings](https://github.com/vllm-project/tpu-inference/blob/fd859cd09d9def7832c8f8c032f5a347b660789c/tpu_inference/models/jax/qwen3.py#L281). (This PR) returns original name if not found in `name_map`.
2. update model file to eliminate inconsistent weight names (e.g. #1480, #1484 ), repeat step 1
3. once mappings is empty, we can get rid of the weight naming conversion logic in `load_hf_model`. This paves the road to refactor weight loading.

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: Added comment for this function

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: Added comment above

lk-chen:
> Thanks for this! Any chance you could add some testing?

I see qwen/deepseek/llama tests all call `load_weights`, so I assume the change is covered there.
Also I manually tested several models (see PR description)

PR #1482, opened 2026-01-17
[jax/torchax unification] Add Linear and Einsum layer

lk-chen:
# Description

Qwen model will be rewritten with the added layer (#1484 ), in order to:
* #1479 : as shown in test, the new layer have same naming as torch
* let flax-nnx path utilize those ops used in vllm path, using similar delegation mechanism (#1485 )

# Tests

```
pytest tests/layers/jax/test_linear.py
```

# Checklist

### DIFF
--- a/tests/layers/jax/test_linear.py
+++ b/tests/layers/jax/test_linear.py
@@ -140,7 +140,7 @@ class TestJaxLinear(unittest.TestCase):
     def test_sharding_assignment(self):
         """Tests sharding assignment of JaxLinear layer."""
 
-        mesh = jax.make_mesh((4, ), ('model', ))
+        mesh = Mesh(jax.devices('cpu')[:1], ("model", ))
         unquantize_config = get_tpu_quantization_config(
             VllmConfig(model_config=ModelConfig(model="Qwen/Qwen3-0.6B")),
             mesh)
--- a/tpu_inference/layers/vllm/quantization/configs.py
+++ b/tpu_inference/layers/vllm/quantization/configs.py
@@ -47,6 +47,11 @@ class VllmQuantLinearConfig(QuantLinearConfig):
             enable_sp=vllm_config.compilation_config.pass_config.enable_sp,
             output_sizes=[layer.output_size])
         self.mesh = mesh
+        self.tp_size = get_mesh_shape_product(self.mesh,
+                                              ShardingAxisName.MLP_TENSOR)
+
+        self.n_shards = get_mesh_shape_product(self.mesh,
+                                               self.weight_sharding[0])
 
         if isinstance(layer, RowParallelLinear):
             self.weight_sharding = P(None, ShardingAxisName.ATTN_HEAD)
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -16,20 +16,12 @@ from typing import Any, Optional
 
 import jax
 import torch
-<<<<<<< HEAD
-
-=======
-import vllm.model_executor.layers.linear as vllm_linear
->>>>>>> 4f8de635 (fix)
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torch.nn.parameter import Parameter
 from torchax.interop import jax_view, torch_view
 from torchax.ops.mappings import t2j
 from vllm.attention.layer import Attention
-<<<<<<< HEAD
 from vllm.model_executor.layers import linear as vllm_linear
-=======
->>>>>>> 4f8de635 (fix)
 from vllm.model_executor.layers.fused_moe import (FusedMoE, FusedMoEConfig,
                                                   UnquantizedFusedMoEMethod)
 from vllm.model_executor.layers.quantization import \
--- a/tpu_inference/layers/common/quantization/configs.py
+++ b/tpu_inference/layers/common/quantization/configs.py
@@ -14,9 +14,6 @@
 
 from jax.sharding import PartitionSpec as P
 
-from tpu_inference.layers.common.sharding import ShardingAxisName
-from tpu_inference.utils import get_mesh_shape_product
-
 
 class QuantLinearConfig:
 
@@ -28,9 +25,4 @@ class QuantLinearConfig:
         self.input_sharding = None
         self.output_sharding = None
 
-        self.tp_size = get_mesh_shape_product(self.mesh,
-                                              ShardingAxisName.MLP_TENSOR)
-
         self.bias_sharding = P(self.weight_sharding[0])
-        self.n_shards = get_mesh_shape_product(self.mesh,
-                                               self.weight_sharding[0])
--- a/tpu_inference/layers/common/quantization/unquantized.py
+++ b/tpu_inference/layers/common/quantization/unquantized.py
@@ -47,15 +47,10 @@ class UnquantizedLinearMethod:
 
     def _apply_split(self,
                      x_jax: jax.Array,
-<<<<<<< HEAD
-                     weight_generator: Sequence[jax.Array],
-=======
-                     weight_generator: Generator[tuple[int, jax.Array], None,
-                                                 None],
->>>>>>> 4f8de635 (fix)
+                     weights: Sequence[jax.Array],
                      bias_jax: Optional[jax.Array] = None) -> jax.Array:
         outs = []
-        for i, weight_jax in enumerate(weight_generator):
+        for i, weight_jax in enumerate(weights):
             out = jnp.einsum("mn,pn->mp", x_jax, weight_jax)
             if bias_jax is not None:
                 out += bias_jax[i]


kyuyeunk:
@gpolovets1 @jrplatin can you help do a first pass review of this pr? I'll do a second pass review when it all looks good from your end.

tpu_inference/layers/jax/linear.py
```diff
@@ -0,0 +1,118 @@
...
+
+class JaxEinsum(nnx.Einsum, JaxModule):
+    """Einsum layer for JAX.
+
+    Args:
+        einsum_str: a string to denote the einsum equation.
+        kernel_shape: the shape of the kernel.
+        bias_shape: the shape of the bias. If this is None, a bias won't be used.
+        param_dtype: Data type for the parameters.
+        quant_config: Quantization configuration.
```
jrplatin, line 33: Nit: add the `prefix` arg here

tpu_inference/layers/jax/linear.py
```diff
@@ -0,0 +1,118 @@
...
+        if self.quant_method is not None:
+            return self.quant_method.apply_jax(self, inputs)
+
+        if self.bias is not None:
+            return jax.numpy.einsum(self.einsum_str, inputs,
+                                    self.weight) + self.bias
+        return jax.numpy.einsum(self.einsum_str, inputs, self.weight)
+
+
+class JaxLinear(JaxEinsum):
```
jrplatin, line 75: Might be a noob question, but how are calling the forward pass / `__call__` method of `JaxEinsum` here? Is it just handled automatically so that `JaxEinsum.__call__` is called when `JaxLinear` is invoked as a method?

tpu_inference/layers/vllm/quantization/__init__.py
```diff
@@ -13,28 +13,35 @@
...
-from tpu_inference.layers.vllm.quantization.configs import VllmQuantConfig
-from tpu_inference.layers.vllm.quantization.fp8 import VllmFp8Config
-from tpu_inference.layers.vllm.quantization.mxfp4 import VllmMxfp4Config
-from tpu_inference.layers.vllm.quantization.unquantized import \
-    VllmUnquantizedConfig
 
 
 def get_tpu_quantization_config(vllm_config: VllmConfig,
                                 mesh: Mesh) -> QuantizationConfig:
+    from tpu_inference.layers.vllm.quantization.awq import VllmAWQConfig
```
jrplatin: Why are these import changes needed?

tpu_inference/layers/vllm/quantization/configs.py
```diff
@@ -40,7 +40,8 @@
 class VllmQuantLinearConfig:
 
     def __init__(self, vllm_config: VllmConfig, mesh: Mesh, layer: LinearBase):
-        assert isinstance(layer, LinearBase)
+        from tpu_inference.layers.jax.linear import JaxEinsum
```
jrplatin: Is there a reason we are lazily importing?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -77,9 +83,9 @@ def get_config_filenames(cls) -> list[str]:
     def from_config(cls, _: dict[str, Any]) -> "VllmUnquantizedConfig":
         return cls()
 
-    def get_quant_method(self, layer: torch.nn.Module,
-                         prefix: str) -> Optional[QuantizeMethodBase]:
-        if isinstance(layer, LinearBase):
+    def get_quant_method(self, layer: Module,
+                         prefix: str) -> Optional[JaxQuantizeMethodBase]:
```
jrplatin: Should this be `JaxQuantizeMethodBase | QuantizeMethodBase`?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -151,51 +162,78 @@ def apply(self,
...
             out = jnp.einsum("mn,pn->mp", x_jax, weight_jax)
-            if bias is not None and not layer.skip_bias_add:
-                out += jax_view(bias[i])
+            if bias_jax is not None:
+                out += bias_jax[i]
 
             outs.append(out)
         out = jnp.concatenate(outs, axis=-1)
-        return torch_view(out)
+        return out
```
jrplatin: Why are you getting rid of the `torch_view` here? Isn't this method shared with TorchAX?

tpu_inference/layers/jax/linear.py
lk-chen: Yes I believe it's how python inheritance works. If child class (JaxLinear) does not overwrite, the method from base class (JaxEinsum) will be used.

tpu_inference/layers/vllm/quantization/__init__.py
lk-chen: The import cannot locate at top of file, in order to prevent circular dependency

tpu_inference/layers/vllm/quantization/configs.py
lk-chen: The import cannot locate at top of file, in order to prevent circular dependency

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: `JaxQuantizeMethodBase` inherits from vllm's `QuantizeMethodBase`

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: `torch_view` is applied outside of this shared part, specifically on L183.

The shared method should be jax-in jax-out

tpu_inference/layers/jax/linear.py
jrplatin: Ah whoops, totally missed that it was inheriting 🤦

tpu_inference/layers/jax/linear.py
```diff
@@ -0,0 +1,118 @@
...
+# limitations under the License.
+
+from typing import Optional
+
+import jax
+from flax import nnx
+
+from tpu_inference.layers.jax import JaxModule
+from tpu_inference.layers.vllm.quantization import JaxQuantizeMethodBase
+from tpu_inference.layers.vllm.quantization.configs import VllmQuantConfig
```
kyuyeunk: for a module / function that is used by both vllm and jax model implementation, can you move it to "layers/common" folder?

tpu_inference/layers/jax/linear.py
```diff
@@ -0,0 +1,118 @@
...
+            self.quant_method.create_weights_jax(self)
+        else:
+            self.quant_method = None
+
+    def __call__(self, inputs: jax.Array) -> jax.Array:
+        if self.quant_method is not None:
+            return self.quant_method.apply_jax(self, inputs)
+
+        if self.bias is not None:
+            return jax.numpy.einsum(self.einsum_str, inputs,
```
kyuyeunk: i think this looks simpler
```
output = jnp.einsum(self.einsum_str, inputs, self.weight)
if self.bias is not None:
    output += self.bias
return output
```

tpu_inference/layers/jax/__init__.py
```diff
@@ -11,3 +11,49 @@
...
+
+import jax
+from flax import nnx
+
+
+class JaxModule(nnx.Module):
+    """Base module for JAX layers, extending flax.nnx.Module.
+    """
+
+    def register_parameter(self, name: str, value: nnx.Param | None) -> None:
```
kyuyeunk: how does `register_parameter` and `create_weights_jax` going to work together to create weights?

tests/layers/jax/test_linear.py
```diff
@@ -0,0 +1,154 @@
...
+        hidden_size = 16
+        intermediate_size = 32
+        vllm_config = VllmConfig(model_config=ModelConfig(
+            model="Qwen/Qwen3-0.6B"))
+
+        # As of vllm0.12.0, vllm.model_executor.parameter.BasevLLMParameter calls
+        # get_tensor_model_parallel_rank() and
+        # get_tensor_model_parallel_world_size() even though disable_tp=True, which
+        # causes error during initialization. So we mock them here.
+        with mock.patch(
```
kyuyeunk: instead of `mock.patch`, can you just do a real initialization? so something like this? https://github.com/vllm-project/tpu-inference/blob/18ee19d8b765dbc21bf235e8bfce51b5aae530cd/tests/layers/vllm/test_unquantized.py#L64-L72

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -90,11 +96,16 @@ def get_quant_method(self, layer: torch.nn.Module,
...
 
 
-class VllmUnquantizedLinearMethod(UnquantizedLinearMethod):
+class VllmUnquantizedLinearMethod(UnquantizedLinearMethod,
+                                  JaxQuantizedLinearMethod):
 
     def __init__(self, linear_config: VllmQuantLinearConfig):
         self.linear_config = linear_config
 
+    def create_weights_jax(self, layer: JaxModule) -> None:
```
kyuyeunk: not a fan of having code designed for `layers/jax` inside the files of `layers/vllm`. we want to decouple any dependency or reference between `layers/vllm` and `layers/jax`. they might share a design pattern but actual execution code should be separated. only logic that are shared between the two should be residing in `layers/common` folder.

tpu_inference/layers/jax/linear.py
jrplatin: Whoop sorry, missed that this was being inherited ...

jrplatin:
LG from my side, thanks!

tpu_inference/layers/jax/linear.py
lk-chen: I think it's easier to do that in a separate PR https://github.com/vllm-project/tpu-inference/pull/1505

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: Done, moved to /common in another PR

tests/layers/jax/test_linear.py
lk-chen: Done, thar works

tpu_inference/layers/jax/__init__.py
lk-chen: Eventually realized `register_parameter` is unnecessary. I originally was copying from torch

tpu_inference/layers/jax/linear.py
lk-chen: Done

tpu_inference/layers/jax/linear.py
lk-chen: I'll remove `prefix` as it's not used

PR #1487, opened 2026-01-18
Optimize ep implementation

kyuyeunk:
# Description

Optimize MoE expert parallelism performance by removing all reduce.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -211,10 +211,10 @@ def expert_sharded_gmm(
     rhs_scale: jax.Array | None,
     rhs_bias: jax.Array | None,
     group_sizes: jax.Array,
-    is_last_expert: bool,
+    is_last_gmm: bool,
     mesh: Mesh,
 ) -> jax.Array:
-    ep_size = get_mesh_shape_product(mesh, ShardingAxisName.MLP_TENSOR)
+    ep_size = get_mesh_shape_product(mesh, ShardingAxisName.EXPERT)
     ep_p_spec = P(ShardingAxisName.EXPERT)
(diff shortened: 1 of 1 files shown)


jrplatin (approved):

vanbasten23:
@kyuyeunk , could you explain where the all-reduce is and how is it removed by "send_sizes"?

kyuyeunk:
> @kyuyeunk , could you explain where the all-reduce is and how is it removed by "send_sizes"?

here are the results
- before: 
image

- after: 
image

explanation
- `recv_sizes` and `send_sizes` have same values: https://github.com/vllm-project/tpu-inference/blob/0b9e2b42a4d6a1209b43b407e76f36574407f9f4/tpu_inference/layers/common/fused_moe_gmm.py#L293
- however, difference is that `send_sizes` is sharded along ep axis while `recv_sizes` is fully replicated: https://github.com/vllm-project/tpu-inference/blob/0b9e2b42a4d6a1209b43b407e76f36574407f9f4/tpu_inference/layers/common/fused_moe_gmm.py#L357-L360
- note that `group_sizes`- which is what `send_sizes` is computed from) is already fully replicated across all shards
- going from fully replicated to ep sharded is a simple local slicing - which is what's already happening on `send_sizes`
- but the issue is that the compiler performs local slicing to convert fully replicated to ep sharded on `send_sizes` first and then tries to obtain the result of `recv_sizes` - which is fully replicated - by invoking collective operations on ep sharded `send_sizes`.
- by setting a constraint that ensures `send_sizes` (and also all the other ops that was created from `send_sizes`)  will be fully replicated will ensure that local slicing to convert fully replicated to ep sharded will happen at the very last step and not affect `recv_sizes`

PR #1493, opened 2026-01-20
Support pooling-model and embedding task

wdhongtw:
# Description

Support pooling-model for embedding task.

vLLM does support [pooling-model > embed-task](https://docs.vllm.ai/en/stable/models/supported_models/#embedding) for
- type A: models that are pooling-models natively
- type B: in-place conversion of generative-models.

This PR covers both types of usage.

From the perspective of attention type (using vLLM notation)

* encoder-only attention (bidirectional attention)
* decoder attention (causal attention)

This PR covers models that use encoder-only attention, and partially covers decoder attention.
*For decoder attention, it's functional when the prompt is not chunked (token-len for that particular req is smaller than max-num-batched-tokens).*

Note: The encoder-attention mechanism itself is not supported by tpu-inference yet.

~To support encoder-only attention, we reuse splash attention kernel from jax library.
The kernel block size is not tuned yet.~

We implements pooling calculation for pooler model, reusing pooler logic from vLLM model implementation`. Also we refactor and adjust the attention metadata building logics, fixing and adjust the logic for tracking pooling parameter for incoming requests. 

~Also upgrade `torchax` to v0.0.11 address some lacked operations that's used by vLLM BERT model implementation.~

~Note: Previously in this PR we try to use `tokamax` splash attention implementation, but we switch to the implementation from jax librarary, as there is no noticeable performance difference and `tokamax` requires an upgrade of `jaxlib` which introduce too much test breaking to be solved here.~

Co-developed-by: Yuyan Peng <yuyanpeng@google.com>

## TODOs

This PR still in progress, TODOs to be resolved by this PR includes at least:

- [x] ~Resolve package conflicts that prevent we from using `tokamax==0.0.8`.~
- [x] Include minimal unit test for pooler-mode / embedding-task.
- [x] Find better location for PoolerFunc definition
- [x] Fix all test error, if any

## Other Notes

- The attention-metadata building logics should be equivalent to previous ones,
  with just the additional case for no-kv-cache.
- It's quite hard, if not impossible, to apply JAX JIT on pooling calculation.
  Since that some `PoolingMethod` and `PoolingCursor` are highly coupled
  with values(tensors) on CPU.
- Lots of change in `tpu_inference/runner/input_batch.py`, ignore leading space
  change is easier to read.

# Tests

No test included yet, we can manually compare the output from GPU and TPU
to ensure the correctness of this implementation.

```sh
python embed.py # generate embed-output.json
python compare.py "<embed-output-tpu>" "<embed-output-gpu>"
```

```text
raw text length: 24
cosine similarity: 0.9999927391643547
raw content: Hello, my name is Alice.

raw text length: 266
cosine similarity: 0.9999763756061478
raw content: In today's fast-paced world, finding a balance between productivity and mindfulness has become more important than ever. As urban landscapes continue to evolve, people are looking for ways to reconnect with nature without losing the convenience of modern technology.

raw text length: 86
cosine similarity: 0.9999711595060252
raw content: 最近の技術革新により、私たちの日常生活は劇的に変化しました。都市の風景は新旧の建築が入り混じり、静かな朝の光が窓から差し込む中で、人々はそれぞれの目的を持って歩き始めます。

raw text length: 37
cosine similarity: 0.9999864471473087
raw content: 최근 기술의 발전과 함께 우리의 일상에는 많은 변화가 찾아왔습니다.
```

See appendix for the testing scripts.

# Checklist

# Appendix

`embed.py`

```python
import json

from vllm import LLM


def main():
    model = "intfloat/multilingual-e5-large-instruct"
    # model = "Qwen/Qwen3-Embedding-0.6B"  # for type B

    llm = LLM(
        model=model,
        runner="pooling",
        max_num_seqs=64,
        max_num_batched_tokens=1024,
        max_model_len=512,
        dtype="float32",
    )

    inputs = [
        "Hello, my name is Alice.",
        "In today's fast-paced world, finding a balance between productivity "
        "and mindfulness has become more important than ever. As urban "
        "landscapes continue to evolve, people are looking for ways to "
        "reconnect with nature without losing the convenience of modern "
        "technology.",
        "最近の技術革新により、私たちの日常生活は劇的に変化しました。"
        "都市の風景は新旧の建築が入り混じり、静かな朝の光が窓から差し込む中で、"
        "人々はそれぞれの目的を持って歩き始めます。",
        "최근 기술의 발전과 함께 우리의 일상에는 많은 변화가 찾아왔습니다.",
    ]
    results = llm.embed(inputs)
    report = dict(zip(inputs, [r.outputs.embedding for r in results]))

    with open("embed-output.json", "w") as f:
        json.dump(report, f, indent=4)


if __name__ == "__main__":
    main()
```

`compare.py`

```python
import json
import sys

import numpy as np


def cosine_similarity(values1: list[float], values2: list[float]) -> float:
    vec1, vec2 = np.array(values1), np.array(values2)
    assert vec1.shape == vec2.shape

    dot_product = np.dot(vec1, vec2)
    norm_vec1, norm_vec2 = np.linalg.norm(vec1), np.linalg.norm(vec2)
    return dot_product / (norm_vec1 * norm_vec2)


if __name__ == "__main__":
    assert len(sys.argv) == 3

    # report: text-document (str) to embedding (list[float])
    def read(path: str) -> dict[str, list[float]]:
        with open(path) as f:
            return json.load(f)

    report1, report2 = read(sys.argv[1]), read(sys.argv[2])
    for key in report1:
        similarity = cosine_similarity(report1[key], report2[key])
        print(f"raw text length: {len(key)}")
        print(f"cosine similarity: {similarity}")
        print(f"raw content: {key}")
        print()
```

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -59,6 +59,7 @@ from tpu_inference.layers.jax.sample.sampling import (compute_logprobs,
 from tpu_inference.layers.jax.sample.sampling_metadata import \
     TPUSupportedSamplingMetadata
 from tpu_inference.logger import init_logger
+from tpu_inference.models.common.interface import PoolerFunc
 from tpu_inference.models.common.model_loader import get_model
 from tpu_inference.models.jax.jax_intermediate_tensor import \
     JaxIntermediateTensors
@@ -277,6 +278,9 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
 
         self.kv_caches: list[jax.Array] = []
         self.layer_name_to_kvcache_index: dict[str, int] = {}
+
+        self.is_pooling_model: bool = self.model_config.runner_type == "pooling"
+        """Generative model or pooling model select different computations."""
 
     def _init_random(self):
         if self.model_config.seed is None:
@@ -507,7 +511,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                                             dtype=np.int64)
 
     def load_model(self):
-        self.model_fn, self.compute_logits_fn, self.combine_hidden_states_fn, multimodal_fns, self.state, self.lora_manager, self.model = get_model(
+        self.model_fn, self.compute_logits_fn, self.pooler_fn, self.combine_hidden_states_fn, multimodal_fns, self.state, self.lora_manager, self.model = get_model(
             self.vllm_config,
             self.rng_key,
             self.mesh,
@@ -542,7 +546,12 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                     f"hbm={common_utils.hbm_usage_gb(self.devices)}GiB")
 
     def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
-        return ("generate", )
+        runner_type = self.model_config.runner_type
+        if runner_type == "generate":
+            return ("generate", )
+        if runner_type == "pooling":
+            return ("embed", )
+        assert False, f"unsupported runner type: {runner_type}"
 
     def get_kv_cache_spec(self):
         return self.kv_cache_manager.get_kv_cache_spec()
@@ -794,6 +803,27 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                 assert isinstance(hidden_states, JaxIntermediateTensors)
                 hidden_states.kv_connector_output = kv_connector_output
                 return hidden_states
+
+            if self.is_pooling_model:
+                seq_lens = self.seq_lens_cpu[:self.input_batch.num_reqs]
+                pooling_metadata = self.input_batch.get_pooling_metadata()
+
+                pooler_fn: PoolerFunc = self.pooler_fn
+                pooler_output = pooler_fn(
+                    hidden_states,
+                    pooling_metadata,
+                    seq_lens,
+                )
+
+                return ModelRunnerOutput(
+                    req_ids=self.input_batch.req_ids,
+                    req_id_to_index=self.input_batch.req_id_to_index,
+                    sampled_token_ids=[],
+                    logprobs=None,
+                    prompt_logprobs_dict={},
+                    pooler_output=pooler_output,
+                )
+
             hidden_states = self._select_from_array_fn(hidden_states,
                                                        logits_indices)
             logits = self.compute_logits_fn(
@@ -1416,10 +1446,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
              sharding=data_parallel_attn_sharding,
          )
 
-        attention_metadata_per_layer: Dict[str, AttentionMetadata] = {}
-        uniform_attention_metadata: AttentionMetadata = None
-        for kv_cache_gid, kv_cache_group in enumerate(
-                self.kv_cache_config.kv_cache_groups):
+        def build_block_table(kv_cache_gid: int) -> jax.Array:
             block_tables = self.block_tables_cpu[kv_cache_gid][:self.
                                                                max_num_reqs]
             for dp_rank in range(dp_size):
@@ -1437,7 +1464,9 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                 (block_tables),
                 sharding=data_parallel_attn_sharding,
             )
+            return block_tables
 
+        def build_attn(block_tables: jax.Array | None) -> AttentionMetadata:
             attention_metadata_gid = AttentionMetadata(
                 input_positions=positions,
                 block_tables=block_tables,
@@ -1449,13 +1478,21 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
             # This is for making these cpu buffers hidden during tracing
             attention_metadata_gid.query_start_loc_cpu = query_start_loc_cpu
             attention_metadata_gid.seq_lens_cpu = seq_lens_cpu
+            return attention_metadata_gid
 
-            if not self.use_hybrid_kvcache:
-                uniform_attention_metadata = attention_metadata_gid
-            else:
-                for layer_name in kv_cache_group.layer_names:
-                    attention_metadata_per_layer[
-                        layer_name] = attention_metadata_gid
+        attention_metadata: AttentionMetadata | dict[str, AttentionMetadata]
+        if len(self.kv_cache_config.kv_cache_groups) <= 1:
+            # Pooling model will not using kv cache
+            no_kv_cache = len(self.kv_cache_config.kv_cache_groups) == 0
+            block_tables = build_block_table(0) if not no_kv_cache else None
+            attention_metadata = build_attn(block_tables)
+        else:
+            attention_metadata = {
+                name: build_attn(build_block_table(gid))
+                for gid, kv_cache_group in enumerate(
+                    self.kv_cache_config.kv_cache_groups)
+                for name in kv_cache_group.layer_names
+            }
 
         # Async scheduling: substitute placeholder tokens for DP
         if self.scheduler_config.async_scheduling and self._pre_async_results is not None:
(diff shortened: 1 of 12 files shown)


requirements.txt
```diff
@@ -5,16 +5,15 @@ pytest-mock
...
-qwix==0.1.1
+qwix==0.1.2
 torchvision==0.24.0
 pathwaysutils
 parameterized
 numba==0.62.1
 runai-model-streamer[s3,gcs]==0.15.0
-tokamax==0.0.4
-immutabledict==4.1.0 # indirect dependency of tokamax
+tokamax==0.0.8
```
kyuyeunk: i wasn't aware that we had tokamax as dependency. and the quick search shows that it's not used anywhere. can we just delete it?

cc: @QiliangCui

wdhongtw: @kyuyeunk 

It's a new dependency I added here in this PR, to enable encoder-only attention functionality to `tpu-inference` project.

The usage is in tpu_inference/layers/vllm/attention.py.
In that file, Yuyan Peng, co-developed me, create an adapter, implementing the required `vllm.v1.attention.backend.AttentionImpl` interface and forward the actual calculation into `tokamax` library.


I would like to use `tokamax` here because it allows us to leverage the GDM team's work, which should be stable and mature for these kinds of applications.

Let me know if you have any questions or concerns. 😳

kyuyeunk: Ah got it. Thanks for the explanation! Will take a look at the pr when I'm available!

wdhongtw:
All test passed, should be good for review. 🙂‍↕️

tpu_inference/layers/vllm/attention.py
```diff
@@ -100,6 +114,286 @@ def get_page_size(vllm_config: VllmConfig) -> int:
         return page_size
 
 
+@final
+class AttentionDispatcher(AttentionImpl[AttentionMetadata]):
```
kyuyeunk: in vllm, there's a concept of `AttentionSelectorConfig`: https://github.com/vllm-project/vllm/blob/e69c990c216c623b1de22f055926602a336f9352/vllm/v1/attention/selector.py#L21-L43

which used like this to select which attention to use:
- https://github.com/vllm-project/vllm/blob/e69c990c216c623b1de22f055926602a336f9352/vllm/platforms/xpu.py#L42-L77
- https://github.com/vllm-project/vllm/blob/e69c990c216c623b1de22f055926602a336f9352/vllm/platforms/cuda.py#L303-L371


on tpu-inference, you can add a code in this function: https://github.com/vllm-project/tpu-inference/blob/fa60b3b8798b46d78b8b7242a99de0059ca266f1/tpu_inference/platforms/tpu_platform.py#L57-L59

is `AttentionDispatcher` achieving same functionality as `AttentionSelect`? if so, maybe use that instead?

tpu_inference/layers/vllm/attention.py
```diff
@@ -100,6 +114,286 @@ def get_page_size(vllm_config: VllmConfig) -> int:
...
+
+    def process_weights_after_loading(self, act_dtype: torch.dtype):
+        # process_weights_after_loading is not part of AttentionImpl[T]
+        # but vLLM does detect it's existence then invoke it.
+        if hasattr(self._attn, "process_weights_after_loading"):
+            self._attn.process_weights_after_loading(act_dtype)
+
+
+@final
+class TokamaxAttentionBackendImpl(AttentionImpl[AttentionMetadata]):
```
kyuyeunk: please separate file for separate attention implementation.

tpu_inference/layers/vllm/attention.py
wdhongtw: Yes I should use it, let me change `TpuPlatform` quickly to address this.

tpu_inference/layers/vllm/attention.py
wdhongtw: No problem, is there any suggestion on the module path/name?

Is `tpu_inference.layers.vllm.attention_jax.py` a good choice here? 😳

tpu_inference/layers/vllm/attention.py
```diff
@@ -285,3 +289,288 @@ def _jax_attn_func(
     outputs = outputs.reshape(q_len, q_compute_dim)
 
     return new_kv_cache, outputs
+
+
+class JaxAttentionBackend(AttentionBackend):
```
kyuyeunk: The name `JaxAttentionBackend` isn't really descriptive. it only refers implementation method but not really a functionality. Also, i think it's simply outright wrong when considering that this implementation also likely utilize Pallas? meaning it's even harder to differentiate `PallasAttentionBackend`. Please use the name like "mla attention", "multi head attention", etc to describe its functionality. (note that i'm also not a fan of `PallasAttentionBackend` as well. if you have a better name for that as well, feel free to change.)

tpu_inference/layers/vllm/attention.py
```diff
@@ -285,3 +289,288 @@ def _jax_attn_func(
...
+
+    @override
+    @staticmethod
+    def get_name() -> str:
+        return "FLASH_ATTN"
```
kyuyeunk: we have already registered `"FLASH_ATTN"` with `PallasAttentionBackend`. Please use a different name.

tpu_inference/platforms/tpu_platform.py
```diff
@@ -59,6 +60,12 @@ def get_attn_backend_cls(cls, selected_backend: "AttentionBackendEnum",
                              **kwargs) -> str:
         from vllm.v1.attention.backends.registry import AttentionBackendEnum
 
+        # For encoder-only attention, only JaxAttentionBackend works for now.
+        if attn_selector_config.attn_type == AttentionType.ENCODER_ONLY:
+            import tpu_inference.layers.vllm.attention as attention
+            cls = attention.JaxAttentionBackend
```
kyuyeunk: instead of this approach, can you use `selected_backend` approach that gets it class path using `selected_backend.get_path()`?

Here is api for registering attention: https://github.com/vllm-project/vllm/blob/10546f925aef5d805e41f0f40c6610d08ff1a037/vllm/v1/attention/backends/registry.py#L206

And here is how we register our existing attention: https://github.com/vllm-project/tpu-inference/blob/4e6e6fb46870b92dafc35ccec587fa2dc5ca4bf7/tpu_inference/layers/vllm/attention.py#L33

tpu_inference/layers/vllm/attention.py
kyuyeunk: i don't think that's a good name. please refer to my similar comment here: https://github.com/vllm-project/tpu-inference/pull/1493#discussion_r2757361618

tpu_inference/platforms/tpu_platform.py
wdhongtw: No problem, will do. 👍

tpu_inference/layers/vllm/attention.py
wdhongtw: I would like to reuse FLASH_ATTN as splash attention should also classified in this group?

Also we can not use arbitrary name as it's a contract from `Attention` class in `vllm/model_executor/layers/attention/attention.py`

```python
self.backend = AttentionBackendEnum[self.attn_backend.get_name()]
```

With current design of `TpuPlatform.get_attn_backend_cls`, we could trigger
registration of different class such that we can reuse the name.

tpu_inference/layers/vllm/attention.py
wdhongtw: Thanks for the suggestion, the name should reflect the ability it provides.

wdhongtw:
Per discussion, need more adjustment on attention kernel implementation. Convert back to draft.

wdhongtw:
Test should be good now, in last CI run all check passed: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/9753/steps/canvas .

Just encounter chex library missing issue in last triggered CI job.

wdhongtw:
Rebased again to leverage main branch change that fix CI job failure issue.

wdhongtw:
Rebased again to address merge conflicts that introduced by recent PRs.

tpu_inference/runner/tpu_runner.py
```diff
@@ -530,7 +534,12 @@ def load_model(self):
                     f"hbm={common_utils.hbm_usage_gb(self.devices)}GiB")
 
     def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
-        return ("generate", )
+        runner_type = self.model_config.runner_type
+        if runner_type == "generate":
+            return ("generate", )
+        if runner_type == "pooling":
+            return ("embed", )
+        assert False, f"unsupported runner type: {runner_type}"
```
lk-chen: raise Error

vipannalla:
Looks good to me.

PR #1505, opened 2026-01-21
[Refactoring] Move some logic from layers/vllm/quantization to layers/common/quantization

lk-chen:
# Description

Prepare for #1482 which will reuse the logic in layers/common

# Tests

CI

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -160,8 +160,8 @@ class VllmUnquantizedLinearMethod(vllm_linear.UnquantizedLinearMethod,
             else:
                 assert isinstance(layer.weight, torch.nn.ParameterList)
 
-                out_jax = self._apply_split(
-                    x_jax, [jax_view(w) for w in layer.weight], bias_jax)
+                out_jax = self._apply_split(x_jax, jax_view(layer.weight),
+                                            bias_jax)
                 out: torch.Tensor = torch_view(out_jax)
 
             if out_sharding := self.linear_config.get_output_sharding(out):


tpu_inference/layers/common/quantization/unquantized.py
```diff
@@ -0,0 +1,62 @@
...
+
+import jax
+from jax import numpy as jnp
+
+from tpu_inference.layers.common.quantization.configs import QuantLinearConfig
+from tpu_inference.layers.common.utils import \
+    slice_sharded_tensor_for_concatenation
+
+
+class UnquantizedLinearMethod:
```
kyuyeunk, line 25: qq: do you plan to apply a similar design for all the other quant methods like `fp8`, `mxfp4`, etc?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -15,35 +15,32 @@
 from typing import Any, Optional
 
 import jax
-import jax.numpy as jnp
 import torch
+import vllm.model_executor.layers.linear as vllm_linear
```
kyuyeunk, line 19: I think it's better to have a consistent syntax - which is to only import functions that is used in here instead of an entire module. wdyt?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -15,35 +15,32 @@
...
     register_quantization_config
 from vllm.model_executor.layers.quantization.base_config import (
     QuantizationConfig, QuantizeMethodBase)
 
+from tpu_inference.layers.common.process_weights.linear_weights import (
+    LinearWeights, process_linear_weights, shard_linear_weights,
+    to_parameter_list)
 from tpu_inference.layers.common.quant_methods import (UNQUANTIZED,
                                                        get_tpu_quant_method)
+from tpu_inference.layers.common.quantization import unquantized as jax_common
```
kyuyeunk: nit: rename to `unquantized_common` or `common_unquantized`. can be confused with files in jax folder.

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -145,58 +143,36 @@ def apply(self,
...
+                bias) if bias is not None and not layer.skip_bias_add else None
             if self.linear_config.fuse_matmuls:
-                out = self._apply_fused(layer, x, bias)
+                weight_jax = jax_view(layer.weight)
+                out_jax = self._apply_fused(x_jax, weight_jax, bias_jax)
+                out: torch.Tensor = torch_view(out_jax)
             else:
-                out = self._apply_split(layer, x, bias)
+
+                def weight_enumerate():
```
kyuyeunk: feel like using generator / yield is unnecessarily complex? list would look much simpler?

tpu_inference/layers/common/quantization/unquantized.py
lk-chen: yes, fp8.py etc. will be added in layers/common/ later

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: This is to avoid name conflict.

I guess you are thinking of
```
from ... import UnquantizedLinearMethod as VllmUnquantizedLinearMethod
```
but later it will conflict with line 90.

I tried all other means, and they are not readable as this. So this is the best I can see.

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: SG, updated

tpu_inference/layers/vllm/quantization/unquantized.py
lk-chen: Done, updated

tpu_inference/layers/vllm/quantization/unquantized.py
kyuyeunk: got it. thanks for the explanation.

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -145,58 +144,32 @@ def apply(self,
...
-                out = self._apply_fused(layer, x, bias)
+                weight_jax = jax_view(layer.weight)
+                out_jax = self._apply_fused(x_jax, weight_jax, bias_jax)
+                out: torch.Tensor = torch_view(out_jax)
             else:
-                out = self._apply_split(layer, x, bias)
+                assert isinstance(layer.weight, torch.nn.ParameterList)
+
+                out_jax = self._apply_split(
+                    x_jax, [jax_view(w) for w in layer.weight], bias_jax)
```
kyuyeunk: `jax_view` supports tree structure. so you can just do `jax_view(layer.weight)`

lk-chen: Updated, thx for the tip

PR #1508, opened 2026-01-22
configure isort to match yapf style

patemotter:
# Description

I had a specific series of imports in a file that formed an endless cycle during pre-commit of sparse_moe.py being changed based on `isort` and then would be changed back to the original format in the next phase by `yapf`. 

This change gets the two formatting tools to play more nicely with each other.


pyproject.toml
```diff
@@ -11,3 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+[tool.isort]
+profile = "google"
```
wdhongtw: nit: maybe a new line here. :D

jrplatin: Added, thanks!

PR #1515, opened 2026-01-22
Add perf regression test for qwen3 coder

vanbasten23:
# Description

This PR adds perf regression tests for qwen3 coder model

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -273,7 +273,6 @@ steps:
            exit 0
          fi
 
-   # TODO(xw32): also check if it is nightly before merging.
    - label: "${LABEL_PREFIX}Perf regression test for qwen3 coder 1k 8k."
      key: ${KEY_PREFIX}test_20
      soft_fail: true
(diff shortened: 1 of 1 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -428,18 +429,26 @@ def _precompile_select_from_array(self) -> None:
             index_paddings = self.runner.num_logits_paddings
         else:
             index_paddings = self.runner.num_reqs_paddings
-        dp_sharding = NamedSharding(self.runner.mesh,
-                                    PartitionSpec(ShardingAxisName.ATTN_DATA))
-        hidden_states_sharding = NamedSharding(
-            self.runner.mesh, PartitionSpec(ShardingAxisName.ATTN_DATA, None))
         dp_size = self.runner.vllm_config.sharding_config.total_dp_size
+        if dp_size > 1:
```
QiliangCui: Is this expected change?

wenxindongwork: @vanbasten23 could you please help me understand the conditional sharding applying to `hidden_states`? It is expected that `indices_sharding` should carry different sharding in the dp vs non-dp case due to the branching logic in `prepare_input` (and this reminds me that I really should refactor the `prepare_input` function to unify those branches). Where is the branching for the sharding on `hidden_states`?

vanbasten23: > Is this expected change?

Yes.

> @vanbasten23 could you please help me understand th...

Here the change is needed to fix the recompilation issue for EP case.

Not sure if I follow your question. Here, 
if dp>1:
    # we shard hidden_states as P('data')
    # we shard indices_sharding as P('data')
    # this is the same as before the change.
elif ep case and dp==1:
    # we shard hidden_states as P('model') because that's what I see in execute_model after the precompilation finishes.
    # we shard indices_sharding as P(None)
else:
    # dp==1 case,
    # we shard hidden_states as P(None)
    # we shard indices_sharding as P(None)
Does it answer your question?

vanbasten23: Will address the recomiplation in a follow-up pr.

PR #1516, opened 2026-01-22
Better moe performance by topk reduction first and psum in fused_moe_gmm

helloworld1:
# Description

Perform local reverse sort and reduction on the topk experts of a token.
The psum size now will be 1/top_k compared with psum first and run reduction later.

Benchmark

seq_len = 8192
num_tokens = B * seq_len
hidden_size = 6144
intermediate_size = 2560
num_experts = 160
topk = 8

Before: Step 15.4ms, psum 9.2ms
After: Step 7.3ms, psum 1.1ms

# Tests
vllm chat to confirm the basic E2E responses. Colab for the layer to compare the output with original code with no major deviation.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -200,7 +200,7 @@ def tensor_sharded_gmm_row_parallel(
         )
 
         # Perform local reverse sort and reduction on the topk experts of
-        # a token. This will save psum size by 1/8 compared with psum first
+        # a token. Psum size will be 1/top_k compared with psum first
         # and run reduction later
         out = out[topk_argsort_revert_indices_local].reshape((-1, topk, n))
         out = out * jnp.expand_dims(topk_weights_local, axis=-1)


tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -185,6 +197,14 @@ def _gmm_all_reduce(lhs, rhs, rhs_scale, rhs_bias, group_sizes):
             tiling=(tm, tk, tn),
             group_offset=jnp.array(0),
         )
+
+        # Perform local reverse sort and reduction on the topk experts of
+        # a token. This will save psum size by 1/8 compared with psum first
```
vanbasten23: Can you explain to me why "This will save psum size by 1/8 compared with psum first and run reduction later" ?

helloworld1: The comment is incorrectly copied from benchmark. Thanks for pointing out. I have updated the comment to be psum size will be 1/top_k compared with ...

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -186,6 +198,14 @@ def _gmm_all_reduce(lhs, rhs, rhs_scale, rhs_bias, group_sizes):
             tiling=(tm, tk, tn),
             group_offset=jnp.array(0),
         )
+
+        # Perform local reverse sort and reduction on the topk experts of
+        # a token. Psum size will be 1/top_k compared with psum first
+        # and run reduction later
+        out = out[topk_argsort_revert_indices_local].reshape((-1, topk, n))
+        out = out * jnp.expand_dims(topk_weights_local, axis=-1)
+        out = out.sum(axis=-2)
```
vanbasten23, line 207: For the same technique, can it be applied to gmm ep?

vanbasten23:
Also, just want to make sure there is a accuracy test covering this case. Other than that, the pr lgtm.

tpu_inference/layers/common/fused_moe_gmm.py
helloworld1: Yes, can be applied there too. I will experiment it a bit there. The benefit is not as significant as TP though.

In the case of EP,
If we assume uniform distribution of tokens to experts
The original EP per rank communication sizes is (total_tokens * top_k / ep_size)
If local reduction is applied, the communication sizes tokens_has_experts_for_a_rank, which can be estimated as total_tokens * (1 - (1 - 1 / ep_size) ^ top_k).

So for qwen coder, the communication size for one rank:
original: 8192 * 8 / 8 = 8192 tokens
local reduction 8192 * (1 - (1 - 1/8) ^ 8) = 8192 * 0.656 = 5376 tokens.
The communication size is only reduced by about 35% in this case.

helloworld1:
> Also, just want to make sure there is a accuracy test covering this case. Other than that, the pr lgtm.

The function is covered in various moe quantization tests:
For example
https://github.com/vllm-project/tpu-inference/blob/d07cca959b3d3d5e52866d1f9de1c901dea0eb2f/tests/layers/vllm/test_fp8.py#L398

tpu_inference/layers/common/fused_moe_gmm.py
vanbasten23: In the case of EP, I assume you'd do something similar as "out.sum(axis=-2)" along the "topk" dim. Then each ep rank communication size becomes "total_tokens/ep_size". So I don't follow why you said "total_tokens * (1 - (1 / ep_size) ^ top_k)". Could you explain more?

helloworld1: Yes, it is basically that. But since EP does not communicate on all token*expert, each shard only need to communicate on token*expert owned by that shard.  We reduce from token*expert to tokens_owned_by_one_shard. And tokens_owned_by_one_shard is not equal to total_tokens / ep_size.

For example ep_size = 2, experts = 4, total_tokens =  8 top_k=2
token  experts  
0   [0, 1]
1   [1, 2]
2   [2, 3]
3   [0, 1]
4   [1, 2] 
5   [2, 3]
6   [0, 1]
7   [1.2]

Then break down by shards
shard = 0, owns expert 0 and 1, and  has tokens
0, 1, 3, 4, 6, 7
total 6 tokens

shard = 1 owns expert 2 and 3, and has tokens
1, 2, 4, 5, 7
total 5 tokens.

So shard = 0 needs to communicate on 6 tokens after local reduction and shard = 1 communicate on 5 tokens.
They are not total_tokens / ep_size = 8 / 2 = 4 tokens. It is more than that normally.

vanbasten23: How did you come up with `total_tokens * (1 - (1 / ep_size) ^ top_k)` then?

helloworld1: That's the number of tokens owned by each shard (Tokens has expert assigned to that shard).
To give a concrete example, ep_size = 8, top_k = 4, for a single token, it has about (1 - 1/8) = 7/8 probability that is NOT selected for a specific EP rank, say rank=0, and since each token has 4 experts (top_k = 4), the probability of all 4 experts NOT owned by the rank is about (7/8) ^ 4. So the probably of any of the 4 experts falls into the current rank=0 is 1 - (7/8)^4. This is how the formula for the probability of one rank owning a token  `1 - (1 - 1/ ep_size) ^ top_k)` coming from.

PR #1520, opened 2026-01-23
[Parallelism Support Matrix Tests] add e2e expert parallelism test

syhuang22:
# Description

This PR adds an **End-to-End (E2E) correctness test for Expert Parallelism (EP)** on TPUs using the `Qwen/Qwen1.5-MoE-A2.7B` model. It verifies that MoE models running with Expert Parallelism produce correct outputs compared to a single-chip baseline.

### Context & Solution
Previously, the `parallelism_support_matrix.csv` listed EP as "unverified" for correctness. This PR fills that gap by implementing a rigorous test suite similar to the existing Pipeline Parallelism (PP) tests.

**Implementation Details:**
- **Test File**: `tests/e2e/test_expert_parallel.py`
- **Methodology**:
  - **Baseline**: Runs `TP=1, EP=1` (Single Chip) to establish ground truth.
  - **Experiment**: Runs `TP=4, EP=4` (4 Chips, Expert Parallel enabled).
  - **Verification**: Compares text output (>90% match) and log probabilities (<1.0 diff) using greedy sampling.
- **Model Choice**: `Qwen/Qwen1.5-MoE-A2.7B` (60 experts) was chosen because it fits comfortably on 4x TPU chips and its expert count (60) is evenly divisible by 4 (15 experts/chip), whereas EP=8 would fail divisibility checks.

### Updates
- Updated `support_matrices/parallelism_support_matrix.csv` to mark EP Correctness as `✅`.

# Tests

Tested on **TPU v4/v7 (4 chips)**.

**Instructions to reproduce:**
```bash
# run via pytest
pytest -vv -s tests/e2e/test_expert_parallel.py

# Result 
tests/e2e/test_expert_parallel.py::test_expert_parallelism_correctness
[Worker] Starting Baseline (TP=1)...
[Worker] Starting EP=4 with GMM Kernel...
✓ Correctness test results:
  Text: 10 matches, 0 mismatches
  Max logprob difference: 2.636719e-01
  Significant logprob mismatches (>1e-3): 0
PASSED
tests/e2e/test_expert_parallel.py::test_expert_parallelism_fused_correctness
[Worker] Starting EP=4 with Fused MoE Kernel...
✓ Correctness test results:
  Text: 9 matches, 1 mismatches
  Max logprob difference: 2.412109e-01
  Significant logprob mismatches (>1e-3): 170
PASSED

### DIFF
--- a/.buildkite/parallelism/EP.yml
+++ b/.buildkite/parallelism/EP.yml
@@ -22,7 +22,8 @@ steps:
       queue: tpu_v6e_queue
     commands:
       - |
-        buildkite-agent meta-data set "EP_CorrectnessTest" "unverified"
+        .buildkite/scripts/run_in_docker.sh \
+          bash -c 'python3 -m pytest -s -v -x /workspace/tpu_inference/tests/e2e/test_expert_parallel.py'
   - label: "Record correctness test result for EP"
     key: "record_EP_CorrectnessTest"
     depends_on: "EP_CorrectnessTest"

### CODE REVIEW

**Comment by syhuang22** [comment]
I tried merging main, but it introduced a regression (EngineDeadError / KeyError) in the V1 engine for MoE models on TPU. I kept this branch on the stable base to ensure the test passes while the upstream issue is resolved.
```
File "vllm/v1/core/sched/scheduler.py", line 1148, in update_from_output
    req_index = model_runner_output.req_id_to_index[req_id]
KeyError: '0-9f35c9d7'

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,180 @@
...
+import os
+import json
+import sys
+import subprocess
+import pytest
+from dataclasses import asdict
+
+# Force V0 engine to avoid V1 regression
+import os
+os.environ['VLLM_USE_V1'] = '0'
```
vanbasten23: why do you need this?

vanbasten23: v0 has been deprecated

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,180 @@
...
+    gmm_results = run_inference_subprocess(parallel_size=4, enable_ep=True, use_fused_moe=False)
+    compare_results(baseline_results, gmm_results, "EP (GMM Kernel)")
+
+@pytest.mark.xfail(reason="Fused MoE Kernel currently has accuracy issues on TPU")
+def test_ep_fused_kernel(baseline_results):
+    """Test Expert Parallelism with Fused MoE Kernel."""
+    fused_results = run_inference_subprocess(parallel_size=4, enable_ep=True, use_fused_moe=True)
+    compare_results(baseline_results, fused_results, "EP (Fused Kernel)")
+
+if __name__ == '__main__':
```
vanbasten23: Since it is following the PP and other examples, I'd suggest to be consistent with other test files. You probably don't need this.

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,180 @@
...
+    env = os.environ.copy()
+    
+    if use_fused_moe:
+        env['USE_MOE_EP_KERNEL'] = '1'
+    else:
+        # Important: Unset it if not fused
+        if 'USE_MOE_EP_KERNEL' in env:
+            del env['USE_MOE_EP_KERNEL']
+
+    cmd = [
```
vanbasten23: Suggest to be consistent with other test files instead of using "subprocess"

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,180 @@
...
+
+# Force V0 engine to avoid V1 regression
+import os
+os.environ['VLLM_USE_V1'] = '0'
+
+# Worker function to run inference (executed when script is run as main)
+def run_inference_worker(parallel_size: int, enable_ep: bool):
+    try:
+        from vllm import LLM, EngineArgs, SamplingParams
+    except ImportError:
```
vanbasten23: You may want to update tpu-inference/.buildkite/parallelism/EP.yml as well.

tests/e2e/test_expert_parallel.py
syhuang22: Will restructure to maintain consistency.

tests/e2e/test_expert_parallel.py
syhuang22: Will restructure to maintain consistency.

tests/e2e/test_expert_parallel.py
syhuang22: I think there were some regression issue back when using v1, ack this suggestion and removed this line.

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,213 @@
...
+                    zip(baseline_logprobs, experiment_logprobs)):
+                # vLLM returns [{token_id: Logprob}, ...]
+                if not base_lp_dict or not exp_lp_dict:
+                    continue
+
+                # Get the top token's logprob from each (assuming top_logprobs=1)
+                t1, lp1 = next(iter(base_lp_dict.items()))
+                t2, lp2 = next(iter(exp_lp_dict.items()))
+
+                if t1 == t2:
```
vanbasten23, line 136: Curious if t1, t2 means the token_idx?

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,213 @@
...
+        test_prompts=test_prompts,
+        sampling_params=sampling_params,
+        tensor_parallel_size=4,
+        enable_expert_parallel=True,
+    )
+
+    _verify_correctness(baseline_outputs, ep_outputs, label="Expert Parallel")
+
+
+@pytest.mark.xfail(reason="Fused MoE accuracy issues on TPU")
```
vanbasten23: what's the failure?

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,213 @@
...
+
+    text_match_rate = text_matches / len(baseline_outputs)
+    assert text_match_rate >= 0.9, f"Text match rate {text_match_rate:.2%} is too low"
+
+    # Log probabilities should be very close (allow small numerical errors)
+    # Raising tolerance slightly for Fused kernels if needed, but keeping strict for now
+    assert max_logprob_diff < 1, f"Max logprob difference {max_logprob_diff} is too large"
+
+
+def test_expert_parallelism_correctness(
```
vanbasten23: suggest to rename it to `test_expert_parallelism_correctness_via_gmm_kernel`

tests/e2e/test_expert_parallel.py
```diff
@@ -0,0 +1,213 @@
...
+        sampling_params=sampling_params,
+        tensor_parallel_size=4,
+        enable_expert_parallel=True,
+    )
+
+    _verify_correctness(baseline_outputs, ep_outputs, label="Expert Parallel")
+
+
+@pytest.mark.xfail(reason="Fused MoE accuracy issues on TPU")
+def test_expert_parallelism_fused_correctness(
```
vanbasten23: suggest to rename it to `test_expert_parallelism_correctness_via_fused_moe_kernel`

tests/e2e/test_expert_parallel.py
syhuang22: These are token id. base_lp_dict is a dictionary like {token_id: LogprobObject}

tests/e2e/test_expert_parallel.py
syhuang22: 8/10 as supposed to be text_match_rate >= 0.9

vanbasten23: maybe print out the output and see if it is sensible. If the difference is like https://github.com/vllm-project/tpu-inference/blob/48aa18c90c86bc8ef9f1ca527d66267e2678f0ff/tests/e2e/test_pipeline_parallel.py#L195-L196, then you can adjust the text_match_rate

syhuang22: Mismatch 1 (Prompt 5) has minor difference 

Baseline: "...do they all play at the same time or do **some**"
EP Fused: "...do they all play at the same time or do **they**" 

Mismatch 2 (Prompt 6): Look like a real real issue here. 

Baseline: "**PoseidonHuman**: Can you tell me more about Poseidon's role..."
EP Fused: "**ApolloHuman**: Can you tell me more about the god of the sea..." 
____
I can relax the text_match_rate to 0.8 for prompt 5, any thoughts on this?

vanbasten23: Sure

syhuang22: The Fused kernel test will still FAIL on logprobs (since the difference is ~1.76)

syhuang22: Current status passed with
[Worker] Starting EP=4 with Fused MoE Kernel...
✓ Correctness test results:
  Text: 9 matches, 1 mismatches
  Max logprob difference: 2.412109e-01
  Significant logprob mismatches (>1e-3): 170
PASSED

Issue #1431, opened 2026-01-09
[Feature]: Add support to vLLM tpu-inference library for Qwen3-VL-30B-A3B model


muskansh-google:
### 🚀 The feature, motivation and pitch

vLLM tpu-inference library is missing support for serving and execution of Qwen3-VL-30B-A3B model

### Additional context

Similar feature was added in https://github.com/vllm-project/tpu-inference/pull/1374

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

PR #1374, opened 2025-12-29
[Feat] Qwen3VL dense

HyperBlaze456:
# Description

## Add Qwen 3 VL dense model support

Followed the huggingface style pytorch implementation.

Few files outside of qwen3_vl.py was changed. Explanation:

### `tpu_inference/runner/multimodal_manager.py`:
1. Added pixel_values_videos list concat
```py
  # Handle video pixel values similarly
  if "pixel_values_videos" in batched_mm_inputs and isinstance(
          batched_mm_inputs["pixel_values_videos"], list):
      batched_mm_inputs["pixel_values_videos"] = torch.cat(
          batched_mm_inputs["pixel_values_videos"], dim=0)
```
Qwen3VL supports both modalities with arbitary # of them.
 
2. Updated imports to use JAX compatible scatter/gather func
PyTorch based scatter_mm_placeholders and gather_mm_placeholders caused an error.
JAX-compatible version was added thus.

### `tpu_inference/models/jax/utils/multi_modal_utils.py`
Added JAX-compatible versions of vLLM's placeholder scatter/gather utilities

scatter_mm_placeholders and gather_mm_placeholders use PyTorch tensor operations (new_full, boolean indexing assignment) that fail on JAX arrays. Additionally, vLLM passes is_embed as a torch.Tensor.

  Added:
  - _to_jax_array(): Converts torch.Tensor and probably numpy arrays to JAX array.
  - scatter_mm_placeholders(): Creates placeholder tensor with NaN at non-embedding positions
  - gather_mm_placeholders(): Extracts embeddings from masked positions

---

FIXES: #1330
https://github.com/vllm-project/tpu-inference/issues/1330

# Tests

Aside from specific unit tests, serving was also tested.

start serve:
```sh
sudo docker exec -it ray-head bash -lc "vllm serve 'Qwen/Qwen3-VL-8B-Thinking' \
    --tensor-parallel-size 16 \
    --max-model-len 16384 \
    --max-num-batched-tokens 32768 \
    --disable-frontend-multiprocessing \
	  --log-error-stack \
    --limit-mm-per-prompt '{\"image\":1,\"video\":1}' \
    --trust-remote-code"
```

test three types of inputs:
```sh
curl http://localhost:8000/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
      "model": "Qwen/Qwen3-VL-8B-Thinking",
      "messages": [
        {"role": "user", "content": "Say hello world, but in a more polite manner."}
      ],
[...]

PR #1522, opened 2026-01-23
Qwen3 vl moe

### LINKED ISSUE #1330: [Feature]: Support for Qwen3-VL-4B and 8B

Do we have the support team for Qwen3-VL

or

When can i expect this?

### Alternatives

_No response_

### Additional context

_No response_

HyperBlaze456:
# Description

Add Qwen3VL Dense and Qwen3VL MoE support

The newly added model definition file and model_loader.py's change are self explanatory.
  - tpu_inference/layers/jax/rope_interface.py
    Adds strict validation for M‑RoPE configs (rope_scaling['mrope_section']) when positions are (3, seq_len). Qwen3‑VL uses M‑RoPE and this prevents silent misconfiguration by enforcing presence, length, non‑negative values, and head‑dim bounds.                                                                                            
  - tpu_inference/models/jax/utils/multi_modal_utils.py                                                                                                                                                                                                          
    Adds JAX‑native scatter_mm_placeholders / gather_mm_placeholders (plus _to_jax_array) so multimodal placeholder packing/unpacking works without the torch‑only vllm.v1.worker.utils dependency. This is required by the JAX multimodal pipeline and used by the manager/cache below.
  - tpu_inference/runner/multimodal_manager.py
    Refactors multimodal handling to support Qwen3‑VL’s inputs/outputs:
      - Normalizes image_grid_thw / video_grid_thw from list/tuple/torch/numpy and flattens (B, N, 3) into (B*N, 3) for JIT‑friendly hashing.
      - Handles pixel_values_videos and video_grid_thw (Qwen3‑VL supports video).
      - Switches to the new JAX placeholder helpers.
      - Accepts encoder outputs as { "embeds": ..., "deepstack": ... } and caches DeepStack features per request.
      - gather_mm_embeddings can now return (mm_embeds, deepstack_embeds) with padding, matching the model’s deepstack merge path.                                                                                                                               
  - tpu_inference/runner/persistent_batch_manager.py
    Adds deepstack_cache lifecycle management; cleans deepstack entries when a request is freed, to avoid leaks.
  - tpu_inference/runner/compilation_manager.py
    Extends the compiled call signature to include deepstack_embeds (passes None for warm‑up). This is required so JIT signatures include the new optional input.
  - tpu_inference/runner/tpu_runner.py
    Plumbs DeepStack end‑to‑end: initializes deepstack_cache, handles tuple return from MultiModalManager, passes deepstack embeddings into model_fn. Also keeps input_ids alongside inputs_embeds for multimodal models (Qwen3‑VL still needs ids for segment/M‑RoPE alignment).

FIXES: #1330 #1431 

# Tests
The test present tests/ does not reflect the new MoE.

Mainly tested with actual serving scenarios. Similar to #1374 's test, done for Qwen3-VL-30B-A3B-Instruct model.

Served with following:
```bash
sudo docker exec -it ray-head bash -lc "vllm serve '/mnt/gcs/hf/Qwen3-VL-30B-A3B-Instruct' \
      --tensor-parallel-size 16 \
      --max-model-len 32768 \
      --max-num-batched-tokens 32768 \
      --disable-frontend-multiprocessing \
      --log-error-stack \
      --limit-mm-per-prompt '{\"image\":2,\"video\":2}' \
      --no-async-scheduling"
```

Example request:

```bash
curl http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "/mnt/gcs/hf/Qwen3-VL-30B-A3B-Instruct",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "Please describe what you see in the image and video."
          },
          {
            "type": "image_url",
            "image_url": {
              "url": "https://samplelib.com/lib/preview/png/sample-boat-400x300.png"
            }
          },
          {
            "type": "video_url",
            "video_url": {
              "url": "https://samplelib.com/lib/preview/mp4/sample-5s.mp4"
            }
          }
        ]
      }
    ],
    "max_tokens": 1024
  }'
```

```json
{"id":"chatcmpl-a39f07e2fdbb13ff","object":"chat.completion","created":1769138203,"model":"/mnt/gcs/hf/Qwen3-VL-30B-A3B-Instruct","choices":[{"index":0,"message":{"role":"assistant","content":"The video shows a park scene on a sunny day. The camera is posit
ioned in a grassy area with large, leafy trees providing shade. A paved path winds through the park. In the foreground, there is a wooden bench with a grey trash can next to it. The camera pans slowly to the left, revealing a road with moving traffic, inclu
ding cars and buses, separated from the park by a low hedge. The sunlight filters through the tree canopy, creating dappled light on the ground.","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null,"reasoning
_content":null},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":11739,"total_tokens":11841,"completion_tokens":102,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}
```

# Checklist


kyuyeunk:
@HyperBlaze456 thanks for the contribution! @kwang3939 can you help review this pr?

muskansh-google:
Thanks @HyperBlaze456 for implementing this change, we have validated the MMMU accuracy for Dense model on v6e and it seems to be matching with accuracy calculated on GPUs. 

Is it possible for you to split the Dense and Moe model changes? Or is that going to be too much involved work? Splitting can hasten the merge of this PR.

Moreover, can you please sync these changes with the latest main branch?

HyperBlaze456:
> Thanks @HyperBlaze456 for implementing this change, we have validated the MMMU accuracy for Dense model on v6e and it seems to be matching with accuracy calculated on GPUs.
> 
> Is it possible for you to split the Dense and Moe model changes? Or is that going to be too much involved work? Splitting can hasten the merge of this PR.
> 
> Moreover, can you please sync these changes with the latest main branch?

Hello, as the MoE shares the VL part with dense, it seems redundant to separate MoE from dense model.

I have updated the branches

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+    is preserved through separate tokens that annotate frame positions.
+
+    Args:
+        grid_thw: Tuple of (T, H, W) for each image/video
+
+    Returns:
+        segment_ids: (total_tokens,) array with segment IDs per frame
+    """
+    segments = []
+    seg_id = 1 # start for 1 to make padding 0
```
muskansh-google: Did you mean "Start from 1 to make padding 0"? Can you elaborate this comment more for a neutral reader? Why do we need to make padding 0?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+            return getattr(self._config, name)
+        except AttributeError:
+            pass
+        if self._text_config is not None:
+            try:
+                return getattr(self._text_config, name)
+            except AttributeError:
+                pass
+        raise AttributeError(
+            f"'{type(self._config).__name__}' object has no attribute '{name}'"
```
muskansh-google: Nit: The AtrributeError message can be more informative indicating that the attribute was not found in neither self._config nor self._text_config

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: Ah, yes. The comment's intent was to start segment id index from 1 to keep rest of the padding's segment id to be 0.

The function `pad_segment_ids_for_attention` consumes the created segmentIds, which follows the Qwen2.5 VL logic. It paddes `0` value to make the vision tokens be the multiple of flash attention block size.

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: I see. I will fix this.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+            # DeepStack: reshape first, then norm
+            x = x.reshape(-1, self.hidden_size)
+            x = self.norm(x)
+        else:
+            # Final merger: norm first, then reshape
+            x = self.norm(x)
+            x = x.reshape(-1, self.hidden_size)
+
+        x = self.linear_fc1(x)
+        x = nnx.gelu(x) # make this configurable?
```
muskansh-google: Can we please remove the comment? There are two options
- If we want to make this configurable, we should ideally do the same for Qwen3VLVisionMLP to maintain internal consistency and parity with Qwen2_5_VisionMLP
- [Qwen2_5_VisionPatchMerger](https://github.com/vllm-project/tpu-inference/blob/184ced40caa7f41e54a252fd7fd3279f9d7ca5aa/tpu_inference/models/jax/qwen2_5_vl.py#L419) hardcodes GELU, not sure if that is intended. 

@kwang3939 Should we make this configurable?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+
+    This module supports both:
+    - Final merger: norm before reshape (use_postshuffle_norm=False)
+    - DeepStack merger: reshape before norm (use_postshuffle_norm=True)
+    """
+
+    def __init__(
+        self,
+        d_model: int, # in
+        context_dim: int, # out
```
muskansh-google: Please either remove the comments, or add it for all the params. IMO, it is not required.

tpu_inference/models/jax/qwen3_vl.py
muskansh-google: Can we use [modeling_flax_utils](https://github.com/vllm-project/tpu-inference/blob/184ced40caa7f41e54a252fd7fd3279f9d7ca5aa/tpu_inference/models/jax/qwen2_5_vl.py#L419) for the activation function. I noticed there are three variants 'gelu', 'quick_gelu', 'gelu_new'. The choice depends on which activation function was used in the original architecture / closest variant i.e Qwen2.5-VL

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+        """
+        L, dim = x.shape
+        C = dim // (
+            self.temporal_patch_size * self.patch_size * self.patch_size
+        )
+        # Reshape to (L, T, H, W, C) for Conv3D with channels_last
+        x = x.reshape(
+            L, C, self.temporal_patch_size, self.patch_size, self.patch_size
+        )
+        # L, T, H, W, C
```
muskansh-google: Nit: This comment is a little confusing as it describes the target state i.e after the transpose operation. Maybe make the current state and target state more clear?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+        """Apply 3D patch embedding.
+
+        Args:
+            x: Input tensor of shape (num_patches, C * T * H * W)
+
+        Returns:
+            Embedded patches of shape (num_patches, hidden_size)
+        """
+        L, dim = x.shape
+        C = dim // (
```
muskansh-google: Should we add an assertion that `dim` is exactly divisible? In case `dim` is not perfectly divisible, it might lead to ShapeMismatch errors deeper in the stack.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+        self.temporal_patch_size = temporal_patch_size
+        self.hidden_size = hidden_size
+
+        kernel_size = (temporal_patch_size, patch_size, patch_size)
+        self.proj = nnx.Conv(
+            in_features=in_channels,
+            out_features=hidden_size,
+            kernel_size=kernel_size,
+            strides=kernel_size,
+            use_bias=True,
```
muskansh-google: Nit: This is a deviation from Qwen2.5-VL which has ```use_bias=False```. Can we explicitly call it out in a comment?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+            patch_size=patch_size,
+            temporal_patch_size=temporal_patch_size,
+            in_channels=in_channels,
+            hidden_size=self.hidden_size,
+            dtype=dtype,
+        )
+
+        # Learned PE, 48 x 48 H W
+        num_position_embeddings = getattr(
+            vision_config, "num_position_embeddings", 2304
```
muskansh-google: Can we use constants with a variable name that signifies what this number means, for e.g SPATIAL_GRID_SIZE=2304. Or GRID_SIZE can be defined as 48, and then pass GRID_SIZE x GRID_SIZE here.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2037 @@
...
+    h_indices = jnp.arange(1, h * 3, 3)
+    result = result.at[..., h_indices].set(freqs[1][..., h_indices])
+
+    w_indices = jnp.arange(2, w * 3, 3)  # [2, 5, 8, ..., 59]
+    result = result.at[..., w_indices].set(freqs[2][..., w_indices])
+
+    return result
+
+
+class Qwen3VLTextRMSNorm(nnx.Module):
```
kwang3939: This `Qwen3VLTextRMSNorm` is functionally repetitive with our existing JaxRmsNorm interface: https://github.com/vllm-project/tpu-inference/blob/dabd50c9a53d5234e1c8e549a532bde11b6728b7/tpu_inference/layers/jax/norm.py#L25

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: I personally don't think it should be configurable. Huggingface has that feature, but it is questionable if anyone would use it in serving scenarios.

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: I was originally confused by the naming of `context_dim`, but thinking again it does seem to be redundant.\

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: I'll move it back, as we can see `C` right at here.

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: qwen3 model series would never actually have that issue as the dim is always divisble

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+    return num_images, num_videos
+
+
+def get_mrope_input_positions(
+    input_tokens: List[int],
+    image_grid_thw: Optional[List[Tuple[int, int, int]]],
+    video_grid_thw: Optional[List[Tuple[int, int, int]]],
+    image_token_id: int,
+    video_token_id: int,
+    vision_start_token_id: int,
```
kwang3939: This `vision_start_token_id` is received as an input but never used.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+    masked_ids = jnp.where(attention_mask == 1, input_ids, -1)
+
+    # Count per sequence
+    num_images = jnp.sum(masked_ids == image_token_id, axis=1)
+    num_videos = jnp.sum(masked_ids == video_token_id, axis=1)
+
+    return num_images, num_videos
+
+
+def get_mrope_input_positions(
```
kwang3939: Could you use a different function name here so it's easy to distinguish from `get_mrope_input_positions` below?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+
+        # Ensure positions are treated as Qwen3-VL does: always 3D positions for MRoPE
+        pos = md.input_positions
+        if pos.ndim == 1:
+            # Expand vanilla positions to (3, T) for text-only
+            pos = jnp.broadcast_to(pos[None, :], (3, pos.shape[0]))
+
+        cos, sin = self.rotary_emb(pos)  # (bs, T, head_dim_original)
+
+        q = apply_rotary_pos_emb_thd_padded(q, cos, sin, self.head_dim_original)
```
kwang3939: We could pass mrope into `model_fn` with `input_positions`. Could you adapt the existing `apply_rope` for interleaving rope to avoid duplicating the implementation of the whole text model?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+        # duplicate the half-dim freqs to produce (bs, seq_len, dim).
+        freqs = jnp.concatenate([freqs, freqs], axis=-1)
+
+        cos = jnp.cos(freqs)
+        sin = jnp.sin(freqs)
+
+        return cos.astype(jnp.bfloat16), sin.astype(jnp.bfloat16)
+
+
+class Qwen3VLTextAttention(nnx.Module):
```
kwang3939: We already have a Qwen3 text attention implementation in: https://github.com/HyperBlaze456/tpu-inference/blob/7e896b2c3c508da2fcf555e9970c9fb36cf05188/tpu_inference/models/jax/qwen3.py#L41. I think we should make use of the existing module by only adapting the rope part instead of reimplementing the whole text model.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+            self.head_dim_original,
+            q_scale=q_scale,
+            k_scale=k_scale,
+            v_scale=v_scale,
+        )
+        o = self.o_proj(outputs)
+        return new_kv_cache, o
+
+
+class Qwen3VLTextMLP(nnx.Module):
```
kwang3939: Similarly, this is also a duplicate of the existing one in `qwen3.py`: https://github.com/HyperBlaze456/tpu-inference/blob/7e896b2c3c508da2fcf555e9970c9fb36cf05188/tpu_inference/models/jax/qwen3.py#L31

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+            raise NotImplementedError(f"Activation function '{hidden_act}' not implemented")
+
+    def __call__(self, x: jax.Array) -> jax.Array:
+        gate_output = self.act_fn(self.gate_proj(x))
+        up_output = self.up_proj(x)
+        down_proj = self.down_proj(gate_output * up_output)
+        return down_proj
+
+
+class Qwen3VLTextDecoderLayer(nnx.Module):
```
kwang3939: This is also a duplicate of: https://github.com/HyperBlaze456/tpu-inference/blob/7e896b2c3c508da2fcf555e9970c9fb36cf05188/tpu_inference/models/jax/qwen3.py#L162.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+
+
+def rotate_half(x: jax.Array) -> jax.Array:
+    """Rotate half the hidden dims of the input for RoPE."""
+    x1 = x[..., : x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2 :]
+    return jnp.concatenate([-x2, x1], axis=-1)
+
+
+def apply_rotary_pos_emb_thd_padded(
```
kwang3939: If we could unify the mrope with the existing `apply_rope`, I think this padding func would not be needed.

tpu_inference/runner/compilation_manager.py
```diff
@@ -240,6 +242,7 @@ def model_fn_wrapper(
                 intermediate_tensors,
                 is_first_rank,
                 is_last_rank,
+                None,
```
kwang3939: This would cause recompilation when deepstack_embeds is not None. Could you add a precompilation path for this case or add a TODO here?

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,2039 @@
...
+            deepstack_features: List of intermediate features for DeepStack
+        """
+        return self.encode_jit(x, grid_thw)
+
+
+
+class Qwen3VLModel(nnx.Module):
+    """Text model for Qwen3VL with MRoPE and DeepStack support."""
+
+    def __init__(
```
kwang3939, line 1259: After fixing the duplication of `DecoderLayer`, I think this initialization could reuse `Qwen3Model` in `qwen3.py`. A wrapper would still be needed to inject DeepStack features in `__call__`, but inheriting would avoid duplicating this initialization.

tpu_inference/runner/multimodal_manager.py
```diff
@@ -16,18 +16,69 @@
...
 from tpu_inference.models.jax.utils.multi_modal_utils import (
-    flatten_embeddings, sanity_check_mm_encoder_outputs)
+    flatten_embeddings, gather_mm_placeholders, sanity_check_mm_encoder_outputs,
+    scatter_mm_placeholders)
 
 if TYPE_CHECKING:
     from tpu_inference.runner.tpu_runner import TPUModelRunner
 
 
+def _normalize_grid_thw(grid_thw: object) -> tuple[tuple[int, int, int], ...]:
```
kwang3939: This is duplicated in both here and the model, with slightly different behaviour. Could you unify this and place it in a shared place, say `multi_modal_utils`?

tpu_inference/runner/multimodal_manager.py
```diff
@@ -156,18 +207,40 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
...
+                    mm_hashes_pos,
+                    encoder_outputs,
+                    deepstack_outputs,
+            ):
+                self.runner.encoder_cache[mm_hash] = output
+                if deepstack_output is not None:
+                    self.runner.deepstack_cache[mm_hash] = deepstack_output
 
     def gather_mm_embeddings(self, scheduler_output: "VllmSchedulerOutput",
                              target_pad_len: int) -> list[jax.Array]:
```
kwang3939: This return type annotation needs to be updated to match the behavior.

tpu_inference/models/jax/utils/multi_modal_utils.py
```diff
@@ -131,6 +131,89 @@ def _merge_multimodal_embeddings(
...
+        return None
+    if isinstance(x, jax.Array):
+        return x
+    # Handle torch.Tensor - convert via numpy
+    if hasattr(x, 'numpy'):
+        return jnp.array(x.numpy())
+    return jnp.array(x)
+
+
+def scatter_mm_placeholders(
```
kwang3939: vLLM has deprecated `gather_mm_placeholders` / `scatter_mm_placeholders` in https://github.com/vllm-project/vllm/pull/33362.  Could you remove these and also update the `multimodal_manager` based on our earlier PR to avoid using placeholder: https://github.com/vllm-project/tpu-inference/pull/1618?

kwang3939 (changes requested):

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: Addressed! Now reusing existing Text MLP

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: Qwen3VLTextDecoderLayer now reuses Qwen3DecoderLayer and only swaps in the VL-specific attention subclass.

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: refactored the dense Qwen3-VL text attention path to reuse Qwen3Attention and kept only the VL-specific rotary application overridden in Qwen3VLTextAttention

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: Qwen3VLModel now reuses Qwen3Model initialization/layer construction, and the remaining custom logic is limited to the VL-specific DeepStack injection path

tpu_inference/models/jax/utils/multi_modal_utils.py
HyperBlaze456: Addressed across c5266d97 and d409a070. Rmoeved the placeholder gather/scatter helpers from multi_modal_utils.py, and updated the DeepStack gathering path to use direct embed-index slicing instead of placeholder-style reconstruction

tpu_inference/runner/multimodal_manager.py
HyperBlaze456: Sure, modified the utils file and imported it.

muskansh-google (approved):

muskansh-google (approved):
Hi @HyperBlaze456, can you please resolve the merge conflicts as well?

tpu_inference/models/jax/qwen3_vl_moe.py
```diff
@@ -0,0 +1,961 @@
...
+)
+
+init_fn = nnx.initializers.uniform()
+
+modeling_flax_utils = FlaxUtils()
+
+logger = init_logger(__name__)
+
+@dataclass(kw_only=True)
+class Qwen3VLMoeTextExperts(nnx.Module):
```
kwang3939: I found this moe text model part is also repetitive with the existing qwen3_moe: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/jax/qwen3_moe.py.
Could you do the same as what you have done for `qwen3_vl.py` to reuse the existing moe model code?

kwang3939 (changes requested):

.omx/logs/hooks-2026-03-15.jsonl
yaochengji, line 1: What are the 3 files under `.omx` folder? Look like they're accidentally submitted.

tpu_inference/models/jax/qwen3_vl.py
```diff
@@ -0,0 +1,1592 @@
...
+        attention_metadata: AttentionMetadata,
+        inputs_embeds: Optional[jax.Array] = None,
+        *args,
+    ) -> Tuple[List[jax.Array], jax.Array, List[jax.Array]]:
+        visual_pos_mask = None
+        deepstack_embeds = None
+        if args:
+            candidate = args[-1]
+            if isinstance(candidate, (list, tuple)):
+                deepstack_embeds = candidate
```
yaochengji: This is fragile, any extra positional argument that happens to be a list would be misinterpreted as deepstack_embeds. It also makes the call signature unclear. Use an explicit keyword argument might be better.

tests/models/jax/test_qwen3_vl.py
yaochengji, line 1: Could you also cover tests for qwen3_vl_moe?

tpu_inference/models/jax/qwen3_vl.py
HyperBlaze456: I see, I will change both functions of qwen3vl and qwen3vl moe to all have explicit keyword arguments to handle this properly.

kwang3939:
Hi @HyperBlaze456, I think my previous comments are addressed! Could you please add the Qwen3-VL model to the CI and the video feature to the feature matrix? You can find the instructions for adding models and features to the CI here: https://github.com/vllm-project/tpu-inference/tree/main/.buildkite#support-matrices. Thanks!

HyperBlaze456:
Hello! I would try to open this branch and add it to CI ASAP as MoE version's weight loading is tangled and tpu spot quotas isn't quite available for testing.

amanseervi:
Hi @HyperBlaze456 , I see that the Qwen3-VL-MoE is experiencing issues. If the dense model is working fine, could you split the PR and submit the dense model first separately?

HyperBlaze456:
Hello, due to my midterm just coming up next week I would be unable to do any tasks for now. And as of 4/18 I see that the torchax path for multimodal models have been fixed, so I would check it and then determine later if the Qwen3 VL must be supported in this separate jax based implementation or no.

PR #1526, opened 2026-01-23
[Kernel]Add reduce-scatter-matmul kernel

rupengliu-meta:
# Description

All tests have passed
截屏2026-02-05 下午10 57 09
 does and how this is a change from
the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/features/Hybrid_kvcache.yml
+++ b/.buildkite/features/Hybrid_kvcache.yml
@@ -24,7 +24,7 @@ steps:
     key: "${TPU_VERSION:-tpu6e}_hybrid_kvcache_CorrectnessTest"
     soft_fail: true
     agents:
-      queue: "${TPU_QUEUE_SINGLE:-tpu_v6e_queue}"
+      queue: "${TPU_QUEUE_MULTI:-tpu_v6e_8_queue}"
     env:
       IS_FOR_V7X: "${IS_FOR_V7X}"
     commands:
--- a/tests/layers/common/test_utils.py
+++ b/tests/layers/common/test_utils.py
@@ -1,0 +1,102 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest import mock
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from absl.testing import absltest
+from jax._src import test_util as jtu
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
+
+from tpu_inference import envs
+from tpu_inference.layers.common.utils import general_device_put
+
+
+class UtilsTest(jtu.JaxTestCase):
+
+    def setUp(self):
+        super().setUp()
+        devices = jax.local_devices()
+        # Create a mesh with available devices.
+        # If running on CPU with 1 device, mesh will be (1, 1).
+        self.num_devices = len(devices)
+        mesh_shape = (self.num_devices, 1)
+        self.mesh = Mesh(np.array(devices).reshape(mesh_shape), ('x', 'y'))
+        self.sharding = NamedSharding(self.mesh, P('x', 'y'))
+
+    def test_general_device_put_single_tensor(self):
+        tensor = jnp.ones((8, 8))
+        with mock.patch.object(envs, 'TPU_MULTIHOST_BACKEND', ''):
+            result = general_device_put(tensor, self.sharding)
+
+        self.assertIsInstance(result, jax.Array)
+        self.assertEqual(result.sharding, self.sharding)
+        self.assertAllClose(result, tensor)
+
+    def test_general_device_put_pytree(self):
+        t1 = jnp.ones((8, 8))
+        t2 = jnp.zeros((16, 4))
+        tree = {'a': t1, 'b': [t2, t1]}
+
+        with mock.patch.object(envs, 'TPU_MULTIHOST_BACKEND', ''):
+            result = general_device_put(tree, self.sharding)
+
+        self.assertIsInstance(result, dict)
+        self.assertIsInstance(result['a'], jax.Array)
+        self.assertIsInstance(result['b'], list)
+        self.assertEqual(result['a'].sharding, self.sharding)
+        self.assertEqual(result['b'][0].sharding, self.sharding)
+        self.assertAllClose(result['a'], t1)
+
+    def test_general_device_put_multihost_single_tensor(self):
+        tensor = jnp.ones((8, 8))
+
+        with mock.patch.object(envs, 'TPU_MULTIHOST_BACKEND', 'ray'):
+            with mock.patch('jax.make_array_from_single_device_arrays'
+                            ) as mock_make_array:
+                mock_make_array.return_value = tensor
+
+                result = general_device_put(tensor, self.sharding)
+
+                mock_make_array.assert_called_once()
+                args, _ = mock_make_array.call_args
+                self.assertEqual(args[0], tensor.shape)
+                self.assertEqual(args[1], self.sharding)
+                # Check that x_split contains device_put result
+                # Since we didn't mock device_put, it runs on the tensor slice.
+                # We can check the length of x_split
+                self.assertEqual(len(args[2]), self.num_devices)
+                self.assertAllClose(result, tensor)
+
+    def test_general_device_put_multihost_pytree(self):
+        t1 = jnp.ones((8, 8))
+        tree = [t1, t1]
+
+        with mock.patch.object(envs, 'TPU_MULTIHOST_BACKEND', 'ray'):
+            with mock.patch('jax.make_array_from_single_device_arrays'
+                            ) as mock_make_array:
+                mock_make_array.return_value = t1
+
+                result = general_device_put(tree, self.sharding)
+
+                self.assertEqual(mock_make_array.call_count, 2)
+                self.assertIsInstance(result, list)
+                self.assertEqual(len(result), 2)
+
+
+if __name__ == "__main__":
+    absltest.main(testLoader=jtu.JaxTestLoader())
--- a/tpu_inference/layers/common/process_weights/linear_weights.py
+++ b/tpu_inference/layers/common/process_weights/linear_weights.py
@@ -22,8 +22,8 @@ from torch.nn import ParameterList
 from torch.nn.parameter import Parameter
 from torchax.tensor import Tensor
 
-from tpu_inference.layers.common.utils import \
-    reorder_concatenated_tensor_for_sharding
+from tpu_inference.layers.common.utils import (
+    general_device_put, reorder_concatenated_tensor_for_sharding)
 from tpu_inference.logger import init_logger
 
 P = PartitionSpec
@@ -187,6 +187,6 @@ def shard_linear_weights(
         key = field.name
         if (weight := getattr(weights, key, None)) is not None:
             sharding = getattr(weight_shardings, key)
-            weight = jax.device_put(weight, sharding)
+            weight = general_device_put(weight, sharding)
             setattr(weights, key, weight)
     return weights
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -16,15 +16,15 @@ from dataclasses import dataclass, fields
 
 import jax
 import jax.numpy as jnp
-from jax.experimental.layout import Format, Layout, with_layout_constraint
+from jax.experimental.layout import Layout, with_layout_constraint
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.tensor import Tensor
 
 from tpu_inference.layers.common.moe import MoEBackend
 from tpu_inference.layers.common.quantization import quantize_tensor
 from tpu_inference.layers.common.sharding import ShardingAxisName
-from tpu_inference.layers.common.utils import \
-    reorder_concatenated_tensor_for_sharding
+from tpu_inference.layers.common.utils import (
+    general_device_put, reorder_concatenated_tensor_for_sharding)
 from tpu_inference.utils import align_to
 
 P = PartitionSpec
@@ -386,6 +386,6 @@ def shard_moe_weights(
         if (weight := getattr(weights, key, None)) is not None:
             layout = getattr(weight_layouts, key)
             sharding = getattr(weight_shardings, key)
-            weight = jax.device_put(weight, Format(layout, sharding))
+            weight = general_device_put(weight, sharding, layout)
             setattr(weights, key, weight)
     return weights
--- a/tpu_inference/layers/common/utils.py
+++ b/tpu_inference/layers/common/utils.py
@@ -14,6 +14,9 @@
 
 import jax
 import jax.numpy as jnp
+from jax.experimental.layout import Format
+
+from tpu_inference import envs
 
 
 def reorder_concatenated_tensor_for_sharding(concatenated_tensor: jax.Array,
@@ -92,3 +95,34 @@ def slice_sharded_tensor_for_concatenation(sharded_tensor: jax.Array,
         start_offset = end_offset
 
     return split_tensors
+
+
+def general_device_put(tensor: jax.Array, sharding, layout=None) -> jax.Array:
+    """
+    Put a tensor onto devices with the given sharding.
+    This method handles both single-host and multi-host cases.
+    """
+
+    def _put(t):
+        multihost_backend = envs.TPU_MULTIHOST_BACKEND
+        if multihost_backend != "ray":
+            return jax.device_put(t, sharding)
+
+        # NOTE: at here, num_global_devices != num_local_devices
+        # meaning we are in multi-host setup. Each host will run the same process
+        # and each process only need to handle the devices accessible to this host.
+        shape = t.shape
+        x_split = [
+            jax.device_put(t[i], device) for device, i in
+            sharding.addressable_devices_indices_map(shape).items()
+        ]
+        global_array = jax.make_array_from_single_device_arrays(shape,
+                                                                sharding,
+                                                                x_split,
+                                                                dtype=t.dtype)
+        if layout is not None:
+            global_array = jax.device_put(global_array,
+                                          Format(layout, sharding))
+        return global_array
+
+    return jax.tree_util.tree_map(_put, tensor)
--- a/tpu_inference/layers/vllm/process_weights/cleanup_sharding.py
+++ b/tpu_inference/layers/vllm/process_weights/cleanup_sharding.py
@@ -31,8 +31,8 @@ from vllm.lora.layers.base_linear import BaseLinearLayerWithLoRA
 from vllm.model_executor.layers.vocab_parallel_embedding import (
     ParallelLMHead, VocabParallelEmbedding)
 
-from tpu_inference import envs
 from tpu_inference.layers.common.sharding import ShardingAxisName
+from tpu_inference.layers.common.utils import general_device_put
 from tpu_inference.logger import init_logger
 from tpu_inference.utils import to_jax_dtype
 
@@ -105,7 +105,7 @@ def _convert_to_torchax_and_shard(tensor: torch.Tensor,
             tensor = jax_view(tensor)
         else:
             tensor = t2j(tensor)
-        return torch_view(_sharded_device_put(tensor, sharding))
+        return torch_view(general_device_put(tensor, sharding))
 
 
 def _shard_tensor_to_tpu_replicated(tensor: torch.Tensor,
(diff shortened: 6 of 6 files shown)


tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,599 @@
...
+    x = jnp.ones((m, k), dtype=jnp.bfloat16)
+    y = jnp.ones((k, n_per_device), dtype=jnp.bfloat16)
+
+    # Setup mesh
+    devices = mesh_utils.create_device_mesh((num_devices, ))
+    mesh = Mesh(devices, axis_names=("x", ))
+
+    # Shard inputs appropriately
+    # x is replicated, y is passed as local shape (replicated, each device has same copy)
+    x_sharding = NamedSharding(mesh, P(None, None))  # x replicated
```
vanbasten23: In the rs matmul, the input x and y should be sharded on the contracting dim, right?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+        return
+
+    devices = mesh_utils.create_device_mesh((num_devices, ))
+    mesh = Mesh(devices, axis_names=("x", ))
+
+    for dtype in [jnp.float32, jnp.bfloat16]:
+        m = 256 * num_devices
+        k = 512
+        n = 256 * num_devices
```
vanbasten23: also put the test in tpu-inference/tests/kernels/collectives to be consistent?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,599 @@
...
+        )
+        # Print full mapping for debugging
+        print("  Full slice mapping:")
+        for device_idx, shard in enumerate(result.addressable_shards):
+            actual_value = float(shard.data[0, 0])
+            inferred_slice = int(actual_value / k) - 1
+            print(f"    Device {device_idx} → Slice {inferred_slice}")
+
+
+def test_reduce_scatter_matmul_different_dtypes():
```
vanbasten23: For this test and the one below `test_reduce_scatter_matmul_larger_dimensions` and `test_reduce_scatter_matmul_lhs_transpose(`, suggest to make it parameterized such as https://github.com/vllm-project/tpu-inference/blob/c2a287c4bcd1fadccdaeec765284558504f0d904/tests/kernels/collectives/all_gather_matmul_kernel_test.py#L23-L28

tpu_inference/kernels/reduce-scatter-matmul/pipelining_diagram.md
```diff
@@ -0,0 +1,262 @@
...
+
+    style D0 fill:#c8e6c9
+    style D1 fill:#c8e6c9
+    style D2 fill:#c8e6c9
+    style D3 fill:#c8e6c9
+```
+
+## Pipeline Timeline (4 Devices, grid_n=1, grid_k=1)
+
+This shows the execution timeline across `global_step_id`:
```
vanbasten23: what's the x-axis? The "global_step_id"?
And why does each MXU have different length 
image
?

tpu_inference/kernels/reduce-scatter-matmul/kernel.py
```diff
@@ -0,0 +1,931 @@
...
+) -> jax.Array:
+    """Reference implementation of reduce-scatter matmul using pure JAX.
+
+    The operation computes:
+        output[device_i] = (x @ y)[device_i * m_per_device : (device_i+1) * m_per_device]
+
+    Each device computes the full matmul and takes its shard of rows.
+    This is a scatter operation along the m dimension.
+
+    Note: In this pattern, x and y are replicated on all devices, so each device
```
vanbasten23: I think for a rs matmul, the API signature should be: (x: [bs, k@TP], y: [out_feature, in_features@TP]) --> [bs@TP, out_features]. Notice that the input contracting dimension should be sharded instead of duplicated. This is also consistent with the out_proj, a RowParallelLinear in which the contracting dim is sharded.

rupengliu-meta:
@vanbasten23 thanks for the review, there is some issue related to the implementation of the kernel, I'm debugging it right now!

tpu_inference/kernels/reduce-scatter-matmul/pipelining_diagram.md
rupengliu-meta: if you open the readme with mermaid, it will show the correct picture, this picture is not very correct lol

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: updated

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: updated

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: updated

tpu_inference/kernels/reduce-scatter-matmul/pipelining_diagram.md
rupengliu-meta: updated

tpu_inference/kernels/reduce-scatter-matmul/kernel.py
rupengliu-meta: updated

rupengliu-meta:
the initial version of the kernel is ready for review, it achieved true computation & communication overlap. As discussed offline, we could start to optimize it together @vanbasten23 @kyuyeunk

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+    print(
+        f"    Expected shape: {expected_np.shape}, dtype: {np.asarray(expected).dtype}"
+    )
+    print(
+        f"    Result range: [{float(result_np.min()):.4f}, {float(result_np.max()):.4f}]"
+    )
+    print(
+        f"    Expected range: [{float(expected_np.min()):.4f}, {float(expected_np.max()):.4f}]"
+    )
+    print(f"    Max abs diff: {float(diff.max()):.4f}")
```
vanbasten23: could you also print the mean value "jnp.mean(jnp.abs(expected_output - actual_output))"?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+    print(f"    Max rel diff: {float(rel_diff.max()):.4f}")
+    print(
+        f"    Result[0,0]: {float(result_np[0,0]):.4f}, Expected[0,0]: {float(expected_np[0,0]):.4f}"
+    )
+
+    # Check if result is a fraction of expected (indicates missing accumulations)
+    if float(expected_np.mean()) > 1e-6:
```
vanbasten23: Could you explain what this tries to do?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+
+    x_sharded = jax.device_put(x, x_sharding)
+    y_sharded = jax.device_put(y, y_sharding)
+
+    result = single_dir_reduce_scatter_matmul_sharded(x_sharded,
+                                                      y_sharded,
+                                                      mesh,
+                                                      axis_name="x")
+
+    def reference_fn(x_shard, y_shard):
```
vanbasten23: Could you put the ref impl to the kernel.py to be consistent with other kernel code? So you don't need to define it in every test.

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+
+        expected = reference_fn_sharded(x_sharded, y_sharded)
+
+        np.testing.assert_allclose(result, expected, rtol=0.05, atol=0.1)
+        print(f"  ✓ device_count={dc} works")
+
+    print("✓ test_different_device_counts passed!")
+
+
+def test_slice_mapping():
```
vanbasten23: Isn't the test "test_single_dir_reduce_scatter_matmul_with_random_values" already tests the slice mapping because the JAX ref impl  returns the source of truth?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+            print(f"    ⚠ MISMATCH on device {device_idx}")
+
+    if all_correct:
+        print("✓ test_slice_mapping passed!")
+    else:
+        raise AssertionError(
+            "test_slice_mapping FAILED - slice mapping is wrong")
+
+
+def run_all_tests():
```
vanbasten23: We usually put the kernel tests to the test folder, such as https://github.com/vllm-project/tpu-inference/blob/main/tests/kernels/collectives/all_gather_matmul_kernel_test.py. 

Also, with pytest, we don't need a function aggregating all the tests. You can see the above test as an example.

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+
+    # Check per-device local shape
+    local_shape = result.addressable_shards[0].data.shape
+    expected_local_shape = (m_per_device, n)
+
+    assert (
+        local_shape == expected_local_shape
+    ), f"Local shape mismatch: got {local_shape}, expected {expected_local_shape}"
+
+    print(
```
vanbasten23: Is it possible to test the output shape in the above basic test?

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
```diff
@@ -0,0 +1,581 @@
...
+
+    _print_array_comparison(result, expected, "Basic Test")
+
+    # Each element should be K (since input is all ones and x @ y.T sums K ones)
+    np.testing.assert_allclose(result, expected, rtol=0.02)
+
+    print("✓ test_single_dir_reduce_scatter_matmul_basic passed!")
+
+
+def test_single_dir_reduce_scatter_matmul_with_random_values():
```
vanbasten23: Maybe you can make this test cover more cases with @parameterized.product:
- more dtypes so you don't need a separate "test_single_dir_reduce_scatter_matmul_different_dtypes"
- more shapes so you don't need a separate "test_single_dir_reduce_scatter_matmul_larger_dimensions"
- more device counts

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
```diff
@@ -0,0 +1,744 @@
...
+            final_copy = pltpu.make_async_copy(
+                src_ref=scratch_ref.at[working_slot],
+                dst_ref=out_ref,
+                sem=copy_sem,
+            )
+            final_copy.start()
+            final_copy.wait()
+
+
+def single_dir_reduce_scatter_matmul(
```
vanbasten23: I think you can drop the "single_dir" wordings since eventually we are going to do bidirectional.

rupengliu-meta:
tests have passed for both kernel 
截屏2026-02-06 下午1 53 38

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: updated

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: there is 1e-6 precision drop after cross device com

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: sure!

rupengliu-meta: I think the function is in the kernel code, reference_fn is just a wrapper for that

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: one will never worry that tests are too much, lol

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: ok, will move the tests!

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: the below tests already covered that!

tpu_inference/kernels/reduce-scatter-matmul/test_reduce_scatter_matmul.py
rupengliu-meta: sure, will update!

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
rupengliu-meta: I added the bidirectional implementation as well

rupengliu-meta:
@vanbasten23 updated! thanks for the review

vanbasten23:
fyi, here is my reduce scatter matmul kernel https://github.com/vllm-project/tpu-inference/tree/xiowei/rs_scatter_matmul_take1

kyuyeunk:
> fyi, here is my reduce scatter matmul kernel https://github.com/vllm-project/tpu-inference/tree/xiowei/rs_scatter_matmul_take1

@vanbasten23 @rupengliu-meta can you share how these two implementations differ at a high level?

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
```diff
@@ -0,0 +1,744 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Single-Direction Reduce-Scatter Matmul with Delayed Computation.
+
+This is a simplified, educational implementation showing the core concept of
+delayed computation in a reduce-scatter fused with matmul.
+
+Setup (4 devices example):
+- Each device has: x[M, K_shard], y[N, K_shard] where K is sharded
```
vanbasten23: No need to check in this single directional kernel?

tpu_inference/kernels/reduce-scatter-matmul/bidirectional_kernel.py
```diff
@@ -0,0 +1,820 @@
...
+
+    def get_left_target_block(step):
+        """LEFT direction: compute TOP half of block (my_id + step + 1) % N."""
+        return mod(my_id + step + 1, num_devices)
+
+    def get_right_target_block(step):
+        """RIGHT direction: compute BOT half of block (my_id - step - 1) % N."""
+        return mod(my_id - step - 1, num_devices)
+
+    def get_m_start_for_block_top(block_idx):
```
vanbasten23: Is it used?

rupengliu-meta:
@kyuyeunk @vanbasten23 is it possible to use my implementation, I already get blockwise quantization fused into the kernel, once this is merged, I will update immediately?

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
rupengliu-meta: we may need this actually, for example, if it's computation bound, the single direction kernel will have better performance. If it's communication bound, the bidirectional will be better, wdyt?

tpu_inference/kernels/reduce-scatter-matmul/bidirectional_kernel.py
rupengliu-meta: ok, forgot to use these helper functions, let me remove them

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
vanbasten23: Not sure if I follow. Even if it is compute bound, the single direction kernel has the same perf as the bidirectional one. Regardless of compute bound or memory bound, we can use the bi-directional one, right?

vanbasten23:
> @kyuyeunk @vanbasten23 is it possible to use my implementation, I already get blockwise quantization fused into the kernel, once this is merged, I will update immediately?

Just sync'd offline: will merge the current implementation, then I'll improve the upon it, then Rupeng will add the blockwise quant.

vanbasten23:
> > fyi, here is my reduce scatter matmul kernel https://github.com/vllm-project/tpu-inference/tree/xiowei/rs_scatter_matmul_take1
> 
> @vanbasten23 @rupengliu-meta can you share how these two implementations differ at a high level?

On a high level, 
- In the current implementation, each device does `x_local @ y_local = [m, k_local] @ [n, k_local]`. Grid=[num_devices]
  - Compute: in each kernel invocation, the kernel takes x_local[m//num_devices, k_local]. Then it split the x_local.shape[0] to be upper and lower `x_local[:m//num_devices//2, k_local]` and `x_local[m//num_devices//2:, k_local]`. Then we do 2 matmuls.
  - Comm: in each kernel invocation, the kernel pass the scratch space to the left and right simultaneously.
- In my implementation, each device again does `x_local @ y_local = [m, k_local] @ [n, k_local]`. Differences are
  - Compute: in each kernel invocation, the kernel takes x_local[m//num_devices, k_local]. Then it split the the y_local.shape[0] (the n dim)
  - Comm: in each kernel invocation, the kernel passes the scratch  to either left or right alternatively.

vanbasten23:
Thanks Rupeng. Just one last comment https://github.com/vllm-project/tpu-inference/pull/1526#discussion_r2784639225

tpu_inference/kernels/reduce-scatter-matmul/single_direction_kernel.py
rupengliu-meta: ok, will remove the single direction one

rupengliu-meta:
thanks! updated!

vanbasten23 (approved):
Thanks @rupengliu-meta !

rupengliu-meta:
I don't have permission to merge the PR, feel free to merge it, thanks!

kyuyeunk:
> I don't have permission to merge the PR, feel free to merge it, thanks!

let me take a very quick look at it and i'll press the merge button :)

tpu_inference/kernels/reduce-scatter-matmul/bidirectional_kernel.py
```diff
@@ -0,0 +1,812 @@
...
+    Final DMA exchange
+    Compute own block contributions:
+      - LEFT:  Compute P₀(B0_TOP) for own block's top half
+      - RIGHT: Compute P₀(B0_BOT) for own block's bot half
+    Accumulate to received data
+    Write output:
+      - TOP half from LEFT direction
+      - BOT half from RIGHT direction
+
+KEY INSIGHTS:
```
kyuyeunk, line 63: what is the rationale for splitting into top and bot? isn't it possible to just send the same block left and right and let the transfer stop at the middle & that will achieve the same desired result?

meaning
- 0 -> 1 -> 2 -> 3 -> 4
- 0 -> 7 -> 6 -> 5 -> 4

also, when a dma size is smaller than a certain threshold, we are often latency bound and not bandwidth bound. meaning, even if both approaches might have same transfer size in total, splitting approach requires more hops & potentially have more latency requirements.

rupengliu-meta: From Claude 😄 
Short answer: It's technically possible in concept, but hard to implement. Here's why:

Problem 1: The Midpoint Collision / Double-Counting
In a reduce-scatter, every device contributes exactly once to each output shard. With the proposed approach, consider what happens at the midpoint device (device 4 in an 8-device ring):

From the LEFT path: device 4 receives a partial sum that already includes contributions from devices 0, 1, 2, 3
From the RIGHT path: device 4 receives a partial sum that already includes contributions from devices 0, 7, 6, 5
Notice device 0's contribution appears in BOTH partial sums. You'd need to subtract it to avoid double-counting. More generally, each originating device's partial product would be double-counted since it was sent in both directions. You'd need special "merge" logic at every device that receives from both directions, to know which contributions have already been summed and to avoid duplication.

Problem 2: The Owner Device Sees Both Paths
In reduce-scatter, each device "owns" a specific output block. Say device 3 owns Block 3. It needs P₀(B3) + P₁(B3) + P₂(B3) + ... + P₇(B3) — exactly once each.

With the current algorithm, this is clean:

LEFT ring accumulates TOP halves, RIGHT ring accumulates BOT halves
Each direction does a complete independent reduce-scatter for its half
No overlap, no double-counting, trivially correct
With the "send same block both ways" approach, you need a complex merge protocol to combine the two partial sums arriving from different directions, subtracting the double-counted contributions. This adds both logic complexity and extra compute.

Problem 3: Asymmetric Step Counts
For the "meet in the middle" idea to halve latency, each direction only needs N/2 hops instead of N-1. But the problem is that different owner devices are at different distances from the two frontiers. Device 0 is 0 hops from the left path start and N-1 hops from the right path end. The "meeting point" is different for every block's owner, creating an asymmetric, hard-to-pipeline schedule.

Problem 4: Compute Collision at the Midpoint
Even if you solve the double-counting, both directions arrive at the midpoint device at the same step and need to compute partial products for the same block and same rows. This means two matmuls writing to the same output region, requiring careful synchronization and merge logic. The current M-split algorithm avoids this entirely — at step 3 (the midpoint), LEFT computes B4_TOP while RIGHT computes B4_BOT — they're writing to different memory regions, so no conflict exists.

kyuyeunk: hmm not sure. isn't that how it's usually implemented? see diagrams in https://jax-ml.github.io/scaling-book/sharding/

rupengliu-meta: yeah, I know that, but I also think what Claudes explained makes sense, it will be tricky to implement the approach you mentioned above, it's possible, but it will not be that straightforward

rupengliu-meta: I think Xiongfei's approach is also similiar to mine, but he is doing one computation and one send per iteration

PR #1529, opened 2026-01-25
Fix pip install command syntax in installation.md

Akasxh:
**Description**: The installation command in installation.md incorrectly uses --version=, which is not a valid pip argument and causes the command to fail.

This PR updates the command to use the standard == specifier for pinning the version.

#### The previous pip was :

`pip install vllm-tpu --version=0.13.2.post6 `

#### Now changed to:

`pip install vllm-tpu==0.13.2.post6`


Akasxh:
@bvrockwell I think this can be merged without any further changes.

PR #1533, opened 2026-01-26
Wrap xla_quantized_matmul with shard map

rupengliu-meta:
# Description
This is to remove all reduce triggered by reduce_max. This also implicitly converts the quantization from per channel into block
截屏2026-01-26 下午12 36 01
This is the latest trace and I can see that the all-reduce is gone
截屏2026-01-26 下午7 21 42

Perf improved around: 7%
截屏2026-01-27 上午10 12 29
截屏2026-01-27 上午10 12 38

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/linear.py
+++ b/tpu_inference/layers/vllm/linear.py
@@ -40,27 +40,27 @@ def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
 
     # NOTE (jacobplatin/kyuyeunk) there have been numeric issues (concerning) NaNs
     # with the kernel and thus we disable it for now.
-    if envs.ENABLE_QUANTIZED_MATMUL_KERNEL:
-        out_axis, in_axis = weight_sharding
-        x_sharding = P(None, in_axis)
-        scale_sharding = P(out_axis, )
-        out_sharding = P(None, out_axis)
+    out_axis, in_axis = weight_sharding
+    x_sharding = P(None, in_axis)
+    scale_sharding = P(out_axis, )
+    out_sharding = P(None, out_axis)
 
-        x = jax.lax.with_sharding_constraint(x,
-                                             NamedSharding(mesh, x_sharding))
+    x = jax.lax.with_sharding_constraint(x, NamedSharding(mesh, x_sharding))
 
-        def wrapper(x, w_q, w_s):
+    def wrapper(x, w_q, w_s):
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/vllm/linear.py
```diff
@@ -40,27 +40,27 @@ def sharded_quantized_matmul(x: jax.Array, w_q: jax.Array, w_s: jax.Array,
 
     # NOTE (jacobplatin/kyuyeunk) there have been numeric issues (concerning) NaNs
     # with the kernel and thus we disable it for now.
-    if envs.ENABLE_QUANTIZED_MATMUL_KERNEL:
```
vanbasten23, line 43: Good catch

kyuyeunk: yes, and this code change maintains the behavior (see line 52). only difference is that xla based kernel is executed inside `shard_map` to get rid of `all reduce`. 

previously, it was doing per-channel quantization. and the issue was that if `x=(b,d)` and `x` was sharded along `d` dimension, it will trigger allreduce during `jnp.max(x, axis=1)` but if this is executed inside `shard_map`, no collective operation is needed. also, we get extra accuracy boost for free because we are essentially doing block, quantization where `num_blocks=tp_size`.

kyuyeunk:
in the description, can you add e2e performance difference between before and after this change?

vanbasten23:
By "all reduce after the reduce_max", do you mean the `all-reduce.370`?

tpu_inference/layers/vllm/linear.py
vanbasten23: Thanks @kyuyeunk for the explanation! Makes sense.

kyuyeunk:
> By "all reduce after the reduce_max", do you mean the `all-reduce.370`?

@rupengliu-meta probably worth renaming the description into "all reduce triggered by reduce_max".

rupengliu-meta:
> > By "all reduce after the reduce_max", do you mean the `all-reduce.370`?
> 
> @rupengliu-meta probably worth renaming the description into "all reduce triggered by reduce_max".

updated

rupengliu-meta:
> By "all reduce after the reduce_max", do you mean the `all-reduce.370`?

yes

rupengliu-meta:
> in the description, can you add e2e performance difference between before and after this change?

will update tmr!

rupengliu-meta:
updated with perf gain, thanks!

PR #1544, opened 2026-01-27
[fix] jax_view cannot handle torch.ParamList

lk-chen:
# Description

As of now, `jax_view` cannot handle `torch.ParamList` and will [return ParamList as is](https://github.com/google/torchax/blob/320ff06408d583030bb28728cc7d1db98a2ab62b/torchax/interop.py#L224)

We didn't catch this issue because the test was not updated to alternate `linear_config`. The bug was introduced in #1505 

This PR fixes above cases.

# Tests

CI, and

```
export CURRENT_VLLM_MODEL=Qwen/Qwen2.5-7B-Instruct
vllm serve $CURRENT_VLLM_MODEL \
    --max-model-len=2048 \
    --disable-log-requests
```

# Checklist

### DIFF
--- a/tests/layers/vllm/test_unquantized.py
+++ b/tests/layers/vllm/test_unquantized.py
@@ -37,6 +37,7 @@ from vllm.model_executor.layers.linear import (ColumnParallelLinear,
                                                RowParallelLinear)
 from vllm.model_executor.model_loader import get_model as vllm_get_model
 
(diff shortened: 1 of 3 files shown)


kyuyeunk:
ohh seems like i was wrong about my last comment regarding not needing for loop because `jax_view` supports tree structure? thanks for catching this @lk-chen!

PR #1548, opened 2026-01-28
Fix hf_transfer pip install issue.

QiliangCui:
# Description
Fix hf_transfer pip install issue.

change the test filter to make sure it runs when requirements changes.


vanbasten23 (approved):
Thanks for the fix!

PR #1553, opened 2026-01-28
Skip fused moe test if device less than TPU7

JiriesKaileh:
# Description

Temporarily disable fused moe test until jax version update lands.

### DIFF
--- a/tests/kernels/fused_moe_v1_test.py
+++ b/tests/kernels/fused_moe_v1_test.py
@@ -26,17 +26,6 @@ from tpu_inference.kernels.fused_moe.v1.kernel import fused_ep_moe, ref_moe
 jax.config.parse_flags_with_absl()
 
 
-def is_tpu_v7_or_newer():
-    try:
-        d = jax.devices()[0]
-        # Check for TPU v7 specific strings
-        is_v7 = "v7" in d.device_kind or "Trillium" in d.device_kind
-        # logical OR with the JAX utility if you trust it for some cases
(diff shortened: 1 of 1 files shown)


tests/kernels/fused_moe_v1_test.py
```diff
@@ -24,6 +26,17 @@
 jax.config.parse_flags_with_absl()
 
 
+def is_tpu_v7_or_newer():
```
kyuyeunk: maybe this approach is simpler?: https://github.com/vllm-project/tpu-inference/blob/04adca196ca4f80005a31da1e92cebe15a3eed44/tests/layers/vllm/test_mxfp4.py#L45-L46

JiriesKaileh: I will try that locally

tests/kernels/fused_moe_v1_test.py
```diff
@@ -109,12 +111,11 @@ def sub_channel_quantize(x, quant_dtype, wsz=256):
 
 
 @jtu.with_config(jax_numpy_dtype_promotion="standard")
+@unittest.skipUnless(jtu.is_device_tpu_at_least(version=7), "Expected TPUv7+")
```
vanbasten23, line 114: previous fix doesn't work?

JiriesKaileh: It works, but I thought I should simplify it. This simpler fix works as well

kyuyeunk (approved):

PR #1561, opened 2026-01-29
Add named scope for gmm

helloworld1:
# Description

Add named scope for gmm for easier debugging and tuning gmm kernel

# Tests

Tested E2E benchmark and tracing

# Checklist

### DIFF
--- a/tpu_inference/kernels/megablox/gmm.py
+++ b/tpu_inference/kernels/megablox/gmm.py
@@ -620,6 +620,7 @@ def gmm(
         )),
         interpret=interpret,
         cost_estimate=cost_estimate,
+        name=f"gmm-m_{m}-k_{k}-n_{n}-tm_{tm}-tk_{tk}-tn_{tn}",
     )
 
     out = call_gmm(


tpu_inference/kernels/megablox/gmm.py
```diff
@@ -617,6 +617,7 @@ def out_transform_indices(n_i, grid_id, b_i, k_i, group_metadata,
         )),
         interpret=interpret,
         cost_estimate=cost_estimate,
+        name=f"gmm-m_{m}-k_{k}-n_{n}-tm_{tm}-tk_{tk}-tn_{tn}",
```
vanbasten23, line 623: where is it used?

vanbasten23: need to do something like https://github.com/vllm-project/tpu-inference/blob/414bf19915913cad54eb2506d6979277e0273264/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py#L1581?

helloworld1: name is basically scope name for pallas kernel. I think we are refactoring out from using jax.named_scope to pallas native scope name. Don't think it is used anywhere else in this kernel so we are OK with just inlining the name here without creating another variable / function?

image

PR #1562, opened 2026-01-29
fused_moe_gmm EP to run two stage local->global reduction

helloworld1:
# Description

fused_moe_gmm run EP with two stage reduction, local reduction on all top_k experts first and then global reduction on all experts.

The original logic is like 
1. After GMM2, the gmm2_output is (token * top_k,  hidden) sorted by experts. Only token, expert pair owned by the EP rank has non zero values.
2. Use ragged-all-to-all to broadcast token,expert pair to all ranks
3. Unsort the token, expert pair into (token, top_k, hidden)
4. Run reduction

The new local reduction logic is like 
1. After GMM2, the gmm2_output is (expert * top_k,  hidden) sorted by experts. Only token, expert pair owned by the EP rank has non zero values.
2. For each EP rank, unsort the gmm2_output into (token, top_k, hidden), Note many token, top_k has zero values since those tokens does not have any expert in that EP rank
3. Run local reduction to sum local tokens to (token, hidden)
4. Run psum for all-reduce the (token, hidden) across the ranks.

This way we will reduce the communication. Originally ragged a2a (65536, 6144) takes about 10ms, and now it is all-reduce (8192, 6144) takes only 2.5ms. This likely due to less communication size and all-reduce being more efficient than ragged a2a.

# Tests

E2E benchmark. Prompt output test of short and long prompts, GMM EP unit tests passing, simple numerical validation between original EP and local reduction passes atol/rtol 1e-4

# Checklist

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -235,6 +235,9 @@ def expert_sharded_gmm(
     rhs_scale: jax.Array | None,
     rhs_bias: jax.Array | None,
     group_sizes: jax.Array,
+    topk_argsort_revert_indices: jax.Array,
+    topk_weights: jax.Array,
+    topk: int,
     is_last_gmm: bool,
     mesh: Mesh,
 ) -> jax.Array:
@@ -260,6 +263,7 @@ def expert_sharded_gmm(
             tiling=(tm, tk, tn),
             group_offset=group_offset[0],
         )
+
         return gmm_res
 
     # The result from gmm on each shard has the same shape, but only the rows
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -302,98 +306,24 @@ def _gmm(lhs, rhs, rhs_scale, rhs_bias, group_sizes, group_offset):
...
         return gmm_res
 
-    # For i-th shard, it is responsible groups (AKA experts) from
-    # i*num_experts_per_shard to (i+1)*num_experts_per_shard We sum them up to
-    # get total rows in that shard, and that is the size for shard to send to
-    # its peers. This is also the number of non-zero rows from the gmm results.
-    # In the working example, send_sizes would be [3, 2, 5, 4].
+    def _top_k_expert_reduction(gmm_res_local):
+        # First run local reduction on topk experts owned by the rank for all tokens
+        token_topk_hidden = gmm_res_local[topk_argsort_revert_indices].reshape(
```
kyuyeunk: nit: `topk_argsort_revert_indices` is also a jax array and not a static value right? if so, maybe it's better to also make it as an argument so it's more explicit?

same goes with other arguments like `topk_weights`.

tpu_inference/layers/common/fused_moe_gmm.py
kyuyeunk, line 422: nit: can be done later, but since now ep and tp path completely diverges after this point, maybe we can create a separate ep / tp function that calls all the gate up, activation fn, gate donwn, and token routing?

tpu_inference/layers/common/fused_moe_gmm.py
helloworld1: Yes, make the function pure

helloworld1: Updated PR for that

tpu_inference/layers/common/fused_moe_gmm.py
helloworld1: Yes, we can refactor it in a later PR to make the logic cleaner

kyuyeunk (approved):

PR #1566, opened 2026-01-29
Fix the empty model_id error in Ray multihost scenario caused by 1477

mrjunwan-lang:
# Description

Reset the weights only in RUNAI scenarios

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/sharding.py
+++ b/tpu_inference/layers/common/sharding.py
@@ -35,7 +35,6 @@ class ShardingAxisNameBase:
     SEQUENCE = ('data', 'attn_dp')
     ATTN_DATA = ('data', 'attn_dp')
     MLP_DATA = 'data'
-    TENSOR = ('model', 'expert')
     ATTN_HEAD = ('model', 'expert')
     ATTN_TENSOR = None
     MLP_TENSOR = ('attn_dp', 'model', 'expert')
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, Optional
+from typing import Any, Callable, Optional
 
 import jax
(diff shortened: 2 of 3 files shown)


tpu_inference/executors/ray_distributed_executor.py
```diff
@@ -356,8 +356,8 @@ def sort_by_driver_then_worker_ip(item: RayWorkerMetaData):
             # This allows each worker to independently invoke `maybe_pull_model_tokenizer_for_runai`
             # and stream the model from GCS.
             if node_id != driver_node_id and hasattr(
-                    self.vllm_config, "model_config") and hasattr(
-                        self.vllm_config.model_config, "model_weights"):
+                    self.vllm_config, "model_config"
+            ) and self.vllm_config.model_config.model_weights:
```
amacaskill: accessing self.vllm_config.model_config.model_weights will cause an AttributeError if model_weights does not exist on the model_config object which is possible. 

Can you please have both: 

`and hasattr(self.vllm_config.model_config, "model_weights") and self.vllm_config.model_config.model_weights`

tpu_inference/worker/tpu_worker.py
```diff
@@ -85,7 +85,7 @@ def __init__(
         # Explicitly trigger RunAI download on the worker if needed.
         # This handles downloading config.json and other non-weight files to the
         # worker's local cache before VllmModelWrapper initialization.
-        if hasattr(self.model_config, "maybe_pull_model_tokenizer_for_runai"):
+        if self.model_config.maybe_pull_model_tokenizer_for_runai:
```
amacaskill: Not sure why this is needed. `if hasattr(self.model_config, "maybe_pull_model_tokenizer_for_runai"):` Checks existence of this method safely. 

`if self.model_config.maybe_pull_model_tokenizer_for_runai:` Checks the existence of this method, but Raises AttributeError if the method is missing. 

We shouldn't throw an error and crash if maybe_pull_model_tokenizer_for_runai isn't set.

PR #1567, opened 2026-01-29
[JAX][DeepSeek] Fix `VLLM_MOE` MoE backend for unquantized DeepSeek model

jrplatin:
# Description

Currently, using the `VLLM_MOE` backend doesn't work with the unquantized DeepSeek model.  This PR fixes that.

# Tests

Verified using the following command:

```
 USE_VLLM_MOE_KERNEL=1  VLLM_MLA_DISABLE=1  NEW_MODEL_DESIGN=1 TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=deepseek-ai/DeepSeek-R1 --max-model-len=5120 --max-num-batched-tokens 128 --max-num-seqs=128 --no-enable-prefix-caching --gpu-memory-utilization 0.95 --tensor-parallel-size 8 --additional_config='{"skip_quantization": true, "sharding": {"sharding_strategy": {"expert_parallelism": 1, "tensor_parallelism": 8}}}' --kv-cache-dtype=fp8  --hf_overrides  '{"num_hidden_layers": 10}'
```

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/moe/moe.py
+++ b/tpu_inference/layers/jax/moe/moe.py
@@ -194,8 +194,8 @@ class MoE(nnx.Module):
             # in the future
             output_TD = fused_moe_func(
                 hidden_states=x_TD,
-                w1=self.kernel_gating_upproj_EFD.value,
-                w2=self.kernel_down_proj_EDF.value,
+                w1=self.kernel_gating_upproj_EDF.value,
+                w2=self.kernel_down_proj_EFD.value,
                 w1_bias=self.w1_bias,
                 w2_bias=self.w2_bias,
                 w1_scale=self.w1_scale,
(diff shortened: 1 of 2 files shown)


tpu_inference/layers/jax/moe/moe.py
```diff
@@ -318,15 +321,18 @@ def __post_init__(self, rngs: nnx.Rngs):
                 "bd2c": 256,
             }
         elif self.moe_backend == MoEBackend.VLLM_MOE:
+            # TODO (jacobplatin): the current GMM kernel expects that w1/w2 have the second and third
+            # dimensions transposed, but this is likely not optimal for DeepSeek, so we will
+            # need to fix this in the future
             self.kernel_gating_upproj_EFD = create_param(
```
bzgoogle: do you consider to update the suffix to reflect the shape? or if it introduces complexity, we can remove the suffix?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -318,15 +321,18 @@ def __post_init__(self, rngs: nnx.Rngs):
...
                 rngs,
-                shape=(E, 2 * F, D),
+                shape=(E, D, 2 * F),
                 dtype=self.dtype,
                 sharding=self.efd_sharding,
                 random_init=self.random_init)
             self.kernel_down_proj_EDF = create_param(
```
bzgoogle: the same as above

tpu_inference/layers/jax/moe/moe.py
jrplatin: Fixed thanks!

PR #1568, opened 2026-01-29
[CI] Fix the MultiModalKwargsItem to align with vllm changes

kwang3939:
# Description

Fix the CI failure caused by: https://github.com/vllm-project/vllm/pull/33331.

Rm redundant logic on `req_id` in `multimodal_manager.py`.

# Tests

Tested locally with `pytest -v -s tests/runner/test_multimodal_manager.py`, `bash tests/e2e/benchmarking/mm_bench.sh`.

# Checklist

### DIFF
--- a/tpu_inference/runner/multimodal_manager.py
+++ b/tpu_inference/runner/multimodal_manager.py
@@ -92,7 +92,7 @@ class MultiModalManager:
             return
 
         # Batch the multi-modal inputs.
-        mm_kwargs = list[MultiModalKwargsItem]()
+        mm_kwargs = list[tuple[str, MultiModalKwargsItem]]()
         # List of tuple (mm_hash, pos_info)
         mm_hashes_pos = list[tuple[str, PlaceholderRange]]()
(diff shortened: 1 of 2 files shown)


patemotter (approved):

tpu_inference/runner/multimodal_manager.py
```diff
@@ -164,8 +164,6 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
                 mm_hashes_pos,
                 encoder_outputs,
         ):
-            if req_id not in self.runner.encoder_cache:
```
mrjunwan-lang, line 167: should we change to  if req_id in self.runner.encoder_cache: ??

kwang3939: I think this logic meant to check if`mm_hash` is included in the `self.runner.encoder_cache` instead of the last `req_id` from the earlier for loop. However, that should be created with the line `self.runner.encoder_cache[mm_hash] = scatter_mm_placeholders` even if it's not in the `self.runner.encoder_cache`.

PR #1571, opened 2026-01-29
[jax] Add `LoadableWithIterator` and `JaxAutoWeightsLoader` utils

lk-chen:
# Description

RFC: #1479 

This PR adds new utils for weight loading. image

In order for jax models work together with vLLM weight loading facilities, model needs to have `load_weights` that consumes iterator (through new `LoadableWithIterator`). Also through `JaxAutoWeightsLoader` we add custom per-param weight loader, instead of vLLM's [default](https://github.com/vllm-project/vllm/blob/dc917cceb877dfd13f98c538c4c96158047d98bd/vllm/model_executor/models/utils.py#L219)

# Tests

New unit tests added

# Checklist

### DIFF
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -714,15 +714,41 @@ class StandardWeightLoader(BaseWeightLoader):
             keep_hf_weight_suffix_when_match=keep_hf_weight_suffix_when_match)
 
 
-def load_nnx_param_from_torch(jax_param: nnx.Param,
-                              torch_weight: torch.Tensor):
-    """Load a nnx.Param from a torch.Tensor."""
+def load_nnx_param_from_reshaped_torch(
+        jax_param: nnx.Param,
+        torch_weight: torch.Tensor,
+        *,
+        reshape_dims: Optional[tuple[int, ...]] = None,
+        permute_dims: Optional[tuple[int, ...]] = None,
+        param_name: str = "Unknown"):
+    """Load a nnx.Param from a torch.Tensor with reshaping and transposing.
+    
+    HuggingFace model almost always store linear layer weights with contracting dimension
+    last, and only support 1D/2D weight tensors. This function reshapes then transposes
+    the torch weight to match the jax_param shape before loading.
+
+    Args:
+        jax_param: The target nnx.Param to load the weight into.
+        torch_weight: The source torch.Tensor weight.
+        reshape_dims: Optional tuple specifying the shape to reshape the torch weight to before permutation. If None, no reshaping is applied.
+        permute_dims: Optional tuple specifying the permutation of dimensions. If None, no-op for 1D tensors and transpose for 2D tensors is applied.
+    """
+    if reshape_dims is not None:
+        torch_weight = torch_weight.reshape(reshape_dims)
+    if permute_dims is None and torch_weight.ndim == 2:
+        permute_dims = (1, 0)
+    if permute_dims is not None:
+        torch_weight = torch_weight.permute(*permute_dims)
+
+    assert tuple(torch_weight.shape) == jax_param.value.shape, \
+        f"Shape mismatch when loading weight '{param_name}': torch {torch_weight.shape} vs jax {jax_param.value.shape}"
+
     spec = jax_param.sharding
     if isinstance(jax_param.sharding, NamedSharding):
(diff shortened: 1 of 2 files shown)


tpu_inference/models/common/model_loader.py
```diff
@@ -200,9 +202,12 @@ def create_sharded_model():
         # Although the created model can already work, we still need to jit
         # the model creation again, otherwise the model forward will have
         # non-trivial overhead in PjitFunction.
-        with mesh:
+        with jax.set_mesh(mesh):
             loader = get_model_loader(vllm_config.load_config)
-            if isinstance(loader, RunaiModelStreamerLoader):
+            if isinstance(model, LoadableWithIterator):
```
jrplatin, line 207: Is this going to break non-refactored models? Did you test with something like Llama3?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -703,3 +712,61 @@ def load_weights(self,
...
+                              torch_weight: torch.Tensor):
+    """Load a nnx.Param from a torch.Tensor."""
+    spec = jax_param.sharding
+    if isinstance(jax_param.sharding, NamedSharding):
+        spec = jax_param.sharding.spec
+    elif isinstance(jax_param.sharding, SingleDeviceSharding):
+        spec = ()
+    mesh = jax_param.mesh if hasattr(jax_param, 'mesh') else None
+
+    jax_param.value = shard_put(t2j(torch_weight, use_dlpack=False).astype(
```
jrplatin, line 753: Might be silly, but is this `astype` safe? Wonder if it might be good to cehck if the `jax_param` and `torch_weight` have the equivalent dtype

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -703,3 +712,61 @@ def load_weights(self,
...
+        jax_param.value.dtype),
+                                spec,
+                                mesh=mesh)
+
+
+def load_nnx_param_from_reshaped_torch(jax_param: nnx.Param,
```
jrplatin: Nit: do you think it might be cleaner just to consolidate to `load_nnx_param_from_torch` and add `reshape` as an arg?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -703,3 +712,61 @@ def load_weights(self,
...
+    """Mixin for models that support loading weights with an iterator.
+    
+    This is replicating what vLLM does for most models, e.g. https://github.com/vllm-project/vllm/blob/8e2a469b3b2f67bc900ed72724fe3f05e3564994/vllm/model_executor/models/gemma3_mm.py#L644-L646
+    """
+
+    def load_weights(self, weights: Iterable[tuple[str,
+                                                   torch.Tensor]]) -> set[str]:
+        if not isinstance(weights, Iterable):
+            # Use next parent class in MRO.
+            return super().load_weights(weights)
```
jrplatin, line 808: This class isn't inherting from anything? How would `super` work?

lk-chen: it's how [MRO](https://docs.python.org/3/howto/mro.html) works, `super()` is just next class in MRO, specifically
```
class A(B, LoadableWithIterator):
  ...
```
super() refers to B.

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: yes, done

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: Good point, I honestly don't know. Will revisit when hitting such case, as of this PR no model is using this path.

tpu_inference/models/common/model_loader.py
lk-chen: only when llama3 inherits `LoadableWithIterator` this path will be used, so no model hits this path as of now

lk-chen:
Hi @jrplatin thanks for the comments, could you take another look? After this one merged, I'll rebase #1586 and send for review.

tpu_inference/models/jax/utils/weight_utils.py
jrplatin: Ohhhh super interesting, thanks for explaining!

PR #1573, opened 2026-01-30
[Bugfix] Fix monolithic select

kyuyeunk:
# Description

- Fix error due to https://github.com/vllm-project/vllm/pull/32954 by adding `_select_monolithic` in `VllmUnquantizedFusedMoEMethod` 

- Remove `ShardingAxisName.TENSOR` that was introduced in https://github.com/vllm-project/tpu-inference/pull/1468 and replaced it with pre-existing sharding annotation `ShardingAxisName.MLP_TENSOR` that is more widely used across code base.

- Additionally, fix sharding annotation where dp sharding annotation was not added in some pre-compilation cases.

# Tests

commands to reproduce.

# Checklist


mrjunwan-lang:
Thanks for the fix!

PR #1574, opened 2026-01-30
Pin the vllm hash to a vllm hash to minitgate test issue.

QiliangCui:
# Description

Pin a hash to mitigate unit test issue


lk-chen:
I made #1575 which doesn't hardcode commit in docker, because developer also need to use the right commit

Issue #1570, opened 2026-01-29
[RFC] pin vLLM revision


lk-chen:
### 🚀 The feature, motivation and pitch

Current CI/CD pulls vLLM from latest main, which often breaks. This
* is affecting dev. velocity by a lot. Examples are #1568 #1543 #1521 etc.
* also makes it hard to reproduce a historical tpu-inference commit, since it's hard to link which tpu-inference commit works with which vllm commit

Proposal:
1. pin the vllm revision, store in a file
2. update [buildkite scripts](https://github.com/search?q=repo%3Avllm-project%2Ftpu-inference%20VLLM_COMMIT_HASH&type=code) to load from above file
3. update contribution doc to ask developers to install the specific revision
4. periodically (ideally everyday) update the pinned revision

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

lk-chen:
With #1575 merged, this RFC can be closed as completed. Following steps are:
* periodically roll forward
* evolve the process to an automated process

PR #1575, opened 2026-01-30
[CI] pin vllm revision

lk-chen:
# Description

See RFC: #1570 

# Tests

As shown in below screenshot, buildkite is checking out commit specified in `.env` instead of latest
image

# Checklist

### DIFF
--- a/.buildkite/vllm_lkg.version
+++ b/.buildkite/vllm_lkg.version
@@ -1,1 +1,1 @@
-d334dd26c4972312d9dc4f04dc7c93cdcacbd5a4
+1a7894dbdfd025da846b0bbfd19ecc4c6e9fcc1b


.env
kyuyeunk, line 1: any particular reason for using .env file? .env file isn't really an obvious place where people expect a dependency commit id to exists. is there a way to utilize requirements.txt file? i think that's where people expect to find any dependency related logics.

.env
```diff
@@ -0,0 +1 @@
+VLLM_COMMIT_HASH=d334dd26c4972312d9dc4f04dc7c93cdcacbd5a4
```
QiliangCui: I think right now the test is good. so, no need to do this? set to empty?

QiliangCui: i assume this for testing?

.buildkite/scripts/bootstrap.sh
```diff
@@ -20,6 +20,13 @@ set -euo pipefail
 echo "--- :git: Checking changed files"
 
 BASE_BRANCH=${BUILDKITE_PULL_REQUEST_BASE_BRANCH:-"main"}
+# load VLLM_COMMIT_HASH from .env file, if not exists, get the latest commit hash from vllm repo
+if [ -f .env ]; then
```
QiliangCui: load from https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/vllm_lkg.version

then, this work is merged to @weiyu0824 's work. 

He has a pipeline that should updating this file

.buildkite/scripts/commit_verified_commit_hashes.sh
```diff
@@ -39,7 +39,13 @@ git fetch origin "${TARGET_BRANCH}"
 git checkout "${TARGET_BRANCH}"
 git reset --hard origin/"${TARGET_BRANCH}"
 
-VLLM_COMMIT_HASH=$(buildkite-agent meta-data get "VLLM_COMMIT_HASH" --default "")
+# load VLLM_COMMIT_HASH from .env file, if not exists, get the latest commit hash from vllm repo
+if [ -f .env ]; then
```
QiliangCui: need to load from file? 

I think only bootstrap.sh needs to load it and then the value is passed to other places. no?

.buildkite/scripts/setup_docker_env.sh
```diff
@@ -73,11 +73,20 @@ setup_environment() {
 
   echo "Cleanup complete."
 
+  # load VLLM_COMMIT_HASH from .env file
+  if [ -f .env ]; then
```
QiliangCui: same here.

i think this value can be loaded from one place, like bootstrap.

.env
lk-chen: I thought it's very typical to store environment variables in `.env`. But @QiliangCui told me there's already vllm_lkg.version file in the repo, I'll use that.

.buildkite/scripts/setup_docker_env.sh
lk-chen: reverted changes in this file. I think I misread the lines previously

.env
lk-chen: I still want to pin a revision in this PR.

.buildkite/scripts/commit_verified_commit_hashes.sh
lk-chen: reverted this file change

.buildkite/scripts/bootstrap.sh
lk-chen: Done, updated to load from that file

QiliangCui (approved):

PR #1598, opened 2026-02-02
Bump Jax to 0.9.0, flax to 0.12.3 and qwix to 0.1.4

patrickji2014:
# Description

Latest libtpu and jax support Ironwood support so we can remove requirements_v7x.txt which use jax 0.8.1 etc. There are other dependencies using 0.8.1 and we have to migrate to jax 0.9.0 instead. 

- A follow up of this PR is investigating whether we need gather/scatter_mm_placeholders after there are removed in https://github.com/vllm-project/vllm/pull/30475/changes . From this PR, we might be able to gain more memory for KV cache. 

# Tests

Run on CI.

# Result

Wait for result.

### DIFF
--- a/tpu_inference/layers/jax/attention/deepseek_v3_attention.py
+++ b/tpu_inference/layers/jax/attention/deepseek_v3_attention.py
@@ -222,8 +222,8 @@ class MLA(nnx.Module):
         """
         md = attention_metadata
         x = jnp.asarray(x, self.dtype)
-        x_SD = nnx.with_sharding_constraint(x, self.activation_attention_td)
-        x_q_TD = nnx.with_sharding_constraint(x, self.activation_q_td)
+        x_SD = jax.lax.with_sharding_constraint(x, self.activation_attention_td)
+        x_q_TD = jax.lax.with_sharding_constraint(x, self.activation_q_td)
 
         with jax.named_scope("q_proj"):
             # Query down projection.
@@ -242,12 +242,12 @@ class MLA(nnx.Module):
                 # Absorb the k up-projection matrix into q
                 q_TNA = jnp.einsum("TNH,ANH -> TNA", q_nope_TNH,
                                    self.kernel_k_up_proj_ANH.value)
-                q_TNA = nnx.with_sharding_constraint(q_TNA, self.query_tnh)
+                q_TNA = jax.lax.with_sharding_constraint(q_TNA, self.query_tnh)
             else:
                 # Concatenate the nope and rope queries.
                 q_TNH = jnp.concatenate([q_nope_TNH, q_rope_TNH], axis=-1)
                 # Multiply the query by scaling factor
-                q_TNH = nnx.with_sharding_constraint(q_TNH, self.query_tnh)
+                q_TNH = jax.lax.with_sharding_constraint(q_TNH, self.query_tnh)
 
         with jax.named_scope("kv_proj"):
             # KV down projection.
@@ -263,7 +263,7 @@ class MLA(nnx.Module):
 
             kv_SA = kv_SA[..., :self.kv_lora_rank]
             kv_SA = self.kv_rms_norm(kv_SA)
-            kv_SA = nnx.with_sharding_constraint(kv_SA, self.keyvalue_skh)
+            kv_SA = jax.lax.with_sharding_constraint(kv_SA, self.keyvalue_skh)
 
             if not self.use_mla_kernel:
                 k_rope_SNH = jnp.broadcast_to(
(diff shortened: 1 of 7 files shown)


lk-chen:
🤣 wait.. why does the description say it's signed off by me?

tpu_inference/runner/tpu_runner.py
```diff
@@ -295,6 +295,7 @@ def _init_mesh(self) -> None:
             self.mesh = self._create_2d_mesh()
 
         logger.info(f"Init mesh | mesh={self.mesh}")
+        jax.set_mesh(self.mesh)
```
kyuyeunk, line 298: random thought: with this single line of code, i think this means that we don't have to carry around `mesh` variable across different data structures / function arguments like these: https://github.com/vllm-project/tpu-inference/blob/28521ee8fbccee7f5d625c4e8690c24cfdde9d39/tpu_inference/layers/vllm/attention.py#L182

and just use `get_concrete_mesh` or `get_abstract_mesh` to get our current mesh like this: https://github.com/jax-ml/jax/blob/a5854e44cb01c1b6aa24d43ab997e0ed77d8dab6/tests/shard_map_test.py#L4200-L4214

kyuyeunk (commented):
can you elaborate more on how this fix works?

also, in the description, attach the doc in the error message for more info: https://flax.readthedocs.io/en/latest/flip/4844-var-eager-sharding.html

tpu_inference/runner/multimodal_manager.py
```diff
@@ -30,6 +29,51 @@
     from tpu_inference.runner.tpu_runner import TPUModelRunner
 
 
+# TODO(patrickji): Investigate whether we can deprecate scatter_mm_placeholders and gather_mm_placeholders
```
jrplatin, line 32: I don't think this is in-scope for this PR -- I'd delegate this to the on-call to triage, and thanks for the effort!

kyuyeunk:
> > random thought: with this single line of code, i think this means that we don't have to carry around `mesh` variable across different data structures / function arguments like these: https://github.com/vllm-project/tpu-inference/blob/28521ee8fbccee7f5d625c4e8690c24cfdde9d39/tpu_inference/layers/vllm/attention.py#L182
> > 
> > and just use `get_concrete_mesh` or `get_abstract_mesh` to get our current mesh like this: https://github.com/jax-ml/jax/blob/a5854e44cb01c1b6aa24d43ab997e0ed77d8dab6/tests/shard_map_test.py#L4200-L4214
> 
> Yes, I think so. I am also wondering why we don't use set_mesh after creating mesh. we should have one static mesh during the whole life of one program run?

i had some concern whether this approach works for multi-host setup as well. but this concern isn't grounded in anything. @mrjunwan-lang, @Lumosis feel free to provide feedback.

tpu_inference/models/jax/qwen3.py
```diff
@@ -228,7 +228,7 @@ def __init__(self, vllm_config: VllmConfig, rng: nnx.Rngs,
             rngs=rng,
             quant_config=vllm_config.quant_config,
         )
-        self.layers = [
+        self.layers = nnx.List([
```
lk-chen, line 231: if you change list to nnx.List here, [this part](https://github.com/vllm-project/tpu-inference/blob/99573b74b10ce335cdf571008daf369565baf828/tpu_inference/layers/jax/__init__.py#L66-L67) might fail... 

Could you rebase to main and run CI to check?

tpu_inference/runner/multimodal_manager.py
patrickji2014: This is resolved in another PR now. Will sync to head

tpu_inference/models/jax/qwen3.py
patrickji2014: Thanks, I add support/checking for nnx.List and it works fine to generate the name.

kyuyeunk:
@patrickji2014, any update? i was hoping to start using updated jax version soon.

patrickji2014:
> @patrickji2014, any update? i was hoping to start using updated jax version soon.

Due to large versions bumps, there are quite some change need to be made. I figure out that it might be easier to update flax to latest and jax to less latest version first(first phase), then upgrade jax to latest version later.

The first phase is jax ==0.8.1 flax==0.12.3
The second phase will be jax=0.9.0 and flax = 0.12.3

What jax version you need?

patrickji2014:
Blocked by this RP https://github.com/vllm-project/tpu-inference/pull/1641. will continue work on this after the PR merged.

patrickji2014:
this work is broke down to multiple smaller pr and they are about to merge. close!

PR #1605, opened 2026-02-03
Implement incremental weight loading and sharding for vLLM models

Lumosis:
# Description

This PR introduces an incremental weight loading mechanism to significantly reduce peak CPU memory usage during model initialization.

Currently, the default loading behavior loads all model weights into CPU RAM before processing and sharding them to the TPU. For large models, this can lead to Host OOMs. This change intercepts the weight loading process to shard and move weights to the TPU (freeing the CPU memory) immediately after a specific layer's parameters are fully loaded.

## Key Changes:

* [tpu_inference/models/vllm/vllm_model_loader.py]:
   * Added IncrementalModelLoader: A new model loader registered under the name "incremental".
   * Added attach_incremental_weight_loader: A utility that traverses the model and wraps the existing weight_loader of parameters.
   * Logic: The wrapper tracks which parameters have been loaded for a specific layer using layer._loaded_weights. Once all required parameters (e.g., weight + bias for Linear, or all experts for MoE) are loaded, it triggers the sharding logic immediately.
 * Sharding Logic Integration:
    * Utilizes process_unquantized_linear_weight for LinearBase layers (handling QKVParallelLinear specifically).
    * Utilizes process_unquantized_moe_weight for FusedMoE layers.

# Tests

Local tested with
* Qwen3-Coder-480B-A35B-Instruct with incremental weight loader
* Qwen/Qwen3-235B-A22B with incremental weight loader
* Qwen3-Coder-480B-A35B-Instruct-FP8 without incremental weight loader
* unsloth/gpt-oss-120b-BF16  without incremental weight loader

# Checklist


tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,127 @@
...
+            if original_loader is None:
+                continue
+            setattr(
+                param, "weight_loader",
+                create_weight_loader(module, original_loader, name,
+                                     param_name))
+
+
+@register_model_loader("incremental")
+class IncrementalModelLoader(DefaultModelLoader):
```
lk-chen, line 68: Could you add comments why we need a dedicated model loader? Is it all the purpose of calling `attach_incremental_weight_loader` before loading?

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    def create_weight_loader(layer, original_loader, layer_name, param_name):
+
+        def weight_loader_wrapper(param: torch.nn.Parameter,
+                                  loaded_weight: torch.Tensor, *args,
+                                  **kwargs):
+            # Loading the weight
+            res = original_loader(param, loaded_weight, *args, **kwargs)
+            print(
```
lk-chen: remove or using logger.debug

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -95,50 +91,10 @@ def __init__(self, linear_config: VllmQuantLinearConfig):
...
-            if bias is not None:
-                layer.bias = Parameter(weights.bias, requires_grad=False)
-        else:
-            layer.weight = to_parameter_list(weights.weight)
-            if bias is not None:
-                layer.bias = to_parameter_list(weights.bias)
+        if not _tensor_is_in_cpu(layer.weight):
+            # Already processed and sharded.
+            return
+        process_unquantized_linear_weight(layer)
```
lk-chen: This PR moves utility functions like `process_unquantized_linear_weight` to a central place (cleanup_sharding.py) which I think is not scalable. I assume the reason is you wan to call `process_unquantized_linear_weight` after a layer is loaded? see my [comment](https://github.com/vllm-project/tpu-inference/pull/1605/changes#r2760713379)

Instead you can:

1. at L78, pass `layer` into `VllmUnquantizedLinearMethod`
2. at L91, assign custom `load_weights` to `layer`, which is used [here](https://github.com/vllm-project/vllm/blob/61e632aea15f76fd1c46354b00f9cac62cd28c4e/vllm/model_executor/models/utils.py#L261)
3. in the custom `load_weights`, you call `process_unquantized_linear_weight` at the end (after all params loaded)

kyuyeunk:
i think this pr needs coordination with this another pr by @jrplatin because they both touches weight loading logic: https://github.com/vllm-project/tpu-inference/pull/1610

lk-chen:
> i think this pr needs coordination with this another pr by @jrplatin because they both touches weight loading logic: #1610

@kyuyeunk I'm overseeing both, #1610 is focusing on refactoring and shouldn't touch any vllm-path weight loading.
While this PR is touching vllm-path only

tpu_inference/layers/vllm/quantization/unquantized.py
kyuyeunk: yers. i have a similar concern as well. moving the logic into the centralized place will not be scalable & we should prefer modular / interface based approach where each classes only have to care about implementing apis that will be invoked by model loader.

also, @Lumosis, you probably already mentioned this during our design discussion, but what was the reason for overriding `weight_loader` instead of `load_weights` like @lk-chen has suggested here?

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,109 @@
...
+from vllm.utils.torch_utils import set_default_torch_dtype
+
+from tpu_inference.layers.vllm.quantization.unquantized import (
+    VllmUnquantizedFusedMoEMethod, VllmUnquantizedLinearMethod)
+
+
+def attach_incremental_weight_loader(model: torch.nn.Module) -> None:
+    """
+    Traverses the model and overrides the weight_loader of each parameter to support incremental loading.
+    This allows processing and sharding of weights after all weights for a parameter have been loaded.
```
lk-chen: "all weights for a **module**"

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,109 @@
...
+        def weight_loader_wrapper(param: torch.nn.Parameter,
+                                  loaded_weight: torch.Tensor, *args,
+                                  **kwargs):
+            # Loading the weight
+            res = original_loader(param, loaded_weight, *args, **kwargs)
+
+            # Processing and sharding
+            # For now, only handle unquantized linear and moe layers.
+            quant_method = layer.quant_method
+            if (isinstance(layer, LinearBase) and isinstance(
```
lk-chen: just need to check `hasattr(quant_method, "maybe_process_weights")`

Lumosis: Good idea

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,101 @@
...
+            original_loader = getattr(param, "weight_loader", None)
+            if original_loader is None:
+                continue
+            setattr(
+                param, "weight_loader",
+                create_weight_loader(module, original_loader, name,
+                                     param_name))
+
+
+@register_model_loader("incremental")
```
lk-chen: tpu-streaming-loader

lk-chen (approved):

tpu_inference/layers/vllm/quantization/unquantized.py
Lumosis: Offline discussed. The current design looks good.

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,101 @@
...
+        def weight_loader_wrapper(param: torch.nn.Parameter,
+                                  loaded_weight: torch.Tensor, *args,
+                                  **kwargs):
+            # Loading the weight
+            res = original_loader(param, loaded_weight, *args, **kwargs)
+
+            # Processing and sharding
+            # For now, only handle unquantized linear and moe layers.
+            quant_method = layer.quant_method
+            if hasattr(quant_method, "maybe_process_weights"):
```
kyuyeunk: not sure if this string based checking is a good design? maybe a better design is to define an class with abstractmethod with `maybe_process_weights` & each `quant_method` defined in tpu-inference will inherit from it and needs to implement its api. and on the weight loader, `quant_method` is an instance of that abstractmethod class, we can invoke the function. wdyt?

tpu_inference/models/vllm/vllm_model_loader.py
```diff
@@ -0,0 +1,101 @@
...
+
+        return weight_loader_wrapper
+
+    for name, module in model.named_modules():
+        # Weight loader will be invoked multiple times for module. In order to determine when all the weights are loaded,
+        # we need to keep track of the loaded weights for each module.
+        module._loaded_weights = set()
+        for param_name, param in module.named_parameters(recurse=False):
+            # Omit parameters that do not have a weight_loader
+            original_loader = getattr(param, "weight_loader", None)
```
kyuyeunk, line 58: similar to above, not really a fan of string based checking. maybe a better approach is to check if `param` is an instance of `BasevLLMParameter` and if it is, invoke the weight loader? https://github.com/vllm-project/vllm/blob/72bb24e2db2acd98a49adcb9e3f1dc6f1bbef4c0/vllm/model_executor/parameter.py#L69

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -94,14 +96,36 @@ class VllmUnquantizedLinearMethod(vllm_linear.UnquantizedLinearMethod,
...
         super().__init__(linear_config)
 
+    def maybe_process_weights(self, layer: torch.nn.Module, param_name: str,
+                              args, kwargs):
+        """Check if all weights are loaded for the layer. If so, process and shard the weights."""
+        if isinstance(layer, vllm_linear.QKVParallelLinear):
+            assert len(args) == 1, "Expecting shard_id as the only argument"
+            shard_id = args[0]
+            # Keep track of loaded weights for QKVLinear, e.g. (('weight', 'q'), ('bias', 'q'), ('weight', 'k'), ('bias', 'k'), ...)
+            layer._loaded_weights.add((param_name, shard_id))
```
lk-chen, line 108: nit, you can just
```
layer._loaded_weights.add(args)
```
without type checking

Lumosis: We need to include param_name (e.g., weight or bias) to ensure uniqueness. Otherwise, if we only use q, k, and v, identical elements will be merged because we are using a set.

tpu_inference/models/vllm/vllm_model_loader.py
Lumosis: Agreed. Done.

tpu_inference/models/vllm/vllm_model_loader.py
Lumosis: Here we cannot do it due to the inconsistency in vllm codebase. E.g. bias is not a BasevLLMParameter: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/linear.py#L510

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -88,20 +99,51 @@ def get_quant_method(self, layer: torch.nn.Module,
         return None
 
 
+class VllmTpuUnquantizedMethod(ABC):
```
kyuyeunk: move this to this file so that other quantization methods (like fp8 quant methods) can also utilize this?: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/quantization/__init__.py

kyuyeunk: and rename it to be more generic like `VllmQuantizationMethod`

Lumosis: Done

kyuyeunk (approved):
Looks really good!

PR #1610, opened 2026-02-03
[JAX][MoE] Refactor JAX MoE layer + consolidate common TorchAX/JAX logic + add UnquantizedMoE Scaffolding

jrplatin:
# Description

In this PR, I do the following
* Consolidate to a single `select_moe_backend` method (the existing one in the TorchAX path) and add testing in `tests/layers/common/test_moe.py`
* Move `tpu_inference/layers/vllm/process_weights/fused_moe_weights.py` to `tpu_inference/layers/common/process_weights/moe_weights.py` , **context**: we want to share these utils between Jax/TorchAX + I added an UnfusedMoEWeights class since the JAX path has a few paths that don't want to fuse weights on (for now)

* Move `‎tpu_inference/layers/common/moe.py‎tpu_inference/layers/vllm/fused_moe.py`  to `tpu_inference/layers/common/moe.py` , **context**: we also want to share the logic in this file between JAX/TorchAX.  I also renamed FusedMoEBackend to MoEBackend since some of the JAX path backends aren't fused (neither weights nor full MoE operations), and updated select_moe_backend to take use_ep as a bool since the JAX path doesn't have an equivalent FusedMoEConfig for now.  Also, rename `fused_moe_apply` to `moe_apply`.

* `tpu_inference/layers/jax/moe/moe.py`: rename `MoE` to `JaxMoE` and add scaffolding of `quant_method` logic that will be wired in a forthcoming PR.

* `tpu_inference/layers/jax/quantization/unquantized.py` add skeleton of `UnquantizedFusedMoEMethod`

# Tests

Checked that `offline_inference` produces the exact same results on the TorchAX path for both FP8/Compressed-Tensor MoE models:

```
python examples/offline_inference.py --model=Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 --tensor-parallel-size=2

Prompt: 'Hello, my name is'
Generated text: ' Wang Yujie. I am a postgraduate student at the School of Computer'
Generated text: ' ____\nA. London\nB. Berlin\nC. Rome\nD.'
Generated text: ' red, orange, yellow, green, blue, indigo, and violet.'
Prompt: 'The future of AI is'
Generated text: " not just about creating smarter machines; it's about ensuring that these machines remain beneficial"
Generated text: ' the head of the executive branch of the federal government of the United States. The'



MODEL_IMPL_TYPE=vllm python examples/offline_inference.py --model=RedHatAI/Qwen3-30B-A3B-FP8-dynamic  --tensor-parallel-size=2
Prompt: 'Hello, my name is'
Generated text: ' Emily. I am a 30-year-old woman. I used to be'
Generated text: ' __________.\nA. Paris\nB. London\nC. Berlin\nD'
Generated text: ' red, orange, yellow, green, blue, indigo, and violet.'
Prompt: 'The future of AI is'
Generated text: " not just about creating more advanced algorithms or more powerful hardware. It's about how"
Generated text: ' the commander-in-chief of the armed forces, and the military is the main tool'
```

Also checked Megablocks + FP8 + JAX path for 10 layers (just to be safe, even though the `quant_method` isn't hooked up):

```
USE_MEGABLOCKS=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=jrplatin/DeepSeek-R1-1D-Subchannel-256-FP8 --max-model-len=5120 --tensor-parallel-size 8 --max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching  --gpu-memory-utilization 0.95 --additional_config='{"sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' --hf_overrides  '{"num_hidden_layers": 10}'

Prompt: 'Hello, my name is'
Generated text: ' Luggagereenwaldorf-language Artsgeanfreemdership Sagarothalamidusapes'
Generated text: ' Coastal Plain Jagdoms offendahirdur gapsutersUIToolseden涎滞留_religious'
Generated text: ' bells缭乱的 multiplyingabasjeevikstal Bharat Swap的电影Tai McNabb二郎 grounding'
Prompt: 'The future of AI is'
Generated text: 'ouwాలుIntegratedCircuitousness borrowed}}_{\\ruzpillarayan Faberrile tapseled'
Generated text: 'bauer组长出不 Acidifizierten472 TwistCabinet inspurerbbeosite渐变udas'
```

I also tested that offline inference outputs match for DeepSeek w/ MLA, vLLM backend, and no quantization:

```



NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=deepseek-ai/DeepSeek-R1 --max-model-len=5120 --tensor-parallel-size 8 --max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching  --gpu-memory-utilization 0.95 --additional_config='{"skip_quantization": true, "sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' --hf_overrides  '{"num_hidden_layers": 5}'

(EngineCore_DP0 pid=925639) WARNING 02-03 15:06:00 [tpu_runner.py:701] Should not schedule a request that does nothing!
Processed prompts: 100%|██████████████████████████████████████████████████████████████████████| 35/35 [00:13<00:00,  2.69it/s, est. speed input: 28.09 toks/s, output: 42.98 toks/s]
Prompt: 'Hello, my name is'
Generated text: 'Entreprise060 Volt熟的 Blaodil碍deb一圈儿的可乐ptusasma Dragons综合治理'
Generated text: 'icro517.lib здравerdquoanatysisجادdestructive aptitudeshaiubuenas'
Generated text: '� anthrop962nevolved stratum Bib491asto马拉松这把骚 auton disorderly operable'
Prompt: 'The future of AI is'
Generated text: 'etteersh980ouzsko-focusedonti-Ammaryerusicalsopteriore点缀 tapsulos'
Generated text: '046人和ocated181matical jurisprudence蜡带回ETTER unrealistic imaginations Chavezము Frojekecd'
```

and DeepSeek w/ MHA, VLLM Fused Kernel, and no quantization:

```
  VLLM_MLA_DISABLE=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=deepseek-ai/DeepSeek-R1 --max-model-len=5120 --tensor-parallel-size 8 --max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching  --gpu-memory-utilization 0.95 --additional_config='{"skip_quantization": true, "sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' --hf_overrides  '{"num_hidden_layers": 10}'

Prompt: 'Hello, my name is'
Generated text: 'hn433 Sheldon recovering bek YugDepth ShieldEnsembleIron Generator葵 Valenciarokము'
Generated text: 'Ketoman shedsova Archaeologicalbzadehlap Mawrok solicrok福祉reet非凡 annotated'
Generated text: '一方Soupupl灵芝CSS Gudeki Forecast weathered banner Pagesrix Calder� Gatebro'
Prompt: 'The future of AI is'
Generated text: 'rozOMO军在后者在前几年的ZOleo行礼 JJ_COMP巧TeXillos pardon Eng'
Generated text: 'rokeKarlsic别无UTE materia Verjekedes访erschuzeanaloty spoil Married'
```

# Checklist

### DIFF
--- a/docs/developer_guides/jax_model_development.md
+++ b/docs/developer_guides/jax_model_development.md
@@ -59,6 +59,8 @@ class NewModel(nnx.Module):
                mesh: jax.sharding.Mesh)
 ```
 
+"# TODO (jacobplatin): we need to update this once the JAX-path refactor is fully done"
+
 The constructor should set the architecture configuration (e.g. num_layers, hidden_size) and initialize the model layers. Layers can be defined from scratch using flax NNX (e.g. [Llama3](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/models/jax/llama3.py)) or can leverage tpu-inference to import or extend commonly used layer types (e.g. [Embedder](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/layers.py#L168), [RMSNorm](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/layers.py#L49), [MoE](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/moe/moe.py#L69), [Attention](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/attention/attention.py#L23), [DenseFFW](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/layers.py#L98C7-L98C15), [TransformerBlock](https://github.com/vllm-project/tpu-inference/blob/31fa76a0187496ec161c634c98ac5eba144cb36c/tpu_inference/layers/jax/transformer_block.py#L15
 )).
 
--- a/tpu_inference/layers/common/moe.py
+++ b/tpu_inference/layers/common/moe.py
@@ -13,61 +13,62 @@
 # limitations under the License.
 
 from enum import Enum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
 import jax
 import jax.numpy as jnp
 from jax.sharding import Mesh
 from vllm.model_executor.layers.fused_moe import FusedMoE
-from vllm.model_executor.layers.fused_moe.config import FusedMoEConfig
 
-from tpu_inference import envs
 from tpu_inference.kernels.fused_moe.v1.kernel import fused_ep_moe
 from tpu_inference.layers.common.fused_moe_gmm import fused_moe_func
 from tpu_inference.logger import init_logger
 
 if TYPE_CHECKING:
-    from tpu_inference.layers.vllm.process_weights.fused_moe_weights import \
-        FusedMoEWeights
+    from tpu_inference.layers.common.process_weights.moe_weights import (
+        FusedMoEWeights, UnfusedMoEWeights)
+    from tpu_inference.layers.jax.moe.moe import JaxMoE
+else:
+    FusedMoEWeights = None
+    UnfusedMoEWeights = None
+    JaxMoE = None
 
 logger = init_logger(__name__)
 
 
-class FusedMoEBackend(Enum):
+class MoEBackend(Enum):
+    # This is using the Fused MoE kernel found in tpu-inference/tpu_inference/kernels/fused_moe/v1/kernel.py
+    # and is production ready.
     FUSED_MOE = "fused_moe"
+    # This is using the GMM kernel found in tpu-inference/tpu_inference/layers/common/fused_moe_gmm.py
+    # and is `expert_sharded_gmm` the GMM calls.  Production ready.
     GMM_EP = "gmm_ep"
+    # Same as GMM_EP but uses `tensor_sharded_gmm_*` for the GMM calls and is production ready
     GMM_TP = "gmm_tp"
+    # Uses a simple dense matmul for the MoE backend and is intended for testing
+    # Only used in the JAX path for now
+    DENSE_MAT = "dense_mat"
+    MEGABLX_GMM = "megablox_gmm"  # only used in the JAX path for now
+
+    @classmethod
+    def fused_moe_backends(cls):
+        """Returns those backends that use fused weights"""
+        return {cls.FUSED_MOE, cls.GMM_EP, cls.GMM_TP}
 
 
-def select_moe_backend(moe: FusedMoEConfig):
-    if envs.USE_MOE_EP_KERNEL:
-        if moe.use_ep:
-            return FusedMoEBackend.FUSED_MOE
-        logger.warning_once(
-            "USE_MOE_EP_KERNEL=1 but expert parallelism is not "
-            "enabled. Falling back to gmm implementation.")
-
-    if moe.use_ep:
-        return FusedMoEBackend.GMM_EP
-
-    # Use default implementation.
-    return FusedMoEBackend.GMM_TP
-
-
-def fused_moe_apply(
-    layer: FusedMoE,
+def moe_apply(
+    layer: FusedMoE | JaxMoE,
     x: jax.Array,
     gating_output: jax.Array,
-    weights: "FusedMoEWeights",
-    moe_backend: FusedMoEBackend,
+    weights: Union[FusedMoEWeights, UnfusedMoEWeights],
+    moe_backend: MoEBackend,
     mesh: Mesh,
     extra_backend_kwargs: dict,
 ) -> jax.Array:
-    assert isinstance(layer, FusedMoE)
 
     with jax.named_scope(layer._get_name()):
         match moe_backend:
-            case FusedMoEBackend.FUSED_MOE:
+            case MoEBackend.FUSED_MOE:
                 subc_quant_w1_sz = None
                 subc_quant_w2_sz = None
                 if weights.w13_weight_scale is not None and weights.w2_weight_scale is not None:
@@ -105,7 +106,7 @@ def fused_moe_apply(
                     b2=weights.w2_bias,
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
-            case FusedMoEBackend.GMM_EP | FusedMoEBackend.GMM_TP:
+            case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
                 output = fused_moe_func(
                     hidden_states=x,
                     w1=weights.w13_weight,
@@ -122,5 +123,12 @@ def fused_moe_apply(
                     activation=layer.activation,
                     scoring_fn=layer.scoring_func,
                 )
+            case MoEBackend.DENSE_MAT:
+                # TODO(jacobplatin): will implement in forthcoming PR
+                raise NotImplementedError
+
+            case MoEBackend.MEGABLX_GMM:
+                # TODO(jacobplatin): will implement in forthcoming PR
+                raise NotImplementedError
 
         return output
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -20,11 +20,11 @@ from jax.experimental.layout import Format, Layout, with_layout_constraint
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.tensor import Tensor
 
+from tpu_inference.layers.common.moe import MoEBackend
 from tpu_inference.layers.common.quantization import quantize_tensor
 from tpu_inference.layers.common.sharding import ShardingAxisName
 from tpu_inference.layers.common.utils import \
     reorder_concatenated_tensor_for_sharding
-from tpu_inference.layers.vllm.fused_moe import FusedMoEBackend
 from tpu_inference.utils import align_to
 
 P = PartitionSpec
@@ -42,6 +42,21 @@ class FusedMoEWeights:
     w2_bias: jax.Array | Tensor | None
 
 
+@jax.tree_util.register_dataclass
+@dataclass
+class UnfusedMoEWeights:
+    """Unfused moe weights. weights can be either jax or torchax array."""
+    w1_weight: jax.Array | Tensor
+    w1_weight_scale: jax.Array | Tensor | None
+    w1_bias: jax.Array | Tensor | None
+    w2_weight: jax.Array | Tensor
+    w2_weight_scale: jax.Array | Tensor | None
+    w2_bias: jax.Array | Tensor | None
+    w3_weight: jax.Array | Tensor
+    w3_weight_scale: jax.Array | Tensor | None
+    w3_bias: jax.Array | Tensor | None
+
+
 def quantize_moe_weights(
     weights: FusedMoEWeights,
     dtype: jnp.dtype,
@@ -53,7 +68,7 @@ def quantize_moe_weights(
         weights: fused moe weights.
         dtype: dtype to perform quantization.
         block_size: Specify block quantization size. If non, use per-channel
-            quantization. If contracting dim is not divisible by block size, 
+            quantization. If contracting dim is not divisible by block size,
             the dim will be automatically padded and corresponding dim on bias
             and the other weight (w13_weight <-> w2_weight) is also padded.
 
@@ -111,7 +126,7 @@ def quantize_moe_weights(
 
 def process_moe_weights(
     weights: FusedMoEWeights,
-    moe_backend: FusedMoEBackend,
+    moe_backend: MoEBackend,
     w13_reorder_size: int | None = None,
     w13_interleave: bool = False,
 ) -> FusedMoEWeights:
@@ -121,8 +136,8 @@ def process_moe_weights(
         weights: fused moe weights.
         moe_backend: backend type the weights should be processed for.
         w13_reorder_size: only used when backend type is GMM_TP. in order to
-            eliminate collective operations when using tensor parallelism, 
-            group w13_weight into w13_reorder_size number of chuncks where each
+            eliminate collective operations when using tensor parallelism,
+            group w13_weight into w13_reorder_size number of chunks where each
             chunk stores both w1 and w3 weights.
         w13_interleave: used when loaded w13_weight is stored in interleaved
             pattern where even index element is w1 and odd index element is w3.
@@ -181,7 +196,7 @@ def process_moe_weights(
         w2_bias = jnp.expand_dims(w2_bias, 1)
 
     match moe_backend:
-        case FusedMoEBackend.FUSED_MOE:
+        case MoEBackend.FUSED_MOE:
             # Kernel expects:
             # w13: (num_experts, 2, hidden_size, intermediate_size)
             # w2: (num_experts, intermediate_size, hidden_size)
@@ -245,7 +260,7 @@ def process_moe_weights(
                     ((0, 0), (0, 0), (0, pad_width_hidden_size)),
                 )
 
-        case FusedMoEBackend.GMM_TP:
+        case MoEBackend.GMM_TP:
             assert w13_reorder_size is not None
             assert intermediate_size % w13_reorder_size == 0
             output_sizes = [intermediate_size, intermediate_size]
@@ -269,9 +284,20 @@ def process_moe_weights(
                     w13_reorder_size,
                     dim=2,
                 )
-        case FusedMoEBackend.GMM_EP:
+        case MoEBackend.GMM_EP:
             # No additional processing is needed for GMM_EP.
             pass
+
+        case MoEBackend.DENSE_MAT:
+            # TODO (jacobplatin)
+            raise NotImplementedError(
+                "process_moe_weights is not yet implemented for dense matmul backend."
+            )
+        case MoEBackend.MEGABLX_GMM:
+            # TODO (jacobplatin)
+            raise NotImplementedError(
+                "process_moe_weights is not yet implemented for megablox gmm backend"
+            )
 
     return FusedMoEWeights(
         w13_weight=w13_weight,
@@ -285,12 +311,12 @@ def process_moe_weights(
 
 def shard_moe_weights(
     weights: FusedMoEWeights,
-    moe_backend: FusedMoEBackend,
+    moe_backend: MoEBackend,
     mesh: Mesh,
 ) -> FusedMoEWeights:
 
     match moe_backend:
-        case FusedMoEBackend.FUSED_MOE | FusedMoEBackend.GMM_EP:
+        case MoEBackend.FUSED_MOE | MoEBackend.GMM_EP:
             ep_sharding = NamedSharding(mesh, P(ShardingAxisName.EXPERT))
             weight_shardings = FusedMoEWeights(
                 w13_weight=ep_sharding,
@@ -300,7 +326,7 @@ def shard_moe_weights(
                 w2_weight_scale=ep_sharding,
                 w2_bias=ep_sharding,
             )
-        case FusedMoEBackend.GMM_TP:
+        case MoEBackend.GMM_TP:
             # When using per-channel, in_dim // block_size == 1. This means we
             # are unable to shard w2_weight_scale along 1st dim. Therefore, we
             # fully replicate it instead.
@@ -336,7 +362,7 @@ def shard_moe_weights(
             )
 
     match moe_backend:
-        case FusedMoEBackend.FUSED_MOE:
+        case MoEBackend.FUSED_MOE:
             weight_layouts = FusedMoEWeights(
                 w13_weight=Layout((0, 1, 2, 3)),
                 w13_weight_scale=Layout((0, 1, 2, 3, 4)),
@@ -345,7 +371,7 @@ def shard_moe_weights(
                 w2_weight_scale=Layout((0, 1, 2, 3)),
                 w2_bias=Layout((0, 1, 2)),
             )
-        case FusedMoEBackend.GMM_TP | FusedMoEBackend.GMM_EP:
+        case MoEBackend.GMM_TP | MoEBackend.GMM_EP:
             weight_layouts = FusedMoEWeights(
                 w13_weight=Layout((0, 1, 2)),
                 w13_weight_scale=Layout((0, 1, 2, 3)),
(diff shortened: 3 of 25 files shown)


tpu_inference/layers/common/moe.py
```diff
@@ -13,61 +13,78 @@
...
 if TYPE_CHECKING:
-    from tpu_inference.layers.vllm.process_weights.fused_moe_weights import \
-        FusedMoEWeights
+    from tpu_inference.layers.common.process_weights.moe_weights import (
+        FusedMoEWeights, UnfusedMoEWeights)
+    from tpu_inference.layers.jax.moe.moe import JaxMoE
+else:
+    FusedMoEWeights = None
+    UnfusedMoEWeights = None
+    JaxMoE = None
```
patemotter, line 34: Can you explain this logic? Not quite sure I follow.

tpu_inference/layers/common/moe.py
```diff
@@ -122,5 +139,12 @@ def fused_moe_apply(
                     activation=layer.activation,
                     scoring_fn=layer.scoring_func,
                 )
+            case MoEBackend.DENSE_MAT:
+                # TODO(jacobplatin): will implement in forthcoming PR
+                raise ValueError
+
+            case MoEBackend.RAGGED_DOT | MoEBackend.MEGABLX_GMM:
+                # TODO(jacobplatin): will implement in forthcoming PR
+                raise NotImplementedError
```
patemotter: Nit: Any specific reason these two errors are different?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -98,7 +99,9 @@ def __call__(self, x_TD: Float):
                                       self.kernel_DE.value)
 
         #TODO: Refactor the Router so that it will always only return router_logits_TE
-        if self.moe_backend == MoEBackend.FUSED_MOE or self.moe_backend == MoEBackend.VLLM_MOE:
+        if self.moe_backend in [
+                MoEBackend.FUSED_MOE, MoEBackend.GMM_EP, MoEBackend.GMM_TP
+        ]:
```
patemotter: Can we make this list a part of the `MoEBackend` class itself as a `supported_moe_backends` to have it self-contained and reusable?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -160,6 +168,10 @@ class MoE(nnx.Module):
     # --- MoE Kernel Specific Attributes ---
     renormalize: bool = True
 
+    # ---- Quantization Specific Attributes ----
+    quant_config: Optional[QuantizationConfig] = None
+    prefix: str = ""
```
patemotter: Nit: I would rename to something like `quant_prefix` just because of how generic/overloaded `prefix` can be.

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -169,6 +181,10 @@ def __call__(self, x_TD: Float):
         Returns:
             Output array of shape (sequence_length, d_model) after passing through MoE.
         """
+        # TODO (jacobplatin): wire this up
```
patemotter: Nit: I know it's a todo, but for posterity I would define what "wire this up" means.

tpu_inference/layers/jax/moe/utils.py
```diff
@@ -286,3 +254,25 @@ def gmm_fn(inputs, kernel, group_sizes, tile_size, moe_backend, dtype,
     if pad_amount > 0:
         output = output[:num_rows, :]
     return output
+
+
+def get_expert_parallelism(expert_axis_name: str, mesh: Mesh) -> int:
+    """
+    Returns the expert parallelism number from the mesh.
+    This will likely be deprecated/refactored in another PR
+    (jacobplatin).
```
patemotter: Nit: Helpful to include a basic answer here to "Why?"

docs/developer_guides/jax_model_development.md
```diff
@@ -59,6 +59,7 @@ class NewModel(nnx.Module):
                mesh: jax.sharding.Mesh)
 ```
 
+# TODO (jacobplatin): we need to update this once the JAX-path refactor is fully done
```
lk-chen: it will become a title, I assume that's not what you wanted?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -149,8 +152,13 @@ class MoE(nnx.Module):
 
     # --- Flags & Configs ---
     apply_expert_weight_before_computation: bool
+    expert_axis_name: str
+    num_expert_parallelism: int
+    use_ep: bool
```
lk-chen: is it necessary to have both `use_ep` and `num_ep` flags? i.e. does `num_ep is None` infer `use_ep = False` ?

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -93,10 +92,10 @@ def __init__(self,
         super().__init__(weight_quant, input_quant, moe)
 
         self.mesh = mesh
-        self.moe_backend = select_moe_backend(self.moe)
+        self.moe_backend = select_moe_backend(self.moe.use_ep)
```
patemotter: Would it make sense to leave this as it was and change the logic inside `select_moe_backend`? I can imagine us quickly outgrowing the single boolean argument.

tpu_inference/models/jax/llama4.py
```diff
@@ -516,13 +516,17 @@ def __init__(self,
...
+            moe_ffw = JaxMoE(
                 dtype=dtype,
                 mesh=self.mesh,
                 num_local_experts=self.num_local_experts,
                 apply_expert_weight_before_computation=True,
                 hidden_size=self.hidden_size,
                 intermediate_size_moe=self.intermediate_size_moe,
+                expert_axis_name=None,
+                num_expert_parallelism=None,
+                # TODO (jacobplatin): maybe try both once/if FusedMoEKernel is ever supported
```
patemotter: Nit: Let's make this todo slightly more positive?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -121,7 +136,7 @@ def process_moe_weights(
         weights: fused moe weights.
         moe_backend: backend type the weights should be processed for.
         w13_reorder_size: only used when backend type is GMM_TP. in order to
-            eliminate collective operations when using tensor parallelism, 
+            eliminate collective operations when using tensor parallelism,
             group w13_weight into w13_reorder_size number of chuncks where each
```
patemotter: Supernit: Can we fix the spelling of "chunks" while we are here?

docs/developer_guides/jax_model_development.md
jrplatin: Oh good catch, thanks!

tpu_inference/layers/common/moe.py
jrplatin: Oh good catch, will consolidate to NIE

tpu_inference/layers/jax/moe/moe.py
jrplatin: Yep, will do!

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
jrplatin: I thought on this but the "issue" is that `FusedMoEConfig` doesn't have an equivalent for the JAX path right now, so I figured this was cleaner, but happy to discuss further

tpu_inference/layers/jax/moe/moe.py
jrplatin: I think so because otherwise we need to have a bunch of `self.num_expert_parallelism > 1` calls scattered throughout

tpu_inference/layers/common/moe.py
jrplatin: It's to get past a circular import error:

```
(EngineCore_DP0 pid=2148637)   File "/mnt/disks/jacobplatin/tpu-inference/tpu_inference/models/common/model_loader.py", line 60, in _get_model_architecture
(EngineCore_DP0 pid=2148637)     from tpu_inference.models.jax.deepseek_v3 import DeepSeekV3
(EngineCore_DP0 pid=2148637)   File "/mnt/disks/jacobplatin/tpu-inference/tpu_inference/models/jax/deepseek_v3.py", line 38, in <module>
(EngineCore_DP0 pid=2148637)     from tpu_inference.layers.common.moe import MoEBackend, select_moe_backend
(EngineCore_DP0 pid=2148637)   File "/mnt/disks/jacobplatin/tpu-inference/tpu_inference/layers/common/moe.py", line 27, in <module>
(EngineCore_DP0 pid=2148637)     from tpu_inference.layers.jax.moe.moe import JaxMoE
(EngineCore_DP0 pid=2148637)   File "/mnt/disks/jacobplatin/tpu-inference/tpu_inference/layers/jax/moe/moe.py", line 27, in <module>
(EngineCore_DP0 pid=2148637)     from tpu_inference.layers.common.moe import MoEBackend, fused_moe_func
(EngineCore_DP0 pid=2148637) ImportError: cannot import name 'MoEBackend' from partially initialized module 'tpu_inference.layers.common.moe' (most likely due to a circular import) (/mnt/disks/jacobplatin/tpu-inference/tpu_inference/layers/common/moe.py)
```

tpu_inference/layers/jax/moe/utils.py
jrplatin: I actually realized it probably won't be, so removing

tpu_inference/layers/jax/moe/moe.py
lk-chen: but you could assign `self.use_ep` in `__post_init__`, user shouldn't need to set both.

(FYI, `parallel_size > 1` check is pretty common in both upstream and tpu-inference)

jrplatin: Ah fair, that's cleaner!

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
patemotter: I am fine deferring to you on this. I would maybe get someone else's opinion just to be sure.

jrplatin: Gotcha, will wait for @kyuyeunk's opinion too!

lk-chen: `FusedMoEConfig` is vllm concept, which we don't want to import from common/ , I'm fine using boolean argument here.

lk-chen (approved):

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
kyuyeunk: hmm not sure if changing this into `self.moe_backend = select_moe_backend(self.moe.use_ep)` is a good approach. but i do agree with other's point about `FusedMoEConfig` being a vllm concept and we don't want to introduce it to `common/`.

how about this:

1. when moving `tpu_inference/layers/vllm/fused_moe.py` to `tpu_inference/layers/common/moe.py`, we also createa file something like `tpu_inference/layers/vllm/moe.py` or `tpu_inference/layers/vllm/moe_interface.py`
2. this new file's primary job is to invokes moe function in  `tpu_inference/layers/common/moe.py`
3. additionally, this new file in vllm directiory will also have a function called `fused_moe_config_to_backend` or something like that - which takes in `FusedMoEConfig` and selects moe backend based on it.
4. we can also have a similar logic on flax/jax backend as well - a function that takes in a config class that only exists in flax/jax backend and selects backend based on that.

this approach addresses both concerns about
- maintaining current behavior of `select_moe_backend`
- do not introduce vllm specific concept (like `FusedMoEConfig`) into `/common` folder.

wdyt?

tpu_inference/layers/common/moe.py
```diff
@@ -13,61 +13,83 @@
...
 logger = init_logger(__name__)
 
 
-class FusedMoEBackend(Enum):
+class MoEBackend(Enum):
     FUSED_MOE = "fused_moe"
     GMM_EP = "gmm_ep"
     GMM_TP = "gmm_tp"
+    DENSE_MAT = "dense_mat"  # only used in the JAX path for now
+    MEGABLX_GMM = "megablox_gmm"  # only used in the JAX path for now
```
kyuyeunk, line 51: made similar comment to @bzgoogle as well, but what is the difference between `MEGABLOX_GMM` and `GMM_TP` / `GMM_EP` that also uses megablox kernel?

tpu_inference/layers/common/moe.py
```diff
@@ -13,61 +13,83 @@
...
+else:
+    FusedMoEWeights = None
+    UnfusedMoEWeights = None
+    JaxMoE = None
 
 logger = init_logger(__name__)
 
 
-class FusedMoEBackend(Enum):
+class MoEBackend(Enum):
```
kyuyeunk, line 39: as we are adding more backends, i think we need use more descriptive names. at least on the comment, can you elaborate how each methods differ? i'm hoping comments explain when each backend should be used for which workload. this will allow users can make educated decision on which backend they should use for their use case.

tpu_inference/layers/common/moe.py
```diff
@@ -13,61 +13,83 @@
...
 
+    @classmethod
+    def fused_moe_backends(cls):
+        """Returns those backends that use fused weights"""
+        return {cls.FUSED_MOE, cls.GMM_EP, cls.GMM_TP}
 
-def select_moe_backend(moe: FusedMoEConfig):
+
+def select_moe_backend(use_ep: bool):
     if envs.USE_MOE_EP_KERNEL:
```
kyuyeunk: probably not the scope of this pr, but maybe it's better to switch this environment variable to string based?

so things like `MOE_BACKEND=gmm_ep`, `MOE_BACKEND=gmm_tp`, `MOE_BACKEND=fused_moe`, etc.

i can imagine a list of `USE_....` list being very long over time.

tpu_inference/layers/common/moe.py
kyuyeunk, line 39: in the comment, can we clearly mark what is ready to be used for production use case? or ideally, remove any backends that offers no real benefit over any other backends. if we do want to keep some, we should clearly mark why it is there & when it should be used. for instance, i would assume some backends offers zero performance benefit but exists purely for numeric debugging purpose.

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -124,7 +125,7 @@ def __post_init__(self, rngs: nnx.Rngs):
 
 # --- Main Class for MoE ---
 @dataclass(kw_only=True)
-class MoE(nnx.Module):
+class JaxMoE(JaxModule):
```
kyuyeunk, line 128: nit: i haven't really got around to it, but ithere was a consensus to change `*/jax` to  `/flax` and change any existing classes in that file to use `Flax` prefix because both torchax & flax path uses jax, and only difference is whether model implementation was using pytorch or flax.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -42,6 +42,21 @@ class FusedMoEWeights:
     w2_bias: jax.Array | Tensor | None
 
 
+@jax.tree_util.register_dataclass
+@dataclass
```
kyuyeunk, line 46: probably not the scope of this pr, but what is the use case of `UnfusedMoEWeights`? meaning, on which use case is it better to use fused moe weight vs. unfused moe weights?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -181,7 +196,7 @@ def process_moe_weights(
         w2_bias = jnp.expand_dims(w2_bias, 1)
 
     match moe_backend:
```
kyuyeunk, line 198: should there be a switch case added for newly added backends as well?

tpu_inference/layers/common/moe.py
```diff
@@ -13,61 +13,83 @@
...
     gating_output: jax.Array,
-    weights: "FusedMoEWeights",
-    moe_backend: FusedMoEBackend,
+    weights: Union[FusedMoEWeights, UnfusedMoEWeights],
+    moe_backend: MoEBackend,
     mesh: Mesh,
     extra_backend_kwargs: dict,
 ) -> jax.Array:
-    assert isinstance(layer, FusedMoE)
+    assert isinstance(layer, (FusedMoE, JaxMoE))
```
kyuyeunk: related to my comment in https://github.com/vllm-project/tpu-inference/pull/1610#discussion_r2762010420, i think we can move assertion to each framework's interface and keep this code base framework agnostic / does not import on any class from vllm / flax.

tpu_inference/layers/common/moe.py
jrplatin: Of course, very good point thanks!

tpu_inference/layers/common/moe.py
jrplatin: Yeah let's address as a fast follow!

tpu_inference/layers/common/moe.py
jrplatin: Definitely!

tpu_inference/layers/jax/moe/moe.py
jrplatin: Yep that's a good point, will change!

lk-chen: My 2 cents here: I used `Jax-` prefix because it's for jax **ecosystem**, including flax/qwix etc.

kyuyeunk: hmm not sure. because i think torchax could be also considered as a part of jax **ecosystem**? so at the end of the day, they are both part of jax ecosystem, and what differentiate them is how the model code was implemented - whether it is leveraging vllm's implementation, or whether it is our own flax implementation. wdyt?

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: I am going to wire that up in the next PR but let me add the `case`s and raise `NotImplementedErrors` for now

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: @bzgoogle  found worse performance when using Fused weights / the vLLM backend as compared to the new `MEGABLOX` backend, which uses unfused weights.  See [here](https://buganizer.corp.google.com/issues/463682303#comment7) for more context

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
jrplatin: Ah ok sure, that makes sense -- will implement

tpu_inference/layers/jax/moe/moe.py
lk-chen: I always thought torchax-path doesn't work with MaxText or Qwix etc.
If that's myth, I'm happy to note it and come back later. But since we already have models (Qwen series) written with these layers, it doesn't make too much sense doing renaming right now (we were targeting due date 2/8 but it's severely delayed)

jrplatin: I agree that Flax is more meaningful since it's fundamentally tied to the that backend's implementation (i.e. `JaxMoE` is really a wrapper around a fundamentally `Flax` approach to MoE), but I also see Linkun's point that it's generic to the ecosystem (but also agree with Kyuyeun that TorchAX _is_ part of that ecosystem too...)

Why don't we punt the discussion for another PR and we can wholesale address it there?

tpu_inference/layers/common/process_weights/moe_weights.py
kyuyeunk: but the gmm op itself is faster with fused? when combing bot up & down
- 75 us for unfused
- 70 us for fused. 

instead, what we see is that all-reduce that comes after much larger for fused moe - which is addressed by these prs
- https://github.com/vllm-project/tpu-inference/pull/1562
- https://github.com/vllm-project/tpu-inference/pull/1516

so the conclusion here is is wrong: the performance diff isn't coming from weight fusion it's actually the token routing logic that comes after it was unoptimized. in fact, when you just look at gmm layers, actually faster with fused moe vs unfused.

tpu_inference/layers/jax/moe/moe.py
kyuyeunk: > But since we already have models (Qwen series) written with these layers, it doesn't make too much sense doing renaming right now

Yeah, i agree. that's why i posted as a "nit". let's have a separate discussion when things settle down.

tpu_inference/layers/common/moe.py
jrplatin: @bzgoogle  can comment further, but the biggest things are:
* `MEGABLOX_GMM` uses unfused weights (gate/uproj are separate)
* `MEGABLOX_GMM` showed better [perf](https://buganizer.corp.google.com/issues/463682303#comment7) on DeepSeek
* I took a cursory look and it looks like `MEGABLOX_GMM` is much reliant on A2A (versus AR) and tokens in `MEGABLOX_GMM` are moved to the device holding the expert whereas `GMM_TP` is broadcasting them/keeping the tokens on the same device

kyuyeunk: > MEGABLOX_GMM is much reliant on A2A (versus AR)

if `GMM_EP` was used(instead of `GMM_TP`), they should be the same.

in general, i'm always trying to remove any code path that doesn't serve any real use or try to merge feature into existing codes to minimize code path divergence.

anyway, it's not the scope of this pr, so let's discuss later.

jrplatin: Totally agree -- I filed a bug here to track (b/481522056)

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Discussed above, will track as a separate issue here: b/481522056

tpu_inference/layers/common/moe.py
jrplatin: Done!

bvrockwell:
I agree with flax vs torch model framing. Both paths use Jax.

It's otherwise confusing for the user.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
jrplatin: Believe I have addressed this!

patemotter (approved):

kyuyeunk (approved):

Issue #1623, opened 2026-02-04
[Feature] Use vLLM-similar mechanism to load FP8 quantized model


lk-chen:
### 🚀 The feature, motivation and pitch

RFC: #1485 

This ticket is to track the overall process of enabling FP8 pre-quantized model loading

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

lk-chen:
close with https://github.com/vllm-project/tpu-inference/pull/1835 merged

PR #1611, opened 2026-02-03
[FP8][Jax] FP8 JAX linear support

patemotter:
# Description

This PR enables support for FP8 linear layers in JAX, bringing the JAX implementation to feature parity with our existing TorchAX-path FP8 support. This change introduces the necessary infrastructure to load, configure, and execute FP8 linear layers with blockwise scaling.

   * Key Changes:
       * `JaxQuantLinearConfig` to parse einsum dimensions and compute weight sharding compatible with FP8 kernels.
       * `Fp8LinearMethod` 
           * Blockwise scale initialization
           * On-the-fly requantization during loading if REQUANTIZE_BLOCK_SIZE is set.
           * Reshaping/sharding of weights for the 2D kernel execution.

# Tests

Added unit tests for scale loading, linear layer configuration, and numerical correctness of the FP8 path. Added simple qwen2-based e2e test that uses FP8. All tests currently pass.

PR #1629, opened 2026-02-05
[JAX][DeepSeek/MoE] Create `UnquantizedFusedMoEBackend` (sans weight loading)

jrplatin:
# Description

In this PR, I implement the core logic for the `UnquantizedMoEBackend` while keeping reverse compatibility (for now) for the Megablocks + Qwix quantized path.   Specifically, I do the following:

* Move all of the MoE forward pass logic to be contained in `UnquantizedFusedMoEBackend.`. The `FP8UnquantizedFusedMoEBackend` should be landing soon to provide full quantization support.
* Wire up the `DENSE_MAT` and `MEGABLOCKS_GMM` paths in `moe_apply`
* Refactor sparse/dense MoE to be self-contained
* Move backend-specific MoE weight logic to `UnquantizedFusedMoEMethod` from the `JaxMoE` class
* Add `process_weights_after_loading` as a (potentially inheritable) method on `QuantizeMethodBase`
* Implement `apply_jax`, `create_jax_weights`, and `process_weights_after_loading` on `UnquantizedFusedMoEBackend`
* Add a `_process_weights_after_loading` function to `model_loader.py` to call `_process_weights_after_loading`

# Tests

Ensured offline inference outputs matched between `main` for all combinations of the following:
* MHA/MLA
* GMM_TP and MEGABLOCKS_GMM
* Quantization (only for MEGABLOCKS_GMM)

# Checklist

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
     PHASED_PROFILING_DIR: str = ""
     PYTHON_TRACER_LEVEL: int = 1
     USE_MOE_EP_KERNEL: bool = False
-    USE_MEGABLOCKS: bool = False
+    USE_UNFUSED_MEGABLOCKS: bool = False
     USE_DENSE_MOE: bool = False
     NUM_SLICES: int = 1
     RAY_USAGE_STATS_ENABLED: str = "0"
@@ -150,11 +150,14 @@ environment_variables: dict[str, Callable[[], Any]] = {
     "USE_MOE_EP_KERNEL":
     env_bool("USE_MOE_EP_KERNEL", default=False),
     # Enable megablocks for JAX sparse matmul for MoE (Mixture of Experts)
-    "USE_MEGABLOCKS":
-    env_bool("USE_MEGABLOCKS", default=False),
-    # Number of TPU slices for multi-slice mesh
+    # using Unfused weights
+    "USE_UNFUSED_MEGABLOCKS":
+    env_bool("USE_UNFUSED_MEGABLOCKS", default=False),
+    # Enable the dense backend for Jax MoE (Mixture of Experts)
+    # NOTE: this is a naive implementation and should not be used in production
     "USE_DENSE_MOE":
     env_bool("USE_DENSE_MOE", default=False),
+    # Number of TPU slices for multi-slice mesh
     "NUM_SLICES":
     lambda: int(os.getenv("NUM_SLICES") or "1"),
     # Enable/disable Ray usage statistics collection
--- a/tpu_inference/layers/common/moe.py
+++ b/tpu_inference/layers/common/moe.py
@@ -127,6 +127,9 @@ def moe_apply(
                 # NOTE: circular import avoidance
                 from tpu_inference.layers.jax.moe.dense_moe import \
                     dense_moe_func
+                assert isinstance(
+                    gating_output,
+                    tuple), "Expected the gating output to be a tuple"
                 assert len(
                     gating_output
                 ) == 2, "Expected the gating output to be have 2 entries: weights and indices"
--- a/tpu_inference/layers/jax/moe/moe.py
+++ b/tpu_inference/layers/jax/moe/moe.py
@@ -191,25 +191,16 @@ class JaxMoE(JaxModule):
         D = self.hidden_size
         F = self.intermediate_size_moe
 
-        # TODO (jacobplatin): we need to clean up the shardings, but the issue is that
-        # the fused backends expect a different sharding.  Ideally, we'd apply this
-        # in `process_weights_after_loading` in the QuantMethods but I was running
-        # into issues just setting `layer.kernel_gating_EDF.sharding = layer.efd_sharding`
-        # as it didn't seem to have any effect.
-        self.kernel_gating_EDF = create_param(
-            rngs,
-            shape=(E, D, F),
-            dtype=self.dtype,
-            sharding=self.edf_sharding if self.moe_backend
-            not in MoEBackend.fused_moe_backends() else self.efd_sharding,
-            random_init=self.random_init)
-        self.kernel_up_proj_EDF = create_param(
-            rngs,
-            shape=(E, D, F),
-            dtype=self.dtype,
-            sharding=self.edf_sharding if self.moe_backend
-            not in MoEBackend.fused_moe_backends() else self.efd_sharding,
-            random_init=self.random_init)
+        self.kernel_gating_EDF = create_param(rngs,
+                                              shape=(E, D, F),
+                                              dtype=self.dtype,
+                                              sharding=self.edf_sharding,
+                                              random_init=self.random_init)
+        self.kernel_up_proj_EDF = create_param(rngs,
+                                               shape=(E, D, F),
+                                               dtype=self.dtype,
+                                               sharding=self.edf_sharding,
+                                               random_init=self.random_init)
         self.kernel_down_proj_EFD = create_param(
             rngs,
             shape=(E, F, D),
--- a/tpu_inference/layers/jax/quantization/__init__.py
+++ b/tpu_inference/layers/jax/quantization/__init__.py
@@ -27,7 +27,6 @@ def get_tpu_quantization_config(vllm_config: VllmConfig):
         UnquantizedConfig
 
     model_config = copy.deepcopy(vllm_config.model_config)
-    # TODO: add fp8
     method_to_config: dict[str | None, type] = {
         None: UnquantizedConfig,
         # We want to support FP8 model (e.g. DeepSeek) in jax path.
@@ -36,9 +35,6 @@ def get_tpu_quantization_config(vllm_config: VllmConfig):
         # TODO(#1623): replace with actual FP8 config when delegation is ready.
         FP8: lambda _: None,
     }
-    # TODO (jacobplatin): temporary workaround for now before FP8 is fully ready for DeepSeek
-    if vllm_config.model_config.hf_config.model_type == "deepseek_v3":
-        method_to_config[FP8] = UnquantizedConfig
 
     if model_config.quantization not in method_to_config:
         raise NotImplementedError(
@@ -70,7 +66,6 @@ class QuantizeMethodBase(ABC):
         Expects create_weights to have been called before on the layer."""
         raise NotImplementedError
 
-    @abstractmethod
     def process_weights_after_loading(self, layer: JaxModule, *args,
                                       **kwargs) -> None:
         """
@@ -80,4 +75,4 @@ class QuantizeMethodBase(ABC):
             layer: The layer to process
         """
 
-        raise NotImplementedError
+        pass
--- a/tpu_inference/layers/jax/quantization/unquantized.py
+++ b/tpu_inference/layers/jax/quantization/unquantized.py
@@ -63,7 +63,8 @@ class UnquantizedFusedMoEMethod(QuantizeMethodBase):
         super().__init__(*args, **kwargs)
         self.extra_backend_kwargs = {}
 
-    def process_weights_after_loading(self, layer: JaxMoE) -> None:
+    def process_weights_after_loading(self, layer: JaxMoE, *args,
+                                      **kwargs) -> None:
         """
         Process weights after loading.
 
@@ -78,7 +79,7 @@ class UnquantizedFusedMoEMethod(QuantizeMethodBase):
             w_gate = layer.kernel_gating_EDF.value
             w_up = layer.kernel_up_proj_EDF.value
 
-            # stack to create a 4d matrix
+            # stack to create a 4d array
             w13_val = jnp.stack([w_gate, w_up], axis=1)
 
             layer.kernel_gating_upproj_E2DF = nnx.Param(w13_val,
@@ -101,9 +102,6 @@ class UnquantizedFusedMoEMethod(QuantizeMethodBase):
                 "bd2c": 256,
             }
 
-            # TODO (jacobplatin): why is this sharding flipped?
-            layer.kernel_down_proj_EFD.sharding = layer.edf_sharding
-
         elif layer.moe_backend in [MoEBackend.GMM_EP, MoEBackend.GMM_TP]:
             w_gate = layer.kernel_gating_EDF.value
             w_up = layer.kernel_up_proj_EDF.value
@@ -117,9 +115,6 @@ class UnquantizedFusedMoEMethod(QuantizeMethodBase):
 
             del layer.kernel_gating_EDF
             del layer.kernel_up_proj_EDF
-
-            # TODO (jacobplatin): why is this sharding flipped?
-            layer.kernel_down_proj_EFD.sharding = layer.edf_sharding
 
     def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
         assert isinstance(layer, JaxMoE)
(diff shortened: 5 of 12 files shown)


tpu_inference/layers/jax/moe/moe.py
```diff
@@ -321,96 +191,40 @@ def __post_init__(self, rngs: nnx.Rngs):
...
-                dtype=self.dtype,
-                sharding=self.efd_sharding,
-                random_init=self.random_init)
-
-        # Default MoE has no bias vectors
-        self.w1_bias, self.w2_bias = (None, None)
-
-        # TODO: Add quantization scale params for VLLM MoE kernel
-        self.w1_scale, self.w2_scale = (None, None)
+        # TODO (jacobplatin): we need to clean up the shardings, but the issue is that
```
jrplatin: Want to address this at some point soon

tpu_inference/layers/jax/quantization/__init__.py
```diff
@@ -35,6 +36,10 @@ def get_tpu_quantization_config(vllm_config: VllmConfig):
         # TODO(#1623): replace with actual FP8 config when delegation is ready.
         FP8: lambda: None,
     }
+    # TODO (jacobplatin): temporary workaround for now before FP8 is fully ready for DeepSeek
```
jrplatin: Very temporary hack

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -49,18 +54,120 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
...
+                "bf": 512,
+                "bd1": 512,
+                "bd2": 512,
+                "btc": 64,
+                "bfc": 256,
+                "bd1c": 256,
+                "bd2c": 256,
+            }
+
+            # TODO (jacobplatin): why is this sharding flipped?
```
jrplatin: @bzgoogle  I'd like to understand why this is the case

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -1533,6 +1471,13 @@ def __init__(self,
         self.vllm_config = vllm_config
         self.rng = nnx.Rngs(rng)
 
+        # TODO (jacobplatin): we will resolve this issue in a forthcoming PR that will refactor weight loading
```
jrplatin: Note

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -49,18 +54,119 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
...
+
+        router_logits = None
+        # Fused weight backends
+        if layer.moe_backend in MoEBackend.fused_moe_backends():
+            # of shape TE, only 1D in this case
+            router_logits = layer.router(x_TD)
+
+            w13_weight = layer.kernel_gating_upproj_E2DF.value if layer.moe_backend == MoEBackend.FUSED_MOE else layer.kernel_gating_upproj_EDF.value
+            w2_weight = layer.kernel_down_proj_EFD.value
+            # TODO (jacobplatin/bzgoogle): we should support bias
```
jrplatin, line 133: Note

tpu_inference/envs.py
```diff
@@ -152,6 +153,8 @@ def _get_bool_env() -> bool:
     "USE_MEGABLOCKS":
     env_bool("USE_MEGABLOCKS", default=False),
     # Number of TPU slices for multi-slice mesh
+    "USE_DENSE_MOE":
+    env_bool("USE_DENSE_MOE", default=False),
```
lk-chen, line 159: Could you add comment like other envs did?

tpu_inference/models/jax/deepseek_v3.py
lk-chen: you can use #1623

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -49,18 +54,119 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
 
 class UnquantizedFusedMoEMethod(QuantizeMethodBase):
     """
-    TODO (jacobplatin): implement in forthcoming PR.
+    Unquantized method for JAXMoE layer.
+
+    TODO (jacobplatin): support weight loading -- currently, model-dependent.
```
lk-chen, line 59: same here, you can use #1623

tpu_inference/models/common/model_loader.py
```diff
@@ -96,6 +96,21 @@ def _get_nnx_model(
     rng: jax.Array,
     mesh: Mesh,
 ) -> nnx.Module:
+    vllm_config.quant_config = get_tpu_quantization_config(vllm_config)
+
+    def _process_weights_after_loading(model: nnx.Module) -> None:
```
jrplatin: Looks like https://github.com/vllm-project/tpu-inference/pull/1611/changes#diff-68dc98fbd69cf60fc9de428c6fb117a12bfd17c70bc13985885f7e2bdec0ee36R573

Is also implementing this, so will need to coordinate with @patemotter

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -49,18 +54,119 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
 
 class UnquantizedFusedMoEMethod(QuantizeMethodBase):
```
lk-chen, line 55: Comparing 2 pattern:
1. There's one UnquantizedFusedMoEMethod, switching to different branches according to `moe_backend`, call `moe_apply` which is still branching according to `moe_backend`
2. Implement `UnquantizedGmmMoEMethod` `UnquantizedMegabloxGmmMoEMethod` separately and call `gmm_moe_apply` `megablox_gmm_moe_apply` each

Is there a reason behind why you choose 1?

tpu_inference/layers/jax/quantization/__init__.py
lk-chen: is it intended to check in?

tpu_inference/layers/common/moe.py
```diff
@@ -124,11 +124,32 @@ def moe_apply(
...
                 )
             case MoEBackend.DENSE_MAT:
-                # TODO(jacobplatin): will implement in forthcoming PR
-                raise NotImplementedError
+                # NOTE: circular import avoidance
+                from tpu_inference.layers.jax.moe.dense_moe import \
+                    dense_moe_func
+                assert len(
+                    gating_output
+                ) == 2, "Expected the gating output to be have 2 entries: weights and indices"
```
patemotter, line 135: Based on the type hints this can be a jax.Array as well as a tuple. Will this work for both cases? I believe len(jax.Array) will return the size of the first dim.

tpu_inference/layers/jax/quantization/unquantized.py
jrplatin: 1 is more in-line with what [TorchAX](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/moe.py#L59) is currently doing (one Method with the multiple backends called via `moe_apply`) so figured I'd keep things consistent

tpu_inference/layers/jax/quantization/__init__.py
jrplatin: Yes, I think so -- otherwise, DeepSeek can't be loaded with the refactored code, but as I noted, this is very temporary and should be removed as soon as FP8 support is done (hopefully by EOW)

kyuyeunk: can you share why deepseek can't be loaded with the refactored code?

tpu_inference/layers/jax/moe/dense_moe.py
```diff
@@ -11,54 +11,139 @@
...
+
+
+def dense_moe_func(weights: UnfusedMoEWeights, x_TD: jax.Array,
+                   gating_output: Tuple[jax.Array, jax.Array],
+                   cast_dtype: jnp.dtype, num_local_experts: int,
+                   apply_expert_weight_before_computation: bool,
+                   activation_ffw_td, hidden_act) -> jax.Array:
+    """
+    Forward pass of the dense MoE layer.  This is a naive implementation
+    and thus should not be used in production.
```
patemotter, line 105: Nit: do we print an explicit logging warning about this anywhere?

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -49,18 +54,119 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
...
+        """
+        if layer.moe_backend == MoEBackend.FUSED_MOE:
+            if layer.edf_sharding:
+                e2df_sharding = (layer.edf_sharding[0], None,
+                                 layer.edf_sharding[1], layer.edf_sharding[2])
+            # fuse the weights into w13: [Gate, Up]
+            w_gate = layer.kernel_gating_EDF.value
+            w_up = layer.kernel_up_proj_EDF.value
+
+            # stack to create a 4d matrix
```
patemotter: The most pedantic nit: 4d tensor or 4d array would be more correct.

tpu_inference/layers/jax/quantization/__init__.py
jrplatin: Sorry, what I said above wasn't super clear -- when I say "refactored code" what I mean is the "refactored code in its current state".  This is a very temporary hack that we'll remove once we actually have the FP8 support, but it just enables us to run the quantized model between now and then (when the refactor is actually done)

tpu_inference/layers/jax/moe/dense_moe.py
jrplatin: I can add a `warning_once`!

tpu_inference/envs.py
jrplatin: Added!

tpu_inference/layers/common/moe.py
jrplatin: Ah what a catch, thanks Pate -- updated!

tpu_inference/envs.py
```diff
@@ -21,7 +21,8 @@
     PHASED_PROFILING_DIR: str = ""
     PYTHON_TRACER_LEVEL: int = 1
     USE_MOE_EP_KERNEL: bool = False
-    USE_MEGABLOCKS: bool = False
+    USE_UNFUSED_MEGABLOCKS: bool = False
```
jrplatin, line 24: I renamed this BTW because I realized `USE_MEGABLOCKS` is too vague -- I _think_ this is a clear name but open to feedback obviously!

tpu_inference/layers/jax/quantization/unquantized.py
lk-chen: Got it. For future reference vLLM use if-branch's to create weight [here](https://github.com/vllm-project/vllm/blob/92e7562a994038c904fea859d90462c7e84a3246/vllm/model_executor/layers/quantization/mxfp4.py#L250-L274)

tpu_inference/layers/jax/quantization/__init__.py
lk-chen: IIUC, DS-R1 needs FP8 einsum to fully work, which is missing right now.

Image

Could you explain (maybe in the TODO comment) if we load DS-R1-FP8 checkpoint, how does it not hit exception at einsum? If there's exception, we should remove this hack (you can have local patch for dev.)

jrplatin: Ah so this branch was based on the current `main` yesterday where the `UnquantizedLinearMethod` wasn't hooked up yet -- if that has been hooked up, then I think you're right in that we'll get issues, but let me double check.  We might be ok because DeepSeek isn't using `JaxLinear` at all anyways

jrplatin: I moved this logic to the DeepSeek modeling file for now, thanks @kyuyeunk !

tpu_inference/models/common/model_loader.py
jrplatin: I took over `_process_modules_after_loading` so it should be easy to merge, regardless of who goes first

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -568,6 +568,22 @@ def load_hf_weights(
     nnx.update(model, params)
 
 
+def process_modules_after_loading(module, mesh):
```
lk-chen: nit: similar to out discussion regarding the config, if this is temporary for ds-v3, you may want to move it to ds_v3.py to control the exclosure of this function

lk-chen: nit: similar to out discussion regarding the config, if this is temporary for ds-v3, you may want to move it to ds_v3.py to control the exclosure of this function

tpu_inference/models/common/model_loader.py
```diff
@@ -96,6 +96,7 @@ def _get_nnx_model(
     rng: jax.Array,
     mesh: Mesh,
 ) -> nnx.Module:
+    vllm_config.quant_config = get_tpu_quantization_config(vllm_config)
```
lk-chen: Line 245 already has this

tpu_inference/layers/jax/quantization/__init__.py
```diff
@@ -27,6 +27,7 @@ def get_tpu_quantization_config(vllm_config: VllmConfig):
         UnquantizedConfig
 
     model_config = copy.deepcopy(vllm_config.model_config)
+    # TODO: add fp8
```
lk-chen: Line 36 should be sufficient

tpu_inference/models/jax/utils/weight_utils.py
jrplatin: I think Pate has this logic too no? So if it's needed for both of us, I figured just add it in the shared weight utils

jrplatin: Ah didn't see [this](https://github.com/vllm-project/tpu-inference/pull/1611/changes#r2766542566) -- so this will be removed?

tpu_inference/models/common/model_loader.py
jrplatin: Ah good catch -- rebased and missed thanks!

tpu_inference/models/jax/utils/weight_utils.py
jrplatin: Moved!

tpu_inference/layers/jax/moe/dense_moe.py
```diff
@@ -11,54 +11,139 @@
...
+                              axis=1)
+    # Some models use the routing scores to weight the data instead of
+    # weighting the expert outputs.
+    if apply_expert_weight_before_computation:
+        with jax.named_scope("pre_computing_weight"):
+            return dense_moe_fwd_preapply_router_weights(
+                weights=weights,
+                x_TD=x_TD,
+                cast_dtype=cast_dtype,
+                activation_ffw_td=activation_ffw_td,
```
patrickji2014, line 140: there is inconsistency in calling dense_moe_fwd_preapply_router_weights, activation_ffw_td is not an arg of the function, it should be activation_ffw_ted. either one need a fix

jrplatin: https://github.com/vllm-project/tpu-inference/pull/1685

Fix out here

PR #1633, opened 2026-02-05
[FP8][Jax] blockwise-quantized fp8 linear method

lk-chen:
# Description

As part of #1623 , this PR adds `Fp8BlockwiseLinearMethod` that:
* loads weight from pre-quantized HF checkpoint (e.g. Qwen3-0.6B-FP8)
* implements forward path

This PR is independent from #1638 (which is for **tensor-wise** quantized model)

# Tests

1.
```
export CURRENT_VLLM_MODEL=Qwen/Qwen3-0.6B-FP8
MODEL_IMPL_TYPE=flax_nnx vllm serve $CURRENT_VLLM_MODEL
```
it errors at RMSNorm, which is expected and means qkv proj are working (both weight loading and forward path)

2. #1640 can reproduce above results in 1

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/quantization/fp8.py
+++ b/tpu_inference/layers/jax/quantization/fp8.py
@@ -42,40 +42,12 @@ class Fp8BlockwiseLinearMethod(QuantizeMethodBase, common_fp8.Fp8LinearMethod):
 
         kernel_shape = layer.kernel_shape
         if len(kernel_shape) > 2:
-            # HF model stores weight in 2-D shape. E.g. for "TD,DNH->TNH", weight shape in HF is (NH, D)
-            x_axis, w_axis = layer.einsum_str.split("->")[0].split(",")
-            contracting_axis = set(x_axis) & set(w_axis)
-            in_features = math.prod([
-                layer.kernel_shape[i] for i, c in enumerate(w_axis)
-                if c in contracting_axis
-            ])
-            out_features = math.prod([
-                layer.kernel_shape[i] for i, c in enumerate(w_axis)
-                if c not in contracting_axis
-            ])
-
-            # E.g. if weight shape is (NH, D), sharding is ('x', None, 'y'), we need to fuse sharding to ('x', 'y')
-            spec = layer.weight.sharding
-            if isinstance(layer.weight.sharding, jax.NamedSharding):
-                spec = layer.weight.sharding.spec
-            elif isinstance(layer.weight.sharding,
-                            jax.sharding.SingleDeviceSharding):
-                spec = ()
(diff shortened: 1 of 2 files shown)


tpu_inference/layers/jax/quantization/configs.py
```diff
@@ -28,3 +28,18 @@ def __init__(self, hf_quant_config: dict):
     def get_quant_method(self, layer: JaxModule,
                          prefix: str) -> Optional[QuantizeMethodBase]:
         raise NotImplementedError
+
+    @classmethod
+    def get_from_keys(cls, config: dict, keys: list, *args):
```
patemotter, line 33: This function could be made a bit clearer in its naming. I would expect a `get_from_keys` function that accepts a list of `keys` to return a list of found or default config values of equal length to `keys`. Returning the first value found also suggests a priority in the ordering of `keys`, which is also not obvious. 

I think the code itself is fine, it's just a clarity concern.

tpu_inference/layers/jax/quantization/fp8.py
```diff
@@ -0,0 +1,218 @@
...
+class Fp8BlockwiseLinearMethod(QuantizeMethodBase, common_fp8.Fp8LinearMethod):
+    """Block-wise Fp8 method for JAX Linear layer."""
+
+    def __init__(self, quant_config: "Fp8Config", layer: JaxEinsum,
+                 linear_config: QuantLinearConfig):
+        common_fp8.Fp8LinearMethod.__init__(self, linear_config)
+        self.quant_config = quant_config
+
+        kernel_shape = layer.kernel_shape
+        if len(kernel_shape) > 2:
```
patemotter, line 44: I think the section of code could be put into a helper function. I wrote something similar that has not yet been added into the codebase. It should largely be a drop-in replacement. 

https://github.com/vllm-project/tpu-inference/blob/f78a68f00d0727f6a60d08c5b0f57a85c4ddb7ea/tpu_inference/layers/jax/quantization/configs.py#L71

tpu_inference/layers/jax/quantization/configs.py
lk-chen: It's replicating upstream https://github.com/vllm-project/vllm/blob/aaa2efbe9861200056a8f2dc31d2482fec89b880/vllm/model_executor/layers/quantization/base_config.py#L118

tpu_inference/layers/jax/quantization/fp8.py
lk-chen: Done, moved to `tpu_inference/layers/common/quantization/configs.py`

PR #1634, opened 2026-02-05
[Jax] refactor to prepare for FP8 linear method support

lk-chen:
# Description

Prepare for #1633 

# Tests

CI

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/quantization/__init__.py
+++ b/tpu_inference/layers/jax/quantization/__init__.py
@@ -45,7 +45,7 @@ def get_tpu_quantization_config(vllm_config: VllmConfig):
     # There are some cases to be supported in the future:
     # 1) Some vision model keep quantization config under text_config
     # 2) overriding through `--hf_overrides`
-    return quant_config.from_config(hg_quant_config)
+    return quant_config(hg_quant_config)
 
 
 class QuantizeMethodBase(ABC):


tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -797,6 +797,16 @@ def __init__(self, model, **kwargs):
 
         super().__init__(model, **kwargs)
 
+    def _load_module(self, base_prefix: str, module: JaxModule,
+                     weights: Iterable) -> Iterable:
+        yield from super()._load_module(base_prefix, module, weights)
+        # Post-process module after loading weights. Unlike vLLM post-process
```
patemotter, line 803: Nit: Similar to a comment on my PR, do we mean vLLM proper or the vLLM path in tpu-inference?

tests/layers/jax/quantization/test_unquantized.py
```diff
@@ -43,7 +43,7 @@ def test_linear_forward_correctness(self, in_features, out_features,
                                use_bias=use_bias)
         y_from_layer = jax_linear(x)
 
-        method = UnquantizedConfig().get_quant_method(jax_linear, prefix='')
+        method = UnquantizedConfig({}).get_quant_method(jax_linear, prefix='')
```
patemotter, line 46: Nit: Why is the `{}` necessary?

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: Here I mean upstream vllm. And vllm path in tpu-inference, is adapting to upstream, which is also non-ideal.

tests/layers/jax/quantization/test_unquantized.py
lk-chen: The base class (in this PR) is updated to consume hf_quant_config dict, `UnquantizedConfig` inherits from that which needs an empty dict here.

PR #1643, opened 2026-02-06
Refactor and clean up attn_dp shardings

wenxindongwork:
# Description

Refactor some attention dp shardings. Impact to performance should be positive and minimal.  

Previously inputs are all-gathered before the embedding layer, now this all-gather is eliminated. 

Previously output hidden state is all-gathered during logits computation, now this all-gather happens in the step_fn. 

# Tests

Verified with e2e profiling. 

# Checklist

### DIFF
--- a/output.txt
+++ b/output.txt
@@ -1,1178 +1,0 @@
-============================= test session starts ==============================
-platform linux -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0 -- /home/wenxindong_google_com/miniconda3/envs/wenxin_test/bin/python3
-cachedir: .pytest_cache
-rootdir: /home/wenxindong_google_com/tpu-inference
-configfile: pyproject.toml
-plugins: mock-3.15.0, jaxtyping-0.3.2, anyio-4.10.0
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/common/sharding.py
```diff
@@ -41,7 +41,7 @@ class ShardingAxisNameBase:
     MOE_TENSOR = ('attn_dp', 'model')
     EXPERT = ('attn_dp', 'expert', 'model')
     EXPERT_DATA = ('data', 'attn_dp', 'expert', 'model')
-    VOCAB = ('model', 'attn_dp', 'expert')
+    VOCAB = ('attn_dp', 'model', 'expert')
```
bzgoogle, line 44: may i know the reason of switching this 2 axises for VOCAB?

bzgoogle (approved):

tpu_inference/layers/common/sharding.py
wenxindongwork: it's just for cosmetics :) everywhere else had attn_dp before model

PR #1644, opened 2026-02-06
[AWQ][MoE] Support AWQ MoE

catswe:
# Description
Enabled AWQ. Ran on v6e-8.

# Changes:
To preserve the memory benefit of AWQ, I dequantized on the fly and added `zero_point` parameters. The changes from adding `zero_point` is similar in logic as `scale` parameters

`tpu_inference/layers/vllm/quantization/awq.py`
- Referenced `vllm/model_executor/layers/quantization/awq_marlin.py`
- Referenced `tpu_inference/layers/vllm/quantization/fp8.py`
- Referenced `tpu_inference/layers/vllm/quantization/mxfp4.py`
- Defined `get_fused_moe_quant_config` since otherwise attribute access error
- Implemented `create_weights` directly and uses `extra_weight_attrs` because upstream has no base AWQ MoE method to inherit from, unlike the other TPU MoE methods

# Tests

Due to changes from https://github.com/vllm-project/tpu-inference/pull/1659, doing `pip install -e .` may be needed before running the following command
```
MODEL_IMPL_TYPE=vllm vllm serve QuantTrio/MiniMax-M2-AWQ --gpu-memory-utilization=0.95 --tensor_parallel_size=8 --enable-expert-parallel --dtype bfloat16 --download-dir /dev/shm
```

Screenshot 2026-02-27 at 6 26 31 PM

# Checklist

### DIFF
--- a/tests/layers/vllm/test_awq.py
+++ b/tests/layers/vllm/test_awq.py
@@ -538,4 +538,8 @@ def test_fused_moe(num_devices, num_tokens, intermediate_size, hidden_size,
             vllm_fused_moe)
         actual = vllm_fused_moe(jax_a, score)
 
-        torch.testing.assert_close(expected, actual, check_device=False)
+        torch.testing.assert_close(expected,
+                                   actual,
+                                   check_device=False,
+                                   atol=1e-1,
+                                   rtol=1e-1)


tpu_inference/kernels/megablox/gmm.py
```diff
@@ -37,6 +37,7 @@ def _validate_args(
     rhs: jnp.ndarray,
     group_sizes: jnp.ndarray,
     rhs_scale: jnp.ndarray | None = None,
+    rhs_zero_point: jnp.ndarray | None = None,
```
kyuyeunk: might be dumb question, but what would be the fundamental difference between `zero_point` and `bias`? meaning, instead of creating new argument `zero_point`, is it possible to just expand `bias` to also cover `zero_point` use case?

tpu_inference/kernels/megablox/gmm.py
```diff
@@ -494,11 +499,22 @@ def _wrapper(x):
...
             for b_i in range(num_quant_blocks_per_tk):
+                loaded_lhs_block = loaded_lhs[..., b_i *
+                                              quant_block_size:(b_i + 1) *
+                                              quant_block_size]
+                loaded_rhs_block = loaded_rhs[b_i *
+                                              quant_block_size:(b_i + 1) *
+                                              quant_block_size, ...]
+
+                if rhs_zero_point is not None:
+                    # TODO(catswe): direct cast from uint4 (awq)
```
kyuyeunk: note that we don't have to strictly follow the same checkpoint numeric as the original weight value. as long as we can verify it's quality neutral, we can just requantize it from uint4 to int4. can you see if that kind of approach is possible & make the code run faster?

catswe: That's actually a good idea. I think it should work better. I'll report back with MMLU benchmarks

catswe: It improved from 44.9 peak tokens/s generation to 45.2 peak tokens/s generation for single request expert parallel on v6e-8. The difference is small but it's there since there's no chain-casting anymore.

kyuyeunk: that's awesome! thanks! have you verified that there is no quality drop as well?

catswe: i don't see any quality difference and there theoretically shouldn't be any quality difference

kyuyeunk: agree. yeah just wanted to make sure in case there's some bug in our code that causes some issues. thanks for checking it!

tpu_inference/kernels/megablox/gmm.py
catswe: so the matmul for awq is `y = x @ (scale * (uint4_weight - uint4_zero_point))`. if you expand the matmul, `y = x @ (scale * uint4_weight) - x @ (scale * uint4_zero_point)`, the zero_point term becomes input-dependent (depends on x). i don't see a way to expand `bias` to also cover `zero_point` use case and it's also per-block/per-group unlike bias. `zero_point` logic is more similar to `scale` than `bias` in that it's per-block/per-group

tpu_inference/kernels/megablox/gmm.py
```diff
@@ -491,11 +496,21 @@ def _wrapper(x):
...
+                loaded_lhs_block = loaded_lhs[..., b_i *
+                                              quant_block_size:(b_i + 1) *
+                                              quant_block_size]
+                loaded_rhs_block = loaded_rhs[b_i *
+                                              quant_block_size:(b_i + 1) *
+                                              quant_block_size, ...]
+
+                if rhs_zero_point is not None:
+                    # NOTE(catswe): cast to bf16 for awq
+                    loaded_rhs_block = loaded_rhs_block.astype(
```
kyuyeunk: couple of thoughts
1. upcasting from 4bit to 16bit is not free
2. compute wise, w16a16 performs worse than w4a16

can you do some analysis on how much performance we are loosing by this upcasting and if that is a lot, brainstorm a way to perform this logic without upcasting?

catswe: I changed to doing the zero-point subtraction after `jnp.dot`, which allows for removing `rhs_zero_point` upcasting. This improved generation throughput from 45.2 tokens/s to 48.6 tokens/s for single request expert parallel on v6e-8!

catswe: Interestingly enough, I still have to keep the logic for converting rhs to `int4` instead of keeping it as `uint4` since there appears to be a Pallas bug that leads to incorrect result for jnp.dot with `uint4` rhs and `bf16` lhs. I filed a bug report here https://github.com/jax-ml/jax/issues/35492

kyuyeunk: can you try manually upcasting uint4 into fp8? for context, technically, there isn't a native hw support for w4a16, but it does have support for w8a16 where w8 as fp8. so w4a16 works by upcasting w4 into fp8 and than perform w8a16 - which is still faster than w16a16. it's possible that this feature is enabled for int4 but not enabled for uint4.

note that you can see the list of hardware supported matmuls here: https://github.com/jax-ml/jax/blob/500bbc2066db9a4103bbff45e0c8fa0477f27d94/jax/_src/pallas/mosaic/tpu_info.py#L196-L235

kyuyeunk: Also, for kernels changes, suggest making a separate pr for it. So something like 'gmm support do asymmetric quantized weight'

kyuyeunk: My suggestion is to change this pr to not make any changes to gmm kernel (so maybe dequantize th weights) and than create a follow up pr that adds asymmetric quantized weight support + reenable awq asymmetric weights.

catswe: > My suggestion is to change this pr to not make any changes to gmm kernel (so maybe dequantize th weights) and than create a follow up pr that adds asymmetric quantized weight support + reenable awq asymmetric weights.

will do! sounds perfectly reasonable to me!

catswe: > can you try manually upcasting uint4 into fp8? for context, technically, there isn't a native hw support for w4a16, but it does have support for w8a16 where w8 as fp8. so w4a16 works by upcasting w4 into fp8 and than perform w8a16 - which is still faster than w16a16. it's possible that this feature is enabled for int4 but not enabled for uint4.
> 
> note that you can see the list of hardware supported matmuls here: https://github.com/jax-ml/jax/blob/500bbc2066db9a4103bbff45e0c8fa0477f27d94/jax/_src/pallas/mosaic/tpu_info.py#L196-L235

casting from `uint4` to any of `jnp.float8_e5m2`, `jnp.float8_e4m3fn` `jnp.float8_e4m3b11fnuz`, or any other fp8 dtype isn't supported in pallas

tpu_inference/layers/vllm/quantization/awq.py
```diff
@@ -219,3 +229,271 @@ def _apply_split(self,
             outs.append(out)
         out = jnp.concatenate(outs, axis=-1)
         return torch_view(out)
+
+
+class VllmAWQMoEMethod(FusedMoEMethodBase):
```
kyuyeunk, line 234: so i just took a look at vllm's implementation and they seem to have a generic [`MoeWNA16Method` ](https://github.com/vllm-project/vllm/blob/a73af584fe6d4c1c2781d537c35e3cc85f58480b/vllm/model_executor/layers/quantization/moe_wna16.py#L214) that is shared across different quantization methods. is this kind of design principal that we can adopt for tpu-inference as well?

for instance, we already have two moe methods for fp8 w8a8. it would be better if we can remove redundant code by having shared moe method implementation.

catswe: are you suggesting to create a shared MoE method scoped to WNA16 quantization (and have AWQ inherit from it) so it can be reused when adding GPTQ support later? or a broader refactor across all MoE methods (FP8, compressed tensors FP8, MXFP4, AWQ) into a shared base class? if it's the first option, that seems pretty sound to me

kyuyeunk: yup yup. the first option. but come to think of it, i think it might be outside of the scope of this pr and it might be better to just create a follow up pr for it? i'll approve it but let me know if you prefer landing the change into this pr or you prefer creating a follow up pr for it.

kyuyeunk (approved):

catswe:
i prefer creating a follow up pr!

kyuyeunk:
> i prefer creating a follow up pr!

thanks! will merge it!

PR #1660, opened 2026-02-09
2D-TP & FP8 KV Support for MLA + replicated attention weights

gpolovets1:
# Description

Updated shardings to support MLA for 2D-TP (including adding new `attn_dp_expert` axis)
Also included support for using replicated attention weights and custom compilation batch sizes.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -569,7 +569,9 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         if self.execute_model_state is not None:
             raise RuntimeError("State error: sample_tokens() must be called "
                                "after execute_model() returns None.")
-        output = self._execute_model(scheduler_output, intermediate_tensors)
+        with jax.set_mesh(self.mesh):
+            output = self._execute_model(scheduler_output,
+                                         intermediate_tensors)
         return output
 
     def sample_tokens(
@@ -1010,7 +1012,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
             logprobs_lists = None
 
         if self.speculative_config:
-            with self.maybe_forbid_compile:
+            with self.maybe_forbid_compile, jax.set_mesh(self.mesh):
                 self.speculative_decoding_manager.propose_draft_token_ids(
                     valid_sampled_token_ids,
                     aux_hidden_states,
@@ -1040,9 +1042,9 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         ret = jax.shard_map(
             select_local_fn,
             mesh=self.mesh,
-            in_specs=(PartitionSpec(ShardingAxisName.MLP_DATA),
-                      PartitionSpec(ShardingAxisName.MLP_DATA)),
-            out_specs=PartitionSpec(ShardingAxisName.MLP_DATA))(
+            in_specs=(PartitionSpec(ShardingAxisName.ATTN_DATA),
+                      PartitionSpec(ShardingAxisName.ATTN_DATA)),
+            out_specs=PartitionSpec(ShardingAxisName.ATTN_DATA))(
                 array, indices_to_select)
 
         return ret
@@ -1219,8 +1221,6 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         dp_size = self.dp_size
         data_parallel_attn_sharding = NamedSharding(
             self.mesh, PartitionSpec(ShardingAxisName.ATTN_DATA))
-        data_parallel_sharding = NamedSharding(
-            self.mesh, PartitionSpec(ShardingAxisName.MLP_DATA))
 
         (req_ids_dp, req_indices_dp, num_scheduled_tokens_per_dp_rank,
          scheduled_tokens_per_dp_rank, num_req_per_dp_rank,
@@ -1408,16 +1408,11 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         logits_indices_cpu = logits_indices
         seq_lens_cpu = seq_lens
 
-        (input_ids, logits_indices) = device_array(
-            self.mesh,
-            (input_ids, logits_indices),
-            sharding=data_parallel_sharding,
-        )
-
-        (positions, query_start_loc, seq_lens,
+        (input_ids, positions, query_start_loc, seq_lens, logits_indices,
          request_distribution) = device_array(
              self.mesh,
-             (positions, query_start_loc, seq_lens, request_distribution),
+             (input_ids, positions, query_start_loc, seq_lens, logits_indices,
+              request_distribution),
              sharding=data_parallel_attn_sharding,
          )
 
--- a/tests/layers/jax/quantization/test_fp8.py
+++ b/tests/layers/jax/quantization/test_fp8.py
@@ -32,6 +32,8 @@ from tpu_inference.layers.jax.moe.moe import JaxMoE
 # yapf: disable
 from tpu_inference.layers.jax.quantization.fp8 import (
     Fp8Config, Fp8TensorwiseLinearMethod)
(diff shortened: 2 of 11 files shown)


tpu_inference/layers/common/sharding.py
```diff
@@ -155,20 +165,25 @@ def from_vllm_config(cls,
...
             # Use attention DP instead to reduce per-device num_kv_heads and
             # eliminate this waste.
-            num_kv_heads_per_device_in_kv_cache = (num_kv_heads * 2) / packing
+            
+            num_kv_heads_per_device_in_kv_cache = max(1, (num_kv_heads * 2) / packing)
             attn_dp = max(
                 int(tensor_parallelism // num_kv_heads_per_device_in_kv_cache),
                 1)
             tensor_parallelism = tensor_parallelism // attn_dp
+            attn_dp_expert = expert_parallelism
```
wenxindongwork, line 179: I'm a bit confused on why we need a new axis `attn_dp_expert` here, could you help me understand why we can't simply use `expert`? would one of `expert` and  `attn_dp_expert` always be one?

bzgoogle: +1. also curious the different usage between expert and attn_dp_expert ?

gpolovets1: It is intended to be the analogy to 'model' & 'attn_dp': https://github.com/vllm-project/tpu-inference/pull/1660/changes#diff-de37130aad6bcb0c3aef2904bc90bc0bf03971954abc55ef5371a0eb2903a68eR177

Since we are sharding the model weights now on both 'model' and 'expert', we want to shard the kv cache on both of these axes and create model * expert DP shards. Without this, total_dp_size would be just 'attn_dp' which does not include the shards from the 'expert' axis.

But if there is a cleaner way to do this than applying the analogy of model/attn_dp, ideas are welcome!

wenxindongwork, line 179: In the JAX path, we don't enable expert parallelism via the `--enable_expert_parallelism` flag, but instead do it via `additional_config`, is this understanding correct? (also is there a reason why we haven't unified these approaches?) 

Wondering how this new change works on the torchax path. In the torchax path, we don't use the `expert` axis. We use the `model` axis instead. In that case, there will be no way to use this 2D tensor sharding in torchax right?

gpolovets1: That's a good question, @jrplatin has been working on unifying the code for DeepSeek. Do we still plan on using the 'expert' axis from additional_config to create the weight shardings?
In general, I have not checked out of the box support on vllm path for DeepSeek (we have been running it through the Jax path) but can add that as a future todo (which may require a few more changes like what you are bringing up with enable_expert_parallelism).

wenxindongwork: this approach makes sense to me! I can't think of any cleaner approach

gpolovets1: Aligned async that follow-up PRs will address other sharding configurations that are also used by torchax.

tpu_inference/layers/common/quantization/__init__.py
```diff
@@ -271,11 +271,13 @@ def static_per_tensor_quantize_tensor(
...
-    value: jax.Array,
-    k_scale: float,
-    v_scale: float,
+    value: jax.Array | None = None,
+    k_scale: float | None = None,
+    v_scale: float | None = None,
 ) -> Tuple[jax.Array, jax.Array]:
     """Static quantize key and value tensors."""
     key = static_per_tensor_quantize_tensor(dtype, key, k_scale)
+    if value is None:
```
jrplatin, line 280: Can you add a comment explaining this case more -- not immediately obvious to me why this is needed, thanks

tpu_inference/layers/common/sharding.py
```diff
@@ -128,14 +134,22 @@ def from_vllm_config(cls,
         sharding_strategy = vllm_config.additional_config.get(
             "sharding", {}).get("sharding_strategy", {})
         parallel_config = vllm_config.parallel_config
-        tensor_parallelism = parallel_config.tensor_parallel_size
+        # Currently tensor_parallelism is also used for other things like determining number of Ray workers.
+        pc_tensor_parallelism = parallel_config.tensor_parallel_size
```
jrplatin, line 138: Nit: maybe write out the full abbreviation? Not obvious what PC/SS are

tpu_inference/runner/tpu_runner.py
```diff
@@ -403,12 +405,22 @@ def _init_inputs(self) -> None:
...
+            "compilation_sizes", [])
         # [16, 32, 64, 128, 256, 512, 1024, 2048]
+        cache_dtype = self.cache_config.cache_dtype
+        if cache_dtype == "auto":
+            cache_dtype = self.dtype
+        kv_cache_dtype = to_jax_dtype(cache_dtype)
+        kv_packing = common_utils.get_dtype_packing(kv_cache_dtype)
         self.num_tokens_paddings = runner_utils.get_token_paddings(
-            min_token_size=max(16, self.dp_size),
+            min_token_size=max(16, self.dp_size * kv_packing),
```
jrplatin, line 418: Why is this needed?

tpu_inference/runner/tpu_runner.py
```diff
@@ -403,12 +405,22 @@ def _init_inputs(self) -> None:
...
             max_token_size=scheduler_config.max_num_batched_tokens *
             self.dp_size,
             padding_gap=vllm_envs.VLLM_TPU_BUCKET_PADDING_GAP)
+        self.num_tokens_paddings = sorted(self.num_tokens_paddings +
```
jrplatin, line 422: Why is this needed?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -1568,11 +1585,24 @@ def _create_deepseek_attention(
                 query_tnh_spec = P(ShardingAxisName.MLP_TENSOR, None, None)
                 keyvalue_skh_spec = P(ShardingAxisName.MLP_TENSOR, None)
                 attn_o_tnh_spec = P(ShardingAxisName.MLP_TENSOR, None, None)
-
+                anh_sharding = (None, ShardingAxisName.MLP_TENSOR, None)
             else:
                 query_tnh_spec = P(None, ShardingAxisName.MLP_TENSOR, None)
                 keyvalue_skh_spec = P(None, ShardingAxisName.MLP_TENSOR, None)
                 attn_o_tnh_spec = P(None, ShardingAxisName.MLP_TENSOR, None)
+            rd_sharding = (ShardingAxisName.MLP_TENSOR, None)
```
jrplatin, line 1613: Are these variables names in the glossary?

tpu_inference/layers/common/quantization/__init__.py
gpolovets1: Hi @jrplatin it's because for MLA and some other models, there is a latent that is shared between both K & V. In these cases, you only need to quantize a single array instead of two arrays.

tpu_inference/runner/tpu_runner.py
gpolovets1: IIRC, for multi host (maybe in conjunction with MLA), there was a bug because the dimension was further divided by kv_packing and made the overall dimension size < 1 (i.e. a fraction). This was necessary to fix it.

tpu_inference/runner/tpu_runner.py
gpolovets1: This is to allow custom compilation shapes that are not necessarily powers of 2. E.g. if your batch size becomes large, then unless it is a power of 2, there could be a very large amount of padding. And your maximum concurrency does not always line up with a power of 2.

tpu_inference/layers/common/sharding.py
gpolovets1: Could you provide an example of what you're thinking? I was initially trying to compromise between length and clarity.

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: What do you mean by glossary? They were already previously defined in line 1612 below. I just allow for the option to keep these weights replicated.

PR #1672, opened 2026-02-09
[CI Bug Fix] Removes the sharding_config from vllm_config before deep copy

wenxindongwork:
# Description

`tpu-inference` dynamically adds a `sharding_config` attribute to VllmConfig, but when the config is deep-copied and passed to vLLM's model loading code, vLLM's `replace()` function in `config/utils.py` tries to validate all attributes against the dataclass fields. Since `sharding_config` is not a field in vLLM's `VllmConfig` dataclass, it raises `ValueError: Field 'sharding_config' not found in VllmConfig`.

The fix removes the `sharding_config` attribute from the deep-copied config before passing it to vLLM for model loading.

This change is safe because `vllm_config_for_load` is a temporary deep copy used only during `vllm_get_model()` and LoRA loading
The original `self.vllm_config` is preserved and still has `sharding_config` for all subsequent operations.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -133,6 +133,12 @@ class VllmModelWrapper:
             self.vllm_config.device_config.slice = slice_config
         assert self.vllm_config.model_config.dtype in TORCH_DTYPE_TO_JAX, "The model_config.dtype must be a PyTorch dtype."
         vllm_config_for_load.device_config.device = "cpu"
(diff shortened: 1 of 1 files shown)


kyuyeunk:
doesn't have to be done in this pr, but i wasn't really a fan of assigning a variable that doesn't exist in `VllmConfig`, would it be possible to create a child class `TpuVllmConfig` that expands the `VllmConfig` to include our custom member variables like `sharding_config`?

PR #1673, opened 2026-02-09
[Deepseek][Jax][MLA] Migrate DS-v3 MLA 2D to use JaxEinsum

patemotter:
# Description

* Migrates the 2D projections inside Deepseek-V3 MLA to use `JaxEinsum`. 
* Removed leading underscore from `_sharded_initializer` and `_scale_initializer` in `tpu_inference/layers/jax/base.py` to indicate they are usable elsewhere. 

Note: the 3d weights (about 10% of the total MLA parameters) remain unchanged in this commit, it will require some additional work if we decide to quantize them. 

# Tests

* Unit tests
```
NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py \
--model=deepseek-ai/DeepSeek-R1 --max-model-len=5120 --tensor-parallel-size 8 \
--max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching  \
--gpu-memory-utilization 0.95 \
--additional_config='{"skip_quantization": true, "sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' \
--hf_overrides  '{"num_hidden_layers": 5}'
```

# Checklist


tests/models/jax/test_deepseek_v3.py
```diff
@@ -175,7 +175,7 @@ def loader(self, mock_config):
     @pytest.mark.parametrize("loaded_key, expected_mapped", [
         ("model.embed_tokens.weight", "embedder.input_embedding_table_VD"),
         ("model.layers.0.self_attn.q_a_proj.weight",
-         "layers.0.self_attn.kernel_q_down_proj_DA"),
+         "layers.0.self_attn.q_down_proj.weight"),
```
jrplatin, line 178: Same comment here and below, but do we not want to follow the "suffix-with-shape" convention any more? Seemed fairly useful to me

jrplatin:
Can you please run this command (on the Megablocks unfused path) as well please?

```
USE_UNFUSED_MEGABLOCKS=1 NEW_MODEL_DESIGN=True TPU_BACKEND_TYPE=jax python examples/offline_inference.py --model=jrplatin/DeepSeek-R1-1D-Subchannel-256-FP8 --max-model-len=5120 --tensor-parallel-size 8 --max-num-seqs 16 --max-num-batched-tokens 128 --no-enable-prefix-caching  --gpu-memory-utilization 0.95 --additional_config='{"sharding": {"sharding_strategy": {"attention_data_parallelism": 8, "enable_dp_attention": true}}}' --hf_overrides  '{"num_hidden_layers": 10}'
```

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -423,6 +446,7 @@ class DeepseekV3MLA(DeepseekV3BaseAttention):
     anh_sharding: Sharding = ()
 
     def setup_specific_layers(self, rngs: nnx.Rngs) -> None:
+        # TODO: refactor 3D projections to use JaxEinsum
```
jrplatin, line 449: TODO here?

jrplatin:
I'd say it'd also probably be good to run a full MMLU run with perf, maybe [this](https://b.corp.google.com/issues/477390236#comment13)?

tests/models/jax/test_deepseek_v3.py
lk-chen: Because the variable name needs to exact match param names from HF checkpoint. Alternative is to add `load_weights` function for attention module and:
```
def load_weights(self, weights_iter):
    for name, weight in weights_iter:
         if "q_down_proj" in name:
            # match to kernel_q_down_proj_DA
            load_nnx_param_from_reshaped_torch(self.kernel_q_down_proj_DA, weight)
         elif "..." in name:
             ...
```
Which seems unnecessary to me, because "einsum_str" contains the shape information already.

Note that for moe I'm keeping "suffix-with-shape", because MoE anyway needs `load_weights` function for weight loading.

lk-chen (commented):
LGTM, let's run some benchmark/evaluation to confirm

tests/models/jax/test_deepseek_v3.py
jrplatin: Ah fair, makes sense

tpu_inference/models/jax/deepseek_v3.py
jrplatin: Discussed offline -- will follow-up on in another PR

patemotter:
> I'd say it'd also probably be good to run a full MMLU run with perf, maybe [this](https://b.corp.google.com/issues/477390236#comment13)?

Combined with the other ongoing PRs we get an MMLU accuracy of .926. Full details are [here](https://buganizer.corp.google.com/issues/476484771#comment5).

PR #1675, opened 2026-02-10
Add e2e correctness test for sequence parallelism

weiyu0824:
# Description
This PR add a sequence parallelism e2e test. Only correctness test is added, since performance gain is not verified yet. 

# Tests
Run pytest -s -v ./tests/e2e/test_sequence_parallelism.py

# Checklist


tests/e2e/test_sequence_parallelism.py
```diff
@@ -0,0 +1,297 @@
...
+import time
+from dataclasses import dataclass, field
+from typing import Optional
+
+import pytest
+from vllm import LLM, SamplingParams
+from vllm.config import CompilationConfig
+
+
+# MODEL_IMPL_TYPE=vllm vllm serve Qwen/Qwen2.5-32B  --seed 42
```
vanbasten23: do you need it?

tests/e2e/test_sequence_parallelism.py
```diff
@@ -0,0 +1,297 @@
...
+
+    start_time = time.time()
+    outputs = llm.generate(test_prompts, sampling_params)
+    elapsed_time = time.time() - start_time
+
+    del llm
+    time.sleep(10)  # Wait for TPUs to be released
+    return outputs, elapsed_time
+
+import numpy as np
```
vanbasten23: can you put it at the top?

tests/e2e/test_sequence_parallelism.py
```diff
@@ -0,0 +1,297 @@
...
+
+def test_sp_correctness(sampling_params: SamplingParams):
+    _test_sequence_parallelism(
+        sampling_params=sampling_params,
+        check_correctness=True,
+        check_performance=False,
+    )
+
+
+# def test_sp_performance(sampling_params: SamplingParams):
```
vanbasten23: take out the perf test for now? Similar comment for `def _check_performance(`

tests/e2e/test_sequence_parallelism.py
```diff
@@ -0,0 +1,297 @@
...
+        sp_text = sp_result.outputs[0].text.strip()
+
+        # Calculate word overlap for fuzzy matching
+        baseline_words = set(baseline_text.split())
+        sp_words = set(sp_text.split())
+        overlap = baseline_words & sp_words
+        match_percent = len(overlap) / len(
+            baseline_words) if baseline_words else 0
+
+        if match_percent >= 0.7:
```
vanbasten23, line 132: I'm not sure if it will be flaky. In EP https://github.com/vllm-project/tpu-inference/blob/771bbe0e376747700aa485bbfc9dcc037fec0956/.buildkite/parallelism/EP.yml#L28, we observed flakiness after James merged the code so we have to keep reducing the threshold.. So we use lm_eval instead. You can monitor the test after the pr is merge.

tests/e2e/test_sequence_parallelism.py
weiyu0824: This is the command that I follow, I can remove this right now.

tests/e2e/test_sequence_parallelism.py
weiyu0824: ok

vanbasten23 (approved):

PR #1712, opened 2026-02-13
Revert "[Kernel]Add reduce-scatter-matmul kernel"

QiliangCui:
Reverts vllm-project/tpu-inference#1526

1. The change has "from experimental.reduce_scatter_matmul import ..." which doesn't seem right.
2. it breaks tests.


kyuyeunk:
okay @rupengliu-meta,  @QiliangCui, let's hold off on merging it since it seems like a simple fix.

rupengliu-meta:
https://github.com/vllm-project/tpu-inference/pull/1713 maybe? @QiliangCui

QiliangCui:
the 1713 fix doesn't seem to work. still fail..

PR #1715, opened 2026-02-13
[JAX][MoE] Add a few clarifying comments on MoE weights

jrplatin:
# Description

Please see title.

# Tests

N/A

# Checklist

### DIFF
--- a/tpu_inference/layers/common/moe.py
+++ b/tpu_inference/layers/common/moe.py
@@ -37,18 +37,30 @@ logger = init_logger(__name__)
 
 
 class MoEBackend(Enum):
-    # This is using the Fused MoE kernel found in tpu-inference/tpu_inference/kernels/fused_moe/v1/kernel.py
+    # FUSED_MOE is using the Fused MoE kernel found in tpu-inference/tpu_inference/kernels/fused_moe/v1/kernel.py
     # and is production ready.
+    # NOTE: for FusedMOE on the JAX/Flax path, we expect the MoE quant method (e.g. UnquantizedFusedMoEMethod) to
+    # create a new `kernel_gating_upproj_E2DF` param that replaces `kernel_gating_EDF` and `kernel_up_proj_EDF`
+    # and thus runs the forward pass on the fused weight.  `kernel_down_proj_EFD` is unchanged.
     FUSED_MOE = "fused_moe"
-    # This is using the GMM kernel found in tpu-inference/tpu_inference/layers/common/fused_moe_gmm.py
+
+    # GMM_EP is using the GMM kernel found in tpu-inference/tpu_inference/layers/common/fused_moe_gmm.py
     # and is `expert_sharded_gmm` the GMM calls.  Production ready.
+    # NOTE: for GMM_EP JAX/Flax path, we expect the MoE quant method (e.g. UnquantizedFusedMoEMethod) to
+    # create a new `kernel_gating_upproj_EDF` param that replaces `kernel_gating_EDF` and `kernel_up_proj_EDF`
+    # and thus runs the forward pass on the fused weight.  `kernel_down_proj_EFD` is unchanged.
     GMM_EP = "gmm_ep"
+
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/common/moe.py
```diff
@@ -39,14 +39,22 @@
 class MoEBackend(Enum):
     # This is using the Fused MoE kernel found in tpu-inference/tpu_inference/kernels/fused_moe/v1/kernel.py
     # and is production ready.
+    # NOTE: for the JAX/Flax path, we expect the MoE quant method (e.g. UnquantizedFusedMoEMethod) to
```
lk-chen: It's sufficient to say "With this backend, gate and up projection weights are expected to fuse into one during weight loading, and forward path will apply on the fused weight"

Also shape is not right, should be [num_experts, 2 * hidden_size, intermediate size]. Let's remove that kind of detail here.

jrplatin: Whoops, good catch on the typo....pushed fix

tpu_inference/layers/common/moe.py
```diff
@@ -39,14 +39,22 @@
...
+    # createa a new `kernel_gating_upproj_E2DF` param that replaces `kernel_gating_EDF` and `kernel_up_proj_EDF`
```
patemotter: Nit: "create a"

tpu_inference/layers/common/moe.py
```diff
@@ -39,14 +39,22 @@
...
+    # and thus runs the forward pass on the fused weight.  `kernel_down_proj_EFD` is unchanged.
     FUSED_MOE = "fused_moe"
     # This is using the GMM kernel found in tpu-inference/tpu_inference/layers/common/fused_moe_gmm.py
     # and is `expert_sharded_gmm` the GMM calls.  Production ready.
+    # NOTE: for the JAX/Flax path, we expect the MoE quant method (e.g. UnquantizedFusedMoEMethod) to
+    # createa a new `kernel_gating_upproj_EDF` param that replaces `kernel_gating_EDF` and `kernel_up_proj_EDF`
```
patemotter: Nit: "create a"

tpu_inference/layers/common/moe.py
```diff
@@ -39,14 +39,22 @@
...
+    # and thus runs the forward pass on the fused weight.  `kernel_down_proj_EFD` is unchanged.
     GMM_EP = "gmm_ep"
     # Same as GMM_EP but uses `tensor_sharded_gmm_*` for the GMM calls and is production ready
     GMM_TP = "gmm_tp"
     # Uses a simple dense matmul for the MoE backend and is intended for testing
     # Only used in the JAX path for now
+    # NOTE: for the JAX/Flax path, there are no change for weights for the unfused backends (DENSE_MAT and MEGABLOX_GMM).
+    # That is, `kernel_gating_EDF`, `kernel_up_proj_EDF`, and `kernel_down_proj_EFD` are unchanged.
```
patemotter, line 60: Can we add some vertical whitespace to this whole collection of comments so it's easier to tell what comment corresponds to what variable?

tpu_inference/layers/common/moe.py
```diff
@@ -39,14 +39,22 @@
 class MoEBackend(Enum):
     # This is using the Fused MoE kernel found in tpu-inference/tpu_inference/kernels/fused_moe/v1/kernel.py
```
patemotter: I know you didn't touch this line, but if we are changing the comments, can we also change all the instances of "this" to the proper name of whatever it is referencing?

tpu_inference/layers/common/moe.py
jrplatin: Will do!

tpu_inference/layers/common/moe.py
jrplatin: Yep!

PR #1716, opened 2026-02-13
[Fixing] Fix CI/CD to make Integration pipeline moving forward

QiliangCui:
# Description

breaks fixed

1) https://github.com/vllm-project/vllm/pull/33843 --> MoE action changed from string to enum.
    a) in the torchax path, use `layer.activation.value` to replace `layer.activation`. 
    b) avoid touching kernel or common so that both jax and torchax path works
2) https://github.com/vllm-project/vllm/pull/34125 --> added new pause_state property in the scheduler interface


tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -761,6 +762,16 @@ def reset_encoder_cache(self) -> None:
             self._get_result_from_queue(rank,
                                         SchedulerCommand.RESET_ENCODER_CACHE)
 
+    @property
+    def pause_state(self) -> PauseState:
+        return self._pause_state
+
+    def set_pause_state(self, pause_state: PauseState) -> None:
+        del pause_state
```
QiliangCui, line 770: @wenxindongwork , Hi Wenxin, https://github.com/vllm-project/vllm/pull/34125 added "pause_state", can I fix it like this as a quick unblocking test? 

Should we support the pause_state somewhere?

QiliangCui: chatted offline. a follow up fix: https://github.com/vllm-project/tpu-inference/pull/1718

tpu_inference/layers/common/moe.py
```diff
@@ -67,6 +67,8 @@ def moe_apply(
 ) -> jax.Array:
 
     with jax.named_scope(layer._get_name()):
+        activation = layer.activation if isinstance(
+            layer.activation, str) else layer.activation.value
```
kyuyeunk, line 71: Are there vLLM code path that still uses string based activation?

QiliangCui: no. it uses the enum.
however, I don't want our lower level code, like kernel.py and fused_moe_gmm.py, which are shared by jax and vllm, reference the enum defined by vllm. 

In this way, our lower level function, like gmm kernel and fused_moe, doesn't depend on vllm but just the string. at the vllm --> lower layer, we transfer the vllm'enum to string. 

This is the cleanest way of fixing it. ( I tried other way too but not clean and need to have jax path depend on vllm code that is not ideal).

kyuyeunk (approved):

tpu_inference/layers/common/moe.py
kyuyeunk: @lk-chen maybe it might be a good idea to change `JaxMoE`'s `activation` variable into vLLM's enum as well?

so this one: https://github.com/vllm-project/tpu-inference/blob/6b4aaa1e25136b6a0af952fcf16c97bdea923455/tpu_inference/layers/jax/moe/moe.py#L218 will be changed into this enum: https://github.com/neuralmagic/vllm/blob/d5fe3f702c2f4392515cdfde6fd0442271c74dda/vllm/model_executor/layers/fused_moe/activation.py#L11-L27

wdyt?

PR #1727, opened 2026-02-14
[FIX] remove strict power-of-two constraint for total_dp

sierraisland:
# Description

Currently, `total_dp` is required to be a power of two due to a strict assertion in the token padding logic. This PR relaxes this constraint by ensuring that the internal `min_token_size` is aligned to the next power of two, regardless of whether the user-provided `dp_size` is a power of two. This maintains the hardware's requirements for memory alignment while supporting arbitrary DP configurations.

**Changes:**
- Used `next_power_of_2` utility in `tpu_inference/runner/utils.py`.
- Updated `TPUModelRunner` initialization to calculate `min_token_size` using this new alignment logic.
- Added comprehensive unit tests to ensure different `dp_size` and `kv_packing` combinations align correctly.

FIX: b/483122466

# Tests

**Unit Tests:** Added `test_min_token_size_alignment` in `tests/runner/test_utils.py`. All tests passed.
   - Command: `pytest tpu-inference/tests/runner/test_utils.py`

# Checklist


tpu_inference/runner/utils.py
```diff
@@ -44,6 +44,10 @@ class InferencePhase(Enum):
     AMBIGUOUS = 3
 
 
+def next_power_of_two(n: int) -> int:
```
kyuyeunk: nit: palls provides the same exact utility function: https://github.com/jax-ml/jax/blob/fff793a03eabfab02baf8d68f63983cb256f5914/jax/_src/pallas/utils.py#L70

kyuyeunk: And also vLLM upstream: https://github.com/vllm-project/vllm/blob/main/vllm/utils/math_utils.py#L15

tpu_inference/runner/tpu_runner.py
```diff
@@ -323,11 +324,18 @@ def _create_single_slice_mesh(self) -> jax.Array:
...
+                mesh_shape,
+                self.devices,
+                allow_split_physical_axes=True,
+            )
+        except (AssertionError, ValueError, RuntimeError) as e:
+            logger.warning(
+                "Physical mesh creation failed (shape=%s, devices=%d). "
+                "Falling back to logical reshape. Error: %s", mesh_shape,
+                len(self.devices), e)
+            return np.array(self.devices).reshape(mesh_shape)
```
wenxindongwork, line 341: @sierraisland should a mesh be returned here?

tpu_inference/runner/tpu_runner.py
wenxindongwork: could you also add the fallback here?

tpu_inference/runner/tpu_runner.py
wenxindongwork: and curious if you know why `mesh_lib.Mesh` works but `jax.make_mesh` doesn't?

sierraisland: Actually, I'd prefer to keep the return type as `np.ndarray` here to remain strictly consistent with JAX's native `mesh_utils.create_device_mesh` and `create_hybrid_device_mesh` APIs.

sierraisland: `jax.make_mesh` performs a strict physical topology check to ensure a perfect rectangular grid on the hardware. On an 8-core host, DP=6 is physically 'irregular,' so it fails that check. `mesh_lib.Mesh` is a low-level API that only cares about logical layout, allowing us to bypass the hardware constraints.

wenxindongwork: awesome, thanks for the explanation, please free feel to add this in the comments

PR #1732, opened 2026-02-17
[Misc] Log error when exception from TPU worker

lk-chen:
# Description

When exception happens in worker, executor might just [capture it](https://github.com/vllm-project/vllm/blob/7967e854da4868872b4b13d2b7f039061fee50fe/vllm/v1/executor/uniproc_executor.py#L94) and, if another error occurs before this exception is thrown, it's hard to know the real exception.

This PR adds error logging

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/utils.py
+++ b/tpu_inference/runner/utils.py
@@ -150,6 +150,13 @@ class ForbidCompile:
         # Store the original function
         self._original_func = pxla._cached_lowering_to_hlo
         original_cached_func = self._original_func
+        # Prepare trackback such that when exception triggers,
+        # we can show where the call that triggered the compilation is.
+        tb = None
+        try:
+            raise Exception()
+        except Exception as e:
+            tb = e.__traceback__
 
         # Create a wrapper
         @functools.wraps(original_cached_func)
@@ -167,7 +174,7 @@ class ForbidCompile:
 
             # Check if a cache miss occurred
             if misses_after > misses_before:
-                raise RuntimeError(self.message)
+                raise RuntimeError(self.message).with_traceback(tb)
 
             return result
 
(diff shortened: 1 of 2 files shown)


tpu_inference/runner/utils.py
```diff
@@ -150,6 +150,13 @@ def __enter__(self):
         # Store the original function
         self._original_func = pxla._cached_lowering_to_hlo
         original_cached_func = self._original_func
+        # Prepare trackback such that when exception triggers,
+        # we can show where the call that triggered the compilation is.
+        tb = None
```
vanbasten23, line 155: Since this function is run for every fwd, will there be any perf implication if we raise an exception every time? Another idea is to raise and throw within the `if misses_after > misses_before:` statement below.

tpu_inference/worker/tpu_worker.py
```diff
@@ -305,7 +305,16 @@ def execute_model(
         # violates the pure abstract contract of the base class. This is a
         # deliberate, temporary compromise for the same reasons outlined in
         # the `get_kv_cache_spec` method.
+        try:
+            return self._execute_model(scheduler_output)
+        except Exception as e:
+            logger.error(f"Error executing model: {e}")
```
vanbasten23, line 311: Somehow, when the async scheduling is on, the real exception is hidden.  But when the async scheduling is off, I can see the real exception, like the one here.

tpu_inference/runner/utils.py
lk-chen: Good call on perf. implication. I cannot put it under `if misses_after > misses_before` because at that time the traceback is already missing.

I can put this PR on hold for now.

PR #1657, opened 2026-02-08
[Jax][Qwen] Several fixes for `Qwen3-4B-2507-Instruct-FP8` in jax-path

lk-chen:
# Description

`Qwen3-4B-2507-Instruct-FP8` is the most achievable model to test FP8 linear for #1623 (for now, because #1633 and #1640 enabled block-wise FP8 linear method, and we are missing FP8-RmsNorm and/or FP8-MoE). This PR contains several fixes for this model:

1. set `dtype=model.dtype`. It's meaningless to set `param_dtype`, because we should [keep the original type from checkpoint](https://github.com/vllm-project/tpu-inference/blob/91d7109627bc968fff535fa223ce29b7a8d302bb/tpu_inference/models/jax/utils/weight_utils.py#L756-L758). Otherwise we load an FP8 checkpoint onto the device as floating type, that loses the whole point of quantization. This PR sets `dtype` of each layer to follow [upstream definition](https://docs.vllm.ai/en/v0.15.1/configuration/engine_args/#-dtype).

2. considers skipped layers, following upstream code (see code comment for details). Model files are updated to pass `prefix` of each module.

3. use `jax.set_mesh` when necessary.

4. Known issue 
  * #1684 
  * #1661 

# Tests

1. CI

2. benchmark and eval
```
# start server
export CURRENT_VLLM_MODEL=Qwen/Qwen3-4B-Instruct-2507-FP8
MODEL_IMPL_TYPE=flax_nnx vllm serve $CURRENT_VLLM_MODEL \
    --max-model-len=8192 \
    --disable-log-requests \
    --no-enable-prefix-caching

# benchmark
export CURRENT_VLLM_MODEL=Qwen/Qwen3-4B-Instruct-2507-FP8
vllm bench serve \
    --model $CURRENT_VLLM_MODEL \
    --num-prompts 1000 \
    --dataset-name random \
    --random-input-len 1024 \
    --random-output-len 4096

# eval
export CURRENT_VLLM_MODEL=Qwen/Qwen3-4B-Instruct-2507-FP8
lm_eval \
    --model local-completions \
    --model_args base_url=http://localhost:8000/v1/completions,model=$CURRENT_VLLM_MODEL \
    --tasks gsm8k_cot \
    --batch_size auto \
    --apply_chat_template \
    --num_fewshot 8
```
result
```
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  1641.95   
[...]

PR #1733, opened 2026-02-17
[Fix] Fix lora test

lk-chen:
# Description

Fix https://github.com/vllm-project/tpu-inference/pull/1657#issuecomment-3895188714

# Tests

```
VLLM_XLA_CHECK_RECOMPILATION=1 VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax pytest -sx -vv tests/lora/test_lora.py -k single
```

# Checklist

### DIFF
--- a/tpu_inference/runner/utils.py
+++ b/tpu_inference/runner/utils.py
@@ -150,13 +150,6 @@ class ForbidCompile:
         # Store the original function
         self._original_func = pxla._cached_lowering_to_hlo
         original_cached_func = self._original_func
-        # Prepare trackback such that when exception triggers,
-        # we can show where the call that triggered the compilation is.
-        tb = None
-        try:
-            raise Exception()
-        except Exception as e:
-            tb = e.__traceback__
 
         # Create a wrapper
         @functools.wraps(original_cached_func)
@@ -174,7 +167,7 @@ class ForbidCompile:
 
             # Check if a cache miss occurred
             if misses_after > misses_before:
-                raise RuntimeError(self.message).with_traceback(tb)
+                raise RuntimeError(self.message)
 
             return result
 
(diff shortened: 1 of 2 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -85,7 +84,8 @@ def capture_model(self) -> None:
             return
         logger.info("Precompile all the subgraphs with possible input shapes.")
 
-        with self.runner.maybe_setup_dummy_loras(self.runner.lora_config):
+        with self.runner.maybe_setup_dummy_loras(
```
vanbasten23, line 87: I guess with this change, the `maybe_setup_dummy_loras` is run under the `self.runner.mesh` so that it is consistent with the code in `tpu_runner`, right?

vanbasten23:
Thanks @lk-chen for the fix!

vanbasten23:
Could you also run a multi-chip test `USE_V6E8_QUEUE=1 VLLM_LOG_LEVEL=INFO VLLM_XLA_CHECK_RECOMPILATION=1 VLLM_USE_V1=1 MODEL_IMPL_TYPE=vllm TPU_BACKEND_TYPE=jax pytest -s -vv tests/lora/test_lora.py -k single` before merging?

tpu_inference/runner/compilation_manager.py
lk-chen: it doesn't matter? `maybe_setup_dummy_loras` is defined [upstream](https://github.com/vllm-project/vllm/blob/dc5fa77a4eb6680339cb77abe713fb22d7795560/vllm/v1/worker/lora_model_runner_mixin.py#L94) which is regardless of whether inside or outside `set_mesh`

The reason of the issue, I think, is one of `_recompile_` is triggered outside of `execute_model` and should not be included in `set_mesh`.

vanbasten23: Not sure what you meant by "_recompile_ is triggered outside of execute_model and should not be included in set_mesh". But the fix moves the `set_mesh` from `_run_compilation` to the upper level `capture_model` so that the `set_mesh` will additionally affect the `set_active_loras`, consistent with what the `set_mesh` does in `tpu_runner.py`.

lk-chen: Got it, that makes sense

PR #1735, opened 2026-02-17
[Bug fix] Minor fix for TPU profiling

jrplatin:
# Description

Addresses a new profiler arg added to vLLM via https://github.com/vllm-project/vllm/pull/33198

# Tests

Tested locally.

# Checklist

### DIFF
--- a/tpu_inference/worker/tpu_worker.py
+++ b/tpu_inference/worker/tpu_worker.py
@@ -354,7 +354,9 @@ class TPUWorker:
         raise NotImplementedError(
             "LoRA is not supported by the JAX worker yet.")
 
-    def profile(self, is_start: bool = True):
+    def profile(self,
+                is_start: bool = True,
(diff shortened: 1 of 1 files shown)


tpu_inference/worker/tpu_worker.py
```diff
@@ -354,7 +354,9 @@ def add_lora(
         raise NotImplementedError(
             "LoRA is not supported by the JAX worker yet.")
 
-    def profile(self, is_start: bool = True):
+    def profile(self,
+                is_start: bool = True,
+                profile_prefix: str | None = None):
```
patemotter, line 359: Do we want to do anything with this variable or is that a decision for later?

jrplatin: Decision for later -- but I wanted to use the explicit var instead of `kwargs` to be consistent with vLLM

PR #1738, opened 2026-02-17
[Fix] fix sharding in fp8.py

lk-chen:
# Description

Sharding should be `(out, in)` in 2D case, i.e. weight and sharding should be permuted altogether

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/jax/deepseek_v3.py
+++ b/tpu_inference/models/jax/deepseek_v3.py
@@ -165,7 +165,6 @@ class DeepseekV3BaseAttention(JaxModule):
             quant_config=self.quant_config,
             param_dtype=self.dtype,
             kernel_init=nnx.with_partitioning(weight_init, self.q_da_sharding),
-            prefix=self.prefix + ".q_a_proj",
         )
 
         self.q_b_proj = JaxEinsum(
@@ -175,7 +174,7 @@ class DeepseekV3BaseAttention(JaxModule):
             quant_config=self.quant_config,
             param_dtype=self.dtype,
             kernel_init=nnx.with_partitioning(weight_init, self.ap_sharding),
-            prefix=self.prefix + ".q_b_proj")
+        )
 
         self.kv_down_proj = JaxEinsum(
             einsum_str="SD,DA -> SA",
@@ -185,7 +184,7 @@ class DeepseekV3BaseAttention(JaxModule):
             param_dtype=self.dtype,
             kernel_init=nnx.with_partitioning(weight_init,
                                               self.kv_da_sharding),
-            prefix=self.prefix + ".kv_a_proj_with_mqa")
+        )
 
         self.o_proj = JaxEinsum(
             einsum_str="TR,RD->TD",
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/common/quantization/configs.py
```diff
@@ -26,19 +26,24 @@ class QuantLinearConfig:
...
-        # Adapted weight sharding for the 2D (contracting, free) dims.
-        # E.g. for "TD,DNH->TNH", if weight sharding is ('x', None, 'y'),
-        # adapted weight sharding will be ('x', 'y') because the sharding on
-        # N and H axis are fused.
-        weight_sharding: tuple
         # Non-contracting, non-batch axis sizes from the weight.
         # E.g. for "TD,DNH->TNH", out_features = (N, H).
         # Used for reshaping right before return from __call__.
         out_features: tuple[int, ...]
+        # Sharding for out_features, extracted from weight sharding.
```
patemotter, line 34: Nit: Can we make this series of comments easier to read by adding a newline to make it easier to see what comment is referring to what variable?

tpu_inference/layers/jax/quantization/fp8.py
```diff
@@ -251,10 +223,10 @@ def create_weights_jax(self, layer: JaxModule, *weight_args, rngs,
             # Weight stays in FP8 and is used with sharded_quantized_batched_matmul.
             param_dtype = jnp.float8_e4m3
             layer.weight = nnx.Param(
-                kernel_init(rngs.params(), layer.kernel_shape, param_dtype),
+                kernel_init(rngs.params(), self.kernel_shape, param_dtype),
                 weight_loader=partial(load_nnx_param_from_reshaped_torch,
                                       permute_dims=None,
-                                      param_name="linear_fp8_weight"),
+                                      param_name=layer.prefix + ".weight"),
```
patemotter, line 229: Do we have any guarantee that `layer.prefix` is not empty when we use it, either here or in other places changed in this PR? By default the parameter is `""`.

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -1513,11 +1515,13 @@ def __init__(self,
                  vllm_config: VllmConfig,
                  rng: jax.Array,
                  mesh: Mesh,
-                 force_random_weights: bool = False):
+                 force_random_weights: bool = False,
+                 prefix: str = ""):
```
patemotter, line 1525: Do we want to require `prefix` instead of a default empty string?

tpu_inference/layers/jax/quantization/fp8.py
lk-chen: Intentionally there's no such guarantee, since it's just used in logging

tpu_inference/models/jax/deepseek_v3.py
lk-chen: Yes, in a later PR we will have `DeepseekV3ForCausalLM` that creates `DeepSeekV3` with prefix "model"

tpu_inference/layers/common/quantization/configs.py
lk-chen: Done

PR #1691, opened 2026-02-11
Update Kernel Blocks Sizes for GMM and Fused EP MOE

QiliangCui:
# Description
1. Add tuned block to gmm kernel. 
2. Update Fused EP MoE Kernel

# Tests

```
2/8 code 
vllm: a75a5b54c7f76bc2e15d3025d61e63cc91c7b0d7
tpu-inference: 3eac95a34411298ed93c6d00e7f7ef2bf681c237

# USE_MOE_EP_KERNEL=0 for gmm
# USE_MOE_EP_KERNEL=1 for fused_ep_moe
USE_MOE_EP_KERNEL=0 \
MODEL_IMPL_TYPE=vllm \
vllm serve \
  --seed=42 \
  --download_dir /mnt/disks/persist \
  --model=BCCard/Qwen3-Coder-480B-A35B-Instruct-FP8-Dynamic \
  --max-model-len=10240 \
  --max-num-batched-tokens=8192 \
  --max-num-seqs=512 \
  --no-enable-prefix-caching \
  --disable-log-requests \
  --tensor-parallel-size=8 \
  --kv-cache-dtype=fp8 \
  --gpu-memory-utilization=0.95 \
  --async-scheduling \
  --enable-expert-parallel

```

Run benchmark.

with gmm

| Scenario (In/Out) | Metric | Base (Total) | Base (Per Chip) | New (Total) | New (Per Chip) | Improvement (%) |
|:---|:---|:---|:---|:---|:---|:---|
| **1024 / 1024** | Request Throughput (req/s) | 1.31 | 0.3275 | 1.61 | 0.4025 | +22.90% |
| | Output Token Throughput (tok/s) | 1,208.16 | 302.04 | 1,492.84 | 373.21 | +23.56% |
| | Total Token Throughput (tok/s) | 2,410.01 | 602.50 | 2,977.87 | 744.47 | +23.56% |
| **8192 / 1024** | Request Throughput (req/s) | 0.70 | 0.1750 | 0.78 | 0.1950 | +11.43% |
| | Output Token Throughput (tok/s) | 645.64 | 161.41 | 720.02 | 180.01 | +11.52% |
| | Total Token Throughput (tok/s) | 5,778.65 | 1,444.66 | 6,444.37 | 1,611.09 | +11.52% |
| **1024 / 8192** | Request Throughput (req/s) | 0.18 | 0.0450 | 0.22 | 0.0550 | +22.22% |
[...]

PR #1741, opened 2026-02-18
Update GMM v1 heuristic to fix moe perf regression

helloworld1:
# Description

There is a regression on gmm v1 performance due to default block sizes too small in https://github.com/vllm-project/tpu-inference/pull/1691

Update the heuristic to reflect the original fused_moe_gmm logic.
Also updated the heuristic name to make it more clear on kernel used.

# Tests

E2E test running qwen3 coder

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/moe/utils.py
+++ b/tpu_inference/layers/jax/moe/utils.py
@@ -22,7 +22,7 @@ from jax.sharding import Mesh
 
 from tpu_inference import envs
 from tpu_inference.kernels.megablox.gmm import gmm as megablox_gmm
-from tpu_inference.layers.common.fused_moe_gmm import \
+from tpu_inference.kernels.megablox.tuned_block_sizes import \
     round_up_to_multiple_of_128_within_limit
 from tpu_inference.layers.common.moe import MoEBackend
 from tpu_inference.layers.jax.layers import FlaxUtils


tpu_inference/kernels/megablox/tuned_block_sizes.py
```diff
@@ -233,13 +233,69 @@
 }
 
 
-def get_default_gmm_block_sizes(m: int, k: int, n: int):
+# TODO (jacobplatin): make this more generic
+def round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
```
QiliangCui, line 237: I think it is the same as: https://github.com/vllm-project/tpu-inference/blob/e71fc1823d9644bd58c97c00bf97e92c8dd06eac/tpu_inference/layers/common/fused_moe_gmm.py#L65

put them together to avoid duplicate code?

QiliangCui (changes requested):

tpu_inference/kernels/megablox/tuned_block_sizes.py
kyuyeunk: or just make it a one liner

```
from jax._src.utils import ceil_of_ratio

def round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
  return min(ceil_of_ratio(x, 128), limit) 
```

kyuyeunk: but it doesn't cover `if x % candidate == 0` so it probably needs some more throught.

helloworld1: The one in fused_moe_gmm can be completely removed.

helloworld1: ceil_of_ratio does not work, it is just -(-x//y), https://github.com/jax-ml/jax/blob/main/jax/_src/util.py#L551
I didn't see a much simpler way  to capture the original logic. Let's keep it as is.

helloworld1: Removed the duplicated code now.

PR #1748, opened 2026-02-18
[Jax][FP8] Enforce load and process weights on CPU

lk-chen:
# Description

Force loading HF weights on CPU, the jax weight should be "device_put" on TPU later.

# Tests

CI

# Checklist

### DIFF
--- a/tpu_inference/layers/jax/moe/moe.py
+++ b/tpu_inference/layers/jax/moe/moe.py
@@ -23,6 +23,7 @@ from jax.sharding import NamedSharding, PartitionSpec
 from jaxtyping import Float
 
 from tpu_inference.layers.common.moe import MoEBackend
+from tpu_inference.layers.common.utils import cpu_mesh_context
 from tpu_inference.layers.jax import JaxModule
 from tpu_inference.layers.jax.base import create_param
 from tpu_inference.layers.jax.layers import FlaxUtils
@@ -30,7 +31,7 @@ from tpu_inference.layers.jax.quantization import QuantizeMethodBase
 from tpu_inference.layers.jax.quantization.configs import QuantizationConfig
 from tpu_inference.logger import init_logger
 from tpu_inference.models.jax.utils.weight_utils import (
-    cpu_mesh_context, jax_array_from_reshaped_torch, shard_put)
+    jax_array_from_reshaped_torch, shard_put)
 
 modeling_flax_utils = FlaxUtils()
 logger = init_logger(__name__)
--- a/tpu_inference/layers/jax/quantization/fp8.py
+++ b/tpu_inference/layers/jax/quantization/fp8.py
@@ -32,6 +32,7 @@ from tpu_inference.layers.common.process_weights.moe_weights import (
     FusedMoEWeights, process_fp8_moe_weights)
 from tpu_inference.layers.common.quantization import fp8 as common_fp8
 from tpu_inference.layers.common.quantization.configs import QuantLinearConfig
+from tpu_inference.layers.common.utils import cpu_mesh, cpu_mesh_context
 from tpu_inference.layers.jax import JaxModule
 from tpu_inference.layers.jax.base import create_param
 from tpu_inference.layers.jax.linear import JaxEinsum
@@ -42,8 +43,8 @@ from tpu_inference.layers.jax.quantization.unquantized import (
     UnquantizedFusedMoEMethod, UnquantizedLinearMethod)
 from tpu_inference.logger import init_logger
 from tpu_inference.models.jax.utils.weight_utils import (
-    cpu_mesh, cpu_mesh_context, jax_array_from_reshaped_torch,
-    load_nnx_param_from_reshaped_torch, shard_put)
+    jax_array_from_reshaped_torch, load_nnx_param_from_reshaped_torch,
+    shard_put)
 
 logger = init_logger(__name__)
 
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -37,15 +37,14 @@ from vllm.config import VllmConfig
 from vllm.model_executor.models.utils import AutoWeightsLoader
 
 from tpu_inference import envs, utils
-from tpu_inference.layers.common.utils import general_device_put
+from tpu_inference.layers.common.utils import (cpu_mesh_context,
+                                               general_device_put)
 from tpu_inference.layers.jax import JaxModule
 from tpu_inference.layers.jax.quantization import QuantizeMethodBase
 from tpu_inference.logger import init_logger
 from tpu_inference.models.jax.utils import file_utils
 
 logger = init_logger(__name__)
-# Lazy initialized, since device might not be ready at import time.
-_cpu_mesh = None
 
 HF_WEIGHTS_FORMAT = "*.safetensors"
 
@@ -220,7 +219,6 @@ def shard_put(x: jax.Array,
     # Single device sharding requires this special handling
     # to avoid the recursive jit error.
     if mesh is None:
-        print(f"clkbp {mesh=} using {get_mesh()=}")
         mesh = get_mesh()
 
     x_mesh = None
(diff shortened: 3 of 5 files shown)


tpu_inference/layers/jax/moe/moe.py
```diff
@@ -308,9 +297,18 @@ def _load_weights(self, weights: Iterable):
...
         }.items():
-            if getattr(param, "_cnt_moe_weights_loaded",
-                       0) == self.num_local_experts:
-                param.value = shard_put(param.value, param.sharding)
-                loaded_names.add(param_name)
+            weights_to_load = param._weights_to_load
+            if all(w is not None for w in weights_to_load):
+                with jax.set_mesh(
+                        jax.make_mesh((1, ), ('x', ),
+                                      devices=jax.devices('cpu'))):
```
patemotter: Can we pull this out into a reusable function or constant that lives in one of the utility files? 

```python
def cpu_mesh():
    return jax.make_mesh((1,), ('x',), devices=jax.devices('cpu'))

OR

_CPU_MESH = jax.make_mesh((1,), ('x',),devices=jax.devices('cpu'))

```

tpu_inference/layers/jax/quantization/fp8.py
```diff
@@ -415,18 +414,11 @@ def load_weights(self, *, layer: JaxMoE, original_load_weights_fn,
...
-                jax_param.value = jax_param.value.at[expert_id].set(jax_weight)
-            except TypePromotionError as e:
-                raise TypePromotionError(
-                    f"Error while loading weight for {torch_name} with {jax_weight.dtype=} {jax_weight.shape=} "
-                    f"into {jax_param.value.dtype=} {jax_param.value.shape=}"
-                ) from e
+
+            jax_weight = jax_array_from_reshaped_torch(
+                torch_weight, reshape_dims=(1, ) +
+                torch_weight.shape)  # add expert dim for concatenation later
```
patemotter: Nit: Can you include some more comments/notes about the rationale so it will make sense as to why this was needed for future readers?

tpu_inference/layers/jax/moe/moe.py
lk-chen: Done, added in weight_util.py

tpu_inference/layers/jax/quantization/fp8.py
lk-chen: Done

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -657,6 +657,18 @@ def transfer_state_with_mappings(src_state,
     return tgt_state
 
 
+def cpu_mesh():
```
kyuyeunk: can you move this to a common utility folder/file? and we can have this code shared in vllm model implementation as well: https://github.com/vllm-project/tpu-inference/blob/4bab9fafec2a8cc8c30909c013bc756aca07b8b3/tpu_inference/models/vllm/vllm_model_wrapper.py#L176-L178

lk-chen: I made that in https://github.com/vllm-project/tpu-inference/pull/1792

tpu_inference/layers/jax/quantization/fp8.py
```diff
@@ -280,38 +287,40 @@ def process_weights_after_loading(self, layer):
...
-        delattr(layer, 'bias')
-
-        if self.linear_config.enable_quantized_matmul_kernel:
-            # The quantized_matmul_kernel expects weight scales shaped (n_out_features, 1, n_blocks) for blockwisze quantization.
-            weights.weight_scale = jnp.expand_dims(
-                jnp.transpose(weights.weight_scale),
-                axis=1,
-            )
+        # Do the re-quant process on CPU to avoid OOM on device.
+        with cpu_mesh_context():
```
vanbasten23, line 291: I guess on torchax path tpu-inference/tpu_inference/layers/vllm/quantization/fp8.py, it's already processing on CPU, is my understanding correct?

lk-chen: Yes in torchax path, weight loading (including `process_weights_after_loading`) is on CPU context.
This PR is for jax path

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -769,12 +778,12 @@ def load_nnx_param_from_reshaped_torch(
     assert tuple(jax_weight.shape) == jax_param.value.shape, \
         f"Shape mismatch when loading weight '{param_name}': torch {jax_weight.shape} vs jax {jax_param.value.shape}"
 
-    spec = jax_param.sharding
-    if isinstance(jax_param.sharding, NamedSharding):
-        spec = jax_param.sharding.spec
-    elif isinstance(jax_param.sharding, SingleDeviceSharding):
+    spec = jax_param.get_metadata().get('sharding', ())
```
kyuyeunk, line 781: can you share what happens if this line does not exist?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -217,17 +219,23 @@ def shard_put(x: jax.Array,
...
-        return jax.device_put(x, shardings)
+        mesh = get_mesh()
+
+    x_mesh = None
+    if isinstance(x.sharding, NamedSharding):
+        x_mesh = x.sharding.mesh
 
     if math.prod(mesh.axis_sizes) == 1:
-        return jax.device_put(x, mesh.devices.flatten()[0])
+        return general_device_put(x,
```
kyuyeunk, line 229: similar to the other comment: what happens if change does not exist?

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -139,6 +141,8 @@ def apply_jax(self, layer: JaxModule, x: jax.Array) -> jax.Array:
 
             w13_weight = layer.kernel_gating_upproj_E2DF.value if layer.moe_backend == MoEBackend.FUSED_MOE else layer.kernel_gating_upproj_EDF.value
             w2_weight = layer.kernel_down_proj_EFD.value
+            w13_weight = jnp.swapaxes(w13_weight, 1, 2)
```
kyuyeunk, line 144: this doesn't seem related to cpu weight loading. was this a bug in the previous pr?

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -108,18 +109,19 @@ def process_weights_after_loading(self, layer: JaxMoE, *args,
...
             w_up = layer.kernel_up_proj_EDF.value
 
             # Fuse the weights into w13: [Gate, Up]
-            w13_val = jnp.concatenate([w_gate, w_up], axis=-1)
+            w13_val = jnp.concatenate([w_gate, w_up], axis=1)
 
             # TODO (jacobplatin): we probably want to make the sharding configurable
             layer.kernel_gating_upproj_EDF = nnx.Param(
-                w13_val, sharding=layer.edf_sharding)
+                shard_put(w13_val, shardings=layer.edf_sharding))
```
kyuyeunk, line 124: why does this need a separate `shard_put` function? why doesn't `sharding` argument work?

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: Line 234 is more critical, if I use `device_put`, disagg. serving test will [break](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/11059#019c84f1-1416-4495-8fdb-8448fccd6afd/L597). Since I already need to change line 234, it's more consistent to use `general_device_put` in all branches

tpu_inference/layers/jax/quantization/unquantized.py
lk-chen: just provide `sharding` to nnx.Param is no-op (Param won't help use shard the value). Here we are already in `process_weight_after_loading` so need to explicitly shard the jax.Arrary

tpu_inference/layers/jax/quantization/unquantized.py
lk-chen: it's silly but 

Image

previously it was [auto transposed](https://github.com/vllm-project/tpu-inference/blob/dabd50c9a53d5234e1c8e549a532bde11b6728b7/tpu_inference/models/jax/utils/weight_utils.py#L736-L739), now we explicitly tranpose.

tpu_inference/models/jax/utils/weight_utils.py
lk-chen: There will be complaint that "sharding" attribute doesn't exist. 

With jax/flax upgrade, the nxx.Param metadata is safer to be accessed this way

kyuyeunk (approved):
seems like there were bunch of interesting issues encountered during implementing this feature so all the changes makes sense. approved!

PR #1761, opened 2026-02-19
[TPU Offload][FIX] Stack KV blocks across all layers for save

juncgu-google:
# Description

Previously, we concatenated kv blocks into a large jax array in each layer for swapping out. This approach requires extra splitting operations when saving them into cpu buffer and mapping them to a certain chunk id per block.

In this PR, we stacked the data of blocks across all layers into a single jax array to eliminate the splitting operation following swap-out. Due to the change of data layout, the function of updating kv cache after swapping in is also modified.

# Tests

pytest -s -v tests/offload/tpu_offload_connector_worker_test.py
pytest -s -v tests/offload/tpu_offload_utils_test.py
pytest -s -v tests/offload/tpu_offload_accuracy_test.py

# Checklist

### DIFF
--- a/tests/offload/tpu_offload_accuracy_test.py
+++ b/tests/offload/tpu_offload_accuracy_test.py
@@ -164,6 +164,7 @@ def test_kv_cache_cpu_offloading_accuracy_larger_than_cpu_ram(
         )
 
 
+# TODO(jcgu): verify the batched_save mode
 # def test_kv_cache_cpu_offloading_batch_save_multi_request(
 #     monkeypatch: pytest.MonkeyPatch,
 #     sampling_config: SamplingParams,
@@ -175,7 +176,7 @@ def test_kv_cache_cpu_offloading_accuracy_larger_than_cpu_ram(
 #     on the host correctly distributes KV blocks to their respective requests.
 #     """
 #     num_requests = 10
-#     swap_op_types = ["pallas", "jax", "jax_copy"]
+#     swap_op_types = ["jax"]
 #     decode_save = "0"
 #     skip_precompile = "1"
 #     batched_save = "1"
--- a/tpu_inference/offload/tpu_offload_connector.py
+++ b/tpu_inference/offload/tpu_offload_connector.py
@@ -1925,7 +1925,6 @@ class TPUOffloadConnectorWorker:
     def _prepare_save_plan(
         self,
         meta: TPUReqMeta,
-        validate: bool = False,
     ) -> tuple[list[int], list[int]] | None:
         """
         Validate and plan the blocks for the save operation for the given request.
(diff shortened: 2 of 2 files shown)


tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1790,50 +1812,32 @@ def _decompose_into_buckets(self, num_blocks: int) -> list[int]:
...
-                else:
-                    chunked_dummy_kv_cpu = jax.tree.map(
-                        lambda flat_layer_cache: jax.lax.split(
-                            flat_layer_cache, split_size_list, axis=0),
-                        dummy_kv_cpu)
 
-                chunked_dummy_kv_tpu = self.swap_in_fn(chunked_dummy_kv_cpu)
-                jax.block_until_ready(chunked_dummy_kv_tpu)
+                # Warm up
+                for _ in range(num_warmup):
```
saikat-royc, line 1828: why do we need > 1 warmup per block size?

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1857,9 +1861,9 @@ def _bucketed_gather_kv_cache(
         GATHER_NUM_BLOCKS.observe(num_blocks)
         if num_blocks == 0:
             return []
-        # return KVCacheManager._jitted_gather_kv_cache(kv_caches, block_ids)
         if num_blocks in BLOCK_SIZE_BUCKETS:
-            return KVCacheManager._jitted_gather_kv_cache(kv_caches, block_ids)
+            return stack_kv_cache_cross_layers(kv_caches, block_ids,
```
saikat-royc, line 1865: block_ids and num_blocks are redundant args to the stack_kv_cache_cross_layers function. we can derive the num_blocks inside the function

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2038,62 +1935,63 @@ def _validate_and_prepare_save(
...
-        full_block_ids = meta.local_block_ids
-        full_token_ids = meta.token_ids
         blocks_to_save = save_spec.src_blocks
         dst_chunks = save_spec.dst_chunks
 
-        num_total_tokens = save_spec.num_total_tokens
-        num_skip_leading_tokens = save_spec.num_skip_leading_tokens
-        assert num_total_tokens <= len(
-            full_token_ids), f"{num_total_tokens} > {len(full_token_ids)}"
+        if validate:
```
saikat-royc: remove the if validate. we can do this assert check always.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2011,24 +1910,22 @@ def _bucketed_jitted_insert_kv_cache_slices(
...
 
             block_offset += decomposed_block_size
 
         return updated_kv_caches
 
-    def _validate_and_prepare_save(
+    def _prepare_save_plan(
         self,
         meta: TPUReqMeta,
+        validate: bool = False,
```
saikat-royc: the validate arg is not called from anywhere with value True. We can remove this arg.
Also check comment below about if validate code block. we can do the checks always I think. This is exepceted to add any signficant latency


Else this function is simply returning the save_spec.src_blocks, save_spec.dst_chunks and in that case do not even need this function call.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2219,14 +2119,15 @@ def _batched_gather_tpu_blocks(
...
-                    self.runner.kv_caches, all_src_blocks_arr)
-            jax.block_until_ready(flat_kv_caches_tpu)
+                gathered_kv_caches_tpu = stack_kv_cache_cross_layers(
+                    self.runner.kv_caches, all_src_blocks_arr,
+                    total_num_blocks_to_save)
+            jax.block_until_ready(gathered_kv_caches_tpu)
         else:
-            flat_kv_caches_tpu = None
+            gathered_kv_caches_tpu = None
         duration = time.time() - start_time
```
saikat-royc: We can move the GATHER_TPU_BLOCKS_CALLS.inc(), duration calculation and GATHER_TPU_BLOCKS_LATENCY.observe() inside the if not self.no_op_gather to track real gather activity,

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2219,14 +2119,15 @@ def _batched_gather_tpu_blocks(
         if not self.no_op_gather:
             all_src_blocks_arr = jnp.array(all_src_blocks)
             if self.use_bucketed_swap_ops:
```
saikat-royc, line 2118: maybe put an assert for the if not self.use_bucketed_swap_ops to avoid any hidden issues. same for other places. or completely remove the  if self.use_bucketed_swap_ops: in this code path.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2219,14 +2119,15 @@ def _batched_gather_tpu_blocks(
...
+                gathered_kv_caches_tpu = self._bucketed_stack_kv_caches(
                     self.runner.kv_caches, all_src_blocks_arr)
             else:
-                flat_kv_caches_tpu = KVCacheManager._jitted_gather_kv_cache(
-                    self.runner.kv_caches, all_src_blocks_arr)
-            jax.block_until_ready(flat_kv_caches_tpu)
+                gathered_kv_caches_tpu = stack_kv_cache_cross_layers(
+                    self.runner.kv_caches, all_src_blocks_arr,
+                    total_num_blocks_to_save)
+            jax.block_until_ready(gathered_kv_caches_tpu)
```
saikat-royc, line 2125: this block_until_ready can align with the  gathered_kv_caches_tpu = stack_kv_cache_cross_layers() invocation.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1874,120 +1878,15 @@ def _bucketed_gather_kv_cache(
                 block_slice = jax.lax.dynamic_slice_in_dim(
```
saikat-royc, line 1878: it would be interesting to see if this is causing any significant additional latency.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2575,8 +2462,8 @@ def _process_completed_saves(self):
         if completed_count > 0:
             duration = time.time() - start_time
             PROCESS_COMPLETED_SAVE_LATENCY.observe(duration)
-            logger.info(f"{completed_count} future(s) processed "
-                        f"in {duration:.4f} seconds.")
+            logger.info(f"collecte {completed_count} save operation "
```
saikat-royc: completed ...

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2638,78 +2526,86 @@ def start_load_kv(self, fwd_ctx: "ForwardContext") -> None:
...
-            first_dst_block = dst_blocks[0]
-            last_dst_block = dst_blocks[-1]
-            try:
-                first_block_idx_in_local = meta.local_block_ids.index(
-                    first_dst_block)
-                last_block_idx_in_local = meta.local_block_ids.index(
-                    last_dst_block)
-                if not (last_block_idx_in_local - first_block_idx_in_local + 1
-                        == len(dst_blocks)):
+            if validate:
```
saikat-royc: let's keep this validation?

tests/offload/tpu_offload_accuracy_test.py
```diff
@@ -107,10 +107,10 @@ def test_kv_cache_cpu_offloading_accuracy_smaller_then_cpu_ram(
     sampling_config: SamplingParams,
     kv_transfer_config: KVTransferConfig,
 ):
-    swap_op_types = ["pallas", "jax", "jax_copy"]
-    decode_saves = ["0", "1"]
+    swap_op_types = ["jax"]
+    decode_saves = ["0"]
     skip_precompile = ["0", "1"]
-    batched_saves = ["0", "1"]
+    batched_saves = ["0"]
```
saikat-royc, line 113: why are we removing batched save test case?

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: I was worried that jit could not correctly trace`len(block_ids)`, and put num_blocks as an input.

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: done

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: the `use_bucketed_swap_ops=True` branch also needs the block_until_ready.

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: done

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: done

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: sorry, I did not get that.
what's the issue of the not self.use_bucketed_swap_ops branch?

tpu_inference/offload/tpu_offload_connector.py
juncgu-google: done

tests/offload/tpu_offload_accuracy_test.py
juncgu-google: The `test_kv_cache_cpu_offloading_batch_save_multi_request` is failed; I want to avoid batched_save logics in the current stage.

tpu_inference/offload/tpu_offload_connector.py
saikat-royc: ack

PR #1766, opened 2026-02-20
feat(tpu): implement KV cache reset and reinitialization for RL workflows

sierraisland:
# Description

In Reinforcement Learning (RL) pipelines (such as GRPO), model weights are synchronized and resharded frequently. Currently, the HBM reserved by the KV cache is not released during these operations, leading to two major issues:
- Stale Data: Stale KV cache values from previous model versions can persist due to constant prefix keys, leading to incorrect distribution rollouts.
- Memory Pressure: High HBM fragmentation (up to ~45%) during weight sync/resharding forces users to significantly reduce --gpu-memory-utilization (often to 0.4), degrading performance.

This PR introduces two new APIs to the TPU runner and worker to manage the KV cache lifecycle explicitly.
- `reset_kv_cache()`: Explicitly deletes each JAX array within the KV cache to immediately free HBM, clears the prefix cache to prevent serving stale values, and logs HBM usage before and after to verify memory reclamation.
- `reinitialize_kv_cache()`: Reallocates empty KV cache arrays using the `KVCacheConfig` saved during the initial setup., and allows inference to resume with a clean state after weight synchronization.

The RL workflow requires a three-step sequence: (1) Delete cache -> (2) Perform heavy weight sync/resharding -> (3) Recreate cache. We have intentionally kept these as two separate RPCs because weight resharding is a memory-intensive operation that needs the HBM buffer freed by the reset step to be available.

Related bug: [b/481371885](https://b.corp.google.com/issues/481371885)

# Tests

1. Unit Tests in tests/runner/test_kv_cache_manager.py:
- test_reset_kv_cache: Verifies that JAX arrays are deleted and the manager state is cleared.
- test_reinitialize_kv_cache: Verifies that fresh arrays are correctly reallocated with expected shapes.
- test_reinitialize_kv_cache_without_init_raises: Ensures proper error handling if reinit is called without prior initialization.
2. End-to-End Verification
Verified using examples/offline_inference.py by performing back-to-back generations with a reset/reinit cycle in between:
```
    print("\n" + "="*30)
    print(">>> PHASE 1: BEFORE RESET")
    print("="*30)

    llm.collective_rpc("reset_kv_cache")
    
    print("\n" + "="*30)
    print(">>> PHASE 2: AFTER RESET (HBM should be low now)")
    print("="*30)

    llm.collective_rpc("reinitialize_kv_cache")

    print("\n" + "="*30)
    print(">>> PHASE 3: AFTER RE-INIT (HBM should be back)")
    print("="*30)
```

# Checklist


examples/offline_inference.py
```diff
@@ -35,6 +35,9 @@ def main(args: dict):
     # Create an LLM
     llm = LLM(**args)
 
+    llm.collective_rpc("reset_kv_cache")
+    llm.collective_rpc("reinitialize_kv_cache")
```
wenxindongwork: can you remove changes to this file?

tpu_inference/core/core_tpu.py
```diff
@@ -784,3 +794,20 @@ def run_busy_loop(self):
 
     def shutdown(self):
         self._orchestrator.shutdown()
+
+    def reset_kv_cache(self) -> None:
+        """Delete KV cache JAX arrays on all workers **and** clear prefix cache.
+        See :pymeth:`DisaggEngineCore.reset_kv_cache` for details.
```
wenxindongwork: could you format the comments here?

tpu_inference/core/core_tpu.py
```diff
@@ -784,3 +794,20 @@ def run_busy_loop(self):
...
+        """
+        for engine in itertools.chain(self._prefill_engines,
+                                      self._decode_engines):
+            engine.collective_rpc("reset_kv_cache")
+            engine.scheduler.reset_prefix_cache()
```
wenxindongwork: can we remove this line?
` engine.scheduler.reset_prefix_cache()`?
we want the same behavior in both the disagg and non disagg cases

examples/offline_inference.py
```diff
@@ -93,6 +93,11 @@ def main(args: dict):
     if profiler_config.profiler == "torch":
         llm.stop_profile()
 
+    llm.collective_rpc("reset_kv_cache")
```
wenxindongwork: @sierraisland second thoughts... do you think `delete_kv_cache` will be a better name?

sierraisland: yes, `delete_kv_cache` is better. Let me change to it.

examples/offline_inference.py
```diff
@@ -93,6 +93,11 @@ def main(args: dict):
     if profiler_config.profiler == "torch":
         llm.stop_profile()
 
+    llm.collective_rpc("delete_kv_cache")
```
wenxindongwork: can we remove all changes to this file?

tpu_inference/core/core_tpu.py
```diff
@@ -784,3 +794,15 @@ def run_busy_loop(self):
 
     def shutdown(self):
         self._orchestrator.shutdown()
+
+    def delete_kv_cache(self) -> None:
+        """Delete KV cache JAX arrays on all workers **and** clear prefix cache. See :pymeth:`DisaggEngineCore.delete_kv_cache` for details."""
```
wenxindongwork: here, we are not clearing the prefix cache. Can you update the comments? also a small nit comment, could we remove the `See :pymeth:`DisaggEngineCore.delete_kv_cache` for details.` part? (and same for reinitialize_kv_cache)

PR #1770, opened 2026-02-20
Update jax to 0.8.3

kyuyeunk:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,8 +5,9 @@ pytest-mock
 absl-py
 numpy
 google-cloud-storage
-jax[tpu]==0.8.1
-jaxlib==0.8.1
+jax[tpu]==0.8.3
+jaxlib==0.8.3
+libtpu==0.0.32
 jaxtyping
 flax==0.12.4
 torchax==0.0.11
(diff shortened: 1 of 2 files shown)


jrplatin:
Have we verified that this won't cause regressions @kyuyeunk ? Also I believe @patrickji2014 is already working on this?

https://github.com/vllm-project/tpu-inference/pull/1737/changes

vipannalla:
I stamped to unblock, but please add CI run (or test details) to show this doesn't regress on perf or correctness.

patrickji2014:
> Have we verified that this won't cause regressions @kyuyeunk ? Also I believe @patrickji2014 is already working on this?
> 
> https://github.com/vllm-project/tpu-inference/pull/1737/changes

Yes I have one that upgrades jax to 0.9.0 but I observe performance regression after upgrade to 0.9.0(captured here b/485649545). 0.8.3 doesn't have this issue. You can merge this in if no failures. I will follow up with an upgrade to 0.9.0 after we figure out why the performance drops in b/485649545

PR #1771, opened 2026-02-20
Fix Value tensor dtype mismatch in Qwen3 attention.

abhinavclemson:
# Description

This PR casts the value tensor (v) to match the key tensor (k)'s data type in Qwen3Attention before passing it to the vLLM ragged paged attention kernel.

Currently, running inference or evaluation with Qwen3 models crashes with the following error:
```
File tpu_inference/kernels/ragged_paged_attention/v3/kernel.py:1261, in static_validate_inputs(queries, keys, values, kv_cache, kv_lens, page_indices, cu_q_lens, distribution, sm_scale, sliding_window, soft_cap, mask_value, q_scale, k_scale, v_scale, chunk_prefill_size, num_kv_pages_per_block, num_queries_per_block, vmem_limit_bytes, debug_mode)
   1259 # Note: we expect the kv quantization happens outside of the RPA kernel.
   1260 if not (kv_cache.dtype == k.dtype == v.dtype):
-> 1261     raise ValueError(
   1262         f"Expected {kv_cache.dtype=} to be equal to {k.dtype=} and {v.dtype=}."
   1263     )
   1264 # Integer kv quantization is currently not supported.
   1265 if not jnp.issubdtype(kv_cache.dtype, jnp.floating):

ValueError: Expected kv_cache.dtype=dtype(bfloat16) to be equal to k.dtype=dtype(bfloat16) and v.dtype=dtype('float32').
```

FIXES: b/484414167

# Tests

Validated on maxtext side: [script](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/maxtext/examples/sft_qwen3_demo.ipynb)

# Checklist

### DIFF
--- a/tpu_inference/models/jax/qwen3.py
+++ b/tpu_inference/models/jax/qwen3.py
@@ -159,6 +159,7 @@ class Qwen3Attention(JaxModule):
 
         # v: (T, K, H)
         v = self.v_proj(x)
+        v = v.astype(k.dtype)
         # o: (T, N, H)
         q_scale = k_scale = v_scale = None
         if self.kv_cache_quantized_dtype:


vipannalla:
By Qwen3, I assume you are referring to Qwen3-32B dense model. cc @jrplatin to also take a look.

tpu_inference/models/jax/qwen3.py
```diff
@@ -159,6 +159,7 @@ def __call__(
 
         # v: (T, K, H)
         v = self.v_proj(x)
+        v = v.astype(k.dtype)
```
kyuyeunk, line 162: curious why the dtype is different at all. can you do some analysis on where the dtype delta is coming from?

jrplatin (changes requested):
Agree with Kyuyeun -- I'd like to understand why this is happening before I approve

tpu_inference/models/jax/qwen3.py
abhinavclemson: Last time I checked. Problem seems to be,on tpu_inference side, in Qwen3Attention class the query (q) and key (k) tensors are explicitly cast to bfloat16 by the apply_rope function, but the value (v) tensor skips RoPE and remains float32.

kyuyeunk: couple of questions
- is my understanding correct that this `__call__` function is set to return `fp32` by default? if that is the case, can you correct the behavior on `JaxEinsum` level?: https://github.com/vllm-project/tpu-inference/blob/4819e3e3f4e6bf9bda2305b05a6294b12b30d7f7/tpu_inference/layers/jax/linear.py#L81-L88
- if that is the case, do you know if the error goes away when using vllm's model implementation? i.e., `MODEL_IMPL_TYPE=vllm`?

PR #1797, opened 2026-02-24
docs: Streamline installation instructions and validation script

RobMulla:
Refines the vLLM TPU installation documentation by recommending `uv` for installation. Unifies the v6e/v7x pip installation into a single command now that jax 0.8.3 removes the branching requirement. Improves the installation verification python script to explicitly print package versions and active jax backends.

### DIFF
--- a/docs/getting_started/quickstart.md
+++ b/docs/getting_started/quickstart.md
@@ -19,7 +19,7 @@ For detailed steps on installing `vllm-tpu` with `pip` or running it as a Docker
 We recommend using [uv](https://docs.astral.sh/uv/) (`uv pip install`) instead of standard `pip` as it improves installation speed.
 
 ```shell
-pip install vllm-tpu
+uv pip install vllm-tpu
 ```
 
 ## Verify Installation


docs/getting_started/quickstart.md
```diff
@@ -16,19 +16,34 @@ Google Cloud TPUs (Tensor Processing Units) accelerate machine learning workload
...
 
-!!! important
-    Until jax=0.8.3 is released, v7x will have different build requirements from previous TPU generations (v6e and prior). As a result, please use the following until consolidation is complete:
-
-**For v6e and prior**
+We recommend using [uv](https://docs.astral.sh/uv/) (`uv pip install`) instead of standard `pip` as it improves installation speed.
 
 ```shell
-pip install vllm-tpu --version=0.13.2.post6
+pip install vllm-tpu
```
jcyang43: Do we want to update this to use uv (given that we just recommended users to use uv a few lines above)

jcyang43 (approved):

docs/getting_started/quickstart.md
RobMulla: Great point! I've updated the quickstart to include uv

PR #1800, opened 2026-02-24
[GMM] Unit tests for upcoming support for misaligned input shapes in gmm_v2

rohan-reddy:
## Description

Adds unit tests for the upcoming implicit padding feature in `gmm_v2`. All tests are currently skipped (`@pytest.mark.skip(reason="padding not yet implemented")`) and will be enabled once the kernel-side padding logic lands.

These tests validate that `gmm_v2()` produces correct results when input dimensions (in_size, out_size) are not aligned to TPU lane boundaries.

## Changes

**`tests/kernels/gmm_test.py`**
- `test_gmm_implicit_padding`: misaligned `in_size`, `out_size` with optional bias (24 cases)
- `test_gmm_weight_quantized_padding`: misaligned shapes with quantized weights (`int8`, `float8_e4m3fn`)

## Testing
All tests are currently skipped. When not skipped, they fail with expected errors:
- `test_gmm_implicit_padding` and `test_gmm_weight_quantized_padding`: `NotImplementedError: Implicit padding along lane dimensions is not supported.` Fails with this error since K, N dims are misaligned, so lane alignment check fails upfront.

## Checklist

- [x] Self-review completed
- [x] New tests added for padding correctness with misaligned shapes (skipped until kernel changes land)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -887,8 +894,21 @@ def gmm_v2(
         rhs_bias = rhs_bias.astype(jnp.float32)
         rhs_bias_spec = pl.BlockSpec(memory_space=pltpu.HBM)
 
+    # Pad inputs to match padded dimensions when needed.
+    m_pad = dims.size_m - orig_m
+    k_pad = dims.size_k - orig_k
+    n_pad = dims.size_n - orig_n
+    if m_pad > 0 or k_pad > 0:
+        lhs = jnp.pad(lhs, ((0, m_pad), (0, k_pad)))
```
kyuyeunk: since this function gets called during inference time, i think this will likely cause performance overhead. can you create an trace file (guide: https://github.com/vllm-project/tpu-inference/blob/main/docs/profiling.md) and share how line of code looks like in the traces? note: please use a large batch size since the overhead from padding will likely be negligible in small batch size.

if it does trigger a performance overhead, please think of a different implementation to remove such overhead.

rohan-reddy: Per our conversation offline, the PR is now updated to have unit tests only pending a follow up with implementation of the shape support

tests/kernels/gmm_test.py
```diff
@@ -297,6 +298,158 @@ def test_gmm_activation_weight_quantized(
 
         self.assertArraysAllClose(actual, expected, atol=1.1, rtol=1.1)
 
+    @parameterized.product(
+        batch_size=[129, 135, 255],
```
kyuyeunk: can you remove non aligned batch size case? this is something that we don't need to support since we will pad it anyway: https://github.com/vllm-project/tpu-inference/blob/c13c91cf18e0be4020a9f6aa8f1df99faf041bf5/tpu_inference/runner/utils.py#L64-L95

tests/kernels/gmm_test.py
```diff
@@ -297,6 +298,158 @@ def test_gmm_activation_weight_quantized(
...
+    @parameterized.product(
+        batch_size=[33, 65],
+        in_size=[256],
+        out_size=[256],
+        num_groups=[4],
+        has_bias=[False],
+        group_offset=[0],
+    )
+    @pytest.mark.skip(reason="padding not yet implemented")
+    def test_gmm_implicit_padding_small_m(self, batch_size, in_size, out_size,
```
kyuyeunk: on a similar note, we probably don't need this unit test.

kyuyeunk (approved):

PR #1807, opened 2026-02-25
fix(tpu): align min_req_size to next_power_of_2 for non-p2 DP support

sierraisland:
# Description

- Problem: TPU inference fails with an AssertionError at runner/utils.py:54 when using non-power-of-two Data Parallel (DP) sizes (e.g., DP=6).
- Fix: Align min_req_size to the next power of two using next_power_of_2(self.dp_size) in tpu_runner.py to satisfy existing padding logic constraints.

FIXES: b/483122466

# Tests

- Added unit test `test_min_req_size_alignment` in test_utils.py.
- Verified offline inference runs successfully with `--data_parallel_size=6`.

# Checklist


wenxindongwork (approved):
LGTM, thanks for the quick fix!

PR #1825, opened 2026-02-28
[Qwen 3] promote_dtype_for_stats flag to optionally casting RMSNorm to FP32

JiriesKaileh:
# Description

Setting promote_dtype_for_stats=False keeps RMSNorm at BF16. 
Setting promote_dtype_for_stats=True casts RMSNorm to FP32.

Tested perf and accuracy differences in this [bug](https://b.corp.google.com/issues/453705319)

### DIFF
--- a/tpu_inference/layers/jax/rope_interface.py
+++ b/tpu_inference/layers/jax/rope_interface.py
@@ -157,8 +157,11 @@ def apply_rope(
         rope_angles = int(rope_proportion * head_dim // 2)
         nope_angles = head_dim // 2 - rope_angles
 
-        fraction = 2 * jnp.arange(0, rope_angles, dtype=jnp.float32) / head_dim
-        timescale = 1.0 / (rope_theta**fraction)
+        # 1. Cast the arange to the working dtype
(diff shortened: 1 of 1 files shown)


tpu_inference/models/jax/qwen3.py
```diff
@@ -97,15 +97,16 @@ def __init__(self,
...
-        )
+        self.q_norm = JaxRmsNorm(self.head_dim,
+                                 epsilon=self.rms_norm_eps,
+                                 dtype=dtype,
+                                 scale_init=nnx.with_partitioning(
+                                     init_fn, (None, )),
+                                 rngs=rng,
+                                 quant_config=quant_config,
+                                 prefix=prefix + ".q_norm",
+                                 promote_dtype_for_stats=False)
```
jrplatin, line 108: Can you please make a quick comment indicating that this maintains RMSNorm in BF16 (as opposed to upcasting to FP32), since it's not immediately obvious from the param name? Thanks!

jrplatin:
Also, please include the 1k/4k and MMLU results here (with repro commands) in the Testing seciton, thanks!

PR #1827, opened 2026-03-01
[Jax] Weight loading: process_weights_after_loading return bool

lk-chen:
# Description

Bookmark visited module, skip if already visited.

# Tests

commands to reproduce.

# Checklist


jrplatin:
LGTM, but not sure I 100% understand why this is needed (I know it _is_ needed it's just not clear to me why)

lk-chen:
> LGTM, but not sure I 100% understand why this is needed (I know it _is_ needed it's just not clear to me why)

@jrplatin This is not super necessary, because we have the alternative to do this kind of bookmarking inside each quant_method (e.g. through a self._processed_weights_after_loading flag, and early return if the flag is marked), but this PR wants to book mark this info at higher level

PR #1839, opened 2026-03-03
Fix recompilation check code

kyuyeunk:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -241,7 +241,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
 
         self.devices = devices
         self.dtype = self.model_config.dtype
-        self.maybe_forbid_compile = runner_utils.ForbidCompile(
+        self.maybe_forbid_compile = runner_utils.no_tracing(
         ) if envs.VLLM_XLA_CHECK_RECOMPILATION else nullcontext()
         self.dp_size = self.vllm_config.sharding_config.total_dp_size
         self.rank = rank
--- a/tests/runner/test_utils.py
+++ b/tests/runner/test_utils.py
@@ -8,15 +8,13 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 import pytest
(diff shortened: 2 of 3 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -241,7 +241,7 @@ def __init__(
 
         self.devices = devices
         self.dtype = self.model_config.dtype
-        self.maybe_forbid_compile = runner_utils.ForbidCompile(
+        self.maybe_forbid_compile = jax.no_tracing(
```
vanbasten23: This "disallow tracing for JIT compilation". How does it raise an Exception when a recompilation happens?

kyuyeunk: this line gets triggered: https://github.com/jax-ml/jax/blob/c0b2687098239f19c53a2d55e2f8ec42eafa3e15/jax/_src/pjit.py#L252-L253

but i found that the error is not correctly caught and vllm throws some random error message that isn't really useful. introducing manual try / except code yields better error message like this. let me see if i can clean up that code and push it to this branch.

```
RuntimeError("re-tracing function compute_logits_func at /home/kyuyeunk_google_com/workspace/tpu-inference/tpu_inference/models/vllm/vllm_model_wrapper.py:292 for `jit`, but 'no_tracing' is set")
```

vanbasten23: Do you know why the previous approach doesn't work? I remember I fixed some recompilation issue caught by the previous approach.

kyuyeunk: not 100% sure what kind of changes were introduced to jax that might have caused the change. but regardless of why, the previous approach was to override the private function (`jax._src.interpreters.pxla._cached_lowering_to_hlo` ) and check if that gets invoked or not - which is really fragile because 1. JAX might reset the overriding 2. it's using private function so no guarantee that it will always work.

Meanwhile, `jax.no_tracing` is an official api provided by JAX to check for recompilation. so it should be much more robust.

PR #1846, opened 2026-03-04
[Jax][Deepseek] Quantize MLA weights to FP8

lk-chen:
# Description

In #1845 we kept MLA weights in unquantized format. This PR re-quantize them to FP8.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/jax/deepseek_v3.py
+++ b/tpu_inference/models/jax/deepseek_v3.py
@@ -40,6 +40,7 @@ from tpu_inference.layers.common.quantization import (dequantize_tensor,
                                                       quantize_kv)
 from tpu_inference.layers.common.sharding import \
     ShardingAxisNameBase as ShardingAxisName
+from tpu_inference.layers.common.utils import cpu_mesh_context
 from tpu_inference.layers.jax import JaxModule
 from tpu_inference.layers.jax.attention.attention import AttentionMetadata
 from tpu_inference.layers.jax.base import _init_fn as init_fn
@@ -534,8 +535,7 @@ class MLAEinsum(JaxEinsum):
             return
         assert self.quant_config is not None
         # After loading, split the weights into k/v
-        with jax.set_mesh(
-                jax.make_mesh((1, ), ('x', ), devices=jax.devices('cpu'))):
+        with cpu_mesh_context():
             dequantized_weight = dequantize_tensor(
                 self.weight,
                 self.weight_scale_inv,
@@ -581,12 +581,17 @@ class MLAEinsum(JaxEinsum):
                 prefix=mla_layer.prefix + ".v_up_proj",
                 quant_config=self.quant_config,
             ))
+        # Cannot apply anh_sharding to scales, otherwise it complains about shape mismatch.
         mla_layer.k_up_proj.weight.value = shard_put(
             k_ANH_weight, self.mla_layer.anh_sharding)
         mla_layer.k_up_proj.weight_scale_inv.value = shard_put(k_N1A_scale, ())
         mla_layer.v_up_proj.weight.value = shard_put(
             v_ANH_weight, self.mla_layer.anh_sharding)
         mla_layer.v_up_proj.weight_scale_inv.value = shard_put(v_N1H_scale, ())
+
+        delattr(self, 'weight')
+        delattr(self, 'weight_scale_inv')
+        delattr(self, 'quant_method')
 
 
 @dataclass(kw_only=True)


tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -477,31 +487,127 @@ def _ragged_paged_attention(q, k, v, cache, seq_lens, block_tables,
...
+            weight_loader = getattr(param, "weight_loader")
+            weight_loader(param, weight)
+            self.loaded.add(name)
+        if len(self.loaded) != len(named_params):
+            return
+        assert self.quant_config is not None
+        # After loading, split the weights into k/v
+        with jax.set_mesh(
+                jax.make_mesh((1, ), ('x', ), devices=jax.devices('cpu'))):
+            dequantized_weight = dequantize_tensor(
```
gpolovets1, line 539: Can you speak more about the mental model for pushing processing steps into deepseek_v3.py?
My understanding is that process_weights_after_loading is called after load_module. So is the only reason this works because 3D process_weights_after_loading is a no-op? Otherwise, I assume process_weights_after_loading would have been double called.

Also, is it cleaner to add an MLA weight generator like I previously tried and add the steps below into fp8.py for 3D weight tensor logic (i.e. what you previously had under self.batch_features)?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -477,31 +487,127 @@ def _ragged_paged_attention(q, k, v, cache, seq_lens, block_tables,
...
+            JaxEinsum(
+                einsum_str="TNA,ANH->TNH",
+                kernel_shape=(A, N, v_head_dim),
+                rngs=nnx.Rngs(0),
+                prefix=mla_layer.prefix + ".v_up_proj",
+                quant_config=self.quant_config,
+            ))
+        mla_layer.k_up_proj.weight.value = shard_put(
+            k_ANH_weight, self.mla_layer.anh_sharding)
+        mla_layer.k_up_proj.weight_scale_inv.value = shard_put(k_N1A_scale, ())
```
gpolovets1, line 587: Why don't we utilize anh_sharding for the scales?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -477,31 +487,127 @@ def _ragged_paged_attention(q, k, v, cache, seq_lens, block_tables,
...
+            mla_layer=self,
+            einsum_str="SA,AL->SL",
+            kernel_shape=(self.kv_lora_rank,
+                          self.N * (self.qk_nope_head_dim + self.v_head_dim)),
             rngs=rngs,
             quant_config=self.quant_config,
             param_dtype=self.dtype,
-            kernel_init=nnx.with_partitioning(weight_init, self.anh_sharding),
-            prefix=self.prefix + ".v_up_proj",
+            kernel_init=nnx.with_partitioning(weight_init, self.ap_sharding),
```
gpolovets1, line 614: Does this mean that we store random weights for kv_b_proj on TPU even though it's a temporary buffer during MLA?

tpu_inference/models/jax/deepseek_v3.py
lk-chen: > My understanding is that process_weights_after_loading is called after load_module

which load_module are you referring to? Neither `DeepseekV3BaseAttention` or `DeepseekV3MLA` has load_module or load_weights

> Otherwise, I assume process_weights_after_loading would have been double called.

`process_weights_after_loading` is operating on self.weight and self.weight_scale_inv, which is not used in forward path. MLA forward path only uses v_up_proj and k_up_proj

> is it cleaner to add an MLA weight generator

I don't think so, that approach took several days and didn't work. This approach works without further debug. I guess that speaks for itself.

tpu_inference/models/jax/deepseek_v3.py
lk-chen: I don't think so, Line 538 explicitly uses CPU context. If kv_b_proj is on TPU, line 539 will have exception

tpu_inference/models/jax/deepseek_v3.py
lk-chen: It will cause error 
```
ValueError: One of device_put args was given the sharding of NamedSharding(mesh=Mesh('data': 1, 'attn_d
p': 8, 'attn_dp_expert': 1, 'expert': 1, 'model': 1, axis_types=(Auto, Auto, Auto, Auto, Auto)), spec=PartitionSpec(None, ('attn_dp
', 'attn_dp_expert', 'model', 'expert'), None), memory_kind=device), which implies that the global size of its dimension 1 should b
e divisible by 8, but it is equal to 1 (full shape: (128, 1, 512))
```

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: I see process_weights_after_loading being called in load_module: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/jax/utils/weight_utils.py#L862.
And load_module calls load_weights first, so wouldn't that mean you would be calling proces_weights_after_loading after your processing logic in MLAEinsum's load_weights?

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: Okay, but i think for this to work we would need to assume that the weight is AG'd before applying the scale? Maybe worth at least adding a comment here about this assumption.

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: Should we keep the init unsharded then?

tpu_inference/models/jax/deepseek_v3.py
lk-chen: a. I wasn't argue how many times `proces_weights_after_loading` will be called here, I said 
> process_weights_after_loading is operating on self.weight and self.weight_scale_inv, which is not used in forward path. MLA forward path only uses v_up_proj and k_up_proj

which means no matter how many times `proces_weights_after_loading` is called, it doesn't affect forward path

b. the line 862 you referenced is exactly there to prevent double calling 
image

tpu_inference/models/jax/deepseek_v3.py
lk-chen: I don't quite follow. Providing "intended sharding" through "kernel_init" is common practice everywhere

tpu_inference/models/jax/deepseek_v3.py
lk-chen: Done

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: I think it's a special case because we don't end up using kv_b_proj during MLA other than to store its weights into k_up_proj/v_up_proj. So it was a bit confusing to me why we set its shardings.
But feel free to keep it if you feel strongly that it follows best practices.

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: Chatted offline and confirmed that process_weights_after_loading will not be called again on k_up_proj/v_up_proj.

Issue #1847, opened 2026-03-04
[Bug]: MLA (DeepSeek V3) crashes with TP > 1


bgchun-fs:
### Your current environment

<details>
<summary>The output of commands above</summary>

```text
N/A
```

</details>

### 🐛 Describe the bug

MLA models (e.g. DeepSeek V3) are not compatible with `tensor_parallel_size > 1`.
There are two places in the code that assume standard multi-head attention behavior, which does not hold for MLA.

**Wrong assertion for MLA**

`get_kv_cache_shape_with_mesh` asserts `actual_num_kv_heads % model_cnt == 0` before the `if use_mla / else` branch. MLA has `num_kv_heads=1` which is not divisible by any TP size greater than 1, so this always crashes for MLA + TP > 1. The assertion only makes sense for standard attention where KV heads are split across TP devices — MLA does not shard by KV heads at all.

**Wrong KV cache sharding for MLA**

Even after fixing the assertion, the MLA KV cache is incorrectly sharded across TP devices on the page dimension. Since `page_indices` remain global, each device only sees `1/TP` of the pages — causing out-of-bounds access in the Pallas kernel. Since MLA has `num_kv_heads=1` and cannot shard on the head dimension, the KV cache should be replicated across TP devices instead.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues and checked the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

lk-chen:
which command are you using? Jax path won't work until https://github.com/vllm-project/tpu-inference/pull/1845

cc @kyuyeunk

kyuyeunk:
FYI, this pr has a command for torchax path: https://github.com/vllm-project/tpu-inference/pull/1785

PR #1849, opened 2026-03-04
Fix mla assertion

bgchun-fs:
# Description

Fix wrong assertion in `get_kv_cache_shape_with_mesh` that causes MLA models (e.g. DeepSeek V3) to crash with `tensor_parallel_size > 1`.

The assert `actual_num_kv_heads % model_cnt == 0` was placed before the `if use_mla / else` branch. MLA has `num_kv_heads=1` which is not divisible by any TP size greater than 1, so it always fired for MLA + TP > 1.
The assertion only makes sense for standard attention where KV heads are split across TP devices — MLA does not shard by KV heads at all.

Move the assert into the `else` block where it belongs, directly above `actual_num_kv_heads // model_cnt` which it protects.

FIXES: #1847

# Tests

Tested with DeepSeek V3 and `tensor_parallel_size > 1`.

# Checklist

### DIFF
--- a/tpu_inference/runner/kv_cache.py
+++ b/tpu_inference/runner/kv_cache.py
@@ -45,17 +45,19 @@ def get_kv_cache_shape_with_mesh(mesh: Mesh,
     axis_name = ShardingAxisName.ATTN_HEAD
     model_cnt = utils.get_mesh_shape_product(mesh, axis_name)
 
-    assert actual_num_kv_heads % model_cnt == 0
     # NOTE(chengjiyao): Currently, the attention kernel is tailored to the
     # specific model, rather than being determined by the head_dim. If new
     # models are introduced with a head_dim of 64, this will require additional
     # model-specific adjustments.
     if use_mla:
+        # No assertion needed: MLA compresses all KV into a single latent vector,
+        # so actual_num_kv_heads is never used in mla.get_kv_cache_shape().
         get_kv_cache_shape_fn = mla.get_kv_cache_shape
         shape = list(
             get_kv_cache_shape_fn(total_num_pages, page_size, actual_head_dim,
                                   kv_dtype))
     else:
+        assert actual_num_kv_heads % model_cnt == 0
         get_kv_cache_shape_fn = (
             rpa_hd64.get_kv_cache_shape if actual_head_dim == 64 \
                 else rpa.get_kv_cache_shape


bgchun-fs:
> does this issue happens even when attention dp is enabled using this argument? `--additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}'`

@kyuyeunk 
When enable_dp_attention is enabled, attn_dp is set to the same value as TP, which internally reduces TP to 1, so the assertion never fires(if enable_dp_attention set to 1). 
- https://github.com/vllm-project/tpu-inference/blob/21fe79fb3639e07f6f00d75c829a6ac40e4a3985/tpu_inference/layers/common/sharding.py#L175-L178

kyuyeunk:
can you check if we can remove the assertion completely if it does not cause an issue for non-mla / mha as well?

bgchun-fs:
> can you check if we can remove the assertion completely if it does not cause an issue for non-mla / mha as well?

@kyuyeunk 
I'm not sure we can always guarantee that all non-MLA models' num_kv_heads is divisible by the configured TP size. I think it is worth keeping it as a safeguard.

kyuyeunk:
> > can you check if we can remove the assertion completely if it does not cause an issue for non-mla / mha as well?
> 
> @kyuyeunk I'm not sure we can always guarantee that all non-MLA models' num_kv_heads is divisible by the configured TP size. I think it is worth keeping it as a safeguard.

hmm i'm just trying to create a more generalized solution. so the question is
1. for simplicity sake, in this context, can we just treat mla as a similar thing as "attention with num kv heads = 1"?
2. if previous statement is true, why do we not need assertion check for mla but need assertion check for "attention with num kv heads = 1"?

bgchun-fs:
> > > can you check if we can remove the assertion completely if it does not cause an issue for non-mla / mha as well?
> > 
> > 
> > @kyuyeunk I'm not sure we can always guarantee that all non-MLA models' num_kv_heads is divisible by the configured TP size. I think it is worth keeping it as a safeguard.
> 
> hmm i'm just trying to create a more generalized solution. so the question is
> 
> 1. for simplicity sake, in this context, can we just treat mla as a similar thing as "attention with num kv heads = 1"?
> 2. if previous statement is true, why do we not need assertion check for mla but need assertion check for "attention with num kv heads = 1"?

@kyuyeunk  that's very sharp approach

Seems MLA path doesn't actually use `actual_num_kv_heads` since all KV information is compressed into a single latent vector. This is evident from the function signatures — `actual_num_kv_heads` is not even passed to `mla.get_kv_cache_shape()`, and the MLA kernel has no `num_kv_heads` parameter at all:
I think the `num_kv_heads=1` assumption is just a code-level simplification. There's no reason to assert on a value that is never used in the MLA path. So we can remove assertion for this case.
```python
# mla
def get_kv_cache_shape(
        total_num_pages,
        page_size,
        kv_dim,
        kv_dtype,
)

# rpa / rpa_hd64
def get_kv_cache_shape(
        total_num_pages,
        page_size,
        actual_num_kv_heads,
        actual_head_dim,
        kv_dtype,
)
```

For the non-MLA path, as above example shows `actual_num_kv_heads` is used to calculate the per-chip KV cache shape. Without the assertion, an invalid TP config (e.g. `num_kv_heads=1, TP=8` → `1 // 8 = 0` heads per chip) would silently produce a broken KV cache instead of a clear error. So I think keep assertion for non-MLA path is worth it.

kyuyeunk:
> > > > can you check if we can remove the assertion completely if it does not cause an issue for non-mla / mha as well?
> > > 
> > > 
> > > @kyuyeunk I'm not sure we can always guarantee that all non-MLA models' num_kv_heads is divisible by the configured TP size. I think it is worth keeping it as a safeguard.
> > 
> > 
> > hmm i'm just trying to create a more generalized solution. so the question is
> > 
> > 1. for simplicity sake, in this context, can we just treat mla as a similar thing as "attention with num kv heads = 1"?
> > 2. if previous statement is true, why do we not need assertion check for mla but need assertion check for "attention with num kv heads = 1"?
> 
> @kyuyeunk that's very sharp approach
> 
> Seems MLA path doesn't actually use `actual_num_kv_heads` since all KV information is compressed into a single latent vector. This is evident from the function signatures — `actual_num_kv_heads` is not even passed to `mla.get_kv_cache_shape()`, and the MLA kernel has no `num_kv_heads` parameter at all: I think the `num_kv_heads=1` assumption is just a code-level simplification. There's no reason to assert on a value that is never used in the MLA path. So we can remove assertion for this case.
> 
> ```python
> # mla
> def get_kv_cache_shape(
>         total_num_pages,
>         page_size,
>         kv_dim,
>         kv_dtype,
> )
> 
> # rpa / rpa_hd64
> def get_kv_cache_shape(
>         total_num_pages,
>         page_size,
>         actual_num_kv_heads,
>         actual_head_dim,
>         kv_dtype,
> )
> ```
> 
> For the non-MLA path, as above example shows `actual_num_kv_heads` is used to calculate the per-chip KV cache shape. Without the assertion, an invalid TP config (e.g. `num_kv_heads=1, TP=8` → `1 // 8 = 0` heads per chip) would silently produce a broken KV cache instead of a clear error. So I think keep assertion for non-MLA path is worth it.

makes sense! can you add a comment that describes why the assertion is not needed for mla case? i've triggered ci, so other than that, as long as ci passes, we can merge it.

bgchun-fs:
@kyuyeunk updated :)

PR #1855, opened 2026-03-04
Used condition variable for event polling

datenglin:
# Description

Before this change, the code uses `sleep` for event polling which has the drawback of potential delay of up to 0.05 seconds. This change uses `threading.Condition` so we can wait for the signal of scheduler having the requests.

# Tests

It passed the Buildkite: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/11630/steps/canvas


datenglin:
OK I think [this error](https://github.com/vllm-project/tpu-inference/actions/runs/22682912849/job/65757993967?pr=1855) is basically about permission. I will try contributing directly to the upstream (original repo) instead of the origin (forked repo). If that works this PR is not needed.

PR #1859, opened 2026-03-04
remove all unnecssary all-gather within attn layer, via correcty sharding of w_uk and w_uv

gxd3:
# Description
remove all unnecssary all-gather within attn layer, via correcty sharding of w_uk and w_uv

(using single-host, 5 layers of DS model for faster test & iteration, using 1.5K decode batch size to simulate max throughput config for singlehost)

before this PR: http://xprof/trace_viewer.html?session_id=gxd-12216786850413704240
5 layer step time: 33.7ms

after this RP http://xprof.corp.google.com/trace_viewer/gxd-1053712665842639110
5 layers step time: 29.7ms

# Tests
Quality test no diff:
python ./tpu-inference/scripts/vllm/benchmarking/benchmark_serving.py --backend vllm --model deepseek-ai/DeepSeek-R1 --dataset-name mmlu --dataset-path /home/gxd_google_com/work-dir/mmlu/data/test --num-prompts 1000 --run_eval --temperature 0

```
Benchmark duration (s):                  33.75     
Total input tokens:                      173929    
Request throughput (req/s):              29.63     
Output token throughput (tok/s):         59.26     
Total token throughput (tok/s):          5212.50   
Mean TTFT (ms):                          18051.63  
Median TTFT (ms):                        16857.51  
P99 TTFT (ms):                           33622.56  
Mean TPOT (ms):                          1395.42   
Median TPOT (ms):                        1450.03   
P99 TPOT (ms):                           1459.62   
Mean ITL (ms):                           1395.42   
Median ITL (ms):                         1450.03   
P99 ITL (ms):                            1459.62   
Mean E2EL (ms):                          19447.04  
Median E2EL (ms):                        18308.11  
P99 E2EL (ms):                           33646.13  
Evaluating MMLU...
[nltk_data] Downloading package punkt to
[nltk_data]     /home/gxd_google_com/nltk_data...
[nltk_data]   Package punkt is already up-to-date!
[nltk_data] Downloading package punkt_tab to
[nltk_data]     /home/gxd_google_com/nltk_data...
[nltk_data]   Package punkt_tab is already up-to-date!

Results
{'accuracy': 0.862, 'gen_num': 1000}
```

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/mla_attention.py
+++ b/tpu_inference/layers/vllm/mla_attention.py
@@ -14,22 +14,24 @@
 import jax
 import torch
 import torchax
-from jax.sharding import PartitionSpec as P, NamedSharding
+from jax.sharding import NamedSharding
+from jax.sharding import PartitionSpec as P
 from torch.nn import Parameter
-from torchax.interop import torch_view, jax_view
+from torchax.interop import jax_view, torch_view
 from vllm.config import CacheConfig
-from vllm.model_executor.layers.attention.attention import get_attention_context
+from vllm.model_executor.layers.attention.attention import \
+    get_attention_context
 from vllm.model_executor.layers.attention.mla_attention import MLAAttention
 from vllm.model_executor.layers.linear import ColumnParallelLinear
-from vllm.model_executor.layers.mla import MLAModules, MultiHeadLatentAttentionWrapper
+from vllm.model_executor.layers.mla import (MLAModules,
+                                            MultiHeadLatentAttentionWrapper)
 from vllm.model_executor.layers.quantization import QuantizationConfig
 from vllm.v1.attention.backend import AttentionType
-from tpu_inference.layers.common.sharding import ShardingAxisName
 
 from tpu_inference import utils
-from tpu_inference.models.vllm.vllm_model_wrapper_context import (
-    get_vllm_model_wrapper_context,
-)
+from tpu_inference.layers.common.sharding import ShardingAxisName
+from tpu_inference.models.vllm.vllm_model_wrapper_context import \
+    get_vllm_model_wrapper_context
 
 
 class VllmTPUMLAAttention(MLAAttention):
@@ -65,8 +67,7 @@ class VllmTPUMLAAttention(MLAAttention):
         self.kv_cache_quantized_dtype = None
         if self.kv_cache_dtype != "auto":
             self.kv_cache_quantized_dtype = utils.get_jax_dtype_from_str_dtype(
-                self.kv_cache_dtype
-            )
+                self.kv_cache_dtype)
 
     def process_weights_after_loading(self, act_dtype: torch.dtype):
         with torchax.default_env():
@@ -79,11 +80,13 @@ class VllmTPUMLAAttention(MLAAttention):
             # Device_put `W_UK_T`, `W_UV` to TPUs
             mesh = self.kv_b_proj.quant_method.linear_config.mesh
             self.W_UK_T = torch_view(
-                jax.device_put(jax_view(self.W_UK_T), NamedSharding(mesh, P(ShardingAxisName.ATTN_HEAD,)))
-            )
+                jax.device_put(
+                    jax_view(self.W_UK_T),
+                    NamedSharding(mesh, P(ShardingAxisName.ATTN_HEAD, ))))
             self.W_UV = torch_view(
-                jax.device_put(jax_view(self.W_UV), NamedSharding(mesh, P(ShardingAxisName.ATTN_HEAD,)))
-            )
+                jax.device_put(
+                    jax_view(self.W_UV),
+                    NamedSharding(mesh, P(ShardingAxisName.ATTN_HEAD, ))))
 
             self.W_UK_T = Parameter(self.W_UK_T, requires_grad=False)
             self.W_UV = Parameter(self.W_UV, requires_grad=False)
--- a/tests/layers/vllm/test_mla_attention.py
+++ b/tests/layers/vllm/test_mla_attention.py
@@ -118,14 +118,16 @@ class TestVllmTPUMLAAttention:
             prefix="test",
         )
 
-        attn.W_UK_T = torchax.tensor.Tensor(jnp.ones((10, 10)), env=torchax.default_env())
-        attn.W_UV = torchax.tensor.Tensor(jnp.ones((10, 10)), env=torchax.default_env())
+        attn.W_UK_T = torchax.tensor.Tensor(jnp.ones((10, 10)),
+                                            env=torchax.default_env())
+        attn.W_UV = torchax.tensor.Tensor(jnp.ones((10, 10)),
+                                          env=torchax.default_env())
         attn.kv_b_proj = MagicMock()
         attn.kv_b_proj.quant_method.linear_config.mesh = mesh
         attn.kv_b_proj.named_parameters.return_value = {}
 
         with patch(
-            "vllm.model_executor.layers.attention.mla_attention.MLAAttention.process_weights_after_loading"
+                "vllm.model_executor.layers.attention.mla_attention.MLAAttention.process_weights_after_loading"
         ) as mock_super_process:
             with torchax.default_env():
                 attn.process_weights_after_loading(torch.float32)


tpu_inference/layers/vllm/backends/flash_attn_mla.py
```diff
@@ -123,7 +123,20 @@ def forward(self, q: torch.Tensor, kv_c_normed: torch.Tensor,
 
         # (B, N, P) x (N, P, L) -> (B, N, L)
         # torch nn param
-        q_nope = jnp.einsum("bnp,npl->bnl", q_nope, jax_view(layer.W_UK_T))
+        q_nope = jax.shard_map(
```
kyuyeunk: I think using lambda here might look a bit messy? also, seems like `q_nope` and `outputs` implementation can be shared? maybe implement one function that can be used for both `q_nope` and `outputs`?

tpu_inference/layers/vllm/mla_attention.py
```diff
@@ -63,14 +64,27 @@ def __init__(
...
             super().process_weights_after_loading(act_dtype)
 
             # NOTE: vLLM dequantizes kv_b_proj weights which causes more memory
             # usage than expected.
+
+            # TODO(gxd3): consider quantize W_UK_T, W_UV back to fp8.
+            # Explicit replicate `W_UK_T`, `W_UV` to all TPU, given MLA layers
+            # are pure DP.
+            mesh = self.kv_b_proj.quant_method.linear_config.mesh
+            self.W_UK_T = torch_view(
```
kyuyeunk, line 82: so this is fully replicating the weight? is this correct?

gxd3: Updated to use `P(ShardingAxisName.ATTN_HEAD,)` so that it look more general. But in typical sharding for MLA (DP for attn layer, EP for MOE layer), these 2 matrix are indeed replicated.

gxd3:
> can you slightly clean up the results section of pr description to make it more readable? maybe use markdown's code block feature?

Yes, updated.

tpu_inference/layers/vllm/backends/flash_attn_mla.py
gxd3: I extract the common part (jax.shard_map) into a shared util to be used by 2 call sites. I used lambda mostly to avoid naming for these small functions / operations :). PTAL, thank you!

tpu_inference/layers/vllm/backends/flash_attn_mla.py
```diff
@@ -121,9 +121,28 @@ def forward(self, q: torch.Tensor, kv_c_normed: torch.Tensor,
...
+                lambda sharded_x, sharded_w: fn(sharded_x, sharded_w),
+                mesh=mesh,
+                in_specs=(
+                    P(ShardingAxisName.ATTN_DATA,),
+                    P(ShardingAxisName.ATTN_HEAD,),
+                ),
+                out_specs=P(
+                    ShardingAxisName.ATTN_DATA,
+                ),
+                check_vma=False,
```
kyuyeunk: is this needed?

gxd3: yes, it's indeed needed,
i tried remove it, then the following error popped up
```
(EngineCore_DP0 pid=2582834)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/vllm/vllm/model_executor/layers/mla.py", line 173, in forward
(EngineCore_DP0 pid=2582834)     attn_out = self.mla_attn(
(EngineCore_DP0 pid=2582834)                ^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/vllm_env/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
(EngineCore_DP0 pid=2582834)     return self._call_impl(*args, **kwargs)
(EngineCore_DP0 pid=2582834)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/vllm_env/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
(EngineCore_DP0 pid=2582834)     return forward_call(*args, **kwargs)
(EngineCore_DP0 pid=2582834)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/tpu-inference/tpu_inference/layers/vllm/mla_attention.py", line 116, in forward
(EngineCore_DP0 pid=2582834)     outputs, new_kv_cache = self.impl.forward(q, kv_c_normed, k_pe,
(EngineCore_DP0 pid=2582834)                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/tpu-inference/tpu_inference/layers/vllm/backends/flash_attn_mla.py", line 140, in forward
(EngineCore_DP0 pid=2582834)     q_nope = sharded_fn(
(EngineCore_DP0 pid=2582834)              ^^^^^^^^^^^
(EngineCore_DP0 pid=2582834)   File "/home/gxd_google_com/work-dir/tpu-inference/tpu_inference/layers/vllm/backends/flash_attn_mla.py", line 126, in sharded_fn
(EngineCore_DP0 pid=2582834)     return jax.shard_map(
(EngineCore_DP0 pid=2582834)            ^^^^^^^^^^^^^^
(EngineCore_DP0 pid=2582834) ValueError: shard_map applied to the function '<lambda>' was given out_specs which require replication which can't be statically inferred given the mesh:
(EngineCore_DP0 pid=2582834) 
(EngineCore_DP0 pid=2582834) The mesh given has shape (1, 8, 1, 1, 1) with corresponding axis names ('data', 'attn_dp', 'attn_dp_expert', 'expert', 'model').
(EngineCore_DP0 pid=2582834) 
(EngineCore_DP0 pid=2582834) out_specs is PartitionSpec(('data', 'attn_dp', 'attn_dp_expert'),) which implies that the corresponding output value is only varying across mesh axes {data,attn_dp,attn_dp_expert} and not {expert,model}, but it was inferred to be possibly varying over {data,attn_dp,attn_dp_expert,expert,model}
(EngineCore_DP0 pid=2582834) 
(EngineCore_DP0 pid=2582834) Check if these output values are meant to be replicated over those mesh axes. If not, consider revising the corresponding out_specs entries. If so, consider disabling the check by passing the check_vma=False argument to `jax.shard_map`.
```

kyuyeunk: hmm do you know why? usually, that error makes sense when we are using kernel where xla cannot figure out sharding. but when using pure xla op, it's a bit weird that xla could not figure out the sharding - in most cases, it usually indicates that we are doing something very weird with sharding. also, any reason why we can't just do `with_sharding_constraint` & need to use `shard_map` instead?

gxd3: reading from this part of the msg:
```
ValueError: shard_map applied to the function '<lambda>' was given out_specs which require replication which can't be statically inferred given the mesh:
The mesh given has shape (1, 8, 1, 1, 1) with corresponding axis names ('data', 'attn_dp', 'attn_dp_expert', 'expert', 'model').
out_specs is PartitionSpec(('data', 'attn_dp', 'attn_dp_expert'),) which implies that the corresponding output value is only varying across mesh axes {data,attn_dp,attn_dp_expert} and not {expert,model}, but it was inferred to be possibly varying over {data,attn_dp,attn_dp_expert,expert,model}
Check if these output values are meant to be replicated over those mesh axes. If not, consider revising the corresponding out_specs entries. If so, consider disabling the check by passing the check_vma=False argument to `jax.shard_map`.
```
I think we should be `If so, consider disabling the check by passing the check_vma=False argument to jax.shard_map` case.


I also try revert the changes in this file, it also works as intended (no unintended AG of weights).
(So, only diffs in tpu_inference/layers/vllm/mla_attention.py is required :)) I revert the changes in this file to keep things simple.
updated xprof: http://xprof.corp.google.com/trace_viewer/gxd-10580911554035166288
(Originally I use jax.shard_map in this file because I wanted to make things more explicit. But since XLA can infer intended sharding correctly, i'll keep things concise and simple)

kyuyeunk: thanks! yeah my philosophy is that if compiler can handle it, let them handle it without being too explicit. because it makes the code much more easier to maintain. plus, compiler might introduce some optimization trick later down the line but we won't be able to leverage it because we made everything too explicit & strict for compiler to do anything with it.

gxd3:
> can you slightly clean up the results section of pr description to make it more readable? maybe use markdown's code block feature?

resolved.

PR #1861, opened 2026-03-05
Skip outer JIT for models that self-manage sharding

khatwanimohit:
# Description
                                                                                                                                                              
 ### Summary                                                                                                                                                                              
                                                                  
  Adds a bypass in _get_nnx_model for model classes that handle their own JIT-sharded initialization internally.

  Previously, all NNX models went through an outer @jax.jit-wrapped create_sharded_model() function. For models that already use out_shardings (or equivalent) during construction,
  this caused nested JIT inlining, which multiplies compilation overhead unnecessarily.

 ### Changes

  tpu_inference/models/common/model_loader.py: Added a new early-return path in _get_nnx_model for model classes decorated with _self_manages_sharding = True. These models are
  instantiated directly under the mesh context, skipping the outer @jax.jit wrapper, while still applying Qwix quantization and KV cache initialization.

  ### How it works

  If a model class sets the class attribute _self_manages_sharding = True, the loader will:
  1. Instantiate the model directly within the mesh context (no outer JIT)
  2. Apply Qwix quantization (apply_to_abstract_model=False)
  3. Call initialize_cache() if present

  All other models follow the existing path unchanged.

 ### Why

  Models that internally use create_nnx_model with out_shardings already emit the correct sharding constraints during construction. Wrapping them again in @jax.jit triggers redundant
  recompilation and is unnecessary overhead.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -166,6 +166,27 @@ def _get_nnx_model(
                                              use_qwix_on_abstract_model=True)
             return jit_model
 
+        if getattr(model_class, '_self_manages_sharding', False):
+            # `_self_manages_sharding` is a class-level boolean flag set to True
+            # by model classes (e.g. MaxText-backed models) that handle their own
+            # JIT-compiled, sharded weight initialization internally — typically by
+            # wrapping construction in jax.jit with explicit out_shardings. For
+            # these models, the standard path below (which wraps create_abstract_model
+            # + with_sharding_constraint in an outer @jax.jit) must be skipped:
+            # adding a second outer jit causes nested JIT inlining that re-traces
+            # and multiplies compilation time without benefit.
+            with mesh:
+                jit_model = model_class(vllm_config, rng, mesh)
+                jit_model = apply_qwix_quantization(
+                    vllm_config,
+                    jit_model,
+                    rng,
+                    mesh,
+                    apply_to_abstract_model=False)
+                if hasattr(jit_model, 'initialize_cache'):
+                    jit_model.initialize_cache()
+            return jit_model
+
         @jax.jit
         def create_sharded_model():
             model = create_abstract_model()


tpu_inference/models/common/model_loader.py
```diff
@@ -164,6 +164,22 @@ def create_jit_model(
                                              use_qwix_on_abstract_model=True)
             return jit_model
 
+        if getattr(model_class, '_self_manages_sharding', False):
```
wenxindongwork, line 169: @khatwanimohit thank you for finding the fix, is `_self_manages_sharding` MaxText specific, or a nnx notion? If it is maxtext specific, I wonder if there could be a more general way of implementing this logic.

khatwanimohit: this is maxtext specific, is it preferable that I add a attribute class in `tpu_inference/models/common/base.py` like this
 ```
class SelfShardingModel(abc.ABC):
      """Mixin for models that manage their own JIT-sharded initialization."""
      pass
```

and then use it like this `MaxTextForCausalLM(nnx.Module, SelfShardingModel)` and the loader checks `issubclass(model_class, SelfShardingModel)`

Comment (wenxindongwork) [reply tpu_inference/models/common/model_loader.py:169]: I'm brainstorming a more general API like this

https://github.com/vllm-project/tpu-inference/compare/oot-init-model 

where maxtext models can initialize like this: 

```
class MaxTextForCausalLM(JaxModule):
    @classmethod
    def create_model(cls, vllm_config, rng, mesh):
        from tpu_inference.models.common.model_loader import (
            jit_model, load_model_weights)
        
        with mesh:
            model = cls(vllm_config, rng, mesh)  # own sharding
            load_model_weights(model, vllm_config, rng)  # reuse standard loader
            # skip jit_model() if not needed, or call it:
            # model = jit_model(model, vllm_config, rng, mesh)
        return model
```

what do you think of this approach? I'm happy to check in the code for tpu-inference. how urgent do you need to have this support in?

khatwanimohit: omitting this jit helps us get around an OOM for Deepseek. Can this be checked in this week ? @wenxindongwork

NicoGrande: @wenxindongwork - I think the proposed API would not work given the different style checkpoints both model libraries are working with. Unless the API exposed is general to both huggingface and MaxText-style checkpoints

jrplatin: Can we please add a quick comment with some more info on what `_self_manages_sharding` is and where it's coming from? It's not obvious that this is for MaxText integration to me, thank you!

khatwanimohit: @jrplatin I added a comment

PR #1875, opened 2026-03-06
test upload unit test duration artifact in nightly build

ernie-chang:
# Description
This PR introduces dynamic test sharding for JAX unit tests to improve CI efficiency and balance. We now use pytest-split to distribute tests across parallel jobs based on their actual execution duration.

# Tests
CI / Unit tests.
generate .test_durations file: https://paste.googleplex.com/4730630665535488
unit test group1: https://paste.googleplex.com/4917425272389632

nightly build: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/13027
buildkite CI: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/16822

# Checklist


.buildkite/.test_durations
```diff
@@ -0,0 +1,1607 @@
+{
```
jcyang43, line 1: Let's not save this file in the project root. You can place this under the `.buildkite` directory. 

(The same applies to your commit script. Update where you're commit this file)

.buildkite/pipeline_jax.yml
```diff
@@ -97,7 +97,7 @@ steps:
...
+            .buildkite/scripts/run_in_docker.sh bash -c "pip install pytest-cov pytest-split; python3 -m pytest -s -v -x --splits 2 --durations-path .test_durations --group 1 --cov-config=/workspace/tpu_inference/.coveragerc --cov tpu_inference tests/ --ignore=/workspace/tpu_inference/tests/layers/vllm/test_compressed_tensors_moe.py --ignore=/workspace/tpu_inference/tests/kernels --ignore=/workspace/tpu_inference/tests/lora --ignore=/workspace/tpu_inference/tests/e2e --ignore=/workspace/tpu_inference/tpu_inference/mock --ignore=/workspace/tpu_inference/tests/models/jax/test_deepseek_v3.py --ignore=/workspace/tpu_inference/tests/layers/jax/moe/test_moe.py; if [ -f .coverage ]; then cp .coverage /tmp/hf_home/.coverage.part1.${TPU_VERSION:-tpu6e}; else exit 1; fi"
```
jcyang43: pip install should not be handled here

.buildkite/pipeline_jax.yml
```diff
@@ -109,7 +109,7 @@ steps:
...
+            .buildkite/scripts/run_in_docker.sh bash -c "pip install pytest-cov pytest-split; python3 -m pytest -s -v -x --splits 2 --durations-path .test_durations --group 2 --cov-config=/workspace/tpu_inference/.coveragerc --cov tpu_inference tests/ --ignore=/workspace/tpu_inference/tests/layers/vllm/test_compressed_tensors_moe.py --ignore=/workspace/tpu_inference/tests/kernels --ignore=/workspace/tpu_inference/tests/lora --ignore=/workspace/tpu_inference/tests/e2e --ignore=/workspace/tpu_inference/tpu_inference/mock --ignore=/workspace/tpu_inference/tests/models/jax/test_deepseek_v3.py --ignore=/workspace/tpu_inference/tests/layers/jax/moe/test_moe.py; if [ -f .coverage ]; then cp .coverage /tmp/hf_home/.coverage.part2.${TPU_VERSION:-tpu6e}; else exit 1; fi"
```
jcyang43: Ditto. pip install should not be handled here

jcyang43 (commented):
Please 
1. Set [this](https://github.com/vllm-project/tpu-inference/blob/a9c6227f5559b85275343633c28a924eecb86bc0/docker/Dockerfile.pypi#L19) to true 
2. Rebase on the latest main (& solve merge conflicts) 
3. Test again and re-generate the artifact

Thanks

.buildkite/nightly_verify.yml
```diff
@@ -89,3 +89,17 @@ steps:
       - GITHUB_PAT
      command:
        - bash .buildkite/scripts/commit_verified_commit_hashes.sh
+
+   - label: "Upload pytest duration artifact"
+     if: build.env("NIGHTLY") == "1"
```
jcyang43: And also if it's on main branch. Since only the main branch has access to the github PAT, there's no need to try to commit on non-main branches.

.buildkite/nightly_verify.yml
jcyang43: Ditto on only for main branch. See comment on L94

jcyang43, line 70: Ditto on only for main branch. See comment on L94

.buildkite/nightly_verify.yml
```diff
@@ -89,3 +89,17 @@ steps:
       - GITHUB_PAT
      command:
        - bash .buildkite/scripts/commit_verified_commit_hashes.sh
+
+   - label: "Upload pytest duration artifact"
+     if: build.env("NIGHTLY") == "1"     
+     key: unit_test_duration
```
jcyang43: Please rename to `update_unit_test_duration_file`

jcyang43 (approved):

.buildkite/nightly_verify.yml
jcyang43: this is still not resolved

jcyang43: this is still not resolved

.buildkite/nightly_verify.yml
jcyang43: this is still not resolved ...

jcyang43 (changes requested):

.buildkite/pipeline_jax.yml
```diff
@@ -120,7 +120,7 @@ steps:
...
+            .buildkite/scripts/run_in_docker.sh bash -c "python3 -m pytest -s -v -x --splits 4 --durations-path .buildkite/.test_durations --group 1 --cov-config=/workspace/tpu_inference/.coveragerc --cov tpu_inference tests/ --ignore=/workspace/tpu_inference/tests/kernels --ignore=/workspace/tpu_inference/tests/lora --ignore=/workspace/tpu_inference/tests/e2e --ignore=/workspace/tpu_inference/tpu_inference/mock --ignore=/workspace/tpu_inference/tests/models/jax/test_deepseek_v3.py; if [ -f .coverage ]; then cp .coverage /tmp/hf_home/.coverage.part1.${TPU_VERSION:-tpu6e}; else exit 1; fi"
```
QiliangCui: instead of copy it 4 times with changing number, i see BK has a feature for "repeating"... for example: 
https://github.com/vllm-project/tpu-inference/blob/cuiq-oot-perf-flakiness/.buildkite/pipeline_dev.yml#L37

then, other place can reference {{matrix}} like https://github.com/vllm-project/tpu-inference/blob/cuiq-oot-perf-flakiness/.buildkite/pipeline_dev.yml#L28. 

but i haven't tried if you can do this in the command line -- i guess it should work, too. have a try?

.buildkite/pipeline_jax.yml
```diff
@@ -136,14 +136,48 @@ steps:
...
+        soft_fail: true
+        artifact_paths: ".coverage.part4.${TPU_VERSION:-tpu6e}"
+        env:
+          USE_PREBUILT_IMAGE: "1"
+          TPU_VERSION: "${TPU_VERSION:-tpu6e}"
+        agents:
+          queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_queue}
+        commands:
+          - rm -f .coverage.part4.${TPU_VERSION:-tpu6e}
+          - |
```
QiliangCui, line 127: what if one added a new test? i think by default the new test is not in the `.test_durations`, right? then, where will it b run? does it default to the last group?

docker/Dockerfile.pypi
```diff
@@ -16,7 +16,7 @@ ARG VLLM_COMMIT_HASH=""
 
 FROM $BASE_IMAGE
 
-ARG IS_TEST="false"
```
QiliangCui, line 19: does this have anything to do with unit test update?

.buildkite/nightly_verify.yml
```diff
@@ -77,7 +78,7 @@ steps:
        - bash .buildkite/scripts/notify_test_results.sh
 
    - label: "Record verified commit hashes"
-     if: build.env("NIGHTLY") == "1"
+     if: build.branch == "main" && build.env("NIGHTLY") == "1"
```
QiliangCui, line 93: is this intentional change?

QiliangCui: oh, i see. people are running NIGHTLY in their own branch... today.

.buildkite/nightly_verify.yml
```diff
@@ -89,3 +90,17 @@ steps:
       - GITHUB_PAT
      command:
        - bash .buildkite/scripts/commit_verified_commit_hashes.sh
+
+   - label: "Upload pytest duration artifact"
+     if: build.branch == "main" && build.env("NIGHTLY") == "1"     
+     key: update_unit_test_duration_file
+     artifact_paths: /mnt/disks/persist/models/.test_durations
+     agents:
+       queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_queue}
```
QiliangCui, line 111: can this be done in CPU machine?

.buildkite/scripts/commit_test_duration.sh
```diff
@@ -0,0 +1,65 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+# --- Configuration ---
+REPO_URL="https://github.com/vllm-project/tpu-inference.git"
+TARGET_BRANCH="main"
```
QiliangCui, line 21: add a comment saying this must happen in main.

.buildkite/scripts/commit_test_duration.sh
```diff
@@ -0,0 +1,65 @@
...
+# --- Check for changes before committing ---
+if git diff --quiet --cached; then
+  echo "No changes to commit. Exiting successfully."
+  exit 0
+else
+  echo "--- Committing changes"
+  git commit -s -m "${COMMIT_MESSAGE}"
+
+  echo "--- Pushing changes to '${TARGET_BRANCH}'"
+  git push "${AUTHENTICATED_REPO_URL}" "HEAD:${TARGET_BRANCH}"
```
QiliangCui, line 65: does it log the AUTHENTICATED_REPO_URL?

QiliangCui: Could you search what is the standard way of doing this safely?

.buildkite/nightly_verify.yml
```diff
@@ -89,3 +90,17 @@ steps:
...
+     secrets:
+       - GITHUB_PAT
+     commands:
+       - |
+         .buildkite/scripts/run_in_docker.sh bash -c "python3 -m pytest -s -v -x --store-durations --cov tpu_inference tests/ --ignore=/workspace/tpu_inference/tests/kernels --ignore=/workspace/tpu_inference/tests/lora --ignore=/workspace/tpu_inference/tests/e2e --ignore=/workspace/tpu_inference/tpu_inference/mock --ignore=/workspace/tpu_inference/tests/models/jax/test_deepseek_v3.py; if [ -f .test_durations ]; then cp .test_durations /tmp/hf_home/.test_durations; else exit 1; fi"
+       - cp /mnt/disks/persist/models/.test_durations .buildkite/
+       - bash .buildkite/scripts/commit_test_duration.sh
```
QiliangCui, line 118: i think we have 3 "nightly". 

default
model=vllm
model=flax_xxx

will there be race condition? i think we actually expects the "default" night. right?

.buildkite/pipeline_jax.yml
```diff
@@ -158,7 +192,7 @@ steps:
             echo "    ." >> .coveragerc
             echo "    /workspace/tpu_inference/" >> .coveragerc
           - |
-            python3 ./coverage.pex combine .coverage.part1.${TPU_VERSION:-tpu6e} .coverage.part2.${TPU_VERSION:-tpu6e}
+            python3 ./coverage.pex combine .coverage.part1.${TPU_VERSION:-tpu6e} .coverage.part2.${TPU_VERSION:-tpu6e} .coverage.part3.${TPU_VERSION:-tpu6e} .coverage.part4.${TPU_VERSION:-tpu6e}
```
QiliangCui: does it work if `python3 ./coverage.pex combine .coverage.part*.${TPU_VERSION:-tpu6e} `?

QiliangCui (changes requested):

.buildkite/pipeline_jax.yml
jcyang43: Yes, when a new test that hasn't been seen in the durations file is added, it will be run. 

What happens is that pytest will assign a default runtime for the new test case (which will be the average runtime of a test case in the current durations file based on their docs). So the splits may not be 100% even, but the durations file will get updated every night by the nightly build to reflect the latest state.

.buildkite/nightly_verify.yml
jcyang43: Yes, we only want to do this for the default. Thanks for the catch Qiliang

docker/Dockerfile.pypi
jcyang43: Nope. But it's a needed change when he moved explicit test dependencies installation from Dockerfile to `requirements_test.txt`

.buildkite/scripts/commit_test_duration.sh
jcyang43: No, it does not log `AUTHENTICATED_REPO_URL`. You can refer to this buildkite [link](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/16034#019dd809-d4f0-4115-8a8a-4bde31e844b1) to see an example of how the logs will look.

---
I looked into this topic when I first joined the team. TLDR the current mechanism is safe: 
- It's configured as a secret on buildkite and only the main branch can read it (only Buildkite admins can modify this setting) 
- If the team accidentally merges some code into the main branch that tries to print out this secret, Buildkite has built-in mechanism to not print out the secret. Can refer to their official doc [here](https://buildkite.com/docs/pipelines/security/secrets/buildkite-secrets#redaction) for details

.buildkite/pipeline_jax.yml
ernie-chang: It can work by using {{matrix}}, thanks for the suggestion.

.buildkite/nightly_verify.yml
ernie-chang: we use tpu agent because of the CPU machine need to additional installation for docker

.buildkite/scripts/commit_test_duration.sh
ernie-chang: Finish the comment for it, thanks for the remind

.buildkite/pipeline_jax.yml
ernie-chang: Yes, it can work by using part* instead of using part1-4, thanks for the suggestion

.buildkite/nightly_verify.yml
ernie-chang: modify the yaml to only the main branch has access to the github PAT

.buildkite/nightly_verify.yml
ernie-chang: modify the yaml to only the main branch has access to the github PAT

ernie-chang: modify the yaml to only the main branch has access to the github PAT

PR #1877, opened 2026-03-06
Format support matrix statuses natively in buildkite shell scripts [Phase 1]

RobMulla:
This PR refactors the support matrix generation scripts to natively output UI-formatted statuses (`✅ Passing`, `⚪ N/A`, `❓ Untested`, `❌ Failing`) instead of relying on downstream Python scripts to parse emojis.

## Changes
1. **`record_step_result.sh`**: Modifies the `case  in` block to output the actual user-facing strings instead of just emojis/raw words. Because this step outputs the data directly into the Buildkite metadata pipeline, every downstream script will natively pull the correct, formatted status without any post-processing required.
2. **`generate_support_matrices.sh`**: 
   - Modifies `process_kernel_matrix_to_pivot` header variables to output `(Corr)` and `(Perf)` instead of `(Correctness)` and `(Performance)` to save space.
   - Updates the `buildkite-agent meta-data get` fallbacks from `N/A` to `⚪ N/A` to match the UI scheme consistently.
3. **Tests**: Added `tests/test_generate_support_matrices.sh` with a mock `buildkite-agent` to guarantee the awk pivoting algorithm and fallback UI strings are correctly formatted before merging.

This sets up the foundation for stripping out the redundant `update_all_matrices.py` and legacy `combine_matrices.py` scripts from the downstream logic.


tests/test_generate_support_matrices.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
```
jcyang43, line 1: Hi @RobMulla, is this test actually passing for you btw? 

Image

RobMulla: I think this is because you are running on mac - newer bash syntax crashes on Mac's native bash. I updated `generate_support_matrices` so it should work now.

jcyang43: I initially suspected it was a mac issue but tested on linux & got the same error XD 
image

But the test is passing with your new changes now, thanks.

jcyang43 (approved):

RobMulla:
Merging re discussion with @jcyang43 @bvrockwell and @ica-chao

PR #1891, opened 2026-03-10
DeepSeek routing fixes on JAX path.

gpolovets1:
# Description
Adding the following fixes to improve accuracy on the Jax path.
  - Applying router activations just once (previously applied an extra time in fused_moe_func).
  - Applying routed_scaling_factor 
  - Using float32 precision during topk selection ([vLLM](https://github.com/vllm-project/vllm/blob/e89a91d9275cd8ac086fe04476b41675a9ebbd5c/vllm/model_executor/layers/fused_moe/cpu_fused_moe.py#L59) reference implementation is also doing this)

# Tests

Locally tested that MMLU went up from 67 to 80 while maintaining perf to <1%.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -303,8 +303,6 @@ def fused_moe_func(
     use_ep: bool,
     activation: str,
     scoring_fn: str,
-    topk_weights: jax.Array | None = None,
-    topk_indices: jax.Array | None = None,
 ) -> jax.Array:
     """Route tokens in hidden_states into each experts based on routing.
 
@@ -323,8 +321,6 @@ def fused_moe_func(
         use_ep: use expert parallelism.
         activation: activation function to perform on the output of w1.
         scoring_fn: scoring function to apply on gating_output.
-        topk_weights: pre-calculated expert weights.
-        topk_indices: pre-calculated expert indices.
 
     Returns:
         Output of moe operation [num_tokens, hidden_size]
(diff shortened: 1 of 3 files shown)


tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -855,9 +856,10 @@ def __init__(self,
             routed_scaling_factor=routed_scaling_factor,
             dtype=dtype,
             moe_backend=moe_backend,
-            activation_ffw_td=P(ShardingAxisName.MLP_DATA, None),
-            ed_sharding=P(None, None),
-            e_sharding=P(None, ),
+            activation_ffw_td=(ShardingAxisName.MLP_DATA, None),
+            ed_sharding=(None, None),
+            e_sharding=(None, ),
```
lk-chen: I guess this line is unnecessary, and you can revert line 1078?

tpu_inference/models/jax/deepseek_v3.py
```diff
@@ -1065,26 +1074,32 @@ def __call__(self, x_TD: Float) -> Tuple[Float, Float]:
                 - indices: Indices of selected experts, shape (sequence, num_experts_per_tok).
         """
         x_TD = jnp.asarray(x_TD, self.dtype)
-        x_TD = lax.with_sharding_constraint(x_TD, self.activation_ffw_td)
+        x_TD = jax.lax.with_sharding_constraint(x_TD,
+                                                P(*self.activation_ffw_td))
+
+        logits_TE = super().__call__(x_TD).astype(jnp.float32)
```
lk-chen, line 1082: Could you add comment with reference of https://github.com/vllm-project/vllm/blob/e89a91d9275cd8ac086fe04476b41675a9ebbd5c/vllm/model_executor/layers/fused_moe/cpu_fused_moe.py#L59 here?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -303,6 +303,8 @@ def fused_moe_func(
     use_ep: bool,
     activation: str,
     scoring_fn: str,
+    topk_weights: jax.Array | None = None,
```
kyuyeunk: not sure if this change is necessary? wouldn't it be possible to make change `DeepSeekV3Router` to return a value that `fused_moe_func` expects?

gpolovets1: fused_moe_func applies global top-k routing whereas DeepSeek needs to use the custom grouped top-k routing.

gpolovets1: Would you prefer if I passed get_topk_func as an argument? It could replace jax.lax.top_k if passed and I think retain the same logical flow.

kyuyeunk: so vllm has a concept called monolithic vs. non-monolithic. and i was wondering if we can leverage this api: https://github.com/vllm-project/vllm/blob/bdd8981dab8d8c6ae88a3f605d04ec5243088e5a/vllm/model_executor/layers/fused_moe/runner/default_moe_runner.py#L505-L525

tpu_inference/models/jax/deepseek_v3.py
gpolovets1: Can you explain what you mean by reverting line 1078?

lk-chen: I meant you don't need to change `P(None, )` to tuple `(None, )` here and in line 1078 convert tuple to PartitionSpec ? `e_sharding` is defined as type of PartitionSpec on line 989

gpolovets1: Good catch, I must have done this a while back and it is unnecessary now. Thanks!

tpu_inference/layers/common/fused_moe_gmm.py
gpolovets1: Hi @kyuyeunk per offline discussion, I am splitting enablement of node-limited routing from this PR since it doesn't seem to affect perf/accuracy. Will follow up if we see evidence of any benefit. Thanks!

PR #1898, opened 2026-03-10
Add subchannel quantization support to GMM V2 and fully deprecate GMM V1

wenxindongwork:
# Description

This PR adds support for subchannel quantization for fp4 and fp8 in the `gmm_v2` kernel.  

This PR also fully replace GMM V1 with GMM v2. 

GPT-OSS 120b TP=2 GMM up

- Before: 511us
- After: 239us

GMM down

- Before: 249us
- After:158us

# Tests

Verified MMLU
Added more tests to `gmm_test.py`

# Checklist

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -753,8 +753,8 @@ def calculate_tiling(
     # to fit rhs into vmem by only adjusting tile_n.
 
     # Decrease tile_n until rhs fits in vmem target.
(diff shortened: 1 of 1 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
wenxindongwork: @kyuyeunk why is `>=` used here?

kyuyeunk: oh good point. just `>` should be more correct.

PR #1904, opened 2026-03-11
Fix: Align Qwen3 dtype handling with Qwen2 for TPU

xianglon-commits:
This PR addresses a `ValueError` caused by a dtype mismatch (bfloat16 vs float32) within the TPU attention kernels when running the Qwen3 model. This issue was encountered when using Tunix with the vLLM-JAX backend on the `lance-ds` branch.

Expected `kv_cache.dtype=dtype(bfloat16)` to be equal to `k.dtype=dtype(bfloat16)` and `v.dtype=dtype(bfloat16)`, but found `v.dtype=dtype('float32')`.

Following the pattern used to fix similar issues in `qwen2.py` (as suggested by the Tunix team, see https://screenshot.googleplex.com/8AwrZ44yv57Bt95), this change renames the `dtype` parameter to `param_dtype` within the `__init__` methods of layers in `tpu_inference/models/jax/qwen3.py`.

This change aims to prevent potential naming conflicts and ensure the intended `bfloat16` data type is consistently propagated and used for model parameters on TPU, resolving the mixed-precision error.

### DIFF
--- a/tpu_inference/models/jax/qwen3.py
+++ b/tpu_inference/models/jax/qwen3.py
@@ -78,7 +78,7 @@ class Qwen3Attention(JaxModule):
         self.q_proj = JaxEinsum(
             "TD,DNH->TNH",
             (self.hidden_size, self.num_heads, self.head_dim),
-            dtype=dtype,
+            param_dtype=dtype,
             kernel_init=nnx.with_partitioning(init_fn, (None, "model", None)),
             rngs=rng,
             quant_config=quant_config,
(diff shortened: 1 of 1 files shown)


kyuyeunk:
would this be related to this pr? https://github.com/vllm-project/tpu-inference/pull/1771

kyuyeunk:
i believe this issue has been resolved by https://github.com/vllm-project/tpu-inference/pull/1899

PR #1913, opened 2026-03-12
RL Integration support matrix

wenxindongwork:
# Description

Setting up the support matrix for RL integration. 

Currently included tests: 
- Deleting & reinitializing kv cache (for reducing HBM pressure duiring weight sync)
- Log prob computation correctness and performance
- Prefix caching performance 

Future APIs & tests to include: 
- Fetch & Update models weights (for weight sync)
- Return MoE expert ids 

# Tests 

Added `test_rl_integration.py`

# Checklist

### DIFF
--- a/support_matrices/feature_support_matrix.csv
+++ b/support_matrices/feature_support_matrix.csv
@@ -13,6 +13,5 @@ Feature,CorrectnessTest,PerformanceTest
 "async scheduler",✅,✅
 "data_parallelism",✅,unverified
 "runai_model_streamer_loader",✅,N/A
-"RL_Integration",unverified,unverified
 "sampling_params",✅,N/A
 "structured_decoding",✅,N/A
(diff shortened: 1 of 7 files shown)


tests/e2e/test_rl_integration.py
```diff
@@ -0,0 +1,416 @@
...
+        t_lp = self._measure_latency(llm, test_prompts, sp_lp)
+
+        overhead = (t_lp - t_no_lp) / t_no_lp if t_no_lp > 0 else 0
+
+        print("✓ Logprobs latency test results:")
+        print(f"  Without logprobs: {t_no_lp:.3f}s")
+        print(f"  With logprobs:    {t_lp:.3f}s")
+        print(f"  Overhead:         {overhead:.1%}")
+
+        assert overhead < 1.00, (
```
QiliangCui, line 328: do we really want to allow 100% overhead?

jcyang43 (approved):
Thanks WenXin! I think the current setup will populate the results into the feature support matrix, but we'll use a separate PR later to add the logic for automatically populate a new RL matrix. Thanks

QiliangCui (approved):

tests/e2e/test_rl_integration.py
wenxindongwork: great point and nope! Sierra is implementing an optimization right now for improving log prob performance, we will update this number once that PR is in.

PR #1918, opened 2026-03-13
fix invalid compilation by adding dummy variable

sixiang-google:
# Description

Fix performance drop caused by https://github.com/vllm-project/tpu-inference/pull/1895

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -639,6 +639,16 @@ class CompilationManager:
 
                 for do_sampling in (False, True):
                     draft_probs = None
+                    # Use a dummy tensor with a unique shape for each logprobs config.
+                    # Currently logprobs=False for rejection_sampler.
+                    logprobs_dummy = False
+                    dummy_shape = (1 if logprobs_dummy else 2, )
+                    _cache_collision_dummy = jnp.zeros(dummy_shape,
+                                                       dtype=jnp.int32)
+                    _cache_collision_dummy = jax.device_put(
+                        _cache_collision_dummy,
+                        NamedSharding(self.runner.mesh, PartitionSpec(None)))
+
                     if do_sampling:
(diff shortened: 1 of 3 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -639,6 +639,16 @@ def _precompile_rejection_sampler(self) -> None:
 
                 for do_sampling in (False, True):
                     draft_probs = None
+                    # Use a dummy tensor with a unique shape for each logprobs config.
+                    # Currently logprobs=False for rejection_sampler.
+                    logprobs_dummy = False
```
QiliangCui, line 644: Will we support "true" in the future? If so,  consider either adding a `TODO: update when logprobs is supported in rejection_sampler`?

tpu_inference/runner/compilation_manager.py
```diff
@@ -639,6 +639,16 @@ def _precompile_rejection_sampler(self) -> None:
...
+                    dummy_shape = (1 if logprobs_dummy else 2, )
+                    _cache_collision_dummy = jnp.zeros(dummy_shape,
+                                                       dtype=jnp.int32)
+                    _cache_collision_dummy = jax.device_put(
+                        _cache_collision_dummy,
```
QiliangCui, line 649: I see in the other file, you're using 
device_array for collision_dummy. 

here, for the _cache_collsion_dummy, you're using device_put. 

does this matter?

tpu_inference/runner/compilation_manager.py
sixiang-google: In compiling rejection_sampler the logprobs is not required

tpu_inference/runner/compilation_manager.py
sixiang-google: This is a small dummy tensor and I don't see a significant performance gap in small models by adding it

PR #1929, opened 2026-03-13
[Kernel] [GMM] Correct truncation bias during LHS quantization

catswe:
# Description

For integer targets, the current code truncates towards zero. This systematically underestimates the magnitude of quantized values. jnp.round() applies round-to-nearest-even before the cast, resulting in better quantization.

# Tests

commands to reproduce.

# Checklist


kyuyeunk:
I don't think rounding operation is free. Do you know how much quality vs. performance trade off exits?

PR #1934, opened 2026-03-15
[Kernel] [GMM] Fix incorrect RHS scale application

catswe:
# Description

For the quantized matmul path, when `q_block_size` (`lhs_cfgs.quant_block_size`) > `rhs_qbs` (`rhs_cfgs.quant_block_size`), it applies only one RHS scale per LHS block instead of one per RHS block, which is incorrect, as one LHS block spans multiple RHS blocks.

https://github.com/vllm-project/tpu-inference/blob/1acbeab4c3685a55be447f8da99b4e29f523c45b/tpu_inference/kernels/megablox/gmm_v2.py#L331-L335

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -932,6 +932,19 @@ def make_gmm_configs(
         dtype=lhs.dtype,
     )
 
+    # NOTE(catswe): when lhs quant_block_size > rhs quant_block_size,
+    # a single lhs block spans multiple rhs quant blocks. the quantized
+    # matmul path iterates at lhs block granularity and applies one rhs
+    # scale per lhs block, so the additional rhs scales within each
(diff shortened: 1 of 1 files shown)


kyuyeunk:
thanks for identifying this issue @catswe! @wenxindongwork, since you've recently worked on subchannel weight support in https://github.com/vllm-project/tpu-inference/pull/1898, can you take a look?

tpu_inference/kernels/megablox/gmm_v2.py
kyuyeunk, line 1: can you add a unit test where it triggers fails if this code does not exist?

catswe: i added a `NotImplementedError` guard since i think that's more appropriate. let me know if you want the unit test still!

PR #1940, opened 2026-03-17
Remove block_until_ready from the transfer_and_register_cpu_chunk code in TPU Host offload connector

saikat-royc:
# Description

the past.

   Eliminates the second synchronous `block_until_ready()` call in the KV
    offload pipeline, enabling fully non-blocking Device-to-Host (D2H)
    transfers. This change decouples the initiation of the data transfer from the
    resource reclamation, ensuring the background thread pool is never
    stalled and the main model loop remains free of XLA synchronization
    bottlenecks.

    Key changes:
    - Asynchronous Dispatch: Modified `_transfer_and_register_cpu_chunks` to
      return a list of JAX futures (unready arrays) and removed the
      `jax.block_until_ready()` stall.
    - Soft Budgeting Polling: Updated `_process_completed_saves` in the main
      feedback loop to implement a non-blocking two-stage completion check:
      1. Python Readiness: Verifies the background thread finished
         dispatching (`future.done()`).
      2. Jax Array Readiness: Performs a non-blocking hardware check
         (`array.is_ready()`) to confirm the TPU finished moving the bytes.
    - Safe Resource Lifecycle: The signal to the Scheduler to "Free" staging
      buffer slots is now strictly gatekept by cpu chunks readiness.
    - Unit test for async save + is_ready() check for the jax array stored
      in cpu backend

FIXES: #123456

# Tests

commands to reproduce.

Manual runs of the kv host offload unit tests in v7 machine

# Checklist

### DIFF
--- a/tpu_inference/offload/tpu_offload_connector.py
+++ b/tpu_inference/offload/tpu_offload_connector.py
@@ -2169,11 +2169,12 @@ class TPUOffloadConnectorWorker:
 
         return gathered_kv_caches_tpu, manifest, total_num_blocks_to_save
 
-    def _transfer_and_register_cpu_chunks(self,
-                                          flat_kv_caches_tpu: Any,
-                                          total_num_blocks_to_save: int,
-                                          manifest: list[SaveReqInfo],
-                                          is_batched: bool = False):
+    def _transfer_and_register_cpu_chunks(
+            self,
+            flat_kv_caches_tpu: Any,
+            total_num_blocks_to_save: int,
+            manifest: list[SaveReqInfo],
+            is_batched: bool = False) -> list[jax.Array]:
         """
         Asynchronously transfers KV blocks from TPU to CPU, unstitches them,
         and registers them with the CPU RAM backend store.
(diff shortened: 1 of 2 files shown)


tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -2443,7 +2449,7 @@ def _async_transfer_task(req_id, *args):
                                                    num_blocks_to_save, [info],
                                                    False)
 
-                self._pending_save_futures.append((future, [info]))
+                self._pending_save_futures.append((future, [info], None))
```
juncgu-google, line 2452: do we still need to put  `_async_transfer_task` in a thread for execution? Since the copy is non-blocking, we can directly call the fuction and append the new future (a tuple of unready_chunks and info) into the `self._pending_save_futures`

saikat-royc: addressed in https://github.com/vllm-project/tpu-inference/pull/2102

saikat-royc:
please review https://github.com/vllm-project/tpu-inference/pull/2102

PR #1948, opened 2026-03-18
Replace deepcopy with copy.

wang2yn84:
# Description

This operations takes 3ms for us, with copy() and it's reduced to <0.4ms. The deepcopy is more expensive than copy. This dictionary is {str, int}, therefore copy is enough.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 import functools
 import logging
 import random
@@ -995,7 +994,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
             # Return Model output to executor
             model_runner_output = ModelRunnerOutput(
                 req_ids=req_ids,
-                req_id_to_index=copy.copy(self.input_batch.req_id_to_index),
+                req_id_to_index=self.input_batch.req_id_to_index.copy(),
                 sampled_token_ids=[],  # Fill in async get
                 logprobs=logprobs_lists,
                 prompt_logprobs_dict=prompt_logprobs_dict,


tpu_inference/runner/tpu_runner.py
```diff
@@ -834,7 +834,7 @@ def _execute_model(
 
                 return ModelRunnerOutput(
                     req_ids=self.input_batch.req_ids,
-                    req_id_to_index=self.input_batch.req_id_to_index,
+                    req_id_to_index=self.input_batch.req_id_to_index.copy(),
```
wenxindongwork: why is the `copy()` needed here?

tpu_inference/runner/tpu_runner.py
```diff
@@ -1063,7 +1063,7 @@ def _sample_from_logits(
 
         model_runner_output = ModelRunnerOutput(
             req_ids=req_ids,
-            req_id_to_index=self.input_batch.req_id_to_index,
+            req_id_to_index=self.input_batch.req_id_to_index.copy(),
```
wenxindongwork: and similarly, why is the copy needed here?

tpu_inference/runner/tpu_runner.py
kyuyeunk: quote from @yuyanpeng-google:

> It's for async scheduler. Since the next step would change the input_batch, it need deepcopy to prevent race condition. When the previous result is ready to get, it need the req_id_to_index(of the previous batch)

yuyanpeng-google: I think this one is not for the AsyncTPUModelRunnerOutput, and we don't need the copy here.

tpu_inference/runner/tpu_runner.py
yuyanpeng-google: I thought it's not necessary for not AsyncTPUModelRunnerOutput.

tpu_inference/runner/tpu_runner.py
wang2yn84: Thank you for the comments, thought it's also used for async before. Removed!

tpu_inference/runner/tpu_runner.py
wang2yn84: Also removed.

wenxindongwork (approved):

Issue #1501, opened 2026-01-21
# [RFC] TPU inference model loaders behind a common runtime interface


qihqi:
## Summary
- `get_flax_model` and `get_vllm_model` currently return 7-tuples of functions and metadata with slight differences, making usage error-prone and hard to extend.
- Proposal: return a `LoadedModel` interface and subclasses that exposes common methods instead of a tuple of callables.
- Call sites will invoke methods on the returned object; optional capabilities are surfaced as optional methods.
- Thought process: a tuple of functions is worse than a NamedTuple of functions in readability. A NamedTuple of functions & states is equivalent of an object with methods.

## Background
- In `tpu_inference/tpu_inference/models/common/model_loader.py`:
  - `get_flax_model(...) -> (model_fn, compute_logits_fn, combine_hidden_states_fn, multimodal_fns, state, lora_manager, model)`
  - `get_vllm_model(...) -> (jit_model, compute_logits_fn, combine_hidden_states_fn=None, multimodal_fns=None, params, lora_manager, model)`
- A TODO notes the returned functions “do not work with all models and this is not very easy to see from the code.”
- Tuple positions differ between flax/vLLM and require intimate knowledge to use safely.

## Goals
- Establish a single, documented runtime interface for both flax_nnx (JAX) and vLLM (PyTorch) loaders.
- Make optional capabilities explicit and discoverable.
- Simplify downstream usage and reduce positional unpacking errors.
- Enable future extensions without breaking callers.

## Proposed interface
Introduce a lightweight runtime class with explicit method signatures (no `*args`, `**kwargs`). Optional capabilities raise `NotImplementedError` unless implemented.

```python
from typing import Any, List, Optional, Sequence, Tuple
import jax

class LoadedModel:
    def run(
        self,
        state: Any,
        kv_caches: List[jax.Array],
        input_ids: Optional[jax.Array],
        attn_metadata: Any,  # AttentionMetadata
        input_embeds: Optional[jax.Array],
        input_positions: jax.Array,
        layer_name_to_kvcache_index: Sequence[Tuple[str, int]],
        lora_metadata: Optional[dict[str, Any]],
        intermediate_tensors: Optional[Any] = None,  # JaxIntermediateTensors
        is_first_rank: bool = True,
        is_last_rank: bool = True,
    ) -> Tuple[List[jax.Array], jax.Array, list[Any]]:
[...]

PR #1952, opened 2026-03-18
[Multi-Modality] support vllm model wrapper mm model

yuyanpeng-google:
# Description

1. add vllm wrapper for multimodal.
2. modify interface of embed_input_ids with related jax model.
3. modify gather_mm_embeddings to get is_mm_embed for the new interface.
4. register function for torch.sdpa through torchax to use flash attention

# Tests

test with vllm model with mm support.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/attention_interface.py
+++ b/tpu_inference/layers/common/attention_interface.py
@@ -66,26 +66,51 @@ def sharded_flash_attention(
     causal: bool = True,
     sm_scale: Optional[float] = None,
     vmem_limit_bytes: int | None = None,
+    use_attention_bias: bool = False,
 ) -> Callable[..., Any]:
-    in_specs = (
-        P("data", "model", None, None),  # q
-        P("data", "model", None, None),  # k
-        P("data", "model", None, None),  # v
-        P(),  # segment_ids
-    )
-    out_specs = P("data", "model", None, None)
+    if use_attention_bias:
+        in_specs = (
+            P("data", "model", None, None),  # q
+            P("data", "model", None, None),  # k
+            P("data", "model", None, None),  # v
+            P("data", "model", None, None),  # attention_bias
+            P(),  # segment_ids
+        )
+        out_specs = P("data", "model", None, None)
 
-    def _flash_attention(q, k, v, segment_ids):
-        return flash_attention(q,
-                               k,
-                               v,
-                               segment_ids=segment_ids,
-                               sm_scale=sm_scale,
-                               causal=causal,
-                               vmem_limit_bytes=vmem_limit_bytes)
+        def _flash_attention_use_ab(q, k, v, attention_bias, segment_ids):
+            return flash_attention(q,
+                                   k,
+                                   v,
+                                   ab=attention_bias,
+                                   segment_ids=segment_ids,
+                                   sm_scale=sm_scale,
+                                   causal=causal,
+                                   vmem_limit_bytes=vmem_limit_bytes)
+
+        attn_fn = _flash_attention_use_ab
+    else:
+        in_specs = (
+            P("data", "model", None, None),  # q
+            P("data", "model", None, None),  # k
+            P("data", "model", None, None),  # v
+            P(),  # segment_ids
+        )
+        out_specs = P("data", "model", None, None)
+
+        def _flash_attention(q, k, v, segment_ids):
+            return flash_attention(q,
+                                   k,
+                                   v,
+                                   segment_ids=segment_ids,
+                                   sm_scale=sm_scale,
+                                   causal=causal,
+                                   vmem_limit_bytes=vmem_limit_bytes)
+
+        attn_fn = _flash_attention
 
     return jax.jit(
-        jax.shard_map(_flash_attention,
+        jax.shard_map(attn_fn,
                       mesh=mesh,
                       in_specs=in_specs,
                       out_specs=out_specs,
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -369,9 +369,21 @@ def get_vllm_model(
     jit_model = model.jit_step_func()
     compute_logits_fn = model.jit_compute_logits_func()
     pooler_fn = model.build_pooler_func()
+
+    multimodal_fns = {
+        "precompile_vision_encoder_fn":
+        getattr(model.model.vllm_model, "precompile_vision_encoder", None),
+        "embed_multimodal_fn":
+        model.wrap_embed_multimodal_func(),
+        "embed_input_ids_fn":
+        model.wrap_embed_input_ids_func(),
+        "get_mrope_input_positions_fn":
+        getattr(model.model.vllm_model, "get_mrope_input_positions", None),
+    }
+
     # the model needs to be returned because lora weights are neither torch.nn.parameter nor torch.nn.buffer. After we load the lora weights and set it to the torch.nn.Module, we can shard it and move it to TPU.
     combine_hidden_states_fn = None
-    return jit_model, compute_logits_fn, pooler_fn, combine_hidden_states_fn, None, params, lora_manager, model
+    return jit_model, compute_logits_fn, pooler_fn, combine_hidden_states_fn, multimodal_fns, params, lora_manager, model
 
 
 def get_model(
--- a/tpu_inference/models/jax/qwen2_5_vl.py
+++ b/tpu_inference/models/jax/qwen2_5_vl.py
@@ -1066,8 +1066,14 @@ class Qwen2_5_VLForConditionalGeneration(nnx.Module):
         return multimodal_embeddings
 
     def embed_input_ids(
-            self, input_ids: jax.Array,
-            multimodal_embeddings: Optional[jax.Array]) -> jax.Array:
+        self,
+        input_ids: jax.Array,
+        multimodal_embeddings: jax.Array | None,
+        *,
+        is_multimodal: jax.Array | None = None,
+    ) -> jax.Array:
+
+        del is_multimodal
 
         if not self.is_first_rank:
             return None
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import copy
+import functools
 import time
 from collections.abc import Sequence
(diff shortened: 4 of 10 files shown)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -287,6 +369,88 @@ def step_fun(
 
         return step_fun
 
+    def wrap_embed_multimodal_func(self):
+        if not hasattr(self.model.vllm_model, "embed_multimodal"):
```
kyuyeunk: not really a fan of string based `hasattr` checking. i feel like there should be a better way to check if a model is multimodal or not?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -287,6 +369,88 @@ def step_fun(
...
+            return None
+
+        # The function cannot be JITted directly due to its dynamic implementation
```
kyuyeunk, line 326: can you elaborate more on this part?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -112,10 +119,85 @@ def __init__(self, vllm_config: VllmConfig, rng: PRNGKey, mesh: Mesh):
...
+            torch.nn.functional.scaled_dot_product_attention,
+            is_jax_function=True,
+            needs_env=False,
+        )
+        @register_function(
+            torch.ops.aten.scaled_dot_product_attention,
+            is_jax_function=True,
+            needs_env=False,
+        )
+        def patched_sdpa(
```
kyuyeunk: what is `patched_sdpa` and does it belong in `vllm_model_wrapper`?

tpu_inference/models/common/model_loader.py
```diff
@@ -369,9 +369,17 @@ def get_vllm_model(
     jit_model = model.jit_step_func()
     compute_logits_fn = model.jit_compute_logits_func()
     pooler_fn = model.build_pooler_func()
+    
+    multimodal_fns = {
```
kyuyeunk, line 373: this is getting really difficult to manage.

can be done in future pr, but i think we need to considering implementing this proposal: https://github.com/vllm-project/tpu-inference/issues/1501

tpu_inference/layers/common/attention_interface.py
```diff
@@ -53,26 +53,49 @@ def sharded_flash_attention(
     causal: bool = True,
     sm_scale: Optional[float] = None,
     vmem_limit_bytes: int | None = None,
+    use_ab: bool = False,
```
kyuyeunk: what is `ab`?

yuyanpeng-google: attention_bias. I though let the name consistent to the flash attention implementation at first. Let me modify the name.

tpu_inference/models/common/model_loader.py
yuyanpeng-google: Thanks for point this out.

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: It's for the multimodal function which directly call `torch.nn.functional.scaled_dot_product_attention` .  The torchax has a default sdpa wrapper, but without flash attention and could easily be OOM.

kyuyeunk: got it. but i don't think this logic belongs in `vllm_model_wrapper`. maybe move this to `tpu_inference/layers` or create a folder called `tpu_inference/ops` and place it there?

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: IIUC, the vllm provides the interface for the model developer, and does not require the implementation need to be compilable. It could have some condition branch depend on the value or etc. Use eager mode at first, and will try to find if there is a more general way to jit them instead model by model.

kyuyeunk: sgtm. note that there is a possibility that a layer that you are having an issue with is a `CustomOp` type. if that's the case, you can override it to our own implementation. example: https://github.com/vllm-project/tpu-inference/pull/1782

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: use `if not self.vllm_config.model_config.is_multimodal_model:` instead

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: Also remove a redundant register op.

tpu_inference/layers/common/attention_interface.py
yuyanpeng-google: modify the name to attention_bias for readability

tpu_inference/runner/tpu_runner.py
```diff
@@ -777,18 +777,18 @@ def _execute_model(
             # Run the multimodal encoder if any.
             # We have the modality embeds at this time.
             self.mm_manager.execute_mm_encoder(scheduler_output)
-            mm_embeds = self.mm_manager.gather_mm_embeddings(
+            mm_embeds, is_mm_embed = self.mm_manager.gather_mm_embeddings(
                 scheduler_output, input_ids.shape[0])
         else:
-            mm_embeds = []
+            mm_embeds = is_mm_embed = None
```
wdhongtw: nit: maybe `mm_embeds, is_mm_embed = None, None`

tpu_inference/runner/multimodal_manager.py
```diff
@@ -195,7 +207,8 @@ def gather_mm_embeddings(self, scheduler_output: "VllmSchedulerOutput",
                 start_idx = max(num_computed_tokens - start_pos, 0)
                 end_idx = min(
                     num_computed_tokens - start_pos + num_scheduled_tokens,
-                    num_encoder_tokens)
+                    num_encoder_tokens,
```
wdhongtw: nit: seems not require here.

tpu_inference/runner/multimodal_manager.py
```diff
@@ -166,9 +167,20 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
 
             self.runner.encoder_cache[mm_hash] = output
 
-    def gather_mm_embeddings(self, scheduler_output: "VllmSchedulerOutput",
-                             target_pad_len: int) -> list[jax.Array]:
+    def gather_mm_embeddings(
```
wdhongtw, line 170: nit: maybe a docstring to describe two response field? Now the two response item have same type so it's a little confusing.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -287,6 +369,89 @@ def step_fun(
...
+            with torchax.default_env():
+                call_kwargs = {}
+                for k, v in kwargs.items():
+                    if isinstance(v, jax.Array):
+                        call_kwargs[k] = torch_view(v)
+                    elif isinstance(v, np.ndarray):
+                        # The "pixel_values" need to be a torch.Tensor.
+                        # Cast it back to torch.Tensor.
+                        call_kwargs[k] = torch_view(jnp.array(v))
+                    elif isinstance(v, torch.Tensor):
```
wdhongtw: Will the last else-branch be used? Should we just assert False there?

tpu_inference/models/jax/qwen2_5_vl.py
```diff
@@ -1066,8 +1066,15 @@ def embed_multimodal(self, image_grid_thw: tuple[tuple[int, int, int],
         return multimodal_embeddings
 
     def embed_input_ids(
-            self, input_ids: jax.Array,
-            multimodal_embeddings: Optional[jax.Array]) -> jax.Array:
+        self,
+        input_ids: jax.Array,
+        multimodal_embeddings: Optional[jax.Array],
```
wdhongtw: nit: signature is not the same as `llama_guard_4.py`

tpu_inference/runner/multimodal_manager.py
```diff
@@ -216,15 +242,34 @@ def gather_mm_embeddings(self, scheduler_output: "VllmSchedulerOutput",
...
+                # use cpu numpy array for inplace modification
+                if is_embed is None:
+                    is_mm_embed_cpu[req_start_pos + start_idx:req_start_pos +
+                                    end_idx] = True
+                else:
+                    # is_embed is torch Tensor in cpu
+                    is_mm_embed_cpu[req_start_pos + start_idx:req_start_pos +
+                                    end_idx] |= is_embed.numpy()
+
+            req_start_idx += num_scheduled_tokens
```
yuyanpeng-google, line 257: Also fix wrong indent here.

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: remove torch.Tensor case merged into else. It could be any type.

tpu_inference/models/jax/qwen2_5_vl.py
yuyanpeng-google: fix

tpu_inference/runner/multimodal_manager.py
yuyanpeng-google: add a docstring

tpu_inference/runner/multimodal_manager.py
yuyanpeng-google: remove the modification

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: Extract the ops to tpu_inference/layers/vllm/ops.py

tpu_inference/runner/tpu_runner.py
yuyanpeng-google: fix

kyuyeunk (approved):
unsure about `multimodal_manager.py` part, but other files look good.

yuyanpeng-google:
Remove `handle_oov_mm_token` per https://github.com/vllm-project/vllm/pull/34858

PR #1953, opened 2026-03-18
fix einsum with batch dimension

yuyanpeng-google:
# Description

fix einsum with batch dimension throw exception.

# Tests

`pytest tests/layers/jax/quantization/test_fp8.py tests/layers/vllm/test_unquantized.py tests/lora/test_layers.py`

# Checklist

### DIFF
--- a/tpu_inference/layers/common/quantization/unquantized.py
+++ b/tpu_inference/layers/common/quantization/unquantized.py
@@ -35,7 +35,7 @@ class UnquantizedLinearMethod:
                      x_jax: jax.Array,
                      weight_jax: jax.Array,
                      bias_jax: Optional[jax.Array] = None,
-                     einsum_str: str = "mn,pn->mp") -> jax.Array:
+                     einsum_str: str = "...mn,pn->...mp") -> jax.Array:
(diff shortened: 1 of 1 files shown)


kyuyeunk:
when does this error happen? were there no unit test to catch this error?

yuyanpeng-google:
> when does this error happen? were there no unit test to catch this error?

I catch the error while testing the multimodal support of some models. It called from `vllm/model_executor/models/siglip.py` .  (https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/siglip.py#L426)

Could you recommend a file I could put the test in?

kyuyeunk:
> > when does this error happen? were there no unit test to catch this error?
> 
> I catch the error while testing the multimodal support of some models. It called from `vllm/model_executor/models/siglip.py` . (https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/siglip.py#L426)
> 
> Could you recommend a file I could put the test in?

thanks for the explanation! for unit test, somewhere in here would work: https://github.com/vllm-project/tpu-inference/tree/main/tests/layers/vllm

PR #1774, opened 2026-02-20
add Gemma3ForCausalLM 

panalexeu:
# Description

This PR adds Gemma 3 support for TPU inference by implementing the `Gemma3ForCausalLM` architecture in JAX/Flax.

The implementation follows the Gemma 3 architecture which introduces a sliding window attention pattern — local attention layers use a bounded sliding window while global attention layers attend to the full sequence. The model alternates between these based on the `sliding_window_pattern` config. It also uses per-layer QK normalization and `gelu_pytorch_tanh` activation.

References for the architecture:
- https://developers.googleblog.com/gemma-explained-whats-new-in-gemma-3/
- https://arxiv.org/pdf/2503.19786
- https://github.com/google-deepmind/gemma

Changes:
- **`tpu_inference/models/jax/gemma3.py`**: Full model implementation — `RMSNorm`, `Gemma3MLP`, `Gemma3Attention`, `Gemma3DecoderLayer`, `Gemma3Model`, and `Gemma3ForCausalLM` with weight loading mappings from HF checkpoints.
- **`tpu_inference/models/common/model_loader.py`**: Registered `Gemma3ForCausalLM` in the model registry.
- **`tpu_inference/kernels/ragged_paged_attention/v3/kernel.py`**: Fixed sliding window handling in the general RPA kernel. References to the first KV block index (accumulator init in `load_with_init`, the `wait_cur_bq` guard, and the `fori_loop` start) were hardcoded to `0` instead of using `bkv_idx_start`. This was already done correctly in `kernel_hd64.py` but was missing here, causing sliding window attention to fail in the main kernel. Ported that same fix over.

Sorry if something is missing — this is one of my first contributions. Happy to polish or rework anything!

# Tests

- Added `tests/models/jax/test_gemma3.py` which tests model init, weight loading, and forward pass for `google/gemma-3-1b-it`.
- Tests run with and without `qwix` int8 quantization rules.
- Verifies layer structure, projection shapes, local vs global attention config (rope theta, sliding window), and output shapes through the full forward pass including logit computation.

Run with:
```
pytest tests/models/jax/test_gemma3.py -v
```

# Checklist

Before submitting this PR, please make sure:
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have made or will make corresponding changes to any relevant documentation.

PR #1820, opened 2026-02-27
[RPA3] Separate kernel to 3 calls and many optimizations

bythew3i:
# Description

This cl made big changes to RPA v3:

1. Separate RPA to 3 calls: decode-only, prefill-only, mixed
2. Allow user to pass 3 sets of block sizes for different cases
3. Skip unnecessary bkv fetching (huge gain for prefill heavy)
4. Improve strided load (resolve SREG spilling issue, handle bank conflicts)
5. Introduce "heuristic block sizes formula" as default block size finder for untuned models. (Note current default formula is built on top of Lllama 3 model)

# Tests

Tested:
```
pytest -x tests/kernels/ragged_paged_attention_kernel_v3_test.py 
```

# Benchmarking
* Llama 3-8: 12.89 -> 14.64 reqs/s (**1.14x**)

Details:

```
VLLM_USE_V1=1 vllm serve meta-llama/Llama-3.1-8B-Instruct  --disable-log-requests --port 8003 --gpu-memory-utilization 0.95 --max-num-batched-tokens 512 --tensor-parallel-size 1 --max-model-len 2048 --max_num_seqs 512 --download_dir=/mnt/pd/models &> /mnt/pd/tmp/serve.log &

vllm bench serve --backend vllm --model meta-llama/Llama-3.1-8B-Instruct --dataset-name random --port=8003 --random-input-len 1800 --random-output-len 128

```

```
Before:
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  77.56     
Total input tokens:                      1799000   
Total generated tokens:                  128000    
Request throughput (req/s):              12.89     
Output token throughput (tok/s):         1650.33   
Peak output token throughput (tok/s):    2199.00   
Peak concurrent requests:                1000.00   
Total token throughput (tok/s):          24845.15  
---------------Time to First Token----------------
Mean TTFT (ms):                          38348.85  
Median TTFT (ms):                        38326.25  
P99 TTFT (ms):                           75559.43  
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          20.09     
Median TPOT (ms):                        20.30     
P99 TPOT (ms):                           20.40     
---------------Inter-token Latency----------------
Mean ITL (ms):                           20.10     
Median ITL (ms):                         20.36     
P99 ITL (ms):                            21.54     
==================================================


After:
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
[...]

PR #1954, opened 2026-03-18
fix RPA kernel for SW attention

panalexeu:
#1774 

> **`tpu_inference/kernels/ragged_paged_attention/v3/kernel.py`**: Fixed sliding window handling in the general RPA kernel. References to the first KV block index (accumulator init in `load_with_init`, the `wait_cur_bq` guard, and the `fori_loop` start) were hardcoded to `0` instead of using `bkv_idx_start`. This was already done correctly in `kernel_hd64.py` but was missing here, causing sliding window attention to fail in the main kernel. Ported that same fix over.

cc: @kwang3939

### DIFF
--- a/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
+++ b/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
@@ -427,7 +427,7 @@ def _ragged_paged_attention_kernel(
         head_acc_ref = acc_ref.at[kv_head_idx, :p.shape[0]]
 
         def load_with_init(ref, init_val):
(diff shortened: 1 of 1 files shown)


kyuyeunk:
Code change looks good to me. @bythew3i can you also take a look?

bythew3i:
It is also fixed in the #1820

kyuyeunk:
@panalexeu can you confirm that the pr fixed the issue? https://github.com/vllm-project/tpu-inference/pull/1820

panalexeu:
@kyuyeunk yes #1820 solves the issue

PR #1959, opened 2026-03-18
use make_array_from_callback API for multihost device put

sixiang-google:
# Description

Transitioned from manually slicing and placing arrays using jax.device_put inside a list comprehension to using JAX's higher-level jax.make_array_from_callback API. https://docs.jax.dev/en/latest/_autosummary/jax.make_array_from_callback.html

This API makes use of JAX internal thread pool for transferring and enables multi-host parallel.

Perf change:
prepare_inputs() 18ms -> 7ms. 

Also: some changes to enable Ray on single host run. For sole debug purposes.

# Tests

commands to reproduce.

# Checklist


tpu_inference/executors/ray_distributed_executor.py
```diff
@@ -364,7 +364,11 @@ def sort_by_driver_then_worker_ip(item: RayWorkerMetaData):
         # Environment variables to copy from driver to workers
         env_vars_to_copy = get_env_vars_to_copy(
             exclude_vars=self.WORKER_SPECIFIC_ENV_VARS,
-            additional_vars=set(current_platform.additional_env_vars),
+            additional_vars=set(current_platform.additional_env_vars) | {
```
Lumosis: Can we add the extra vars here: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/platforms/tpu_platform.py#L49

PR #1965, opened 2026-03-18
Fix jax path device_put for deepseek

Lumosis:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist


tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -230,14 +230,19 @@ def shard_put(x: jax.Array,
         x_mesh = x.sharding.mesh
 
     if math.prod(mesh.axis_sizes) == 1:
-        return general_device_put(x,
-                                  mesh.devices.flatten()[0],
-                                  source_mesh=x_mesh)
+        return jax.device_put(x, mesh.devices.flatten()[0])
```
gxd3, line 233: curious: why use jax.device_put instead of general_device_put here?

gxd3 (approved):

tpu_inference/models/jax/utils/weight_utils.py
Lumosis: general_device_put requires the input to be a instance of NamedSharding. `mesh.devices.flatten()[0]` is device instead of mesh.devices.flatten()[0], so we cannot use general_device_put directly here

PR #1974, opened 2026-03-19
Add fix to CompilationManager to support Qwen-3VL multimodal input and add support to multi_modal_inference script

muskansh-google:
This PR improves multimodal inference script to support Qwen3-VL model, and applies critical fixes to the CompilationManager to ensure proper JAX tracing. 

Key Changes
- multi_modal_inference.py:
  - Multi-Image Support: Added a --test-multi-image flag and supporting logic to test multiple images per prompt.
  - Model Expansion: Generalized the runner to support Qwen3-VL in addition to Qwen2.5-VL.
  - Configuration Updates: Disabled chunked prefill, increased default GPU memory utilization to 0.85, and increased max_tokens to 1024.
  - Bug Fix: Added fuse_minimax_qk_norm: False to pass_config to fix a Pydantic validation error in recent vLLM versions.

- compilation_manager.py:
  - Tracing Fix: Passed an explicit boolean dummy_is_multimodal mask during precompilation to handle mixed text/multimodal batches without runtime recompilation.
  - DeepStack Support: Added logic to compute dimensions for DeepStack visual concatenation.

# Tests

### Command used
```
python examples/multi_modal_inference.py \
  --model Qwen/Qwen3-VL-8B-Instruct \
  --test-multi-image \
  --max-model-len 8192
```
Logs
```
The two images depict very different scenes, contrasting natural beauty with urban culture:

**First Image:**
*   **Subject:** A view of the **Tokyo Skytree** in Japan.
*   **Composition:** The iconic, modern tower is seen from a low angle, framed by the vibrant pink blossoms of cherry trees in full bloom.
*   **Atmosphere:** The scene is serene and picturesque, capturing the beauty of spring in Japan. The bright blue sky and soft-focus flowers create a dreamy, almost magical quality.

**Second Image:**
*   **Subject:** The **Chinatown Gate (Chinatown Archway)** in San Francisco, USA.
*   **Composition:** The image shows a street-level view looking through the ornate, traditional Chinese archway. A red "STOP" sign is prominently in the foreground, and a black SUV is driving past.
*   **Atmosphere:** This is a bustling, urban scene. The presence of the stop sign, the car, and the surrounding storefronts (like "OPTUS" and "yes") grounds the image in a modern, everyday city environment, contrasting with the traditional architecture of the gate.

In summary, the first image is a tranquil, natural scene featuring a landmark in Japan, while the second is a dynamic, urban street scene featuring a cultural landmark in the United States.
```

### Testing video modality

Need to manually lower the threshold in [flash attention kernel](https://www.google.com/url?q=https://github.com/vllm-project/tpu-inference/blob/3c1ce2af11544a80eff469155e4291537609b4a1/tpu_inference/kernels/flash_attention/kernel.py%23L135&sa=D&source=buganizer&usg=AOvVaw0-Y-vuJGheDX4-UEqCMKnq) to `16384`

#### vLLM Serve Command
```
vllm serve Qwen/Qwen3-VL-8B-Instruct \
  --max-model-len 4096 \
  --disable-uvicorn-access-log \
  --max-num-seqs 128 \
  --tensor-parallel-size 1 \
  --trust-remote-code \
  --limit-mm-per-prompt '{"image": 1}' \
  --no-enable-prefix-caching --gpu_memory_utilization 0.85 > vllm_server.log 2>&1 &
```

#### vLLM Bench Command
```
vllm bench serve \
    --backend openai-chat \
    --model Qwen/Qwen3-VL-8B-Instruct \
    --dataset-name random-mm \
    --num-prompts 1 \
    --random-mm-bucket-config '{(3000, 1000, 1): 1.0}' \
    --random-mm-limit-mm-per-prompt '{"image": 1, "video": 1}' \
    --random-mm-base-items-per-request 1 \
    --random-input-len 505 \
    --random-output-len 450 \
    --ignore-eos
```


tpu_inference/runner/persistent_batch_manager.py
```diff
@@ -176,16 +177,27 @@ def update_states(self, scheduler_output: "VllmSchedulerOutput",
...
-                        self.requests[req_id].prompt_token_ids,
-                        hf_config=hf_config,
-                        image_grid_thw=image_grid_thw,
-                        video_grid_thw=video_grid_thw,
-                        second_per_grid_ts=second_per_grid_ts,
-                        audio_feature_lengths=audio_feature_lengths,
-                        use_audio_in_video=use_audio_in_video,
-                    )
+                # Simple if-else logic for Qwen3-VL vs Qwen2.5-VL
+                if "qwen3" in self.model_config.model.lower():
```
kyuyeunk: this kind of model specific logic doesn't seem like a good idea? instead, can you name it based on functionality? for instance, insted of `if model == qwen3`, do `if config.perform_mrope_positioning`.

 also, even if you were to use this kind o logic, please use `architectures` field of model config with strict string matching and not `in`: https://huggingface.co/Qwen/Qwen2.5-VL-32B-Instruct/blob/main/config.json#L3

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -398,6 +637,16 @@ def embed_input_ids_func(
 
         return embed_input_ids_func
 
+    def wrap_embed_input_ids_func(self):
+        if not self.vllm_config.model_config.is_multimodal_model:
+            return None
+
+        is_qwen3_vl = type(self.model.vllm_model).__name__ == "Qwen3VLForConditionalGeneration"
```
kyuyeunk: same comment as below.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -381,15 +575,60 @@ def embed_input_ids_func(
...
+                        "call_kwargs": call_kwargs,
+                    },
+                    tie_weights=False,
+                )
+
+                return jax_view(output_from_torch)
+
+        return embed_input_ids_func
+
+    def _wrap_qwen3_vl_embed_input_ids_func(self):
```
kyuyeunk: similar comment as below. please don't use model specific naming. instead, please name it based on functionality.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -146,6 +148,88 @@ def _patch_vllm_ops(self):
             needs_env=False,
         )
 
+        @register_function(
+            torch.ops.vllm.torch_sdpa_wrapper,
+            is_jax_function=True,
+            needs_env=False,
+        )
+        def patched_vllm_vit_sdpa(
```
kyuyeunk: please define the op in this file instead: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/ops.py

kyuyeunk (changes requested):

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -381,15 +575,60 @@ def embed_input_ids_func(
                 else:
                     call_args = (torch_view(input_ids), )
 
+                call_kwargs = {
+                    "is_multimodal": torch_view(is_multimodal),
+                }
+                if handle_oov_mm_token:
+                    call_kwargs["handle_oov_mm_token"] = handle_oov_mm_token
```
yuyanpeng-google: https://github.com/vllm-project/vllm/pull/34858

I thought this args  `handle_oov_mm_token` was removed.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: Done, moved the op to the suggested file and using ```register_torch_dispatch_op``` to register the dispatch operation

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: Missed this while resolving merge conflicts, removed it's usage.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: Done, have refactored the code to use common code and define logic based on functionality instead of model.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: Done

tpu_inference/runner/persistent_batch_manager.py
muskansh-google: Done, changed the conditional logic to check functionality instead of model name.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -375,21 +505,26 @@ def embed_input_ids_func(
...
+                    elif always_wrap_list:
+                        # Specific fix for Qwen3-VL which expects mm_embeds to be a list of Tensors.
+                        torch_mm_embeds = [torch_view(mm_embeds)]
                     else:
                         torch_mm_embeds = torch_view(mm_embeds)
                     call_args = (torch_view(input_ids), torch_mm_embeds)
                 else:
                     call_args = (torch_view(input_ids), )
 
+                call_kwargs = {
```
kyuyeunk: why change this?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -360,10 +484,16 @@ def embed_multimodal_func(
...
     def wrap_embed_input_ids_func(self):
+        # The function cannot be JITted directly due to its dynamic implementation
+
         if not self.vllm_config.model_config.is_multimodal_model:
             return None
 
-        # The function cannot be JITted directly due to its dynamic implementation
+        always_wrap_list = False
+        architectures = getattr(self.vllm_config.model_config.hf_config, "architectures", [])
+        if "Qwen3VLForConditionalGeneration" in architectures:
```
kyuyeunk: hmm still not really a fan of this approach. is there any characteristics of a model architecture - that is agnostici to the model name (i.e., string value) itself - that can be used to determine if this feature is needed or not?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -329,11 +443,21 @@ def embed_multimodal_func(
             image_grid_thw: Any,
             **kwargs,
         ) -> Any:
-            # Del args for jax path. Need to refactor the function call.
-            del image_grid_thw
+            inner_model = getattr(self.model, "vllm_model", self.model)
+            method = getattr(inner_model, "embed_multimodal", None)
```
kyuyeunk: please avoid using string based `getattr`. and do something like `if model == multimodal: do x`. if the member variable is not found even when `if model == multimodal` is true, we can just simply raise an error.

similar comment for all the other lines that uses `getattr`

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -329,11 +443,21 @@ def embed_multimodal_func(
...
+            sig = inspect.signature(method) if method else None
+            has_var_keyword = any(p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values()) if sig else False
```
kyuyeunk: same comment as above. please avoid any string based approach.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -161,6 +172,109 @@ def _apply_pp_patch(self):
...
+        from tpu_inference.distributed.jax_parallel_state import \
+            get_pp_group as jax_get_pp_group
+
+        logger.info("Applying Qwen3-VL stateless Deepstack patches")
+
+        # 1. Override setter to avoid in-place mutation error
+        orig_set_deepstack = getattr(vllm_model, "_set_deepstack_input_embeds", None)
+        if orig_set_deepstack is not None:
+            def patched_set_deepstack(deepstack_input_embeds):
+                indexes = getattr(vllm_model.config.vision_config, "deepstack_visual_indexes", [])
```
kyuyeunk: same comment as all the others. please avoid using `getattr`

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: Not relevant now.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: It's difficult to base the conditional logic on just functionality like is_multimodal or some other logic, because all of this code is very specific to Qwen3-VL. For example, this logic is not found in even Qwen2.5-VL. 

I am not able to come up with any logic which does not use either model name or based on functionality, any suggestions for this?

Instead of using ```getattr```, I think conditioning on model specific config name would be more explicit and clear for a reader. But I am keeping it as it is for now, and we can think of coming up with model agnostic code in a follow-up.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: It's difficult to base the conditional logic on just functionality like is_multimodal or some other logic, because all of this logic is very specific to Qwen3-VL. For example, this logic is not found in even Qwen2.5-VL.
we plan to implement model agnostic architecture as a follow-up. It's not easy to reconcile model specific logic for different models using just functionality logic.

I am resolving for now, feel free to reopen if you have any suggestions on how we can make this better.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: The method _apply_qwen3_vl_patches is only specific for Qwen3-VL. We plan to make change to model agnostic architecture as a follow-up.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: It's difficult to base the conditional logic on just functionality like is_multimodal or some other logic, because all of this logic is very specific to Qwen3-VL. For example, this logic is not found in even Qwen2.5-VL. 
we plan to implement model agnostic architecture as a follow-up. It's not easy to reconcile model specific logic for different models using just functionality logic.

I am resolving for now, feel free to reopen if you have any suggestions on how we can make this better.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -30,8 +31,9 @@
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.interop import jax_view, torch_view
 from torchax.ops.mappings import TORCH_DTYPE_TO_JAX
-from torchax.ops.ops_registry import register_torch_function_op
-from vllm.config import VllmConfig, set_current_vllm_config
+from torchax.ops.ops_registry import (register_torch_dispatch_op,
+                                         register_torch_function_op)
+from vllm.config import VllmConfig, set_current_vllm_config, set_current_vllm_config
```
kwang3939: There are two `set_current_vllm_config` here.

examples/multi_modal_inference.py
```diff
@@ -69,7 +79,8 @@ def run_qwen2_5_vl(questions: list[str], modality: str,
 
 
 model_example_map = {
-    "qwen2_5_vl": run_qwen2_5_vl,
+    "Qwen/Qwen2.5-VL-3B-Instruct": run_qwen_vl,
```
kwang3939: I think specify `Qwen2.5-VL-3B-Instruct` might be not general enough in case we want to test `Qwen2.5-VL-7B`. Could you make this more general?

examples/multi_modal_inference.py
```diff
@@ -215,23 +249,33 @@ def parse_args():
...
     data = mm_input["data"]
     questions = mm_input["questions"]
 
-    # NOTE: Currently, only Qwen2.5-VL is supported. If later we want to support a model with new chat template, we may need to change this
-    req_data = model_example_map["qwen2_5_vl"](questions, modality, args)
+    # NOTE: Currently, only Qwen2.5-VL and Qwen3-VL is supported. If later we want to support a model with new chat template, we may need to change this
+    model_key = args.model
+
+        
+    req_data = model_example_map[model_key](questions, modality, args)
```
kwang3939: Could you add a check to see if model_key is in the map and raise error if not?

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -83,3 +83,79 @@ def scaled_dot_product_attention(
...
+
+        q_seg = jnp.broadcast_to(q_seg, (batch, q_seg.shape[0]))
+        kv_seg = jnp.broadcast_to(kv_seg, (batch, kv_seg.shape[0]))
+
+        from tpu_inference.kernels.flash_attention.kernel import SegmentIds
+        seg_ids = SegmentIds(q=q_seg, kv=kv_seg)
+    else:
+        seg_ids = None
+
+    from tpu_inference.layers.common.attention_interface import \
```
kwang3939: This is already imported at the top of this file.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -155,6 +164,109 @@ def _apply_pp_patch(self):
...
+        orig_set_deepstack = getattr(vllm_model, "_set_deepstack_input_embeds", None)
+        if orig_set_deepstack is not None:
+            def patched_set_deepstack(deepstack_input_embeds):
+                indexes = getattr(vllm_model.config.vision_config, "deepstack_visual_indexes", [])
+                
+                if not hasattr(vllm_model, "_deepstack_tensors"):
+                    vllm_model._deepstack_tensors = {}
+
+                if isinstance(deepstack_input_embeds, dict):
+                    vllm_model._deepstack_tensors.update(deepstack_input_embeds)
```
kwang3939: One concern here: do we need to reset this `_deepstack_tesnors` to `{}` if `deepstack_input_embeds` is None? Otherwise it may retain the values from the previous forward path.

tpu_inference/runner/persistent_batch_manager.py
```diff
@@ -176,16 +177,30 @@ def update_states(self, scheduler_output: "VllmSchedulerOutput",
...
-                        hf_config=hf_config,
-                        image_grid_thw=image_grid_thw,
-                        video_grid_thw=video_grid_thw,
-                        second_per_grid_ts=second_per_grid_ts,
-                        audio_feature_lengths=audio_feature_lengths,
-                        use_audio_in_video=use_audio_in_video,
-                    )
+                # Qwen3-VL uses a different method signature and takes in mm_features as an argument.
+                import inspect
+                takes_mm_features = "mm_features" in inspect.signature(get_mrope_input_positions_fn).parameters
```
kwang3939: Could you use `supports_kw` here instead of `inspect.signature`?
Also, I think this can be moved outside the for loop: https://github.com/muskansh-google/tpu-inference/blob/dd49ffcca0d74b577ea88512bd45742ee04a67ea/tpu_inference/runner/persistent_batch_manager.py#L132.

tpu_inference/models/vllm/patches/__init__.py
```diff
@@ -0,0 +1 @@
+# This file makes the directory a Python package.
```
lk-chen: we need copy right disclaimer here, and qwen3_vl.py https://github.com/vllm-project/tpu-inference/blob/4e332d39f789544411d014238e7f47c57941ea9d/tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py#L1-L13

tpu_inference/runner/multimodal_manager.py
```diff
@@ -177,6 +177,13 @@ def gather_mm_embeddings(
                 start_pos = pos_info.offset
                 num_encoder_tokens = pos_info.length
 
+                if hasattr(pos_info, "is_embed") and pos_info.is_embed is not None:
```
lk-chen: I assume this is debug only?

tpu_inference/runner/tpu_runner.py
```diff
@@ -586,11 +586,18 @@ def load_model(self):
         disable_mm_from_limits = False
         if self.model_config.is_multimodal_model:
             mm_limits = self.model_config.multimodal_config.limit_per_prompt
+            
             image_limit = mm_limits.get("image")
             video_limit = mm_limits.get("video")
-            image_count = image_limit.count if image_limit else 0
-            video_count = video_limit.count if video_limit else 0
-            disable_mm_from_limits = image_count == 0 and video_count == 0
+
```
lk-chen: This part has been fixed https://github.com/vllm-project/tpu-inference/pull/2213

lk-chen (commented):
I think
- tpu_inference/layers/common/quantization/unquantized.py
- tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py

are self-contained, and can be moved into 2 separate PRs for easier review?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -465,11 +497,21 @@ def embed_input_ids_func(
             is_multimodal: jax.Array | None = None,
         ) -> jax.Array:
             with torchax.default_env():
+                if mm_embeds is not None and is_multimodal is not None:
```
lk-chen: This embedding truncate logic can also be a standalone PR.
Also if I read [this part](https://github.com/vllm-project/vllm/blob/bcc2306cefa4179c548d3e638e7a22a88d281733/vllm/model_executor/models/interfaces.py#L341-L351) and [this](https://github.com/vllm-project/vllm/blob/bcc2306cefa4179c548d3e638e7a22a88d281733/vllm/model_executor/models/interfaces.py#L406) correctly, `is_multimodal` must be provided when `mm_embeds` is provided

tpu_inference/runner/multimodal_manager.py
muskansh-google: Yes, removed

tpu_inference/runner/tpu_runner.py
muskansh-google: Removed.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -427,12 +436,25 @@ def embed_multimodal_func(
...
-                    logger.warning(f"Expect torch.Tensor, got {type(v)}")
-                    return v
-                return v.to(device="jax")
+            # Pop it so it doesn't go through the 'move' loop if we want to handle it specially
+            image_grid_thw = kwargs.pop("image_grid_thw", None)
+
+            def move(v: Any) -> Any:
+                import numpy as np
+                if isinstance(v, np.ndarray):
+                    if str(v.dtype) == "bfloat16":
```
kyuyeunk: You are using string to check dtype? Can you elaborate why?

kyuyeunk (commented):
It's hard to provide feed back to all of them, but I get the impression that this pr isn't really a high quality.

For instance, using string based checking should really really be avoided and instead use object type based checking. But this pr has excessive used of that.

There so much more but it's a bit difficult to comment on all of them.

tpu_inference/models/vllm/vllm_model_wrapper.py
muskansh-google: I apologize, this code was a result of mistake I made while resolving merge conflicts. This is not required.

muskansh-google:
> It's hard to provide feed back to all of them, but I get the impression that this pr isn't really a high quality.
> 
> For instance, using string based checking should really really be avoided and instead use object type based checking. But this pr has excessive used of that.
> 
> There so much more but it's a bit difficult to comment on all of them.

I appreciate the feedback. The PR was in a bit of a 'work-in-progress' state while I was splitting it and trying to resolve merge conflicts, and I forgot to change it to "Draft" state. The PR is now updated with the final changes. It should be much cleaner now.

tpu_inference/runner/compilation_manager.py
```diff
@@ -124,25 +124,50 @@ def capture_model(self) -> None:
...
+        import functools
         for num_tokens in self.runner.num_tokens_paddings:
-            hidden_size = self.runner.vllm_config.model_config.get_hidden_size(
-            )
+            hidden_size = self.runner.vllm_config.model_config.get_hidden_size()
+            hf_conf = self.runner.vllm_config.model_config.hf_config
+
+            # Identify multimodal embedding size
+            mm_hidden_size = hidden_size
+            try:
```
kyuyeunk: this try/except isn't really a good pattern. can't you just do `isinstance` with `Qwen3VLVisionConfig` and `Qwen3VLConfig` instead?

example:
- https://github.com/vllm-project/vllm/blob/bf45e6d0a558da2b8d7b60efb07b4aa394f3b60b/vllm/model_executor/models/qwen3_vl.py#L521
- https://github.com/vllm-project/vllm/blob/bf45e6d0a558da2b8d7b60efb07b4aa394f3b60b/vllm/model_executor/models/qwen3_vl.py#L42-L45

kyuyeunk:
> > It's hard to provide feed back to all of them, but I get the impression that this pr isn't really a high quality.
> > For instance, using string based checking should really really be avoided and instead use object type based checking. But this pr has excessive used of that.
> > There so much more but it's a bit difficult to comment on all of them.
> 
> I appreciate the feedback. The PR was in a bit of a 'work-in-progress' state while I was splitting it and trying to resolve merge conflicts, and I forgot to change it to "Draft" state. The PR is now updated with the final changes. It should be much cleaner now.

thanks for the clarification! yes it does look much better!

tpu_inference/runner/compilation_manager.py
muskansh-google: Done. Have used "getattr" instead of "isinstance" to make it conditional on feature instead of Model config class. Does this look better? If not, I can just use `isinstance` instead.

muskansh-google:
The changes for pre-compilation got merged in a separate PR. CLosing this PR for now. Will raise the changes in the example script in a separate PR.

PR #1979, opened 2026-03-20
round kv cache to allow expert sharding

khatwanimohit:

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/kv_cache_manager.py
+++ b/tpu_inference/runner/kv_cache_manager.py
@@ -394,9 +394,19 @@ class KVCacheManager:
                 num_blocks = kv_cache_tensor.size // page_size_bytes
                 if duplicate_shared_layers:
                     num_blocks //= num_shared_layers
-                dp_size = self.runner.vllm_config.sharding_config.total_dp_size
-                # num_blocks must be a multiple of dp_size
-                num_blocks = (num_blocks // dp_size) * dp_size
+                sharding_config = self.runner.vllm_config.sharding_config
+                if self.use_mla and not self.runner.vllm_config.additional_config.get(
+                        "sharding", {}).get("sharding_strategy", {}).get(
+                            "enable_dp_attention", False):
+                    # MLA KV cache is sharded with MLP_TENSOR = (attn_dp, attn_dp_expert, model, expert)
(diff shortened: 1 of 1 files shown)


tpu_inference/runner/kv_cache_manager.py
```diff
@@ -303,9 +303,17 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
             num_blocks = kv_cache_tensor.size // page_size_bytes
-            dp_size = self.runner.vllm_config.sharding_config.total_dp_size
-            # num_blocks must be a multiple of dp_size
-            num_blocks = (num_blocks // dp_size) * dp_size
+            sharding_config = self.runner.vllm_config.sharding_config
+            if self.use_mla:
+                # MLA KV cache is sharded with MLP_TENSOR = (attn_dp, attn_dp_expert, model, expert)
+                divisor = (sharding_config.attn_dp_size *
+                           sharding_config.attn_dp_expert_size *
+                           sharding_config.tp_size *
```
gpolovets1: Wouldn't this be the same as sharding_config.total_dp_size if self.use_mla is True? https://github.com/vllm-project/tpu-inference/blob/b39f07b2ed2f357d2309859f3be9724fd50d7d96/tpu_inference/layers/common/sharding.py#L173

if self.use_mla is True then TP is transferred to attn_dp_size and EP is transferred to attn_dp_expert_size.

khatwanimohit: since we are running maxtext model here through vllm, we are controlling the shardings and therefore this code of transferring tp/ep to attn_dp_* is not being run.

gpolovets1: Do you mean you are setting enable_dp_attention=False?

khatwanimohit: yes, we are setting that to false and just using DP + EP shardings

gpolovets1: Maybe you can add that in the conditional for clarity (i.e. enable_dp_attention == False)?

khatwanimohit: added that check

PR #1981, opened 2026-03-20
Add optional small graph compilation control var

sixiang-google:
# Description

In local disagg we are launching 2 different engines which might have the same mesh structure but land on different sets of device. Some hlo on 2 devices hash collide and would result in device out of bound during execution. Having jax_persistent_cache_min_entry_size_bytes on effect would avoid such collision

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -52,9 +52,12 @@ class CompilationManager:
             logger.info("Enabling JAX compile cache.")
             jax.config.update("jax_compilation_cache_dir",
                               vllm_envs.VLLM_XLA_CACHE_PATH)
-            # Ensure small compiled function are cached as well.
-            jax.config.update("jax_persistent_cache_min_entry_size_bytes", -1)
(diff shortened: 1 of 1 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -44,17 +44,22 @@
...
         self._sampling_precompiled = False
         self._gather_logprobs_precompiled = False
         if not vllm_envs.VLLM_DISABLE_COMPILE_CACHE:
             logger.info("Enabling JAX compile cache.")
             jax.config.update("jax_compilation_cache_dir",
                               vllm_envs.VLLM_XLA_CACHE_PATH)
-            # Ensure small compiled function are cached as well.
-            jax.config.update("jax_persistent_cache_min_entry_size_bytes", -1)
-            jax.config.update("jax_persistent_cache_min_compile_time_secs", -1)
+            if cache_small_functions:
```
gxd3: nit: to simplify things, how about 
if VLLM_XLA_CHECK_RECOMPILATION:
    .......

PR #1987, opened 2026-03-20
[TPU] Support Async Logprobs with Lazy Materialization

sierraisland:
# Description

Previously, logprobs were calculatedimmediately after the sampling step in each iteration. This caused a CPU overhead would block and wait for the TPU to finish computing and transferring logprob data.

Since Tunix team do not need the partial results, this PR introduces Lazy Materialization for logprobs in the TPU ModelRunner. By moving the data retrieval and conversion logic into the asynchronous output object, we effectively hide the synchronization latency between the TPU and the Host.

FIXES: b/489119662

# Tests

Benchmark result matched the one without logprobs.
decode_heavy and prefill_heavy do not have large logprobs overheads.

# Checklist


tpu_inference/runner/tpu_runner.py
```diff
@@ -193,27 +201,38 @@ def _substitute_placeholder_token(
...
-        ]
+        log_token_ids = log_token_ids[logits_indices_selector]
+        logprobs_arr = logprobs_arr[logits_indices_selector]
+        selected_token_ranks = selected_token_ranks[logits_indices_selector]
 
     return LogprobsLists(
-        logprob_token_ids=np.asarray(log_token_ids_list),
-        logprobs=np.asarray(logprobs_list),
-        sampled_token_ranks=np.asarray(selected_token_ranks_list),
+        logprob_token_ids=np.array(log_token_ids.tolist()),
```
wenxindongwork, line 233: any reason why we are changing np.asarray to np.array here?

sierraisland: The change from `np.asarray` to `np.array(...tolist())` is a defensive measure to guarantee a deep copy and fully decouple the data from JAX's memory management. If a JAX array has been moved to the host (via `device_get`) and is C-contiguous, `np.asarray` creates a NumPy array that points directly to the JAX internal buffer. Calling `.tolist()` on a JAX array performs an explicit, element-wise conversion into a native Python list. With `np.array(...tolist())`, once the list is created, it has zero connection to the XLA/JAX backend buffers.

PR #1997, opened 2026-03-23
[Multi-Modal] Skip dedicated MM logic and properly move tensors

wdhongtw:
- Wrap QWen 2.5 VL jax specific logic into a conditional branch.
- Tree-map the MM kwargs into "jax" accelerator for torchax path.

# Description

MultiModal models yields a PyTree of `torch.Tensor`, and the PyTree of `torch.Tensor` then got feed into `embed_multimodal` of each model.
(See `MultiModalInputs` and `MultiModalFieldElem` from vLLM source tree for more details.)

In path common to jax/torchax, we should not do processing that's specific to particular path and/or model,
so in MultiModalManager (part of TPURunner), we hide the logic beind `'image_grid_thw' in batched_mm_inputs` condition.
Not perfect but works for now.

Also in torchax path, we avoding convert from torch.Tensor to numpy.ndarray (part of the logic that skipped above) then jax.Array then back to torch.Tensor,
just move the tensor into jax accelerator directly.

# Tests

Serve any model in torchax path and check if multimodel requests, e.g. a request with text and image, still works.

# Checklist

### DIFF
--- a/tpu_inference/runner/multimodal_manager.py
+++ b/tpu_inference/runner/multimodal_manager.py
@@ -113,32 +113,34 @@ class MultiModalManager:
         for _, num_items, mm_kwargs_group in group_mm_kwargs_by_modality(
                 mm_kwargs):
             batched_mm_inputs = mm_kwargs_group
-            # Convert torch tensors to numpy arrays that JAX can handle.
-            if "pixel_values" in batched_mm_inputs and isinstance(
-                    batched_mm_inputs["pixel_values"], list):
-                batched_mm_inputs["pixel_values"] = torch.cat(
-                    batched_mm_inputs["pixel_values"], dim=0)
-
             image_grid_thw = ()
-            for key, value in batched_mm_inputs.items():
-                if isinstance(value, torch.Tensor):
-                    if key == 'image_grid_thw':
-                        # change it to tuple of tuples to make it hashable for JIT
-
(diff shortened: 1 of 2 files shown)


tpu_inference/runner/multimodal_manager.py
```diff
@@ -113,32 +113,34 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
...
-                            tuple(row) for row in grid_thw_reshaped.tolist())
-
-                        continue
-
-                    if value.dtype == torch.bfloat16:
-                        batched_mm_inputs[key] = value.to(
-                            torch.float32).numpy().astype(jnp.bfloat16)
-                    else:
-                        batched_mm_inputs[key] = value.numpy()
+            # FIXME(wdhongtw): QWen 2.5 VL specific logic, need to be moved.
```
lk-chen: Let's use [best practice](http://shortn/_U68OHinvqL) to create an issue with details, and add to TODO here

lk-chen (approved):
LGTM with nit

tpu_inference/runner/multimodal_manager.py
wdhongtw: Thanks Linkun, I like this work flow! 👍

wdhongtw: TODO comments changed accordingly.

PR #1999, opened 2026-03-23
Replace multiprocessing.Queue with raw Pipe connections in DP scheduler to eliminate GIL contention

wenxindongwork:
## Summary

This PR replaces `multiprocessing.Queue` with raw `multiprocessing.Pipe` connections in the DP scheduler's inter-process communication.

`multiprocessing.Queue` uses background feeder threads for `put()`, which causes GIL contention when DP is large. Each DP rank had its own input queue plus a separate output queue per command type (`Dict[str, Queue]`), creating a large number of background threads competing for the GIL.

## Solution
- Each worker now has one input pipe and one output pipe (2 connections per rank instead of 1 + N queues)
- Use `cloudpickle.dumps()`/`loads()` directly with `send_bytes()`/`recv_bytes()` to eliminate all background feeder threads
- Add `_send_command()` and `_get_result()` helper methods for clean serialization/deserialization with timing instrumentation

## Test Plan
- Updated `tests/core/test_dp_scheduler.py` 
- Successful `tests/e2e/test_data_parallel.py`

### DIFF
--- a/tests/core/test_dp_scheduler.py
+++ b/tests/core/test_dp_scheduler.py
@@ -757,8 +757,8 @@ class TestDPScheduler:
 
         mock_process_0 = MagicMock()
         mock_process_1 = MagicMock()
-        mock_process_0.is_alive = MagicMock(return_value=False)
-        mock_process_1.is_alive = MagicMock(return_value=False)
+        mock_process_0.is_alive = MagicMock(return_value=True)
+        mock_process_1.is_alive = MagicMock(return_value=True)
         scheduler.processes = [mock_process_0, mock_process_1]
 
         scheduler.shutdown()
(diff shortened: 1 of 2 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -286,11 +286,10 @@ def _precompile_substitute_placeholder_token(self) -> None:
...
                                                       jnp.int32, dp_sharding)
                 # Need align to the sampling output
-                next_tokens = self._create_dummy_tensor(
-                    (num_reqs, ),
-                    jnp.int32,
-                    sharding=NamedSharding(self.runner.mesh, PartitionSpec()),
-                )
+                next_tokens = self._create_dummy_tensor((num_reqs, ),
+                                                        jnp.int32,
+                                                        sharding=dp_sharding)
```
wenxindongwork, line 291: @sixiang-google  FYI I had to make this change to avoid recompilation errors.

PR #2001, opened 2026-03-23
Hybrid MoE, combining EP and TP

zhangamy-crypto:
# Description

Add hybrid expert and tensor parallelism support in MoE; currently maps expert sharding to attention's DP sharding.

# Tests
Tested on Qwen3 Coder and for prefill, improved throughput and latency.

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -492,6 +492,7 @@ def expert_parallel_gmm(
     ep_size = get_mesh_shape_product(mesh, ShardingAxisName.EXPERT)
     ep_p_spec = P(ShardingAxisName.EXPERT)
     data_p_spec = P(ShardingAxisName.MLP_DATA)
+    x_p_spec = P(ShardingAxisName.EXPERT_DATA)
     ep_data_p_spec = P(ShardingAxisName.EXPERT_DATA)
     attn_data_p_spec = P(ShardingAxisName.ATTN_DATA)
     num_experts = w1.shape[0]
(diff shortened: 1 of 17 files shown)


kyuyeunk:
@guowei-dev @gxd3 can you help take a look and review this pr?

gxd3 (approved):

tpu_inference/layers/common/moe.py
```diff
@@ -151,6 +153,7 @@ def moe_apply(
                     renormalize=layer.renormalize,
                     mesh=mesh,
                     use_ep=layer.use_ep,
+                    use_hybrid=(moe_backend == MoEBackend.GMM_HYBRID),
```
guowei-dev, line 157: Currently, use_ep and use_hybrid are set independently with no guard. However, when both are enabled, _process_tokens_locally takes the EP branch and includes the model axis. At the same time, the hybrid weights shard experts over ep_axis only without model_axis, possibly leading to inconsistency issue. Should we enforce an explicit guard to prevent this?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -532,6 +551,101 @@ def expert_parallel_gmm(
...
+    if w2_scale is None:
+        w2_scale_spec = None
+    elif num_blocks == 1:
+        w2_scale_spec = P(ep_axis, None, None, None)
+    else:
+        w2_scale_spec = P(ep_axis, tp_axis, None, None)
+
+    w2_bias_spec = None if w2_bias is None else P(ep_axis, None, None)
+
+    data_out_spec = P(ep_axis)
```
guowei-dev: Should we also branch on if scatter_results -> P(ATTN_DATA) / elif enable_rs_kernel -> P(EXPERT_DATA) , similar to  tensor_parallel_gmm/expert_parallel_gmm? Otherwise, when there’s no DP-attention, the reduction becomes a plain psum(ep+tp) and every shard already holds the full, identical output, but the fixed P(ep_axis) would still tell shard_map each shard is a distinct ep-slice.

lk-chen (approved):

PR #2020, opened 2026-03-25
[MoE][Unquantized] reuse process_unquantized_moe_weights from torchax

lk-chen:
# Description

`process_moe_weights` does [padding](https://github.com/vllm-project/tpu-inference/blob/2f7640057816392d8917cfcad0e9dc775118b19b/tpu_inference/layers/common/process_weights/moe_weights.py#L347) for w13, this PR applies same processing for jax-path (unquantized)

# Tests

```
MODEL_IMPL_TYPE=flax_nnx vllm serve Qwen/Qwen3-30B-A3B-Instruct-2507 \
  --max-model-len=2048 \
  --no-enable-prefix-caching
```
then send query with curl, response looks good

# Checklist


tpu_inference/layers/common/quantization/unquantized.py
```diff
@@ -59,3 +66,35 @@ def _apply_split(
             outs.append(out)
         out = jnp.concatenate(outs, axis=-1)
         return out
+
+
+@jax.jit(static_argnames=('mesh', 'activation', 'moe_backend'))
+def process_unquantized_moe_weights(
```
kyuyeunk, line 72: i think putting this logic in `unquantized.py` file makes sense.

can you apply a similar change to `process_fp8_moe_weights` and move it to `fp8.py`?

tpu_inference/layers/common/quantization/unquantized.py
```diff
@@ -59,3 +66,35 @@ def _apply_split(
...
+    *,
+    mesh: Mesh,
+    moe_backend: MoEBackend,
+    activation: MoEActivation,
+    w13_weight: jax.Array,
```
kyuyeunk, line 77: use `FusedMoEWeights` data structure like in here? https://github.com/vllm-project/tpu-inference/blob/2f7640057816392d8917cfcad0e9dc775118b19b/tpu_inference/layers/common/process_weights/moe_weights.py#L502

tpu_inference/layers/common/quantization/unquantized.py
lk-chen: Intended to make this PR only for unquantized model, FP8 change will come in another PR

tpu_inference/layers/common/quantization/unquantized.py
lk-chen: err maybe not? `FusedMoEWeights` is used on line 89 in this file

The purpose of `process_unquantized_moe_weights` was (I thought) to have a dedicated interface for unquantized model. Specifically, there's no "_scale" parameter. Using `FusedMoEWeights` is against this purpose, and in that case why not just jit `process_moe_weights` ?

tpu_inference/layers/common/quantization/unquantized.py
lk-chen: realized we are already doing that https://github.com/vllm-project/tpu-inference/blob/99a7ce458ebdfc763d457234300a2937549d8c4c/tpu_inference/layers/jax/quantization/fp8.py#L534

PR #2024, opened 2026-03-25
feat(torchax, mm): add env for jit some submodule in mm model for bet…

yuyanpeng-google:
# Description

feat(torchax, mm): add env for jit some submodule in mm model for bet…

Since the embed_multimodal include the preprocessing, which cannot and should not be jitted directly,
add envs to give a chance to jit
the module in the mm.

- JITTED_MM_MODULE_KEYS for the module to jit
- REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES for the custom pytree class.

e.g.,
```
export JITTED_MM_MODULE_KEYS="model.vision_tower.encoder"
export REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES="transformers.modeling_outputs.BaseModelOutputWithPast"
```

# Tests

Run the mm model using vllm implementation. See the module specified are been jitted.

# Checklist

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -38,6 +38,8 @@ if TYPE_CHECKING:
     FORCE_MOE_RANDOM_ROUTING: bool = False
     SC_KERNEL_THRESHOLD: int = 16777216
     SC_KERNEL_COL_CHUNK_SIZE: int = 1024
+    JITTED_MM_MODULE_KEYS: list[str] = []
+    REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES: list[str] = []
 
 
 def env_with_choices(
@@ -113,6 +115,25 @@ def env_bool(env_name: str, default: bool = False) -> Callable[[], bool]:
                 f"Valid options: '0', '1', 'true', 'false', 'True', 'False'.")
 
     return _get_bool_env
+
+
+def env_str_list(env_name: str) -> Callable[[], list[str]]:
+    """
+    Accepts a comma-separated string and returns a list of strings.
+
+    Args:
+        env_name: Name of the environment variable
+        default: Default list of strings if not set
+    """
+
+    def _get_str_list_env() -> list[str]:
+        value = os.getenv(env_name)
+        if value is None or value == "":
+            return []
+
+        return [v.strip() for v in value.split(",")]
+
+    return _get_str_list_env
 
 
 environment_variables: dict[str, Callable[[], Any]] = {
@@ -214,6 +235,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: int(os.getenv("SC_KERNEL_THRESHOLD") or "16777216"),
     "SC_KERNEL_COL_CHUNK_SIZE":
     lambda: int(os.getenv("SC_KERNEL_COL_CHUNK_SIZE") or "3072"),
+    "JITTED_MM_MODULE_KEYS":
+    env_str_list("JITTED_MM_MODULE_KEYS"),
+    "REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES":
+    env_str_list("REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES"),
 }
 
 
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -43,6 +43,7 @@ from vllm.sequence import IntermediateTensors
 from vllm.v1.outputs import PoolerOutput
 from vllm.v1.pool.metadata import PoolingMetadata
 
+from tpu_inference import envs
 from tpu_inference.distributed.jax_parallel_state import \
     get_pp_group as jax_get_pp_group
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
@@ -55,6 +56,7 @@ from tpu_inference.logger import init_logger
 from tpu_inference.models.common.interface import PoolerFunc
 from tpu_inference.models.jax.jax_intermediate_tensor import \
     JaxIntermediateTensors
+from tpu_inference.models.vllm.experimental.model_patcher import patch_mm_model
 from tpu_inference.models.vllm.vllm_model_wrapper_context import (
     get_vllm_model_wrapper_context, set_vllm_model_wrapper_context)
 from tpu_inference.runner.lora_utils import replace_lora_metadata
@@ -236,6 +238,19 @@ class VllmModelWrapper:
         params_and_buffers = shard_model_to_tpu(self.model, self.mesh)
 
         self._pooler: Pooler | None = self.model.pooler
+
+        if self.vllm_config.model_config.is_multimodal_model:
+            # NOTE: It patch mm models to be JITtable within some submodule.
+            # Caution: the submodule params_and_buffers would be put into
+            # the wrapper directly. params_and_buffers should be sharded to tpu
+            # and would not be used in the function args.
+            self.model, params_and_buffers = patch_mm_model(
+                self.model,
+                params_and_buffers,
+                jitted_mm_module_keys=envs.JITTED_MM_MODULE_KEYS,
+                register_mm_module_custom_pytree_classes=envs.
+                REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES,
+            )
 
         loading_end = time.time()
         total_loading_time = loading_end - loading_start
--- a/tpu_inference/models/vllm/experimental/__init__.py
+++ b/tpu_inference/models/vllm/experimental/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/tpu_inference/models/vllm/experimental/model_patcher.py
+++ b/tpu_inference/models/vllm/experimental/model_patcher.py
@@ -1,0 +1,142 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Patch vllm models to be TPU friendly or JITtable.
+
+Currently, it provides a way to JIT compile specific submodules of a
+multimodal model using `torchax.interop.JittableModule`. This is useful
+for offloading heavy vision or audio encoders to an Jitted graph while keeping
+the rest of the model in eager mode.
+
+"""
+
+import functools
+import importlib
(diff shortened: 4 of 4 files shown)


wdhongtw (commented):
LGTM, most issues are nit picking.

But I'm not confident with this direction, may need some other reviewer approval.

wdhongtw (changes requested):

tpu_inference/envs.py
```diff
@@ -204,6 +226,10 @@ def _get_bool_env() -> bool:
     lambda: int(os.getenv("JAX_PROFILER_SERVER_PORT") or "9999"),
     "USE_BATCHED_RPA_KERNEL":
     env_bool("USE_BATCHED_RPA_KERNEL"),
+    "JIT_MM_MODULE_KEYS":
```
wdhongtw: I'am afraid introducing permanent environment variable that are coupled with change implementation details.

These should be noted strongly that they are experimental interface and likely to be changed.

tpu_inference/envs.py
```diff
@@ -112,6 +114,26 @@ def _get_bool_env() -> bool:
     return _get_bool_env
 
 
+def env_str_list(env_name: str,
+                 default: list[str] = []) -> Callable[[], list[str]]:
```
wdhongtw: nit: Although we use `default.copy()` to avoid bug here, I thinks simply remove `default` arg (and return empty list) can cover the use cases.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
...
+                    jitted_module.params[inner_k] = torch_view(v)
+                elif inner_k in jitted_module.buffers:
+                    jitted_module.buffers[inner_k] = torch_view(v)
+                else:
+                    raise ValueError(f"Unexpect key {k} in params_and_buffers")
+            else:
+                new_params_and_buffers[k] = v
+        params_and_buffers = new_params_and_buffers
+
+    return model, params_and_buffers
```
wdhongtw: nit: can just return new_params_and_buffers

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
...
+        jitted_module = JittableModule(getattr(cur_module, target_module))
+        setattr(cur_module, target_module, jitted_module)
+
+        # params_and_buffers is a dict. for each key with prefix of the module,
+        # add `._model` to the prefix since JittableModule wrapper put origin into _model.
+        new_params_and_buffers = {}
+        prefix = f"vllm_model.{'.'.join(module_names[1:])}"
+        new_prefix = f"{prefix}._model"
+        for k, v in params_and_buffers.items():
+            if k.startswith(prefix):
```
wdhongtw: nit: maybe invert the condition and reduce one level of indent.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
...
+        if len(module_names) <= 1:
+            raise ValueError(
+                f"jit submodule only, but not the whole model. get {module_names=}"
+            )
+
+        # Start from 1 because 'model' is our root
+        for name in module_names[1:-1]:
+            cur_module = getattr(cur_module, name)
+
+        target_module = module_names[-1]
```
wdhongtw: nit: target_module_name?

target_module look like same type as cur_module and jitted_module.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
...
+        The patched model and the updated parameters and buffers.
+    """
+    if not jit_mm_module_keys:
+        return model, params_and_buffers
+
+    # Flatten custom pytree to jax for jit
+    # eg. transformers.modeling_outputs.BaseModelOutputWithPast
+    def _flatten_model_output(obj):
+        return obj.to_tuple(), type(obj)
+
```
wdhongtw: This part is still depends on the behavior of `ModelOutput` and `@dataclass`,
so `REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES` setting seems not working in all cases.

Maybe let the register logic be more specific and remove`REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES`, or give a more specific name for the environment variable?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
...
+    Caution: the submodule params_and_buffers would be put into
+    the wrapper directly. params_and_buffers should be sharded to tpu
+    and would not be used in the function args.
+
+    Args:
+        model: The vLLM runner model.
+        params_and_buffers: The parameters and buffers of the model.
+        jit_mm_module_keys: A list of module paths (e.g., "model.vision_tower.encoder") 
+          to be wrapped with `torchax.interop.JittableModule` for JAX JIT compilation.
+        register_mm_module_custom_pytree_classes: A list of fully qualified class names
```
wdhongtw: nit: will be good to document the limitation of being `ModelOutput` and `@dataclass`.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
 logger = init_logger(__name__)
 
 
+def _patch_mm_model(
+    model: "_VllmRunner",
+    params_and_buffers: dict[str, torchax.torch.Tensor],
+    jit_mm_module_keys: Sequence[str],
+    register_mm_module_custom_pytree_classes: Sequence[str],
+) -> tuple["_VllmRunner", dict[str, torchax.torch.Tensor]]:
+    """Jit some modules in the multimodal,
```
wdhongtw: nit: periods at the end for first line.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,99 @@
 logger = init_logger(__name__)
 
 
+def _patch_mm_model(
+    model: "_VllmRunner",
+    params_and_buffers: dict[str, torchax.torch.Tensor],
+    jit_mm_module_keys: Sequence[str],
```
wdhongtw: nit: the variable names does not looks like nouns.

wdhongtw (changes requested):

tpu_inference/envs.py
yuyanpeng-google: Agree. @kyuyeunk could you please suggest where to label "experimental feature" to users, how any more suitable place for it? Thanks.

yuyanpeng-google: `--additional-config` at vllm args could be more suitable. But it still need a place for the document.

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: I would like to keep the origin approach. The args in/out pattern should indicate the ownership are transferred and could be modified inplace as local variables.

Add more explanation in the docstring.

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: I don't think there is a easy way to provide a general way to register all cases. Even the `BaseModelOutputWithPast` may be fail if it has cache. 

It provides a chance to register naive class like `@dataclass` of all `torch.Tensor` without modify the codes. 

I prefer not to register a specific class if it is not used. It depend on each model implementation and the submodule to jit.

tpu_inference/envs.py
yuyanpeng-google: remove the default

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: add early return

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: fix

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: add more explanation

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: fix

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: change the name to jitted_mm_module_keys

tpu_inference/models/vllm/vllm_model_wrapper.py
wdhongtw: Understood, let's go with this limitation.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,104 @@
...
+    if not jitted_mm_module_keys:
+        return model, params_and_buffers
+
+    # Flatten custom pytree to jax for jit
+    # eg. transformers.modeling_outputs.BaseModelOutputWithPast
+    def _flatten_model_output(obj):
+        return obj.to_tuple(), type(obj)
+
+    def _unflatten_model_output(aux, children):
+        return aux(*children)
```
yuyanpeng-google: The flatten and unflatten are not correct. The obj.to_tuple() from https://github.com/huggingface/transformers/blob/main/src/transformers/utils/generic.py#L464 would not return None elements, and the unflatten would be wrong

yuyanpeng-google: Fix the issue and change the flatten, unflatten to ModelOutput like.

tpu_inference/envs.py
kyuyeunk, line 144: doesn't this `allow_csv` already meet this need?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -62,6 +64,107 @@
 logger = init_logger(__name__)
 
 
+def _patch_mm_model(
```
kyuyeunk: this is getting too big. can you create a separate multi modal related files instead of putting everything in this file?

tpu_inference/envs.py
yuyanpeng-google: The new env could be a list, and I thought env_with_choices is for one value?

tpu_inference/envs.py
kyuyeunk: i think env is fine. for instance, we use the same flag for `USE_BATCHED_RPA_KERNEL` which we've denoted as experiemtnal: https://github.com/vllm-project/tpu-inference/blob/2f7640057816392d8917cfcad0e9dc775118b19b/tpu_inference/kernels/experimental/batched_rpa/__init__.py#L16

but maybe printing a really big logger warning might be a good compromise?

tpu_inference/envs.py
kyuyeunk: ah okay, so `env_with_choices(allow_csv=True)` will fail if the string is not in the list right? okay, sgtm.

yuyanpeng-google: Yes. We cannot define the choices for general purposes since each module may have its own name; also, we want to preserve flexibility by using commas to separate multiple items needing to be jitted or registered.

tpu_inference/envs.py
yuyanpeng-google: Sounds good. Let me add an experimental folder and add some logger warning.

yuyanpeng-google: Move the patch_mm_model to `tpu_inference/models/vllm/experimental/model_patcher.py` and add logger.warn

tpu_inference/models/vllm/vllm_model_wrapper.py
yuyanpeng-google: Move the patch_mm_model to `tpu_inference/models/vllm/experimental/model_patcher.py`

wdhongtw (approved):
LGTM

kyuyeunk (approved):

dgouju:
Can you document (or even maybe automate?) how to identify `JITTED_MM_MODULE_KEYS` and `REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES` values for a given model? Otherwise it sounds complicated to use?
Thank you!

PR #2028, opened 2026-03-25
[FP8][MoE] Move FP8 MoE weight requantization from CPU to TPU with cache clearing

rohan-reddy:
## Summary

Moves FP8 MoE weight requantization from CPU to TPU by default, with a fix for the HBM regression that caused the original feature to be reverted.

**Root cause of the revert:** @jax.jit-compiled requantization programs reserve HBM (bytes_reserved) for XLA program buffers. These reservations persisted after weight loading completed, permanently consuming HBM that would otherwise be allocated to KV cache.

**Fix:** Call jax.clear_caches() in tpu_runner.py after model loading completes. This frees XLA program reservations from the requantization JIT, since those compiled programs are never needed again during inference.

**Update (v2):** Removed the MOE_REQUANTIZE_ON_TPU env flag and CPU fallback path. TPU requantization is now always enabled, no branching logic, no config needed.

## Changes

1. `moe_weights.py`: TPU requantization path using shard_map + lax.scan (replaces the old CPU dequant/quantize path)
2. `jax/quantization/fp8.py`: Shard FP8 weights to TPU before calling process_fp8_moe_weights
3. `vllm/quantization/fp8.py`: Same for vLLM code path
4. `tpu_runner.py`: Unconditionally call jax.clear_caches() after weight loading

### Memory metrics

| Metric | Main branch | With TPU requant + clear |
|---|---|---|
| HBM/chip | 4.46 GiB | 4.46 GiB |
| KV cache tokens | 1,033,600 | 1,033,664 (+64) |

Zero HBM regression with clear_caches.

### Loading time (warm-start)

| Config | Time |
|---|---|
| Main (CPU requant) | 25.30s |
| TPU requant + clear | ~18-19s |

~25% loading speedup.

### Test results

299 passed, 52 skipped, 0 failures (jax fp8, jax moe, vllm moe, vllm fp8 test suites).

## Notes

- jax.clear_caches() clears all JIT compilation caches. At the call site (after weight loading, before inference compilation), only loading-related JIT artifacts exist. Inference functions get compiled on first use regardless.
- Large model validation (Qwen3-235B, DeepSeek-R1) requires v7x access. Qwen3-235B OOMs on v6e-8 on main branch (pre-existing).

### DIFF
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -17,6 +17,7 @@ from functools import partial
 import jax
 import jax.numpy as jnp
 from jax.experimental.layout import Layout, with_layout_constraint
+from jax.experimental.shard_map import shard_map
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.tensor import Tensor
 
@@ -412,15 +413,21 @@ def process_moe_weights(
     )
 
 
-def shard_moe_weights(
+def _get_moe_weight_shardings(
     weights: FusedMoEWeights,
     moe_backend: MoEBackend,
     mesh: Mesh,
 ) -> FusedMoEWeights:
+    """Build sharding specs for MoE weights based on the backend type.
+
+    Returns a FusedMoEWeights where each field is a NamedSharding.
+    Used by both shard_moe_weights (for device_put) and
+    process_fp8_moe_weights (for sharding constraints inside JIT).
+    """
     match moe_backend:
         case MoEBackend.FUSED_MOE | MoEBackend.GMM_EP:
             ep_sharding = NamedSharding(mesh, P(ShardingAxisName.EXPERT))
-            weight_shardings = FusedMoEWeights(
+            return FusedMoEWeights(
                 w13_weight=ep_sharding,
                 w13_weight_scale=ep_sharding,
                 w13_bias=ep_sharding,
@@ -437,7 +444,7 @@ def shard_moe_weights(
                 w2_weight_scale_p_spec = P()
             else:
                 w2_weight_scale_p_spec = P(None, ShardingAxisName.MLP_TENSOR)
-            weight_shardings = FusedMoEWeights(
+            return FusedMoEWeights(
                 w13_weight=NamedSharding(
                     mesh,
                     P(None, None, ShardingAxisName.MLP_TENSOR),
@@ -462,6 +469,15 @@ def shard_moe_weights(
                     P(None, None, None),
                 ),  # (num_experts, 1, out_dim)
             )
+
+
+def shard_moe_weights(
+    weights: FusedMoEWeights,
+    moe_backend: MoEBackend,
+    mesh: Mesh,
+) -> FusedMoEWeights:
+
+    weight_shardings = _get_moe_weight_shardings(weights, moe_backend, mesh)
 
     match moe_backend:
         case MoEBackend.FUSED_MOE:
@@ -491,6 +507,56 @@ def shard_moe_weights(
             weight = general_device_put(weight, sharding, layout=layout)
             setattr(weights, key, weight)
     return weights
+
+
+def shard_fp8_moe_weights_to_tpu(
+    weights: FusedMoEWeights,
+    mesh: Mesh,
+    source_mesh: Mesh | None = None,
+) -> FusedMoEWeights:
+    """Shard FP8 MoE weights onto TPU before requantization.
+
+    Transfers FP8 weights from CPU to TPU with expert-dimension sharding
+    so that the subsequent dequant/requant in process_fp8_moe_weights runs
+    on TPU in parallel across experts. This avoids OOM (no single TPU holds
+    the full unsharded weight) and is much faster than CPU requantization.
+
+    For meshes without an EXPERT axis (e.g. GMM_TP), falls back to the
+    first mesh axis to distribute experts across devices.
+
+    Args:
+        weights: FP8 MoE weights (typically on CPU).
+        mesh: The TPU device mesh for inference.
+        source_mesh: The mesh the weights currently reside on (e.g.
+            cpu_mesh()). None when weights are plain CPU arrays.
+
+    Returns:
+        FusedMoEWeights sharded across TPU devices.
+    """
+    expert_axis = ShardingAxisName.EXPERT
+    if isinstance(expert_axis, str):
+        if expert_axis in mesh.axis_names:
+            shard_axis = expert_axis
+        else:
+            shard_axis = mesh.axis_names[0]
+    else:
+        if all(a in mesh.axis_names for a in expert_axis):
+            shard_axis = expert_axis
+        else:
+            shard_axis = mesh.axis_names[0]
+    ep_sharding = NamedSharding(mesh, P(shard_axis))
+
+    result_fields = {}
+    for field in fields(FusedMoEWeights):
+        key = field.name
+        weight = getattr(weights, key)
+        if weight is not None:
+            result_fields[key] = general_device_put(weight,
+                                                    ep_sharding,
+                                                    source_mesh=source_mesh)
+        else:
+            result_fields[key] = None
+    return FusedMoEWeights(**result_fields)
 
 
 @jax.jit(static_argnames=(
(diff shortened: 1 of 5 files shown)


kyuyeunk:
> jax.clear_caches() clears ALL JIT compilation caches. At the point where it is called (after weight loading, before inference compilation), only loading-related JIT artifacts exist. Inference functions get compiled on first use regardless.

this looks a bit scary? does it mean that we have to a compilation gets triggered every time a model is launched? even if it has been launched & compiled before?

rohan-reddy:
@kyuyeunk Good question, I checked this to see what the behavior is. `jax.clear_caches()` only clears the in-process Python JIT cache, not the on-disk XLA persistent compilation cache (`jax_compilation_cache_dir`).

Verified on Qwen3-30B-A3B-FP8 / v6e-8 with `MOE_REQUANTIZE_ON_TPU=1`:

| | Run 1 (cold cache) | Run 2 (warm cache) |
|---|---|---|
| Model load | 45.07s | 32.05s |
| First inference (XLA compile) | ~16s | ~5s |
| Total wall time | 97s | 83s |
| Cache files in `xla_cache/` | 3 (created) | 3 (reused) |
| HBM after init | 4.46 GiB/chip | 4.46 GiB/chip |
| KV cache blocks | 16,151 | 16,151 |

`jax.clear_caches()` ran after model loading in both runs. The persistent cache survived -- run 2 hit the cache (3x faster first inference, same 3 cache files, no new compilations). HBM and KV cache capacity are identical.

The persistent cache is configured automatically by `CompilationManager.__init__()` via `jax.config.update("jax_compilation_cache_dir", VLLM_XLA_CACHE_PATH)`, which defaults to `~/.cache/vllm/xla_cache`.

kyuyeunk:
thanks for the information!

note that we found jax/flax path to have the same high hbm usage issue and `jax.clear_cache()` trick fixed the issue! so thanks for sharing the insight: https://github.com/vllm-project/tpu-inference/pull/2073 cc: @Lumosis 

out of curiosity, do you know if the models that had oom during weight loading time in the previous version now works?

rohan-reddy:
@kyuyeunk Oh awesome! 
As far as the OOM during weight loading I think the one that was showing OOM previously was Deepseek R1 but I believe we did not test it again since the latest change I had made to further reduce some usage. 
I don’t have access to v7 to retest, but I would be curious to see the result too!

Also, I mostly was able to test vllm / torchax path in the recent test runs by the way, because I didn’t find a good model that defaults to / is supported on Jax path on v6e-8. Could you suggest one to test and I’ll run it and add to the description for parity?
Thanks!

kyuyeunk:
i have tested this feature out and confimed that it's working correctly. no extra hbm usage. in fact, i think it brings in more hbm thanks to the clearing?

@lk-chen, can you check if it works for deepseek as well? in general, i'm not fan of too much branching logics / configs. so if it doesn't have any negative impact for other uses cases, it would be great if we can make it the default for all the use cases.

lk-chen:
I'm hitting
```
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108]   File "/mnt/pd/envs/moe-env/lib/python3.12/site-packages/vllm/model_e
xecutor/layers/quantization/utils/quant_utils.py", line 327, in scaled_dequantize
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108]     x_s = prep_scale_for_group_broadcast(x_s, x_q, group_shape)
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108]           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108]   File "/mnt/pd/envs/moe-env/lib/python3.12/site-packages/vllm/model_e
xecutor/layers/quantization/utils/quant_utils.py", line 250, in prep_scale_for_group_broadcast
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108]     raise ValueError(
(EngineCore pid=3925343) ERROR 03-30 06:06:32 [core.py:1108] ValueError: 1D scale with shape torch.Size([32768]) cannot be broadcas
t to x with shape torch.Size([32768, 512]), group_shape=(128, 128)
```

when loading Deepseek on TPU v7x with
```
MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 MOE_REQUANTIZE_BLOCK_SIZE=512 NEW_MODEL_DESIGN=1 MODEL_IMPL_TYPE=vllm vllm serve deepseek-ai/DeepSeek-R1 --max-model-len=5120 --max-num-batched-tokens=1024 --max-num-seqs=192 --kv-cache-dtype=fp8 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --async-scheduling --enable-expert-parallel --additional_config='{"compilation_sizes": [1536], "sharding": {"sharding_strategy": {"enable_dp_attention": true}}}'
```

lk-chen:
interesting... hit another error
```
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]   File "/home/lkchen_google_com/worktree/moe/tpu_inference/layers/comm
on/fused_moe_gmm.py", line 392, in fused_moe_func
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]     x = tensor_parallel_gmm(
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]         ^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]   File "/home/lkchen_google_com/worktree/moe/tpu_inference/layers/comm
on/fused_moe_gmm.py", line 174, in tensor_parallel_gmm
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]     return jax.shard_map(
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]            ^^^^^^^^^^^^^^
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108] ValueError: shard_map applied to the function 'functools.partial(<func
tion moe_gmm_local at 0x79bd3becb100>, activation='silu', topk=8, parallelism='tp')' was given argument arrays with axis sizes that
 are not evenly divisible by the corresponding mesh axis sizes:
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108]
(EngineCore pid=3953468) ERROR 03-30 06:25:01 [core.py:1108] The mesh given has shape (1, 8, 1, 1, 1) with corresponding axis names
 ('data', 'attn_dp', 'attn_dp_expert', 'expert', 'model').
```

when loading jax-path (with `MODEL_IMPL_TYPE=flax_nnx`). But @rohan-reddy you should be able to reproduce, because jax-path accepts `--hf-overrides '{"num_hidden_layers": 5}'`, i.e.
```
MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 MOE_REQUANTIZE_BLOCK_SIZE=512 NEW_MODEL_DESIGN=1 MODEL_IMPL_TYPE=flax_nnx vllm serve deepseek-ai/DeepSeek-R1 --max-model-len=5120 --max-num-batched-tokens=1024 --max-num-seqs=192 --kv-cache-dtype=fp8 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --async-scheduling --enable-expert-parallel --additional_config='{"compilation_sizes": [1536], "sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' --hf-overrides '{"num_hidden_layers": 5}'
```

rohan-reddy:
@lk-chen Thanks for testing! I looked into both errors and believe they are pre-existing issues on main:

**Error 1 (vllm path - scale broadcasting):**
Lower confidence on this one since I could not reproduce it on our hardware (v6e-8 is too small for full DeepSeek R1, and --hf-overrides causes a different error before reaching MLA init). Based on code reading: the error is in MLA attention's process_weights_after_loading, not in MoE weight processing. get_and_maybe_dequant_weights passes the model-level weight_block_size=(128, 128) as group_shape to scaled_dequantize, but kv_b_proj uses per-channel FP8 quantization with a 1D scale [32768], not block quantization. prep_scale_for_group_broadcast can't broadcast a 1D scale with block-level group_shape. Seems this would affect any FP8-quantized DeepSeek model with MLA regardless of MoE requantization settings.

**Error 2 (jax path - shard_map):**
Reproduced this on v6e-8 on upstream main (synced, no PR changes) with --hf-overrides '{"num_hidden_layers": 5}'. tensor_parallel_gmm tries to shard w2_scale dim 1 (size = intermediate_size / block_size = 2048 / 512 = 4) across MLP_TENSOR axes (size 8 in dp_attention mode). 4 is not divisible by 8. Possible fix: replicate the scale dimension when num_blocks < tp_size instead of sharding it.

I can work on fixes for both in a separate PR if that would be helpful.

gpolovets1:
I'm in the process of adding an e2e test, so looking into the current breakages.

gpolovets1:
These commands look off for the Jax path. EP is not being triggered here and it also looks like the MLA kernel is turned off. Here is an updated command that should work:

```
VLLM_MLA_DISABLE=0 MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 MOE_REQUANTIZE_BLOCK_SIZE=512 NEW_MODEL_DESIGN=1 MODEL_IMPL_TYPE=flax_nnx vllm serve deepseek-ai/DeepSeek-R1 --max-model-len=5120 --max-num-batched-tokens=1024 --max-num-seqs=192 --kv-cache-dtype=fp8 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=1 --async-scheduling --additional_config='{"compilation_sizes": [1536], "replicate_attn_weights": "True",  "sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 8, "tensor_parallelism": 1}}}' --hf-overrides '{"num_hidden_layers": 5}'
```

lk-chen:
> These commands look off for the Jax path. EP is not being triggered here and it also looks like the MLA kernel is turned off. Here is an updated command that should work:
> 
> ```
> VLLM_MLA_DISABLE=0 MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 MOE_REQUANTIZE_BLOCK_SIZE=512 NEW_MODEL_DESIGN=1 MODEL_IMPL_TYPE=flax_nnx vllm serve deepseek-ai/DeepSeek-R1 --max-model-len=5120 --max-num-batched-tokens=1024 --max-num-seqs=192 --kv-cache-dtype=fp8 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=1 --async-scheduling --additional_config='{"compilation_sizes": [1536], "replicate_attn_weights": "True",  "sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 8, "tensor_parallelism": 1}}}' --hf-overrides '{"num_hidden_layers": 5}'
> ```

With this command, at commit 38a365d3ccd4f387eca701f6b5cd37612bd927cb, I tested DS-R1 load successfully on v7 @rohan-reddy

rohan-reddy:
Thanks for testing @lk-chen 
Curious what the runtime difference was roughly? And if there was any KV cache impact?

rohan-reddy:
@kyuyeunk Just want to confirm, given the results discussed, should I update this PR to eliminate the branching config and always requantize on TPU?

kyuyeunk:
sgtm! yeah, let's just enable this feature by default. not really a fan of too many options as it hurts maintainability.

rohan-reddy:
@kyuyeunk Awesome sounds good! Updated with no branching

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -530,34 +596,121 @@ def process_fp8_moe_weights(
...
+                w2_bias=None,
+            ),
+            moe_backend=moe_backend,
+            w13_reorder_size=w13_reorder_size,
+            w13_interleave=w13_interleave,
+        )
+        return (out.w13_weight, out.w13_weight_scale, out.w2_weight,
+                out.w2_weight_scale)
+
+    w13_q, w13_s, w2_q, w2_s = shard_map(
```
kyuyeunk, line 690: why use shard map? i prefer just keeping things as spmd when using jax unless it's stricly necessary.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -530,34 +596,121 @@ def process_fp8_moe_weights(
...
+    hidden_size = align_to(orig_hidden_size, w13_block_size)
+    intermediate_size = align_to(orig_intermediate_size, w2_block_size)
+    w13_pad = ((0, 2 * (intermediate_size - orig_intermediate_size)),
+               (0, hidden_size - orig_hidden_size))
+    w2_pad = ((0, hidden_size - orig_hidden_size),
+              (0, intermediate_size - orig_intermediate_size))
+
+    # Determine which mesh axis the expert dim is sharded across.
+    expert_axis = ShardingAxisName.EXPERT
+    if isinstance(expert_axis, str):
```
kyuyeunk, line 621: why is this logic needed?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -530,34 +596,121 @@ def process_fp8_moe_weights(
...
+            return carry, (w13_q, w13_s_new, w2_q, w2_s_new)
+
+        def _reshape_to_batches(x):
+            return x.reshape(n_batches, scan_batch_size, *x.shape[1:])
+
+        def _reshape_from_batches(x):
+            return x.reshape(n_local, *x.shape[2:])
+
+        xs = jax.tree.map(_reshape_to_batches, (w13, w13_scale, w2, w2_scale))
+        _, (w13_q, w13_s, w2_q, w2_s) = jax.lax.scan(_requant_expert_batch,
```
kyuyeunk, line 668: i think better approach is to keep existing `process_fp8_moe_weights` function as-is without any modification and create a wrapper function that invokes `process_fp8_moe_weights` using `jax.lax.scan`.

rohan-reddy: I originally had tried lax.scan only with keeping process_fp8_moe_weights unchanged, more details are [here](https://github.com/vllm-project/tpu-inference/pull/1842#issuecomment-4025141731). That version caused OOM on Deepseek R1 on v7 when Linkun tested whereas the current version succeeded. I believe shard map is needed to ensure the XLA compilation is based on the local dimension per shard and not global dimension (i.e. if there are 16 experts per shard, then the reservation for inputs should be based on that, not on the 128 experts across all 8 devices). I have some data about that [here](https://github.com/vllm-project/tpu-inference/pull/1842#issuecomment-4027992496).

tpu_inference/layers/common/process_weights/moe_weights.py
rohan-reddy: I added this to cover some edge cases, shard map needs to know which mesh axis to partition experts across. In GMM_EP there is an explicit EXPERT axis, in GMM_TP there isn't so we fall back to the first mesh axis.

tpu_inference/layers/common/process_weights/moe_weights.py
rohan-reddy: Added some details [in the below comment](https://github.com/vllm-project/tpu-inference/pull/2028#discussion_r3053185766)

jrplatin:
Since this is stale, following up here: https://github.com/vllm-project/tpu-inference/pull/2612

PR #2033, opened 2026-03-25
Add the missing TPU_VERSION var to env to tests

yiw-wang:
Add the TPU_VERSION as env var to the tests setup

The v6e and v7x now has separate image build workflow and artifacts now. Due to the fact that the TPU_VERSION is missing for these 5 tests, the v7x tests will try to pull the v6e(fallback) image and these tests become flaky:
- It may succeed when there is a v6e test run happened ahead for the same commit hash, even though it is using the v6e image for test.
- It may fail when the v7x test run happened first and there is no -v6e image

fixes b/496287779

# Tests

Using the CI tests to verify the test fix will work.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -170,6 +170,7 @@ steps:
         soft_fail: true
         env:
           USE_PREBUILT_IMAGE: "1"
+          TPU_VERSION: "${TPU_VERSION:-tpu6e}"
         agents:
           queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_queue}
         commands:
@@ -196,6 +197,7 @@ steps:
         soft_fail: true
         env:
           USE_PREBUILT_IMAGE: "1"
+          TPU_VERSION: "${TPU_VERSION:-tpu6e}"
         agents:
           queue: ${TPU_QUEUE_MULTI:-tpu_v6e_8_queue}
         commands:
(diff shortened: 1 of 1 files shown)


jcyang43:
Sorry I overlooked this yesterday: Why do we need separate pre-built images for v6e & v7x? We can use the same docker image on both v6e & v7x (unless the team recently introduced a change that requires different docker image for different TPU version). @yiw-wang & @QiliangCui, WDYT? 

---
I think we should still add the TPU version env to those test cases that are missing. But we can probably create another ticket/PR to use a single pre-built image for both v6e & v7x

yiw-wang:
Just FYI this version(add the TPU_VERSION as the global env var) won't work, from the failed tests we can see all tpu6e and tpu7x tests' TPU_VERSION is changed to tpu7x. 

So it turns out that, the BuildKite's state merge logic doesn't work as we expected. If the env var is set at the Step level, the env won't change by these two times "buildkite-agent pipeline upload". But if the env var is set at the Global level, the env will be merged and override, to tpu7x in our case.

Reverting to the original implementation.

yiw-wang:
> Sorry I overlooked this yesterday: Why do we need separate pre-built images for v6e & v7x? We can use the same docker image on both v6e & v7x (unless the team recently introduced a change that requires different docker image for different TPU version). @yiw-wang & @QiliangCui, WDYT?
> 
> I think we should still add the TPU version env to those test cases that are missing. But we can probably create another ticket/PR to use a single pre-built image for both v6e & v7x

A single image build is the original plan, but @QiliangCui raised a long term concern that the v6e and v7x may need a different image/lib version in the future, even this is not the case now.

Since this is not related to this PR, we can sync on the chat.

yiw-wang:
Closing this PR in favor of a clean recreation. We suspect some hidden CI state or linter cache was poisoned by the file diff history here. Will open a fresh PR with the exact same changes.

PR #2034, opened 2026-03-25
[Kernel][GMM] Clean up

kyuyeunk:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -14,6 +14,7 @@
 
 import dataclasses
 import functools
+from abc import ABC, abstractmethod
 from typing import Any, Callable, Tuple
 
 import jax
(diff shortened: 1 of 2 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -877,24 +798,32 @@ def kernel_main(
             dims=cfgs.dims,
         )
 
+    (lhs_spec, rhs_spec), out_spec = generate_block_specs(metadata_ref, cfgs)
+
+    if cfgs.fuse_act is not None:
+        rhs_up_ref = jax.tree.map(lambda x: x.at[..., cfgs.out_size_n:],
+                                  rhs_ref)
+        rhs_ref = FusedWeightsRef(gate=rhs_ref, up=rhs_up_ref)
```
patrickji2014, line 806: a little bit confused here, suppose n = 128, the fusedWeightsRef.up.shape[-1] is 64 and fusedWeightsRef.gate.shape[-1] is 128. fusedWeightsRef.get_weight has shape[-1] == 192?

kyuyeunk: `gate.shape[-1]` and `up.shape[-1]` will always be the same. also, they have to be multiple of 128, so `up.shape[-1] = 64` is not possible. so `get_weight.shape[-1]` will be 256.

also, this pr adds `output_size_n` which will be the half of n value if fuse activation is enabled.

coolkp: They have same block shape, just from different offset (half of concatted weight shape and there is assertion for its divisibility)

patrickji2014: thanks.

PR #2044, opened 2026-03-26
transcedentals cost in cost estimate

coolkp:
# Description

Add transcendental ops to cost estimate and make gelu approximate explicit

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -15,7 +15,6 @@
 import dataclasses
 import functools
 from abc import ABC, abstractmethod
-from enum import StrEnum, auto
 from typing import Any, Callable, Tuple
 
 import jax
@@ -23,15 +22,6 @@ import jax.numpy as jnp
 from jax import lax
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
-
-# Enums.
-
-
-class ActivationFn(StrEnum):
-    SILU = auto()
-    GELU = auto()
-    SWIGLUOAI = auto()
-
 
 # Util.
 
@@ -49,7 +39,7 @@ def swigluoai(gate: jax.Array,
     return (up + 1.0) * glu
 
 
-def apply_act_fn(acc: jax.Array, fuse_act: ActivationFn | None):
+def apply_act_fn(acc: jax.Array, fuse_act: str | None):
     """Applies a fused activation function to the accumulator.
 
     This function is used when an activation function is fused with the matrix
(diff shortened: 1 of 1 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -988,11 +988,18 @@ def get_cost_estimate(cfgs: GmmConfigs):
     out_bytes = dims.size_m * cfgs.out_size_n * cfgs.out_dtype.itemsize
 
     total_bytes = lhs_bytes + rhs_bytes + out_bytes
-
+    transcendentals = 0
+    if cfgs.fuse_act is not None:
+        # gelu is 1 because approximate=true
+        transcendentals = dims.size_m * cfgs.out_size_n
```
kyuyeunk: very nit & won't change any functionality, but change to `transcendentals +=`

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -988,11 +988,27 @@ def get_cost_estimate(cfgs: GmmConfigs):
...
-
+    transcendentals = 0
+    if cfgs.fuse_act is not None:
+        # gelu is 1 because approximate=true (tanh) vs (1/x, exp)
+        transcendentals_multiplier = 1 if cfgs.fuse_act == "gelu" else 2
+        act_transcendentals = dims.size_m * cfgs.out_size_n
+        act_transcendentals *= transcendentals_multiplier
+        transcendentals += act_transcendentals
+        act_alu_flops = {
+            "silu": 5,  # -, +, /, *, *
```
kyuyeunk: especially since this string is used in other place as well, should we create a proper enum for it? https://github.com/vllm-project/tpu-inference/blob/6be21f9fc13de620b20ab97d0977b45790c06cb6/tpu_inference/kernels/megablox/gmm_v2.py#L42-L74

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -23,6 +24,18 @@
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
 
+# Enums.
+
+
+class ActivationFn(str, enum.Enum):
```
kyuyeunk: Use "StrEnum" instead.

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -23,6 +24,18 @@
...
 
+# Enums.
+
+
+class ActivationFn(StrEnum):
+    SILU = "silu"
+    GELU = "gelu"
+    SWIGLUOAI = "swigluoai"
+
+    def __str__(self) -> str:
```
kyuyeunk: you can remove `def __str__(self)` since you are using `StrEnum`. Also, you can just set `SILU = auto()` as well: https://docs.python.org/3/library/enum.html#enum.StrEnum

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -1133,7 +1169,7 @@ def gmm_v2(
     acc_dtype: jnp.dtype | None = None,
     maybe_quantize_lhs: bool = True,
     zero_initialize: bool = True,
-    fuse_act: str | None = None,
+    fuse_act: ActivationFn | str | None = None,
```
kyuyeunk: Nit: I think we can remove 'str' type check since 'ActivationFn' already encapsulates it.

PR #2046, opened 2026-03-26
[RPA] Add bidirectional attention support for multimodal

kwang3939:
# Description

For models required to use bidirectional attention in both vision encoder and language model: when image tokens are given as inputs to the language model, the square masking of images is applied to image tokens and needs to be calculated and added dynamically. This is expected to work with both causal and sliding_window cases.

Implementations:
1. The ranges of vision tokens of each request are calculated from `multimodal_manager.py` with `calc_mm_prefix_ranges` and passed into `attention_metadata` to pass into RPA kernel. 
2. In the RPA kernel, square masks are calculated and applied based on the locations of vision tokens. Noticed that the `start_bkv_idx` needs to be adjusted for sliding_window case, when the applied square masks is larger than the window size.
3. Two unit tests are added to check the correctness.

Referred to the changes in vLLM flex attention: https://github.com/Isotr0py/vllm/blob/b71fbd06e215a1a09600220c947a1bb2d5494de9/vllm/v1/attention/backends/flex_attention.py#L467. The `mm_prefix_range` is created as `dict[int, list[tuple[int, int]]]` in vLLM. 
Due to the static requirement, `mm_prefix_range` is padded based on the upper bound of number of images per prompt, which is a required setting for testing multimodal models: `--limit-mm-per-prompt`. (This part can be further refined once having the support of video.)
The `is_mm_prefix` check and fallback can also be further adjusted based on the model configs.

# Tests

Tested with `blaze` for kernel unit tests.
Tested e2e with multimodal models.

# Checklist

### DIFF
--- a/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
+++ b/tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
@@ -80,6 +80,7 @@ def ref_ragged_paged_attention(
     skip_kv_mask: bool = False,
     sm_scale: float = 1.0,
     sliding_window: int | None = None,
+    mm_prefix_range: jax.Array | None = None,
     soft_cap: float | None = None,
     out_dtype: Any = None,
     mask_value: float | None = None,
@@ -192,6 +193,24 @@ def ref_ragged_paged_attention(
             mask = q_span >= kv_span
             if sliding_window is not None:
                 mask = jnp.logical_and(mask, q_span < kv_span + sliding_window)
+            if mm_prefix_range is not None:
+                max_mm_prefix_ranges = mm_prefix_range.shape[0] // (
+                    max_num_seqs * 2)
+                mm_mask = jnp.zeros(mask.shape, dtype=jnp.bool_)
+                for r in range(max_mm_prefix_ranges):
+                    start = mm_prefix_range[i * max_mm_prefix_ranges * 2 +
+                                            r * 2]
+                    end = mm_prefix_range[i * max_mm_prefix_ranges * 2 +
+                                          r * 2 + 1]
+                    is_valid = start != -1
+                    doc_mask_q = jnp.logical_and(start <= q_span, q_span
+                                                 <= end)
+                    doc_mask_kv = jnp.logical_and(start <= kv_span, kv_span
+                                                  <= end)
+                    current_mm_mask = jnp.logical_and(doc_mask_q, doc_mask_kv)
+                    mm_mask = jnp.logical_or(
+                        mm_mask, jnp.logical_and(is_valid, current_mm_mask))
+                mask = jnp.logical_or(mask, mm_mask)
             attn = jnp.where(mask, attn, mask_value)
         attn = jax.nn.softmax(attn, axis=-1).astype(v.dtype)
 
@@ -205,7 +224,9 @@ def ref_ragged_paged_attention(
     return result, kv_cache
 
 
-def get_smem_estimate_bytes(max_num_seqs, pages_per_seq):
+def get_smem_estimate_bytes(max_num_seqs,
+                            pages_per_seq,
+                            max_mm_prefix_ranges=0):
     total_bits = (
         # kv_lens_ref: i32[max_num_seqs]
         align_to(max_num_seqs, 128) * 32 +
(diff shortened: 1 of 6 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -425,6 +425,12 @@ def _init_phased_profiling(self) -> None:
     def _init_mm(self) -> None:
         self.is_multimodal_model = None
         self.uses_mrope = self.model_config.uses_mrope
+        mm_config = getattr(self.model_config, "multimodal_config", None)
```
lk-chen: nit, just `getattr(...., {})`
and next line: `.. = mm_config.get(.., {})`

kyuyeunk:
@bythew3i, can you review this rpa pr?

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -1569,6 +1642,7 @@ def ragged_paged_attention(
     page_indices: jax.Array,  # i32[max_num_seqs * pages_per_seq]
     cu_q_lens: jax.Array,  # i32[max_num_seqs + 1]
     distribution: jax.Array,  # i32[3]
+    mm_prefix_range: jax.Array | None = None,
```
bythew3i: Can you please make it keyword argument instead of positional argument? (eg., putting after sliding_window param?) Also please update other signatures! Thanks!

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
bythew3i, line 1717: Please add a comment to explain what is `mm_prefix_range`

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -192,6 +193,18 @@ def ref_ragged_paged_attention(
             mask = q_span >= kv_span
             if sliding_window is not None:
                 mask = jnp.logical_and(mask, q_span < kv_span + sliding_window)
+            if mm_prefix_range is not None:
+                max_mm_prefix_ranges = mm_prefix_range.shape[1]
+                mm_mask = jnp.zeros(mask.shape, dtype=jnp.bool_)
+                for r in range(max_mm_prefix_ranges):
+                    start = mm_prefix_range[i, r, 0]
```
bythew3i: It seems the start is always the LBD and end is always the UBD, why not just find the max of all starts and min of all ends before calling the kernel?

So the mm_range could just be in shape (max_num_seqs, 2)...and then flatten to 1d to SMEM (shape in (2 * max_num_seqs, )

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -299,6 +314,7 @@ def _ragged_paged_attention_kernel_loop(
     sem_ids_ref,  # [3] (bq_sem_idx, bkv_sem_idx, bo_sem_idx)
     bo_ids_ref,  # [4] (bo_sem_0_seq_idx, bo_sem_1_seq_idx, bo_sem_0_bo_idx, bo_sem_1_bo_idx)
     bkv_update_ids_ref,  # [6] (bkv_sem_0_seq_idx, bkv_sem_1_seq_idx, bkv_sem_0_offset, bkv_sem_1_offset, bkv_sem_0_sz, bkv_sem_1_sz)
+    mm_prefix_range_ref, # [3] (max_num_seqs, max_mm_prefix_ranges, 2)
```
bythew3i: the last dim will be padded to 128 so it will waste a lot of SMEM (given SMEM in total has 1MB).... please flatten it to 1D to maximize the SMEM utilization.

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
kwang3939: Hi @bythew3i , thanks for the review!

The reason I didn't use the global min and max is to support multiple images per sequence. If we have disjoint multiple images (e.g., text -> image 1 -> text -> image 2 -> text), we need to apply two distinct square masks (possibly in different sizes) just on vision tokens instead of one big square covering vision tokens and the text tokens in between. 
I will update and flatten it to 1D (shape in (max_num_seqs * max_mm_prefix_ranges * 2,).

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
kwang3939: Got it! I will flatten the `mm_prefix_range` to 1D: (max_num_seqs * max_mm_prefix_ranges * 2). Will let you know once I have updated and tested it.

kyuyeunk:
@kwang3939 is this ready for additional review?

kwang3939:
Hi @kyuyeunk, it's not ready yet. I'm oncall this week. Will back work on it soon!

PR #2059, opened 2026-03-27
feat(tpu-inference): add fused_moe gather-add checkin for E2E perf

clee1994:
# Description
This PR introduces a Sparse Core (SC) kernel for MoE gather-reduce operations and provides runtime configurability via centralized environment variables. It replaces the standard JAX-based indexing and reduction logic with a low-level MLIR-based SC kernel to improve end-to-end performance on TPU when handling large token batches (especially useful since it will free up the TensorCore).

# Relevant Details and Context
- Why this change is being made: In Mixture-of-Experts (MoE) models, the final local reduction step—which involves gathering expert outputs based on routing indices and applying top-k weights and locally reducing — is usually split up into a gather on SparseCore and logic on TC which requires 4 HBM access (gather load, gather write, logic load and logic write). With a SC kernel we leverage the random memory access of the SC and its compute capabilities. The standard Jax approach can become a performance bottleneck especially as the number of tokens increases.
- The problem being solved: Standard JAX operations for non-contiguous "gathers" and subsequent weighted reductions can lead to high HBM pressure and suboptimal execution times on TPU.
- Why this is a good solution: By offloading these operations to the specialized Sparse Core (SC) hardware, we can execute the gather and weighted summation much more efficiently (overlapping memory access and compute on SC  - double buffer). This implementation provides a specialized sc_gather_reduce kernel that leverages low-level MLIR for maximum performance.
- Implementation details:
  - New Kernel: Added a comprehensive MLIR-based implementation of the sc_gather_reduce kernel in tpu_inference/kernels/sparse_core/gather_reduce.py.
  - Integration: Updated moe_gmm_local in tpu_inference/layers/common/fused_moe_gmm.py to use the SC kernel when the token count exceeds a specified threshold.
  - Safe Configuration: Following recommended practices, we extended tpu_inference/envs.py to manage two new configuration parameters:
     - SC_KERNEL_THRESHOLD: Controls the token count threshold for kernel activation (default: 8192).
     - SC_KERNEL_COL_CHUNK_SIZE: Configures the column chunk size for kernel optimization (default: 3072).
   - Refinement: Updated the moe.py backend logic to pull these values from the envs module, ensuring a clean and typed interface for these parameters.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/kernels/gather/gather_reduce.py
+++ b/tpu_inference/kernels/gather/gather_reduce.py
@@ -16,29 +16,25 @@ import array
 from typing import Any
 
 import jax
-from jax import core
 import jax.experimental.pallas.tpu as pltpu
+import jax.numpy as jnp
+from jax import core
 from jax.experimental import mosaic
 from jax.experimental.mosaic.dialects import tpu
-import jax.numpy as jnp
 from jaxlib.mlir import ir
-from jaxlib.mlir.dialects import arith
-from jaxlib.mlir.dialects import func
-from jaxlib.mlir.dialects import memref
-from jaxlib.mlir.dialects import scf
-from jaxlib.mlir.dialects import vector
+from jaxlib.mlir.dialects import arith, func, memref, scf, vector
 
 
 class VectorTypeHelper:
-  """Helper to create VectorType with a specific element type."""
+    """Helper to create VectorType with a specific element type."""
 
-  def __init__(self, element_type_fn):
-    self.element_type_fn = element_type_fn
+    def __init__(self, element_type_fn):
+        self.element_type_fn = element_type_fn
 
-  def __getitem__(self, shape):
-    if isinstance(shape, int):
-      shape = [shape]
-    return ir.VectorType.get(shape, self.element_type_fn())
+    def __getitem__(self, shape):
+        if isinstance(shape, int):
+            shape = [shape]
+        return ir.VectorType.get(shape, self.element_type_fn())
 
 
 _I32 = VectorTypeHelper(lambda: ir.IntegerType.get_signless(32))
@@ -47,24 +43,23 @@ _BF16 = VectorTypeHelper(lambda: ir.BF16Type.get())
 
 
 def is_supported_by_sc_gather_reduce(x_shape: int,
-                                     sc_kernel_threshold: int) -> bool: 
-  if x_shape > sc_kernel_threshold and pltpu.get_tpu_info().generation == 7: 
-    return True
-  return False
+                                     sc_kernel_threshold: int) -> bool:
+    if x_shape > sc_kernel_threshold and pltpu.get_tpu_info().generation == 7:
+        return True
+    return False
 
-@jax.jit(
-    static_argnames=[
-        "reduce_group_size",
-        "single_sc",
-        "col_chunk_size",
-        "loop_unroll_factor_1",
-        "loop_unroll_factor_2",
-        "loop_unroll_factor_3",
-        "loop_parallel_access_1",
-        "loop_parallel_access_2",
-        "loop_parallel_access_3",
-    ],
-)
+
+@jax.jit(static_argnames=[
+    "reduce_group_size",
+    "single_sc",
+    "col_chunk_size",
+    "loop_unroll_factor_1",
+    "loop_unroll_factor_2",
+    "loop_unroll_factor_3",
+    "loop_parallel_access_1",
+    "loop_parallel_access_2",
+    "loop_parallel_access_3",
+], )
 def sc_gather_reduce(
     op: jax.Array,
     idx: jax.Array,
@@ -81,7 +76,7 @@ def sc_gather_reduce(
     loop_parallel_access_2: bool = True,
     loop_parallel_access_3: bool = False,
 ) -> jax.Array:
-  """Performs a gather-reduce operation on SparseCore.
+    """Performs a gather-reduce operation on SparseCore.
 
   This kernel groups rows of the operand `op` based on `idx`, sums them up,
   and scatters the results. The gather and add operations are performed in fp32,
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
 
-    token_hidden = token_topk_hidden.sum(axis=-2)
 
-    reduction_axis = (ShardingAxisName.MLP_TENSOR
-                      if parallelism == "tp" else ShardingAxisName.EXPERT)
-    # Then global reduction on all ranks for all tokens and all experts
-    return jax.lax.psum(token_hidden, axis_name=reduction_axis)
+    if gmm1_res.shape[0] > 8 * 1024:
+        gmm2_res = gmm_wrapper(gmm1_res, w2, w2_scale, w2_bias, group_sizes,
+                            group_offset, preferred_element_type=jnp.dtype(jnp.float32))
```
kyuyeunk: just use `jnp.float32`?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
+            )[topk_argsort_revert_indices].reshape(-1, topk)
+            topk_weights = jnp.where(mask, topk_weights, 0)
+
+        inds = topk_argsort_revert_indices
+        topk_weights = topk_weights.flatten().reshape(-1, 128)
+
+        token_hidden = gather_reduce_sc.sc_gather_reduce(
+            op=gmm2_res,
+            idx=inds,
+            reduce_group_size=8,
```
kyuyeunk: please make this an argument based on topk value

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
-        )[topk_argsort_revert_indices].reshape(-1, topk, 1)
-        token_topk_hidden = jnp.where(mask, token_topk_hidden, 0.0)
 
-    token_hidden = token_topk_hidden.sum(axis=-2)
 
-    reduction_axis = (ShardingAxisName.MLP_TENSOR
-                      if parallelism == "tp" else ShardingAxisName.EXPERT)
-    # Then global reduction on all ranks for all tokens and all experts
-    return jax.lax.psum(token_hidden, axis_name=reduction_axis)
+    if gmm1_res.shape[0] > 8 * 1024:
```
kyuyeunk: how high confidence do we have that this threshold is always safe? or do we want to make it a configurable value?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
+        inds = topk_argsort_revert_indices
+        topk_weights = topk_weights.flatten().reshape(-1, 128)
+
+        token_hidden = gather_reduce_sc.sc_gather_reduce(
+            op=gmm2_res,
+            idx=inds,
+            reduce_group_size=8,
+            single_sc=False,
+            topk_weights=topk_weights,
+            col_chunk_size=3072,
```
kyuyeunk: should this be tunable?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
+            topk_weights = jnp.where(mask, topk_weights, 0)
+
+        inds = topk_argsort_revert_indices
+        topk_weights = topk_weights.flatten().reshape(-1, 128)
+
+        token_hidden = gather_reduce_sc.sc_gather_reduce(
+            op=gmm2_res,
+            idx=inds,
+            reduce_group_size=8,
+            single_sc=False,
```
kyuyeunk: is there case when we want this to be `True` if not, maybe it's better to completely remove this argument?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +117,62 @@ def moe_gmm_local(
...
+            topk_weights=topk_weights,
+            col_chunk_size=3072,
+        )
+        reduction_axis = (ShardingAxisName.MLP_TENSOR
```
kyuyeunk: move this out of if else statement since this variable can be shared with logic from else statement as well.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: I tried that and it errored out at https://github.com/clee1994/tpu-inference/blob/fix-scatter-reduce/tpu_inference/kernels/megablox/gmm_v2.py#L978 (that it doesn't have itemsize attribute). but you can see two lines above they do the same (fp32_bytes = jnp.dtype(jnp.float32).itemsize).

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: done with latest commit.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: not very high, but made it configurable now.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: made it a parameter, should be a divisor of the 2nd dim of input matrix.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: if you want to do a split offloading with SC, e.g. SC1 does the kernel and SC2 does something else (weight all gather or so) it might be useful. but removed for now.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: done.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +119,58 @@ def moe_gmm_local(
...
+            )[topk_argsort_revert_indices].reshape(-1, topk, 1)
+            token_topk_hidden = jnp.where(mask, token_topk_hidden, 0.0)
+
+        token_hidden = token_topk_hidden.sum(axis=-2)
 
     reduction_axis = (ShardingAxisName.MLP_TENSOR
                       if parallelism == "tp" else ShardingAxisName.EXPERT)
     # Then global reduction on all ranks for all tokens and all experts
-    return jax.lax.psum(token_hidden, axis_name=reduction_axis)
+    return jax.lax.psum(token_hidden, axis_name=reduction_axis).astype(x.dtype)
```
kyuyeunk, line 179: this approach will use fp32 all reduce, but if we do this instead: `jax.lax.psum(token_hidden.astype(x.dtype), axis_name=reduction_axis)` we should perform bf16 allreduce and be faster right?

clee1994: token_hidden should be bf16 in both cases, e.g. either our kernel takes fp32 inputs and outputs bf16 (kernel supports bf16 and fp32 inputs but always outputs bf16) or the non kernel path only works with bf16.

clee1994:
just added unit tests.

tests/kernels/gather_reduce_test.py
```diff
@@ -0,0 +1,235 @@
...
+import jax.numpy as jnp
+import numpy as np
+
+from tpu_inference.kernels.sparse_core import gather_reduce as gather_reduce_sc
+try:
+    import xprof
+except ImportError:
+    xprof = None
+
+os.environ["LIBTPU_INIT_ARGS"] = "--xla_tpu_use_tc_device_shape_on_sc=true"
```
kyuyeunk: i would assume this must be enabled to use this kernel? If so, would it make sense that this is always enabled when using tpu-inference through this file? https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/env_override.py

hmmm. though i think this might make it always default even if user tries to disable it.

tpu_inference/layers/common/fused_moe_gmm.py
kyuyeunk: got it. i think that's probably a bug in gmmv2. i can take a look at it later, but in the meantime, i think `jnp.float32.dtype` probably looks more cleaner.

tpu_inference/envs.py
```diff
@@ -204,6 +206,10 @@ def _get_bool_env() -> bool:
     lambda: int(os.getenv("JAX_PROFILER_SERVER_PORT") or "9999"),
     "USE_BATCHED_RPA_KERNEL":
     env_bool("USE_BATCHED_RPA_KERNEL"),
+    "SC_KERNEL_THRESHOLD":
+    lambda: int(os.getenv("SC_KERNEL_THRESHOLD") or "8192"),
+    "SC_KERNEL_COL_CHUNK_SIZE":
```
kyuyeunk, line 211: how critical is this value? is it possible to create some basic default value or heuristic and get good performance in most cases?

kyuyeunk:
> just added unit tests.

triggered ci.

tpu_inference/kernels/gather/gather_reduce.py
kyuyeunk, line 1: nit: we also have other gather kernel in the works. that one is for permute while this one is for unpermute. would it make sense to just use `tpu_inference/kernels/gather/` folder to place both kernels? https://github.com/vllm-project/tpu-inference/pull/1994

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -51,7 +54,8 @@ def gmm_wrapper(lhs,
         group_sizes=group_sizes,
         group_offset=group_offset[0],
         zero_initialize=False,
-        fuse_act=fuse_act,
+        fuse_act=None,
```
kyuyeunk: i think this is a bug?

clee1994: fixed.

tpu_inference/kernels/gather/gather_reduce.py
clee1994: renamed folder to gather.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +119,58 @@ def moe_gmm_local(
...
+        gmm2_res = gmm_wrapper(gmm1_res, w2, w2_scale, w2_bias, group_sizes,
+                            group_offset, preferred_element_type=x.dtype)
+
+        # First run local reduction on topk experts owned by the rank for all tokens
+        token_topk_hidden = gmm2_res[topk_argsort_revert_indices].reshape(
+            (-1, topk, gmm2_res.shape[-1]))
+        token_topk_hidden = token_topk_hidden * jnp.expand_dims(topk_weights,
+                                                                axis=-1)
+
+        local_group_size = w1.shape[0]
```
kyuyeunk: `valid_rows_mask` logic is shared with both kernel and non-kernel path. can you move this out of if / else to reduce duplicate code?

clee1994: moved outside the if/else.

tpu_inference/envs.py
clee1994: i think the value is probably okay, but we could sweep it for an optimum.

kyuyeunk (approved):

tests/kernels/gather_reduce_test.py
clee1994: done.

tpu_inference/layers/common/fused_moe_gmm.py
clee1994: changed to jnp.float32.dtype

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -112,31 +119,55 @@ def moe_gmm_local(
...
+            gmm1_res.shape[0],
             group_sizes,
             group_offset,
             group_offset + local_group_size,
-        )[topk_argsort_revert_indices].reshape(-1, topk, 1)
-        token_topk_hidden = jnp.where(mask, token_topk_hidden, 0.0)
+        )[topk_argsort_revert_indices]
+
+
+    if gmm1_res.shape[0] > sc_kernel_threshold and jax.devices()[0].device_kind == 'TPU7x':
```
kyuyeunk: please use a logic like this & create the function inside the kernel: 
- https://github.com/vllm-project/tpu-inference/blob/879712df3be7fc9c6e1622acaad8ba09e94cb52a/tpu_inference/kernels/megablox/gmm_v2.py#L1051-L1055
- https://github.com/vllm-project/tpu-inference/blob/879712df3be7fc9c6e1622acaad8ba09e94cb52a/tpu_inference/layers/common/fused_moe_gmm.py#L66

also, instead of device kind, please use `pltpu.get_tpu_info()`

kyuyeunk: also, moving batch shape checking logic to that function would be a good idea s well.

clee1994: created a function used pltpu and also do the batch shape checking inside it.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -69,6 +74,11 @@ def valid_rows_mask(batch_size: int, group_sizes: jax.Array,
     return jnp.where(jnp.logical_and(token_start <= index, index < token_end),
                      True, False)
 
+def is_supported_by_sc_gather_reduce(x_shape: bool,
```
kyuyeunk: can you move this function inside of `tpu_inference/kernsl/gather/gather_reduce.py`?

also, type hints for `x_shape` ad `sc_kernel_threshold` is wrong.

clee1994: fixed and moved.

kyuyeunk:
Merge is failing because of formatting issue found by pre commit. Can you take a look?

clee1994:
ran pre-commit locally and got
```
addlicense...............................................................Passed
isort....................................................................Passed
yapf.....................................................................Passed
ruff.....................................................................Passed
clang-format.........................................(no files to check)Skipped
PyMarkdown...............................................................Passed
Lint GitHub Actions workflow files.......................................Passed
pip-compile..........................................(no files to check)Skipped
shellcheck...............................................................Passed
Detect missing __init__.py...............................................Passed
Check for spaces in all filenames........................................Passed
Suggestion...............................................................Passed
```

submitted a commit with all the changes.

kyuyeunk:
When running locally, by default, precommit only runs formatting check for files changed by last commit. You should tweak the argument to make it check for all the files in this pr.

PR #2062, opened 2026-03-27
[Benchmark] Update benchmark_serving.py: Support OpenAI Chat Completions API

wdhongtw:
# Description

Support OpenAI Chat Completions API, to enable more benchmark dataset that sending multimodality input.

- Generic the request handling through ASYNC_REQUEST_FUNCS
- Extract benchmark context that's identical across all requests
- Simplify null async semaphore usage
- Other refactoring

# Tests

N/A

# Checklist

### DIFF
--- a/scripts/vllm/benchmarking/backend_request_func.py
+++ b/scripts/vllm/benchmarking/backend_request_func.py
@@ -12,11 +12,11 @@ import os
 import sys
 import time
 import traceback
-from dataclasses import dataclass, field
-from typing import Literal, Optional, Union
+from typing import Any, Literal, Optional, Protocol, Union
 
 import aiohttp
 import huggingface_hub.constants
+from benchmark_core import BenchmarkContext, RequestFuncOutput, SampleRequest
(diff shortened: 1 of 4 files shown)


scripts/vllm/benchmarking/backend_request_func.py
```diff
@@ -72,52 +40,118 @@ async def start_stop_profile(base_url: str, action: Literal["start", "stop"]):
...
+    # XXX: not a good pattern to mutate the field after object creation.
+    output.input_request = request_func_input
+
+    if pbar:
+        pbar.update(1)
+
+    return output
+
+
+class _ExtractFunc(Protocol):
```
lk-chen, line 130: is it defined only to annotate the type of L140?

In that case you can use Callable type
```
from typing import Callable

...
  extract_from: Callable[[Any], str],
```

wdhongtw: Yes, but here I would like to add more information through docstring, as Any provides too little information here.

PR #2073, opened 2026-03-27
[jax] Fix memory leak to reduce peak ram usage when loading jax model.

Lumosis:
# Description

Tested with Qwen3-30b. Reduced the peak ram usage from 65 GB to 30 GB.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -779,12 +779,15 @@ def assign_and_shard_param(jax_param: nnx.Param,
     elif isinstance(spec, SingleDeviceSharding):
         spec = ()
     param_mesh = jax_param.get_metadata().get("mesh") or mesh
+    shape = jax_weight.shape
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/jax/quantization/fp8.py
```diff
@@ -293,6 +293,7 @@ def process_weights_after_loading(self, layer: JaxEinsum) -> bool:
             delattr(layer, 'weight')
             delattr(layer, 'weight_scale_inv')
             delattr(layer, 'bias')
+            jax.clear_caches()
```
lk-chen: The change in fp8.py and unquantized.py can be simplified by adding
```
jax.clear_caches()
```
right after [this line](https://github.com/vllm-project/tpu-inference/blob/c141a23a70b2a259b95acc62d188ba1b8843c71a/tpu_inference/models/jax/utils/weight_utils.py#L882)

PR #2075, opened 2026-03-28
Use Chat Completions for MMMU-Pro and align message format

wdhongtw:
# Description

Use Chat Completions for MMMU-Pro so we can send images.

Also align message format with other evaluation tool.

# Tests

Run the benchmark tool and verify it runs without problem.

```sh
python scripts/vllm/benchmarking/benchmark_serving.py --model <model-id-here> --dataset-name mmmu_pro --mmmu-pro-subset 'standard (10 options)' --mmmu-pro-output-len=8192 --run-eval --num-prompts 1730 --backend vllm-chat --endpoint /v1/chat/completions
```

# Checklist

### DIFF
--- a/scripts/vllm/benchmarking/benchmark_serving.py
+++ b/scripts/vllm/benchmarking/benchmark_serving.py
@@ -483,6 +483,13 @@ def main(args: argparse.Namespace):
         trust_remote_code=args.trust_remote_code,
     )
 
+    if args.dataset_name == "mmmu_pro":
+        message = (
+            "MMMU-Pro must use --backend vllm-chat "
+            "and should also use --endpoint=/v1/chat/completions."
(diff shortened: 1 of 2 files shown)


scripts/vllm/benchmarking/benchmark_serving.py
```diff
@@ -483,6 +483,9 @@ def main(args: argparse.Namespace):
         trust_remote_code=args.trust_remote_code,
     )
 
+    if args.dataset_name == "mmmu_pro":
+        assert args.backend == "vllm-chat"
```
lk-chen: maybe add actionable error message: "MMMU-Pro is multi-modal dataset, which requires using --endpoint=/v1/chat/completions"

PR #2082, opened 2026-03-30
 Replace jnp.clip(..., a_min=..., a_max=...) with jnp.clip(..., min=.…)

QiliangCui:
# Description
[JAX] Replace jnp.clip(..., a_min=..., a_max=...) with jnp.clip(..., min=..., max=...).

a_min and a_max are deprecated parameter names to jax.numpy.clip.


gpolovets1:
I believe there is another instance of this in: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/gmm_v2.py#L37

QiliangCui:
> I believe there is another instance of this in: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/gmm_v2.py#L37

Thank you @gpolovets1 !

 updated!

PR #2105, opened 2026-04-01
Add torchax support for EAGLE-3

Lumosis:
# Description

This PR introduces support for running EAGLE3 speculative decoding when using the TorchAX (vLLM) backend for the draft model (DRAFT_MODEL_IMPL_TYPE=vllm). Previously, draft model execution was hardcoded/tailored entirely around native JAX (flax_nnx) layers.

This updates the compilation manager, the Eagle3 proposer, and the VLLM model wrapper to ensure parameters natively required by TorchAX (such as layer_name_to_kvcache_index) are properly propagated down to the draft models.

# Key Changes
* VLLM Model Wrapper (vllm_model_wrapper.py):
  * Implemented jit_draft_step_func specifically to handle EAGLE3's draft execution path (model_fn) within TorchAX.
  * Added jit_combine_hidden_states_func to successfully bind the combine_hidden_states operation to torch.func.functional_call, which is required for EAGLE's proposer initialization.
* Compilation Manager (compilation_manager.py):
  * Updated draft_model_fn_wrapper parameters to correctly pass along layer_name_to_kvcache_index, ensuring KV caches align properly across layers when using partitioned VLLM draft deployments.
* EAGLE3 Drafter (eagle3.py):
  * Added explicit implementation-handling logic (envs.DRAFT_MODEL_IMPL_TYPE) to dynamically reuse the target model's embed_tokens.weight dictionary keys depending on if the active draft layout uses vllm backend structures or native flax_nnx Pytree layouts.

# Tests

To run torchax path eagle-3.
E.g:
```
MODEL_IMPL_TYPE=vllm DRAFT_MODEL_IMPL_TYPE=vllm vllm serve --model=meta-llama/Llama-3.3-70B-Instruct --max-model-len=8192 --no-enable-prefix-caching --tensor-parallel-size=2 --max-num-seqs=1 --speculative_config '{"model": "/mnt/disks/ranlihao/hub/models--yuhuili--EAGLE3-LLaMA3.3-Instruct-70B/snapshots/9659b9a88511bff6ff09878107fd998a0c08c3f3", "num_speculative_tokens": 2, "method": "eagle3", "draft_tensor_parallel_size": 1}' --no-async-scheduling
```

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -42,6 +42,8 @@ from vllm.model_executor.models.interfaces_base import is_pooling_model
 from vllm.sequence import IntermediateTensors
 from vllm.v1.outputs import PoolerOutput
 from vllm.v1.pool.metadata import PoolingMetadata
+from vllm.v1.worker.gpu.spec_decode.eagle.eagle3_utils import \
+    set_eagle3_aux_hidden_state_layers
 
 from tpu_inference import envs
 from tpu_inference.distributed.jax_parallel_state import \
@@ -112,13 +114,19 @@ class VllmModelWrapper:
     mesh: Mesh
     model: _VllmRunner
 
-    def __init__(self, vllm_config: VllmConfig, rng: PRNGKey, mesh: Mesh):
+    def __init__(self,
+                 vllm_config: VllmConfig,
+                 rng: PRNGKey,
+                 mesh: Mesh,
+                 is_draft_model: bool = False):
         self.vllm_config = vllm_config
         self.rng = rng
         self.mesh = mesh
+        self.is_draft_model = is_draft_model
 
         self.vllm_config.quant_config = get_tpu_quantization_config(
             self.vllm_config, self.mesh)
+        self.model_config = vllm_config.speculative_config.draft_model_config if is_draft_model else vllm_config.model_config
         self._apply_pp_patch()
         self._patch_vllm_ops()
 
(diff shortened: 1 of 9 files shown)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -231,8 +250,14 @@ def load_weights(self):
             replace_set_lora(vllm_model)
 
         static_forward_context = vllm_config_for_load.compilation_config.static_forward_context
-        self.vllm_config.compilation_config.static_forward_context = static_forward_context
-        self.vllm_config.compilation_config.static_all_moe_layers = vllm_config_for_load.compilation_config.static_all_moe_layers
+        self.vllm_config.compilation_config.static_forward_context.update(
```
gxd3, line 253: just for my learning, what does these 2 lines do, and why we need to change them?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -231,8 +250,14 @@ def load_weights(self):
...
+            static_forward_context)
+        self.vllm_config.compilation_config.static_all_moe_layers.extend(
+            vllm_config_for_load.compilation_config.static_all_moe_layers)
+
+        if self.vllm_config.speculative_config and self.vllm_config.speculative_config.method == "eagle3" and not self.is_draft_model:
```
gxd3, line 258: same here,  what does it do,?

gxd3 (approved):

tpu_inference/models/vllm/vllm_model_wrapper.py
Lumosis: The draft model of eagle-3 requires hidden states from target model as input. The line enables target model to output its hidden states so that we can feed the draft model.

tpu_inference/models/vllm/vllm_model_wrapper.py
Lumosis: The previous workflow is:

clear static_forward_context -> load target model (static_forward_context is updated while loading) -> clear static_forward_context ->  load draft model (static_forward_context for draft model is updated while loading).

If we don't store the static_forward_context for target model, errors will occur when we compile the target model.

tpu_inference/models/vllm/vllm_model_wrapper.py
gxd3: Make sense, thanks for the explanation!

tpu_inference/models/vllm/vllm_model_wrapper.py
gxd3: Thanks for the explanation!

PR #2115, opened 2026-04-02
[TPU] Support processed_logprobs Feature and Alignment with GPU

sierraisland:
# Description

This PR implements the `processed_logprobs` feature for the TPU, providing feature parity with the GPU implementation. Previously, TPU only supported extracting log-probabilities from raw logits, missing the ability to output logprobs that reflect sampling transformations.

Related bug: b/493265228

# Tests

Added TestProcessedLogprobs to verify `Temperature Scaling`, `Top-K/Top-P Masking`, `Greedy Fallback`.
Confirmed E2E output using offline_inference.py with Llama-3.2-1B-Instruct. Command: `VLLM_XLA_CHECK_RECOMPILATION=1 SKIP_JAX_PRECOMPILE=0 python examples/offline_inference.py --logprobs-mode processed_logprobs`

# Checklist

### DIFF
--- a/tests/layers/jax/sample/test_sampling.py
+++ b/tests/layers/jax/sample/test_sampling.py
@@ -148,8 +148,9 @@ class TestProcessedLogprobs:
     @staticmethod
     def _get_fake_mesh():
         """Create a fake mesh for testing purposes."""
-        devices = jax.devices("cpu")
-        mesh_shape = (1, )
+        devices = jax.devices()
+        num_devices = len(devices)
+        mesh_shape = (num_devices, )
         axis_names = (ShardingAxisName.ATTN_DATA, )
         device_mesh = mesh_utils.create_device_mesh(mesh_shape, devices)
         return Mesh(device_mesh, axis_names)


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -26,6 +26,43 @@
 _SAMPLING_EPS = 1e-5
 
 
+def _apply_sampling_transforms(
```
wenxindongwork, line 29: when returning processed log prob, this `_apply_sampling_transforms` function would be called twice, which is wasteful. An alternative that avoids calling this expensive function twice is to do this: 

1. return `processed_logits` in the `sample()` function. 
2.  if return processed log prob, call `compute_logprobs(processed_logits)`, else, call `compute_logprobs(raw_logits)`. 

This will also simply the code quite a bit. wdyt? 

E.g. 

```
  next_tokens, processed_logits = sample(
      step_rng,
      self.mesh,
      logits,
      tpu_sampling_metadata,
  )
```

wenxindongwork (approved):
LGTM, thank you so much!

PR #2118, opened 2026-04-02
Fix fork PR permissions

ylangtsou:
# Description

Fix the permission issue for forked PRs that caused a 403 error when updating the PR body.

# Tests
Tested and passed locally.

# Checklist

### DIFF
--- a/.github/workflows/pre-PR.yml
+++ b/.github/workflows/pre-PR.yml
@@ -82,7 +82,10 @@ jobs:
           fi
 
       - name: Update PR body
-        if: env.TEMPLATE_CONTENT != ''
+        # Ignore 403 errors for forked PRs (read-only GITHUB_TOKEN)
+        if: |
+          env.TEMPLATE_CONTENT != '' &&
+          github.event.pull_request.head.repo.full_name == github.repository
         uses: peter-evans/create-or-update-comment@v4
         with:
           issue-number: ${{ github.event.pull_request.number }}


.github/workflows/pre-PR.yml
```diff
@@ -15,19 +15,34 @@
 name: Apply Conditional PR Template
 
 on:
-  pull_request:
+  # Use pull_request_target instead of pull_request because:
+  # pull_request gives read-only GITHUB_TOKEN for fork PRs, causing 403 on PR body update.
+  # pull_request_target runs in base branch context, granting read/write GITHUB_TOKEN.
+  pull_request_target:
```
jcyang43: Please see comment below regarding `CHANGED_FILES`. Together with this `pull_request_target` change, there is a security risk

.github/workflows/pre-PR.yml
jcyang43, line 52: Previously, lines like this are fine because `pull_request` gives read-only token scope and has no access to repo secrets. 

With `pull_request_target` however, it will have access to repo secrets and the token scope is read/write. Hence, we need to be very careful with this kind of change. 

For example, an attacker can submit a PR with malicious bash characters in the file name, and places like this will be susceptible to command injection

jcyang43 (changes requested):

.github/workflows/pre-PR.yml
```diff
@@ -83,8 +83,29 @@ jobs:
...
+            echo "Created comment id '$COMMENT_ID' on issue '$ISSUE_NUMBER'."
+
+          elif [ "$HTTP_CODE" -eq 403 ]; then
+            echo "403 permission denied, ignoring for forked PRs"
+            exit 0
+
+          else
+            echo "Unexpected error: $HTTP_RESPONSE"
+            exit 1
+          fi
```
jcyang43: Hmm, can we just make use of `continue-on-error`? 

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error

---
Or better yet, use an if condition to only run the step for PRs from non forks. Something like [this](https://github.com/orgs/community/discussions/25217#discussioncomment-3246902) will do

ylangtsou: continue-on-error will pass all the error. Let me try the if condition

jcyang43 (approved):

PR #2137, opened 2026-04-03
[Kernel][Gather] Introduce sparse core ragged gather

kyuyeunk:
# Description

Introduce sparse core ragged gather kernel into moe layer.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/kernels/gather/ragged_gather.py
+++ b/tpu_inference/kernels/gather/ragged_gather.py
@@ -1,0 +1,289 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+from jax.experimental.pallas import tpu_sc as plsc
+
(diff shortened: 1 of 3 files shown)


tpu_inference/kernels/gather/ragged_gather.py
```diff
@@ -0,0 +1,289 @@
...
+        num_cols += 1
+    return pl.cdiv(hidden_size, (num_cols * num_lanes)) * num_lanes
+
+
+@jax.jit
+def ragged_gather(x: jax.Array, indices: jax.Array, start: jax.Array,
+                  end: jax.Array) -> jax.Array:
+    """Perform gather on indices within dynamic array start and end."""
+
+    assert x.ndim == 2
```
coolkp: nit: can you add assertion message here

coolkp (approved):

tpu_inference/kernels/gather/ragged_gather.py
kyuyeunk: done.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -279,6 +280,7 @@ def expert_parallel_gmm(
     num_experts_per_shard = num_experts // ep_size
     group_offset = jnp.arange(0, num_experts, num_experts_per_shard)
 
+    x_p_spec = P(ShardingAxisName.EXPERT_DATA)
```
gobbleturk, line 283: Why do we use this sharding? Now the tokens are sharded (likely on their wrong expert), don't we still have to all-gather later? I don't see where the all gather happens - unless it is in the ragged gather kernel

gobbleturk: Explanation: This sharding is used for the inputs of gmm matmuls AFTER the sorting (e.g. after ragged sort kernel) - in this case the activations should be sharded by expert. There are different shardings used below for the ragged sort kernel that have the inputs replicated and outputs sharded here https://github.com/vllm-project/tpu-inference/blob/6d98fff0f89f07ebe783c54bf5da117a01a12142/tpu_inference/layers/common/fused_moe_gmm.py#L532

PR #2144, opened 2026-04-05
Fix: Resolve MLA KV cache sharding and DeepSeek V3 shape mismatches

oliverlee209:
1. `tpu_inference/layers/common/attention_interface.py`:
   - Corrected the output sharding spec order in `_mla_ragged_paged_attention` to match the return order. This resolves the `IndivisibleError` caused by `jax.shard_map` applying the attention output sharding spec to the KV cache.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/attention_interface.py
+++ b/tpu_inference/layers/common/attention_interface.py
@@ -513,9 +513,9 @@ def mla_attention(
         P(ShardingAxisName.ATTN_DATA),  # md.distribution
     )
     out_specs = (
+        P(ShardingAxisName.MLP_TENSOR),  # kv cache
         attn_o_tnh_sharding
-        or P(ShardingAxisName.MLP_TENSOR, None, None),  # attn output
-        P(ShardingAxisName.MLP_TENSOR)  # kv cache
+        or P(ShardingAxisName.MLP_TENSOR, None, None)  # attn output
     )
 
     def _mla_ragged_paged_attention(q, q_rope, k, k_rope, cache, *args):


oliverlee209:
Hi team,

This PR addresses a few critical compilation and runtime errors (`IndivisibleError` and `TypeError`) encountered when running DeepSeek V3 (MLA architecture) on TPUs. 

Here are the key fixes:
1. `attention_interface.py`: Corrected the return tuple order in `_mla_ragged_paged_attention` to prevent `jax.shard_map` from applying the attention output's sharding spec to the KV cache matrix.
2. `model_loader.py`: Added dynamic `NamedSharding` logic for the KV cache when MLA is enabled.
3. `deepseek_v3.py`: Explicitly sliced `attn_output` and `mlp_output` to `[:seq_len, :]` to resolve shape mismatches caused by padded precompilation buckets.

Could a maintainer please help review? 

I'm happy to make any adjustments if needed. Thanks for your time!

gxd3:
> * This resolves the `IndivisibleError` caused by `jax.shard_map` applying the attention output sharding spec to the KV cache

Hi,
about `This resolves the IndivisibleError caused by jax.shard_map applying the attention output sharding spec to the KV cache`, could you share the command you're using to bring up DeepSeekV3 model?

oliverlee209:
> > * This resolves the `IndivisibleError` caused by `jax.shard_map` applying the attention output sharding spec to the KV cache
> 
> Hi, about `This resolves the IndivisibleError caused by jax.shard_map applying the attention output sharding spec to the KV cache`, could you share the command you're using to bring up DeepSeekV3 model?

Hi, 
The command is:
`vllm serve --model=${POD_DATA_DIR} --tensor-parallel-size 16 --max-model-len 4096 --max-num-seqs 512 --max-num-batched-tokens 4096 --no-enable-prefix-caching --kv-cache-dtype=fp8 --enable-expert-parallel --no-async-scheduling --gpu-memory-utilization 0.90`

where `${POD_DATA_DIR}` is the mount path to the DeepSeek V3 FP8 model weights. I wrapped this command in the Leader pod of a Kubernetes LeaderWorkerSet and set the environment variables `VLLM_MLA_DISABLE` to `false` and `MODEL_IMPL_TYPE` to `flax_nnx`.

tpu_inference/models/common/model_loader.py
```diff
@@ -301,10 +301,18 @@ def get_flax_model(
             vllm_config.model_config.hf_config)
     jit_model = _get_nnx_model(model_class, vllm_config, rng, mesh)
     vllm_config.model_config.dtype = original_dtype
-    kv_cache_sharding = NamedSharding(
-        mesh,
-        PartitionSpec(ShardingAxisName.ATTN_DATA, None,
-                      ShardingAxisName.ATTN_HEAD))
+
+    use_mla = getattr(vllm_config.model_config, "use_mla", False)
+    if use_mla:
```
gxd3: if using MLA, we'll need to use data parallelism for attention layer, which will require `NEW_MODEL_DESIGN=1` in your server startup command.

If so, this file seem not requiring changing.

tpu_inference/layers/common/attention_interface.py
gxd3: could we swap the order of this instead?

There are a few caller that reply on `new_kv_cache, outputs` being returned from `mla_attention()`,
e.g.
https://github.com/vllm-project/tpu-inference/blob/07072e1274a19a259840a21b39a4c4b098f778ac/tpu_inference/layers/vllm/backends/flash_attn_mla.py#L177

gxd3:
> > > * This resolves the `IndivisibleError` caused by `jax.shard_map` applying the attention output sharding spec to the KV cache
> > 
> > 
> > Hi, about `This resolves the IndivisibleError caused by jax.shard_map applying the attention output sharding spec to the KV cache`, could you share the command you're using to bring up DeepSeekV3 model?
> 
> Hi, The command is: `vllm serve --model=${POD_DATA_DIR} --tensor-parallel-size 16 --max-model-len 4096 --max-num-seqs 512 --max-num-batched-tokens 4096 --no-enable-prefix-caching --kv-cache-dtype=fp8 --enable-expert-parallel --no-async-scheduling --gpu-memory-utilization 0.90`
> 
> where `${POD_DATA_DIR}` is the mount path to the DeepSeek V3 FP8 model weights. I wrapped this command in the Leader pod of a Kubernetes LeaderWorkerSet and set the environment variables `VLLM_MLA_DISABLE` to `false` and `MODEL_IMPL_TYPE` to `flax_nnx`.

Thanks, replied with some comments. Seem that you'll like use enable MLA, if so, you probably need to add `NEW_MODEL_DESIGN=1` in the server startup command, and use data parallel for the attention layer.

tpu_inference/models/common/model_loader.py
oliverlee209: Yes, I also set `NEW_MODEL_DESIGN` to `1`, but  I encountered the error
`(EngineCore pid=1843) ERROR 04-04 08:28:46 [core.py:1108] jax._src.sharding.IndivisibleError: Sharding NamedSharding(mesh=Mesh('data': 1, 'attn_dp': 1, 'attn_dp_expert': 1, 'expert': 1, 'model': 16, axis_types=(Auto, Auto, Auto, Auto, Auto)), spec=P(('data', 'attn_dp', 'attn_dp_expert'), None, ('model', 'expert')), memory_kind=device) implies that array axis 2 is partitioned 16 times, but the dimension size is 4 (full shape: (1219, 256, 4, 640), per-dimension tiling factors: [1, 1, 16, 1] should evenly divide the shape)
`. 
There appears to be a need for a condition to apply the correct sharding when MLA is enabled.

oliverlee209:
> > > > * This resolves the `IndivisibleError` caused by `jax.shard_map` applying the attention output sharding spec to the KV cache
> > > 
> > > 
> > > Hi, about `This resolves the IndivisibleError caused by jax.shard_map applying the attention output sharding spec to the KV cache`, could you share the command you're using to bring up DeepSeekV3 model?
> > 
> > 
> > Hi, The command is: `vllm serve --model=${POD_DATA_DIR} --tensor-parallel-size 16 --max-model-len 4096 --max-num-seqs 512 --max-num-batched-tokens 4096 --no-enable-prefix-caching --kv-cache-dtype=fp8 --enable-expert-parallel --no-async-scheduling --gpu-memory-utilization 0.90`
> > where `${POD_DATA_DIR}` is the mount path to the DeepSeek V3 FP8 model weights. I wrapped this command in the Leader pod of a Kubernetes LeaderWorkerSet and set the environment variables `VLLM_MLA_DISABLE` to `false` and `MODEL_IMPL_TYPE` to `flax_nnx`.
> 
> Thanks, replied with some comments. Seem that you'll like use enable MLA, if so, you probably need to add `NEW_MODEL_DESIGN=1` in the server startup command, and use data parallel for the attention layer.

Sorry for not mentioning this earlier, but I also set `NEW_MODEL_DESIGN` to `1`.

tpu_inference/layers/common/attention_interface.py
oliverlee209: If we do it this way, do we also need to change the return order here?
https://github.com/vllm-project/tpu-inference/blob/844f9e7f49153209a2d16068598d0c3225ede935/tpu_inference/kernels/mla/v2/kernel.py#L1583

gxd3: https://github.com/vllm-project/tpu-inference/blob/844f9e7f49153209a2d16068598d0c3225ede935/tpu_inference/layers/common/attention_interface.py#L521-L540
The code here look right to me without requiring changes.

tpu_inference/models/common/model_loader.py
gxd3: In the server startup command,
could you try add `--additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 8, "tensor_parallelism": 1}}, "sparse_matmul": "True", "replicate_attn_weights": "True"}'` as well to see if the error go away?

gxd3: expert_parallelism should be 16 if you're using 16 devices.

tpu_inference/layers/common/attention_interface.py
oliverlee209: My apologies, I misread that.


> https://github.com/vllm-project/tpu-inference/blob/844f9e7f49153209a2d16068598d0c3225ede935/tpu_inference/layers/common/attention_interface.py#L521-L540
> 
> 
> The code here look right to me without requiring changes.

My apologies, I misread that.

oliverlee209: > could we swap the order of this instead?
> 
> There are a few caller that reply on `new_kv_cache, outputs` being returned from `mla_attention()`, e.g.
> 
> https://github.com/vllm-project/tpu-inference/blob/07072e1274a19a259840a21b39a4c4b098f778ac/tpu_inference/layers/vllm/backends/flash_attn_mla.py#L177

I have modified it. Please check, thanks!

gxd3: Thanks!

tpu_inference/models/common/model_loader.py
oliverlee209: I tried the flag you provided, but encountered the error
`(EngineCore pid=1839) (RayWorkerWrapper pid=2186) ERROR 04-06 01:37:19 [ray_utils.py:87] NotImplementedError: Unsupported moe backend: MoEBackend.FUSED_MOE! Currently supported: [<MoEBackend.GMM_EP: 'gmm_ep'>, <MoEBackend.GMM_TP: 'gmm_tp'>]`

If you need more detailed logs, please let me know!

gxd3: I just ran the following startup command on HEAD (without any changes), and verified it work:

```
VLLM_MLA_DISABLE=0 MOE_REQUANTIZE_BLOCK_SIZE=256 MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 USE_UNFUSED_MEGABLOCKS=0 NEW_MODEL_DESIGN=1 TPU_BACKEND_TYPE=jax MODEL_IMPL_TYPE=flax_nnx vllm serve gs://ranlihao-deepseek/model/r1 --served-model-name deepseek-ai/DeepSeek-R1  --gpu-memory-utilization=0.95 --max-model-len=5120 --max-num-batched-tokens=1024 --max-num-seqs=192 --kv-cache-dtype=fp8    --no-enable-prefix-caching     --tensor-parallel-size 16     --additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 16, "tensor_parallelism": 1}}, "sparse_matmul": "True", "replicate_attn_weights": "True"}'   --async-scheduling
```
(some are optional, e.g. MOE_REQUANTIZE_BLOCK_SIZE, MOE_REQUANTIZE_WEIGHT_DTYPE)

Could you give it a try?

oliverlee209: I still got the same error. I used the image `vllm/vllm-tpu:nightly-20260403-350d385-d9408ff`. Should I use this image `nightly-20260405-6e64193-968ed02`?

oliverlee209: I wonder if there are any differences between DeepSeek R1 and DeepSeek V3 that could be causing this error.

gxd3: Pasting the commit I used in case it's helpful:
Vllm: 4dd49b06f81af2238ac5a86cfb0b7220083eb125
Tpu-inference: 55b4c5ffecbdb61aa2e23c58463c7734b1d06584 

The latest error is still `Unsupported moe backend: MoEBackend.FUSED_MOE! ` ?

oliverlee209: Yes, I still encountered this error. 
Could I get the image tag from you? My environment is air gapped.

gpolovets1: The error seems to be related to the env variables being set: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/moe/utils.py#L269
Maybe you can add a print statement to confirm the env variables being used?

gpolovets1: FYI you may also be getting an error because of recent commits requiring an updated [libtpu version](https://github.com/vllm-project/tpu-inference/commit/fe442ba25c8ea40ecabcd1fa2a04c6d918e4747d#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552). So you may want to use a newer docker after all (e.g. from today/yesterday).

oliverlee209: > The error seems to be related to the env variables being set: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/moe/utils.py#L269 Maybe you can add a print statement to confirm the env variables being used?

Thanks for your suggestion! I found I set `USE_MOE_EP_KERNEL` to `1`. Now the error is fixed. I am testing the command you provided.

oliverlee209: > In the server startup command, could you try add `--additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 8, "tensor_parallelism": 1}}, "sparse_matmul": "True", "replicate_attn_weights": "True"}'` as well to see if the error go away?

I wonder why the `"tensor_parallelism": 1` is set in the `--additional_config`.

gxd3: ""expert_parallelism": 8, "tensor_parallelism": 1" is to indicate to use expert parallelism for MOE layers

oliverlee209: This is the command I used:

`vllm serve --model=${POD_DATA_DIR} --tensor-parallel-size 16 --max-model-len 5120 --max-num-seqs 192 --max-num-batched-tokens 1024 --no-enable-prefix-caching --kv-cache-dtype=fp8 --enable-expert-parallel --async-scheduling --gpu-memory-utilization 0.90 --additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true, "expert_parallelism": 16, "tensor_parallelism": 1}}, "sparse_matmul": "True", "replicate_attn_weights": "True"}'`

Could I ask what is the difference between `--tensor-parallel-size 16` and `"tensor_parallelism": 1` in `--additional_config`?

I noticed that:
1. `tensor_parallel_size` is assigned to `pc_tensor_parallelism`
2. `tensor_parallelism` is assigned to `ss_tensor_parallelsim`
according to
https://github.com/vllm-project/tpu-inference/blob/70c79bbe366a094a4531f5f9c7f4f6f3f55663b2/tpu_inference/layers/common/sharding.py#L155-L156

I’m also curious about the difference between `pc_tensor_parallelism` and `ss_tensor_parallelsim`. My understanding is that if I am using 16 devices, I should set the tensor parallel size to 16. Is that correct? What is the reason why we need to set `"tensor_parallelism": 1` in the `--additional_config`? Many thanks!

gxd3: I agree this part is a bit messy.

what i usually do is look at the server log when it start up, it will print out mesh, axises size, something like: "[tpu_runner.py:303] Init mesh | mesh=Mesh('data': 1, 'attn_dp': 16, 'attn_dp_expert': 1, 'expert': 1, 'model': 1, axis_types=(Auto, Auto, Auto, Auto, Auto))",

Then look at the table in https://github.com/vllm-project/tpu-inference/blob/70c79bbe366a094a4531f5f9c7f4f6f3f55663b2/tpu_inference/layers/common/sharding.py#L33-L51, as well as sharding annotations in the tpu_inference/models/jax/deepseek_v3.py for the important arrays, to see if the sharding is same as expectaion (TP or EP or DP etc)

Hope this helps :)

oliverlee209: > I agree this part is a bit messy.
> 
> what i usually do is look at the server log when it start up, it will print out mesh, axises size, something like: "[tpu_runner.py:303] Init mesh | mesh=Mesh('data': 1, 'attn_dp': 16, 'attn_dp_expert': 1, 'expert': 1, 'model': 1, axis_types=(Auto, Auto, Auto, Auto, Auto))",
> 
> Then look at the table in
> 
> https://github.com/vllm-project/tpu-inference/blob/70c79bbe366a094a4531f5f9c7f4f6f3f55663b2/tpu_inference/layers/common/sharding.py#L33-L51
> 
> , as well as sharding annotations in the tpu_inference/models/jax/deepseek_v3.py for the important arrays, to see if the sharding is same as expectaion (TP or EP or DP etc)
> Hope this helps :)

Thank you so much! 
When I set `--max-model-len 4096`, `--max-num-seqs 512`, and `--max-num-batched-tokens 4096`, I encountered the error
 `jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: E1000: CompileTimeHbmOom: XLA:TPU compile permanent error. Ran out of memory in memory space hbm. Used 95.94G of 94.75G hbm. Exceeded hbm capacity by 1.20G.`. 
 However, I was able to run these exact configurations successfully using the code from my PR.

gxd3: "However, I was able to run these exact configurations successfully using the code from my PR."
could you paste the `Init mesh | mesh=Mesh...` logging from your server start up?

want to confirm what the differences between
PartitionSpec(ShardingAxisName.ATTN_DATA, None,
                      ShardingAxisName.ATTN_HEAD)
and PartitionSpec(ShardingAxisName.MLP_TENSOR)

Thanks!

oliverlee209 (commented):
Please review it! Thanks!

gxd3 (approved):

gpolovets1 (approved):

PR #2147, opened 2026-04-06
Added Compressed Tensors wNa16 scheme

NotDevinLe:
# Description

Loaded the model as 8 int4's packed into a single int32. However, when processing the weights, I made it so that the model expands such that each int4 becomes an int8 that persists in memory. This is because the TPU doesn't have any native int4 operations so the kernel will have to automatically do slow bit shift operations and the speed is significantly slower. During the apply, the implementation follows standard de-quantization logic.

# Tests

Ran for expected output on RedHatAI/Qwen2.5-VL-7B-Instruct-quantized.w4a16 and also ran unit tests.


tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -76,9 +82,14 @@ def get_moe_method(
         if quant_config._is_fp8_w8a8(weight_quant, input_quant):
             return VllmCompressedTensorsW8A8Fp8MoEMethod(
                 weight_quant, input_quant, layer.moe_config, quant_config.mesh)
-        else:
-            raise RuntimeError(
-                f"Unsupported FusedMoe scheme: {weight_quant}, {input_quant}")
+        if quant_config._is_wNa16_group_channel(weight_quant, input_quant):
+            return VllmCompressedTensorsWNA16MoEMethod(
```
dmmolitor, line 86: It looks like this class was removed in the last commit on this branch and needs to be removed here as well. 

What was the reason for removing MoE support?

NotDevinLe: I'm still trying to run some tests for the MoE models, but they're taking a lot longer to set up than I expected. I had some unit tests that it passed, but I wanted to run it on actual models to ensure correctness. I should have it done soon.

dmmolitor:
Did you test the quality? For example, benchmarking against one of the datasets reported for [RedHatAI/Qwen2.5-VL-7B-Instruct-quantized.w4a16](https://huggingface.co/RedHatAI/Qwen2.5-VL-7B-Instruct-quantized.w4a16) and comparing to the existing VLLM numbers?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_wNa16.py
```diff
@@ -0,0 +1,346 @@
...
+        if num_bits not in WNA16_SUPPORTED_BITS:
+            raise ValueError(
+                f"Unsupported num_bits = {num_bits}. "
+                f"Supported num_bits = {WNA16_SUPPORTED_BITS}")
+
+    @classmethod
+    def get_min_capability(cls) -> int:
+        return 0
+
+    def create_weights(
```
kyuyeunk, line 80: we rarely create a function from scratch and always try to leverage existing vllm's function - which significantly helps with the maintainability. is there an existing function in vllm side that we can subclass or reuse?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_wNa16.py
```diff
@@ -0,0 +1,346 @@
...
+
+        if self.has_g_idx:
+            weight_g_idx = RowvLLMParameter(
+                data=torch.empty(input_size_per_partition, dtype=torch.int32),
+                input_dim=0,
+                weight_loader=weight_loader,
+            )
+            layer.register_parameter("weight_g_idx", weight_g_idx)
+
+    def _unpack_packed_tensor(self, packed: jax.Array) -> jax.Array:
```
kyuyeunk, line 173: please move this function to this file: https://github.com/vllm-project/tpu-inference/blob/703373d00501d7bef2b822f0901d3a10eaa50d94/tpu_inference/layers/common/quantization/__init__.py#L73

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_wNa16.py
```diff
@@ -0,0 +1,346 @@
...
+            if weights.zero_point is not None:
+                layer.weight_zero_point_unpacked = to_parameter_list(weights.zero_point)
+            if weights.bias is not None:
+                layer.bias = to_parameter_list(weights.bias)
+                
+        # Save sort_indices to permute the input 'x' in the forward pass
+        if sort_indices is not None:
+            layer.sort_indices = Parameter(torch_view(sort_indices), requires_grad=False)
+
+    def _dequantize_to_bf16(self, weight_unpacked: jax.Array, weight_scale: jax.Array,
```
kyuyeunk, line 265: please reuse existing function: https://github.com/vllm-project/tpu-inference/blob/703373d00501d7bef2b822f0901d3a10eaa50d94/tpu_inference/layers/common/quantization/__init__.py#L85

PR #2152, opened 2026-04-06
Add support for FP8 activation in fused MoE

dawnhan1111:
This change introduces FP8 quantization for the hidden_states All-Gather operation within the fused Mixture of Experts (MoE) layer. Previously, activations were always processed in BF16.

The primary motivation is to reduce the communication overhead of the All-Gather, which is a major bottleneck in MoE layers. By quantizing the activations to FP8 before the All-Gather and dequantizing afterwards, we significantly reduce the amount of data transferred between devices.

The e4m3fn FP8 format was chosen to maintain reasonable precision for the forward pass, coupled with a separate FP32 scale factor calculated per token to handle the dynamic range.

This change yields a ~1.96x speedup on the All-Gather operation (from 558.79 us to 283.35 us) based on XProf benchmarks.

The implementation involves:

-  Calculating a per-token scaling factor based on the maximum absolute value of the hidden states.
-  Quantizing the hidden_states to FP8 using this scale.
-  Performing the All-Gather on the FP8 tensor and the FP32 scale.
-  Dequantizing the hidden_states back to the original dtype using the gathered scale within a jax.shard_map to ensure the operation is sharded correctly.

# Tests
VLLM_USE_V1=1 NEW_MODEL_DESIGN=1  TPU_BACKEND_TYPE=jax MODEL_IMPL_TYPE=vllm VLLM_MLA_DISABLE=0 MOE_REQUANTIZE_BLOCK_SIZE=512 MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 vllm serve --model gs://tpu-commons-ci/deepseek/r1 --generation-config /workspace/generation_configs/DeepSeek-R1 --served-model-name deepseek-ai/DeepSeek-R1 --load-format=runai_streamer   --seed 42  --max-num-seqs 448  --max-num-batched-tokens 1024  --tensor-parallel-size 8  --no-enable-prefix-caching  --download_dir /mnt/disks/persist  --max-model-len 2048 --kv-cache-dtype=fp8 --enable-expert-parallel --gpu-memory-utilization=0.95 --additional_config=\{\"sharding\":\ \{\"sharding_strategy\":\ \{\"enable_dp_attention\":\ true\}\}\}  --async-scheduling

bf16 accuracy :

Serving Benchmark Result :
Successful requests:                     5000
Benchmark duration (s):                  312.02
Total input tokens:                      1015147
Total generated tokens:                  10000
Request throughput (req/s):              16.02
Output token throughput (tok/s):         32.05
Total token throughput (tok/s):          3285.47

Time to First Token
Mean TTFT (ms):                          210849.76
Median TTFT (ms):                        225937.70
P99 TTFT (ms):                           310241.93

Time per Output Token (excl. 1st token)
Mean TPOT (ms):                          2203.92
Median TPOT (ms):                        1203.22
P99 TPOT (ms):                           1366.31

Inter-token Latency
Mean ITL (ms):                           2203.92
Median ITL (ms):                         1203.22
P99 ITL (ms):                            1366.31

End-to-end Latency
Mean E2EL (ms):                          213053.68
Median E2EL (ms):                        227144.85
P99 E2EL (ms):                           311447.88

Evaluating MMLU...
Results
{'accuracy': 0.876, 'gen_num': 5000}

fp8 accuracy: 

Serving Benchmark Result:
Successful requests:                     5000
Benchmark duration (s):                  163.99
Total input tokens:                      1015147
Total generated tokens:                  10000
Request throughput (req/s):              30.49
Output token throughput (tok/s):         60.98
Total token throughput (tok/s):          6251.41

Time to First Token
Mean TTFT (ms):                          83104.57
Median TTFT (ms):                        75970.88
P99 TTFT (ms):                           161554.93

Time per Output Token (excl. 1st token)
Mean TPOT (ms):                          1201.78
Median TPOT (ms):                        1204.70
P99 TPOT (ms):                           1231.51

Inter-token Latency
Mean ITL (ms):                           1201.78
Median ITL (ms):                         1204.70
P99 ITL (ms):                            1231.51

End-to-end Latency
Mean E2EL (ms):                          84306.36
Median E2EL (ms):                        77175.77
P99 E2EL (ms):                           162781.22

Evaluating MMLU...
Downloading builder script: 4.20kB [00:00, 18.0MB/s]
Results

{'accuracy': 0.8742, 'gen_num': 5000}

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
     JITTED_MM_MODULE_KEYS: list[str] = []
     REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES: list[str] = []
     RAGGED_GATED_DELTA_RULE_IMPL: str = "ragged_gated_delta_rule_chunked"
+    MOE_ALL_GATHER_ACTIVATION_DTYPE: str = ""
 
 
 def env_with_choices(
@@ -248,6 +249,8 @@ environment_variables: dict[str, Callable[[], Any]] = {
     env_with_choices(
         "RAGGED_GATED_DELTA_RULE_IMPL", "ragged_gated_delta_rule_chunked",
         ["ragged_gated_delta_rule_ref", "ragged_gated_delta_rule_chunked"]),
+    "MOE_ALL_GATHER_ACTIVATION_DTYPE":
+    lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
 }
 
 
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -24,6 +24,7 @@ import tpu_inference.envs as envs
 from tpu_inference.kernels.gather import gather_reduce as gather_reduce_sc
 from tpu_inference.kernels.gather.ragged_gather import ragged_gather
 from tpu_inference.kernels.megablox.gmm_v2 import gmm_v2
+from tpu_inference.layers.common.quantization import quantize_tensor
 from tpu_inference.layers.common.sharding import ShardingAxisName
 from tpu_inference.utils import get_mesh_shape_product
 
@@ -353,6 +354,28 @@ def expert_parallel_gmm(
     )
 
 
+def _apply_all_gather_fp8(hidden_states: jax.Array, mesh: Mesh,
+                          dtype: jnp.dtype) -> jax.Array:
+    hidden_states_q, scale = quantize_tensor(
+        jnp.float8_e4m3fn,
+        hidden_states,
+        axis=-1,
+    )
+    # quantize_tensor squeezes the scale if axis is int. We need to expand it back.
+    scale = jnp.expand_dims(scale, -1)
+
+    # Dequantize if needed
+    return jax.shard_map(
+        lambda x, s: (x.astype(jnp.float32) * s).astype(dtype),
+        mesh=mesh,
+        in_specs=(
+            P(ShardingAxisName.MLP_DATA, None),
+            P(ShardingAxisName.MLP_DATA, None),
+        ),
+        out_specs=P(ShardingAxisName.MLP_DATA, None),
+    )(hidden_states_q, scale)
+
+
 @jax.jit(static_argnames=(
     "topk",
     "renormalize",
@@ -362,6 +385,7 @@ def expert_parallel_gmm(
     "scoring_fn",
     "sc_kernel_threshold",
     "sc_kernel_col_chunk_size",
+    "all_gather_fp8",
 ))
 def fused_moe_func(
     hidden_states: jax.Array,
@@ -380,6 +404,7 @@ def fused_moe_func(
     scoring_fn: str,
     sc_kernel_threshold: int,
     sc_kernel_col_chunk_size: int,
+    all_gather_fp8: bool = False,
 ) -> jax.Array:
     """Route tokens in hidden_states into each experts based on routing.
 
@@ -475,6 +500,9 @@ def fused_moe_func(
 
     x_out_spec = (P(ShardingAxisName.EXPERT_DATA)
                   if use_ep else P(ShardingAxisName.MLP_DATA))
+    if all_gather_fp8:
+        hidden_states = _apply_all_gather_fp8(hidden_states, mesh, dtype)
+
     x, group_sizes, topk_argsort_revert_indices = jax.shard_map(
         _process_tokens_locally,
         mesh=mesh,
--- a/tpu_inference/layers/common/moe.py
+++ b/tpu_inference/layers/common/moe.py
@@ -24,6 +24,7 @@ from tpu_inference import envs
 from tpu_inference.kernels.fused_moe.v1.kernel import fused_ep_moe
 from tpu_inference.layers.common.fused_moe_gmm import fused_moe_func
 from tpu_inference.logger import init_logger
+from tpu_inference.utils import to_jax_dtype
 
 if TYPE_CHECKING:
     from tpu_inference.layers.common.process_weights.moe_weights import (
@@ -122,6 +123,13 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+                # Check if activation_dtype was passed via kwargs or as an environment variable
+                activation_dtype = extra_backend_kwargs.get(
+                    "activation_dtype", envs.MOE_ALL_GATHER_ACTIVATION_DTYPE)
+                all_gather_fp8 = (bool(activation_dtype)
+                                  and to_jax_dtype(activation_dtype)
+                                  == jnp.float8_e4m3fn)
+
                 output = fused_moe_func(
                     hidden_states=x,
                     w1=weights.w13_weight,
@@ -139,6 +147,7 @@ def moe_apply(
                     scoring_fn=layer.scoring_func,
                     sc_kernel_threshold=envs.SC_KERNEL_THRESHOLD,
                     sc_kernel_col_chunk_size=envs.SC_KERNEL_COL_CHUNK_SIZE,
+                    all_gather_fp8=all_gather_fp8,
                 )
             case MoEBackend.DENSE_MAT:
                 # NOTE: circular import avoidance
--- a/scripts/multihost/benchmarks/torchax/run_deepseek_v3_1k_8k.sh
+++ b/scripts/multihost/benchmarks/torchax/run_deepseek_v3_1k_8k.sh
@@ -7,7 +7,7 @@ SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
 NIGHTLY_SCRIPT="${SCRIPT_DIR}/../../nightly_benchmarking.sh"
 
 # Adjust model-path, max-seqs, and code-hash below when officially serving DeepSeek.
-bash "${NIGHTLY_SCRIPT}" \
+MOE_ALL_GATHER_ACTIVATION_DTYPE=fp8 bash "${NIGHTLY_SCRIPT}" \
   --model-path "/mnt/disks/checkpoint/hub/models--deepseek-ai--DeepSeek-R1/snapshots/56d4cbbb4d29f4355bab4b9a39ccb717a14ad5ad/" \
   --model-name "DeepSeek-R1" \
   --tokenizer "deepseek-ai/DeepSeek-R1" \


tpu_inference/layers/common/moe.py
```diff
@@ -122,6 +122,13 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+                # Check if activation_dtype was passed via kwargs or as an environment variable
+                activation_dtype = extra_backend_kwargs.get(
+                    "activation_dtype",
+                    os.environ.get("VLLM_TPU_ACTIVATION_DTYPE", "")
```
gxd3: this is a new environment variable?
as i can't find it in https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/envs.py

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -411,38 +338,43 @@ def _process_tokens_locally(hidden_states_local, topk_indices_local):
...
-                shard_output_end,
-            )
-        else:
-            x = hidden_states_local[token_indices_sorted]
-
         return x, group_sizes_local, topk_argsort_revert_indices
 
-    x_out_spec = (P(ShardingAxisName.EXPERT_DATA)
-                  if use_ep else P(ShardingAxisName.MLP_DATA))
+    if scatter_gather_fp8:
```
gxd3: nit: could we put this into a helper function? :)

tpu_inference/layers/common/moe.py
dawnhan1111: This is a new variable that isn't currently in `envs.py`. I considered reusing `MOE_REQUANTIZE_WEIGHT_DTYPE`, but since that specifically governs weight storage rather than activation quantization for communication, it felt technically inaccurate. I can add it to `envs.py`, unless you have a preference for a different existing variable.

gxd3: Thanks for the explanation!
please add the new environment variable to `envs.py` in this PR as well, perhaps name it as `MOE_ALL_GATHER_ACTIVATION_DTYPE` so that it's more specific?

tpu_inference/layers/common/fused_moe_gmm.py
gxd3: Thanks!

tpu_inference/layers/common/moe.py
```diff
@@ -122,6 +123,13 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+                # Check if activation_dtype was passed via kwargs or as an environment variable
+                activation_dtype = extra_backend_kwargs.get(
+                    "activation_dtype",
+                envs.MOE_ALL_GATHER_ACTIVATION_DTYPE
+                )
+                scatter_gather_fp8 = (activation_dtype == "fp8")
```
gxd3: should this be `float8_e4m3fn`?

tpu_inference/envs.py
```diff
@@ -17,38 +17,29 @@
...
     LAYOUT_Q_PROJ_AS_NDH: bool = False
-    USE_JAX_PROFILER_SERVER: bool = False
-    JAX_PROFILER_SERVER_PORT: int = 9999
-    USE_BATCHED_RPA_KERNEL: bool = False
-    FORCE_MOE_RANDOM_ROUTING: bool = False
-    SC_KERNEL_THRESHOLD: int = 16777216
-    SC_KERNEL_COL_CHUNK_SIZE: int = 1024
-    JITTED_MM_MODULE_KEYS: list[str] = []
-    REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES: list[str] = []
+    MOE_ALL_GATHER_ACTIVATION_DTYPE: str = ""
```
gxd3, line 45: set default to be `float8_e4m3fn` as well?

gxd3 (approved):

tpu_inference/layers/common/moe.py
```diff
@@ -122,6 +123,26 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+<<<<<<< Updated upstream
```
kyuyeunk: merge conflict?

dawnhan1111: Thanks for pointing it out, took some time resolved the merge conflict in this PR

tpu_inference/layers/common/moe.py
dawnhan1111: Done

tpu_inference/envs.py
dawnhan1111: Acknowledged.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -326,6 +326,36 @@ def expert_parallel_gmm(
     )
 
 
+def _apply_scatter_gather_fp8(
+    hidden_states: jax.Array, mesh: Mesh, dtype: jnp.dtype
+) -> jax.Array:
+    fp8_max = jnp.finfo(jnp.float8_e4m3fn).max.astype(jnp.float32)
+    # Calculate scale per token
+    local_amax = jnp.max(jnp.abs(hidden_states), axis=-1, keepdims=True)
```
kyuyeunk: we already have a file with all the quantization related functions. suggest reusing that function instead: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/quantization/__init__.py

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -326,6 +326,36 @@ def expert_parallel_gmm(
...
+    scale = jnp.where(local_amax == 0.0, 1.0, local_amax / fp8_max)
+    scale = jnp.maximum(scale, 1e-30)
+
+    # Quantize
+    hidden_states = jnp.clip(
+        hidden_states / scale, -fp8_max, fp8_max
+    ).astype(jnp.float8_e4m3fn)
+
+    # BARRIER: Prevent XLA from fusing the dequantize and pulling the All-Gather ABOVE the quantization
```
kyuyeunk: at which hlo pass does the fusion happens? if you can pinpoint the hlo pass, maybe better approach is to specify an xla flag to change that pass's behavior.

tpu_inference/layers/common/moe.py
```diff
@@ -122,6 +122,13 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+                # Check if activation_dtype was passed via kwargs or as an environment variable
+                activation_dtype = extra_backend_kwargs.get(
+                    "activation_dtype",
+                    envs.MOE_ALL_GATHER_ACTIVATION_DTYPE
+                )
+                scatter_gather_fp8 = (activation_dtype == "float8_e4m3fn")
```
kyuyeunk: use `to_jax_dtype` instead of string based matching.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -326,6 +326,36 @@ def expert_parallel_gmm(
...
+    hidden_states, scale = jax.lax.optimization_barrier((hidden_states, scale))
+
+    # Dequantize if needed
+    return jax.shard_map(
+        lambda x, s: (x.astype(jnp.float32) * s).astype(dtype),
```
kyuyeunk, line 369: i'm a bit confused here. what does this function do? it seems like it's just performing quantization and immediately perform dequantization inside the shard map? i thought it was performing quantized allgather but it seems like in specs & out specs doesn't change?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -326,6 +326,36 @@ def expert_parallel_gmm(
...
+        mesh=mesh,
+        in_specs=(
+            P(ShardingAxisName.MLP_DATA, None),
+            P(ShardingAxisName.MLP_DATA, None),
+        ),
+        out_specs=P(ShardingAxisName.MLP_DATA, None),
+        check_vma=False,
```
kyuyeunk: unless it is strictly necessary, recommend not using `check_vma=False` as it completely byapsses compiler's safe guard that guanratees sharding is applied correctly.

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: You are right that the Python code shows immediate dequantization. However, the intent is to let XLA perform the optimization. And XLA pushes the dequantization pass after the All-Gather in the subsequent steps, so the communication actually happens in FP8 in the compiled HLO.

We tried another way in microbenchmark, which is put the dequantizaion after shardmap. into `_process_tokens_locally`, the code will looks like below and the way shard_map is written leans toward being explicit rather than concise

```
def _process_tokens_locally(
      hidden_states_local, topk_indices_local, scale_local=None
  ):
   ...
    if scale_local is not None:
      # Dequantize to bfloat16 before gather to ensure we have a single,
      # standard-width gather fusion (multiple of 128). This matches the
      # faster bfloat16 throughput and avoids inefficient pad fusions.
      hidden_states_dequant = (
          hidden_states_local.astype(jnp.float32) * scale_local
      ).astype(jnp.bfloat16)
      x = hidden_states_dequant[token_indices_sorted % num_tokens_local]
    else:
      x = hidden_states_local[token_indices_sorted]

...
x, group_sizes, topk_argsort_revert_indices = jax.shard_map(
        _process_tokens_locally,
        mesh=mesh,
        in_specs=(
            P(ShardingAxisName.MLP_DATA, None),
            P(ShardingAxisName.MLP_DATA, None),
            P(ShardingAxisName.MLP_DATA, None),
        ),
        out_specs=(
            P(ShardingAxisName.MLP_DATA, None),
            P(ShardingAxisName.MLP_DATA),
            P(ShardingAxisName.MLP_DATA),
        ),
        check_vma=False,
    )(hidden_states_fp8, topk_indices, scale)
  else:
    x, group_sizes, topk_argsort_revert_indices = jax.shard_map(
        _process_tokens_locally,
        mesh=mesh,
        in_specs=(
            P(ShardingAxisName.MLP_DATA, None),
            P(ShardingAxisName.MLP_DATA, None),
        ),
        out_specs=(
            P(ShardingAxisName.MLP_DATA, None),
            P(ShardingAxisName.MLP_DATA),
            P(ShardingAxisName.MLP_DATA),
        ),
        check_vma=False,
    )(hidden_states, topk_indices)
...
```

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: Done

tpu_inference/layers/common/moe.py
dawnhan1111: Thanks for the comment! It's a good idea. Done

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: Pinpointing the exact HLO pass is a good idea. I use the optimization_barrier here because it provides a local, targeted fix for this specific MoE kernel. Disabling or modifying an XLA fusion pass via global flags risks affecting fusion behavior and performance in other unrelated parts of the model. The barrier ensures safety and maintainability without global side-effects.
Let me know what you think!

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: Thanks for the comment! Updated code and reuse `quantize_tensor` here.

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: I got the barrier removed here and it works fine now. All-gather is quantized correctly.

tpu_inference/layers/common/moe.py
```diff
@@ -122,6 +123,13 @@ def moe_apply(
                     **extra_backend_kwargs,
                 )[:, :actual_hidden_size]
             case MoEBackend.GMM_EP | MoEBackend.GMM_TP:
+                # Check if activation_dtype was passed via kwargs or as an environment variable
+                activation_dtype = extra_backend_kwargs.get(
+                    "activation_dtype", envs.MOE_ALL_GATHER_ACTIVATION_DTYPE)
+                scatter_gather_fp8 = (bool(activation_dtype)
```
gxd3: do we need the `bool(activation_dtype) and` part?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -475,6 +500,9 @@ def _process_tokens_locally(hidden_states_local, topk_indices_local):
 
     x_out_spec = (P(ShardingAxisName.EXPERT_DATA)
                   if use_ep else P(ShardingAxisName.MLP_DATA))
+    # if scatter_gather_fp8:
```
gxd3: do we need to keep the `if....`? (it has been commented out)

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -353,6 +354,28 @@ def expert_parallel_gmm(
     )
 
 
+def _apply_scatter_gather_fp8(hidden_states: jax.Array, mesh: Mesh,
```
gxd3: maybe naive question:
why we call "scatter" here?
isn't FP8 for all_gather?

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: Yes, we still need this, added it back.

tpu_inference/envs.py
dawnhan1111: After the offline discussion, we decide to change the default value back to NULL, and only enable it as needed for now.

tpu_inference/layers/common/moe.py
dawnhan1111: Based on the offline discussion, we still want to keep it, read the environment type, check if it's empty, if not, check if it's fp8. Then we enable the feature as needed.

Now I only enable it in `torchax/deepseek_v3_1k_8k.sh`, please let me know if it's better to add it in `nightly_benchmarking.sh` instead that more models can enable it.
I think it's good to start from one model and probably enable for more models as the next step

tpu_inference/layers/common/fused_moe_gmm.py
dawnhan1111: That’s a fair distinction. I used "scatter" to describe the MoE dispatch logic conceptually, but since the implementation relies on the `all_gather` primitive, renaming it provides better clarity. I've refactor the name here.

gxd3 (approved):

PR #2183, opened 2026-04-08
Add configurable logic for gdn_attention. Improve jax gdn chunked impl

helloworld1:
# Description

* Add configurable logic for gdn_attention through env variable e.g. RAGGED_GATED_DELTA_RULE_IMPL=ragged_gated_delta_rule_chunked 
* Integrated triangle_solver kernel 
* Parallelize the scan in decode only case

FIXES: #123456

# Tests
* Tested offline_inference.py script and output sensible outputs
* Benchmarked the performance

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/kernels/gdn/triangle_solver.py
+++ b/tpu_inference/kernels/gdn/triangle_solver.py
@@ -1,0 +1,127 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+
+
+def local_forward_substitution(A, b):
+    """Solves A X = B for unit lower triangular matrix A using forward substitution.
+
+    Args:
+      A: A tensor of shape (B, N, N) representing a batch of unit lower triangular
+        matrices.
+      b: A tensor of shape (B, N, K) representing the right-hand side.
+
+    Returns:
+      A tensor of shape (B, N, K) representing the solution X.
+    """
+    B, N, K = b.shape
+    x_list = []
+    for i in range(N):
+        b_i = b[:, i, :]
+        if i == 0:
+            x_i = b_i
+        else:
+            stacked_x = jnp.stack(x_list, axis=1)  # (B, i, K)
+            all_prev_A = A[:, i, :i]  # (B, i)
+            prev_sum = jnp.sum(all_prev_A[..., None] * stacked_x,
+                               axis=1)  # (B, K)
+            x_i = b_i - prev_sum  # (B, K) for the row i
+        x_list.append(x_i)
+    x = jnp.stack(x_list, axis=1)  # (B, N, K)
+    return x
+
+
+def decompose_triangular_matrix_inverse_pallas_kernel(A_ref,
+                                                      x_ref,
+                                                      *,
+                                                      block_size=16):
+    A = A_ref[...]
+    # Matrix dimension
+    B, N, _ = A.shape
+    num_blocks = N // block_size
+
+    # same as lower_triangle_solver_pallas_kernel but 2d block wise
+    # AX = I, solve for X block wise. X = I - sum(AX_prev)
+    for i in range(num_blocks):
+        start, end = i * block_size, (i + 1) * block_size
+        e_block = jnp.eye(N, dtype=A.dtype)[start:end, :]
+        e_block = jnp.broadcast_to(e_block, (B, block_size, N))
+        if i == 0:
+            target_b = e_block
+        else:
+            interaction_A = A[:, start:end, :start]
+            solved_x = x_ref[:, :start, :]
+            prev_sum = jnp.matmul(interaction_A,
+                                  solved_x,
+                                  precision=jax.lax.Precision.HIGHEST)
+            target_b = e_block - prev_sum
+
+        local_A = A[:, start:end, start:end]
+        x_block = local_forward_substitution(local_A, target_b)
+        x_ref[..., start:end, :] = x_block
+
+
+def decompose_triangular_matrix_inverse_pallas(A,
+                                               *,
+                                               n_block_size=64,
+                                               block_size=16):
+    """Inverts unit lower triangular matrices using a block-wise approach in Pallas.
+
+    This function solves A X = I for X, where A is a unit lower triangular matrix.
+    It uses a block-wise Gaussian elimination approach to improve performance.
+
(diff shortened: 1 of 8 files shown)


tpu_inference/kernels/gdn/triangle_solver.py
jrplatin, line 1: Can we please add better docstrings / typing hints here? Thanks!

tpu_inference/layers/vllm/ops/gdn_attention.py
```diff
@@ -71,21 +96,28 @@ def run_jax_gdn_attention_local(
...
         - The output tensor of shape `(num_tokens, n_v * d_v)`.
     """
-    # Ensure query_start_loc is monotonically increasing to handle padded slots
+
+    # Ensure query_start_loc is monotonically increasing. This is required to
+    # handle cases where query_start_loc might be padded with trailing zeros.
     query_start_loc = jnp.maximum.accumulate(query_start_loc)
 
-    out_mixed_qkv, new_conv_state = ragged_conv1d(
+    # Switch conv implementaion based on config once we have more than 1 impl
```
jrplatin: Is this a TODO?

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
```diff
@@ -17,6 +17,8 @@
 import jax.numpy as jnp
 from jax import lax
 
+import tpu_inference.kernels.gdn.triangle_solver as triangle_solver
```
jrplatin, line 20: nit: do we need the "as" here?

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
```diff
@@ -116,31 +120,46 @@ def pack_inputs_single_stream(
...
-                               dtype=unpadded_data.dtype)
-        packed_data = packed_data.at[padded_indices_valid].set(unpadded_data)
-        return packed_data
-
-    packed_query = pad_and_concatenate(query)
-    packed_key = pad_and_concatenate(key)
-    packed_value = pad_and_concatenate(value)
-    packed_g = pad_and_concatenate(g)
-    packed_beta = pad_and_concatenate(beta)
+    max_packed_tokens = (max_packed_tokens + chunk_size -
```
jrplatin, line 131: nit: this seems formatted weird / hard to read?

tpu_inference/kernels/gdn/triangle_solver.py
```diff
@@ -0,0 +1,106 @@
...
+                                                      x_ref,
+                                                      *,
+                                                      block_size=16):
+    A = A_ref[...]
+    # Matrix dimension
+    B, N, _ = A.shape
+    num_blocks = N // block_size
+
+    # same as lower_triangle_solver_pallas_kernel but 2d block wise
+    # AX = I, solve for X block wise. X = I - sum(AX_prev)
```
jrplatin, line 61: Is it not an issue that we don't init `x_ref` explicitly? Just curious

tpu_inference/kernels/gdn/triangle_solver.py
```diff
@@ -0,0 +1,106 @@
...
+        if i == 0:
+            x_i = b_i
+        else:
+            stacked_x = jnp.stack(x_list, axis=1)  # (B, i, K)
+            all_prev_A = A[:, i, :i]  # (B, i)
+            prev_sum = jnp.sum(all_prev_A[..., None] * stacked_x,
+                               axis=1)  # (B, K)
+            x_i = b_i - prev_sum  # (B, K) for the row i
+        x_list.append(x_i)
+    x = jnp.stack(x_list, axis=1)  # (B, N, K)
```
jrplatin, line 47: Also curious, is it not more performant to allocatre an empty array and then update it in-place rather than doing `jnp.stack`?

tpu_inference/layers/vllm/ops/gdn_attention.py
```diff
@@ -106,10 +156,7 @@ def run_jax_gdn_attention_local(
         dt_bias,
         query_start_loc,
         state_indices,
-        n_kq,
```
jrplatin, line 109: Why did you pull these out of the common args if both impls pass them? Does the antipcated kernel not accept these?

tpu_inference/kernels/gdn/triangle_solver.py
helloworld1: I haven't tried but I don't think it is better since first allocating an empty array (set to 0) + gather is usually more costly than stack (Which is concatenate + reshape in HLO)

tpu_inference/kernels/gdn/triangle_solver.py
helloworld1: No need since all values will be overwritten. It is more performant this way

tpu_inference/layers/vllm/ops/gdn_attention.py
helloworld1: Probably, it is a more like note to remind to change once we have the kernel ragged_conv1d.

tpu_inference/layers/vllm/ops/gdn_attention.py
helloworld1: Good call. Chengji's code does not use n_kq / n_v etc so I don't want to force those optional parameters. I will probably create an interface requirement by using Callable or Protocol

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
helloworld1: Yes, in python importing just allow you to use fully name. You have to alias it for short name.
For example
import jax.numpy

You cannot do
numpy.Array()

You must do
jax.numpy.Array()

tpu_inference/layers/vllm/ops/gdn_attention.py
jrplatin: Gotcha, definitely a nit but would be good to add `TODO` explicitly since we aren't using `ragged_gdn_impl` at all right now

tpu_inference/layers/vllm/ops/gdn_attention.py
jrplatin: Oh definitely not a big deal but was just curious, feel free to resolve as you see fit

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
helloworld1: Yea, ugly code. Let me try to fix in the internal code. The internal formatting did this way

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
jrplatin: Oh nevermind, I see what you're saying -- I was thinking `from tpu_inference.kernels.gdn import triangle_solver` but I guess same difference

tpu_inference/kernels/gdn/triangle_solver.py
helloworld1: Yes, let me add.

tpu_inference/layers/vllm/ops/ragged_gated_delta_rule_chunked.py
helloworld1: Auto-formatter always did this format due to we internally use 80 line length.

jrplatin (approved):

PR #2187, opened 2026-04-09
Enable DP mode for Qwen3.5

ShobhitBehl:
# Description

Fix issues with data parallelism for Qwen3.5;
 - Replace `mesh.shape[...]` with `get_mesh_shape_product`.
 - Add `ATTN_DATA` sharding specs to GDN Attention and KV cache.
 - Fix sharding for M-RoPE positions.
 - Use SPMD representations to align block size.

# Tests
 - CI
 - `NEW_MODEL_DESIGN=1   USE_MOE_EP_KERNEL="0" MODEL_IMPL_TYPE="vllm" python examples/offline_inference.py --model Qwen/Qwen3.5-397B-A17B-FP8 --tensor-parallel-size=8 --additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' --block-size=256 --limit-mm-per-prompt='{"image": 0, "video": 0}' --kv-cache-dtype=fp8 --max-model-len=10240 --max-num-batched-tokens=8192  --max-num-seqs=512 --gpu-memory-utilization=0.8 --async-scheduling --chat-template-kwargs='{"enable_thinking": false}'`
 - `VLLM_USE_AOT_COMPILE=0 MODEL_IMPL_TYPE=vllm python examples/offline_inference.py --model=Qwen/Qwen3.5-397B-A17B-FP8 --tensor-parallel-size=8  --limit-mm-per-prompt '{"image": 0, "video": 0}' --block-size=256 --gpu-memory-utilization=0.95 --top_p=1.0 --top_k=-1 --temperature=0 --kv-cache-dtype=fp8  --chat-template-kwargs='{"enable_thinking": false}'`
commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/gdn_attention.py
+++ b/tpu_inference/layers/common/gdn_attention.py
@@ -31,6 +31,7 @@ from tpu_inference.layers.common.ragged_gated_delta_rule_chunked import \
 from tpu_inference.layers.common.ragged_gated_delta_rule_ref import \
     ragged_gated_delta_rule as ragged_gated_delta_rule_ref
 from tpu_inference.layers.common.sharding import ShardingAxisName
+from tpu_inference.utils import get_mesh_shape_product
 
 
 class RaggedConv1dImpl(enum.Enum):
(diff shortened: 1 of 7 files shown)


tpu_inference/layers/common/gdn_attention.py
```diff
@@ -211,31 +212,35 @@ def run_jax_gdn_attention(
         - The output tensor of shape `(num_tokens, n_v * d_v)`.
     """
     in_specs = (
-        P(None, ShardingAxisName.ATTN_HEAD),  # j_mixed_qkv
-        P(None, ShardingAxisName.ATTN_HEAD),  # j_b
-        P(None, ShardingAxisName.ATTN_HEAD),  # j_a
-        P(None, None, ShardingAxisName.ATTN_HEAD),  # conv_state
-        P(None, ShardingAxisName.ATTN_HEAD, None, None),  # recurrent_state
+        P(ShardingAxisName.ATTN_DATA,
```
jrplatin, line 213: So if we have both TP and attention_DP enabled, will this shard on the states on both or will `ATTN_HEAD` be None?

tpu_inference/platforms/tpu_platform.py
```diff
@@ -283,14 +283,21 @@ def update_block_size_for_backend(cls, vllm_config: VllmConfig) -> None:
...
-            backend_cls = cls._find_non_ssm_backend(vllm_config)
-            if backend_cls is None:
-                return
-            cls._align_hybrid_block_size(vllm_config, backend_cls)
-
-        return
+        # Since TPU uses an SPMD representation for layers and KV caches, we
+        # temporarily set the tensor parallel size to 1. This prevents vLLM from
+        # incorrectly partitioning the number of KV heads during hybrid block
+        # size alignment.
```
jrplatin: Just curious -- is this necessarily related to the changes from this PR or orthogonal?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -409,23 +410,22 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
                 assert kv_cache_tensor.size % page_size_bytes == 0
                 num_blocks = kv_cache_tensor.size // page_size_bytes
 
-            for j, layer_name in enumerate(kv_cache_tensor.shared_by):
-                layer_spec = layer_name_to_spec[layer_name]
+            if self.use_mla and not self.runner.vllm_config.additional_config.get(
```
jrplatin, line 413: After https://github.com/vllm-project/tpu-inference/pull/2222, there shouldn't be a case where we are using MLA and _not_ using attention DP -- does that change things here?

tpu_inference/layers/common/gdn_attention.py
ShobhitBehl: IIUC it should shard both depending on values in some scenario when DP=4 and TP=2. But I haven't tried such a case.

tpu_inference/runner/kv_cache_manager.py
ShobhitBehl: I am not sure, I've only refactored this code from earlier lines[415 - 427]. (Pulled this code out of the loop and replaced hardcoded `sharding_config` fetches to `get_mesh_shape_product` which should be equivalent) 

This is an unrelated code clean up.

tpu_inference/platforms/tpu_platform.py
ShobhitBehl: It's related.

`_align_hybrid_block_size` in vLLM code always uses `tensor_parallel_size` to shard the KV heads. If `enable_dp_attention` is set, this is incorrect since the heads are not sharded and the padded page size that's 

This fixes the computation for all cases.

tpu_inference/runner/kv_cache_manager.py
jrplatin: Oh got it -- let's keep it for now then

tpu_inference/layers/common/gdn_attention.py
jrplatin: Ah had to dig a bit but it turns out that DP is set inverse to the [attn_dp size](https://github.com/vllm-project/tpu-inference/blob/618dea5f5c0ca556a6c76a2e1cc130ff6a30893c/tpu_inference/layers/common/sharding.py#L196), so this should be fine, thanks!

jrplatin (approved):

tpu_inference/platforms/tpu_platform.py
ShobhitBehl: Realised that setting it to `1` for all cases might end up overestimating block size. Instead, overriding it from sharding config now.

PR #2188, opened 2026-04-09
Separate MoE compressed tensor schemes as done in VLLM

dmmolitor:
# Description

Split separate MoE compressed tensor schemes into their own dedicated files, aligning the directory structure with both linear compressed tensor schemes and the upstream vLLM codebase. Previously, all compressed tensor MoE schemes were housed in a single file, which will become challenging as more quantization schemes are introduced.

**Why is this change being made & problem being solved:**
- Consolidating multiple quantization schemes in a single file creates merge conflicts and reduces code readability. 
- Moving MoE quantization method classes into specialized files improves modularity and aligns with upstream vLLM conventions (specifically vLLM PR #38960), making the addition of new MoE quantization schemes cleaner.

**Specific Implementation Details:**
- Created a new `compressed_tensors_moe/` subdirectory.
- Migrated the base `VllmCompressedTensorsMoEMethod` class to `compressed_tensors_moe/compressed_tensors_moe.py`.
- Migrated `VllmCompressedTensorsW8A8Fp8MoEMethod` class to `compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py`.
- Exposed all relevant classes via `compressed_tensors_moe/__init__.py` to maintain full backward compatibility with existing module imports.

# Tests

This is a refactor only and is covered by existing tests.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/__init__.py
+++ b/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/__init__.py
@@ -1,0 +1,23 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tpu_inference.layers.vllm.quantization.compressed_tensors.compressed_tensors_moe.compressed_tensors_moe import \
+    VllmCompressedTensorsMoEMethod
+from tpu_inference.layers.vllm.quantization.compressed_tensors.compressed_tensors_moe.compressed_tensors_moe_w8a8_fp8 import \
+    VllmCompressedTensorsW8A8Fp8MoEMethod
+
+__all__ = [
+    "VllmCompressedTensorsMoEMethod",
+    "VllmCompressedTensorsW8A8Fp8MoEMethod",
+]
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tpu_inference.layers.vllm.quantization.compressed_tensors.compressed_tensors_moe.compressed_tensors_moe import \
+    VllmCompressedTensorsMoEMethod
+from tpu_inference.layers.vllm.quantization.compressed_tensors.compressed_tensors_moe.compressed_tensors_moe_w8a8_fp8 import \
+    VllmCompressedTensorsW8A8Fp8MoEMethod
+
```
jrplatin, line 19: Dumb question but why is this needed?

dmmolitor: It isn't completely necessary, but makes it so we don't need to udpate existing imports of these classes in tpu-inference.

jrplatin: What do you mean, sorry?

dmmolitor: We can remove these, but will need to update the imports in tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors.py and tests/layers/vllm/test_compressed_tensors_moe.py to reflect the added folder and separated files. (This is any easy change. I am happy with either approach.) 

[VLLM imports only the base class in the init.](https://github.com/neuralmagic/vllm/blob/8d825b87d6590ca971823890f9705988b8709add/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors_moe/__init__.py#L1)

jrplatin: Oh I see -- I think we should go with the approach that's more cannonical with vLLM / minimizes the diff between the two, which I guess would be the current code then ?

jrplatin (approved):

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/__init__.py
dmmolitor: Yes. VLLM only imports the base CompressedTensorsMoEMethod in this file and does not include the classes from the individual schemes.  We could update this to match that more closely

PR #2195, opened 2026-04-09
Fix to resolve W8A8BlockFp8LinearOp import error

JiriesKaileh:
# Description

Resolves the following error: 

```
ImportError: cannot import name 'W8A8BlockFp8LinearOp' from 'vllm.model_executor.layers.quantization.utils.fp8_utils' (/workspace/vllm/vllm/model_executor/layers/quantization/utils/fp8_utils.py)”
```
full trace: https://paste.googleplex.com/5743386385580032

Fixes part of this bug [b/500817770](https://b.corp.google.com/issues/500817770)


QiliangCui (approved):

tpu_inference/layers/vllm/quantization/mxfp4.py
```diff
@@ -25,13 +25,13 @@
 from vllm.model_executor.layers.fused_moe.activation import MoEActivation
 from vllm.model_executor.layers.fused_moe.config import (
     FusedMoEConfig, FusedMoEQuantConfig, mxfp4_w4a16_moe_quant_config)
-from vllm.model_executor.layers.fused_moe.oracle.mxfp4 import Mxfp4MoeBackend
```
jrplatin, line 28: Why is this change needed?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
```diff
@@ -61,16 +57,6 @@ def __init__(
         self.is_static_input_scheme = is_static_input_scheme
         self.weight_block_size = self.weight_quant.block_structure
 
-        if self.weight_block_size is not None:
```
jrplatin, line 64: Why are we wholesale removing this? Is it somewhere else in the codebase now?

tpu_inference/layers/vllm/quantization/mxfp4.py
JiriesKaileh: This change isn't needed. I'll revert.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
JiriesKaileh: I did this specifically to get the dashboards up and running again. It seems that all the W8A8BlockFp8LinearOp related code in vLLM has been removed. The only trace of it I saw was in this file: https://github.com/vllm-project/vllm/blob/19ec9a0a6276d5af449b3011dac2944e8af37ee0/benchmarks/kernels/benchmark_block_fp8_gemm.py#L53 .

it seems that W8A8Fp8LinearOp can be created as such: 

```
# Create W8A8BlockFp8LinearOp instance
    weight_group_shape = GroupShape(block_n, block_k)
    act_quant_group_shape = GroupShape(1, block_k)  # Per-token, per-group quantization

    linear_op = init_fp8_linear_kernel(
        weight_quant_key=create_fp8_quant_key(
            static=True, group_shape=weight_group_shape
        ),
        activation_quant_key=create_fp8_quant_key(
            static=False, group_shape=act_quant_group_shape
        ),
        out_dtype=torch.get_default_dtype(),
        module_name="build_w8a8_block_fp8_runner",
    )
```

I'll see if this works

jrplatin: Looks like [this](https://screenshot.googleplex.com/3oLmtqdpFErp7Dc) might be the preferred solution rather than removing it

jrplatin (changes requested):

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
JiriesKaileh: I included that change in my latest commit

jrplatin (approved):

PR #2199, opened 2026-04-10
Add "torch-backend=cpu" in the installation doc

yaochengji:
# Description

Then it will only install torch-cpu, excluding those cuda related packages

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -86,7 +86,7 @@ For debugging or development purposes, you can install `tpu-inference` from sour
 
     ```shell
     cd vllm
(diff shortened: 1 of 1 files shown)


bvrockwell:
oh wow! Thanks Chengji

cc @RobMulla we'll want to correct this everywhere (including upstream and in tpu-recipes if exists).

bvrockwell (approved):

yaochengji:
> oh wow! Thanks Chengji
> 
> cc @RobMulla we'll want to correct this everywhere (including upstream and in tpu-recipes if exists).

Np! CI and dockerfile also need to be modified.

PR #2205, opened 2026-04-10
Move gdn_attention to tpu_inference/layers/common, remove unnecessary ops

helloworld1:
# Description

* Move gdn_attention  to tpu_inference/layers/common
* Move torch related logic to gdn_attention_custom_op.py
* Clean up the in_proj op, which was removed in vllm upstream

# Tests
offline_inference.py script gives sensible output
tested benchmark

# Checklist


tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
kyuyeunk, line 1: Now that there is less room for confusion, rename this into 'gdn_attention.py'?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -148,8 +148,6 @@ def _patch_vllm_ops(self):
             needs_env=False,
         )
 
-        patch_ops.gdn_attention.apply_gated_delta_net_torch_ops_patch()
```
jrplatin, line 151: Good catch in that we no longer need this, but don't we still need to implement `gdn_in_proj` in `VllmGatedDeltaNetAttention`?

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
helloworld1: I would prefer no name collision since it is easier to work with IDE, searching and reading code. Even AI to @ the correct file. 

E.g. the op name for fused moe is fused_moe while Jax is fused_moe_gmm. While confusing, the names can still be differentiated easily.

I can rename this file to gdn_attention_op.py to make the name shorter. wdyt? @kyuyeunk

tpu_inference/models/vllm/vllm_model_wrapper.py
helloworld1: gdn_in_proj is currently removed in vllm upstream. So we don't need it right now.

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
kyuyeunk: Yup! That proposed name does sound good to me!

kyuyeunk (approved):

PR #2216, opened 2026-04-11
[Pathways Dummy Weight Loader] Avoid OOM when loading embedding and moe layer

wenxindongwork:
# Description

Fixes HBM OOM during loading embedding and MoE layers when DP is large. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -136,16 +136,19 @@ class VllmUnquantizedEmbeddingMethod(UnquantizedEmbeddingMethod):
         self.mesh = mesh
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
-        weight = t2j(layer.weight, use_dlpack=False)
-        weight = general_device_put(
-            weight,
-            NamedSharding(self.mesh, P(ShardingAxisName.MLP_TENSOR, None)))
+        weight_sharding = NamedSharding(self.mesh,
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -136,16 +136,19 @@ def __init__(self, mesh):
         self.mesh = mesh
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
-        weight = t2j(layer.weight, use_dlpack=False)
-        weight = general_device_put(
-            weight,
-            NamedSharding(self.mesh, P(ShardingAxisName.MLP_TENSOR, None)))
+        weight_sharding = NamedSharding(self.mesh,
+                                        P(ShardingAxisName.MLP_TENSOR, None))
+        weight = _load_weight_for_layer(layer, "weight", weight_sharding)
```
pv97, line 141: Does this do an extra device put? https://github.com/vllm-project/tpu-inference/blob/ea05d69d0ee9b711449c2233b925ea3e8ac1bcc9/tpu_inference/layers/vllm/quantization/unquantized.py#L63

wenxindongwork: when pathways is not used, the original behavior is preserved. when pathways is used, there will be two device_put  (one in `_load_weight_for_layer`, and one in `general_device_put`), but the second one will be a no-op.

pv97 (approved):

PR #2227, opened 2026-04-13
Add comment to reach out if changing model load time logs

sethiay:
# Description

Add comment to reach out incase changing model load time logs.

A couple of teams/products rely on these log lines to extract the model load times metrics. Changing the log lines will break the systems/tests/pipelines built on top of it. 
Having a comment will remind the contributors to check with the teams in the linked ticket before making changes to the log lines or inform them about the change, so they can make changes on their side timely.

# Tests
N/A, just an addition of a comment.

# Checklist


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -276,6 +278,8 @@ def load_weights(self):
 
         loading_end = time.time()
         total_loading_time = loading_end - loading_start
+        # There are systems that depends on this log line. 
+        # Please reach out on b/481726029 if making changes to this line.
```
jrplatin: Please don't leave internal comments here just because external contributors can't view with

You might also want to make this stronger, like:

# WARNING: please do NOT remove the following logging ....

jrplatin: In lieu of b/481726029, I'd recommend putting your gihub username

jrplatin (changes requested):

tpu_inference/models/vllm/vllm_model_wrapper.py
sethiay: Done.

jrplatin (approved):

PR #2236, opened 2026-04-14
Support for w4afp8 linear layers (compressed tensors)

dmmolitor:
# Description

Adds support for compressed-tensors format models that utilize int4 weight and fp8 activation linear layers (W4A8).

* New Quantization Scheme: Implemented VllmCompressedTensorsW4A8Fp8 in tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py ([ source](https://github.com/vllm-project/tpu-inference/compare/main...dmmolitor:tpu-inference:int4_linear?expand=1&content_ref=class+vllmcompressedtensorsw4a8fp8+compressedtensorsw4a8fp8)).
* Weight Handling: Supports symmetric quantization with uint4 storage.

# Tests

Comprehensive tests have been added in tests/layers/vllm/test_compressed_tensors_w4a8_fp8.py.

Accuracy for  RedHatAI/DeepSeek-R1-Distill-Qwen-32B-quantized.w4a16 on the MMLU dataset was 'accuracy': 0.7978 (with fp8 activations) compared to the [80.91](https://huggingface.co/RedHatAI/DeepSeek-R1-Distill-Qwen-32B-quantized.w4a16) with fp16 activations reported on HuggingFace.

# Checklist


tests/layers/vllm/test_compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,431 @@
...
+    "nm-testing/SmolLM-1.7B-Instruct-quantized.w4a16",
+]
+
+
+def ref_w4a8_fp8_dynamic(x: torch.Tensor, w_float: torch.Tensor,
+                         b: Optional[torch.Tensor]):
+
+    x_q, x_s = test_utils.ref_quantize_fp8(x, dtype=torch.float8_e4m3fn)
+
+    out = torch.einsum('bd,fd->bf', x_q.to(torch.float32),
```
jrplatin: Just curious -- why are casting both to FP32?

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -132,7 +132,7 @@ def quantized_matmul_kernel(
     # allowing the compiler to overlap output fusion and packing overhead with MXU computation
     # TODO(amandaliang): use pltpu.get_tpu_info().mxu_column_size when JAX version is newer
     compute_tile_n = MXU_SIZE * n_lane_multiplier
-    steps_n = out_block_size // compute_tile_n
+    steps_n = max(1, out_block_size // compute_tile_n)
```
jrplatin, line 135: Why is this needed?

tpu_inference/layers/common/linear.py
```diff
@@ -80,7 +81,8 @@ def sharded_quantized_matmul(x: jax.Array,
         scale_sharding = P(out_axis, )
     out_sharding = P(ShardingAxisName.ATTN_DATA, out_axis)
 
-    x_q_dtype = _get_x_q_dtype(w_q.dtype)
```
jrplatin, line 83: I notice in the code we have [this](https://github.com/dmmolitor/tpu-inference/blob/2ca3ff57c10bbb43a62002343465a8fb7c679cf4/tpu_inference/layers/common/linear.py#L33) comment, maybe that's a cleaner approach?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,288 @@
...
+        self.quant_type = W4A8_SUPPORTED_TYPES_MAP[weight_quant.num_bits]
+
+        self.weight_quant = weight_quant
+        self.out_dtype = torch.get_default_dtype()
+        self.is_static_input_scheme = is_static_input_scheme
+        self.weight_block_size = self.weight_quant.block_structure
+
+        self.linear_config = linear_config
+
+    def create_weights(
```
jrplatin, line 92: Took a cursory look at the vLLM code and it seems like this is nearly identical -- why do we need to specifically override this?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,288 @@
...
+                                                          dtype=torch.int64),
+                                         weight_loader=weight_loader)
+
+        layer.register_parameter("weight_packed", weight)
+        layer.register_parameter("weight_scale", weight_scale)
+        layer.register_parameter("weight_shape", weight_shape)
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+
+        # TODO(dmolitor): Handle the case where weights are stored unpacked.
```
jrplatin: Should we throw an error/be more defensive about this case?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,288 @@
...
+
+            if processed_weights.weight_scale is not None and processed_weights.weight_scale.ndim == 2:
+                processed_weights.weight_scale = jnp.expand_dims(
+                    jnp.transpose(processed_weights.weight_scale, (1, 0)), 1)
+
+            return processed_weights
+
+        weights = process_uint4_linear_weights(uint_weight, weight_scale, bias)
+        weights = torch_view(
+            shard_linear_weights(
```
jrplatin, line 215: Have we checked that the HBM usage (given the configured shardings) are as expected?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,288 @@
...
+        self.has_g_idx = weight_quant.actorder == ActivationOrdering.GROUP
+
+        if self.num_bits == 4:
+            self.wtype = scalar_types.uint4
+        else:
+            raise ValueError(
+                f"Unsupported num_bits = {weight_quant.num_bits}. "
+                f"Supported num_bits = {W4A8_SUPPORTED_TYPES_MAP.keys()}")
+
+        if is_static_input_scheme:
```
jrplatin, line 76: I see w8a8_fp8 has this support in TPU Inference -- any reason we can't do the same here?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
```diff
@@ -0,0 +1,288 @@
...
+        x_jax = jax_view(x)
+        weight_jax = jax_view(layer.weight)
+        weight_scale_jax = jax_view(layer.weight_scale)
+
+        outs = sharded_quantized_matmul(x_jax,
+                                        weight_jax,
+                                        weight_scale_jax,
+                                        self.linear_config.weight_sharding,
+                                        mesh=self.linear_config.mesh,
+                                        x_q_dtype=jnp.float8_e4m3fn)
```
jrplatin, line 254: Why is `x_q_dtype` needed here? Same below

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
dmmolitor: The parameter handling is essentially the same, but we don't want the [kernel selection](https://github.com/neuralmagic/vllm/blob/f02b3269e7e0ebfdc62f39db2169d88074512d6d/vllm/model_executor/layers/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py#L99) that the vLLM does in the create_weights function. The VLLM kernel selection raises an error on TPU.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
dmmolitor: This is needed to resolve the ValueError: "Need at least one array to concatenate." for the case when compute_tile_n < out_block_size.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
dmmolitor: We could add static support here as well. I omitted since we aren't currently using it (avoiding YAGNI), but I can add it if preferred.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
dmmolitor: The default activation dtype is int8 when the weights are integers in sharded_quantized_matmul. We need a way to indicate using int4xfp8 instead of int4xint8.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
dmmolitor: I agree, this is preferred. I think the best way to do this requires a small update on the vllm side. We would want to check for `"format": "pack-quantized"`  in the quantization config. See this [example](https://huggingface.co/nm-testing/Qwen3-Coder-30B-A3B-Instruct-W4A16-awq/blob/main/config.json).
Unfortunately, [VLLM only adds this to the `target_scheme_map` if the format is specified in the quantization group's`quant_config` and not in the outer `quantization_config`](https://github.com/vllm-project/vllm/blob/747256bb5d645f28579bdb1a913f3584b6a246f6/vllm/model_executor/layers/quantization/compressed_tensors/compressed_tensors.py#L307). I can add a PR to VLLM to update this with a TODO comment in TPU-Inference. 

Currently, if a model uses int4 without the packed format, it will raise an error during the model loading - complaining about missing weights. Eg
```
(EngineCore pid=3286778)   File "/mnt/pd/dmolitor/vllm/vllm/model_executor/models/qwen2.py", line 501, in load_weights
(EngineCore pid=3286778)     param = params_dict[name]
(EngineCore pid=3286778)             ~~~~~~~~~~~^^^^^^
(EngineCore pid=3286778) KeyError: 'layers.0.mlp.gate_up_proj.weight'
```
since `create_weights` expects `weight_packed` instead of `weight` here.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
dmmolitor: Not yet. Will do this in a follow up.

tests/layers/vllm/test_compressed_tensors_w4a8_fp8.py
dmmolitor: They need to be the same dtype. Updated to only cast x_q to bfloat16 to match w_float's existing dtype. The original version matched analogous function in tests/layers/vllm/test_compressed_tensors_w8a8_fp8.py.

tpu_inference/layers/common/linear.py
dmmolitor: This change lets you explicitly pass in the activation data type when calling this from a LinearMethod (scheme). If you wanted w4a4, you would pass in x_q_dtype = int4. The _get_x_q_dtype now provides defaults if x_q_dtype is not passed explicitly.

We could require explicitly passing in a dtype and omit _get_x_q_dtype entirely.

I added x_q_dtype to the docstring.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
jrplatin: Fair!

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a8_fp8.py
jrplatin: Got it, seems like we do something similar already for `VllmCompressedTensorsW8A8Fp8` so makes sense!

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
jrplatin: I don't know the kernel well enough but does the fact you need to add this (i.e. `compute_tile_n < out_block_size`) indicate that something is wrong with your implementation or is it ok that `compute_tile_n < out_block_size` ?

dmmolitor: The default tuned value for the out_block_size is 128 in /tpu-inference/tpu_inference/kernels/quantized_matmul/tuned_block_sizes.py. Since the lookup table currently doesn't have int4xfp8 entries, all matmuls are using the default tile sizes. Since this is less than the hardcoded MXU_SIZE=256, for int4xfp8 models, we always hit this compute_tile_n < out_block_size. I expect this will reduce when tuned tile sizes are added for int4xfp8 matmuls.

jrplatin: Got it, seems reasonable thanks!

jrplatin (approved):

PR #2240, opened 2026-04-14
Generalise einsum operation to support input of any dimension

muskansh-google:
# Description

Generalise the einsum operation to accept input of any dimension instead of strict 2 dimension.

Previously, the notation assumed the input x_jax is a 2D tensor of shape (M, N). This was breaking implementation when x_jax was of a different dimension (for example 3D input for Qwen3-VL). 

Also added a docstring describing the methods and the dimensions of input and output for readability.

# Tests

Manually tested that after this change the error was resolved. 

# Checklist


lk-chen (approved):

wdhongtw:
Just a side note.

For GPU (From `UnquantizedLinearMethod.apply`, to `linear_batch_invariant`, then an result in
`vllm/model_executor/layers/batch_invariant.py` `matmul_batch_invariant`.)
The corresponding implementation handles dimension combination of (2, 2), (3, 3), (3, 2), (2, 3) and (4, 4).

See <https://github.com/vllm-project/vllm/blob/fcb31c1ac3f5af4701d45007cb2b66d084328f39/vllm/model_executor/layers/batch_invariant.py#L606>

PR #2252, opened 2026-04-15
cleanup vllm ops

kyuyeunk:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import copy
-import functools
 import time
 from collections.abc import Sequence
 from contextlib import nullcontext
@@ -30,8 +29,6 @@ from flax.typing import PRNGKey
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.interop import jax_view, torch_view
 from torchax.ops.mappings import TORCH_DTYPE_TO_JAX
-from torchax.ops.ops_registry import (register_torch_dispatch_op,
-                                      register_torch_function_op)
 from vllm.config import VllmConfig, set_current_vllm_config
 from vllm.forward_context import set_forward_context
 from vllm.lora.layers import BaseLayerWithLoRA
@@ -51,7 +48,6 @@ from tpu_inference.distributed.jax_parallel_state import \
     get_pp_group as jax_get_pp_group
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
 from tpu_inference.layers.common.sharding import ShardingAxisName
-from tpu_inference.layers.vllm import ops as patch_ops
 from tpu_inference.layers.vllm.process_weights.cleanup_sharding import \
     shard_model_to_tpu
 from tpu_inference.layers.vllm.quantization import get_tpu_quantization_config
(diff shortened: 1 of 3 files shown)


kyuyeunk:
@yuyanpeng-google @weiyu0824 @muskansh-google all ci seems to pass. can i get a confirmation that this approach still works for torchax multimodals?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -50,7 +48,6 @@
     get_pp_group as jax_get_pp_group
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
 from tpu_inference.layers.common.sharding import ShardingAxisName
-from tpu_inference.layers.vllm import ops as patch_ops
```
yuyanpeng-google, line 54: Miss `import tpu_inference.layers.vllm.ops.scaled_dot_product_attention` and the op is not correctly registered.

yuyanpeng-google (approved):
LGTM, with a miss import.

Sorry about that the test cannot catch this feature. The ops is used by Gemma4 but tpu-inference doesn't support the vllm version that including Gemma4 for now.

I test manually, it should work.

tpu_inference/models/vllm/vllm_model_wrapper.py
kyuyeunk: - that gets hooked in to vllm and always gets triggered when vllm model launches: https://github.com/vllm-project/tpu-inference/blob/987a396a5226df3ab9d2461fd1fdabe5dbb27dc7/setup.py#L97
- which calls this: https://github.com/vllm-project/tpu-inference/blob/987a396a5226df3ab9d2461fd1fdabe5dbb27dc7/tpu_inference/layers/vllm/__init__.py#L20
- which imports this: https://github.com/vllm-project/tpu-inference/blob/987a396a5226df3ab9d2461fd1fdabe5dbb27dc7/tpu_inference/layers/vllm/ops/__init__.py#L15-L16

yuyanpeng-google: https://github.com/vllm-project/tpu-inference/blob/987a396a5226df3ab9d2461fd1fdabe5dbb27dc7/tpu_inference/layers/vllm/__init__.py#L15

I only see the custom_ops here, but not ops? Maybe need import ops here, too?

I discovered the missing import while testing the PR and realized that the ops was not registered. It must has missed something.

kyuyeunk: yeah, i just added it because i realized i forgot 😅 but it should be available in the pr - as you can see here: https://github.com/vllm-project/tpu-inference/blob/c7ef082ed649859870d69cbe0b53a1e20a46fcea/tpu_inference/layers/vllm/__init__.py#L16

muskansh-google:
> @yuyanpeng-google @weiyu0824 @muskansh-google all ci seems to pass. can i get a confirmation that this approach still works for torchax multimodals?

Hi @kyuyeunk, I tried patching the changes from your PR onto my changes for Qwen3-VL and got this error - https://paste.googleplex.com/5343092178812928#l=8

kyuyeunk:
> > @yuyanpeng-google @weiyu0824 @muskansh-google all ci seems to pass. can i get a confirmation that this approach still works for torchax multimodals?
> 
> 
> 
> Hi @kyuyeunk, I tried patching the changes from your PR onto my changes for Qwen3-VL and got this error - https://paste.googleplex.com/5343092178812928#l=8

The looks like just a jax jit issue. Just adding static arg name argument should fix it.

PR #2270, opened 2026-04-16
[Jax][MoE] Sharding fix in UnquantizedFusedMoEMethod for Jax native

amanseervi:
# Description

Fixes TPU OOM during MoE weight loading in the JAX native path.
Previously, `process_weights_after_loading` assigned the processed fused weights (w13_weight, w2_weight) directly to nnx.Param without explicit sharding. This caused JAX to replicate  unsharded MoE tensors across every TPU leading to OOM.

### Solution:
We call `shard_moe_weights` after process_unquantized_moe_weights. This distributes the weights across the mesh (using the existing TP partition spec) before creating the Flax parameters. 

# Tests

- Verified the model Qwen3-30B-A3B loads successfully without TPU OOM.
```bash
SKIP_JAX_PRECOMPILE=1 MODEL_IMPL_TYPE=flax_nnx vllm serve "Qwen/Qwen3-30B-A3B"   --tensor-parallel-size 4   --max-model-len 8192 --max-num-batched-tokens 8192   --trust-remote-code   --gpu-memory-utilization 0.85   --port 8000
```

-  Verified the model works correctly by sending requests

# Checklist

### DIFF
--- a/tests/models/jax/test_qwen3_moe.py
+++ b/tests/models/jax/test_qwen3_moe.py
@@ -31,14 +31,16 @@ from tpu_inference.models.jax.qwen3_moe import Qwen3MoeForCausalLM
 
 class TestQwen3MoeForCausalLM:
 
-    @pytest.mark.parametrize("model_name", [
-        "Qwen/Qwen3-30B-A3B-Instruct-2507",
-        "Qwen/Qwen3-30B-A3B-Instruct-2507-FP8",
+    @pytest.mark.parametrize("model_name,pp_rank,pp_world_size,load_format", [
+        (model, rank, world, fmt)
+        for model in [
+            "Qwen/Qwen3-30B-A3B-Instruct-2507",
+            "Qwen/Qwen3-30B-A3B-Instruct-2507-FP8",
+        ]
+        for rank, world in [(0, 1), (0, 4), (1, 4), (3, 4)]
+        for fmt in ["skip_layers_model_loader_for_test", "jax_dummy"]
+        if not (world == 1 and fmt == "jax_dummy")
     ])
-    @pytest.mark.parametrize("pp_rank,pp_world_size", [(0, 1), (0, 4), (1, 4),
-                                                       (3, 4)])
-    @pytest.mark.parametrize(
-        "load_format", ["skip_layers_model_loader_for_test", "jax_dummy"])
     def test_model_loading(
             self,
             model_name,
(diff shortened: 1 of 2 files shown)


muskansh-google (commented):
LGTM

lk-chen (approved):

amanseervi:
The Tests are failing with : 
AssertionError: Peak device memory during load_weights(Qwen/Qwen3-30B-A3B-Instruct-2507) exceeded threshold: peak_delta=2.612 GB > max=2.000 GB (before=13.786 GB, after=13.773 GB, peak=16.398 GB, 6504 samples)

Should I change the peak delta threshold to a little higher?

lk-chen:
> The Tests are failing with : AssertionError: Peak device memory during load_weights(Qwen/Qwen3-30B-A3B-Instruct-2507) exceeded threshold: peak_delta=2.612 GB > max=2.000 GB (before=13.786 GB, after=13.773 GB, peak=16.398 GB, 6504 samples)
> 
> Should I change the peak delta threshold to a little higher?

cc @kyuyeunk @rohan-reddy I'm not sure this is expected

rohan-reddy:
Would need to dig into it further but it looks like a memory regression
Is Qwen 30B very close to the HBM limit (is this v6e-8)? I thought it fits comfortably. Maybe we want to check a tight HBM fit model on this change like Deepseek R1 on v7, checking no OOM during loading and no decrease in KV cache size?

kyuyeunk (approved):

amanseervi:
@rohan-reddy the specific test failing is : `tests/models/jax/test_qwen3_moe.py::TestQwen3MoeForCausalLM::test_model_loading[jax_dummy-0-1-Qwen/Qwen3-30B-A3B-Instruct-2507]`

which I believe is trying to load part of the model on 1 device.  Overall the 30B model loads fine even on v6e-4. 
Agreed that we should be checking on a model which just about fits on the devices, since the peak memory issue is only going to bite in those cases

rohan-reddy:
Got it, I think qwen3 235B FP8 on v6e-8, or deepseek r1 on v7, would be a good litmus test if we are worried about this

lk-chen:
it's just the "jax_dummy" case fail, and no failure from "skip_layers_model_loader_for_test"

I think it's due to "jax_dummy" loader does sharding [here](https://github.com/vllm-project/tpu-inference/blob/fb9a1206390cc287f365bba2231d2b8dfdd9c975/tpu_inference/models/jax/utils/weight_utils.py#L970). I suggest disable "jax_dummy" for [0-1](https://github.com/vllm-project/tpu-inference/blob/fb9a1206390cc287f365bba2231d2b8dfdd9c975/tpu_inference/models/jax/utils/weight_utils.py#L970) but keep [0-4][1-4][3-4]

PR #2271, opened 2026-04-16
Recolor question mark symbol

mtsokol:
Hi @RobMulla,

This patch changes ❓ to ❔ in the tables for a better visual effect.


RobMulla:
As discussed. This is a great idea - however the white ? does not show well in the light themes.

mtsokol:
Decided that❔isn't the best fit here. We can either:
- Use a different emoji,
- Have our own pngs for status symbols - here we have space to design sth suitable for light and dark mode.

PR #2278, opened 2026-04-16
Avoid recurrent_state and conv_state copy to optimize GDN perf

helloworld1:
# Description

Avoid recurrent_state and conv_state copy to optimize GDN perf. 

This significant improved E2E decoding performance by about 60% for 1K/8K qwen3.5 case.

# Tests
Tested offline prompts and gives exactly the same output as before.
Benchmarked the performance and no regression on all cases.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/ragged_conv1d_jax.py
+++ b/tpu_inference/layers/common/ragged_conv1d_jax.py
@@ -17,7 +17,8 @@ import jax
 import jax.numpy as jnp
 
 
-@jax.jit(static_argnames=("kernel_size", ), donate_argnames=("conv_state", ))
+# Donate conv_state to avoid "copy" op by XLA
+@jax.jit(donate_argnames=("conv_state", ), static_argnames=("kernel_size", ))
 @jax.named_scope("ragged_conv1d_jax")
 def ragged_conv1d(
     x,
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/common/ragged_conv1d_jax.py
```diff
@@ -13,18 +13,22 @@
 # limitations under the License.
 """Ragged convolution Jax implementation."""
 
+import jax
 import jax.numpy as jnp
 
 
+@jax.jit(static_argnames=("kernel_size", ), donate_argnames=("conv_state", ))
```
jrplatin: Why do we need to JiT this separately? Isn't this already within the JiT boundary?

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -572,6 +572,18 @@ def ragged_gated_delta_rule_decode_only(
     return updated_recurrent_state.astype(recurrent_state.dtype), outputs
 
 
+@jax.jit(
```
jrplatin, line 576: Same comment here?

tpu_inference/layers/common/ragged_conv1d_jax.py
helloworld1: Added more comments here

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
helloworld1: Added more comments here

jrplatin (approved):

PR #2283, opened 2026-04-16
Instrument batch stats collection

Caslyn:
# Description

This PR introduces continuous batch composition logging by adding a new `AggregatedStatsLogger` class and integrating it into the existing `PhasedBasedProfiler` pipeline. 

**Why `AggregatedStatsLogger` is needed:**
We need longitudinal visibility into the server's behavior over its entire lifespan. By continuously logging the composition of *every* batch (number of prefill/decode tokens, padding overhead, total scheduled tokens) to a `.jsonl` file, we can perform long-term analysis on serving efficiency, traffic patterns, and padding waste. It periodically flushes these statistics to local storage or GCS.

**Why it is implemented as a separate class from `PhasedBasedProfiler`:**
1. **Separation of Concerns:** `PhasedBasedProfiler` acts as a state machine designed to trigger deep hardware profiling (`jax.profiler`) under specific threshold conditions. `AggregatedStatsLogger`, on the other hand, is purely responsible for continuous, low-overhead data collection, file I/O, and external storage syncing.
2. **Modularity & Maintainability:** Encapsulating the temporary file management, `atexit` cleanup hooks, and non-blocking `subprocess` calls for GCS synchronization keeps the `PhasedBasedProfiler` clean and focused solely on JAX tracing logic.

**Specific Implementation Details:**
- `AggregatedStatsLogger` writes stats in JSONL format and periodically syncs to GCS (using non-blocking `subprocess.Popen`).
- It is instantiated inside `PhasedBasedProfiler` only for the primary worker (`worker_rank == 0`) and when enabled via the `ENABLE_AGGREGATED_STATS_LOGGER` environment variable.

# Tests

- [x] Verified that `all_batches_stats_YYYY_MM_DD_HH_MM_SS.jsonl` is correctly generated locally when enabled.
- [x] Tested GCS syncing behavior to ensure it flushes properly at the defined `flush_interval` without blocking the main inference loop.
- [x] Verified `atexit` hooks cleanly upload the remaining logs upon server shutdown.

# Checklist


tpu_inference/envs.py
```diff
@@ -99,30 +99,42 @@ def _get_validated_env() -> str | None:
...
     Args:
         env_name: Name of the environment variable
         default: Default boolean value if not set
+        requires: List of environment variables that must be set if this is True.
     """
 
     def _get_bool_env() -> bool:
         value = os.getenv(env_name)
         if value is None or value == "":
-            return default
```
jrplatin, line 121: Why is this change needed?

tpu_inference/envs.py
```diff
@@ -260,6 +272,10 @@ def _get_str_list_env() -> list[str]:
                      ]),
     "MOE_ALL_GATHER_ACTIVATION_DTYPE":
     lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
+    "ENABLE_AGGREGATED_STATS_LOGGER":
+    env_bool("ENABLE_AGGREGATED_STATS_LOGGER",
+             default=False,
+             requires=["PHASED_PROFILING_DIR"]),
```
jrplatin: I wonder if it's useful to have the stats logging _without_ the need for phased-based profiling -- how hard would that be to implement?

tpu_inference/runner/utils.py
```diff
@@ -284,30 +287,130 @@ def determine_phase_from_batch_composition_stats(
...
+        os.makedirs(base_dir, exist_ok=True)
+        return target, target
+
+    def _sync_to_gcs(self,
+                     local_file: str,
+                     target_file: str,
+                     blocking: bool = False) -> None:
+        """Helper to sync local file to GCS."""
+        if target_file.startswith("gs://") and os.path.exists(local_file):
+            cmd = ["gcloud", "storage", "cp", local_file, target_file]
```
jrplatin: Is this the best way to do it? Is there not a pythonic way? Also worried about auth failures potentially -- have you tested this?

tpu_inference/runner/utils.py
```diff
@@ -284,30 +287,130 @@ def determine_phase_from_batch_composition_stats(
...
+            if blocking:
+                subprocess.run(cmd,
+                               stdout=subprocess.DEVNULL,
+                               stderr=subprocess.DEVNULL)
+            else:
+                subprocess.Popen(cmd,
+                                 stdout=subprocess.DEVNULL,
+                                 stderr=subprocess.DEVNULL)
+
+    def log(self, batch_composition_stats: dict) -> None:
```
jrplatin, line 366: Can you add docstrings (with a brief description) to each function here please? Thanks!

tpu_inference/runner/utils.py
```diff
@@ -284,30 +287,130 @@ def determine_phase_from_batch_composition_stats(
...
         current_phase: The current phase.
+        worker_rank: The rank of the current worker process.
+        aggregated_stats_logger: An instance of AggregatedStatsLogger to log stats continuously.
     """
 
-    def __init__(self, profile_dir: str):
+    def __init__(self,
+                 profile_dir: str,
+                 worker_rank: int = 0,
+                 flush_interval: int = 100):
```
jrplatin, line 452: Is this arbitrary or how did you land on 100?

tpu_inference/envs.py
Caslyn: I don't think it will be difficult to separate them. The `AggregatedStatsLogger` is coupled with the `PhaseBasedProfiler` because it ingests the same batch stats data at every step and dumps its collected info into the `PROFILE_DIR`. But decoupling them makes sense since they serve different ends. I can do this in this PR or a subsequent one, if you have a preference?

tpu_inference/envs.py
Caslyn: This enables the added code on L#131-L136, which enforces the dependency between two environment variables. This will error if `ENABLE_AGGREGATED_STATS_LOGGER` is set without `PHASED_PROFILING_DIR`. This requirement is needed because the `AggregatedStatsLogger` is currently baked into the `PhaseBasedProfiler` and uses its directory, but I would back out this change if we decouple the two per [this thread](https://github.com/vllm-project/tpu-inference/pull/2283/changes#r3190773526).

tpu_inference/runner/utils.py
Caslyn: Mostly arbitrary, but maybe reasonable? This will flush the aggregated stats file every 100 steps/batches that are recorded (and the file will get flushed before it's closed as well). I keep in mind higher values lead to more buffering and lower values lead to CPU overhead, but otherwise I'm open to changing it.

tpu_inference/runner/utils.py
Caslyn: Done - 20fa5ef6a7bf18f117ea407b5f33d4aee4a56df0

tpu_inference/runner/utils.py
Caslyn: Tested, yes :) This will fail silently if the host environment isn't authenticated, and the user will need to `gcloud auth login` on the host machine (though I think for GCE/GKE this information automatically gets propagated). efe629cd597f60f628e83564d49eed470d47a852 uses the google cloud API in a hopefully more idiomatic way, and it will also inspect the environment for credentials.

tpu_inference/envs.py
jrplatin: I think it's fine for now!

jrplatin (approved):

PR #2286, opened 2026-04-16
PhaseBasedProfiler supports PREFILL_ONLY and DECODE_ONLY phases

Caslyn:
# Description

This PR introduces two new inference phases, `PREFILL_ONLY` and `DECODE_ONLY`, to provide finer-grained categorization of batch compositions. 

**Details and Context:**
- **Why is this change being made?** Previously, pure prefill batches (100% prefill tokens) and pure decode batches (100% decode tokens) were lumped into the broader `PREFILL_HEAVY` and `DECODE_HEAVY` buckets. Separating them out allows for more precise and granular profiling.
- **The problem being solved:** When analyzing performance profiles or hardware utilization, pure phases often exhibit different execution characteristics compared to mixed "heavy" phases (e.g., a batch that is 95% prefill and 5% decode). This change allows the `PhasedBasedProfiler` and external telemetry scripts to isolate and analyze these pure phases independently.
- **Specific implementation:**
  - Added `PREFILL_ONLY` and `DECODE_ONLY` to the `InferencePhase` enum in `tpu_inference/runner/utils.py`.
  - Updated `determine_phase_from_batch_composition_stats` to return these new phases when the prefill ratio is exactly `1.0` or `0.0` respectively. 
  - Refactored the `elif` chain in `determine_phase_from_batch_composition_stats` to use early returns for cleaner control flow.
  - Registered the new phases in `PhasedBasedProfiler`'s `inference_phase_seen` dictionary so they are actively traced during profiling.

# Tests

I updated the existing unit tests in `tests/runner/test_utils.py` to verify the new phase logic.

Commands to reproduce locally:
```bash
pytest tests/runner/test_utils.py::test_determine_phase_from_batch_composition_stats
pytest tests/runner/test_utils.py::test_phased_profiler_handles_all_phases

PR #2287, opened 2026-04-16
Add batch number and phase to stats

Caslyn:
# Description

This PR enriches the runtime statistics tracked by `get_batch_composition_stats` to include the global `batch_no` and the dynamically evaluated inference `phase` for each step. 

**Details and Context:**
- **Why is this change being made?** Before this change, offline statistics processing lack temporal ordering data or explicit phase indicators. 
- **The problem being solved:** By injecting the `batch_no` and `phase` directly at the time of execution in the TPU worker, we reduce the complexity of downstream analytics and ensure that any changes to phase-determination thresholds are automatically reflected in the logs without needing to update auxiliary tools.
- **Specific Implementation Details:**
  - Added a `batch_counter` state to `TPUModelRunner`, which increments sequentially just prior to calling the telemetry function.
  - Passed `batch_no` into `get_batch_composition_stats`.
  - The returned `stats` dictionary now natively calls `determine_phase_from_batch_composition_stats(stats).name` to append the textual representation of the current phase.

# Tests

Updated the `test_get_batch_composition_stats` test in `tests/runner/test_utils.py` to:
1. Supply a dummy `batch_no` argument.
2. Assert that both `batch_no` and `phase` are calculated and populated correctly in the output dictionary.

Commands to reproduce locally:
```bash
pytest tests/runner/test_utils.py::test_get_batch_composition_stats


Caslyn:
FYI the PR diff includes changes in https://github.com/vllm-project/tpu-inference/pull/2286, and once that gets merged I can rebase this so this PR will be easier to review, if folks prefer to wait. (I guess `gh stack` can't stack existing or draft PRs)

tpu_inference/runner/utils.py
jrplatin, line 199: Can you update docstrings please? Thanks!

jrplatin:
Yep, let's get that first PR in first and then we can merge, thanks!

jrplatin (approved):

tpu_inference/runner/utils.py
Caslyn: Done - bd69040

PR #2294, opened 2026-04-17
Add ragged scatter for unpermute in MOE

gxd3:
# Description

Add ragged scatter kernel for unpermute in MOE.

The interface of the ragged-scatter kernel:
```
def ragged_scatter(
    x: jax.Array, indices: jax.Array, start: jax.Array, end: jax.Array
) -> jax.Array:

  This function performs a gather operation equivalent to `x[indices]` for
  indices that fall within the range `[start, end)`. For indices outside this
  range, the behavior is undefined.
```

# Tests
Run the DS-V3 1k/8k benchmark e2e: 2122 -> 2392 t/s/c (12.7% improvement E2E).
Quality verified with MMLU: https://paste.googleplex.com/5509706039820288

The unpermute reduce from ~260us to ~45us
after: https://screenshot.googleplex.com/jLcVFdeMANWnEwK
before: https://screenshot.googleplex.com/7bdiZPWSpvDbsHc

# Checklist


tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -194,8 +191,21 @@ def moe_gmm_local(
                                group_offset,
                                preferred_element_type=x.dtype)
 
+        if parallelism == "ep":
```
kyuyeunk: instead of checking `parallelism == "ep"`, similar to line 213 or 153, can you use `local group < global group` check to trigger the code? this approach will be more generic.

gxd3: SG, updated.

kyuyeunk (approved):

tpu_inference/kernels/sparse_core/ragged_scatter.py
```diff
@@ -0,0 +1,441 @@
...
+        # Only execute the kernel instance when there is a least one valid row
+        # to process.
+        @pl.when(row_tile_start < total_num_rows)
+        def _():
+            inner_kernel(block_id, core_id, col_id)
+
+    kernel_wrapper()
+
+
+def calculate_col_size(hidden_size: int) -> int:
```
kyuyeunk, line 266: oh this needs fixing: https://github.com/vllm-project/tpu-inference/pull/2224

kyuyeunk (changes requested):

tpu_inference/kernels/sparse_core/ragged_scatter.py
gxd3: Good catch!
Updated.

kyuyeunk (approved):
thanks you for the quick fix!

PR #2064, opened 2026-03-27
Reorganization of ToC and docs linting

mtsokol:
Hi @RobMulla,

Here's a new structure of the docs for vLLM-TPU. It's divided into three main sections:
- Getting Started - a minimal set of actions to run an inference tasks with vLLM on a TPU.
- Features and Usage - for users with supported HF models, deploying it on a single TPU manually, or GCE cluster.
- Developers Guide - for power users / developers wanting to implemented a missing feature or run their custom model.

Additionally this PR includes:
- Linting markdown files (fixing headers, html links, codeblocks language annotation).
- `What are you trying to do today?` cards on the main page.
- Splitting `recommended_models_and_features` into `recommended_models` and `recommended_features` pages.

PR #2301, opened 2026-04-17
Add contributing guides to the docs page

mtsokol:


mtsokol:
@RobMulla I resolved conflicts - this one is ready from my side!

RobMulla (commented):
Looks good to me!

RobMulla:
@mtsokol - It looks like this is failing pre-commit hooks due to markdown formatting. Can you edit and confirm the pre-commit passes locally?

mtsokol:
@RobMulla pymarkdown lint config required an additional option to accept 4 spaces indentation for nested lists.
We need 4 instead of 2 because mkdocs supports 4 spaces (2 isn't enough to make a nested list).
Now it's ready.

.pre-commit-config.yaml
```diff
@@ -56,7 +56,7 @@ repos:
   rev: v0.9.29
   hooks:
   - id: pymarkdown
-    args: [fix]
+    args: [--set, 'plugins.ul-indent.indent=$#4', fix]
```
RobMulla, line 59: I'd like to avoid editing the pre-commit config if possible. Could we try instead try adding something like this to the pyproject.yml and see if it works?

```
[tool.pymarkdown]
plugins.ul-indent.indent = 4
```

Also would like to make sure this change won't impact all the existing markdown files passing pre-commit.

Would it be easier just to fix the markdown files to conform?

mtsokol:
@RobMulla I updated the branch - it's ready now!

RobMulla (commented):
LGTM

RobMulla (approved):
LGTM

bvrockwell (approved):

mtsokol:
@bvrockwell I don't think failing jobs are related (as other PRs fail with similar ones) so I think this needs to be merged manually.

bvrockwell:
@QiliangCui @RobMulla seems to be blocked?

mtsokol:
Replaced by https://github.com/vllm-project/tpu-inference/pull/2064

PR #2324, opened 2026-04-20
GLM-5.1-FP8 MLA multi-host + multi-host weight loader + FP8 MoE direct path

yiqiliu2:
# Description

Related: #2339.

Adds TPU v4-64 multi-host support for `MLAAttention`-based FP8 MoE models (GLM-5.1-FP8) across `(EP, TP, PP)` configurations, plus three FP8 MoE weight-prep improvements the same end-to-end run depends on (direct FP8 path, 3D / 4D `w2_scale` shape handling, GMM_TP `w2_weight_scale` global-axis reconstruction).

**Scope.** This PR targets **correctness** — unblocking end-to-end serving of GLM-5.1-FP8 on v4-64 and exposing the same path to other FP8 MLA MoE models. Performance is currently below expectation; performance optimization will land in follow-up PRs.

## Contents

- [Fixes likely not TPU v4-specific](#fixes-likely-not-tpu-v4-specific)
- [1. MLA attention on TPU v4](#1-mla-attention-on-tpu-v4)
- [2. Multi-host weight loader](#2-multi-host-weight-loader)
- [3. FP8 MoE direct path and scale-shape robustness](#3-fp8-moe-direct-path-and-scale-shape-robustness)
- [Behavior by scenario](#behavior-by-scenario)
- [Opt-in switches introduced](#opt-in-switches-introduced)
- [Tests](#tests)
- [Known limitations](#known-limitations)

---

# Fixes likely not TPU v4-specific

The table below lists the fixes in this PR that we *believe* are not TPU v4-specific — each one is a bug fix or feature whose underlying cause (multi-host sharding, vLLM's MLA path, the FP8 MoE loader, etc.) should recur on v5e / v5p / v6e / v7 too. We've only actually exercised them on v4-64 because that is the hardware we can access; reviewers with access to other generations are welcome to confirm. Sections 1.1 – 1.3 and the `tpu_generation() == 4` gates inside §3.2 are intentionally excluded — those are v4-only workarounds. Where non-v4 kernel behavior matters (e.g. FP8 MXU support on v5e / v5p / v6e), that caveat is noted inline in the relevant subsection.

| Fix | Primary file(s) | See |
| --- | --- | --- |
| `expert`-axis all-gather before `o_proj` for MLA under expert parallelism | `layers/vllm/backends/flash_attn_mla.py` | §1.4 |
| MLA weight absorb under `VllmFp8LinearMethod` (`kv_b_proj` PWAL + dequant path) | `layers/vllm/custom_ops/mla_attention.py` | §1.5 |
| MLA startup check relaxation (DP attention no longer required) | `platforms/tpu_platform.py` | §1.6 |
| GLM MoE architecture registry + PP disable list | `models/common/model_loader.py` | §1.7 |
| MLA cross-shard Q/K all-gather (fixes empty output on shard-spanning prompts and garbage output at small token buckets) | `layers/common/attention_interface.py` | §1.8 |
| Mesh-aware EP expert filter (multi-host) | `models/vllm/vllm_model_loader.py` | §2.1 |
| Process-local weight materialization for `jax.Array` construction | `layers/common/utils.py` + call sites | §2.2 |
| `FORCE_LOGICAL_RESHAPE` consistency between loader and runner | `models/vllm/vllm_model_loader.py` | §2.3 |
| PP group metadata patching for multi-host PP | `worker/tpu_worker.py` | §2.4 |
| Ray executor compiled-DAG bypass for PP > 1 | `executors/ray_distributed_executor.py` | §2.5 |
| KV cache config indexed by JAX `process_index` | `worker/tpu_worker.py` | §2.6 |
| `--load-format tpu_streaming_loader` for large-checkpoint multi-host loads | `models/vllm/tp_selective_loader.py` (new) | §2.7 |
| Runtime diagnostics: HBM fallback, `CompilationTimes` compat, env-var whitelist | `utils.py`, `worker/tpu_worker.py`, `platforms/tpu_platform.py` | §2.8 |
| TP-selective loader unsupported-config guard (`EP < process_count`) | `models/vllm/vllm_model_loader.py` | §2.9 |
| Direct FP8 MoE path (opt-in) | `layers/common/process_weights/moe_weights.py`, `layers/vllm/quantization/fp8.py` | §3.1 |
| `w2_scale` 3D / 4D dispatch in `tensor_parallel_gmm` + native 2D block scale in `gmm_v2` | `layers/common/fused_moe_gmm.py`, `kernels/megablox/gmm_v2.py` | §3.2 |
| GMM_TP `w2_weight_scale` global-axis reconstruction under streaming vs default loader | `layers/common/process_weights/moe_weights.py` | §3.3 |

Each topic below is organized as **Reason**, **How it works**, **Changes** so reviewers can read the intent first and the file pointers last.

---

# 1. MLA attention on TPU v4

Everything needed to make MLA attention produce correct output on v4. §1.1–1.3 are v4-only kernel tuning (FP8-on-MXU fallback, VMEM-fit block sizes, fp32 softmax). §1.4–1.8 are bug fixes in the MLA code path itself and would recur on any TPU generation with a similar mesh / quantization config.

## 1.1 FP8 operands on the v4 MXU

- **Reason.** Mosaic v4 rejects FP8 right-hand-side matmuls (`CompileTimeMosaicUnsupportedRhsType`), so the v2 MLA kernel cannot consume FP8 weights directly.
- **How it works.** FP8 stays resident in HBM for storage; the cast to bf16 fires per-tile inside Pallas so only the active tile becomes bf16. Non-v4 generations keep the direct-FP8 path.
- **Changes.** `kernels/mla/v2/kernel.py` — cast FP8 tile to bf16 in Pallas VMEM before each MXU dot, guarded by a v4 check.

## 1.2 v4 block sizes

- **Reason.** v4 VMEM is ~16 MB vs 32 MB+ on v5/v6; the default block sizes leave no room for scratch and spill.
- **How it works.** The new case selects tile dimensions that fit v4's VMEM. Other generations fall through to the existing default.
- **Changes.** `kernels/ragged_paged_attention/v3/kernel.py` and `tuned_block_sizes{,_hd64}.py` — `case 4` with smaller `num_kv_pages_per_block` / `num_queries_per_block`. `layers/common/attention_interface.mla_attention` — v4 arm picks the matching sizes.

## 1.3 fp32 softmax on v4

- **Reason.** On deep stacks (GLM-5.1 78 layers) bf16 softmax scores saturate after a handful of MLA layers and attention outputs degrade.
- **How it works.** Higher-precision scores prevent saturation in the exponent; the rest of the kernel is unchanged. v5/v6 keep the kernel default.
- **Changes.** `layers/common/attention_interface.py` — v4 arm of `mla_attention` passes `s_dtype=jnp.float32`.

## 1.4 `expert`-axis all-gather before `o_proj`

- **Reason.** When expert parallelism splits attention heads across `('model', 'expert')`, the `W_UV` einsum output is sharded on both axes, but `o_proj` (RowParallelLinear) only all-reduces `model`. Each rank sees only `1/EP` of the attention output and residuals collapse.
- **How it works.** The sharding constraint triggers an all-gather along the `expert` axis, so `o_proj` receives the full head-dim output. Non-EP meshes skip the gather (no-op).
- **Changes.** `layers/vllm/backends/flash_attn_mla.py` — `with_sharding_constraint(P(None, 'model'))` before `o_proj`, gated on `get_mesh_shape_product(mesh, ATTN_HEAD) > mesh.shape['model']`.

## 1.5 MLA absorb under `VllmFp8LinearMethod`

- **Reason.** `MLAAttention.process_weights_after_loading` (parent) needs CPU-resident fp8 weights to absorb, but three things block it on `kv_b_proj`:
  - vLLM invokes the child's PWAL first and materialises a multi-host `jax.Array` whose local-shape slice `general_device_put` then rejects.
  - `VllmFp8LinearMethod.__init__` forces `use_marlin=True` to bypass upstream `scaled_quantize`, but the same flag routes absorb through a generic identity-matmul fallback that feeds `torch.eye` into `jax_view` and asserts.
  - `scaled_dequantize` returns a plain `torch.Tensor` that `jax_view` rejects.
- **How it works.** The no-op lets the parent's absorb drive the weight transform end-to-end on CPU fp8 tensors; the `use_marlin` flip steers to the "simple fp8" branch that handles CPU fp8 directly; `t2j` handles the post-dequant plain tensor.
- **Changes.** `layers/vllm/custom_ops/mla_attention.py` — patch `kv_b_proj.quant_method.process_weights_after_loading` to a no-op in `VllmMLAAttention.__init__`; temporarily flip `use_marlin=False` around `get_and_maybe_dequant_weights`, restore on `finally`; post-dequant `torch.Tensor`s go through `t2j(tensor, use_dlpack=False)` instead of `jax_view`.

## 1.6 MLA startup check

- **Reason.** With 1.4 in place MLA correctness no longer depends on DP attention, but the startup check was asserting it.
- **How it works.** Users who already set `enable_dp_attention=true` keep their previous behavior.
- **Changes.** `platforms/tpu_platform.py` — only require `NEW_MODEL_DESIGN=1`.

## 1.7 GLM registry

- **Reason.** JAX has no GLM DSA implementation, and GLM's JAX-side PP plumbing is not wired up.
- **How it works.** The preferred-architectures registry routes inference through the vLLM path; the PP disabled list prevents attempts to wire up an unsupported PP for this arch.
- **Changes.** `models/common/model_loader._VLLM_PREFERRED_ARCHITECTURES` gains `GlmMoeDsaForCausalLM`; `_PP_DISABLED_MODELS` gains `Glm4MoeLiteForCausalLM`.

## 1.8 MLA cross-shard Q / K all-gather

- **Reason.** MLA's `shard_map` splits Q / new-K on the token axis via `P(ShardingAxisName.MLP_TENSOR, ...)` — a multi-axis product that goes up to 32-way on v4-64 TP=4 EP=8 — while `cu_q_lens` / `kv_lens` come in via `P(ShardingAxisName.ATTN_DATA)` which is 1-way (replicated) for this config. Each shard's kernel therefore computes `q_span = kv_len - q_len + bq_idx*bq_sz + local_iota` from its shard-local iota 0, treating local `q[0]` as global position 0.

  This only produces correct attention when every real token of a sequence falls into the first shard. As soon as real tokens span shards — a 17-token prompt at bucket=512 on a 32-way token mesh, or any prompt at bucket < MLP_TENSOR-size — the shards past the first apply the wrong causal mask and the last real query's hidden state is corrupted. The sampler then picks token 0 (`'!'` in the GLM tokenizer) for every bucket-below-512 test prompt, or lands on `<|endoftext|>` for longer prompts that cross the shard boundary.

  This manifests on any TPU generation with a multi-axis token-sharded MLA mesh.
- **How it works.** Inside the shard_map body, all-gather `q / q_rope / new_kv_c / new_k_pe` along the `MLP_TENSOR` axis tuple so every shard sees the full prefix; call the kernel with the full tensors; then `lax.dynamic_slice` the full attention output back to this shard's token range before returning, keeping `out_spec = P(MLP_TENSOR, None, None)` valid. The kv cache write via `input_output_aliases` uses the replicated page_indices — with the same full new_kv on every shard, each shard writes an equivalent copy into its local cache slice and subsequent decode reads are self-consistent. When the MLP_TENSOR product is 1 (single-host or non-sharded meshes) the guard skips the all-gather entirely.
- **Changes.** `layers/common/attention_interface.mla_attention` — compute `_mlp_tensor_axes` and `_mlp_tensor_size` from the mesh; inside `_mla_ragged_paged_attention` do `jax.lax.all_gather(..., axis_name=_mlp_tensor_axes, axis=0, tiled=True)` for each of `q / q_rope / k / k_rope` when the product is > 1, then `lax.dynamic_slice` the output down to the shard's token range using a flat shard index built from `jax.lax.axis_index` over the same axes. `layers/common/utils.py` — add the missing `init_logger` at module scope so the `general_device_put` tp_selective_loader ImportError path (§2.2) logs its intended warning instead of raising `NameError`.
- **Cost.** ~2 × `max_num_tokens` × `hidden` bytes of cross-device communication per attention call (`q + q_rope + k + k_rope`). At `max_num_batched_tokens=4096` on v4-64 this is 4–8 MB/shard/call through ICI — negligible against the weight-load bandwidth and the MoE gmm all-gathers.

---

# 2. Multi-host weight loader

Correctness and memory fixes for multi-host serving: per-host TP / EP weight slicing (§2.1, §2.3, §2.7), keeping JAX and Ray / vLLM distributed state in sync across frameworks (§2.4–§2.6), runtime diagnostics and a guard against unsupported configs (§2.8, §2.9). §2.2 is the shared plumbing — each host supplies only its local slice of a sharded `jax.Array` instead of the full tensor, so a ~744 GB FP8 MoE checkpoint no longer has to fit in per-host RAM.

## 2.1 Mesh-aware EP expert filter

- **Reason.** The previous filter early-returned on single-host and projected onto a single `expert` axis. In multi-host with a full `attn_dp × attn_dp_expert × expert` mesh product every rank got the wrong expert ids; on single-host all ids dropped.
- **How it works.** The function reconstructs the full mesh with `mesh_utils.create_device_mesh` (or a row-major fallback), flat-iterates `(attn_dp, attn_dp_expert, expert, model)` in mesh order, and emits the ordered list of global expert ids owned by each local device. The order matters because `make_array_from_process_local_data` slices the host buffer in the same order.
- **Changes.**
  - `models/vllm/vllm_model_loader.py` — `_compute_mesh_aware_local_expert_ids`. Caller `_init_tpu_ep_weight_filter` loses the `process_count <= 1` early-return.
  - `models/vllm/vllm_model_wrapper.py` — the legacy code unconditionally disabled `enable_expert_parallel` during weight load to avoid vLLM's sharding-aware loader. Under `NEW_MODEL_DESIGN` we need expert parallel **on** during load so the MoE backend and sharding config stay aligned with the runtime mesh; otherwise this filter never fires. The disable is preserved for the legacy path via `if not envs.NEW_MODEL_DESIGN`.

## 2.2 Process-local materialization

- **Reason.** On multi-host, `jax.device_put` needs the full global tensor on every process. A 744 GB FP8 MoE checkpoint does not fit in host RAM.
- **How it works.** Each rank supplies only its local slice plus the full-tensor shape; JAX reconstructs the global sharded `jax.Array` without any host-side collective. Single-host keeps `jax.device_put`.
- **Changes.** `layers/common/utils.general_device_put` gains a `global_shape` kwarg; when passed, `make_array_from_process_local_data` is used. Call sites: `models/vllm/vllm_model_wrapper.py`, `utils.py`, `layers/common/process_weights/linear_weights.py`, `layers/common/process_weights/moe_weights.py`. The Linear caller only passes `global_shape` for the `weight` field since scale / zero_point / bias have their own shapes. Symmetric helpers on the read side — `runner/tpu_runner.py::_jax_array_to_host` / `_jax_replicated_array_to_host` — handle the `is_fully_addressable=False` case when pulling sampled tokens / logprobs back to host.

## 2.3 FORCE_LOGICAL_RESHAPE consistency

- **Reason.** The runner has always supported `FORCE_LOGICAL_RESHAPE=1` to bypass the physical-optimal mesh reorder; the loader did not. With the env set, loader and runner would pick different device orderings and misalign expert ids.
- **How it works.** Both sites pick `np.array(devices).reshape(mesh_shape)` when the env is set, so the same process-to-device mapping is observed on both sides.
- **Changes.** `models/vllm/vllm_model_loader.py` — honor `FORCE_LOGICAL_RESHAPE=1` identically to `runner/tpu_runner.py`.

## 2.4 PP group metadata

- **Reason.** vLLM initialises its distributed group with `pp=1` (single-process gloo), so `rank_in_group == 0` on every host. `make_layers` / `is_first_rank` / `is_last_rank` return wrong values for PP > 1.
- **How it works.** The patched values make `get_pp_indices` and `is_first_rank` / `is_last_rank` report the real per-host PP rank. PP = 1 is untouched.
- **Changes.** `worker/tpu_worker.py` — when `pipeline_parallel_size > 1`, override `rank_in_group`, `world_size`, `rank`, `ranks`, and `parallel_config.rank` using the JAX process index.

## 2.5 Ray executor dispatch

- **Reason.** JAX's intermediate-tensor transfer is incompatible with Ray's compiled-DAG channels; PP ≥ 2 deadlocks inside the DAG.
- **How it works.** PP > 1 falls back to direct-actor calls; PP = 1 keeps the compiled DAG.
- **Changes.** `executors/ray_distributed_executor.py` — `use_ray_compiled_dag = (pp_size <= 1)`.

## 2.6 KV cache config indexing

- **Reason.** `RayWorkerWrapper.initialize_from_config` used Ray's `global_rank` to pick the per-rank KV cache config, but Ray scheduling order does not match JAX `process_index` order, so ranks could pick the wrong config on multi-host.
- **How it works.** The same index is used on both sides of the JAX ↔ Ray boundary, so the config per rank is consistent.
- **Changes.**
  - `worker/tpu_worker.py` — index by `self.worker.rank` (JAX `process_index`).
  - Same JAX ↔ Ray ordering issue shows up for the EP rank: vLLM / Ray's actor order doesn't match JAX's `process_index`, so the loader's per-host EP slice would be keyed wrong. The Ray executor (`executors/ray_distributed_executor.py`) now pre-sets `TPU_EP_RANK` on each actor's env; `worker/tpu_worker.py` falls back to `jax.process_index()` if the executor didn't set it; `models/vllm/vllm_model_loader.py` reads it for EP shard selection. Single-host keeps the `jax.process_index()` fallback.

## 2.7 `--load-format tpu_streaming_loader`

- **Reason.** For multi-host runs on large FP8 MoE checkpoints (whether pure TP or EP), vLLM's default loader reads the full checkpoint into CPU RAM on every host before JAX shards it, OOMing host RAM on ~744 GB models. The EP weight filter only skips non-local experts; it does not slice TP-parallel non-MoE weights, so those still drive host RAM up.
- **How it works.** Each host reads only its own TP slice via `safetensors.safe_open(f).get_slice(name)[slice_obj]` and drops the CPU buffer as soon as the tensor is on HBM. JAX rebuilds the global sharded array from each host's local slice via 2.2. Gated by `--load-format tpu_streaming_loader` or `TPU_TP_SELECTIVE_LOAD=1`; unset preserves upstream behaviour.
- **Changes.** `models/vllm/tp_selective_loader.py` (new, ~1000 lines) registered as `tpu_streaming_loader`. Pieces: plan builder, resized `param.data`, monkey-patched `FusedMoE._load_w{13,2}` and `_ColumnvLLMParameter` / `RowvLLMParameter` to honor `is_sharded_weight`, plus a registry that feeds `tp_full_shape` (keyed by `tensor.data_ptr()`) back into `general_device_put`.

## 2.8 Runtime diagnostics

- **Reason.** Four runtime / compatibility shims plus the env-var whitelist for Ray workers: the HBM accounting fallback, the "returned 0 for N devices" log fix, the `CompilationTimes` import fallback, the per-worker chip count under device sub-slicing, and the env-var list itself.
- **How it works.** The HBM fallback iterates live arrays and counts per-device shard bytes. The `CompilationTimes` fallback keeps `max(compilation_times)` working either way. The env-var list is the vLLM mechanism for making env vars visible inside Ray workers.
- **Changes.**
  - `utils.hbm_usage_bytes` — fall back to `jax.live_arrays()` when `memory_stats()['bytes_in_use']` returns 0 (true on several PJRT runtimes).
  - `utils.hbm_usage_gb` — record `n_zero` before filling zeros so the "returned 0 for N devices" log is correct.
  - `worker/tpu_worker.py` — `@dataclass(order=True)` `CompilationTimes` fallback for vLLM builds that no longer export the symbol; `determine_available_memory()` filters `self.devices` by `process_index == jax.process_index()` so the per-worker chip count stays correct under device sub-slicing (equivalent to the old `jax.local_devices()` path in the normal multi-host case); live-arrays breakdown probe gated behind `TPU_HBM_PROBE=1` at DEBUG.
  - `platforms/tpu_platform.TpuPlatform.additional_env_vars` — add the new loader / debug knobs so they propagate into Ray worker processes.
  - `layers/vllm/quantization/mxfp4.py` — try/except fallback for upstream vLLM's `GptOssMxfp4Config` → `Mxfp4Config` rename. Without this, whichever vLLM pin mismatches the one name we import fails at quantization-registry import time and the whole server fails to start (GLM-5.1-FP8 included, even though it doesn't use mxfp4 itself).

## 2.9 TP-selective unsupported-config guard

- **Reason.** `EP < process_count` means multiple hosts share one EP slot; EP filter slices MoE weights but not TP-parallel non-MoE weights. Each co-located host loads the full TP=N non-MoE tensors, which may OOM on large models and silently produces wrong sharding on small ones.
- **How it works.** Raise `NotImplementedError` when `ep_total > 1 and ep_total < process_count and tp_global > 1` inside `_init_tp_weight_filter`. Full composition with the EP filter is a follow-up.
- **Changes.** `models/vllm/vllm_model_loader._init_tp_weight_filter`.

---

# 3. FP8 MoE direct path and scale-shape robustness

Loader-side fixes so the FP8 MoE kernel can consume checkpoint weights directly: skip the `FP8 → fp32 → FP8` dequant / requant roundtrip at load time (§3.1), handle either 3D or 4D `w2_scale` layouts at dispatch (§3.2), and reconstruct the global intermediate axis correctly under streaming vs default loaders (§3.3). The kernel-level v4 workarounds that sit underneath these dispatch changes are discussed inline in §3.1 / §3.2.

## 3.1 Opt-in direct FP8 path

- **Reason.** The FP8 MoE loader runs every checkpoint weight through `FP8 → fp32 → FP8` before the GMM kernel. On GLM-5.1-FP8 (256 experts × 75 MoE layers) the fp32 intermediate is the single largest contributor to startup time, and the round trip introduces quantization error the kernel would otherwise avoid by consuming the original FP8 values and their 2D block scales.
- **How it works.** The direct path hands FP8 + block scale straight to `process_moe_weights` → `shard_moe_weights` → kernel, which already has native 2D block-scale support (`scale_n_block_size` in `gmm_v2`). Both the direct and legacy branches run under `cpu_mesh_context()` so load-time transforms allocate on CPU, not TPU HBM. The chunk env var caps the per-expert chunk size on the legacy requantize path and has no effect on the direct path.
- **Changes.**
  - `layers/common/process_weights/moe_weights.process_fp8_moe_weights_direct` — returns the checkpoint FP8 tensor plus its 2D block scale unchanged, skipping the dequant / requant round trip.
  - `layers/vllm/quantization/fp8.VllmFp8MoEMethod` — routes its MoE weights through this path when `envs.MOE_SKIP_REQUANTIZATION` is set.
  - `envs.py` — declares `MOE_SKIP_REQUANTIZATION` (`VLLM_MOE_SKIP_REQUANTIZATION`, default false) and `MOE_REQUANTIZE_EXPERT_CHUNK_SIZE` (default 8).
  - The legacy chunked path rounds `expert_chunk_size` up to an equal divisor so every chunk reuses the jitted shape.
- **Kernel path by generation.** v4 reaches the kernel via a `tpu_generation() == 4` dequant-first branch in `gmm_v2` / `quantized_matmul/util.py` that casts FP8→BF16 per-tile in Pallas VMEM (needed because v4 MXU rejects FP8 matmul with Mosaic E2001). Non-v4 gens fall through to `jnp.matmul(fp8, fp8, ...)`. We have only exercised the direct path on v4-64. Per public docs v7 (Ironwood) is the first generation with native FP8 MXU; v6e is listed as emulating FP8, and v5e / v5p lack native FP8 entirely, so the non-v4 fallthrough may also hit Mosaic E2001 and require extending the v4 gate. Reviewers with v5e / v5p / v6e / v7 access are welcome to confirm which gate is correct on each.

## 3.2 `w2_scale` shape handling in `tensor_parallel_gmm`

- **Reason.** `w2_scale` can arrive in one of two shapes depending on how the checkpoint was produced: 4D `[num_experts, num_blocks_n, intermediate_size, 1]` or 3D `[num_experts, num_blocks_n, intermediate_size]`. The dispatcher previously assumed 4D; 3D checkpoints failed with `ValueError: in_specs too long`.
- **How it works.** The dispatcher builds the correct `in_specs` rank, and the kernel does per-block scalar lookup for 2D scales instead of per-element N expansion.
- **Changes.** `layers/common/fused_moe_gmm.tensor_parallel_gmm` — selects the `PartitionSpec` tuple from `w2_scale.ndim`. `kernels/megablox/gmm_v2.py` — native 2D block-scale support via `scale_n_block_size`; 3D `w2_scale` no longer needs expand-to-4D at dispatch. The FP8 dequant-first branches in `gmm_v2` and `kernels/quantized_matmul/util.py` are gated on `tpu_generation() == 4`; non-v4 gens fall through to native `jnp.matmul(fp8, fp8)`. Only v4 has been tested — per public docs v7 (Ironwood) is the first with native FP8 MXU, v6e emulates FP8, and v5e / v5p lack it, so the gate may need to extend to those gens after someone exercises them. `kernels/quantized_matmul/util.py` — accept scale shape `[n]` or `[n, 1]` (shows up when 3D block scales are sliced per expert).

## 3.3 GMM_TP `w2_weight_scale` global-axis reconstruction

- **Reason.** `shard_moe_weights` (GMM_TP path) reconstructed the global intermediate-axis size of `w2_weight_scale` by multiplying the local shape by `process_count`. This is correct only under `tpu_streaming_loader`, where the tensor is pre-sliced `1/process_count` per host. Under the default loader the tensor arrives full on every host, so the multiplication inflated `global_axis1` by `process_count`, making the downstream shard / repeat / replicate decision wrong by the same factor.
- **How it works.** Detect which regime applies via `tpsl.is_enabled()` and `process_count > 1`: multiply only when the tensor was pre-sliced; otherwise use the local shape directly as the global size.
- **Changes.** `layers/common/process_weights/moe_weights.shard_moe_weights` (GMM_TP branch) — `_is_pre_sliced` flag gates the multiplication. Single-host paths are unaffected (`process_count == 1`).

---

# Behavior by scenario

| Scenario | Behavior |
| --- | --- |
| Single host, no Ray, no EP, no PP, no MLA | Unchanged. |
| TPU v5e / v5p / v6e / v7 (any mode) | Unchanged — v4-specific branches are gated on `tpu_generation() == 4` and skipped on other gens. |
| Single-host MLA, any generation | `kv_b_proj` PWAL no-op and absorb fixes apply (bug fix). |
| Multi-host MLA, any generation | Q / K are all-gathered inside the shard_map body so cross-shard queries see the full prefix (§1.8). |
| Multi-host EP on v4 | Expert placement matches physical mesh slots; `o_proj` all-gathers `expert`; KV cache config keyed by `process_index`. |
| Multi-host PP | PP group metadata patched; compiled DAG bypassed. |
| `--load-format tpu_streaming_loader` passed | Weights stream into HBM per Linear layer. |
| `--load-format tpu_streaming_loader` not passed | Upstream default loader. |
| `VLLM_MOE_SKIP_REQUANTIZATION` unset (default) | Legacy `FP8 → fp32 → FP8` path. |
| `VLLM_MOE_SKIP_REQUANTIZATION=1` | Direct FP8 path; kernel consumes checkpoint FP8 + 2D block scale. |
| Checkpoint with 4D `w2_scale` | `tensor_parallel_gmm` dispatches 4D spec (as before). |
| Checkpoint with 3D `w2_scale` | `tensor_parallel_gmm` dispatches 3D spec; `gmm_v2` consumes 2D block scale natively. |
| New `TPU_*` env vars not set | Not consulted. |

---

# Opt-in switches introduced

| Switch | Purpose |
| --- | --- |
| `--load-format tpu_streaming_loader` (CLI) | Stream per-Linear shards to HBM; releases CPU buffer immediately. |
| `TPU_TP_SELECTIVE_LOAD=1` (env) | Same streaming loader via env for deployment scripts that cannot pass `--load-format`. |
| `TPU_TRUNCATE_LAYERS=N` (env) | Debug knob; truncate model to first N transformer blocks. |
| `TPU_GCS_WEIGHT_LOAD` (env, default `1`) | Picks the I/O path for weights on a gcsfuse mount: `1` = direct GCS HTTP Range GETs (bypass gcsfuse), `0` = read through the gcsfuse mount with `safetensors.safe_open`. Neither is universally faster — which wins depends on how the mount is configured (e.g. gcsfuse with `--enable-buffered-read` beats direct Range on our v4-64 pod; the default wins on a plain mount). |
| `FORCE_LOGICAL_RESHAPE=1` (env) | Bypass physical-optimal mesh reorder in both runner and loader. |
| `VLLM_MOE_SKIP_REQUANTIZATION=1` (env) | Activate the direct FP8 MoE path. |
| `MOE_REQUANTIZE_EXPERT_CHUNK_SIZE=N` (env, default 8) | Cap per-expert chunk on the legacy requantize path. |
| `TPU_MIN_TOKEN_BUCKET=N` (env, default 512) | Minimum token-padding bucket. With §1.8 in place, the compile-level floor (mesh product of `attn_dp × attn_dp_expert × model × expert`) is the true lower bound — e.g. 32 on v4-64 TP=4 EP=8 is safe. Pre-§1.8 buckets below 512 produced garbage; keeping the 512 default for upstream's target config until more configs are tested. |
| `TPU_HBM_PROBE=1` (env) | Enable the `jax.live_arrays()` HBM breakdown probe at DEBUG level. |
| `TPU_EP_RANK` (env, auto-set) | Set by the Ray executor on each actor to preserve EP shard order under multi-host; the worker falls back to `jax.process_index()` if unset. Do not override manually. |

---

# Tests

## Unit

```bash
pytest tests/models/common/test_model_loader.py -v
pytest tests/models/common/test_vllm_model_wrapper.py -v
pytest tests/runner/test_tpu_runner.py tests/runner/test_tpu_runner_dp.py tests/runner/test_tpu_runner_mesh.py -v
pytest tests/executors/test_ray_distributed_executor.py -v
pytest tests/layers/common/test_sharding.py -v
pytest tests/layers/common/test_moe_weights.py -v
pytest tests/layers/vllm/test_fp8.py -v
```

`test_moe_weights.py::test_direct_fp8_path` and `test_fp8.py::test_skip_requantization_env` exercise the new FP8 MoE path.

## End-to-end — TPU v4-64, GLM-5.1-FP8, TP=4 EP=8

Multi-host serving needs a Ray head + 7 workers before `vllm serve` can be started. The PR ships launcher scripts that do this; reviewers do not need to reproduce the Ray bring-up by hand.

```bash
# On the coordinator/head host:
MODEL_PATH=<path-to-GLM-5.1-FP8>         \
VENV=<python-venv-with-vllm-tpu-inf>     \
COORDINATOR_IP=<head-host-IP>            \
WORKERS="<ip1> <ip2> <ip3> <ip4> <ip5> <ip6> <ip7>" \
  bash scripts/multihost/tpu-v4/launch_glm51_tp4_ep8.sh
```

The script is a thin wrapper that exports the env defaults this config needs (`NEW_MODEL_DESIGN=1`, `VLLM_MOE_SKIP_REQUANTIZATION=1`, `TPU_TP_SELECTIVE_LOAD=1`, `TPU_GCS_WEIGHT_LOAD=0`, `ENABLE_EXPERT_PARALLEL=1`, `TP_SIZE=4`, `EP_SIZE=8`, `ENABLE_DP_ATTENTION=0`, `KV_CACHE_DTYPE=auto`) and delegates to `launch_ray_nodocker.sh`, which starts the Ray head, brings up the 7 workers, and runs `vllm serve` with the flags that map to the Known limitations below (`--kv-cache-dtype auto`, `--no-enable-prefix-caching`, `--load-format tpu_streaming_loader`). Required env vars are validated up front via `${VAR:?...}`; no pod-specific defaults are committed to the repo. `MODEL_PATH` must be reachable on every host — we use a gcsfuse mount at the same path on all 8 hosts.

If the worker hosts do not already have matching `tpu-inference` / vLLM checkouts, sync them first with `scripts/multihost/tpu-v4/rsync_workers.sh` (same required env vars plus `SSH_USER`, `VENV_DIR`, `TPU_INFERENCE_DIR`).

Observed startup time on the target pod varies: weight load is ~10–18 min (dominated by slowest host; we've seen individual hosts range from 4 min to 25 min at ~6–15 s/shard depending on gcsfuse cache warmth) and compile is ~5–8 min, for a total of ~15–25 min to `Application startup complete`.

## Smoke (5 prompts)

`scripts/multihost/tpu-v4/smoke_test_glm51.sh` fires 5 prompts (`CLEAN`, `BROKEN`, `CAPITAL`, `MATH`, `CODE`) against the running server. The script requires `MODEL` (the id that `vllm serve` registered — by default this is the checkpoint path) and defaults to `PORT=8000`.

Each response is checked byte-for-byte against a golden captured from a separate PP=8 TP=4 run on the same checkpoint. On the target config we observe 4/5 strict byte-match: `BROKEN`, `CAPITAL`, `MATH`, `CODE` match the PP=8 golden exactly. `CLEAN` (`"2+3="`) returns a semantically-correct continuation (e.g. `'5\n3+5=8\n5+'`) that differs in tokenization from the PP=8 golden (`'5\n5+4=9\n9+'`), so it registers as strict-match FAIL while being well-formed output. Pre-§1.8 the `BROKEN` prompt returned `''` and the `TPU_MIN_TOKEN_BUCKET<512` case returned `!!!!`; both are resolved.

## 3D `w2_scale` regression

A checkpoint with 3D `w2_scale` previously raised `ValueError: in_specs too long` at dispatch. After this PR the same launch succeeds without any extra flags.

---

# Known limitations

- FP8 KV cache on EP needs `--kv-cache-dtype auto` for coherent output on some prompts. The `jnp.isinf` guards in `kernels/mla/v2/kernel.py` cover the all-masked padding case but not this content-triggered NaN. The `launch_glm51_tp4_ep8.sh` wrapper sets `KV_CACHE_DTYPE=auto` by default so this is handled for the target config.
- `--load-format tpu_streaming_loader` + `--kv-cache-dtype fp8` raises `NotImplementedError`. Use `--kv-cache-dtype auto` when the streaming loader is enabled.
- Direct FP8 path only validated on GLM-5.1-FP8; a checkpoint with a different block-scale layout may need §3.2's dispatcher extended.
- 2D block-scale `gmm_v2` kernel only exercised on v4; see §3.1 / §3.2 for the non-v4 testing caveat.
- Smoke test `CLEAN` prompt produces a semantically-correct continuation that differs in tokenization from the PP=8 golden, so it registers as byte-match FAIL (expected; see §Tests / Smoke).
- Token throughput is below the hardware ceiling; performance work will land separately.

---

# Checklist


yiqiliu2:
@copilot review

yiqiliu2:
@copilot review

yiqiliu2:
Now code should be working and doc should be comprehensive.

yuanfz98:
@bvrockwell Hello this PR is waiting for code review for 3 weeks. Could you check it ?

kyuyeunk:
@yuanfz98 i don't think this pr is in a state that can be reviewed?

yuanfz98:
@yiqiliu2 thanks for your pr and can you fix CI so that they can start code review ?

yiqiliu2:
Sorry for not fully completing the CI flow!

Cleaned up just now by rebasing onto current `upstream/main` (was 28+ commits behind), and DCO green now.

The only remaining check failure is `buildkite/tpu-inference-ci/pr`, which exits early because the PR is missing the `ready` label. Could a maintainer please add it so CI can actually run? Happy to address review feedback after that.

@yuanfz98 @kyuyeunk @bvrockwell — thanks for the patience and help!

kyuyeunk:
- 33 files changed
- added 4,400 lines
- deleted 334 lines

i'm going to have to pass on reviewing this pr.

yiqiliu2:
> * 33 files changed
> * added 4,400 lines
> * deleted 334 lines
>
> i'm going to have to pass on reviewing this pr.

Thanks a lot for taking a look @kyuyeunk — very fair concern on the size. Here's the +4,400 / -334 broken down so the real review surface is clear.

**Table 1 — by line type (code split by review weight):**

| | Lines | % | |
|---|---|---|---|
| Comments + docstrings | 1,198 | 27% | skip |
| Blank lines | 401 | 9% | skip |
| Code — low-risk: 3 new test files (258) + 4 new `launch_*` / `smoke_*` / `rsync_workers` scripts (353) + `requirements.txt` (1) | 612 | 14% | coverage + multi-host run/smoke tooling, not engine code |
| Code — additive new modules: `tp_selective_loader.py` (~712) + `vllm_model_loader.py` (~303) | 1,015 | 23% | reads top-to-bottom; doesn't touch existing code paths; implements §2.1 / §2.7 / §2.9 (+ the loader half of §2.3) |
| Code — changes to existing source files (23 files) | 1,174 | 27% | **the part to scrutinize → Table 2** |
| **Total** | **4,400** | | |

**Table 2 — what the 1,174 lines of changes to existing files are:**

| | ~code lines | |
|---|---|---|
| v4-only kernel tuning (`kernels/mla/v2/kernel.py` `+162/-140` reworking of an existing kernel, `ragged_paged_attention/v3/kernel.py`, `tpu_generation()==4` gates in `gmm_v2` / `quantized_matmul`) | ~190 | §1.1–1.3; inert on v5e/v5p/v6e/v7 |
| env-var whitelist additions, asserts, logging (spread thin across files) | ~90 | mechanical |
| genuinely new cross-cutting logic | ~890 | → Table 3 |

**Table 3 — where the ~890 cross-cutting lines land (every chunk `§`-tagged):**

| file | ~code | § |
|---|---|---|
| `process_weights/moe_weights.py` | ~234 | §3.1 direct FP8 MoE path · §3.3 GMM_TP `w2_weight_scale` global-axis reconstruction |
| `worker/tpu_worker.py` | ~122 | §2.4 PP group metadata · §2.6 KV cache config indexing · §2.8 runtime diagnostics |
| `layers/vllm/quantization/fp8.py` | ~98 | §3.1 — direct FP8 MoE path |
| `layers/vllm/custom_ops/mla_attention.py` | ~86 | §1.5 — MLA weight absorb under `VllmFp8LinearMethod` |
| `utils.py` | ~63 | §2.8 — runtime HBM fallback + diagnostics |
| `runner/tpu_runner.py` | ~56 | §2.3 — `FORCE_LOGICAL_RESHAPE` consistency (runner side) + mesh / EP wiring |
| `layers/common/attention_interface.py` | ~49 | §1.8 cross-shard Q/K all-gather (+ §1.2 v4 block-size selection) |
| `layers/common/utils.py` | ~45 | §2.2 — process-local weight materialization |
| `kernels/megablox/gmm_v2.py` (non-v4 part) | ~37 | §3.2 — native 2D block scale / `w2_scale` shape handling |
| `executors/ray_distributed_executor.py` | ~31 | §2.5 — Ray executor compiled-DAG bypass for PP > 1 |
| `layers/vllm/backends/flash_attn_mla.py` | ~19 | §1.4 — `expert`-axis all-gather before `o_proj` |
| the rest (~9 small files, ≤30 lines each) | ~50 | §1.6 MLA startup check · §1.7 GLM registry · §3.2 `w2_scale` dispatch in `fused_moe_gmm` · `e_score_correction_bias` plumbing · launcher defaults · env whitelist · … |

So the "needs careful engineering review" surface is realistically **~900 lines of new logic across ~11 changed files**, plus a higher-level top-to-bottom pass over the two new loader modules (~1,080 + ~460 lines, self-contained). §1.4 / §1.8 (the MLA correctness fixes) and §2.1 / §2.7 (mesh-aware EP filter + streaming loader) are the load-bearing correctness pieces; §1.1–1.3 and the `tpu_generation()==4` gates in §3.2 are v4-only workarounds; §2.8 is mostly diagnostics.

It's actually hard to split this into smaller PRs — the §1.x MLA fixes, the §2.x multi-host loader, and the §3.x FP8 MoE weight-prep changes are the exact set the GLM-5.1-FP8 end-to-end run exercises together. For example, without the updated weight loader, the v4-64 pod will OOM in host memory. That is why a partial PR will not pass an end-to-end test. But the PR is structured for per-`§` review — each section has a Reason / How it works / Changes block.

bvrockwell:
Thanks for the PR. My first thought is have you tried this same exact implementation on v6e?

yiqiliu2:
> Thanks for the PR. My first thought is have you tried this same exact implementation on v6e?

Thanks for asking @bvrockwell! Unfortunately we **only have access to a v4 pod** via the TPU Research Cloud program, for a research project on LLM behavior. We spent some time to make models we want to test runnable and loads faster. Along the process, I think some of our code changes may potentially make tpu-inference better, so filed this PR😊.

The TRC program **does allow queued spot v6e**, but recently **the allocation queue is usually longer than expected and the preemption is sooner than expected**🥲, which does not allow a stable debug-test cycle on a project with long cold start times.

yuanfz98:
@bvrockwell Well I think this PR is valueable as it supports a new model especially it's one of the most powerful open source models. Although tpu-inference is mainly targeting v6e and v7x it's also good to be compatible with tpuv4. If further someone wants to have it on v6e, it will be another PR.
Hoping to see it merged.

bvrockwell:
> @bvrockwell Well I think this PR is valueable as it supports a new model especially it's one of the most powerful open source models. Although tpu-inference is mainly targeting v6e and v7x it's also good to be compatible with tpuv4. If further someone wants to have it on v6e, it will be another PR. Hoping to see it merged.

I'll see if @vipannalla or @chishuen have interest in testing on newer chips, but as I said we should be aiming to validate across more TPU generations before treating this as complete.

zorrofox:
Tested this PR on **TPU v6e-64 (GKE, 16 hosts × 4 chips, topology 8x8)** with the official `zai-org/GLM-5.1-FP8` checkpoint, TP4/EP16 (closest analogue to your TP4/EP8). Ray multi-host on K8s IndexedJob worked out of the box. Sharing findings on the v6e-specific questions you raised in §3.1/3.2, plus a few bugs hit along the way.

## §3.1/3.2 — `tpu_generation()==4` FP8 gate on v6e

**v6e does NOT reject `bf16 × fp8` in Pallas/Mosaic** (no E2001), so the fallthrough path compiles. However the resulting `gmm_v2` kernel triggers **massive register-allocator spill** (193 MB spill into a 128 MB VMEM budget):

```
RESOURCE_EXHAUSTED: Ran out of memory in memory space vmem. Used 269.00M of 127.94M vmem.
  1. Size: 193.47M  XLA label: register allocator spill slots in HLO :: gmm_v2-...-tk_6144-tn_2048
  2/3. Size: 36.00M  Shape: u8[37748736]  (FP8 RHS triple-buffer)
```

`_gmm_vmem_estimate()` returns ~76 MB for this config (tile_m=64, tile_k=6144, tile_n=2048, fuse_act=silu), well under the `0.9 × 128 MB` default — but the estimate cannot account for the spill the v6e backend generates when emulating FP8 in the MXU loop.

Dropping `vmem_limit_bytes` to `0.25 × capacity` shrinks `tile_n` to 768 but the kernel still needs **102.27 M scoped vmem** (XLA's own E1001 says "It should not be possible to run out of scoped vmem"). The spill does not converge with smaller tiles — it's coming from the inner-loop per-block dequant at `gmm_v2.py:450-465` (`rhs_block.astype(f32) * scale → astype(bf16) → matmul`, repeated 48× per K, ×N/mxu_size).

Tried both branches of `should_dequantize_before_matmul` (gmm_v2.py:192) — same 102.27 M scoped allocation either way, so the spill is structural to the v6e lowering of this kernel with `f8e4m3fn` RHS, not specific to the inner-loop vs one-shot dequant choice.

**Net answer to §3.1/3.2**: Mosaic on v6e accepts `f8e4m3fn` operands (no E2001), so the existing `tpu_generation()==4` gates in `quantized_matmul/util.py` / `mla/v2/kernel.py` *probably* don't need extending. But the `gmm_v2` direct-FP8 MoE path is unusable on v6e as-is — it would need either a v6e-specific tile/kernel rework, or the simpler route of dequantizing FP8→BF16 in `process_weights` (i.e. `VLLM_MOE_SKIP_REQUANTIZATION=0` equivalent that dequants instead of requantizing) before `gmm_v2` sees the weights.

## Other issues found (not v6e-specific)

1. **`moe_weights.py:766` `@partial(jax.jit, ...)` with no `from functools import partial`** — module-level `NameError` at import. Trivial fix.

2. **DSA indexer path is triggered with the official GLM-5.1-FP8 checkpoint.** The official `config.json` includes `index_topk=2048`, so vLLM `deepseek_v2.py` sets `is_v32=True` and:
   - `torch.empty(..., device='tpu')` (line ~1217) fails — torch doesn't recognize `'tpu'` as a device string outside torchax context
   - `mla_attention.py:419` calls `self.indexer(...)` → indexer's `k_norm(...).type_as(x)` hits a torchax `'View' object has no attribute '_elem'`
   - Setting `is_v32=False` instead breaks weight loading (`KeyError: 'model.layers.0.self_attn.indexer.k_norm.weight'` — the checkpoint ships indexer weights)

   Workaround used: keep `is_v32=True` (so indexer params are created/loaded) but gate the forward call in `mla_attention.py` (`if False:`). A cleaner fix would be a `TPU_DISABLE_DSA_INDEXER` env that no-ops the indexer forward.

3. **`TPU_TRUNCATE_LAYERS` doesn't propagate to KV-cache `num_layers`** — with TRUNCATE=4 the model loads 12.19 GiB/chip, but `kv_cache_manager` still sizes for 78 layers, computes 685 056 tokens, then OOMs on `create_kv_caches`. Had to drop `--gpu-memory-utilization` to 0.5 when using TRUNCATE.

4. **GKE deployment**: the bare-VM `launch_ray_nodocker.sh` translated cleanly to a K8s IndexedJob (rank-0 = `ray start --head` + `vllm serve`, ranks 1-15 = `ray start --address=<head>:6379 --block`, headless Service for DNS). Ray auto-detected `ray.io/tpu-worker-id` from GKE labels — no extra plumbing needed. Happy to share the manifest if useful.

## Config

- Image: vllm@`8189a159` (PR fork's `vllm_lkg.version`) + tpu-inference@`b05ba6e4`
- `vllm serve /weights --tensor-parallel-size 4 --enable-expert-parallel --additional-config '{"sharding":{"sharding_strategy":{"expert_parallelism":16}}}' --load-format tpu_streaming_loader --max-model-len 4096 --max-num-seqs 8 --gpu-memory-utilization 0.5 --no-enable-prefix-caching`
- env: `TPU_MULTIHOST_BACKEND=ray NEW_MODEL_DESIGN=1 VLLM_MOE_SKIP_REQUANTIZATION=1 TPU_TP_SELECTIVE_LOAD=1`

yiqiliu2:
@zorrofox Thank you so much for joining the effort! Added you to the fork just in case if you would like to commit and ship this PR together.

I am currently on travel so it may take me some time to look into it in detail (also not currently having v6e right now). But from the first glance, I think we should force all generations without **Native** FP8 support to go through the BF16 path (which is original FP8+scale in HBM -> dequant to BF16 in VMEM), so we can skip the emulation, maintain good load speed, and maintain the most precision info.

Also, thank you @yuanfz98 for supporting this matter!

yiqiliu2:
Though later we may need to take care of the dequant performance overhead.

Edit: the TPU Builder program has granted me some credit, I guess in this case I can also try v6 sometime.

zorrofox:
@yiqiliu2 Pushed two commits to the branch:
- `7bb7765` — the missing `functools.partial` import
- `6ec804f` — `TPU_DISABLE_DSA_INDEXER` env gate (creates/loads indexer params but skips the forward call; the launch script sets it by default)

Re: forcing non-native-FP8 generations through the BF16-in-VMEM path — I want to flag that on v6e the `gmm_v2.py:417` one-shot dequant branch (`should_dequantize_before_matmul=True`) *appears* to already be the active path (`quant_block_size=128 < mxu_column_size`), and that's the config that produced the 102 M scoped allocation. I tried forcing it via `<=` and got byte-identical 102.27 M, but I didn't independently confirm the patch propagated to the Ray actor process, so treat that with some uncertainty. Will re-verify with an explicit `mxu_column_size` / branch-taken print once I have v6e cycles again (reservation expires 05-24, currently shared).

If the in-VMEM dequant is indeed already what's running, the next thing I'd try is dequant **outside** the `pallas_call` (per-forward, in HBM, before `gmm_v2` sees the RHS) so the kernel body never touches `f8e4m3fn` — trades HBM bandwidth for avoiding the spill entirely.

The `torch.empty(device='tpu')` in vLLM's `deepseek_v2.py` is upstream of this repo so I left it out; happy to add a torchax-side workaround if you'd rather not wait on a vLLM fix.

zorrofox:
Follow-up with v6e cycles available:

**Confirmed** the in-VMEM dequant branch is already active on v6e — added a diagnostic print in `tile_info()`:
```
[gmm_v2-diag] mxu_col=256 qbs=128 dequant_before=True tm=64 tk=6144 tn=768 est=29M limit=32M
```
So `should_dequantize_before_matmul = (128 < 256) = True` from the start; the 102.27 M scoped allocation is what `gmm_v2.py:417`'s one-shot `tiled_rhs.astype(f32).reshape * scale` lowers to on v6e (vs the 29 M estimate). The "force BF16-in-VMEM" path is the one that's failing.

**Tried dequant outside the `pallas_call`** by intercepting in `fused_moe_gmm.py:gmm_wrapper`:
```python
def _dequant_fp8_rhs(lhs, rhs, rhs_scale):
    e, k, n = rhs.shape
    rhs = rhs.astype(jnp.float32).reshape(e, rhs_scale.shape[1], -1, n)
    return (rhs * rhs_scale).reshape(e, k, n).astype(lhs.dtype)

def gmm_wrapper(lhs, rhs, rhs_scale, ...):
    if envs.MOE_DEQUANT_FP8_BEFORE_GMM and rhs_scale is not None and rhs.dtype == jnp.float8_e4m3fn:
        rhs = _dequant_fp8_rhs(lhs, rhs, rhs_scale)
        rhs_scale = None
    gmm_res = gmm_v2(...)
```
Trace succeeds — `gmm_v2` now sees `rhs_dtype=bfloat16 has_scale=False tm=128 tk=6144 tn=1024` and the VMEM error is gone. But the subsequent `Precompile backbone num_tokens=512` warmup crashes the Ray actors with `SYSTEM_ERROR ... connection error code 2. End of file` (no Python traceback, process killed) ~2 min into compile/warmup. Haven't isolated whether that's introduced by the HBM-side dequant or a pre-existing issue downstream of where the earlier runs were dying. Left the dequant diff uncommitted for now since the warmup crash isn't ruled out; will push once it's clean.

zorrofox:
Got it working e2e on v6e-64 — pushed `1a473e5` with the HBM-side dequant (`MOE_DEQUANT_FP8_BEFORE_GMM`). Smoke test 4/5 strict-match, same as the v4 baseline.

Two more issues found and worked around (not yet fixed in code):

**1. `TPU_TRUNCATE_LAYERS` causes per-host SPMD divergence.** `moe.py:113` does `t2j(layer.e_score_correction_bias)` which inlines the tensor as an HLO constant. With TRUNCATE, layers ≥N have that Parameter unloaded (`torch.empty`-initialized random memory), so each host inlines a different `f32[256] constant({...})` → 9 distinct `jit_step_fun` fingerprints across 16 hosts → `Core halted: unexpected peer with different launch id` at first collective. Workaround: don't use TRUNCATE on multi-host. Proper fix would be to pass `e_score_correction_bias` as a jit arg instead of `t2j`-inlining it.

**2. KV-cache budget over-counts ~4× at TP4/EP16.** `tpu_worker.py:487` reports `total_hbm_avail` per-host (4 chips), but `create_kv_caches` allocates the resulting token count per-chip — so 78 layers × per-layer-size lands at ~33 G/chip when avail says 8 G/chip is free. No `gpu-memory-utilization` value sits in the working window (0.43 trips the avail≤0 check, 0.7 OOMs at layer ~44). Workaround: `--num-gpu-blocks-override 256`.

Working config on v6e-64 (TP4/EP16):
```
env: NEW_MODEL_DESIGN=1 VLLM_MOE_SKIP_REQUANTIZATION=1 MOE_DEQUANT_FP8_BEFORE_GMM=1 \
     TPU_DISABLE_DSA_INDEXER=1 TPU_TP_SELECTIVE_LOAD=1 TPU_MULTIHOST_BACKEND=ray
vllm serve <model> --tensor-parallel-size 4 --enable-expert-parallel \
  --additional-config '{"sharding":{"sharding_strategy":{"expert_parallelism":16}}}' \
  --load-format tpu_streaming_loader --max-model-len 2048 --max-num-seqs 8 \
  --gpu-memory-utilization 0.9 --num-gpu-blocks-override 256 \
  --no-enable-prefix-caching --trust-remote-code
# + sed deepseek_v2.py: device=self.device → "cpu" (vLLM-side, not in this repo)
```

Perf is ~195-200ms TPOT @ cc1-8 / 1K-in / 128-out — the per-forward HBM dequant of 75 MoE layers is the obvious tax. Correctness-only for now, as the PR description says.

yuanfz98:
> Thanks for the PR. My first thought is have you tried this same exact implementation on v6e?

@bvrockwell Since that this pr can pass the smoke test in v6e I think it's time to move this PR forward. Thanks.

bvrockwell:
> > Thanks for the PR. My first thought is have you tried this same exact implementation on v6e?
> 
> @bvrockwell Since that this pr can pass the smoke test in v6e I think it's time to move this PR forward. Thanks.

tagging @vipannalla

zorrofox:
Housekeeping push:

- **DCO**: retroactively signed off my three earlier commits (`7bb7765`→`8811fc93`, `6ec804f`→`29852c60`, `1a473e5`→`c48f32d9`) — content unchanged, only the `Signed-off-by:` trailer added. This should clear the DCO check. Apologies for missing `-s` the first time around.
- **`c2822bd`** — minimal fix for the `TPU_TRUNCATE_LAYERS` multi-host SPMD divergence I reported earlier: exempt `e_score_correction_bias` (~1KB/layer) from the PP/truncate skip filter so the value `t2j` inlines as an HLO constant is identical across hosts. The fuller fix (thread it through as a jit input alongside the MoE weights) is left as a TODO for when there's hardware to validate.

Not addressed here:
- **KV-cache 4× over-count at TP4/EP16** — root cause spans the vLLM-side consumer of `determine_available_memory()`; needs a live pod to verify, keeping the `--num-gpu-blocks-override` workaround for now.
- **Rebase onto main** — I probed it locally: 689 commits behind, first PR commit alone conflicts in 8 files / 24 hunks (13 in `mla/v2/kernel.py` — upstream has substantially reworked that kernel). Doable but it's a multi-hour merge that then needs a full v4+v6e re-validation. @yiqiliu2 happy to pair on this when you're back and have v4 cycles; I can cover the v6e leg.

PR #2327, opened 2026-04-20
[Disagg/qwen3.5] disagg support for qwen3.5 (1/n HostKVPoolHMA)

wyzhang:
Add a class `HostKVPoolHMA` to support d2h host kv pool for HMA. This class will be used in a follow up PR by the TPU conenctor implementation supporting HMA.

Why:
Qwen3.5 is a hybrid full attention + Mamba model. Each Mamba layer's kv-cache is a tuple (ssm_state, conv_state) with different shape and dtype than attention. The current HostKVPool assumes a single uniform jax array shape per layer, which doesn't fit for HMA.

The `HostKVPoolHMA` allocates a flat list of jax arrays per buffer, where Mamba tuples are expanded into individual entries, e.g.
ssm_state,   (mamba layer 0)
conv_state,  (mamba layer 0)
ssm_state,   (mamba layer 1)
conv_state,  (mamba layer 1)
ssm_state,   (mamba layer 2)
conv_state,  (mamba layer 2)
attn_state,  (attn  layer 3)
...  (repeat the above)

### DIFF
--- a/tpu_inference/distributed/host_kv_pool_hma.py
+++ b/tpu_inference/distributed/host_kv_pool_hma.py
@@ -1,0 +1,85 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import queue
+import time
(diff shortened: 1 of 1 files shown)


mrjunwan-lang (approved):

tpu_inference/distributed/host_kv_pool_hma.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        per_array_shapes: List[Tuple[int, ...]],
+        per_array_dtype: List[jnp.dtype],
+        per_array_host_sharding: List[jax.sharding.NamedSharding],
+    ) -> List[jax.Array]:
+        host_zeros = [
+            np.zeros(shape, dtype=dtype)
+            for shape, dtype in zip(per_array_shapes, per_array_dtype)
+        ]
+        return jax.device_put(host_zeros, per_array_host_sharding)
```
mrjunwan-lang, line 85: @wyzhang , the device_put may be slow here. have you tried how long it will take for 64 buffers?

PR #2329, opened 2026-04-20
[Gemma 4] Layer specific head dimension logic for quantization

JiriesKaileh:
Added layer specific head dimension logic to account for global_head_dim=512 (full_attention) and head_dim=256 (sliding_attention). Not assigning specific head dimensions to corresponding layers result in a dimension mismatch errors.


tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin, line 1: Thanks for this work Jiries! Can we please update / create new unit tests for this file?

JiriesKaileh: Will do!

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -139,16 +139,27 @@ def qwix_quantize_nnx_model(model: nnx.Module, qwix_config: List[dict],
...
+    head_sizes = kv_cache_head_size
+    if isinstance(head_sizes, int):
+        head_sizes = (head_sizes, ) * num_hidden_layers
+
+    num_kv_heads_tuple = kv_cache_num_kv_heads
+    if isinstance(num_kv_heads_tuple, int):
+        num_kv_heads_tuple = (num_kv_heads_tuple, ) * num_hidden_layers
+
+    kv_caches = []
+    for i in range(num_hidden_layers):
```
jrplatin, line 152: Are we sure that that calling `create_kv_caches` in the `for i in range(num_hidden_layers)` is equivalent to the old call (where we don't have an outter `for-loop`?)

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -291,51 +302,93 @@ def apply_qwix_quantization(
...
+            "Could not find 'num_hidden_layers' in hf_config, hf_config.text_config, or hf_config.vision_config."
         )
 
+    # Determine per-layer kv_heads and head_size if applicable
+    num_kv_heads_list = []
+    head_size_list = []
+
+    text_config = getattr(hf_config, "text_config", hf_config)
+    layer_types = getattr(text_config, "layer_types", None)
+    if layer_types:
```
jrplatin, line 351: What is layer types?

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -291,51 +302,93 @@ def apply_qwix_quantization(
...
+        for i in range(num_hidden_layers):
+            layer_type = layer_types[i] if i < len(
```
jrplatin: What is the length of layer_types with regards to num_hidden_layers?

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -291,51 +302,93 @@ def apply_qwix_quantization(
...
+            if not is_sliding:
+                head_dim_original = getattr(
+                    text_config, "global_head_dim",
+                    getattr(text_config, "head_dim", 0))
+            else:
+                head_dim_original = getattr(text_config, "head_dim", 0)
+            head_size_list.append(utils.get_padded_head_dim(head_dim_original))
+
+            # Determine num_kv_heads
+            use_k_eq_v = ((not is_sliding)
```
jrplatin: This variable name is kind of confusing, maybe clarify

tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -291,51 +302,93 @@ def apply_qwix_quantization(
...
+            if use_k_eq_v:
+                kv_heads = getattr(
+                    text_config, "num_global_key_value_heads",
+                    getattr(text_config, "num_key_value_heads", 0))
+                if not kv_heads:
+                    kv_heads = getattr(text_config, "num_key_value_heads", 0)
+            else:
+                kv_heads = getattr(text_config, "num_key_value_heads", 0)
+            num_kv_heads_list.append(
+                utils.get_padded_num_heads(kv_heads, mesh.shape["model"]))
```
jrplatin: Why `mesh.shape["model"]?` Seems hardcoded?

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: `create_kv_caches()` uses the `layer_names` list to determine the number of arrays to allocate with `sharded_allocate()`. Calling `create_kv_caches()`  once with a list of length N is the same as calling it N times with a list of length 1 and aggregating the results, as long as the other parameters (num_kv_heads, head_size, etc.) are the same.

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: As seen in the [config](https://huggingface.co/google/gemma-4-31B-it/blob/main/config.json) of Gemma 4, `layer_types` is a list that indicates the type of attention used as each hidden layer. The list looks like this: 
[
"sliding_attention",
"sliding_attention",
"sliding_attention",
"sliding_attention",
"sliding_attention",
"full_attention",
"sliding_attention",
"sliding_attention",
...
]

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: Typically, len(`layer_types`) == len(`num_hidden_layers`). I added that line to make the code robust against errors, but perhaps this is going overboard and might lead to silent errors. I think I'll remove it and just throw an error instead if there is a length mismatch.

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: Renaming to `should_use_global_kv_heads`

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: You're right, I replaced `mesh.shape["model"]` with `sharding_size` where `sharding_size = utils.get_mesh_shape_product(mesh, ShardingAxisName.ATTN_HEAD)` to maintain generality.

scripts/vllm/benchmarking/benchmark_dataset.py
```diff
@@ -284,8 +284,10 @@ def sample(
 
             if self.use_chat_template:
                 messages = [{
-                    "role": "system",
-                    "content": "Reasoning effort: high"
+                    "role":
+                    "system",
+                    "content":
+                    "Reasoning effort: high, skip thinking process and must only give short answer in something like A, B, C, D."
```
jrplatin: I wouldn't change this explicitly -- I think a better solution is to keep the `content` as it but make this configurable from the command line, e.g. -- chat-template-system-prompt= "Reasoning..."`

tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin: Ok, might be good to add this test in case it doesn't exist already

tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin: I see, good to add an example explaining this because it's not obvious

tpu_inference/models/jax/utils/qwix/qwix_utils.py
jrplatin: Where did `utils.get_mesh_shape_product(mesh, ShardingAxisName.ATTN_HEAD)` come from?

scripts/vllm/benchmarking/benchmark_dataset.py
JiriesKaileh: Done

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: From `get_kv_cache_shape_with_mesh` in `kv_cache.py`, particularly these lines: 

```
axis_name = ShardingAxisName.ATTN_HEAD
model_cnt = utils.get_mesh_shape_product(mesh, axis_name)
```

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: Done

tpu_inference/models/jax/utils/qwix/qwix_utils.py
JiriesKaileh: Done

jrplatin (approved):

PR #2336, opened 2026-04-20
[Disagg/qwen3.5] disagg support for qwen3.5 (3/n tpu connector for HMA)

wyzhang:
Add `TPUConnectorHMA`, an HMA-aware KV connector for hybrid attention + Mamba model like qwen 3.5.

Why
The current TPUConnector assumes a single kv cache group with uniform layer shape, dtype, and sharding. Qwen3.5 is a hybrid attention + Mamba model whose kv cache manager partitions layers into multiple groups (3 Mamba kv cache group3 + 1 full attention kv cache group). The kv cache jax array shapes are different for Mamba vs Full attention.

Key changes to support HMA
* Block ids become list of list. One per kv cache group rather than a single list
* kv cache is a list containing a mix of jax array tuple for Mamba layer and single jax array for Attention layer. Flatten this by expanding Mamba tuple to individual entires, so to use the same mechanism for pulling from prefill server.
* No-trim, no re-prefill on decode. Prefill's recurrent state summarizes the whole prompt, we cannot trim partial/tail block from prefill.
* kv host pool will use the new `HostKVPoolHMA` to offload send buffers to pinned host DRAM.
* Observability. Log key events during transfer.

### DIFF
--- a/tpu_inference/distributed/tpu_connector_hma.py
+++ b/tpu_inference/distributed/tpu_connector_hma.py
@@ -1,0 +1,596 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import functools
+import time
+from typing import TYPE_CHECKING, Any, Optional
+
+import jax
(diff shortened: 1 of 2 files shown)


tpu_inference/distributed/tpu_connector_hma.py
```diff
@@ -0,0 +1,593 @@
...
+
+                # Build one index array per group. JAX must see the same
+                # device_array call across nodes for collective consistency.
+                # TODO(xiang): pad block_ids to avoid recompilation
+                indices_per_group = [
+                    device_array(self.mesh, np.array(ids))
+                    for ids in req_meta.local_block_ids
+                ]
+                conn = self._maybe_build_kv_connection(req_meta)
+                self.reqs_pulling[req_id] = self.pull_executor.submit(
```
ReneeZhuGG: reqs_pulling was updated in PR#2311

wyzhang: Thanks. Updated the code. Fixed.

mrjunwan-lang (approved):

PR #2366, opened 2026-04-22
[Bug][Qwen3.5] Fix KV cache OOB block IDs in hybrid attn+mamba models

qizzzh:
## Summary
  Fixes a silent KV-cache corruption on TPU for hybrid attention + mamba
  models (e.g. Qwen3.5-A3B / A17B) that manifests as gibberish output when
  2+ requests run concurrently at modest `gpu_memory_utilization`.

  ## Root cause

  vLLM's hybrid allocator sizes one `KVCacheTensor` per `shared_by` group
  (Qwen3.5: 1 full-attn + 3 linear-attn = 4 layers sharing a tensor),
  relying on **GPU-style byte-level reinterpretation** — the four layers
  point at disjoint slots within the same physical tensor, and CUDA
  kernels reinterpret the bytes as attention KV or mamba state.

  TPU can't do that: `jax.Array`s are strongly typed, so tpu-inference
  **duplicates** each of the 4 layers into its own physical tensor. That
  divides the `shared_by` budget four ways, leaving each per-layer cache
  with roughly `1 / (num_attn_per_shared + num_mamba_per_shared)` the
  block count that vLLM's scheduler thinks exists (~3.5× smaller for
  Qwen3.5).

  When the scheduler hands out a block ID past the per-layer slot count,
  JAX's `dynamic_update_slice_in_dim` silently clips it to the last valid
  slot. Multiple requests' mamba recurrent states collapse onto the same
  slot → corrupted state → gibberish. Attention layers have the same
  mismatch but usually only manifest when IDs climb high (concurrency
  high, memory low).

  ## Fix

  Two coordinated changes in `tpu_inference/runner/kv_cache_manager.py`:

  1. **Pad every layer spec's `page_size_padded`** to the full per-
     `shared_by` footprint (`num_attn_groups × attn_page + num_mamba_groups
     × mamba_unpadded`). vLLM's uniform-page-size check passes, and its
     `num_blocks` calculation now reflects the duplicated-tensor reality.

  2. **Set `num_gpu_blocks_override`** to exactly match the block count
     that main's buggy-but-memory-accurate formula would land on. This
     eliminates a 1-block flooring-boundary drift between vLLM's naive
     padded calc and main's effective TPU allocation, so peak HBM usage is
     identical to main bit-for-bit.

  `attn_page` consistently uses `get_attention_page_size_bytes` (which
  accounts for dtype-specific packing like fp8) on both sides of the
  `tensor.size / total_group_page` computation, so fp8 KV cache models
  (397B-A17B) work too.

Eval results
```
|mmlu_pro           |      2|custom-extract|      |exact_match|↑  |0.8179|±  |0.0230|
| - biology         |      3|custom-extract|     5|exact_match|↑  |0.8500|±  |0.0819|
| - business        |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - chemistry       |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|
| - computer_science|      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|
| - economics       |      3|custom-extract|     5|exact_match|↑  |0.9500|±  |0.0500|
| - engineering     |      3|custom-extract|     5|exact_match|↑  |0.7000|±  |0.1051|
| - health          |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - history         |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|
| - law             |      3|custom-extract|     5|exact_match|↑  |0.7000|±  |0.1051|
| - math            |      3|custom-extract|     5|exact_match|↑  |0.9500|±  |0.0500|
| - other           |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - philosophy      |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - physics         |      3|custom-extract|     5|exact_match|↑  |1.0000|±  |0.0000|
| - psychology      |      3|custom-extract|     5|exact_match|↑  |0.9000|±  |0.0688|

|mmlu_pro|      2|custom-extract|      |exact_match|↑  |0.8179|±  | 0.023|

### DIFF
--- a/tpu_inference/runner/kv_cache_manager.py
+++ b/tpu_inference/runner/kv_cache_manager.py
@@ -56,6 +56,13 @@ class KVCacheManager:
         # from the KV cache of `shared_kv_cache_layers[layer_name]`.
         self.shared_kv_cache_layers: dict[str, str] = {}
         self.use_mla = self.runner.model_config.use_mla
+        # Set by `update_mamba_page_size_padded` for hybrid attention+mamba
+        # models. When set, every attention layer spec reports this as its
+        # `page_size_padded` so vLLM sees a uniform page size across groups
+        # and computes `num_blocks` that matches what each layer actually gets
+        # on the TPU side (where we duplicate the shared tensor per layer
+        # because mamba and attention caches have different shapes).
+        self._hybrid_uniform_page_size_bytes: int | None = None
 
     def _create_attention_spec(
             self,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tpu_inference/runner/kv_cache_manager.py [hunk 1]
```diff
@@ -67,38 +74,72 @@ def _create_attention_spec(
             page_size_bytes = get_attention_page_size_bytes(
                 self.runner.mesh, block_size, num_kv_heads, head_size,
                 self.runner.kv_cache_dtype, True)
+            padded = (self._hybrid_uniform_page_size_bytes
```
jrplatin: nit: `padded` -> `page_size_padded`?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -67,38 +74,72 @@ def _create_attention_spec(
...
                                     dtype=self.runner.kv_cache_dtype,
                                     cache_dtype_str=self.runner.vllm_config.
                                     cache_config.cache_dtype,
-                                    page_size_padded=int(page_size_bytes))
+                                    page_size_padded=padded)
         else:
             page_size_bytes = get_attention_page_size_bytes(
                 self.runner.mesh, block_size, num_kv_heads, head_size,
                 self.runner.kv_cache_dtype, False)
+            padded = (self._hybrid_uniform_page_size_bytes
```
jrplatin: Same here

tpu_inference/runner/kv_cache_manager.py
qizzzh: Done

tpu_inference/runner/kv_cache_manager.py
qizzzh: Done

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -122,12 +163,156 @@ def update_mamba_page_size_padded(
...
+        # for e.g. 12 sw + 13 full). For every `shared_by` tensor,
+        # exactly one layer per group ends up inside, so
+        # `num_shared = num_attn_groups + num_mamba_groups`.
+        num_attn = len(attn_modules)
+        num_mamba = len(mamba_modules)
+        min_count = min(num_attn, num_mamba)
+        max_count = max(num_attn, num_mamba)
+        # Match vLLM exactly: float comparison, no int() truncation (matters
+        # at e.g. min=3, max=4, where 4 < 4.5 but 4 < int(4.5)==4 differs).
+        if max_count < min_count * 1.5:
```
jrplatin, line 225: Can you explain this logic a bit? Kind of hard to follow to me

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -122,12 +163,156 @@ def update_mamba_page_size_padded(
...
+        size equal the per-layer tensor's leading dimension exactly.
+
+        No internal safety margin is applied — `gpu_memory_utilization` is
+        the knob users already have for reserving headroom. Adding a
+        silent reduction here would conflict with their explicit budget.
+
+        Skipped if the user has explicitly set `num_gpu_blocks_override` or
+        if HBM usage isn't readable (e.g. in tests without real devices).
+        Spec padding alone still fixes the OOB bug in that case; only the
+        ~1-block-per-tensor flooring-boundary precision is lost.
```
jrplatin, line 286: nit: can you please add args / return values here? Thanks!

jrplatin (approved):

tpu_inference/runner/kv_cache_manager.py
qizzzh: Done

tpu_inference/runner/kv_cache_manager.py
qizzzh: I've rewritten the surrounding comment block to be more concrete. TL;DR: group_size is the number of layers vLLM packs into each equal-sized group, and there's a tradeoff — pick max_count and you get fewer groups (often 1 per type) but the smaller side wastes space padding up to max; pick min_count and there's no padding but the larger side splits into ceil(max/min) groups. The 1.5× rule is vLLM's heuristic: pick max only when padding is minor (max < 1.5 × min), else min. Two concrete cases: 12 sw + 13 full → max (1 group each); 10 attn + 30 mamba → min (1 attn + 3 mamba groups).

PR #2371, opened 2026-04-22
[CI] Fix Correctness tests for Multimodal Inputs

kwang3939:
# Description

Fix the error `pass_config.fuse_minimax_qk_norm pydantic ValidationError (None not valid bool)` occurred in v6 and v7 Multimodal Inputs tests.

# Tests

Tested locally on v7 with `python3 -m pytest -s -v tests/e2e/test_multi_modal_inference.py`.

# Checklist


lk-chen (approved):
LGTM. Would you share which upstream change breaks it?

kwang3939:
> LGTM. Would you share which upstream change breaks it?

@lk-chen , I didn't check which exact commit broke it. But I think it started to be broken from Apr 18, so for vllm it's likely between Apr 7 to Apr 13.

PR #2390, opened 2026-04-24
[TPU KV Offloading] [Feat] KV cache offloading to host memory

juncgu-google:
# Description

TL;DR, add the feature of offloading KV cache to host cpu buffer (similar to the [native CPU offloading in vLLM](https://github.com/vllm-project/vllm/issues/19854)).

This PR allows offloading computed KV cache (at the granularity of block / page) (of prompt tokens or even including the generated tokens) to the host CPU buffer and bringing them back to TPU HBM when there are cache hits, to avoid re-compute. 

## implementation
Following the [general kv connector interfaces](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_transfer/kv_connector/v1/base.py) and the logic of [native CPU offloading in vLLM](https://github.com/vllm-project/vllm/issues/19854), it introduces a  [TPUOffloadConnector](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/tpu_offload_connector.py#L388), which is the central component of managing and executing the offloading logic. Within the `TPUOffloadConnector`, there are:
1. [TPUOffloadConnectorScheduler](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/tpu_offload_connector.py#L519): colocates with the vLLM scheduler, makes KV cache swap decisions, and keeps track of running requests and their in-flight swap operations.   
   - [OffloadManager](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/offload_manager.py#L129): manages the logical kv cache space on host CPU buffer, and applys LRU eviction policy.
   - [StagingBufferManager](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/offload_manager.py#L269): manages a logical staging buffer in the TPU's HBM, which is used as a temporary buffer (in HBM) for data being transferred between the TPU and the CPU. Without the control of using temporary buffer in HBM can easily lead to TPU OOM.  
 2. [TPUOffloadConnectorWorker](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/tpu_offload_connector.py#L1210): executes the swap decisions (embedded in the SchedulerOutput) from `TPUOffloadConnectorScheduler`.     
    - [LocalCPUBackend](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/offload/cpu_backend.py#L26): a simple key-value map in host CPU memory for hosting the offloaded KV cache blocks / pages.

## usage
This feature can be used by setting the `kv_tranfer_config` in vLLM engine:
`--kv-transfer-config '{"kv_connector":"TPUOffloadConnector","kv_connector_module_path":"tpu_inference.distributed.offload.tpu_offload_connector","kv_role":"kv_both"}'`
example: [examples/offload/gke/pod_tpu_commons_cpu_offload.yaml](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/examples/offload/gke/pod_tpu_commons_cpu_offload.yaml#L35)

And, it can be configured through the following [environment variables](https://github.com/vllm-project/tpu-inference/blob/ff512cc815476b47fcb878d1dbd4f0a00f18bbac/tpu_inference/envs.py#L266-L283) (we will move them into the `KVTransferConfig.kv_connector_extra_config`):
1. `TPU_OFFLOAD_SKIP_JAX_PRECOMPILE`: skipping pre-compiling the swap functions, default=0. We would suggest to turn on pre-compile. All swap operations are applied at block-granularity; when swap pre-compile is turned on, we will break a request of swap into multiple swap-operations following the predefined bucket size list (1 block, 2 blocks, 4 blocks, 8 blocks, 16 blocks) to avoid re-compile (thanks to @saikat-royc).
2. `TPU_OFFLOAD_NUM_CPU_CHUNKS`: host CPU buffer capacity in terms of number of chunks (equivalent to kv cache blocks / pages), default=1024.
3. `TPU_OFFLOAD_NUM_STAGING_BLOCKS`: the size of the staging buffer in the TPU HBM, default=128.
4. `TPU_OFFLOAD_DECODE_SAVE`: save the KV cache of generated (decode) tokens, default=False.

# Tests

* All tests: `pytest -sv tests/offload/`
* Accuracy tests: `pytest -sv tests/offload/tpu_offloading_accuracy_test.py`, `pytest -sv tests/offload/tpu_offloading_multi_request_accuracy_test.py`
* Performance test: `pytest -sv tests/offload/tpu_offloading_performance_test.py`

# Checklist


sixiang-google (approved):

cle5eland:
@juncgu-google A couple questions for you! Did you consider how this could be extended to be a tier 1 implementation for the Tiered Offloading Connector in base vLLM? Put another way, IIUC (I'm new to the team) we can't use the Tiered Offloading framework (design [here](https://github.com/vllm-project/vllm/issues/38260)) in the tpu-inference library because the tier 1 GPU -> CPU offloading implementation in vllm isn't compatible with TPUs. My goal is to make it so that any arbitrary tier 2 offloading manager (ssd, object store, etc) that's compatible with native vllm would work on tpus also. What I've gathered is that that having a tier 1 offload implementation that would enable TPU -> CPU offload (which is what you've implemented) but is ALSO compatible with the Tier 1 interface for the tiered offloading manager would give this for free.

Let me know your thoughts!

juncgu-google:
Hi @cle5eland,

1. no, we can not directly use the tiered offloading connector in vllm. Similar to this implementation, we have to implement a new tiered offloading connector in tpu-inference.
2. we are moving to kv offloading using [tpu-sync](https://github.com/google/tpu-sync) (https://github.com/vllm-project/tpu-inference/pull/3029). Inside tpu-sync, we will have tiered storage being supported.

cle5eland:
@juncgu-google right the first tier would need to be reconstructed (possibly as a wrapper around the work you've already done), but we could make it so that all secondary tiers are compatible with TPUs.

PR #2392, opened 2026-04-24
Extend attn_dp_expert to emulate attn_dp.

NicoGrande:
# Description

This PR modified `attn_dp_expert` to emulate the behavior of `attn_dp`. Concretely, this PR allows the "expert" mesh physical axis to extend beyond the number of KV heads in a model by assigning the quotient to "attn_dp_expert". 

This is helpful for models that shard the expert dimension of MoE layers over the "expert" physical axis.  

# Tests

Added a new unit test in `tests/layers/common/test_sharding.py` to validate this behavior. Additionally, this has been tested as part of the e2e RL efforts on Qwen3-235B-A22B with the plugin MaxText model (which shards experts using the "expert" axis).

# Checklist

### DIFF
--- a/tpu_inference/layers/common/sharding.py
+++ b/tpu_inference/layers/common/sharding.py
@@ -194,8 +194,23 @@ class ShardingConfigManager:
                 int(tensor_parallelism // num_kv_heads_per_device_in_kv_cache),
                 1)
             tensor_parallelism = tensor_parallelism // attn_dp
-            attn_dp_expert = expert_parallelism
-            expert_parallelism = 1
+
+            # If Attention DP is active or TP perfectly saturates the KV heads limit,
+            # prioritize TP for KV heads and shift all expert parallelism to attn_dp_expert.
+            is_kv_fully_sharded_by_tp = (attn_dp > 1) or (
+                tensor_parallelism == num_kv_heads_per_device_in_kv_cache)
+
+            if is_kv_fully_sharded_by_tp:
+                attn_dp_expert = expert_parallelism
+                expert_parallelism = 1
+            else:
+                # Otherwise, shard KV heads over the expert axis.
+                # Use the remaining KV heads per device as the divisor.
+                shard_divisor = num_kv_heads_per_device_in_kv_cache // tensor_parallelism
+                attn_dp_expert = max(1,
+                                     int(expert_parallelism // shard_divisor))
+                expert_parallelism //= attn_dp_expert
+
         else:
             attn_dp = 1
             attn_dp_expert = 1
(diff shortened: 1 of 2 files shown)


tpu_inference/layers/common/sharding.py
```diff
@@ -194,8 +194,13 @@ def from_vllm_config(cls,
                 int(tensor_parallelism // num_kv_heads_per_device_in_kv_cache),
                 1)
             tensor_parallelism = tensor_parallelism // attn_dp
-            attn_dp_expert = expert_parallelism
-            expert_parallelism = 1
+
```
wenxindongwork, line 197: could you add a unit test in `tests/layers/common/test_sharding.py`?

wenxindongwork (approved):
LGTM! also asked George to take a quick look.

tpu_inference/layers/common/sharding.py
NicoGrande: Done! Added one more test after the attn_dp unit test

tpu_inference/layers/common/sharding.py
```diff
@@ -194,8 +194,13 @@ def from_vllm_config(cls,
...
+            # Extend attn_dp_expert to to behave like attn_dp. This is helpful for
+            # MoE model implementations that shard experts using the "expert" axis.
+            attn_dp_expert = max(
+                int(expert_parallelism // num_kv_heads_per_device_in_kv_cache),
```
gpolovets1: Could this be overcounting `num_kv_heads_per_device_in_kv_cache`? E.g. if there is a case where heads is sharded by TP then `num_kv_heads_per_device_in_kv_cache` would already be considered when calculating `attn_dp` so we would be double counting it if we also applied in `attn_dp_expert`.

NicoGrande: I added conditional logic to check for this case: If TP fully shards the number of KV heads, then we keep the old logic. Otherwise, we fall back to sharding KV heads over expert and assigning the remainder to attn_dp_expert. I also handle the case where TP partially shards the number of KV heads.

tpu_inference/layers/common/sharding.py
```diff
@@ -194,8 +194,23 @@ def from_vllm_config(cls,
...
+            is_kv_fully_sharded_by_tp = (attn_dp > 1) or (
+                tensor_parallelism == num_kv_heads_per_device_in_kv_cache)
+
+            if is_kv_fully_sharded_by_tp:
+                attn_dp_expert = expert_parallelism
+                expert_parallelism = 1
+            else:
+                # Otherwise, shard KV heads over the expert axis.
+                # Use TP as the divisor if available; fallback to the KV heads per device.
+                shard_divisor = tensor_parallelism if tensor_parallelism > 1 else num_kv_heads_per_device_in_kv_cache
```
gpolovets1: Isn't shard_divisor in this case = num_kv_heads_per_device_in_kv_cache // tensor_parallelism?

NicoGrande: Thanks @gpolovets1 ! Just updated the code to reflect this

gpolovets1 (approved):

PR #2395, opened 2026-04-24
[DP] Remove CPU overhead by reducing update_from_output payload size 

wenxindongwork:
# Description

When DP is large, CPU processing time starts to exceed TPU processing time. One of the main CPU bottlenecks is the IPC between main process and DP worker processes during the `update_from_output` function. 

This PR reduces the amount of data serialized over IPC pipes between the main process and per-rank scheduler workers for this function via two optimizations. 

1. Each worker now maintains a local `_cached_scheduler_outputs` deque. When `SCHEDULE` is called, the output is cached locally in the worker before being sent back to the parent.
`UPDATE_FROM_OUTPUT` no longer receives a (scheduler_output, model_runner_output) tuple — it receives only the `ModelRunnerOutput` and pops the corresponding `SchedulerOutput` from its local cache.

2. Previously, every rank received a `ModelRunnerOutput` containing the full global `sampled_token_ids`. Workers had to ignore entries not belonging to them. Now, each rank receives only its own data sliced by `_split_model_output_by_rank`. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/core/sched/dp_scheduler.py
+++ b/tpu_inference/core/sched/dp_scheduler.py
@@ -14,6 +14,7 @@
 
 import atexit
 import copy
+import gc
 import multiprocessing
 import multiprocessing.reduction
 import os
@@ -27,6 +28,7 @@ from time import time
 from typing import Any, Dict, List, Optional, Tuple
 
 import cloudpickle
+import numpy as np
 import torch
 from vllm.config import VllmConfig
 from vllm.multimodal import MULTIMODAL_REGISTRY, MultiModalRegistry
@@ -38,7 +40,7 @@ from vllm.v1.core.sched.scheduler import Scheduler
 from vllm.v1.engine import EngineCoreOutputs
 from vllm.v1.kv_cache_interface import KVCacheConfig
 from vllm.v1.metrics.stats import PrefixCacheStats, SchedulerStats
-from vllm.v1.outputs import DraftTokenIds, ModelRunnerOutput
+from vllm.v1.outputs import DraftTokenIds, LogprobsLists, ModelRunnerOutput
 from vllm.v1.request import Request
 from vllm.v1.structured_output import StructuredOutputManager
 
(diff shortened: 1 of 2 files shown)


tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -453,36 +472,56 @@ def _get_result(self,
...
+            if gc_was_enabled:
+                gc.disable()
+
             result = cloudpickle.loads(raw_bytes)
+            deserialize_time = time()
+
+            if gc_was_enabled:
+                gc.enable()
+
+            deserialize_time = time()
```
QiliangCui: deserialize_time is asigned twice. is the first one necessary?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -179,14 +184,20 @@ def _send_result(result):
...
                         scheduler_output, model_runner_output)
                     _send_result(result)
 
                 case SchedulerCommand.GET_GRAMMAR_BITMASK:
-                    scheduler_output = data
-                    result = scheduler.get_grammar_bitmask(scheduler_output)
+                    if _cached_scheduler_outputs:
+                        cached_output = _cached_scheduler_outputs[-1]
+                    else:
+                        cached_output = None
```
QiliangCui: in old logic, can scheduler_output be None? 
If not, should we `assert scheduler_output is not None`?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -179,14 +184,20 @@ def _send_result(result):
...
+                    result = (scheduler.get_grammar_bitmask(cached_output)
+                              if cached_output is not None else None)
                     _send_result(result)
```
QiliangCui, line 198: same here.. from reading the old logic, it seems the result should never be None. 

If so, assert it?

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -1127,4 +1230,4 @@ def update_vllm_config_for_dp_scheduler(vllm_config: Any) -> None:
         else:
             vllm_config.scheduler_config._original_scheduler_cls = Scheduler
 
-        vllm_config.scheduler_config.scheduler_cls = DPScheduler
+        vllm_config.scheduler_config.scheduler_cls = DPScheduler
```
QiliangCui: add the trailing newline.

QiliangCui (approved):

PR #2396, opened 2026-04-24
[Qwen3.5] Enable jittable vision tower for Qwen3.5

lk-chen:
# Description

- Add `GridTHW` data structure, to avoid jit error `This concrete value was not available in Python because it depends on the value of the argument kwargs['image_grid_thw']`
- Add jit and precompile for certain models (e.g. Qwen3.5 moe) (`experimental/vision_tower_jit.py`)
- fix test

It doesn't work with
- Qwen3-VL
- Gemma4, because its [pooler_mask](https://github.com/huggingface/transformers/blob/c472755e79aac54d675845bff5e5c821c21260af/src/transformers/models/gemma4/modeling_gemma4.py#L2015) is dynamic shape

# Tests

server
```bash
HF_HUB_OFFLINE=1 MODEL_IMPL_TYPE=vllm vllm serve \
    --model=Qwen/Qwen3.5-35B-A3B \
    --tensor-parallel-size=8 \
    --limit-mm-per-prompt '{"image": 4, "video": 0}' \
    --block-size=256 \
    --max-model-len=8192 \
    --max-num-seqs=256
```

query
```bash
curl http://localhost:8000/v1/chat/completions     -H "Content-Type: application/json"     -d '{
        "model": "Qwen/Qwen3.5-35B-A3B",
        "messages": [
            {
                "role": "user",
                "content": [
                    {
                        "type": "image_url",
                        "image_url": {"url": "https://news.bbc.co.uk/media/images/38107000/jpg/_38107299_ronaldogoal_ap_300.jpg"}
                    },
                    {
                        "type": "text",
                        "text": "plz describe the image above"
                    }
                ]
            }
        ],
        "max_tokens": 25,
        "temperature": 0.0
    }'
```

Qwen3.5-397B needs to skip precompile to avoid OOM
```bash
SKIP_JAX_PRECOMPILE=1 HF_HUB_OFFLINE=1 MODEL_IMPL_TYPE=vllm vllm serve \
    --model=Qwen/Qwen3.5-397B-A17B \
    --tensor-parallel-size=8 \
    --limit-mm-per-prompt '{"image": 4, "video": 0}' \
    --block-size=256 \
    --max-model-len=4096 \
    --gpu-memory-utilization=0.98 \
    --max-num-seqs=16
```

Also checked 2.5-VL on jax path
```bash
HF_HUB_OFFLINE=1 vllm serve \
      --model=Qwen/Qwen2.5-VL-7B-Instruct \
      --tensor-parallel-size=4 \
      --limit-mm-per-prompt '{"image": 4, "video": 0}' \
      --block-size=256 \
      --max-model-len=8192 \
      --max-num-seqs=256
```

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/experimental/vision_tower_jit.py
+++ b/tpu_inference/models/vllm/experimental/vision_tower_jit.py
@@ -32,7 +32,7 @@ JITTABLE_ARCHS = {
 
 
 def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
-                                    vllm_config: VllmConfig):
+                                    vllm_config: VllmConfig) -> Callable:
     """Conditionally wrap `embed_multimodal_func_jax` with jax.jit based on the VllmConfig.
 
     Args:
@@ -85,7 +85,7 @@ def maybe_precompile_vision_encoder_fn(
 
     def precompile_fn(run_compilation_fn: Callable) -> None:
         # Deferred import avoids circular dependencies at module load time.
-        from tpu_inference.runner.multimodal_manager import GridTHW
+        from tpu_inference.models.vllm.vllm_model_wrapper import GridTHW
 
         for num_patches in num_patches_paddings:
             # Split num_patches into (h, w) by distributing bits evenly.
--- a/tpu_inference/models/vllm/experimental/vision_tower_jit.py
+++ b/tpu_inference/models/vllm/experimental/vision_tower_jit.py
@@ -45,7 +45,8 @@ def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
         logger.info_once(
             f"JIT-compiling embed_multimodal_func_jax for architectures: {archs & JITTABLE_ARCHS}"
         )
-        return jax.jit(embed_multimodal_func_jax)
+        return jax.jit(static_argnames=("image_grid_thw", "video_grid_thw",
+                                        "grid_thw"))(embed_multimodal_func_jax)
     else:
         return embed_multimodal_func_jax
 
--- a/tpu_inference/models/vllm/experimental/vision_tower_jit.py
+++ b/tpu_inference/models/vllm/experimental/vision_tower_jit.py
@@ -1,0 +1,47 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Utilities to support JIT compilation of VisionTower.
+
+from typing import Callable
+
+import jax
+from vllm.config import VllmConfig
+
+from tpu_inference.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
+                                    vllm_config: VllmConfig):
+    """Conditionally wrap `embed_multimodal_func_jax` with jax.jit based on the VllmConfig.
+
+    Args:
+        embed_multimodal_func_jax: The JAX function to be potentially JIT-compiled.
+        vllm_config: The VllmConfig instance containing the configuration.
+    """
+    archs = set(vllm_config.model_config.hf_config.architectures)
+
+    JITTABLE_ARCHS = {
+        "Qwen3_5MoeForConditionalGeneration",
+    }
+
+    if archs & JITTABLE_ARCHS:
+        logger.info_once(
+            f"JIT-compiling embed_multimodal_func_jax for architectures: {archs & JITTABLE_ARCHS}"
+        )
+        return jax.jit(embed_multimodal_func_jax)
+    else:
+        return embed_multimodal_func_jax
--- a/tpu_inference/models/vllm/experimental/vision_tower_jit.py
+++ b/tpu_inference/models/vllm/experimental/vision_tower_jit.py
@@ -14,14 +14,21 @@
 
 # Utilities to support JIT compilation of VisionTower.
 
-from typing import Callable
+import math
+from typing import Any, Callable, Optional
 
 import jax
+import jax.numpy as jnp
 from vllm.config import VllmConfig
 
 from tpu_inference.logger import init_logger
 
 logger = init_logger(__name__)
+
+# Architectures whose embed_multimodal function is safe to wrap with jax.jit.
+JITTABLE_ARCHS = {
+    "Qwen3_5MoeForConditionalGeneration",
+}
 
 
 def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
@@ -34,10 +41,6 @@ def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
     """
     archs = set(vllm_config.model_config.hf_config.architectures)
 
-    JITTABLE_ARCHS = {
-        "Qwen3_5MoeForConditionalGeneration",
-    }
-
     if archs & JITTABLE_ARCHS:
         logger.info_once(
             f"JIT-compiling embed_multimodal_func_jax for architectures: {archs & JITTABLE_ARCHS}"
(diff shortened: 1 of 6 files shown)


tpu_inference/runner/multimodal_manager.py
```diff
@@ -111,6 +168,20 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
...
                 mm_kwargs):
+            for k, v in mm_kwargs_group.items():
+                if k in ("image_grid_thw", "video_grid_thw", "grid_thw"):
+                    mm_kwargs_group[k] = GridTHW(v.tolist())
+                else:
+
+                    def move_to_jax(x):
+                        if isinstance(x, torch.Tensor):
+                            with torchax.default_env():
+                                return torchax.interop.jax_view(
```
kyuyeunk: is this different than `t2j`?

tpu_inference/runner/multimodal_manager.py
```diff
@@ -111,6 +168,20 @@ def execute_mm_encoder(self, scheduler_output: "VllmSchedulerOutput"):
         encoder_outputs = []
         for _, num_items, mm_kwargs_group in group_and_batch_mm_kwargs(
                 mm_kwargs):
+            for k, v in mm_kwargs_group.items():
+                if k in ("image_grid_thw", "video_grid_thw", "grid_thw"):
+                    mm_kwargs_group[k] = GridTHW(v.tolist())
+                else:
+
+                    def move_to_jax(x):
```
kyuyeunk: i've seen similar function used in different file as well: https://github.com/vllm-project/tpu-inference/blob/ecc3b78f5b7a00f257b689db02e9d0c1364d342c/tpu_inference/models/vllm/vllm_model_wrapper.py#L413

do we need to move this into commonutlity file instead? https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/utils.py

also, is there no native operations like this in torchax?

tpu_inference/runner/multimodal_manager.py
```diff
@@ -29,6 +31,61 @@
...
+    def prod(self, dim=-1):
+        if dim in (-1, 1):
+            return np.array([row[0] * row[1] * row[2] for row in self])
+        raise NotImplementedError(f"GridTHW.prod({dim}) not supported")
+
+    def __repr__(self):
+        return f"GridTHW({tuple(self)})"
+
+
+# Register as a JAX pytree leaf (no children) so jax.tree_util.tree_map
```
kyuyeunk: why is this needed? what happens if we just treat this class as a static argument? doesn't that also avoid jax traversing its elements?

tpu_inference/models/vllm/experimental/vision_tower_jit.py
kyuyeunk, line 1: can you give a high level summary on what this file does?

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -0,0 +1,119 @@
...
+    if not (archs & JITTABLE_ARCHS):
+        return None
+
+    vc = vllm_config.model_config.hf_config.vision_config
+    patch_input_dim = (vc.in_channels * vc.temporal_patch_size *
+                       vc.patch_size * vc.patch_size)
+    spatial_merge_unit = vc.spatial_merge_size**2
+    max_patches = (vllm_config.scheduler_config.max_num_batched_tokens *
+                   spatial_merge_unit)
+    min_shift = 4  # 1 << 4 = 16 patches minimum
```
jrplatin, line 135: This seems hard-coded? Can you explain some of this logic more?

lk-chen: They are copied from [jax-path](https://github.com/vllm-project/tpu-inference/blob/ecc3b78f5b7a00f257b689db02e9d0c1364d342c/tpu_inference/models/jax/qwen2_5_vl.py#L485)

Yes the current version may only work for Qwen3.5, but vision encoder of each model is just different from each other, and I guess the best solution might just to have per-modal implementation of pre-compile. Before we have conclusion, I'll just put it in experimental folder.

tpu_inference/models/vllm/experimental/vision_tower_jit.py
lk-chen: This file provides utilities to jit and precompile [embed_multimodal](https://github.com/vllm-project/vllm/blob/a474da28131f61684849b31e29af0eebaaedc383/vllm/model_executor/models/interfaces.py#L153) from a vllm model.

It's hard to guarantee jit can work for all models (see main description) depends on the implementation details. So we check architecture and only jit when the architecture is whitelisted

tpu_inference/runner/multimodal_manager.py
lk-chen: The quoted line is moved from original file to here in this PR.

tpu_inference/runner/multimodal_manager.py
lk-chen: Tested and seems to work with `jit(static_argnames=...`

tpu_inference/runner/multimodal_manager.py
lk-chen: Tried
```
with jax.set_mesh(jax.sharding.get_mesh()):
        return t2j(x)
```
and got error
```
Got argument kwargs['pixel_values'] of VllmModelWrapper.wrap_embed_multimodal_func.<locals>.embed_multimodal_func_jax with shape bfloat16[308,1536] and device ids [0] on platform CPU and jit's context mesh with device ids [0, 1, 4, 5, 6, 7, 2, 3] on platform TPU
```

kyuyeunk: i think you need to use `t2j(x, use_dlpack=False)`. otherwise, you'll get torch tensor with a jax wrapper and not jax tensor with a torch wrapper.

lk-chen: Thanks, done using t2j

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -107,6 +110,49 @@ def compute_logits(self, hidden_state: torch.Tensor) -> torch.Tensor:
         return self.vllm_model.compute_logits(hidden_state)
 
 
+class GridTHW(tuple):
```
kyuyeunk: not sure if `vllm_model_wrapper.py` is a good place to put this logic.

in general, i see that `vllm_model_wrapper.py` is being really crowded with so many logic. can you think of a different place to put this logic? maybe in the newly created `tpu_inference/models/vllm/experimental/vision_tower_jit.py`?

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -0,0 +1,120 @@
...
+    max_shift = max(min_shift, (max(max_patches, 1) - 1).bit_length())
+    num_patches_paddings = [1 << i for i in range(min_shift, max_shift + 1)]
+
+    from torchax.ops.mappings import TORCH_DTYPE_TO_JAX
```
kyuyeunk: use `to_jax_dtype` instead: https://github.com/vllm-project/tpu-inference/blob/31e3d71f88077f715a6d3cf6e3598b189cc56de8/tpu_inference/utils.py#L63

also, why is this import in the middle of the file?

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -0,0 +1,120 @@
...
+import jax.numpy as jnp
+from vllm.config import VllmConfig
+
+from tpu_inference.logger import init_logger
+
+logger = init_logger(__name__)
+
+# Architectures whose embed_multimodal function is safe to wrap with jax.jit.
+JITTABLE_ARCHS = {
+    "Qwen3_5MoeForConditionalGeneration",
```
kyuyeunk: is it possible to use type matching instead of string matching? https://github.com/huggingface/transformers/blob/3146914673a4ebff2f5e06ace68fbddacdebe90b/src/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py#L2010

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -0,0 +1,120 @@
...
+            # For any power-of-2 num_patches = 2^k: h=2^(k//2), w=2^(k-k//2).
+            k = int(round(math.log2(num_patches)))
+            h = 1 << (k // 2)
+            w = 1 << (k - k // 2)
+
+            dummy_pixel_values = jnp.ones((num_patches, patch_input_dim),
+                                          dtype=jax_dtype)
+            dummy_image_grid_thw = GridTHW([(1, h, w)])
+
+            try:
```
kyuyeunk: feels hacky? there should be a more fundamental fix to handle this OOM?

tpu_inference/models/vllm/vllm_model_wrapper.py
lk-chen: done, moved to vision_tower_jit

tpu_inference/models/vllm/experimental/vision_tower_jit.py
lk-chen: done

tpu_inference/models/vllm/experimental/vision_tower_jit.py
lk-chen: Done

tpu_inference/models/vllm/experimental/vision_tower_jit.py
lk-chen: removed try-catch, there's a bug where I used * instead of /

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -397,44 +402,70 @@ def draft_step_fun(
...
+                                                  self.vllm_config)
+
+    @functools.cache
     def wrap_embed_multimodal_func(self):
         if not self.vllm_config.model_config.is_multimodal_model:
             return None
 
-        # The function cannot be JITted directly due to its dynamic implementation
-        def embed_multimodal_func(
+        def embed_multimodal_func_jax(
```
kyuyeunk, line 420: what's the difference between `embed_multimodal_func_jax` vs. `embed_multimodal_func_torch`?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -28,7 +29,7 @@
 from flax.typing import PRNGKey
 from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torchax.interop import jax_view, torch_view
-from torchax.ops.mappings import TORCH_DTYPE_TO_JAX
+from torchax.ops.mappings import TORCH_DTYPE_TO_JAX, t2j
```
kyuyeunk, line 31: nit: not related to this pr, but do you know if this `TORCH_DTYPE_TO_JAX` can be replaced with `to_jax_dtype`?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -397,44 +402,70 @@ def draft_step_fun(
...
+    ) -> Optional[Any]:
+        """Return a precompile function for the vision encoder, or None."""
+        if not self.vllm_config.model_config.is_multimodal_model:
+            return None
+        embed_multimodal_fn = self.wrap_embed_multimodal_func()
+        return maybe_precompile_vision_encoder_fn(params, embed_multimodal_fn,
+                                                  self.model.vllm_model,
+                                                  self.vllm_config)
+
+    @functools.cache
```
kyuyeunk: what is the purpose of `functools.cache`?

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -0,0 +1,178 @@
...
+                num_patches=num_patches,
+            )
+
+    return precompile_fn
+
+
+def maybe_prepare_for_jit(kwargs: dict, vllm_model) -> dict:
+    """Convert certain kwargs to JIT-friendly formats, if needed.
+    
+    Specifically, convert "image_grid_thw", "video_grid_thw", and "grid_thw" to GridTHW instances, which are tuple subclasses that can be hashed in jax.jit.
```
kyuyeunk: please keep 80 lines limit. and make similar fixes if you see any others like this.

tpu_inference/models/vllm/vllm_model_wrapper.py
lk-chen: `_torch` takes torch.Tensor input, `_jax` takes jax.Array (and GridTHW) input which is jittable

tpu_inference/models/vllm/vllm_model_wrapper.py
lk-chen: yes, `to_jax_dtype` uses `t2j_dtype` from torchax, which is just looking up this map

tpu_inference/models/vllm/vllm_model_wrapper.py
lk-chen: I think it was somehow auto completed, removed

tpu_inference/models/vllm/experimental/vision_tower_jit.py
lk-chen: Done!

kyuyeunk (approved):

PR #2083, opened 2026-03-30
Pipeline psum and sc kernel

clee1994:
# Description

Start with a short description of what the PR does and how this is a change from
the past.

The rest of the description includes relevant details and context, examples:
- why is this change being made,
- the problem being solved and any relevant context,
- why this is a good solution,
- some information about the specific implementation,
- shortcomings of the solution and possible future improvements.

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: #123456

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2401, opened 2026-04-25
Enable Gather All Reduce Pipelining.

kyuyeunk:
# Description

Add below 3 XLA flags to pipeline gather / all-reduce.
1. Prevent XLA from undoing all-reduce tiling.
2. Enforce all-reduce to be done in tensor core.
3. Increase memory limit to allow extra memory usage from pipeline.

Additionally, reverting many of changes from previous PR as it has been the source of performance degradataion: https://github.com/vllm-project/tpu-inference/pull/2083 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -286,7 +286,17 @@ class VllmModelWrapper:
                 "xla_tpu_all_gather_collective_matmul_mode":
                 "post_spmd_conservative",
                 "xla_tpu_reduce_scatter_collective_matmul_mode":
-                "post_spmd_conservative"
+                "post_spmd_conservative",
+                # Add below 3 XLA flags to pipeline gather / all-reduce.
+                # 1: Prevent XLA from undoing all-reduce tiling.
+                "xla_jf_crs_combiner_threshold_in_bytes": 0,
+                # 2: Enforce all-reduce to be done in tensor core.
+                "xla_tpu_enable_sparse_core_collective_offload_all_reduce":
+                False,
+                # 3. Increase memory limit to allow extra memory usage from
+                # pipeline.
+                # TODO(kyuyeunk): Investigate XLA scheduler memory limit.
+                "xla_tpu_scheduler_percent_shared_memory_limit": 1000,
             },
             static_argnames=(
                 "layer_name_to_kvcache_index",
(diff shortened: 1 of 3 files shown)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -286,7 +286,11 @@ def jit_step_func(self):
                 "xla_tpu_all_gather_collective_matmul_mode":
                 "post_spmd_conservative",
                 "xla_tpu_reduce_scatter_collective_matmul_mode":
-                "post_spmd_conservative"
+                "post_spmd_conservative",
+                "xla_jf_crs_combiner_threshold_in_bytes": 0,
```
helloworld1, line 292: Crs combiner is usually good for performance to combine all reduces etc. Wonder if we can we only set it on the gmm part so it won't apply to other parts?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -286,7 +286,11 @@ def jit_step_func(self):
...
+                "xla_tpu_enable_sparse_core_collective_offload_all_reduce":
```
helloworld1, line 294: Same here. Sparse core AR can usually overlap better, but on gmm we want to control manually. This may impact other parts.

tpu_inference/models/vllm/vllm_model_wrapper.py
kyuyeunk: if we are building completely generic framework, i would agree with you, but we basically know where every all-reduce would be & how it will be used, etc. and my understanding is that there isn't a use case in our code base that we want to combine all-reduce or want to use sparse core ar.

but still, if there is a way to set xla flag for only specific part i think that is preferred way, but i don't think there is. so alternative is to have some kind of branching logic where xla flag is applied for this case while don't applied for other case, etc, which would be a bit annoying to maintain.

vanbasten23:
Could you update the PR description on what these 2 flags try to achieve?

kyuyeunk:
> Could you update the PR description on what these 2 flags try to achieve?

done. added in the comment as well.

tpu_inference/models/vllm/vllm_model_wrapper.py
helloworld1: Does it work by using something like `jax.jit(compiler_options={"xla_tpu_enable_sparse_core_collective_offload_all_reduce": False})` on the moe fuction?

kyuyeunk: last time i tried, compiler options can only be done at the top-most jax jit. which makes sense because if you look at xla compilation process, it's not really taking one function by function, it's applying optimization across entire model.

PR #2405, opened 2026-04-25
[FP8 TorchAX Quantization] Add support for bypassing requantization for FP8 2D pre-quantized checkpoints

jrplatin:
# Description

The main goal of this PR to enable loading 2D-blockwise quantized models, like [Qwen3-Coder](https://huggingface.co/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8), without the need to re-quantize weights.  This may be useful to avoid quality loss.

Specifically, I do the following:

 * Kernel Optimizations:
    * GMM v2: Added a VMEM dequantization path in the GMM kernel for cases where the quantization block size is smaller than the TPU MXU size (< 128). This prevents performance issues with small contracting dimensions while maintaining accuracy.
       * Quantized Matmul: Updated `xla_quantized_matmul `to detect and handle 2D scale tensors by performing dequantization before the matmul operation.
   * Weight Processing and Sharding:
       * Linear Weights: Updated shard_linear_weights to correctly shard 2D scale tensors `[out_blocks, in_blocks]` following the weight sharding pattern.
       * MoE Weights: Enhanced MoE processing (`process_w13_for_gmm` and `process_fp8_moe_weights`) to handle scale_ratio adjustments for block-quantized scales, ensuring proper padding
         and reordering when re-quantization is disabled.
   * Infrastructure:
       * Added the `DISABLE_WEIGHT_REQUANTIZATION` toggle in `tpu_inference/envs.py`.

# Tests

* Kernel Verification:
    * `test_gmm_weight_quantized_small_block`: Validates accuracy when dequantizing small blocks (e.g., 64) inside the GMM kernel.
       * `test_xla_quantized_matmul_2d_scale`: Confirms correct handling of 2D scales in the XLA matmul utility.
   * Weight Processing:
       * Added `tests/layers/common/test_process_weights.py` to verify sharding logic for both 2D and 3D scale tensors.
       * Tested `process_w13_for_gmm with scale_ratio` > 1 to ensure scales are correctly repeated and aligned with padded weights.
   * Unit Tests:
       * Added `test_2d_block_quantization` to verify the core `quantization/dequantization` logic for 2D blocks.
       * Verified the new environment variable behavior in `test_envs.py`.

Also ran perf comparing requant w/ no requant and seeing about a ~20% throughput drop, which is expected and should be reasonable IMO so I think I am fine to share the feature branch with Meta:

| Workload (320 requests, 64 conc) | Metric | Requant (default, baseline) | No Requant (this PR) | Requant Advantage |
| :--- | :--- | :--- | :--- | :--- |
|  **1k / 1k** | **Duration & Throughput** | | | |
| *(294k in / 295k out)*| Benchmark duration (s) | **148.52** | 183.72 | 19.2% faster |
| | Request throughput (req/s) | **2.15** | 1.74 | **+23.6%** |
| | Output token throughput (tok/s) | **1992.25** | 1610.51 | **+23.7%** |
| | Total Token throughput (tok/s) | **3974.09** | 3212.60 | **+23.7%** |
| | **Latency (ms)** | | | |
| | Mean TTFT (Prefill) | **529.44** | 626.80 | -15.5% (lower is better) |
| | Mean TPOT (Decode) | **30.35** | 37.59 | -19.3% |
| | Mean E2E Latency | **28583.93** | 35371.64 | -19.2% |
| | | | | |
|  **8k / 1k** | **Duration & Throughput** | | | |
| *(2.35M in / 295k out)*| Benchmark duration (s) | **275.93** | 332.27 | 17.0% faster |
| | Request throughput (req/s) | **1.16** | 0.96 | **+20.8%** |
| | Output token throughput (tok/s) | **1072.30** | 890.49 | **+20.4%** |
| | Total Token throughput (tok/s) | **9597.34** | 7970.12 | **+20.4%** |
| | **Latency (ms)** | | | |
| | Mean TTFT (Prefill) | **3348.84** | 3883.60 | -13.8% |
| | Mean TPOT (Decode) | **54.44** | 65.64 | -17.1% |
| | Mean E2E Latency | **53788.98** | 64701.77 | -16.9% |
| | | | | |
|  **1k / 8k** | **Duration & Throughput** | | | |
| *(294k in / 2.37M out)*| Benchmark duration (s) | **1139.55** | 1415.93 | 19.5% faster |
| | Request throughput (req/s) | **0.28** | 0.23 | **+21.7%** |
| | Output token throughput (tok/s) | **2083.21** | 1676.58 | **+24.3%** |
| | Total Token throughput (tok/s) | **2341.50** | 1884.45 | **+24.3%** |
| | **Latency (ms)** | | | |
| | Mean TTFT (Prefill) | **493.68** | 594.11 | -16.9% |
| | Mean TPOT (Decode) | **29.30** | 36.41 | -19.5% |
| | Mean E2E Latency | **217903.95** | 270679.64 | -19.5% |

CMDs:

```
# NOTE: add DISABLE_WEIGHT_REQUANTIZATION=1  to skip requant
USE_BATCHED_RPA_KERNEL=1    MODEL_IMPL_TYPE=vllm nohup vllm serve Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8 --max-model-len=9216 --max-num-batched-tokens=8192 --max-num-seqs=512 --kv-cache-dtype=fp8 --no-enable-prefix-caching --gpu-memory-utilization=0.8 --tensor-parallel-size=8 --async-scheduling --block-size=256 --enable-expert-parallel &

# InferenceMax client:
https://paste.googleplex.com/4913959054475264
```

# Checklist

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -183,10 +183,16 @@ class InputConfigs:
     def should_dequantize_before_matmul(self) -> bool:
         """Whether to dequantize before matmul.
 
-        Dequantization is required when the quant block size is smaller than the
-        MXU column size. This allows for the dequantized values to be computed
-        within VMEM, which may need to higher performance for small block sizes,
-        e.g. 64 or 32.
+        Dequantization is preferred when the quant block size is smaller than the
+        MXU column size. In the standard quantized matmul flow, the contracting
+        dimension is tied to the quantization block size. If the block size is
+        small (e.g., 16 or 32 for MX formats), it severely underutilizes the MXU
+        (which expects 128 or 256), causing poor performance.
+
+        By dequantizing the weights inside VMEM first, we can reshape them and
+        perform a regular bf16 matmul that fully utilizes the MXU capacity. This
+        retains high performance while allowing us to keep the original, non-requantized
+        weights in HBM.
         """
         if not self.has_scale or self.quant_block_size is None:
             return False
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -368,12 +368,12 @@ def inner_kernel(
             mxu_size = pltpu.get_tpu_info().mxu_column_size
             rhs_qbs = cfgs.rhs_cfgs.quant_block_size
 
-            # Dequantize inside VMEM to avoid small contracting dimension in jnp.matmul.
+            # This should only be taken in the case where we don't requantize
+            # the scales and thus we need to dequantize inside VMEM to avoid small contracting
+            # dimmensions
             if cfgs.rhs_cfgs.has_scale and rhs_qbs < mxu_size:
                 tiled_rhs_scale = tiled_rhs_ref.get_scale().astype(
                     acc_ref.dtype)
-                # tiled_rhs: [tile_k, tile_n]
-                # tiled_rhs_scale: [num_blocks, 1, tile_n]
                 num_blocks = cfgs.num_quant_blocks_per_tile_k
                 tiled_rhs_dequant = tiled_rhs.astype(acc_ref.dtype).reshape(
                     num_blocks, rhs_qbs, rhs_tile_n)
@@ -435,7 +435,9 @@ def inner_kernel(
             # MXU ops for the next [tile_m, mxu_size].
             mxu_size = pltpu.get_tpu_info().mxu_column_size
 
-            # Dequantize inside VMEM to avoid small contracting dimension in jnp.matmul.
+            # Same as the unquantized matmul path -- this is where we don't requantize
+            # the scales and thus we need to dequantize inside VMEM to avoid small contracting
+            # dimmensions
             rhs_qbs = cfgs.rhs_cfgs.quant_block_size
             if cfgs.rhs_cfgs.has_scale and rhs_qbs < mxu_size:
                 tiled_rhs_scale = tiled_rhs_ref.get_scale().astype(
@@ -484,7 +486,7 @@ def inner_kernel(
 
                     block_acc = jnp.matmul(
                         block_lhs_q,
-                        block_rhs.astype(preferred_element_type),
+                        block_rhs,
                         preferred_element_type=preferred_element_type,
                     ).astype(acc_ref.dtype)
 
@@ -981,7 +983,7 @@ def validate_inputs(
     if rhs_scale is not None:
         num_quant_blocks = rhs_scale.shape[1]
         assert rhs_scale.shape == (size_group, num_quant_blocks, 1, size_n)
-        assert size_k % num_quant_blocks == 0, f"{size_k=}, {num_quant_blocks=}"
+        assert size_k % num_quant_blocks == 0, f"{size_k=} but received {num_quant_blocks=}"
 
     assert group_offset.shape == (1, )
 
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -367,30 +367,54 @@ def inner_kernel(
             # Unquantized matmul path.
             mxu_size = pltpu.get_tpu_info().mxu_column_size
             rhs_qbs = cfgs.rhs_cfgs.quant_block_size
-            for start_n in range(0, rhs_tile_n, mxu_size):
-                end_n = min(rhs_tile_n, start_n + mxu_size)
-                col_size = end_n - start_n
 
-                acc_n = jnp.zeros((cfgs.tiles.tile_m, col_size),
-                                  dtype=acc_ref.dtype)
-                for b_id in range(cfgs.num_quant_blocks_per_tile_k):
-                    k_start = b_id * rhs_qbs
-                    k_end = k_start + rhs_qbs
+            # Dequantize inside VMEM to avoid small contracting dimension in jnp.matmul.
+            if cfgs.rhs_cfgs.has_scale and rhs_qbs < mxu_size:
+                tiled_rhs_scale = tiled_rhs_ref.get_scale().astype(
+                    acc_ref.dtype)
+                # tiled_rhs: [tile_k, tile_n]
+                # tiled_rhs_scale: [num_blocks, 1, tile_n]
+                num_blocks = cfgs.num_quant_blocks_per_tile_k
+                tiled_rhs_dequant = tiled_rhs.astype(acc_ref.dtype).reshape(
+                    num_blocks, rhs_qbs, rhs_tile_n)
+                tiled_rhs_dequant = tiled_rhs_dequant * tiled_rhs_scale
+                tiled_rhs = tiled_rhs_dequant.reshape(cfgs.tiles.tile_k,
+                                                      rhs_tile_n)
 
-                    block_acc = jnp.matmul(
-                        tiled_lhs[:, k_start:k_end],
-                        tiled_rhs[k_start:k_end, start_n:end_n],
+                for start_n in range(0, rhs_tile_n, mxu_size):
+                    end_n = min(rhs_tile_n, start_n + mxu_size)
+                    col_size = end_n - start_n
+                    acc_n = jnp.matmul(
+                        tiled_lhs,
+                        tiled_rhs[:, start_n:end_n],
                         preferred_element_type=jnp.float32,
                     ).astype(acc_ref.dtype)
+                    acc_list.append(acc_n)
+            else:
+                for start_n in range(0, rhs_tile_n, mxu_size):
+                    end_n = min(rhs_tile_n, start_n + mxu_size)
+                    col_size = end_n - start_n
 
-                    if cfgs.rhs_cfgs.has_scale:
-                        tiled_rhs_scale = tiled_rhs_ref.get_scale()
-                        block_acc *= tiled_rhs_scale[b_id, :,
-                                                     start_n:end_n].astype(
-                                                         acc_ref.dtype)
+                    acc_n = jnp.zeros((cfgs.tiles.tile_m, col_size),
+                                      dtype=acc_ref.dtype)
+                    for b_id in range(cfgs.num_quant_blocks_per_tile_k):
+                        k_start = b_id * rhs_qbs
+                        k_end = k_start + rhs_qbs
 
-                    acc_n += block_acc
-                acc_list.append(acc_n)
+                        block_acc = jnp.matmul(
+                            tiled_lhs[:, k_start:k_end],
+                            tiled_rhs[k_start:k_end, start_n:end_n],
+                            preferred_element_type=jnp.float32,
+                        ).astype(acc_ref.dtype)
+
+                        if cfgs.rhs_cfgs.has_scale:
+                            tiled_rhs_scale = tiled_rhs_ref.get_scale()
+                            block_acc *= tiled_rhs_scale[b_id, :,
+                                                         start_n:end_n].astype(
+                                                             acc_ref.dtype)
+
+                        acc_n += block_acc
+                    acc_list.append(acc_n)
         else:
             # Quantized matmul path.
             lhs_q_dtype = cfgs.lhs_cfgs.quant_dtype
@@ -410,6 +434,19 @@ def inner_kernel(
             # inner loop which can be used to pipeline subsequent VPU or VST ops with
             # MXU ops for the next [tile_m, mxu_size].
             mxu_size = pltpu.get_tpu_info().mxu_column_size
+
+            # Dequantize inside VMEM to avoid small contracting dimension in jnp.matmul.
+            rhs_qbs = cfgs.rhs_cfgs.quant_block_size
+            if cfgs.rhs_cfgs.has_scale and rhs_qbs < mxu_size:
+                tiled_rhs_scale = tiled_rhs_ref.get_scale().astype(
+                    acc_ref.dtype)
+                num_blocks = cfgs.num_quant_blocks_per_tile_k
+                tiled_rhs_dequant = tiled_rhs.astype(acc_ref.dtype).reshape(
+                    num_blocks, rhs_qbs, rhs_tile_n)
+                tiled_rhs_dequant = tiled_rhs_dequant * tiled_rhs_scale
+                tiled_rhs = tiled_rhs_dequant.reshape(cfgs.tiles.tile_k,
+                                                      rhs_tile_n)
+
             for start_n in range(0, rhs_tile_n, mxu_size):
                 end_n = min(rhs_tile_n, start_n + mxu_size)
                 col_size = end_n - start_n
@@ -447,14 +484,14 @@ def inner_kernel(
 
                     block_acc = jnp.matmul(
                         block_lhs_q,
-                        block_rhs,
+                        block_rhs.astype(preferred_element_type),
                         preferred_element_type=preferred_element_type,
                     ).astype(acc_ref.dtype)
 
                     block_acc *= block_scale.astype(acc_ref.dtype)
 
-                    # Apply rhs subchannel scale per quant block.
-                    if cfgs.rhs_cfgs.has_scale:
+                    # Apply rhs subchannel scale per quant block if it was not dequantized earlier.
+                    if cfgs.rhs_cfgs.has_scale and not (rhs_qbs < mxu_size):
                         b_id = start_k // cfgs.rhs_cfgs.quant_block_size
                         rhs_scale_slice = tiled_rhs_ref.get_scale()
                         block_acc *= rhs_scale_slice[b_id, :,
@@ -944,7 +981,7 @@ def validate_inputs(
     if rhs_scale is not None:
         num_quant_blocks = rhs_scale.shape[1]
         assert rhs_scale.shape == (size_group, num_quant_blocks, 1, size_n)
-        assert size_k % num_quant_blocks == 0
+        assert size_k % num_quant_blocks == 0, f"{size_k=}, {num_quant_blocks=}"
 
     assert group_offset.shape == (1, )
 
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -383,7 +383,6 @@ def inner_kernel(
 
                 for start_n in range(0, rhs_tile_n, mxu_size):
                     end_n = min(rhs_tile_n, start_n + mxu_size)
-                    col_size = end_n - start_n
                     acc_n = jnp.matmul(
                         tiled_lhs,
                         tiled_rhs[:, start_n:end_n],
@@ -983,7 +982,8 @@ def validate_inputs(
     if rhs_scale is not None:
         num_quant_blocks = rhs_scale.shape[1]
         assert rhs_scale.shape == (size_group, num_quant_blocks, 1, size_n)
-        assert size_k % num_quant_blocks == 0, f"{size_k=} but received {num_quant_blocks=}"
+        assert (size_k % num_quant_blocks == 0
+                ), f"{size_k=} should be divisible by {num_quant_blocks=}"
 
     assert group_offset.shape == (1, )
 
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -411,7 +411,8 @@ def inner_kernel(
                         preferred_element_type=jnp.float32,
                     ).astype(acc_ref.dtype)
 
-                    if not cfgs.rhs_cfgs.should_dequantize_before_matmul:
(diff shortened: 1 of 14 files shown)


vanbasten23:
Just to clarify, in the perf metrics in the PR description, the "No Requant" column is what this PR implements, right?

jrplatin:
> Just to clarify, in the perf metrics in the PR description, the "No Requant" column is what this PR implements, right?

Correct!

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -196,12 +233,18 @@ def _pad_tensor(t):
     # 3. Concatenate and Reorder for avoiding TP sharding comms
     w13_concat = jnp.concatenate([padded_w1, padded_w3], axis=concat_dim)
     if padded_output_sizes is not None:
-        return reorder_concatenated_tensor_for_sharding(
+        padded_output_sizes_adj = [
```
vanbasten23: What's "adj"?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -163,28 +163,65 @@ def process_w13_for_gmm(tensor,
...
+            tensors. Finally, it upsamples (repeats) the resulting tensor by this ratio
+            so the final scales align 1-to-1 with the dimensions of the full weight tensor.
+
+    Returns:
+        The processed JAX array, appropriately padded and dimensionally aligned
+        for the target MoE hardware backend.
+    """
+
+    if not envs.DISABLE_WEIGHT_REQUANTIZATION:
+        assert scale_ratio == 1, "If not requantizing, scale_ratio should be 1!"
```
vanbasten23: nit: ""If not requantizing..." should be "If requantizing..."

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -163,28 +163,65 @@ def process_w13_for_gmm(tensor,
...
+        tensor: The input JAX array. Can be the actual weight tensor or its
+            corresponding block-quantized scale tensor.
+        concat_dim: The axis dimension along which W1 and W3 are concatenated.
+        config: A `W13PaddingConfig` object containing the unscaled sizes and padding
+            amounts calculated based on the full *weight* tensor dimensions.
+        padded_output_sizes: Optional list of sizes for the padded W1 and W3 blocks.
+            If provided, triggers a reordering of the concatenated tensor for optimal
+            TP sharding.
+        name: String identifier used for logging tensor shapes.
+        scale_ratio: The dimensional reduction factor of the `tensor` compared to the
```
vanbasten23: Could you add a comment on who set this value and how it is set? IIUC, scale_ratio is the quant block size?

vanbasten23:
Could you explain to me why this PR may be useful to avoid quality loss?

tpu_inference/kernels/quantized_matmul/util.py
```diff
@@ -85,6 +85,22 @@ def xla_quantized_matmul(
     Returns:
         Output of the quantized matmul.
     """
+    skip_scale = False
+    if w_scale is not None and w_scale.ndim == 2:
+        # If w_scale is 2D, we assume 2d-blockwise quantization and thus we
+        # want to de-quantize first before the matmul.
```
vanbasten23: Could you add a comment on why we should de-quant before the matmul, like the scale is not constant along the contracting axis?

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -179,6 +179,20 @@ def should_bitcast(self) -> bool:
         bits = jax.dtypes.itemsize_bits(self.dtype)
         return bits < 8
 
+    @property
+    def should_dequantize_before_matmul(self) -> bool:
+        """Whether to dequantize before matmul.
+
+        Dequantization is required when the quant block size is smaller than the
```
vanbasten23: similarly, can you add more comment on why dequant is needed before the matmul, unlike the current kernel behavior?

vanbasten23: also please make sure to update the internal kernel file

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -361,15 +409,29 @@ def process_moe_weights(
                                         name="w13_weight")
 
             if w13_weight_scale is not None:
+                scale_ratio = w13_weight.shape[2] // w13_weight_scale.shape[3]
```
vanbasten23: check it's divisable?

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Adjusted, I can write it out!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Yep, will do!

tpu_inference/kernels/megablox/gmm_v2.py
jrplatin: Yep, done already: https://critique.corp.google.com/cl/908950015

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Will do!

jrplatin:
> Could you explain to me why this PR may be useful to avoid quality loss?

Basically, it's because when we requantize, we are fundamentally changing the nature of the checkpoint and Meta observed quality loss when doing so -- my understanding is that there could be rounding / other errors being introdcued by dequantizing and requantizing, so this helps skip that potentially lossy step

tpu_inference/kernels/megablox/gmm_v2.py
jrplatin: Done!

tpu_inference/kernels/quantized_matmul/util.py
jrplatin: Done!

vanbasten23 (approved):

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -160,9 +160,14 @@ def shard_linear_weights(
 
     weight_sharding = NamedSharding(mesh, weight_p_spec)
     bias_sharding = NamedSharding(mesh, bias_p_spec)
-    if isinstance(weights.weight_scale, (jax.Array, Tensor)) and len(
-            weights.weight_scale.shape) == 3:
-        num_blocks = weights.weight_scale.shape[0]
+    # If weight_scale is a list (split weights), we check the first element.
+    sample_scale = weights.weight_scale[0] if isinstance(
+        weights.weight_scale,
+        list) and weights.weight_scale else weights.weight_scale
```
kyuyeunk: nit: what happens if it's a tuple?

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -160,9 +160,14 @@ def shard_linear_weights(
...
+
+    if isinstance(sample_scale,
+                  (jax.Array, Tensor)) and len(sample_scale.shape) == 3:
```
kyuyeunk: isn't it more common to use `isinstance(sample_scale, jax.Array | Tensor)`? using `(jax.Array, Tensor)` is a bit confusing because it looks like it's checking tuple.

also, can't we just do `jax_view(tensor)` at the start of this function and make all the function logic check for `jax.Array`? having a lot of branching logic makes it really prone to errors.

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -172,6 +177,11 @@ def shard_linear_weights(
         out_axis = weight_p_spec[0]
         weight_scale_p_spec = P(in_axis, None, out_axis)
         weight_scale_sharding = NamedSharding(mesh, weight_scale_p_spec)
+    elif isinstance(sample_scale, (jax.Array, Tensor)) and len(
```
kyuyeunk: same comment as above.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -163,28 +163,70 @@ def process_w13_for_gmm(tensor,
...
+            amounts calculated based on the full *weight* tensor dimensions.
+        padded_output_sizes: Optional list of sizes for the padded W1 and W3 blocks.
+            If provided, triggers a reordering of the concatenated tensor for optimal
+            TP sharding.
+        name: String identifier used for logging tensor shapes.
+        scale_ratio: The dimensional reduction factor of the `tensor` compared to the
+            base weight tensor. This value is calculated and provided by the caller
+            (e.g., `process_moe_weights`) by dividing the base weight's dimension size
+            by the scale tensor's corresponding dimension size.
+            - When using the standard requantization flow (`DISABLE_WEIGHT_REQUANTIZATION=False`),
```
kyuyeunk: using 1 to denote no requantization is really prone to error. suggest using `scale_ratio: int | None` instead. also, the name `scale_ratio` sounds a bit confusing. i was a bit confused because i thought it was quantization scale. can't we just use the term `block_size`? also, what happens if a block size is asymmetric 2d? so things like `(128, 64)`?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -163,28 +163,70 @@ def process_w13_for_gmm(tensor,
...
+    # Adjust config values for scale tensors with reduced dimensions
+    intermediate_size = config.intermediate_size // scale_ratio
+    local_intermediate_size = config.local_intermediate_size // scale_ratio
+    pad_amount = config.pad_amount // scale_ratio
+    padded_intermediate_size = config.padded_intermediate_size // scale_ratio
 
     # 1. Split into W1 and W3
-    w1 = tensor[..., :config.intermediate_size]
-    w3 = tensor[..., config.intermediate_size:]
+    w1 = tensor[..., :intermediate_size]
```
kyuyeunk: feels like it's better to just have `config` also take the quantization block size into consideration?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -243,10 +291,12 @@ def process_moe_weights(
         w13_weight = jnp.concat([w1_weight, w3_weight], axis=1)
 
         if w13_weight_scale is not None:
-            w1_weight_scale = w13_weight_scale[:, ::2, :]
-            w3_weight_scale = w13_weight_scale[:, 1::2, :]
-            w13_weight_scale = jnp.concat([w1_weight_scale, w3_weight_scale],
-                                          axis=1)
+            # If scale is block-quantized along the inter dimension, adjust stride
```
kyuyeunk: type: inner?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -243,10 +291,12 @@ def process_moe_weights(
...
+            if w13_weight_scale.shape[1] == w13_weight.shape[1]:
```
kyuyeunk, line 286: bit confused here: why is this logic treated differently?

jrplatin: When w13_interleave is true, the checkpoint has W1 and W3 weights interleaved element-by-element along axis 1. The current code assumes the scale tensor had a 1-to-1 mapping with the weights along that axis and de-interleaved it using a stride of 2 (w13_weight_scale[:, ::2, :]).

IIUC, this is fine for 1D per-channel scales because scale.shape[1] == weight.shape[1].
However, with 2D block-wise scales, shape[1] is the number of blocks (out_blocks), not the number of elements. If we apply [:, ::2, :] to the block dimension, we aren't separating W1 and W3 elements -- we'd just be incorrectly grabbing alternating blocks (e.g. sending Block 0 entirely to W1, Block 1 entirely to W3),  which is wrong.

This check ensures we only de-interleave the scales if they are 1D/per-channel and actually map 1-to-1 with the interleaved weight elements

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Agreed! Will change `scale_ratio` to `block_size` and use `None` for the requant (assume that's what you meant) case

As for asymmetric 2D blocks (like (128, 64)), the logic should work because process_w13_for_gmm only splits and pads along the concat_dim axis, so it only needs the block size for that specific axis.

If we look at where we call this from in process_moe_weights, we calculate the ratio specifically for the dimension we are slicing: `scale_ratio = w13_weight.shape[2] // w13_weight_scale.shape[3]`

So if the 2D block size is (128, 64), the caller just computes and passes 64. The 128 on the other axis doesn't matter here since we aren't slicing or padding along it

jrplatin: Actually, for the `block_size` / `scale_ratio` default being None, I think keeping the default at `1` makes sense because otherwise I'll need to have a bunch of branching/extra logic to check if `block_size` is None and thus divide by 1 in those cases?

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Agreed, this will simplify things a lot actually!

tpu_inference/layers/common/process_weights/moe_weights.py
kyuyeunk: thanks! so if i understand correctly, when block size is even (multiple of 2), we don't need to un-interleave it since 1 scale will have both odd and even N dim right?

i don't think this config is likely to happen, but hypothetically, what happens if block size is odd? like block size = 1? will this assumption still hold? if not, would adding `assert block_size % 2 == 0` be safe? i think this is kind of config is very unlikely but want to avoid a model silently causing numeric errors - because this kind of error won't even cause a model to crash or anything - which makes it even difficult to detect.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -343,53 +389,115 @@ def process_moe_weights(
...
+                padded_output_sizes=padded_output_sizes,
+                name="w13_weight")
 
             if w13_weight_scale is not None:
-                w13_weight_scale = process_w13_tp(tensor=w13_weight_scale,
-                                                  concat_dim=3,
-                                                  name="w13_weight_scale")
-
+                block_size = None
+                if envs.DISABLE_WEIGHT_REQUANTIZATION:
```
kyuyeunk: nit: i think have thins as an argument to the function & make this function stateless is cleaner?

tpu_inference/layers/common/process_weights/moe_weights.py
kyuyeunk, line 705: is it possible to have disable requantized path & requantize path share the same function exit point / `process_moe_weights` function?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -343,53 +389,115 @@ def process_moe_weights(
...
+                    padded_output_sizes=padded_output_sizes,
+                    name="w13_bias")
+            if w2_weight_scale is not None:
+                block_size = None
+                if envs.DISABLE_WEIGHT_REQUANTIZATION:
+                    # check if cleanly divisible
+                    assert hidden_size % w2_weight_scale.shape[3] == 0
+                    block_size = hidden_size // w2_weight_scale.shape[3]
+                if block_size is not None and block_size > 1:
+                    w2_weight_scale = jnp.repeat(w2_weight_scale,
```
kyuyeunk, line 459: so even when 2d block quantization is used, scale will be repeated to become full shape? is it correct?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -343,53 +389,115 @@ def process_moe_weights(
...
+                    concat_dim=2,
+                    config=pad_config_weight,
+                    padded_output_sizes=padded_output_sizes,
+                    name="w13_bias")
+            if w2_weight_scale is not None:
+                block_size = None
+                if envs.DISABLE_WEIGHT_REQUANTIZATION:
+                    # check if cleanly divisible
+                    assert hidden_size % w2_weight_scale.shape[3] == 0
+                    block_size = hidden_size // w2_weight_scale.shape[3]
```
kyuyeunk: this kind of pattern is used very frequently. maybe its better to declare it at the top of the function or have it as an argument?

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Correct and added comment!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Move it to the top, good suggestion!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Done!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Yep, added, thanks!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Correct and added an assertion to catch!

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -142,29 +141,71 @@ class W13PaddingConfig:
 
 def get_w13_padding_config(intermediate_size: int,
                            reorder_size: int,
-                           align: int = 128) -> W13PaddingConfig:
-    """Calculates padded dimensions and pad amounts for w13 tensors."""
+                           align: int = 128,
+                           block_size: int | None = None) -> W13PaddingConfig:
```
kyuyeunk: so if my understanding correct, `block_size=1` is equivalent to 1d block quantization? if so, sorry for changing my mind on this, but can you revert to using `block_size = 1` to denote 1d block quantization? also, change the name to `outer_block_size`? i think i got confused because i thought `block_size` here meant inner quantization block size. and i realized that using `outer_block_size=1` can simplify the code base to represent both 2d block & 1d block quantization.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -343,53 +404,95 @@ def process_moe_weights(
...
-                                          concat_dim=2,
-                                          name="w13_bias")
+                w13_bias = process_w13_for_gmm(tensor=w13_bias,
+                                               concat_dim=2,
+                                               config=pad_config_weight,
+                                               name="w13_bias")
+
+            if w2_weight_scale is not None:
+                if w2_block_size is not None and w2_block_size > 1:
+                    # GMM currently expects scales to be broadcasted to full shape along the contracting dimension when skipping requantization
```
kyuyeunk: nit: please keep 80 character limit.

similar comment for everywhere else.

kyuyeunk (commented):
okay, i think the overall logic looks good! only one comment and that should be it!

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -510,54 +613,92 @@ def process_fp8_moe_weights(
...
     w13_reorder_size = get_mesh_shape_product(mesh,
                                               ShardingAxisName.MLP_TENSOR)
-    weights = quantize_moe_weights(
-        FusedMoEWeights(
+
+    if envs.DISABLE_WEIGHT_REQUANTIZATION:
+        logger.info_once("Disabled weight requantization")
+
+        in_block_size = weight_block_size[
+            1] if weight_block_size is not None else 128
```
kyuyeunk: oh wait. in block size is being defaulted to 128?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -510,54 +613,92 @@ def process_fp8_moe_weights(
...
+        if w13_weight_scale is not None and w13_weight_scale.ndim == 3:
+            # out_dim = 2 * inter, in_dim = hidden
+            # we want (experts, out_blocks, in_blocks)
+            # check if it is (experts, in_blocks, out_blocks)
+            in_blocks_13 = w13_weight.shape[2] // in_block_size
+            if w13_weight_scale.shape[
```
kyuyeunk: brackets being separated looks really ugly. do you know a better way to fortmat it?

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Whoops good catch, will assert that the incoming `weight_block_size` is not None

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Fixed!

kyuyeunk (approved):

PR #2408, opened 2026-04-27
[Bug][Qwen3.5] Match GPU FLA precision in GDN to close CoT accuracy gap

qizzzh:
Closes the TPU↔GPU quality gap on Qwen3.5-397B GPQA-Diamond CoT (zero-shot, flexible-extract): TPU goes from 0.7929 to 0.8485 (full N=198), slightly above the 0.8343 GPU reference. Mirroring GPU FLA's fp32 precision discipline across the GDN kernel restores parity with negligible perf impact (~0% throughput on a 30-prompt × 512-token decode bench, 957.55 → 956.47 output tokens/sec).

Five fixes, all in the GDN path used by Qwen3-Next / Qwen3.5:

1. has_initial_state plumbing (correctness) GPU's gdn_linear_attn._forward_core does initial_state[~has_initial_state, ...] = 0 for prefills with no prior context, and the equivalent for the conv state via causal_conv1d_fn(has_initial_state=...). vLLM's mamba block manager does NOT zero freed slots — it relies on the kernel to honor the flag. The TPU GDN kernel was reading recurrent_state[slot] / conv_state[slot] unconditionally, so once a freed slot got reused, the new request would silently consume the previous tenant's state and corrupt its entire GDN trajectory.

   Derive `has_initial_state = (seq_lens - query_lens) > 0` from the existing attn metadata in gdn_attention_op.py and plumb it through run_jax_gdn_attention → ragged_conv1d_jax and the chunked + ref ragged_gated_delta_rule. Each masks the gathered slot state to zero where ~has_initial_state. Decode-only branch ignores it because decodes always have prior context.

2. recurrent_gated_delta_rule_step in fp32 with Precision.HIGHEST The decode-side recurrent step had been running at TPU's default precision (bf16 multiply, fp32 accumulate), accumulating ~sqrt(d_k) × bf16_ULP ≈ 1% per-element error in the recurrent state on every decode token. Over the long generations used for CoT eval (max_gen_toks=32768) this drift compounds enough to plausibly flip CoT answers vs. the GPU run, though we did not ablate per-fix to prove it is the dominant contributor.

   GPU FLA's fused_sigmoid_gating_delta_rule_update kernel loads q/k/v/state as fp32 and keeps the entire recurrent update in fp32 registers. Mirror that on TPU with explicit fp32 input casts and Precision.HIGHEST in the two einsums (k_state, q_state). The matmul is 8×128×128 — even at 6× cost (bf16x6) it's a rounding error against the 397B-parameter MoE forward pass, which is HBM-bound.

3. ragged_conv1d_jax fp32 accumulation GPU's causal_conv1d_fn loads x and weights as fp32 before the kernel-size-element sum. The TPU loop was accumulating in bf16, contributing ~1-2% per-channel relative error on every GDN layer's conv output. Promote x, weights, and the gathered conv state to fp32 inside the loop; cast the final out back to the input dtype. The state writeback path stays bf16 (it's a pure shuffle of the last kernel_size-1 tokens, no arithmetic).

4. l2norm fp32 sum-of-squares reduction (chunked + ref) GPU FLA's l2norm_fwd reduces in fp32. We initially left this in bf16 because on a small --limit=30 GPQA sample it appeared the bf16 path was fine. On the full 198-question Diamond the bf16-l2norm + bf16-sigmoid variant landed at 0.8232 (below GPU's 0.8343) and produced ~37% slower wall-clock generation, so the fp32 path is restored.

5. sigmoid(b) for the beta gate in fp32 (chunked + ref) GPU's fused_gdn_gating_kernel casts b to fp32 before sigmoid. Restored alongside (4) for the same reason.

Empirical evidence we have for which fixes matter:

   no fixes (baseline):      0.7929 ± 0.0289   full Diamond
   fixes (1)+(2)+(3):         0.8232 ± 0.0272   full Diamond  (+0.030)
   all 5 fixes:                     0.8485 ± 0.0255   full Diamond  (+0.056,
                                               +0.025 over the
                                               (1)+(2)+(3) subset)

Tests (tests/layers/common/test_gdn_attention.py)
   + test_has_initial_state_zeros_stale_slot_{chunked,ref}: a new prefill landing on a slot pre-populated with random nonzero state must produce the same output and final state as on a fresh-zero slot when has_initial_state=False.
   + test_has_initial_state_preserves_continuation_{chunked,ref}: a chunked prefill in two halves (step A from zero, step B from step A's slot with has_initial_state=True) must match a single-shot prefill of the concatenated stream.
   + test_l2norm_fp32_internal_more_precise_than_bf16_{chunked,ref}: on a bf16 input where sum-of-squares is well above 1.0, the fp32 reduction is at least 2× closer to a fp64 reference than a bf16-only reduction. Catches silent regressions to bf16.
   + test_l2norm_returns_input_dtype_{chunked,ref}: the helpers compute in fp32 internally but must return the input dtype so downstream `compute_dtype` casts in pack_inputs_single_stream keep their layout assumptions. 14/14 GDN tests pass on TPU v7x-8.

Eval (TPU v7x-8, Qwen/Qwen3.5-397B-A17B-FP8,
gpqa_diamond_cot_zeroshot --apply_chat_template seed=42, full N=198)

   Prior TPU baseline:           0.7929 ± 0.0289
   GPU reference:                 0.8343
   This commit (full Diamond):   0.8485 ± 0.0255   ← +0.014 vs GPU

Perf (Qwen3.5-397B-A17B-FP8, TPU v7x-8, 30 prompts × 512 forced output tokens, ignore_eos=True)

   baseline:  957.55 output tokens/sec
   this fix:  956.47 output tokens/sec   (−0.11%)

### DIFF
--- a/tests/layers/common/test_gdn_attention.py
+++ b/tests/layers/common/test_gdn_attention.py
@@ -19,6 +19,10 @@ from absl.testing import parameterized
 
 from tpu_inference.layers.common.gdn_attention import (
     GdnAttentionConfig, RaggedGatedDeltaRuleImpl, run_jax_gdn_attention_local)
+from tpu_inference.layers.common.ragged_gated_delta_rule_chunked import \
+    l2norm as l2norm_chunked
+from tpu_inference.layers.common.ragged_gated_delta_rule_ref import \
+    _l2_normalize as l2_normalize_ref
 
 
 class GDNAttentionTest(parameterized.TestCase):
@@ -155,6 +159,12 @@ class GDNAttentionTest(parameterized.TestCase):
             ],
         )
 
+        # All sequences in this test start from a fresh slot; the existing
+        # parametrizations don't exercise prefix-cache-hit / chunked-prefill
+        # continuation. `has_initial_state=False` reproduces the prior
+        # behavior (zero initial state regardless of slot contents).
+        has_initial_state = jnp.zeros((max_reqs, ), dtype=jnp.bool_)
+
         common_kwargs = dict(
             mixed_qkv=mixed_qkv,
             b=b,
@@ -168,6 +178,7 @@ class GDNAttentionTest(parameterized.TestCase):
             query_start_loc=q_loc,
             state_indices=state_indices,
             distribution=distribution,
+            has_initial_state=has_initial_state,
             n_kq=n_kq,
             n_v=n_v,
             d_k=kq_head_dim,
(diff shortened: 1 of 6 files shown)


tpu_inference/layers/common/gdn_attention.py
```diff
@@ -151,17 +161,34 @@ def run_jax_gdn_attention_local(
...
-        dt_bias,
-        query_start_loc,
-        state_indices,
-        distribution,
-    )
+    if config.ragged_gated_delta_rule_impl == RaggedGatedDeltaRuleImpl.FUSED:
+        # The fused kernel does not yet honor `has_initial_state`. Fall
+        # through with the existing call signature; the chunked/ref paths
+        # used in production already mask via the new argument below.
+        new_recurrent_state, output = ragged_gdn_impl(
```
jrplatin, line 174: Are these call signatures identical except for the `has_initial_state` ? Probably nit but might be simpler to pack/unpack the args as a tuple? E.g.:


```
args = (
    out_mixed_qkv,
    b,
    a,
    recurrent_state,
    A_log,
    dt_bias,
    query_start_loc,
    state_indices,
    distribution,
)

# Append the 10th argument if not using the FUSED implementation
if config.ragged_gated_delta_rule_impl != RaggedGatedDeltaRuleImpl.FUSED:
    args += (has_initial_state,)

new_recurrent_state, output = ragged_gdn_impl(*args)
```

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -11,7 +11,46 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Ragged gated delta rule packed JAX implementation."""
+"""Ragged gated delta rule packed JAX implementation.
+
+Precision discipline (Qwen3.5 397B GPQA fix, 2026-04-26)
```
jrplatin: I think this is probably comment over-kill and maybe just better for the PR -- can we simplify / streamline this a bit?

tpu_inference/layers/common/gdn_attention.py
qizzzh: Thanks for the review. I considered this but decided to leave the explicit form. The duplication is intentional: each call site clearly shows what that impl receives, and a grep for ragged_gdn_impl( lands directly on the per-impl signature. With *args the call site loses that. And further edits to the args may be fragile.

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -11,7 +11,46 @@
...
+  ALL load-bearing — do not revert without re-running the full eval:
+    - ``recurrent_gated_delta_rule_step``: fp32 inputs + HIGHEST matmul.
+    - ``ragged_conv1d_jax``: fp32 conv accumulation.
+    - ``l2norm`` (chunked + ref): fp32 sum-of-squares reduction.
+    - ``sigmoid(b)`` for the beta gate (chunked + ref): fp32 input.
+    - ``has_initial_state`` plumbing in mixed-prefill init carry.
+
+What we ablated (full GPQA-Diamond, N=198, flexible-extract):
+    no fixes:                0.7929
+    fixes (1)+(2)+(3):       0.8232    (+0.030 over baseline)
```
helloworld1: The comment here is overly verbose and does not explain what fix (1) (2) (3) are. Would you please simplify the comments just saying what this module do and the implementation details goes to the python doc of the run_gdn_attention function

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -463,24 +529,52 @@ def recurrent_gated_delta_rule_step(
...
+    every decode step, and over the long generation budgets used here
+    (``max_gen_toks=32768``) it contributes to the GPQA-Diamond gap
+    vs. GPU. See the module docstring for the full ablation; we did
+    not isolate this fix's individual contribution.
+    """
     B, H, d_k = query.shape
     d_v = value.shape[-1]
 
+    orig_dtype = query.dtype
+    query = query.astype(jnp.float32)
```
jrplatin, line 504: So just to clarify -- this is matching what vLLM is doing? Can you share the code pointer? Thanks!

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -463,24 +529,52 @@ def recurrent_gated_delta_rule_step(
...
     scale = d_k**-0.5
     query = query * scale
 
     exp_g = jnp.exp(g)
 
-    k_state = jnp.einsum("bhd, bhdm -> bhm", key, state)
+    k_state = jnp.einsum("bhd, bhdm -> bhm",
+                         key,
+                         state,
+                         precision=jax.lax.Precision.HIGHEST)
```
jrplatin, line 524: Will this not always just be FP32?

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -538,7 +632,10 @@ def ragged_gated_delta_rule_decode_only(
     req_indices = jnp.clip(token_idx, 0, max_reqs - 1)
     valid_mask = token_idx < distribution[2]
 
-    beta = jax.nn.sigmoid(b_reshaped)
+    # See comment on the same expression in
+    # `ragged_gated_delta_rule_mixed_prefill` for why sigmoid runs in
+    # fp32.
+    beta = jax.nn.sigmoid(b_reshaped.astype(jnp.float32))
```
jrplatin, line 593: So this is matching what vLLM is doing on GPU?

tpu_inference/layers/common/ragged_gated_delta_rule_ref.py
```diff
@@ -185,7 +213,10 @@ def scan_fn(carry, xs):
         key_reshaped = curr_k.reshape(B, T, n_kq, d_k)
         value_reshaped = curr_v.reshape(B, T, n_v, d_v)
 
-        beta = jax.nn.sigmoid(curr_b)
+        # Cast b to fp32 before sigmoid to match GPU's
+        # `fused_gdn_gating_kernel`. See module docstring of
+        # `ragged_gated_delta_rule_chunked.py` for context.
+        beta = jax.nn.sigmoid(curr_b.astype(jnp.float32))
```
jrplatin, line 218: Will this return FP32 or BF16?

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
```diff
@@ -104,6 +104,23 @@ def gdn_attention_core_tpu(
...
+    # slots without clearing them, so a new request that lands on a slot
+    # previously held by another request would silently consume that
+    # request's recurrent state. GPU's GDN backend handles this via
+    # `initial_state[~has_initial_state, ...] = 0` in
+    # `gdn_linear_attn._forward_core`; without the equivalent on TPU, every
+    # slot reuse corrupts the new request's GDN trajectory.
+    #
+    # context_len = seq_len - query_len. If > 0, prior tokens exist.
+    j_seq_lens = jax_view(attn_metadata.seq_lens)
+    query_lens = q_loc[1:] - q_loc[:-1]
```
helloworld1: We want to constrain the gdn_attention_op to torch code and torchax translation. Would be better to pass in j_seq_lens to run_jax_gdn_attention() function and handle the computation in run_jax_gdn_attention so the logic can be unit tested.

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: Yes - matches fused_sigmoid_gating_delta_rule_update_kernel in vllm/model_executor/layers/fla/ops/fused_sigmoid_gating.py.

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: My understanding is with Precision.DEFAULT, fp32 inputs to a TPU MXU matmul still get downconverted to bf16 before the multiply (1-pass bf16 mul, fp32 accumulate). So fp32-cast inputs alone don't buy us fp32  precision on the matmul — we also need Precision.HIGHEST (which uses bf16x6 ≈ fp32). I added a one-line inline comment.

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: Done

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: Yes — matches fused_gdn_gating_kernel in vllm/model_executor/layers/mamba/gdn_linear_attn.py

tpu_inference/layers/common/ragged_gated_delta_rule_ref.py
qizzzh: fp32. jax.nn.sigmoid preserves input dtype, so casting curr_b to fp32 before the call gives an fp32 result.

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
helloworld1: BTW, the has_initial_state computation can run after sharding in run_jax_gdn_attention_local since run_jax_gdn_attention_local is unit tested

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
```diff
@@ -463,24 +529,52 @@ def recurrent_gated_delta_rule_step(
     beta: jnp.ndarray,
     state: jnp.ndarray | None = None,
 ) -> tuple[jnp.ndarray, jnp.ndarray]:
-    """Single-step recurrent update for decode."""
+    """Single-step recurrent update for decode.
+
+    Computes the gated-delta-rule step in float32 to match GPU FLA's
```
helloworld1: The float32 comment should not be in the function python doc. Would you put it in line 551 when you convert the types?

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: It's update in the latest commit.

tests/layers/common/test_gdn_attention.py
```diff
@@ -196,3 +207,334 @@ def test_run_jax_gdn_attention_local(self, max_reqs, lengths, q_loc,
...
+                                   atol=1e-5)
+
+    @parameterized.named_parameters(
+        dict(
+            testcase_name="chunked",
+            test_config=GdnAttentionConfig(
+                ragged_gated_delta_rule_impl=RaggedGatedDeltaRuleImpl.CHUNKED),
+        ),
+        dict(
+            testcase_name="ref",
```
helloworld1, line 350: Are you comparing ref with ref? We can remove this test case. Leaving @parameterized.named_parameters here is fine since we will need to compare the kernel impl later

qizzzh: The test isn't comparing impls — within one parametrization both runs use the same impl. What differs between the two runs is the conv_state / recurrent_state contents (fresh-zero vs nonzero stale) with has_initial_state=False in both. The assertion is that the masking makes the kernel output independent of stale slot contents — that's the direct regression test for the production bug we hit (vLLM's mamba pool reuses freed slots without clearing them, and we previously consumed the stale state). The parametrization ensures both chunked and ref honor the masking.

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
qizzzh: Done

tpu_inference/layers/common/ragged_gated_delta_rule_chunked.py
qizzzh: Done

jrplatin (approved):

helloworld1 (approved):

tpu_inference/layers/common/ragged_gated_delta_rule_ref.py
```diff
@@ -103,6 +107,7 @@ def ragged_gated_delta_rule(
     query_start_loc,
     state_indices,
     distribution,
+    has_initial_state,
```
patrickji2014, line 110: Hi Qi, this causes failure below. what is the expectation for the has_initial_state in these test. should we add more test to cover it in fused_gdn_kernel_test.py . 

>       ref_state, ref_o = ragged_gated_delta_rule_ref(
--
mixed_qkv.astype(jnp.float32),
b.astype(jnp.float32),
a.astype(jnp.float32),
h0.astype(jnp.float32),
A_log[None, None, :],  # (1,1,H_v) to match curr_a rank in ref
dt_bias[None, None, :],  # (1,1,H_v) to match curr_a rank in ref
cu_seqlens,
state_indices,
distribution_ref,
n_kq=H_qk,
n_v=H_v,
d_k=K,
d_v=V,
)
E       TypeError: ragged_gated_delta_rule() missing 1 required positional argument: 'has_initial_state'
 
tests/kernels/fused_gdn_kernel_test.py:134: TypeError

qizzzh: I think this PR should fix that - https://github.com/vllm-project/tpu-inference/pull/2431

PR #2413, opened 2026-04-27
[FP8] Resolving HBM spike by forcing loading on CPU before processing

aashishrampal-lab:
# Description

the past.

This PR addresses the HBM spike issue faced while running FP8 quantization models with FusedMoe layers.

While working on running Deepseekv3.1 on v6e-64(multihost environment) I encountered the following error:
image

This is my manifest.yaml:
[manifest.yaml](https://github.com/user-attachments/files/27137293/manifest.yaml)

On some investigation I found that the moe layers are being put in the memory without any sharding [here](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/moe/moe.py#L312)

But in the [process_weights_after_loading](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/quantization/fp8.py#L473), the  weights are being taken from cpu rather than from the TPU and being put into the memory with correct sharding strategy.

The MOE layer tensor shape [256, 2048, 7168] exactly match the 3.5 GiB size of memory being requested.

The error is encountered on v6e because it has a much smaller HBM compared to v7x.

The proposed solution involes skipping shard.put() for FP8FusedMoe Models. Simply removing the line will cause errors for [unquantized](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/quantization/unquantized.py#L70) models as they worker on tensors on tpu.

# Tests

I tested the above manifest.yaml with the updated image and was able to run deepseekv3.1 without the aforementioned issue.


tpu_inference/layers/jax/moe/moe.py
lk-chen: a better approach is to add `*, mesh: jax.sharding.Mesh | None = None` argument for `_load_weights` and pass `mesh` to `shard_put`. If fp8.py, pass cpu_mesh

aashishrampal-lab: Okay i will test this solution and update the pr accordingly.

aashishrampal-lab: Hey @lk-chen I have made and tested the change

lk-chen (approved):

lk-chen (approved):

lk-chen (approved):

PR #2416, opened 2026-04-27
Compact mamba KV cache + GDN op req-slot indexing

qizzzh:
Mamba state is recurrent (one slot per active request), but PR #2366's correctness fix uniformly allocates `num_blocks` slots for every layer group, leaving ~76 GiB of mamba HBM idle on Qwen3.5-397B-A17B-FP8 TP=8. This change reclaims that HBM for the attention pool and unlocks higher concurrency without changing dtype, sharding, or any user-facing API.

KVCacheManager:
* New `_maybe_set_compact_mamba_num_blocks_override` (env `COMPACT_MAMBA_KV_CACHE`, default ON). Caps mamba alloc at `max_num_reqs+1` slots, computes the matching attn `num_gpu_blocks` for the remaining HBM, and exits the legacy two-step floor when it succeeds. Falls back cleanly when HBM probe fails (CPU-only tests) or the user pins `num_gpu_blocks_override`.
* `initialize_kv_cache` allocates mamba arrays with the smaller `_mamba_num_blocks` while attention layers keep `num_blocks`.

GDN op (`gdn_attention_core_tpu`):
* Index mamba state by persistent-batch slot (`arange(max_reqs)`) instead of `block_tables[:, 0]`. Required: `block_tables[:, 0]` returns values in [0, num_attn_blocks) which now exceeds the smaller mamba arrays' leading dim.
* Stale-state guard: zero conv/recurrent state for slots whose request has `num_computed_tokens == 0`, mirroring GPU vLLM's `has_initial_state = context_lens > 0`. Without this, slot reuse leaks the previous occupant's state into the next prefill.

Verified on Qwen3.5-397B-A17B-FP8 / TP=8 / v7x with 1k-in / 8k-out:
* num_gpu_blocks_override 1013 -> 1732 (+71% attn pool).
* Output throughput 1047 -> 1237 tok/s (+18%) at conc=64.
* Mean TPOT 56.6 -> 48.1 ms.
* 0 errors across 128 prompts.

Tests: 29 pass in tests/runner/test_kv_cache_manager.py (3 new), 3 pass in tests/layers/vllm/test_gdn_attention_op.py, 6 pass in tests/layers/common/test_gdn_attention.py.

MMLU-Pro looks good

```
|mmlu_pro           |      2|custom-extract|      |exact_match|↑  |0.8314|±  |0.0139|
| - biology         |      3|custom-extract|     5|exact_match|↑  |0.9200|±  |0.0388|
| - business        |      3|custom-extract|     5|exact_match|↑  |0.8600|±  |0.0496|
| - chemistry       |      3|custom-extract|     5|exact_match|↑  |0.9000|±  |0.0429|
| - computer_science|      3|custom-extract|     5|exact_match|↑  |0.8200|±  |0.0549|
| - economics       |      3|custom-extract|     5|exact_match|↑  |0.9400|±  |0.0339|
| - engineering     |      3|custom-extract|     5|exact_match|↑  |0.7000|±  |0.0655|
| - health          |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0571|
| - history         |      3|custom-extract|     5|exact_match|↑  |0.7400|±  |0.0627|
| - law             |      3|custom-extract|     5|exact_match|↑  |0.7200|±  |0.0641|
| - math            |      3|custom-extract|     5|exact_match|↑  |0.9600|±  |0.0280|
| - other           |      3|custom-extract|     5|exact_match|↑  |0.7200|±  |0.0641|
| - philosophy      |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0571|
| - physics         |      3|custom-extract|     5|exact_match|↑  |0.9400|±  |0.0339|
| - psychology      |      3|custom-extract|     5|exact_match|↑  |0.8200|±  |0.0549|

|mmlu_pro|      2|custom-extract|      |exact_match|↑  |0.8314|±  |0.0139|

### DIFF
--- a/tests/runner/test_input_batch.py
+++ b/tests/runner/test_input_batch.py
@@ -247,16 +247,18 @@ def test_swap_states_only_swaps_active_tokens(input_batch: InputBatch):
 
 
 def test_mamba_state_indices_unique_per_request(input_batch: InputBatch):
-    """Each request must get its own mamba state slot id."""
-    reqs = [create_dummy_request(f"req-{i}") for i in range(4)]
-    for req in reqs:
-        input_batch.add_request(req)
+    """Two concurrent requests must never receive the same mamba slot id —
+    if they did, the GDN op would write both requests' recurrent state
+    into the same physical cache slot and corrupt one of them. Also
+    verifies slot 0 is reserved (vLLM's null block convention)."""
+    for i in range(4):
+        input_batch.add_request(create_dummy_request(f"req-{i}"))
 
     slots = input_batch.mamba_state_indices_cpu[:input_batch.num_reqs].tolist()
     assert len(set(slots)) == len(slots), \
-        f"Mamba slots should be unique per request, got: {slots}"
-    # Slot 0 is reserved as the null block.
-    assert all(s >= 1 for s in slots), f"Slot 0 must be reserved: {slots}"
+        f"Slots not unique across concurrent requests: {slots}"
+    assert all(s >= 1 for s in slots), \
+        f"Slot 0 is the null block and must not be assigned: {slots}"
 
 
 def test_mamba_state_indices_freed_on_remove(input_batch: InputBatch):
--- a/tests/runner/test_input_batch.py
+++ b/tests/runner/test_input_batch.py
@@ -246,6 +246,81 @@ def test_swap_states_only_swaps_active_tokens(input_batch: InputBatch):
     assert input_batch.num_tokens_no_spec[1] == 6
 
 
+def test_mamba_state_indices_unique_per_request(input_batch: InputBatch):
+    """Each request must get its own mamba state slot id."""
+    reqs = [create_dummy_request(f"req-{i}") for i in range(4)]
+    for req in reqs:
+        input_batch.add_request(req)
+
+    slots = input_batch.mamba_state_indices_cpu[:input_batch.num_reqs].tolist()
+    assert len(set(slots)) == len(slots), \
+        f"Mamba slots should be unique per request, got: {slots}"
+    # Slot 0 is reserved as the null block.
+    assert all(s >= 1 for s in slots), f"Slot 0 must be reserved: {slots}"
+
+
+def test_mamba_state_indices_freed_on_remove(input_batch: InputBatch):
+    """Removing a request must return its slot id to the free pool so the
+    next add reuses it (rather than running the pool dry)."""
+    req = create_dummy_request("req-0")
+    input_batch.add_request(req)
+    slot_first = int(input_batch.mamba_state_indices_cpu[0])
+
+    input_batch.remove_request("req-0")
+    # condense() with empty_indices=[0] would early-return because num_reqs==0,
+    # so we just verify the slot is back in the pool.
+    assert slot_first in input_batch._free_mamba_slots
+
+    new_req = create_dummy_request("req-new")
+    input_batch.add_request(new_req)
+    # The most-recently-freed slot is at the end of the free list — pop()
+    # returns it, so the new request gets the same id back.
+    assert int(input_batch.mamba_state_indices_cpu[0]) == slot_first
+
+
+def test_mamba_state_indices_follow_condense(input_batch: InputBatch):
+    """When condense moves a request to a different persistent-batch slot,
+    its mamba state id must follow it — otherwise the GDN op reads stale
+    state from the recurrent_state cache. This is the bug that broke gsm8k
+    accuracy on Qwen3.5-397B (only triggered when requests finish out of
+    order, which `--ignore-eos` benchmarks suppress)."""
+    reqs = [create_dummy_request(f"req-{i}") for i in range(4)]
+    for req in reqs:
+        input_batch.add_request(req)
+
+    # Snapshot the slot id assigned to each request before any churn.
+    slot_for_req = {
+        rid: int(input_batch.mamba_state_indices_cpu[idx])
+        for rid, idx in input_batch.req_id_to_index.items()
+    }
+
+    # Remove the lower-indexed requests so condense has to move the higher
+    # ones down: [req-0, req-1, req-2, req-3] → [req-3, req-2, _, _].
+    input_batch.remove_request("req-0")
+    input_batch.remove_request("req-1")
+    input_batch.condense(sorted([0, 1], reverse=True))
+
+    # After condense, indexing-by-persistent-slot must yield the same slot
+    # id each request had before the move.
+    assert input_batch.req_id_to_index["req-3"] == 0
+    assert input_batch.req_id_to_index["req-2"] == 1
+    assert int(input_batch.mamba_state_indices_cpu[0]) == slot_for_req["req-3"]
+    assert int(input_batch.mamba_state_indices_cpu[1]) == slot_for_req["req-2"]
+
+
+def test_mamba_state_indices_swap(input_batch: InputBatch):
+    """swap_states swaps the request mappings, so the slot ids must swap too."""
+    input_batch.add_request(create_dummy_request("req-0"))
+    input_batch.add_request(create_dummy_request("req-1"))
+    slot0 = int(input_batch.mamba_state_indices_cpu[0])
+    slot1 = int(input_batch.mamba_state_indices_cpu[1])
+
+    input_batch.swap_states(0, 1)
+
+    assert int(input_batch.mamba_state_indices_cpu[0]) == slot1
+    assert int(input_batch.mamba_state_indices_cpu[1]) == slot0
+
+
 def test_all_greedy_property(input_batch: InputBatch):
     """Tests the `all_greedy` property."""
     # Initially true
--- a/tests/runner/test_kv_cache_manager.py
+++ b/tests/runner/test_kv_cache_manager.py
@@ -917,6 +917,115 @@ class TestKVCacheManager:
                 f"layer {name} has page_size_bytes={spec.page_size_bytes} "
                 f"but expected uniform={expected_uniform}")
 
+    def test_compact_mamba_override_caps_mamba_and_grows_attn(self):
+        """Compact-mamba: with HBM available, mamba should be capped at
+        max_num_reqs+1 (rounded to divisor) and num_gpu_blocks_override
+        should grow beyond what the legacy single-num_blocks path gives."""
+        from tpu_inference.runner.kv_cache_manager import KVCacheManager
+
+        manager = KVCacheManager(self.runner)
+        # Force the use_mla branch off so divisor uses ATTN_DATA.
+        manager.use_mla = False
+        # 304 GiB of available HBM, 4-device mesh — match the Qwen3.5 TP=8
+        # situation as closely as a CPU mesh allows. Returns (used, limit)
+        # per device summing to total_used=0, total_limit=304 GiB.
+        avail_per_device = 304 * (1 << 30) // 4
+
+        with patch(
+                "tpu_inference.runner.kv_cache_manager.utils.hbm_usage_bytes",
+                return_value=[(0, avail_per_device)] * 4):
+            self.runner.cache_config.gpu_memory_utilization = 1.0
+            self.runner.cache_config.num_gpu_blocks_override = None
+            # Pretend the model wants 256 max in-flight requests; mamba
+            # should be capped at 257 (rounded up to divisor=1).
+            self.runner.scheduler_config = MagicMock(max_num_seqs=256)
+            self.runner.max_num_reqs = 256
+            # Big mamba (4 MiB / state) and modest attn page (1 MiB / block)
+            # so the compact reshuffle is a clear win.
+            attn_page = 1 << 20  # 1 MiB
+            unpadded_mamba = 4 << 20  # 4 MiB
+            # Set the state that update_mamba_page_size_padded would set if
+            # we were calling it from a real run; the compact override
+            # function reads `_hybrid_uniform_page_size_bytes` for its
+            # legacy-comparison sanity check.
+            uniform = attn_page + 3 * unpadded_mamba
+            manager._hybrid_uniform_page_size_bytes = uniform
+            manager._maybe_set_compact_mamba_num_blocks_override(
+                attn_page_size_bytes=attn_page,
+                unpadded_mamba_page_size_bytes=unpadded_mamba,
+                num_attn_groups=1,
+                num_mamba_groups=3,
+                num_attn_layers=15,
+                num_mamba_layers=45,
+                group_size=15,
+            )
+
+        assert manager._mamba_num_blocks == 257
+        assert self.runner.cache_config.num_gpu_blocks_override is not None
+        # Attention pool must beat the legacy single-num_blocks calc — the
+        # whole point of the optimization. Quick sanity:
+        # legacy = avail / (uniform × group_size).
+        legacy = (304 * (1 << 30)) // (uniform * 15)
+        assert self.runner.cache_config.num_gpu_blocks_override > legacy
+
+    def test_compact_mamba_override_falls_back_when_hbm_unreadable(self):
+        """If hbm_usage_bytes raises (e.g., on CPU), compact override is a
+        no-op and the legacy path takes over."""
+        from tpu_inference.runner.kv_cache_manager import KVCacheManager
+
+        manager = KVCacheManager(self.runner)
+        manager.use_mla = False
+        manager._hybrid_uniform_page_size_bytes = 1 << 20
+
+        with patch(
+                "tpu_inference.runner.kv_cache_manager.utils.hbm_usage_bytes",
+                side_effect=RuntimeError("no devices")):
+            self.runner.cache_config.num_gpu_blocks_override = None
+            self.runner.scheduler_config = MagicMock(max_num_seqs=256)
+            self.runner.max_num_reqs = 256
+            manager._maybe_set_compact_mamba_num_blocks_override(
+                attn_page_size_bytes=1 << 20,
+                unpadded_mamba_page_size_bytes=4 << 20,
+                num_attn_groups=1,
+                num_mamba_groups=3,
+                num_attn_layers=15,
+                num_mamba_layers=45,
+                group_size=15,
+            )
+
+        assert manager._mamba_num_blocks is None
+        assert self.runner.cache_config.num_gpu_blocks_override is None
+
+    def test_compact_mamba_override_respects_user_override(self):
+        """If the user already set num_gpu_blocks_override, compact path
+        must not clobber it."""
+        from tpu_inference.runner.kv_cache_manager import KVCacheManager
+
+        manager = KVCacheManager(self.runner)
+        manager.use_mla = False
+        manager._hybrid_uniform_page_size_bytes = 1 << 20
+
+        with patch(
+                "tpu_inference.runner.kv_cache_manager.utils.hbm_usage_bytes",
+                return_value=[(0, 304 * (1 << 30) // 4)] * 4):
+            self.runner.cache_config.num_gpu_blocks_override = 999
+            self.runner.scheduler_config = MagicMock(max_num_seqs=256)
+            self.runner.max_num_reqs = 256
+            manager._maybe_set_compact_mamba_num_blocks_override(
+                attn_page_size_bytes=1 << 20,
+                unpadded_mamba_page_size_bytes=4 << 20,
+                num_attn_groups=1,
+                num_mamba_groups=3,
+                num_attn_layers=15,
+                num_mamba_layers=45,
+                group_size=15,
+            )
+
+        # User's override must survive untouched and we must skip the
+        # compact path so initialize_kv_cache uses legacy sizing.
+        assert manager._mamba_num_blocks is None
+        assert self.runner.cache_config.num_gpu_blocks_override == 999
+
     def test_get_kv_cache_spec_pure_attention_no_cache_config_updates(self):
         mock_attn = MagicMock(spec=MambaBase)
         layers = {'layer.0': mock_attn}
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: tests/runner/test_input_batch.py [hunk 1]
```diff
@@ -246,6 +246,81 @@ def test_swap_states_only_swaps_active_tokens(input_batch: InputBatch):
...
+
+def test_mamba_state_indices_freed_on_remove(input_batch: InputBatch):
+    """Removing a request must return its slot id to the free pool so the
+    next add reuses it (rather than running the pool dry)."""
+    req = create_dummy_request("req-0")
+    input_batch.add_request(req)
+    slot_first = int(input_batch.mamba_state_indices_cpu[0])
+
+    input_batch.remove_request("req-0")
+    # condense() with empty_indices=[0] would early-return because num_reqs==0,
```
jrplatin, line 272: I don't follow this

tests/runner/test_kv_cache_manager.py
```diff
@@ -917,6 +917,115 @@ def mamba_type(self):
...
                 f"but expected uniform={expected_uniform}")
 
+    def test_compact_mamba_override_caps_mamba_and_grows_attn(self):
+        """Compact-mamba: with HBM available, mamba should be capped at
+        max_num_reqs+1 (rounded to divisor) and num_gpu_blocks_override
+        should grow beyond what the legacy single-num_blocks path gives."""
+        from tpu_inference.runner.kv_cache_manager import KVCacheManager
+
+        manager = KVCacheManager(self.runner)
+        # Force the use_mla branch off so divisor uses ATTN_DATA.
```
jrplatin: Don't follow this either

tests/runner/test_kv_cache_manager.py
```diff
@@ -917,6 +917,115 @@ def mamba_type(self):
...
+                num_attn_groups=1,
+                num_mamba_groups=3,
+                num_attn_layers=15,
+                num_mamba_layers=45,
+                group_size=15,
+            )
+
+        assert manager._mamba_num_blocks == 257
+        assert self.runner.cache_config.num_gpu_blocks_override is not None
+        # Attention pool must beat the legacy single-num_blocks calc — the
```
jrplatin: What do you mean by "beat"?

tests/runner/test_kv_cache_manager.py
```diff
@@ -917,6 +917,115 @@ def mamba_type(self):
...
+        # whole point of the optimization. Quick sanity:
+        # legacy = avail / (uniform × group_size).
+        legacy = (304 * (1 << 30)) // (uniform * 15)
+        assert self.runner.cache_config.num_gpu_blocks_override > legacy
+
+    def test_compact_mamba_override_falls_back_when_hbm_unreadable(self):
+        """If hbm_usage_bytes raises (e.g., on CPU), compact override is a
```
jrplatin: Raises?

tpu_inference/layers/common/attention_metadata.py
```diff
@@ -44,6 +45,10 @@ class AttentionMetadata(object):
     query_start_loc: jax.Array = None
     # (3,)
     request_distribution: jax.Array = None
+    # (max_num_seqs,) int32 — per-request physical mamba slot id; stable
```
jrplatin: This is a bit hard to follow / isn't obvious to most users I imagine, maybe just simplify what this is trying to say

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
```diff
@@ -93,13 +93,19 @@ def gdn_attention_core_tpu(
...
-    # Map physical cache blocks
-    flat_block_tables = jax_view(attn_metadata.block_tables)
-    max_reqs = attn_metadata.seq_lens.shape[0]
-    max_blocks_per_req = flat_block_tables.shape[0] // max_reqs
-    block_tables_2d = jnp.reshape(flat_block_tables,
-                                  (max_reqs, max_blocks_per_req))
-    state_indices = block_tables_2d[:, 0].astype(jnp.int32)
+    # Index mamba state by the per-request slot id maintained by
+    # `InputBatch.mamba_state_indices_cpu`, **not** by the persistent-batch
+    # position. The two diverge whenever `condense` shuffles requests
```
jrplatin: What is "condense"? Might be good to expand on that

tpu_inference/runner/input_batch.py
```diff
@@ -134,6 +134,21 @@ def __init__(
 
         self.request_distribution: list[int] = [0, 0, 0]
 
+        # Mamba slot tracking for hybrid attn+mamba models. Each request gets
+        # a unique mamba state slot id when it enters the batch and keeps it
+        # for the request's lifetime. The mapping `req_idx -> mamba_slot_id`
+        # is stored in `mamba_state_indices_cpu`. Why: vLLM's `condense` moves
```
jrplatin: Can you add a permalink to `condense`? Thanks!

tpu_inference/envs.py
```diff
@@ -281,6 +282,14 @@ def _get_str_list_env() -> list[str]:
     # kv offload to dram: prometheus metrics log interval in seconds
     "TPU_OFFLOAD_METRICS_LOG_INTERVAL":
     lambda: int(os.getenv("TPU_OFFLOAD_METRICS_LOG_INTERVAL", "10")),
+    # Cap mamba layer allocation at `max_num_seqs+1` slots and redirect freed
+    # HBM into more attention blocks. Mamba state is recurrent (one slot per
+    # active request), so allocating num_blocks slots wastes most of them.
+    # Big throughput win on hybrid attn+mamba models like Qwen3.5. Disable
+    # only if you hit a regression that traces back to this.
```
jrplatin: I don't see why we need to toggle this -- why would we want to enable a bug?

tpu_inference/envs.py
```diff
@@ -281,6 +282,14 @@ def _get_str_list_env() -> list[str]:
...
+    "COMPACT_MAMBA_KV_CACHE":
+    lambda: os.getenv("COMPACT_MAMBA_KV_CACHE", "1") not in
```
helloworld1: Why this is not using env_bool("COMPACT_MAMBA_KV_CACHE", default=True),

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+            return
+
+        # Sanity check: redirecting HBM into attention should be a strict
+        # win over the legacy single-num_blocks layout. If it isn't (rare
+        # corner case where the legacy num_blocks already saturates), keep
+        # the legacy path.
+        legacy_num_blocks = avail // (self._hybrid_uniform_page_size_bytes *
+                                      group_size)
+        if attn_num_blocks <= legacy_num_blocks:
+            logger.debug(
```
helloworld1: Feels like a warning to me since it has performance impact.

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+                "<= legacy_num_blocks=%d. Falling back to legacy override.",
+                attn_num_blocks, legacy_num_blocks)
+            return
+
+        cache_config.num_gpu_blocks_override = int(attn_num_blocks)
+        self._mamba_num_blocks = int(mamba_num_blocks)
+
+        logger.info(
+            "Compact-mamba KV cache: num_gpu_blocks_override=%d (attn) and "
+            "_mamba_num_blocks=%d (vs legacy num_blocks≈%d). HBM split: "
```
helloworld1: Would you make the variable name more clear? "legacy" means nothing for people has no idea of history. Probably call full_alloc for old vs compact_alloc for new

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+    def _maybe_set_compact_mamba_num_blocks_override(
+            self, attn_page_size_bytes: int,
+            unpadded_mamba_page_size_bytes: int, num_attn_groups: int,
+            num_mamba_groups: int, num_attn_layers: int, num_mamba_layers: int,
+            group_size: int) -> None:
+        """Cap mamba allocation at `max_num_seqs+1` slots, redirect freed HBM
+        into more attention blocks, and pin `num_gpu_blocks_override` to the
+        resulting attention pool size.
+
+        Why this is a big win
```
helloworld1: Instead of claiming "win" in the comment, explain in a more concise way 
1. When compact allocation should / should not be enabled
2. Pro / cons for compact vs full allocation

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+        ---------------------
+        Mamba state is recurrent: one slot per *active request* per layer,
+        independent of context length. With Qwen3.5 (15 attn + 45 mamba
+        layers) the legacy code allocates `num_blocks` mamba slots per layer
+        even though only ~`max_num_seqs` are ever live at once. On TP=8
+        with 307 GiB available, that wastes ~80 GiB on mamba slots that
```
helloworld1: Leave the benchmark result in PR comments. Use more concise math formula showing how much saving is expected

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
-        # budget against imprecision in vLLM's `avail` estimate.
+        # Pin vLLM's num_blocks. Two cases:
+        #  (a) compact-mamba: cap each mamba layer at `max_num_seqs+1` slots
+        #      (one per active request, plus the null block) and redirect the
+        #      freed HBM into more attention blocks. This is the common path
+        #      and gives a substantial throughput win on hybrid models where
+        #      mamba layers dominate the per-block footprint (e.g., Qwen3.5
+        #      where each block currently spends ~50% of its bytes on mamba
+        #      state that is never indexed past `max_num_seqs`).
+        #  (b) legacy: keep TPU and vLLM's num_blocks identical (the original
```
jrplatin: Same as above/below -- I don't see why we should support this if the compact-mamba option is stricly better?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
                                             group_size)
 
+    def _maybe_set_compact_mamba_num_blocks_override(
+            self, attn_page_size_bytes: int,
+            unpadded_mamba_page_size_bytes: int, num_attn_groups: int,
+            num_mamba_groups: int, num_attn_layers: int, num_mamba_layers: int,
+            group_size: int) -> None:
+        """Cap mamba allocation at `max_num_seqs+1` slots, redirect freed HBM
+        into more attention blocks, and pin `num_gpu_blocks_override` to the
+        resulting attention pool size.
```
jrplatin: This description is too verbose -- can we just simplify this down to what this function does and, maybe, how it solves an issue?

tpu_inference/layers/common/attention_metadata.py
```diff
@@ -44,6 +45,10 @@ class AttentionMetadata(object):
...
+    # across condense moves. Only consumed by hybrid attn+mamba models
+    # (e.g. Qwen3.5); None for pure-attention models.
+    mamba_state_indices: jax.Array | None = None
```
helloworld1, line 55: Could you explain why it is only applicable to hybrid model? Is it usable for pure linear attention model like mamba 2/ 3?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+                "num_mamba_groups=%d × mamba_unpadded=%d exceeds per-tensor "
+                "budget %d. Falling back to legacy override.",
+                mamba_num_blocks, num_mamba_groups,
+                unpadded_mamba_page_size_bytes, avail_per_tensor)
+            return
+
+        attn_num_blocks = attn_per_tensor_avail // (num_attn_groups *
+                                                    attn_page_size_bytes)
+        attn_num_blocks = (attn_num_blocks // divisor) * divisor
+        if attn_num_blocks <= 0:
```
helloworld1, line 398: Should you display a logger.warn or error  here if attn_num_blocks are 0 that we cannot allocate any?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+        total_limit = sum(limit for _, limit in hbm_usage)
+        total_used = sum(used for used, _ in hbm_usage)
+        gpu_mem_util = cache_config.gpu_memory_utilization
+        avail = int(total_limit * gpu_mem_util - total_used)
+        if avail <= 0:
+            return
+
+        # Sharding divisor: num_blocks for each layer must be a multiple of
+        # the per-axis device count so JAX shardings don't round up. MLA
+        # uses MLP_TENSOR; everything else uses ATTN_DATA. Match
```
jrplatin: Can you add a peramlink / reference to where you are getting this info from?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+        # `initialize_kv_cache`.
+        if self.use_mla and not self.runner.vllm_config.additional_config.get(
+                "sharding", {}).get("sharding_strategy", {}).get(
+                    "enable_dp_attention", False):
+            divisor = common_utils.get_mesh_shape_product(
+                self.runner.mesh, ShardingAxisName.MLP_TENSOR)
+        else:
+            divisor = common_utils.get_mesh_shape_product(
+                self.runner.mesh, ShardingAxisName.ATTN_DATA)
+        if divisor <= 0:
```
jrplatin: When would this be the case?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+            (mamba_num_blocks + divisor - 1) // divisor) * divisor
+
+        # Compute attn_num_blocks that fits the remaining HBM after mamba.
+        avail_per_tensor = avail // group_size
+        mamba_per_tensor = (num_mamba_groups * mamba_num_blocks *
+                            unpadded_mamba_page_size_bytes)
+        attn_per_tensor_avail = avail_per_tensor - mamba_per_tensor
+        if attn_per_tensor_avail <= 0:
+            # Even mamba alone exceeds budget — the mamba cap is too high
+            # for this configuration. Fall back to the legacy override.
```
jrplatin: Not sure I 100% follow what's going on here -- if the compact Mambda alone exceeds the budget, then how would the legacy / non-compact work?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+            logger.debug(
+                "Compact-mamba override skipped: mamba_num_blocks=%d × "
+                "num_mamba_groups=%d × mamba_unpadded=%d exceeds per-tensor "
+                "budget %d. Falling back to legacy override.",
+                mamba_num_blocks, num_mamba_groups,
+                unpadded_mamba_page_size_bytes, avail_per_tensor)
+            return
+
+        attn_num_blocks = attn_per_tensor_avail // (num_attn_groups *
+                                                    attn_page_size_bytes)
```
jrplatin, line 396: `attn_page_size_bytes` is per-group, not per-layer right?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -252,18 +263,201 @@ def update_mamba_page_size_padded(
...
+
+        logger.info(
+            "Compact-mamba KV cache: num_gpu_blocks_override=%d (attn) and "
+            "_mamba_num_blocks=%d (vs legacy num_blocks≈%d). HBM split: "
+            "attn=%d layers × %d blocks × %d B = %.2f GiB; mamba=%d layers "
+            "× %d slots × %d B = %.2f GiB; total=%.2f GiB / avail=%.2f GiB.",
+            attn_num_blocks, mamba_num_blocks, legacy_num_blocks,
+            num_attn_layers, attn_num_blocks, attn_page_size_bytes,
+            num_attn_layers * attn_num_blocks * attn_page_size_bytes /
+            (1 << 30), num_mamba_layers, mamba_num_blocks,
```
jrplatin: Nit: but I think we prefer `2 ** 30` rather than `1 << 30`

tpu_inference/envs.py
qizzzh: Removed it. It's mostly to fallback if we hit problems.

tpu_inference/envs.py
qizzzh: Removed it and make it the only path.

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/layers/common/attention_metadata.py
qizzzh: Updated the comment: pure-mamba models would also use this field - there's nothing hybrid-specific about it. Today only hybrid models exercise it because we don't run a pure-mamba model on this stack yet.

tpu_inference/layers/common/attention_metadata.py
qizzzh: Rewrote: shape/dtype/range up front, then who consumes it (mamba/GDN ops) and why we don't go through block_tables (mamba pool is smaller than attn pool under compact-mamba sizing)

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
qizzzh: Expanded with a permalink to upstream vLLM's condense. The comment now spells out the two reasons we use slot ids: (1) compact-mamba's smaller leading dim, (2) condense moves requests between persistent-batch slots and the slot id has to follow the request.

tpu_inference/runner/input_batch.py
qizzzh: done

tpu_inference/runner/kv_cache_manager.py
qizzzh: Done

tpu_inference/runner/kv_cache_manager.py
qizzzh: Updated

tpu_inference/runner/kv_cache_manager.py
qizzzh: get_mesh_shape_product returns 1 for an absent axis, so this only guards against a pathological mesh shape producing 0. Replaced with divisor = max(divisor, 1) and a comment saying so.

tpu_inference/runner/kv_cache_manager.py
qizzzh: Good point - legacy would overflow too, so the fallback was meaningless. Removed it. We now logger.warning with the math and skip the override; if the config genuinely doesn't fit, vLLM will OOM with its uniform layout and the operator gets a clear signal rather than a silent perf cliff.

tpu_inference/runner/kv_cache_manager.py
qizzzh: Per-block per-attention-layer. Each kv-cache tensor backs num_attn_groups attention layers, so the per-tensor attention cost is num_attn_groups × N_attn × attn_page_size_bytes. Added an inline comment to make this explicit.

tpu_inference/runner/kv_cache_manager.py
qizzzh: Done

tests/runner/test_input_batch.py
qizzzh: Rewrote the docstring to spell out the why: two concurrent requests sharing a slot id would have the GDN op write both their recurrent states into the same physical cache slot. Slot 0 is reserved for vLLM's null-block convention.

tests/runner/test_kv_cache_manager.py
qizzzh: Rewrote with a _run_compact_mamba_override helper that pins the Qwen3.5-shaped layer counts in one place. The assertion now compares against the exact expected_attn derived from the sizing formula instead of "must beat legacy".

tests/runner/test_kv_cache_manager.py
qizzzh: Removed the framing. The legacy path is gone, so there's nothing to beat — the test now asserts the precise expected number of attention blocks from the sizing math.

tests/runner/test_kv_cache_manager.py
qizzzh: Renamed the test test_compact_mamba_override_skipped_when_hbm_probe_fails and rewrote the docstring: when hbm_usage_bytes raises (which happens on CPU-only test machines because there's no real device to query), compact-mamba silently skips the override.

helloworld1 (approved):

jrplatin (approved):

Issue #899, opened 2025-10-20
[Feature]: Add embedding model functionality to tpu-inference


bvrockwell:
### 🚀 The feature, motivation and pitch

copied from [feature request in vllm upstream](https://github.com/vllm-project/vllm/issues/20869)

Great first issue!

Should enable for both torchax and jax (if needed).

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

carlesoctav:
I’ve already done some preliminary coding and research on how we might achieve it, but I’m still not entirely sure what the best approach is. Could someone review my plan and give feedback?
For now, I’ll try integrating the Torchax version first:

I’ll add a new metadata class similar to `TPUSamplingMetadata`, called `TPUPoolingMetadata`, and construct it in `runner._prepare_input()`. This metadata will handle pooling, analogous to how sampling works in `layers/jax/sample`. Initially, I thought we could follow the same pattern as the sampler, creating a pooling function that takes `hidden_states` and `TPUPoolingMetadata` as arguments (see next paragraph for the issue).

The problem is that when using `vllm_model_wrapper`, pooling is actually part of the Torch model (under `model.pooler`). In contrast, sampling doesn’t depend on the Torch model, it just uses logits and sampling metadata. The pooler requires both `hidden_states` and the (Torch) pooling metadata. Because of this model dependency, should we instead add a new `jit_pool_fn` inside `VLLMModelWrapper` rather than mirroring how `sample` works (which is implemented purely in JAX even though we use Torchax)?

```python
    def jit_pool_fn(self):
        @jax.jit
        def pool_fn(
            self,
            params_and_buffers: Any,
            hidden_states: jax.Array,
            pooling_metadata: TPUPoolingMetadata,
            lora_metadata,
        ) -> list[torch.Tensor | None]:
            if not self._is_pooling_model:
                raise RuntimeError("Pooling requested for a non-pooling model")

            hidden_states_torch = torch_view(hidden_states)
            lora_metadata = torch_view(lora_metadata)

            with torchax.default_env(), set_vllm_model_wrapper_context(
                    kv_caches=None, mesh=self.mesh):
                original_lora_metadata = replace_lora_metadata(
                    self.model, lora_metadata, self.vllm_config.lora_config)

[...]

PR #2420, opened 2026-04-27
Support StepPool and Qwen3-Embedding (8B) for long-sequence workloads

anthonsu:
This PR enables full support for the Qwen3-Embedding-8B model on v7x (Ironwood) TPUs. It specifically addresses the needs of long-context (4K+ tokens) embedding tasks by implementing StepPool support and a hybrid execution path.

#### Key Changes

* **StepPool Implementation:** Added support to handle Chunked Prefill in the pooling layer, allowing 4K+ tokens to be pooled correctly across steps.
* **Hybrid Architecture:** Integrated the native vLLM PyTorch Pooler into the JAX (flax_nnx) backbone. This allows us to use standard pooling logic while benefiting from TPU acceleration for the transformer blocks.
* **Persistence & Preemption:** Migrated pooling states to CachedRequestState in input_batch.py to ensure metadata survives request preemption.
* **Vocab Padding:** Implemented unified padding (e.g., 151665 -> 151666 for TP=2) to prevent IndivisibleError on Ironwood hardware.

Verified the implementation by comparing TPU outputs with CPU baseline outputs for a 7K token sequence.

#### TODOs
- [x] Implement metadata population for `prompt_token_ids`.
- [x] Fix persistence in `CachedRequestState`.
- [x] Add automated routing for pooling tasks in `model_loader.py`.
- [x] Verify precision on v7x and v6e hardware.
- [x] Updated `model`, `kernel` and `feature` matrices tables of v7x and v6e with embedding related entries.

#### Other Notes
*   **CPU-First Strategy:** Final pooling aggregation is performed on the CPU to avoid the massive HBM overhead (~8.5 GB for 256 seqs) and JAX JIT recompilation hazards associated with dynamic token filtering on TPU.
*   **Dependency:** This PR depends on the vLLM fix for `StepPool.forward`. https://github.com/vllm-project/vllm/pull/41049 (LKG version needs update once the dependency PR checked-in)

#### Tests
Verified the implementation  on **v7x (Ironwood)** and **v6e (Trillium)** using 7K token sequences. Achieved **~0.9997 cosine similarity** compared to CPU baselines.

| Input Type | v7x Cosine Similarity | v6e Cosine Similarity |
| :--- | :--- |:--- |
| English (Short) | 0.9995546074 | 0.9996544555|
| English (Long) | 0.9997845670 | 0.9997664366|
| Japanese (Long) | 0.9997769296 | 0.9997792912|
| Korean (Long) | 0.9997860590 | 0.9998040753|

#### Appendix
The verification was performed using a custom `embed_script.py` with `max_num_batched_tokens=512` and long sequence multi-lingual inputs.

Addresses #899

### DIFF
--- a/.buildkite/models/Qwen_Qwen3-Embedding-8B.yml
+++ b/.buildkite/models/Qwen_Qwen3-Embedding-8B.yml
@@ -1,0 +1,89 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pipeline-name: Qwen/Qwen3-Embedding-8B
+# pipeline-type: embedding
+steps:
+  - label: "${TPU_VERSION:-tpu6e} Unit tests for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_UnitTest"
+    soft_fail: true
+    agents:
+      queue: "${TPU_QUEUE_SINGLE:-tpu_v6e_queue}"
+    commands:
+      - .buildkite/scripts/run_in_docker.sh python3 -m pytest -s -v /workspace/tpu_inference/tests/models/jax/test_qwen3.py::TestQwen3ForCausalLM::test_vocab_padding_for_sharding
+  - label: "${TPU_VERSION:-tpu6e} Record unit test result for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_record_Qwen_Qwen3-Embedding-8B_UnitTest"
+    depends_on: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_UnitTest"
+    env:
+      CI_TPU_VERSION: "${TPU_VERSION:-tpu6e}"
+      CI_TARGET: "Qwen/Qwen3-Embedding-8B"
+      CI_STAGE: "UnitTest"
+      CI_CATEGORY: "embedding"
+    agents:
+      queue: cpu
+    commands:
+      - |
+        .buildkite/scripts/record_step_result.sh ${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_UnitTest
+
+  - label: "${TPU_VERSION:-tpu6e} Accuracy for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Accuracy"
+    depends_on: "${TPU_VERSION:-tpu6e}_record_Qwen_Qwen3-Embedding-8B_UnitTest"
+    soft_fail: true
+    agents:
+      queue: "${TPU_QUEUE_SINGLE:-tpu_v6e_queue}"
+    env:
+      TEST_MODEL: Qwen/Qwen3-Embedding-8B
+      TENSOR_PARALLEL_SIZE: "${TENSOR_PARALLEL_SIZE_SINGLE:-1}"
+    commands:
+      - |
+        .buildkite/scripts/run_in_docker.sh python3 -m pytest -s -v /workspace/tpu_inference/tests/e2e/test_step_pooling.py
+  - label: "${TPU_VERSION:-tpu6e} Record accuracy result for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_record_Qwen_Qwen3-Embedding-8B_Accuracy"
+    depends_on: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Accuracy"
+    env:
+      CI_TPU_VERSION: "${TPU_VERSION:-tpu6e}"
+      CI_TARGET: "Qwen/Qwen3-Embedding-8B"
+      CI_STAGE: "Accuracy/Correctness"
+      CI_CATEGORY: "embedding"
+    agents:
+      queue: cpu
+    commands:
+      - |
+        .buildkite/scripts/record_step_result.sh ${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Accuracy
+
+  - label: "${TPU_VERSION:-tpu6e} Performance benchmarks for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Benchmark"
+    depends_on: "${TPU_VERSION:-tpu6e}_record_Qwen_Qwen3-Embedding-8B_Accuracy"
+    soft_fail: true
+    agents:
+      queue: "${TPU_QUEUE_SINGLE:-tpu_v6e_queue}"
+    env:
+      TEST_MODEL: Qwen/Qwen3-Embedding-8B
+      TENSOR_PARALLEL_SIZE: "${TENSOR_PARALLEL_SIZE_SINGLE:-1}"
+    commands:
+      - |
+        buildkite-agent meta-data set "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Benchmark" "unverified"
+  - label: "${TPU_VERSION:-tpu6e} Record performance benchmark result for Qwen/Qwen3-Embedding-8B"
+    key: "${TPU_VERSION:-tpu6e}_record_Qwen_Qwen3-Embedding-8B_Benchmark"
+    depends_on: "${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Benchmark"
+    env:
+      CI_TPU_VERSION: "${TPU_VERSION:-tpu6e}"
+      CI_TARGET: "Qwen/Qwen3-Embedding-8B"
+      CI_STAGE: "Benchmark"
+      CI_CATEGORY: "embedding"
+    agents:
+      queue: cpu
+    commands:
+      - |
+        .buildkite/scripts/record_step_result.sh ${TPU_VERSION:-tpu6e}_Qwen_Qwen3-Embedding-8B_Benchmark
--- a/support_matrices/nightly/v6e/default/model_support_matrix.csv
+++ b/support_matrices/nightly/v6e/default/model_support_matrix.csv
@@ -1,4 +1,5 @@
 Model,Type,UnitTest,Accuracy/Correctness,Benchmark
+"Qwen/Qwen3-Embedding-8B",Embedding,❓ Untested,❓ Untested,❓ Untested
 "Qwen/Qwen2.5-VL-7B-Instruct",Multimodal,✅ Passing,✅ Passing,✅ Passing
 "Qwen/Qwen3-Omni-30B-A3B-Instruct",Multimodal,❓ Untested,❓ Untested,❓ Untested
 "Qwen/Qwen3-VL-8B-Instruct",Multimodal,❓ Untested,❓ Untested,❓ Untested
--- a/support_matrices/release/v6e/default/feature_support_matrix.csv
+++ b/support_matrices/release/v6e/default/feature_support_matrix.csv
@@ -15,5 +15,5 @@ Feature,CorrectnessTest,PerformanceTest
 "multi-host",❓ Untested,❓ Untested
 "runai_model_streamer_loader",✅ Passing,❓ Untested
 "sampling_params",✅ Passing,❓ Untested
+"Step Pooling (Embedding)",✅ Passing,❓ Untested
 "structured_decoding",✅ Passing,❓ Untested
-"Step Pooling (Embedding)",✅,✅
--- a/support_matrices/release/v6e/default/feature_support_matrix.csv
+++ b/support_matrices/release/v6e/default/feature_support_matrix.csv
@@ -16,3 +16,4 @@ Feature,CorrectnessTest,PerformanceTest
 "runai_model_streamer_loader",✅ Passing,❓ Untested
 "sampling_params",✅ Passing,❓ Untested
 "structured_decoding",✅ Passing,❓ Untested
+"Step Pooling (Embedding)",✅,✅
--- a/support_matrices/release/v6e/default/feature_support_matrix.csv
+++ b/support_matrices/release/v6e/default/feature_support_matrix.csv
@@ -15,5 +15,4 @@ Feature,CorrectnessTest,PerformanceTest
 "multi-host",❓ Untested,❓ Untested
 "runai_model_streamer_loader",✅ Passing,❓ Untested
 "sampling_params",✅ Passing,❓ Untested
-"Step Pooling (Embedding)",✅ Passing,❓ Untested
 "structured_decoding",✅ Passing,❓ Untested
--- a/tests/e2e/test_step_pooling.py
+++ b/tests/e2e/test_step_pooling.py
@@ -1,0 +1,57 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from vllm import LLM
+
+
+def test_step_pooling_e2e():
+    model_name = "Qwen/Qwen3-Embedding-8B"
+
+    # CI-optimized small-scale chunking configuration
+    max_model_len = 2048
+    max_num_batched_tokens = 256
+    max_num_seqs = 1
+
+    try:
+        llm = LLM(model=model_name,
+                  runner="pooling",
+                  max_num_seqs=max_num_seqs,
+                  max_model_len=max_model_len,
+                  max_num_batched_tokens=max_num_batched_tokens,
+                  dtype="bfloat16",
+                  trust_remote_code=True,
+                  load_format="dummy")
+    except Exception as e:
+        pytest.skip(f"Skipping test: {e}")
+
+    # Scaling inputs to exceed max_num_batched_tokens to force chunked prefill
+    # 1024 tokens exceeds 256 max_num_batched_tokens.
+    # We use a repeating token string to force the engine to split the prefill into multiple passes.
+    base_input = "Hello, my name is Alice. "
+
+    # Approx 5 words per repetition, so ~200 repetitions generates ~1000 words/tokens.
+    long_input = base_input * 200
+
+    inputs = [long_input]
+
+    try:
+        results = llm.embed(inputs)
+
+        assert len(results) == 1
+        assert results[0].outputs.embedding is not None
+        assert len(results[0].outputs.embedding) > 0
+
+    except Exception as e:
+        pytest.fail(f"Embedding execution failed during chunked prefill: {e}")
--- a/tests/e2e/test_step_pooling.py
+++ b/tests/e2e/test_step_pooling.py
@@ -13,10 +13,10 @@
 # limitations under the License.
 
 import pytest
-from vllm import LLM
 
 
 def test_step_pooling_e2e():
+    from vllm import LLM
     model_name = "Qwen/Qwen3-Embedding-8B"
 
     # CI-optimized small-scale chunking configuration
--- a/tests/models/jax/utils/test_weight_utils.py
+++ b/tests/models/jax/utils/test_weight_utils.py
@@ -113,3 +113,110 @@ class TestJaxAutoWeightsLoader:
                                    jax_output,
                                    rtol=1e-3,
                                    atol=1e-2)
+
+    def test_weight_prefix_mapping(self):
+        """Test that 'model.' is prepended correctly based on module structure."""
+        from unittest.mock import patch, MagicMock
+        from tpu_inference.models.jax.utils.weight_utils import JaxAutoWeightsLoader
+        
+        class MockModel(JaxModule):
+            def __init__(self):
+                super().__init__()
+                self.model = MagicMock() # Simulate having 'model' attribute
+                
+        model = MockModel()
+        loader = JaxAutoWeightsLoader(model)
+        
+        weights = [
+            ("embed_tokens.weight", torch.zeros((2, 2))),
+            ("lm_head.weight", torch.zeros((2, 2))),
+            ("visual.weight", torch.zeros((2, 2))),
+            ("model.layers.0.weight", torch.zeros((2, 2))),
+        ]
+        
+        with patch('tpu_inference.models.jax.utils.weight_utils.AutoWeightsLoader._load_module') as mock_super_load:
+            mock_super_load.return_value = iter([]) # Yield nothing
+            list(loader._load_module("", model, weights))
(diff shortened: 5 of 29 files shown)


bvrockwell:
Thanks @anthonsu !!! Could we also add to the support matrices please? I believe this would be under "embedding" (new type)

cc @jcyang43 @saltysoup

anthonsu:
> Thanks @anthonsu !!! Could we also add to the support matrices please? I believe this would be under "embedding" (new type)
> 
> cc @jcyang43 @saltysoup

Sure. ~~I'll have the support matrices updated too~~ 
Done. Updated `model`, `kernel` and `feature` matrices tables of both **v7x** and **v6e** with `embedding` related entries.

The dependency vLLM fix PR was merged (https://github.com/vllm-project/vllm/pull/41049), ~~I'll update LKG version to that commit as well~~ 
Done.

@bvrockwell @saltysoup

bvrockwell:
> > Thanks @anthonsu !!! Could we also add to the support matrices please? I believe this would be under "embedding" (new type)
> > cc @jcyang43 @saltysoup
> 
> Sure. ~I'll have the support matrices updated too~ Done. Updated `model`, `kernel` and `feature` matrices tables of both **v7x** and **v6e** with `embedding` related entries.
> 
> The dependency vLLM fix PR was merged ([vllm-project/vllm#41049](https://github.com/vllm-project/vllm/pull/41049)), ~I'll update LKG version to that commit as well~ Done.
> 
> @bvrockwell @saltysoup

Amazing! Thank you for adding a new model class :)

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -540,21 +540,32 @@ def compute_pooler_output(
...
+                    num_scheduled_tokens,
                     torch.tensor(seq_lens),
                     device=torch_states.device,
                 )
                 outputs: list[torch.Tensor] = self._pooler(
                     torch_states,
                     pooling_metadata,
                 )
+
+                # Qwen3-Embedding requires L2 normalization.
```
kyuyeunk: please don't add any model specific logics in `vllm_model_wrapper.py`. this file should be completely model agnostic.

tpu_inference/runner/tpu_runner.py
```diff
@@ -482,12 +482,20 @@ def _init_inputs(self) -> None:
         self.requests: dict[str, CachedRequestState] = {}
         # mm_hash ->  encoder_output
         self.encoder_cache: dict[str, jax.Array] = {}
+
+        tp_size = self.mesh.shape.get("model", 1)
```
kyuyeunk: always use `ShardingAxisName`: https://github.com/anthonsu/tpu-inference/blob/49deddcf685f6df164d69b11b9a2f7cc7c09d919/tpu_inference/runner/compilation_manager.py#L143

tpu_inference/runner/tpu_runner.py
```diff
@@ -482,12 +482,20 @@ def _init_inputs(self) -> None:
...
+        if tp_size == 1 and self.vllm_config.parallel_config is not None:
+            tp_size = self.vllm_config.parallel_config.tensor_parallel_size
```
kyuyeunk, line 503: why is this overriding it?

tpu_inference/models/common/model_loader.py
```diff
@@ -259,16 +262,38 @@ def create_sharded_model():
                     model_weights = vllm_config.model_config.model_weights
                 weights_iterator = loader._get_weights_iterator(
                     model_weights, vllm_config.model_config.revision)
+
+                # Quick Extraction Pass for Pooler weights
+                pooler_weights = {}
+
+                def intercepted_iterator():
+                    for name, weight in weights_iterator:
+                        if name.startswith("pooler."):
```
kyuyeunk: this string based approach looks extremely fragile.

tpu_inference/models/common/interface.py
```diff
@@ -23,6 +23,7 @@ def __call__(
         hidden_states: jax.Array,
         pooling_metadata: PoolingMetadata,
         seq_lens: np.ndarray,
+        num_scheduled_tokens: Optional[np.ndarray] = None,
```
kyuyeunk: please follow https://peps.python.org/pep-0604/ and use `np.ndarray | None = None`. same comment for all the other places.

tpu_inference/runner/persistent_batch_manager.py
```diff
@@ -151,6 +152,8 @@ def update_states(self, scheduler_output: "VllmSchedulerOutput",
                 num_computed_tokens=new_req_data.num_computed_tokens,
                 output_token_ids=[],
                 lora_request=new_req_data.lora_request,
+                pooling_states=PoolingStates()
+                if new_req_data.pooling_params is not None else None,
```
kyuyeunk: please declare the variable outside of the function argument.

tpu_inference/runner/input_batch.py
```diff
@@ -260,7 +269,9 @@ def collect_sampling(sampling_params: SamplingParams) -> None:
 
         if pooling_params := request.pooling_params:
             self.pooling_params[req_id] = pooling_params
-            self.pooling_states[req_id] = PoolingStates()
+            self.pooling_states[
```
kyuyeunk: i'm seeing similar pattern of checking if `request.pooling_states` is None or not and perform something. doesn't look like a good pattern

tpu_inference/models/vllm/vllm_model_wrapper.py
anthonsu: Done. Removed the model-specific comments regarding Qwen3-Embedding and L2 normalization to ensure vllm_model_wrapper.py remains strictly model-agnostic.

tpu_inference/runner/tpu_runner.py
anthonsu: Fixed. Replaced hardcoded "model" strings with ShardingAxisName.MODEL in both tpu_runner.py and compilation_manager.py

tpu_inference/models/common/model_loader.py
anthonsu: I have refactored the weight interception logic to use a more robust regex matching mechanism (re.match(r"^(?:model\.)?pooler\.(.*)$", name)).

tpu_inference/runner/persistent_batch_manager.py
anthonsu: Fixed. Extracted the pooling_states initialization logic into a separate variable declared.

tpu_inference/models/common/interface.py
anthonsu: Done.

support_matrices/release/v6e/default/feature_support_matrix.csv
```diff
@@ -15,4 +15,5 @@ Feature,CorrectnessTest,PerformanceTest
 "multi-host",❓ Untested,❓ Untested
 "runai_model_streamer_loader",✅ Passing,❓ Untested
 "sampling_params",✅ Passing,❓ Untested
+"Step Pooling (Embedding)",✅ Passing,❓ Untested
```
jcyang43: Hi @anthonsu, thanks for the PR!  

Please don't update these support matrices under the release directory directly. They get updated based on CI results whenever we roll out an official release. They're supposed to represent the support status of the current latest release at any given moment.

Do you mind following the guide [here](https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/README.md#adding-a-new-feature-to-ci) to add tests for these on the CI so they will be updated automatically in the future? Thanks!

anthonsu: @jcyang43 , thanks for sharing, I am reading your CI link. ~~I will get feature support matrices update through CI~~ Done.

.buildkite/models/Qwen_Qwen3-Embedding-8B.yml
```diff
@@ -0,0 +1,63 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pipeline-name: Qwen/Qwen3-Embedding-8B
+# pipeline-type: Embedding
+steps:
+  - label: "${TPU_VERSION:-tpu6e} Unit tests for Qwen/Qwen3-Embedding-8B"
```
jrplatin, line 18: Do we not want to run a full E2E test for this model?

support_matrices/nightly/v6e/default/model_support_matrix.csv
jrplatin, line 1: I thought this was auto-populated? cc @jcyang43

jrplatin: Same comment with the other CSV changes

tests/e2e/test_step_pooling.py
```diff
@@ -0,0 +1,74 @@
...
+
+import pytest
+from vllm import LLM
+
+
+def test_step_pooling_e2e():
+    model_name = "Qwen/Qwen3-Embedding-8B"
+
+    # CI vs Manual/v7x paths via environment variable
+    is_manual = os.environ.get("TEST_MANUAL", "false").lower() == "true"
```
jrplatin: Why is this needed? Why not just use the same settings for both? I assume they would both fundamentally test the same thing, no?

tpu_inference/runner/tpu_runner.py
```diff
@@ -496,12 +496,18 @@ def _init_inputs(self) -> None:
         self.requests: dict[str, CachedRequestState] = {}
         # mm_hash ->  encoder_output
         self.encoder_cache: dict[str, jax.Array] = {}
+
+        tp_size = self.mesh.shape.get(ShardingAxisName.MODEL, 1)
+
+        self.vocab_size = common_utils.align_to(model_config.get_vocab_size(),
```
jrplatin, line 507: @wenxindongwork Is this the best way to get TP size?

tpu_inference/runner/input_batch.py
```diff
@@ -280,9 +289,8 @@ def collect_sampling(sampling_params: SamplingParams) -> None:
         if sampling_params := request.sampling_params:
             collect_sampling(sampling_params)
 
-        if pooling_params := request.pooling_params:
```
jrplatin, line 283: Why are we omitting the if statement here now?

tpu_inference/runner/compilation_manager.py
```diff
@@ -64,6 +64,13 @@ def _create_dummy_tensor(self,
                              dtype: Any,
                              sharding: Optional[NamedSharding] = None) -> Any:
         """Helper to create dummy tensors for precompilation."""
+        if len(shape) > 1:
+            tp_size = self.runner.mesh.shape.get(ShardingAxisName.MODEL, 1)
+            if tp_size == 1 and self.runner.vllm_config.parallel_config is not None:
```
jrplatin: When would this happen?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -583,21 +583,28 @@ def compute_pooler_output(
             hidden_states: jax.Array,
             pooling_metadata: PoolingMetadata,
             seq_lens: np.ndarray,
+            num_scheduled_tokens: Optional[np.ndarray] = None,
         ) -> PoolerOutput:
             assert self._pooler is not None, "Model does not support pooling"
 
+            if num_scheduled_tokens is None:
+                num_scheduled_tokens = seq_lens
```
jrplatin, line 593: Can you explain this case please?

tests/models/jax/utils/test_weight_utils.py
```diff
@@ -95,10 +95,7 @@ def test_load_from_safetensors(self):
                 model_config.revision = None
 
                 loader = get_model_loader(
-                    LoadConfig(load_format="safetensors",
-                               model_loader_extra_config={
-                                   "enable_weights_track": False
-                               }))
+                    LoadConfig(load_format="safetensors"))
```
jrplatin, line 98: Why is this change needed?

tpu_inference/layers/common/sharding.py
```diff
@@ -52,6 +52,9 @@ class ShardingAxisNameBase:
     MODEL_1 = 'model'
     MODEL_2 = 'expert'
 
+    # Standard name for the model sharding axis.
```
jrplatin, line 55: @wenxindongwork Can you comment on this ?

tpu_inference/models/common/model_loader.py
```diff
@@ -270,6 +274,8 @@ def create_sharded_model():
                 del vllm_config.model_config.runai_model_weights_iterator
             else:
                 model.load_weights(rng)
+            if hasattr(vllm_config, "pytorch_pooler"):
```
jrplatin, line 288: Am I reading this right that this negates what we do on line 255?

tpu_inference/models/common/model_loader.py
```diff
@@ -400,10 +421,44 @@ def combine_hidden_states(graphdef, state, hidden_states):
         get_mrope_input_positions_fn=get_mrope_input_positions_fn,
     )
 
+    if pooler is not None:
+        import torchax
```
jrplatin, line 436: Are these imports here to avoid a circular import?

support_matrices/nightly/v6e/default/model_support_matrix.csv
jcyang43: Yes, the nightly support matrices are auto-populated. Thanks Jacob

tests/e2e/test_step_pooling.py
anthonsu: The dual configuration is designed to balance logic coverage with CI resource limits. Using a smaller max_num_batched_tokens (256, according to the latest buildkite result, we may still need to make the number smaller) in CI still forces the 'Chunked Prefill' logic, but prevents OOM on shared CI nodes. 

I manually verified every time on my v7x machine with the full long input. So I share the manual path in the script for anyone who needs to run full-scale token verification. 

If you believe this dual config is not necessary, I can clean the manual path with no problem.

.buildkite/models/Qwen_Qwen3-Embedding-8B.yml
anthonsu: @jrplatin , did you mean `want` or `not want`?

e2e testing was added `tests/e2e/test_step_pooling.py.` This YAML file is intended to trigger that test during CI to ensure the 'Embedding' status is driven by real execution results.

support_matrices/nightly/v6e/default/model_support_matrix.csv
anthonsu: Got it. I will remove these csv inputs as well.

tpu_inference/runner/tpu_runner.py
anthonsu: This aligns with the standardized axis naming convention requested in earlier comments and uses the device mesh as the single source of truth. Just FYI.

tpu_inference/runner/input_batch.py
anthonsu: This is part of the 'Null Object Pattern' refactor (from Kyuyeun's previous comment) `pooling_states` is now initialized with a default `PoolingStates` object via default_factory, so it is never `None`. This allows us to linearize the logic and avoid fragmented conditional guards

tpu_inference/models/common/model_loader.py
anthonsu: Yes, these are strategically placed to prevent circular dependency issues between the model loader and the JAX/TorchAX interoperability modules during the initial setup

tpu_inference/runner/compilation_manager.py
anthonsu: This case was originally added to handle scenarios where the configuration might specify a parallel size that isn't yet reflected in the active mesh during pre-compilation. However, to address the earlier feedback on using the mesh as the single source of truth, I have refactored this to rely strictly on `ShardingAxisName.MODEL.`

This condition check is not necessary now. Let me clean it.

tpu_inference/models/vllm/vllm_model_wrapper.py
anthonsu: This logic ensures that for pooling-only models where chunking is not required (or for the initial full prefill), the scheduled tokens correctly match the full sequence length. It guarantees that the pooler receives the complete set of hidden states for the entire prompt in a single step when appropriate.

tests/models/jax/utils/test_weight_utils.py
anthonsu: This was a necessary adjustment for the unit tests following the refactoring of JaxAutoWeightsLoader. Since the loader now dynamically prepends `model.` to keys, disabling the explicit weights tracking in this specific test prevents metadata conflicts during the weight mapping process.

tpu_inference/layers/common/sharding.py
anthonsu: Just following up on this: I added this attribute to the `ShardingAxisName` base class to satisfy the requirement of avoiding hardcoded strings across the runner and manager. It allows us to use `ShardingAxisName.MODEL` consistently.

tpu_inference/models/common/model_loader.py
anthonsu: Not exactly. Line 255 handles the standard weight loading for the backbone, while this conditional block ensures that the specialized pooler weights are intercepted and mapped correctly into the hybrid path. It's an additive step to ensure the PyTorch pooler receives its parameters without interfering with the JAX backbone weights.

tpu_inference/models/common/model_loader.py
```diff
@@ -111,6 +112,7 @@ def _get_nnx_model(
     vllm_config: VllmConfig,
     rng: jax.Array,
     mesh: Mesh,
+    pooler: Any | None = None,
```
jrplatin: What is a pooler? Might be worthwhile to add docstrings here

tpu_inference/models/common/model_loader.py
```diff
@@ -400,10 +421,44 @@ def combine_hidden_states(graphdef, state, hidden_states):
...
+        from vllm.v1.pool.metadata import PoolingMetadata
+
+        def compute_pooler_output(
+            hidden_states: jax.Array,
+            pooling_metadata: PoolingMetadata,
+            seq_lens: np.ndarray,
+            num_scheduled_tokens: np.ndarray | None = None,
+        ):
+            # Performance optimization: use torch_view and move to CPU non-blocking
+            torch_states = torch_view(hidden_states)
```
jrplatin, line 447: Why is this a performance optimization? It seems weird to me to have Torch/TorchAX specific code in the `common` path? Wonder if @kyuyeunk has comments

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -823,6 +824,22 @@ def load_nnx_param_from_reshaped_torch(
             f"Failed to convert torch weight for '{param_name}' ({torch_weight.shape}) to JAX array, with reshape_dims={reshape_dims} and permute_dims={permute_dims}"
         ) from e
 
+    if jax_weight.shape != jax_param.value.shape:
```
jrplatin, line 825: This shouldn't be the default IMO -- I think the only desirable case is for Embedding models? Why is this needed in the first place?

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -891,7 +913,31 @@ def _add_loadable_non_param_tensors(self, module: JaxModule,
 
     def _load_module(self, base_prefix: str, module: JaxModule,
                      weights: Iterable) -> Iterable:
-        yield from super()._load_module(base_prefix, module, weights)
+
+        def _map_weights(w_iter):
```
jrplatin, line 948: Why is this needed?

tpu_inference/models/jax/qwen3.py
```diff
@@ -330,13 +333,16 @@ def __init__(self, vllm_config: VllmConfig, rng_key: jax.Array,
             prefix="model",
         )
         model_config = vllm_config.model_config
-        if not model_config.hf_config.tie_word_embeddings:
+        is_pooling = vllm_config.model_config.runner_type == "pooling"
+        if not model_config.hf_config.tie_word_embeddings and not is_pooling:
             if self.model.is_last_rank:
                 vocab_size = model_config.get_vocab_size()
+                tp_size = vllm_config.parallel_config.tensor_parallel_size if vllm_config.parallel_config is not None else 1
```
jrplatin, line 340: I see this tp_size being calculated in a bunch of places -- wonder if it makes sense to pull out to a common util

tests/e2e/test_step_pooling.py
jrplatin: I think it's probably unnecesary!

anthonsu: I cleaned the manual version and reduced chunk size to trigger step pool with smaller input loads. Let us see how the new test goes.

tpu_inference/models/common/model_loader.py
anthonsu: `pooler` refers to the pooler instance who handles embedding logics during the hybrid execution path.

Sounds good, I will add a docstring here.

tpu_inference/models/common/model_loader.py
anthonsu: This uses `torch_view` from torchax to provide a zero-copy bridge for hidden states. It's a performance win because it avoids costly data duplication when transferring JAX arrays from TPU to the CPU-based pooler. I've ensured it's isolated to the hybrid execution logic.

tpu_inference/models/jax/utils/weight_utils.py
anthonsu: This is specifically to handle Vocab Padding for hardware alignment on v7x/v6e. It prevents IndivisibleError when a checkpoint's vocab size doesn't match the TP-aligned requirements of the TPU model (that was highlighted in the description).

tpu_inference/models/jax/utils/weight_utils.py
anthonsu: This override is necessary to inject the _map_weights logic into the loading stream. It allows us to dynamically prepend the model. prefix to backbone weight keys to match the flax_nnx attribute structure, while selectively protecting root-level parameters like visual or pooler.

tpu_inference/models/jax/utils/weight_utils.py
anthonsu: I agree that global automatic padding can be risky. The current implementation includes a safety check to ensure we only pad when the checkpoint is a strict subset of the expected shape.

I can further restrict this logic to only trigger for vocab-related layers or when Embedding models are explicitly being served, preventing any unintended side effects on standard generative models. 

I am going to make this quick change. Let me know what do you think? @jrplatin

tpu_inference/models/jax/qwen3.py
anthonsu: Introduced new common util helper.

tpu_inference/models/common/model_loader.py
anthonsu: docstring added.

tpu_inference/models/jax/utils/weight_utils.py
anthonsu: Done.

anthonsu:
@kyuyeunk , @jrplatin , @jcyang43 , @bvrockwell , all Buildkite tests passed, the CI/CD build is green.

~~I am going to rebase two latest fixes in the main branch and kick off another build to double verify~~ Done.

- https://github.com/vllm-project/tpu-inference/pull/2488
- https://github.com/vllm-project/tpu-inference/pull/2487

Rebased done, all Buildkite reports are green.

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -821,6 +822,42 @@ def load_nnx_param_from_reshaped_torch(
...
         ) from e
 
+    if jax_weight.shape != jax_param.value.shape:
+        # Retrieve current config to determine if the model task is embedding/pooling
+        is_embedding_task = False
+        try:
+            is_embedding_task = (get_current_vllm_config().model_config.
+                                 runner_type == "pooling")
+        except Exception:
+            pass
```
jrplatin: If there is an exception, I imagine we should log and return so no change occurs, no?

jrplatin (approved):

tpu_inference/models/jax/utils/weight_utils.py
anthonsu: It makes sense to add another log line here to indicate if any vllm config retrieve error happens and `is_embedding_task` is guaranteed to be `false` in such case. Done with the change.

PR #2422, opened 2026-04-27
Add support for mistral 3 large 

karan:
# Description

Adds support for mistral 3 large, tested on v7x-16.

# Tests

Full model (takes 90+ mins to start up):

```bash
VLLM_DISABLE_SHARED_EXPERTS_STREAM=0 \
MOE_REQUANTIZE_BLOCK_SIZE=512 \
NEW_MODEL_DESIGN=1 \
MODEL_IMPL_TYPE=vllm \
HF_HOME="/mnt/disks/persist/hf" \
TPU_BACKEND_TYPE=jax \
vllm serve \
    --model mistralai/Mistral-Large-3-675B-Instruct-2512 \
    --download-dir /mnt/disks/persist/hf/hub \
    --max-model-len 5120 \
    --max-num-seqs 16 \
    --no-enable-prefix-caching \
    --tensor-parallel-size 16 \
    --kv-cache-dtype "fp8" \
    --no-async-scheduling \
    --gpu-memory-utilization 0.90 \
    --enable-expert-parallel \
    --additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' \
    --trust-remote-code \
    --tokenizer-mode mistral \
    --config_format mistral \
    --load_format mistral \
    --port 8000
```

Text-to-text:

```bash
$ curl http://localhost:8000/v1/chat/completions \
    -d '{
      "model": "mistralai/Mistral-Large-3-675B-Instruct-2512",
      "messages": [
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "What are the top 5 most popular programming languages? Be brief."}
      ],
      "temperature": 0.7
}'
{"id":"chatcmpl-a0adc67a720a1011","object":"chat.completion","created":1777305131,"model":"mistralai/Mistral-Large-3-675B-Instruct-2512","choices":[{"index":0,"message":{"role":"assistant","content":"Here are the top 5 most popular programming languages as of recent rankings:\n\n1. **Python**\n2. **JavaScript**\n3. **Java**\n4. **C#**\n5. **C/C++**\n\n(Source: TIOBE Index, Stack Overflow Developer Survey, and GitHub's Octoverse)","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":25,"total_tokens":91,"completion_tokens":66,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}r
```

Image-to-text:

```bash
$ curl http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "mistralai/Mistral-Large-3-675B-Instruct-2512",
    "messages": [
      {
        "role": "user",
        "content": [
          { "type": "text", "text": "What is in this image?" },
          {
            "type": "image_url",
            "image_url": { "url": "https://a-z-animals.com/media/swan-3.jpg" }
          }
        ]
      }
    ],
    "max_tokens": 50
  }'
{"id":"chatcmpl-8ce6a7fd895be7f4","object":"chat.completion","created":1777305287,"model":"mistralai/Mistral-Large-3-675B-Instruct-2512","choices":[{"index":0,"message":{"role":"assistant","content":"The image shows a swan, more specifically a mute swan (Cygnus olor), swimming on a body of water. Mute swans are known for their elegant, all-white plumage, orange beak with a black base,","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null},"logprobs":null,"finish_reason":"length","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":897,"total_tokens":947,"completion_tokens":50,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null}
```

vLLM benchmark:

```bash
$ vllm bench serve --model mistralai/Mistral-Large-3-675B-Instruct-2512 --dataset-name random --random-input-len 1024 --random-output-len 1024 --num-prompts 320 --ignore-eos --skip-chat-template

Successful requests:                     320
Benchmark duration (s):                  115.43
Total input tokens:                      327680
Total generated tokens:                  327680
Request throughput (req/s):              2.77
Output token throughput (tok/s):         2838.89
Peak output token throughput (tok/s):    5054.00
Peak concurrent requests:                320.00
Total token throughput (tok/s):          5677.77
Mean TTFT (ms):                          20404.53
Median TTFT (ms):                        8991.31
P99 TTFT (ms):                           70148.74
Mean TPOT (ms):                          56.90
Median TPOT (ms):                        59.80
P99 TPOT (ms):                           61.81
Mean ITL (ms):                           56.90
Median ITL (ms):                         52.38
P99 ITL (ms):                            55.80
```

Correctness check

```bash
lm_eval --model local-completions \
	--model_args model=mistralai/Mistral-Large-3-675B-Instruct-2512,base_url=http://localhost:8000/v1/completions,num_concurrent=16,max_length=5120 \
	--tasks gsm8k  --apply_chat_template --num_fewshot 8 \
	--limit 100
```

Result

```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     8|exact_match|↑  | 0.93|±  |0.0256|
|     |       |strict-match    |     8|exact_match|↑  | 0.10|±  |0.0302|
```

There's no official results that I can find for this specific benchmark and model, but anything >70% is considered great. Low `strict-match` score is expected for instruct models, but the very high `flexible-extract` is a good sign.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/process_weights/linear_weights.py
+++ b/tpu_inference/layers/common/process_weights/linear_weights.py
@@ -114,6 +114,8 @@ def process_linear_weights(
     else:
 
         def slice_tensor(tensor):
+            if tensor is None:
+                return None
             tensors = []
             start = 0
             for size in output_sizes:
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -533,6 +533,12 @@ def process_moe_weights(
                 "process_moe_weights is not yet implemented for megablox gmm "
                 "backend")
 
+    # Covert scales to jax arrays (they may be torch.Tensors)
+    if w13_weight_scale is not None:
+        w13_weight_scale = jnp.array(w13_weight_scale)
+    if w2_weight_scale is not None:
+        w2_weight_scale = jnp.array(w2_weight_scale)
+
     return FusedMoEWeights(
         w13_weight=w13_weight,
         w13_weight_scale=w13_weight_scale,
@@ -674,10 +680,10 @@ def process_fp8_moe_weights(
         weights = FusedMoEWeights(
             w13_weight=w13_weight,
             w13_weight_scale=w13_weight_scale,
-            w13_bias=None,
+            w13_bias=weights.w13_bias,
             w2_weight=w2_weight,
             w2_weight_scale=w2_weight_scale,
-            w2_bias=None,
+            w2_bias=weights.w2_bias,
         )
 
     else:
@@ -715,10 +721,10 @@ def process_fp8_moe_weights(
             FusedMoEWeights(
                 w13_weight=w13_weight,
                 w13_weight_scale=None,
-                w13_bias=None,
+                w13_bias=weights.w13_bias,
                 w2_weight=w2_weight,
                 w2_weight_scale=None,
-                w2_bias=None,
+                w2_bias=weights.w2_bias,
             ),
             desired_quant_dtype,
             requant_block_size,
--- a/tpu_inference/layers/vllm/custom_ops/mla_attention.py
+++ b/tpu_inference/layers/vllm/custom_ops/mla_attention.py
@@ -92,10 +92,25 @@ class VllmMLAAttention(MLAAttention):
 
             # NOTE: vLLM dequantizes kv_b_proj weights which causes more memory
             # usage than expected.
-
             # quantize W_UK_T, W_UV back to cache type and transfer
             # `W_UK_T`, `W_UV` to TPUs
-            mesh = self.kv_b_proj.quant_method.linear_config.mesh
+            # Try to get the mesh from the first possible path.
+            quant_method = getattr(self.kv_b_proj, 'quant_method', None)
+            linear_config = getattr(quant_method, 'linear_config', None)
+            mesh = getattr(linear_config, 'mesh', None)
+            if mesh is None:
+                # If the first path failed, mesh will be None. Try the second path.
+                scheme = getattr(self.kv_b_proj, 'scheme', None)
+                linear_config = getattr(scheme, 'linear_config', None)
+                mesh = getattr(linear_config, 'mesh', None)
+
+            if mesh is None:
+                # If mesh is still None after trying all paths, raise an error.
+                raise ValueError(
+                    "Could not find JAX Mesh. Failed to access "
+                    "'.quant_method.linear_config.mesh' or "
+                    "'.scheme.linear_config.mesh' on the kv_b_proj layer.")
+
             sharding = NamedSharding(mesh, P(ShardingAxisName.ATTN_HEAD, ))
             self.W_UK_T, self.W_UK_T_scale = quantize_tensor(
                 self.kv_cache_quantized_dtype, jax_view(self.W_UK_T), axis=1)
--- a/tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
+++ b/tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
@@ -39,6 +39,9 @@ def scaled_dot_product_attention(
     if enable_gqa is not False:
         raise NotImplementedError("patched_sdpa does not support enable_gqa")
 
+    if scale is None:
+        scale = 1.0
+
     mesh = jax.sharding.get_abstract_mesh()
 
     # Q, K, V shapes: (batch, num_heads, seq_len, head_dim)
@@ -105,6 +108,9 @@ def vllm_vit_sdpa(
     query = jnp.swapaxes(query, 1, 2)
     key = jnp.swapaxes(key, 1, 2)
     value = jnp.swapaxes(value, 1, 2)
+
+    if scale is None:
+        scale = 1.0
 
     mesh = jax.sharding.get_abstract_mesh()
 
--- a/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
+++ b/tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
@@ -12,26 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import jax
 import torch
 from compressed_tensors.quantization import QuantizationArgs
 from jax.sharding import Mesh
 from torch.nn.parameter import Parameter
 from torchax.interop import jax_view, torch_view
 from vllm.model_executor.layers.fused_moe import FusedMoE, FusedMoEConfig
-from vllm.model_executor.layers.fused_moe.activation import MoEActivation
 from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors_moe.compressed_tensors_moe_w8a8_fp8 import \
     CompressedTensorsW8A8Fp8MoEMethod
 
 from tpu_inference.layers.common.moe import MoEBackend
 from tpu_inference.layers.common.process_weights.moe_weights import (
-    FusedMoEWeights, process_moe_weights, shard_moe_weights)
-from tpu_inference.layers.common.sharding import ShardingAxisName
+    FusedMoEWeights, process_fp8_moe_weights, shard_moe_weights)
 from tpu_inference.layers.vllm.interface.moe import (
     select_moe_backend_from_fused_moe_config, vllm_moe_apply)
 from tpu_inference.layers.vllm.quantization.configs import VllmQuantConfig
 from tpu_inference.logger import init_logger
-from tpu_inference.utils import get_mesh_shape_product, t2j
+from tpu_inference.utils import t2j
 
 logger = init_logger(__name__)
 
@@ -94,40 +91,26 @@ class VllmCompressedTensorsW8A8Fp8MoEMethod(CompressedTensorsW8A8Fp8MoEMethod,
         else:
             w13_bias = w2_bias = None
 
-        @jax.jit
-        def process_fp8_moe_weights(
-            w13_weight: jax.Array,
-            w13_weight_scale: jax.Array,
-            w13_bias: jax.Array | None,
-            w2_weight: jax.Array,
-            w2_weight_scale: jax.Array,
-            w2_bias: jax.Array | None,
-        ) -> FusedMoEWeights:
-            w13_interleave = layer.activation == MoEActivation.SWIGLUOAI
-            w13_reorder_size = get_mesh_shape_product(
-                self.mesh, ShardingAxisName.MLP_TENSOR)
+        # Support for blockwise quantization (e.g. 128x128 blocks) used by
+        # DeepSeek V3 / Mistral 3 Large models.
+        weight_block_size = None
+        if self.weight_quant.block_structure is not None:
+            weight_block_size = tuple(self.weight_quant.block_structure)
 
-            return process_moe_weights(
-                weights=FusedMoEWeights(
-                    w13_weight=w13_weight,
-                    w13_weight_scale=w13_weight_scale,
-                    w13_bias=w13_bias,
-                    w2_weight=w2_weight,
-                    w2_weight_scale=w2_weight_scale,
-                    w2_bias=w2_bias,
-                ),
-                moe_backend=self.moe_backend,
-                w13_reorder_size=w13_reorder_size,
-                w13_interleave=w13_interleave,
-            )
-
+        input_weights = FusedMoEWeights(
+            w13_weight=w13_weight,
+            w13_weight_scale=w13_weight_scale,
+            w13_bias=w13_bias,
+            w2_weight=w2_weight,
+            w2_weight_scale=w2_weight_scale,
+            w2_bias=w2_bias,
+        )
         weights = process_fp8_moe_weights(
-            w13_weight,
-            w13_weight_scale,
-            w13_bias,
-            w2_weight,
-            w2_weight_scale,
-            w2_bias,
+            input_weights,
+            moe_backend=self.moe_backend,
+            mesh=self.mesh,
+            activation=layer.activation.value,
+            weight_block_size=weight_block_size,
         )
         weights = torch_view(
             shard_moe_weights(weights, self.moe_backend, self.mesh))
--- a/tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
+++ b/tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
@@ -22,13 +22,8 @@ from compressed_tensors.quantization import (QuantizationArgs,
 from jax.sharding import NamedSharding, PartitionSpec
 from torch.nn.parameter import Parameter
 from torchax.interop import jax_view, torch_view
-from vllm.model_executor.kernels.linear import (FP8ScaledMMLinearKernel,
-                                                register_linear_kernel)
-from vllm.model_executor.kernels.linear.scaled_mm import \
-    FP8ScaledMMLinearLayerConfig
 from vllm.model_executor.layers.quantization.compressed_tensors.schemes.compressed_tensors_w8a8_fp8 import \
     CompressedTensorsW8A8Fp8
-from vllm.platforms import PlatformEnum
 
 from tpu_inference.layers.common.linear import sharded_quantized_matmul
 from tpu_inference.layers.common.process_weights.linear_weights import (
@@ -36,6 +31,8 @@ from tpu_inference.layers.common.process_weights.linear_weights import (
     to_parameter_list)
 from tpu_inference.layers.common.quantization import (dequantize_tensor,
                                                       quantize_tensor)
+from tpu_inference.layers.common.quantization.fp8 import \
+    process_blockwise_fp8_linear_weights
 from tpu_inference.layers.common.utils import \
     slice_sharded_tensor_for_concatenation
 from tpu_inference.layers.vllm.quantization.configs import \
@@ -48,37 +45,6 @@ P = PartitionSpec
 logger = init_logger(__name__)
 
 
-class TpuFP8ScaledMMLinearKernel(FP8ScaledMMLinearKernel):
-    """Stub FP8 kernel registered for TPU platform.
-
-    VllmCompressedTensorsW8A8Fp8 fully overrides apply_weights and
-    process_weights_after_loading, so apply_scaled_mm is never called.
-    This class exists solely so that init_fp8_linear_kernel (called by the
-    upstream create_weights) can select a kernel without raising a KeyError
-    for the TPU platform.
-    """
-
-    @classmethod
-    def is_supported(
-            cls,
-            compute_capability: int | None = None) -> tuple[bool, str | None]:
-        return True, None
-
-    @classmethod
-    def can_implement(
-            cls, c: FP8ScaledMMLinearLayerConfig) -> tuple[bool, str | None]:
-        return True, None
-
-    def apply_scaled_mm(self, *, A, B, out_dtype, As, Bs, bias,
-                        output_shape) -> torch.Tensor:
-        raise NotImplementedError(
-            "TpuFP8ScaledMMLinearKernel.apply_scaled_mm should never be called"
-        )
-
-
-register_linear_kernel(TpuFP8ScaledMMLinearKernel, PlatformEnum.TPU, "fp8")
-
-
 class VllmCompressedTensorsW8A8Fp8(CompressedTensorsW8A8Fp8):
 
     def __init__(
@@ -87,12 +53,43 @@ class VllmCompressedTensorsW8A8Fp8(CompressedTensorsW8A8Fp8):
         is_static_input_scheme: bool,
         linear_config: VllmQuantLinearConfig,
     ):
+        # Per https://github.com/vllm-project/vllm/pull/32929,
+        # init_fp8_linear_kernel is now called by super().__init__
+        # but does not support TPU backends as expected.
+        # use_marlin was also changed to be determined via isinstance(self.fp8_linear, MarlinFP8ScaledMMLinearKernel).
+        # We need to monkeypatch init_fp8_linear_kernel and explicitly set use_marlin = True
+        # in order to bypass using native vLLM's vllm/vllm/model_executor/layers/quantization/utils/quant_utils.py:scaled_quantize.
+        from vllm.model_executor.layers.quantization.compressed_tensors.schemes import \
+            compressed_tensors_w8a8_fp8 as vllm_ct_fp8
+        vllm_ct_fp8.init_fp8_linear_kernel = lambda *args, **kwargs: None
+
         CompressedTensorsW8A8Fp8.__init__(
             self,
             weight_quant=weight_quant,
             is_static_input_scheme=is_static_input_scheme)
+        self.use_marlin = True
 
         self.linear_config = linear_config
+
+    def create_weights(
+        self,
+        layer: torch.nn.Module,
+        input_size_per_partition: int,
+        output_partition_sizes: list[int],
+        input_size: int,
+        output_size: int,
+        params_dtype: torch.dtype,
+        **extra_weight_attrs,
+    ):
+        # Per https://github.com/vllm-project/vllm/pull/33892, use_marlin is set again
+        # in vllm/model_executor/layers/quantization/fp8.py `create_weights`.
+        # The flag is set on whether a specific type of GPU kernel is being used which
+        # means it is set to False for TPU.
+        # We need to return it back to True here.
+        super().create_weights(layer, input_size_per_partition,
+                               output_partition_sizes, input_size, output_size,
+                               params_dtype, **extra_weight_attrs)
+        self.use_marlin = True
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
         weight = t2j(layer.weight, use_dlpack=False)
@@ -132,7 +129,14 @@ class VllmCompressedTensorsW8A8Fp8(CompressedTensorsW8A8Fp8):
                 weight, weight_scale = quantize_tensor(jnp.float8_e4m3fn,
                                                        weight, None)
             else:
-                weight_scale = jnp.squeeze(weight_scale, -1)
+                # Handle 2D blockwise scales correctly for quantized_matmul_kernel
+                # which expects 3D dims for blockwise ops.
+                # In standard per-channel quantization, the weight_scale is
+                # typically 2D with a singleton dimension which can be safely
+                # squeezed to 1D. However, in blockwise quantization, the
+                # scale is truly 2D.
+                if weight_scale.ndim == 2:
+                    weight_scale = jnp.squeeze(weight_scale)
 
             return process_linear_weights(
                 LinearWeights(
@@ -147,7 +151,37 @@ class VllmCompressedTensorsW8A8Fp8(CompressedTensorsW8A8Fp8):
                 per_tensor=per_tensor,
             )
 
-        weights = process_fp8_linear_weights(weight, weight_scale, bias)
+        if self.weight_block_size is not None:
+            weights = process_blockwise_fp8_linear_weights(
+                weight=weight,
+                weight_scale=weight_scale,
+                bias=bias,
+                weight_block_size=tuple(self.weight_block_size),
+                requant_block_size=self.linear_config.requant_block_size,
+                output_sizes=tuple(self.linear_config.output_sizes),
+                requant_weight_dtype=self.linear_config.requant_weight_dtype,
+                fuse_matmuls=self.linear_config.fuse_matmuls,
+                n_shards=self.linear_config.n_shards)
+        else:
+            weights = process_fp8_linear_weights(weight, weight_scale, bias)
+
+        if self.linear_config.enable_quantized_matmul_kernel and weights.weight_scale.ndim == 2:
+            if weights.weight_scale.shape[
+                    0] > 1 and weights.weight_scale.shape[1] > 1:
+                # The quantized_matmul_kernel expects weight scales shaped (n_blocks, 1, n_out_features) for blockwisze quantization.
+                if self.weight_block_size is not None:
+                    # process_blockwise_fp8_linear_weights returns weight_scale shaped (n_out_features, n_blocks)
+                    # We need it shaped (n_blocks, 1, n_out_features)
+                    weights.weight_scale = jnp.expand_dims(
+                        jnp.transpose(weights.weight_scale),
+                        axis=1,
+                    )
+                else:
+                    weights.weight_scale = jnp.expand_dims(
+                        jnp.transpose(weights.weight_scale),
+                        axis=1,
+                    )
+
         weights = torch_view(
             shard_linear_weights(
                 weights,
--- a/tpu_inference/models/vllm/experimental/model_patcher.py
+++ b/tpu_inference/models/vllm/experimental/model_patcher.py
@@ -28,6 +28,7 @@ import jax
 import torchax
 from torchax.interop import JittableModule, torch_view
 
+from tpu_inference import envs
 from tpu_inference.logger import init_logger
 
 if TYPE_CHECKING:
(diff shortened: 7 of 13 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -1180,7 +1170,7 @@ def gmm_v2(
             group_offset = group_offset[None]
 
     if vmem_limit_bytes is None:
-        vmem_limit_bytes = int(pltpu.get_tpu_info().vmem_capacity_bytes * 0.9)
+        vmem_limit_bytes = int(pltpu.get_tpu_info().vmem_capacity_bytes * 0.85)
```
karan: I was running into vmem ooms, so I opted for a bit more headroom for gmm here

tpu_inference/platforms/tpu_platform.py
```diff
@@ -219,6 +219,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         if compilation_config.backend == "":
             compilation_config.backend = "openxla"
 
+        # Dump all keys from text_config to hf_config if they don't exist
```
karan: This is needed because Mistral/Pixtral don't follow config's convention - https://huggingface.co/mistralai/Mistral-Large-3-675B-Instruct-2512/blob/main/params.json

This was a shortcut I took and I'm open to feedback here to improve readability here. One option I thought of was to do this within the generic_patcher (check if the model is mistral).

karan (commented):
Starting a review to collect feedback (do note that it takes 90min for the server to start up so iterations will be slower).

tpu_inference/platforms/tpu_platform.py
gpolovets1: This recent PR may apply to you as well: https://github.com/vllm-project/tpu-inference/pull/2417

karan: Thanks for the link - yes it seems like it would make my hack here moot.

karan:
Verified correctness after rebasing to main.

```
local-completions ({'model': 'mistralai/Mistral-Large-3-675B-Instruct-2512', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 16, 'max_length': 5120}), gen_kwargs: ({}), limit: 100.0, num_fewshot: 8, batch_size: 1
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     8|exact_match|↑  | 0.94|±  |0.0239|
|     |       |strict-match    |     8|exact_match|↑  | 0.09|±  |0.0288|
```

tpu_inference/kernels/megablox/gmm_v2.py
karan: With 0.9 I get vmem oom.

```bash
(EngineCore pid=2088)   File "/workspace/tpu_inference/tpu_inference/runner/compilation_manager.py", line 260, in _precompile_backbone_helper
(EngineCore pid=2088)     self._run_compilation(
(EngineCore pid=2088)   File "/workspace/tpu_inference/tpu_inference/runner/compilation_manager.py", line 87, in _run_compilation
(EngineCore pid=2088)     result = fn(*args, **call_kwargs)
(EngineCore pid=2088)              ^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=2088)   File "/workspace/tpu_inference/tpu_inference/runner/compilation_manager.py", line 249, in model_fn_wrapper
(EngineCore pid=2088)     kv_caches, hidden_states, _ = self.runner.model_fn(
(EngineCore pid=2088)                                   ^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=2088) jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: E1001: CompileTimeScopedVmemOom:
(EngineCore pid=2088) Ran out of memory in memory space vmem while allocating on stack for %gmm_v2-g_8-m_256-k_7168-act_silu-n_8192-tm_128-tk_7168-tn_1408.1 = bf16[256,4096]{1,0:T(8,128)(2,1)S(1)} custom-call(%convert_reduce_fusion.57, %bitcast.102, %slice.2949, %param.1388, %param.1389), custom_call_target="tpu_custom_call", operand_layout_constraints={s32[128]{0}, s32[1]{0}, bf16[256,7168]{1,0}, f8e4m3fn[8,7168,8192]{2,1,0}, f32[8,14,1,8192]{3,2,1,0}}, frontend_attributes={kernel_metadata={
(EngineCore pid=2088) "acc_dtype":"bfloat16",
(EngineCore pid=2088) "dims.size_group":8,
(EngineCore pid=2088) "dims.size_k":7168,
(EngineCore pid=2088) "dims.size_lhs_group":128,
(EngineCore pid=2088) "dims.size_lhs_sublane":16,
(EngineCore pid=2088) "dims.size_m":256,
(EngineCore pid=2088) "dims.size_n":8192,
(EngineCore pid=2088) "fuse_act":"silu",
(EngineCore pid=2088) "lhs_cfgs.dtype":"bfloat16",
(EngineCore pid=2088) "lhs_cfgs.has_bias":false,
(EngineCore pid=2088) "lhs_cfgs.has_scale":false,
(EngineCore pid=2088) "lhs_cfgs.quant_block_size":512,
(EngineCore pid=2088) "lhs_cfgs.quant_dtype":"float8_e4m3fn",
(EngineCore pid=2088) "out_dtype":"bfloat16",
(EngineCore pid=2088) "rhs_cfgs.dtype":"float8_e4m3fn",
(EngineCore pid=2088) "rhs_cfgs.has_bias":false,
(EngineCore pid=2088) "rhs_cfgs.has_scale":true,
(EngineCore pid=2088) "rhs_cfgs.quant_block_size":512,
(EngineCore pid=2088) "rhs_cfgs.quant_dtype":"float8_e4m3fn",
(EngineCore pid=2088) "tiles.tile_k":7168,
(EngineCore pid=2088) "tiles.tile_m":128,
(EngineCore pid=2088) "tiles.tile_n":1408,
(EngineCore pid=2088) "zero_init":false
(EngineCore pid=2088) }}, metadata={op_name="jit(step_fun)/VllmFusedMoE/jit(fused_moe_func)/shard_map/jit(gmm_v2)/gmm_v2-g_8-m_256-k_7168-act_silu-n_8192-tm_128-tk_7168-tn_1408/pallas_call" stack_frame_id=576}. Scoped allocation with size 62.28M and limit 57.60M exceeded scoped vmem limit by 4.68M. It should not be possible to run out of scoped vmem -
(EngineCore pid=2088) See https://openxla.org/xla/errors/error_1001 for more details.
```

gpolovets1: Hmm I think @kyuyeunk wrote the tile size heuristics.

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
```diff
@@ -131,8 +131,8 @@ def quantized_matmul_kernel(
     # n_lane_multiplier > 1 could improve perf by reducing loop overhead and increasing instruction-level parallelism,
     # allowing the compiler to overlap output fusion and packing overhead with MXU computation
     # TODO(amandaliang): use pltpu.get_tpu_info().mxu_column_size when JAX version is newer
-    compute_tile_n = MXU_SIZE * n_lane_multiplier
-    steps_n = max(1, out_block_size // compute_tile_n)
+    compute_tile_n = min(out_block_size, MXU_SIZE * n_lane_multiplier)
```
gpolovets1: Is this still needed? You were probably triggering this because you were using the checkpoint's [3D scales](https://github.com/clee1994/tpu-inference/blob/pipeline_psum/tpu_inference/layers/common/linear.py#L73) before requantizing.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -167,12 +167,25 @@ def process_w13_for_gmm(tensor,
     """helper to split, pad, concatenate, and reorder w13 tensors."""
 
     # 1. Split into W1 and W3
-    w1 = tensor[..., :config.intermediate_size]
-    w3 = tensor[..., config.intermediate_size:]
+    # Dynamically find the split point. If it's a blocked scale, its last dimension
```
gpolovets1: Was this written again before you added requantization into 2D scale? In this case the scale should not be blocked on the intermediate (sublane) dimension.

tpu_inference/layers/common/sharding.py
```diff
@@ -170,6 +170,9 @@ def from_vllm_config(cls,
 
         if enable_dp_attention:
             # Replicate attention layer when num_kv_heads < TP
+            # Ensure ShardingAxisName is initialized before modifying _cls
+            _ = ShardingAxisName.ATTN_HEAD
+            ShardingAxisName._cls.ATTN_HEAD = None
```
gpolovets1: Why is this needed?

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -41,6 +41,10 @@ def scaled_dot_product_attention(
 
     mesh = jax.sharding.get_abstract_mesh()
 
+    if scale is None:
```
gpolovets1, line 42: I'm not sure if this should be changed.. IIUC some models have stopped explicitly performing the sqrt because I think it could be implicitly learned.

tpu_inference/layers/vllm/process_weights/cleanup_sharding.py
```diff
@@ -57,14 +56,13 @@ def shard_model_to_tpu(model: torch.nn.Module,
...
+        all_tensors = {**params, **buffers}
 
-        # For other weight tensors, repliate them on all the TPU chips.
-        params, buffers = pytree.tree_map_only(
-            _tensor_is_in_cpu,
-            lambda tensor: _shard_tensor_to_tpu_replicated(tensor, mesh),
-            (params, buffers))
-
-        return {**params, **buffers}
+        return {
```
gpolovets1: Why is this needed?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
```diff
@@ -77,6 +80,13 @@ def apply_scaled_mm(self, *, A, B, out_dtype, As, Bs, bias,
 
 
 register_linear_kernel(TpuFP8ScaledMMLinearKernel, PlatformEnum.TPU, "fp8")
+if hasattr(vllm_linear, "_POSSIBLE_FP8_BLOCK_KERNELS"):
```
gpolovets1: Why is this needed?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
```diff
@@ -132,7 +142,11 @@ def process_fp8_linear_weights(
                 weight, weight_scale = quantize_tensor(jnp.float8_e4m3fn,
                                                        weight, None)
             else:
-                weight_scale = jnp.squeeze(weight_scale, -1)
+                if weight_scale.ndim == 2:
```
gpolovets1, line 138: Does this have the same effect as jnp.squeeze(weight)?

tpu_inference/models/vllm/experimental/generic_patcher.py
gpolovets1, line 1: Implementing tpu_inference/layers/vllm/custom_ops/ is our standard process of overriding default vLLM logic.

tpu_inference/models/vllm/experimental/generic_patcher.py
```diff
@@ -0,0 +1,94 @@
...
+
+def _patch_rms_norm():
+    """Patch vLLM's RMSNorm to avoid mixed-math errors in eager mode."""
+    try:
+        from vllm.model_executor.layers.layernorm import RMSNorm
+
+        if not hasattr(RMSNorm, "_original_forward_native"):
+            RMSNorm._original_forward_native = RMSNorm.forward_native
+
+            def _patched_rms_norm_forward_native(self, x, residual=None):
```
gpolovets1: Hmm I don't think we've had an issue with RMSNorm. What issue did you run into?

tpu_inference/models/vllm/experimental/generic_patcher.py
```diff
@@ -0,0 +1,94 @@
...
+                )
+
+            RMSNorm.forward_native = _patched_rms_norm_forward_native
+            logger.info("Patched vLLM's RMSNorm.forward_native.")
+    except Exception as e:
+        logger.warning(f"Failed to patch vLLM's RMSNorm: {e}")
+
+
+def _patch_default_unquantized_gemm():
+    """Patch vLLM's default_unquantized_gemm to unwrap parameters."""
```
gpolovets1: What do you mean by unwrap parameters? This sounds like something `process_weights_after_loading` should be handling?

tpu_inference/kernels/quantized_matmul/blockwise_kernel.py
karan: Good catch! I've removed this unnecessary code

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Yes, this is also moot now that we requantize.

tpu_inference/layers/common/sharding.py
karan: I added this because modifying `ShardingAxisName._cls.ATTN_HEAD` before it's been accessed elsewhere caused an `AttributeError`. So this just forces initialization of the internal `_cls` state.

I'll test without it and see if it's still necessary

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
karan: Ah, good point. I've reverted this change

tpu_inference/layers/vllm/process_weights/cleanup_sharding.py
karan: I think I did this when I was trying to make this work on a single node. Reverted as this is moot too now

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
karan: To make vLLM's [_POSSIBLE_FP8_BLOCK_KERNELS](https://github.com/vllm-project/vllm/blob/10558f5f4608f825d06018def1317e7d3a96d6fe/vllm/model_executor/kernels/linear/__init__.py#L188) be aware of TPU support for fp8 linear layer. Otherwise it would just fail.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
karan: Yes, good catch! I'll update this

tpu_inference/models/vllm/experimental/generic_patcher.py
karan: Ack - I'll move this to tpu_inference/layers/vllm/custom_ops

tpu_inference/models/vllm/experimental/generic_patcher.py
karan: It was needed because passing a `torch.nn.Parameter` into `RMSNorm.forward_native` in eager mode was causing mixed-math crash

tpu_inference/models/vllm/experimental/generic_patcher.py
karan: Similar to above, but ack on the refactor. I've just removed it for now as I'm running an e2e test; if parameters still need unwrapping, I'll move that logic into `process_weights_after_loading`

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
karan: Seems like this is required for multimodal requests, otherwise they fail with `TypeError: unsupported operand type(s) for *=: 'DynamicJaxprTracer' and 'NoneType'`.

tpu_inference/kernels/megablox/gmm_v2.py
gpolovets1: In general, I wouldn't update the default for gmm since it's used for a bunch of models. You may want to chat with Kyuyeun about either updating the heuristic code or setting an override for Mistral somewhere else.

tpu_inference/layers/common/process_weights/linear_weights.py
```diff
@@ -96,26 +96,41 @@ def process_linear_weights(
     if output_sizes is None:
         output_sizes = [weight.shape[dim]]
 
+    def get_scaled_sizes(tensor):
```
gpolovets1: Do we need this if we are using 1D blockwise quantization?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -190,9 +191,6 @@ def _pad_tensor(t):
     padded_w1 = _pad_tensor(w1)
     padded_w3 = _pad_tensor(w3)
 
-    logger.info(f"{name}_w1 shape after padding: {padded_w1.shape}")
```
gpolovets1, line 248: Did you remove these on purpose?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -402,6 +400,11 @@ def process_moe_weights(
                 "process_moe_weights is not yet implemented for megablox gmm backend"
             )
 
+    if w13_weight_scale is not None:
```
gpolovets1, line 537: Why do we need to create a copy?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -524,12 +527,15 @@ def process_fp8_moe_weights(
                               if requant_block_size_from_env else None)
 
     moe_logging_str = (
-        f"[MoE requantization]: re-quantizing MoE weights to {desired_quant_dtype}"
+        f"[MoE requantization]: re-quantizing MoE weights {w13_weight.shape} to {desired_quant_dtype}"
```
gpolovets1: Did you mean to print w13_weight.shape? We aren't changing the shape here, just the output dtype and scales.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -524,12 +527,15 @@ def process_fp8_moe_weights(
...
     )
     if requant_block_size is not None:
         moe_logging_str += f" with block size {requant_block_size}"
+    if hasattr(w13_weight, "sharding"):
+        moe_logging_str += f" | sharding: {w13_weight.sharding}"
     logger.info(moe_logging_str)
 
+    jax.debug.print("[Execution] Starting dequantization")
```
gpolovets1: Will this result in the same print for every layer? Seems a bit excessive if so.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -543,18 +549,21 @@ def process_fp8_moe_weights(
     w13_interleave = activation == "swigluoai"
     w13_reorder_size = get_mesh_shape_product(mesh,
                                               ShardingAxisName.MLP_TENSOR)
+    jax.debug.print("[Execution] Starting quantization")
```
gpolovets1: Same comment about printing here.
Also can't you use the logger instead of jax.debug.print to print the logging messages?

tpu_inference/layers/vllm/custom_ops/mla_attention.py
```diff
@@ -77,10 +77,15 @@ def process_weights_after_loading(self, act_dtype: torch.dtype):
 
             # NOTE: vLLM dequantizes kv_b_proj weights which causes more memory
             # usage than expected.
+            if hasattr(self.kv_b_proj, "scheme"):
+                mesh = self.kv_b_proj.scheme.linear_config.mesh
```
gpolovets1: Shouldn't the linear projections be associated with a quant_method?

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
gpolovets1: Should you just set it to 1 then?

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -106,6 +111,9 @@ def vllm_vit_sdpa(
     key = jnp.swapaxes(key, 1, 2)
     value = jnp.swapaxes(value, 1, 2)
 
+    if scale is None:
```
gpolovets1, line 112: Set to 1 here as well.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
```diff
@@ -94,40 +92,30 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+            w13_weight=w13_weight,
+            w13_weight_scale=w13_weight_scale,
+            w13_bias=w13_bias,
+            w2_weight=w2_weight,
+            w2_weight_scale=w2_weight_scale,
+            w2_bias=w2_bias,
+        )
+
+        # JIT the global processing function to run fast on TPU
+        _jitted_process = jax.jit(process_fp8_moe_weights,
```
gpolovets1: process_fp8_moe_weights is already decorated with jit in its definition.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
gpolovets1: It may be because you need the same monkeypatches I added to FP8 [here](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/quantization/fp8.py#L111) and [here](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/quantization/fp8.py#L125).

tpu_inference/models/vllm/experimental/model_patcher.py
```diff
@@ -66,6 +66,56 @@ def patch_mm_model(
     Returns:
         The patched model and the updated parameters and buffers.
     """
+    # Monkey patch vLLM's _merge_multimodal_embeddings to support JAX/torchax
+    # broadcasting limitations with boolean masks.
+    try:
+        import vllm.model_executor.models.utils as vllm_utils
```
gpolovets1: Does this affect all multimodal models or just Mistral?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -455,6 +455,20 @@ def embed_input_ids_func(
                     else:
                         torch_mm_embeds = torch_view(mm_embeds)
                     assert is_multimodal is not None
+                    # Replicate is_multimodal to avoid multi-host fetching errors during boolean indexing
```
gpolovets1: Again just wondering if this is for all multimodal models or Mistral-specific.

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Yes - I found it to be helpful for such a large model to know that _something_ was happening, but I can remove it for now.

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Same as above (removed)

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Oh no this was unintentional, restored

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Fair - reverted

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
karan: Ah right! I'll remove the jit here

tpu_inference/models/vllm/experimental/model_patcher.py
karan: It's for all mm models, this specific change is to prevent a JAX tracing deadlock with boolean indexing.

tpu_inference/models/vllm/vllm_model_wrapper.py
karan: This is also for all multimodal models afaict and should affect all multi-host topologies

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
karan: Thanks for linking to that. Stubbing out `init_fp8_linear_kernel` entirely might work too, however the approach in this PR I think is in line with the intent of [the refactor](https://github.com/vllm-project/vllm/pull/33892) that added the per-platform block kernels.

tpu_inference/kernels/megablox/gmm_v2.py
karan: Yes, I agree. One idea George had offline was to use an env var. For now I'm moving to that (defaulting to the status quo of 0.9).

tpu_inference/layers/common/process_weights/linear_weights.py
karan: Right, I don't think we need this anymore

tpu_inference/layers/common/process_weights/moe_weights.py
karan: We do not - removed

tpu_inference/layers/vllm/custom_ops/mla_attention.py
karan: Done

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
karan: Done - though I'm curious why?

EDIT: I actually would like to understand why it should be 1 here and above. My understanding is that

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
karan: Actually, jitting seems to be needed here.

Removing it forced JAX to eagerly execute the layer operations in global memory, allocating and releasing massive intermediate fp32 tensors causing an hbm oom.

karan:
I've tested this PR as of commit [d2848ac](https://github.com/vllm-project/tpu-inference/pull/2422/commits/d2848acaffea71467fa6c1dc2f8fc93824815e12) which reverts the need to change the gmm kernel, as the new tiling logic that resolves the vmem issue without needing to change the mem allocation directly. That logic is being sent out tomorrow for review.

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
gpolovets1: Setting to 1 would be a no-op. Some models do not use sqrt(dim) to scale the attention so 1 would be a more innocuous default.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
gpolovets1: Hmm that's pretty strange - we call `process_fp8_moe_weights` in fp8 quantization implementations and they don't need to wrap it with jit again.

tpu_inference/models/vllm/vllm_model_wrapper.py
gpolovets1: Are you sure these changes don't impact perf or accuracy in the other multimodal or multi host models?

tpu_inference/layers/vllm/custom_ops/layernorm.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from vllm.model_executor.layers.layernorm import RMSNorm
+
+
+@RMSNorm.register_oot
```
gpolovets1: Can you check if this gets triggered for all of the models? IIUC we may need to set a flag to ignore certain custom op implementations otherwise they automatically override the vLLM implementation (and almost all the models are using RMSNorm).

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -402,6 +403,11 @@ def process_moe_weights(
                 "process_moe_weights is not yet implemented for megablox gmm backend"
             )
 
+    if w13_weight_scale is not None:
+        w13_weight_scale = jnp.array(w13_weight_scale)
+    if w2_weight_scale is not None:
```
kyuyeunk, line 539: What is the purpose of this line?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -528,6 +534,8 @@ def process_fp8_moe_weights(
     )
     if requant_block_size is not None:
         moe_logging_str += f" with block size {requant_block_size}"
+    if hasattr(w13_weight, "sharding"):
```
kyuyeunk: Please don't use string based checking

tpu_inference/layers/vllm/custom_ops/__init__.py
```diff
@@ -12,13 +12,13 @@
...
 from tpu_inference.layers.vllm.custom_ops import embedding as embedding
 from tpu_inference.layers.vllm.custom_ops import fused_moe as fused_moe
 from tpu_inference.layers.vllm.custom_ops import \
     gdn_attention_op as gdn_attention_op
+from tpu_inference.layers.vllm.custom_ops import layernorm as layernorm
 from tpu_inference.layers.vllm.custom_ops import linear as linear
 from tpu_inference.layers.vllm.custom_ops import mla_attention as mla_attention
 from tpu_inference.layers.vllm.custom_ops import rope as rope
-
-# Register custom op to vLLM so that vLLM model implementation will instantiante
```
kyuyeunk, line 23: Why delete this comment?

tpu_inference/layers/vllm/custom_ops/mla_attention.py
```diff
@@ -77,10 +77,15 @@ def process_weights_after_loading(self, act_dtype: torch.dtype):
 
             # NOTE: vLLM dequantizes kv_b_proj weights which causes more memory
             # usage than expected.
+            if hasattr(self.kv_b_proj.quant_method, "linear_config"):
```
kyuyeunk: No string based checking

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
kyuyeunk, line 1: What is the purpose of this change?

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
kyuyeunk: Not a good approach. I agree with @gpolovets1 's comment.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
kyuyeunk: Why is this needed?

kyuyeunk (commented):
Haven't got into multimodal

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
karan: Got it - thanks for explaining. Changed to 1.

tpu_inference/layers/vllm/custom_ops/__init__.py
karan: Oh it's just moved at the top of the imports - is that wrong?

I moved it back down for now

tpu_inference/layers/common/process_weights/moe_weights.py
karan: The scales were arriving as `torch.Tensor`s, so I have to convert them to jax arrays here. Added a comment explaining this.

Is there some place more upstream where fixing that would be better?

tpu_inference/layers/common/process_weights/moe_weights.py
karan: Sure - I'll remove this change for now (it was primarily added to help me debug)

tpu_inference/layers/vllm/custom_ops/layernorm.py
karan: Thanks - I did validate and yes it is used for all models. I've put this behind an env var (and validated both with a qwen3 and mistral models - both with and without the env var).

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
karan: Changed to 1

tpu_inference/layers/vllm/custom_ops/mla_attention.py
karan: Switched to `getattr` based access instead.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
karan: Good question @kyuyeunk. Mistral 3 Large uses blockwise FP8 quantization with true 2d scales which need handling here.

I also added a comment here explaining this.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
karan: There's 2 changes here mainly:

1. blockwise quantization support (needed for the model)
2. reducing model bring up time by offloading moe weight transformations to the jax device (though this is now reverted)

I've added comments for these as well

tpu_inference/models/vllm/vllm_model_wrapper.py
karan: I actually think that after https://github.com/vllm-project/tpu-inference/pull/2504/changes#diff-1aa20442c7dfbac299386b46a7ce3009199c85f891708f66e365df5d6b7ffcb1, my change here might be moot[0] so I'll undo my change here.

[0] It removes boolean indexing which was causing an issue on multi-host set up.

tpu_inference/models/vllm/experimental/model_patcher.py
karan: I've wrapped this in an env var as well to reduce the blast radius of this patch.

tpu_inference/layers/vllm/quantization/compressed_tensors/compressed_tensors_moe/compressed_tensors_moe_w8a8_fp8.py
karan: I've removed the jit here again because double jitting should definitely not be required. I've been able to test with 5 layers for now and it works okay for now. Once this PR goes in, I'll test with the full model once again and see if/why there's an OOM.

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py
karan: Okay I've gone with the linked approach.

gpolovets1 (approved):

tpu_inference/layers/common/process_weights/moe_weights.py
kyuyeunk: just do `jax_view(w2_weight_scale)`?

PR #2426, opened 2026-04-28
[TPU KV Offload][Feat] Option to override host memory kind

sangam-jindal:
Use an ENV variable to override host memory kind type. By default defaults to pinned_host.

# Description

This change introduces an environment variable `TPU_OFFLOAD_USE_UNPINNED_HOST` to allow overriding the host memory kind of host buffer  for KV offloading.

Default behavior: It defaults to `pinned_host` if the environment variable is not set.

# Tests

Added unit tests in tests/offload/tpu_offload_connector_worker_test.py to verify:

- Correct override behavior when the env var is set.
- Defaulting to pinned_host when unset.
- Error handling for invalid values.
- Tested locally with QWEN32B model  run with sglang bench and checked the logs to ensure both pinned/unpinned are good.

[patchset - 2]
Addresses comment regarding juncgu-google@ .

### DIFF
--- a/tests/offload/tpu_offload_connector_worker_test.py
+++ b/tests/offload/tpu_offload_connector_worker_test.py
@@ -226,57 +226,134 @@ class TestTPUOffloadConnectorWorker(jtu.JaxTestCase):
         )
 
     @parameterized.named_parameters(
+        # Pinned Host Cases
         dict(
-            testcase_name="_single_block",
+            testcase_name="_single_block_pinned",
             num_blocks_to_save=1,
             num_requests=1,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_requests_single_block",
+            testcase_name="_multi_requests_single_block_pinned",
             num_blocks_to_save=1,
             num_requests=6,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_blocks",
+            testcase_name="_multi_blocks_pinned",
             num_blocks_to_save=5,
             num_requests=1,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_requests_multi_blocks",
+            testcase_name="_multi_requests_multi_blocks_pinned",
             num_blocks_to_save=16,
             num_requests=6,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_blocks_with_compile_jax",
+            testcase_name="_multi_blocks_with_compile_jax_pinned",
             num_blocks_to_save=5,
             num_requests=1,
             use_precompiled_swap_ops=True,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_requests_single_block_with_compile_jax",
+            testcase_name=
+            "_multi_requests_single_block_with_compile_jax_pinned",
             num_blocks_to_save=1,
             num_requests=6,
             use_precompiled_swap_ops=True,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_multi_requests_multi_blocks_with_compile_jax",
+            testcase_name=
+            "_multi_requests_multi_blocks_with_compile_jax_pinned",
             num_blocks_to_save=5,
             num_requests=6,
             use_precompiled_swap_ops=True,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_final_save",
+            testcase_name="_final_save_pinned",
             num_blocks_to_save=1,
             num_requests=1,
             is_final_save=True,
             skip_save=False,
+            use_unpinned_host=False,
         ),
         dict(
-            testcase_name="_final_skip_save",
+            testcase_name="_final_skip_save_pinned",
             num_blocks_to_save=0,
             num_requests=1,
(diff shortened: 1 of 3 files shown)


sangam-jindal:
@juncgu-google  Added a patch for flipping memory kind via env variable. Please review (I am unable to add to reviewers section for some reason).

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1287,12 +1288,20 @@ def register_runner(self, runner: TPUModelRunner):
             self.device_sharding = kv_layer.sharding
             self.num_kv_blocks = self.shape[0]
 
+            host_memory_kind = os.getenv('KV_OFFLOAD_HOST_MEMORY_KIND', 'pinned_host')
```
juncgu-google: please move the env definition to `tpu_inference/envs.py`; you can use [TPU_OFFLOAD_NUM_CPU_CHUNKS](https://github.com/vllm-project/tpu-inference/blob/fb9a1206390cc287f365bba2231d2b8dfdd9c975/tpu_inference/envs.py#L270) as an example.

juncgu-google (commented):
please update the pr description; and test this feature in tests/offload/tpu_offload_worker/utils_test.py

tpu_inference/offload/tpu_offload_connector.py
sangam-jindal: Thanks. Used envs.

sangam-jindal:
> please update the pr description; and test this feature in tests/offload/tpu_offload_worker/utils_test.py

Sure . So I added this in the worker_test assuming that was the intended path (I couldn't find the path you mentioned). But I am happy to move it to another file if I misunderstood the path you mentioned. The following passes and I was able to observed prints flipped towards the right toggle.

```
PYTHONPATH=.${PWD}/../vllm python3 -m unittest tests.offload.tpu_offload_connector_worker_test
```

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -100,6 +100,7 @@
      internal tracking sets (`_reqs_being_saved`).
 """
 import copy
+import os
```
juncgu-google: I think we do not need `import os` in this file.

tpu_inference/envs.py
```diff
@@ -281,6 +282,9 @@ def _get_str_list_env() -> list[str]:
     # kv offload to dram: prometheus metrics log interval in seconds
     "TPU_OFFLOAD_METRICS_LOG_INTERVAL":
     lambda: int(os.getenv("TPU_OFFLOAD_METRICS_LOG_INTERVAL", "10")),
+    # kv offload to dram: host memory kind for kv cache tensors
+    "KV_OFFLOAD_HOST_MEMORY_KIND":
+    lambda:os.getenv("KV_OFFLOAD_HOST_MEMORY_KIND", "pinned_host"),
```
juncgu-google: Let's make this env var a boolean for now (since we only have "pinned" vs. "unpinned"). For example `TPU_OFFLOAD_USE_UNPINNED_HOST = False` (default).

tests/offload/tpu_offload_connector_worker_test.py
```diff
@@ -689,3 +689,27 @@ def body_fun(i, val):
         current_device_val = np.array(
             worker.runner.kv_caches[0][block_to_save])
         self.assertArraysEqual(current_device_val, np.array(corruption_val))
+
```
juncgu-google, line 822: I think we should also add "pinned" / "unpinned" host option into the tests for [save](https://github.com/vllm-project/tpu-inference/blob/5db2d7476d619548d583c63262e232f3957861d0/tests/offload/tpu_offload_connector_worker_test.py#L282) and load in the worker_test.py, and also the tests for gather / [update](https://github.com/vllm-project/tpu-inference/blob/5db2d7476d619548d583c63262e232f3957861d0/tests/offload/tpu_offload_utils_test.py#L142) in the utils_test.py,

sangam-jindal: Sure. Good point. Added & ran the tests for both worker and utils.

tpu_inference/envs.py
sangam-jindal: Good idea. Done.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1290,13 +1290,15 @@ def register_runner(self, runner: TPUModelRunner):
             # kv_caches[0].sharding.spec in update_kv_caches_one at runtime,
             # which can differ due to buffer donation side effects
             self.cached_kv_sharding_spec = kv_layer.sharding.spec
+            host_memory_kind = "unpinned_host" if envs.TPU_OFFLOAD_USE_UNPINNED_HOST else "pinned_host"
```
juncgu-google: please make `host_memory_kind` be an attribute of the `TPUOffloadConnectorWorker`, and move it to __init__()

sangam-jindal: Sure thing. Done.

juncgu-google (approved):

PR #2429, opened 2026-04-28
Add a test case using lm_eval's mmlu dataset and fix the parameter pa…

CienetStingLin:
# Description

Add an accuracy test case using the mmlu dataset with model Llama-3.3-70B
Fix the parameter passing issue when running the mmlu test, and enhance the default value when parser_case cannot retrieve a value
Increase the length limit of column `RunBy` in the Spanner DB

# Tests

[Buildkite CI](https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/206)

# Checklist


jcyang43 (approved):

.buildkite/benchmark/lm_eval/math500/run.sh
```diff
@@ -16,30 +16,28 @@
...
 
+unset MODEL_IMPL_TYPE VLLM_XLA_CHECK_RECOMPILATION
+
 mkdir -p "$LOG_DIR"
 
 CMD=(
     lm_eval
     --model vllm
-    --model_args "pretrained=$MODEL_NAME,tensor_parallel_size=${TP_SIZE:-8},dtype=auto,max_model_len=2048,gpu_memory_utilization=0.98"
+    --model_args "pretrained=$MODEL_NAME,tensor_parallel_size=${TP_SIZE:-8},dtype=auto,max_model_len=$MAX_MODEL_LEN,gpu_memory_utilization=0.98"
```
QiliangCui2023: `${TP_SIZE:-8}` — `TP_SIZE` is not exported anywhere in the pipeline (parser_case.py exports `TENSOR_PARALLEL_SIZE` at L204/L222, run_bm.sh logs `TENSOR_PARALLEL_SIZE` at L169). This will always silently fall back to 8. Use `${TENSOR_PARALLEL_SIZE:-8}` to match `mmlu/run.sh:40`.

.buildkite/benchmark/lm_eval/mlperf/run.sh
```diff
@@ -16,30 +16,28 @@
...
 
+unset MODEL_IMPL_TYPE VLLM_XLA_CHECK_RECOMPILATION
+
 mkdir -p "$LOG_DIR"
 
 CMD=(
     lm_eval
     --model vllm
-    --model_args "pretrained=$MODEL_NAME,tensor_parallel_size=${TP_SIZE:-8},dtype=auto,max_model_len=2048,gpu_memory_utilization=0.98"
+    --model_args "pretrained=$MODEL_NAME,tensor_parallel_size=${TP_SIZE:-8},dtype=auto,max_model_len=$MAX_MODEL_LEN,gpu_memory_utilization=0.98"
```
QiliangCui2023: Same as math500/run.sh:40 — `TP_SIZE` is never set, so this always falls back to 8. Use `${TENSOR_PARALLEL_SIZE:-8}`.

QiliangCui2023 (commented):
`TP_SIZE` is silently never set — `parser_case.py` only exports `TENSOR_PARALLEL_SIZE` (see `parser_case.py:204, 222`), and `run_bm.sh:169` only logs `TENSOR_PARALLEL_SIZE`. So `math500/run.sh` and `mlperf/run.sh` will always use the `:-8` fallback, regardless of what the JSON specifies. `mmlu/run.sh` already uses the correct name. See inline.

.buildkite/benchmark/lm_eval/math500/run.sh
CienetStingLin: Fixed, thanks for the review

.buildkite/benchmark/lm_eval/mlperf/run.sh
CienetStingLin: Fixed

QiliangCui2023 (approved):

QiliangCui (approved):

docker/Dockerfile
```diff
@@ -70,7 +70,7 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install -e tests/v
 RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install \
     git+https://github.com/thuml/depyf.git \
     pytest-asyncio \
-    "lm-eval[api]==0.4.12" \
+    "lm-eval[api,math]==0.4.12" \
```
CienetStingLin, line 73: Noticed that the version of lm-eval[api] has been updated, so lm-eval[math] is merged together rather than written separately to avoid version conflicts.

PR #2431, opened 2026-04-28
[Bug][Qwen3.5] Honor has_initial_state in fused GDN kernel

qizzzh:
PR #2408 plumbed `has_initial_state` through the chunked / ref GDN paths to fix stale-slot leakage on freshly-allocated mamba slots, but explicitly skipped the fused Pallas kernel ("does not yet honor has_initial_state"). With compact-mamba shrinking the mamba pool to `max_num_reqs+1`, slot reuse is the common case and the FUSED path silently consumes a freed slot's prior state on every brand-new prefill — the same kind of silent corruption that hit the chunked path before PR #2408.

Closes that gap on the FUSED path:

- `fused_gdn_recurrent_kernel.py`: add a `has_initial_state` SMEM input; after the DMA load, `@pl.when(is_new_seq & ~has_init)` overwrites `h_bufs[buf_idx]` with zeros so the recurrent update starts from zero instead of stale tenant data. The decode kernel is unchanged — decodes always have a valid prior state.
- `fused_gdn_kernel_wrapper.py`: forward `has_initial_state` through `fused_gdn`, `_dispatch_with_distribution`, and the `ragged_gated_delta_rule` adapter. Default `None` materializes a static all-True ones array, preserving the pre-fix behaviour for any callers that don't manage slot reuse.
- `gdn_attention.py`: flip the FUSED branch to pass `has_initial_state`, matching the chunked / ref branches.

Perf (apples-to-apples microbench on v7x-8, median of 200 reps, H_qk=2 H_v=8 K=V=128):

  pure_decode_128       2.437 -> 2.447 ms  (+0.4%)
  mixed_64dec_4x256pre  3.668 -> 3.664 ms  (-0.1%)
  pure_prefill_4x1024   5.879 -> 5.877 ms  (-0.0%)
  decode_256            4.237 -> 4.197 ms  (-0.9%)

All within ±1% noise — the SMEM input + conditional adds no measurable cost, and `_zero_h0` only fires on `is_new_seq` for slots flagged False.

Tests (5 new in tests/kernels/fused_gdn_kernel_test.py + 2 FUSED parametrize entries on the existing dispatcher-level `test_has_initial_state_*` tests in tests/layers/common/test_gdn_attention.py; all pass on TPU v7x-8):

- `test_has_initial_state_zeros_stale_slot` — fused-with-stale h0 vs fused-with-fresh h0, the strongest direct test that the kernel zeros stale state.
- `test_has_initial_state_pattern_matches_ref_all_true` — True path preserves the loaded h0.
- `test_has_initial_state_pattern_matches_ref_alternating` — per-slot SMEM lookup (catches off-by-one or uniform-apply bugs).
- `test_has_initial_state_long_sequence_multi_block` — kernel zeros only on `is_new_seq`; subsequent blocks of the same prefill must not re-zero (or the running state gets wiped mid-sequence).
- `test_has_initial_state_decode_plus_new_prefill` — decode kernel + recurrent kernel interaction; decode-slot state survives the recurrent kernel zeroing prefill slots.

Eval - MMLU-Pro looks good

|mmlu_pro           |      2|custom-extract|      |exact_match|↑  |0.8321|±  |0.0224|
| - biology         |      3|custom-extract|     5|exact_match|↑  |0.8500|±  |0.0819|
| - business        |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - chemistry       |      3|custom-extract|     5|exact_match|↑  |0.8500|±  |0.0819|
| - computer_science|      3|custom-extract|     5|exact_match|↑  |0.8500|±  |0.0819|
| - economics       |      3|custom-extract|     5|exact_match|↑  |0.9000|±  |0.0688|
| - engineering     |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - health          |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|
| - history         |      3|custom-extract|     5|exact_match|↑  |0.7500|±  |0.0993|
| - law             |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|
| - math            |      3|custom-extract|     5|exact_match|↑  |0.9500|±  |0.0500|
| - other           |      3|custom-extract|     5|exact_match|↑  |0.7000|±  |0.1051|
| - philosophy      |      3|custom-extract|     5|exact_match|↑  |0.9000|±  |0.0688|
| - physics         |      3|custom-extract|     5|exact_match|↑  |1.0000|±  |0.0000|
| - psychology      |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0918|

|mmlu_pro|      2|custom-extract|      |exact_match|↑  |0.8321|±  |0.0224|

### DIFF
--- a/tests/kernels/fused_gdn_kernel_test.py
+++ b/tests/kernels/fused_gdn_kernel_test.py
@@ -314,15 +314,18 @@ class FusedGdnKernelTest(jtu.JaxTestCase):
                             check_dtypes=False)
 
     @parameterized.named_parameters(
-        # `all_true` covers the True-preserves-loaded-state path;
-        # `alternating` exercises the per-slot SMEM lookup (catches
-        # off-by-one or uniformly-applied conditional). The all-False
-        # case is covered by `test_has_initial_state_zeros_stale_slot`
-        # (which is a stronger test — fused-with-stale vs
-        # fused-with-fresh, directly validating the kernel's zeroing
-        # rather than just matching ref). Extra positional patterns
-        # (first-only / last-only / etc.) add CI time without new
-        # coverage given the kernel's per-seq SMEM-indexed conditional.
+        # Two patterns chosen for orthogonal coverage:
+        #   * `all_true` — every slot uses its loaded h0 (no
+        #     zeroing); the continuation case.
+        #   * `alternating` — has_initial_state varies by slot, so
+        #     the kernel must look up the right SMEM entry per
+        #     seq; catches off-by-one or uniformly-applied
+        #     conditional bugs.
+        # The all-False case is covered by
+        # `test_has_initial_state_zeros_stale_slot`, which directly
+        # compares fused-with-stale vs fused-with-fresh rather than
+        # matching the ref. Extra patterns (first-only, last-only,
+        # ...) add CI time without new coverage.
         dict(testcase_name="all_true", pattern=[True, True, True, True]),
         dict(testcase_name="alternating", pattern=[True, False, True, False]),
     )
--- a/tests/kernels/fused_gdn_kernel_test.py
+++ b/tests/kernels/fused_gdn_kernel_test.py
@@ -105,6 +105,7 @@ class FusedGdnKernelTest(jtu.JaxTestCase):
         use_dt_bias=False,
         lower_bound=None,
         atol=1e-2,
+        has_initial_state_active=None,
     ):
         rng = np.random.RandomState(42)
         q, k, v, a, b, A_log, h0, cu_seqlens, state_indices, N = _make_inputs(
@@ -121,6 +122,24 @@ class FusedGdnKernelTest(jtu.JaxTestCase):
 
         dt_bias = (jnp.array(rng.randn(H_v).astype(np.float32))
                    if use_dt_bias else jnp.zeros(H_v, dtype=jnp.float32))
+
+        # `has_initial_state_active` is a python list of bools for the
+        # ACTIVE sequences only (length N); we pad to `max_num_req` with
+        # True for the padded slots (their state isn't read by either
+        # impl, but the array shape must match `state_indices`). Default
+        # behaviour (None) is all-True, which matches the pre-fix path
+        # and keeps the existing baseline tests valid.
+        max_num_req_padded = state_indices.shape[0]
+        if has_initial_state_active is None:
+            has_initial_state = jnp.ones((max_num_req_padded, ),
+                                         dtype=jnp.bool_)
+        else:
+            assert len(has_initial_state_active) == N, (
+                f"has_initial_state_active must have length N={N}, got "
+                f"{len(has_initial_state_active)}")
+            full = np.ones(max_num_req_padded, dtype=np.bool_)
+            full[:N] = np.array(has_initial_state_active, dtype=np.bool_)
+            has_initial_state = jnp.array(full)
 
         # ── Reference (ragged_gated_delta_rule_ref) ──
         mixed_qkv = jnp.concatenate(
@@ -141,6 +160,7 @@ class FusedGdnKernelTest(jtu.JaxTestCase):
             cu_seqlens,
             state_indices,
             distribution_ref,
+            has_initial_state,
             n_kq=H_qk,
             n_v=H_v,
             d_k=K,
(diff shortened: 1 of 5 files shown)


tpu_inference/layers/common/gdn_attention.py
```diff
@@ -168,9 +170,11 @@ def run_jax_gdn_attention_local(
         )
 
     if config.ragged_gated_delta_rule_impl == RaggedGatedDeltaRuleImpl.FUSED:
-        # The fused kernel does not yet honor `has_initial_state`. Fall
-        # through with the existing call signature; the chunked/ref paths
-        # used in production already mask via the new argument below.
+        # All three impls (fused / chunked / ref) honor `has_initial_state`
+        # — the fused recurrent kernel zeros h0 in VMEM for slots whose
+        # flag is 0. The branch is preserved (rather than unified into a
```
jrplatin: Not sure I really understand this / if this is needed

qizzzh: Ah stale comment during the development. Updated.

tests/kernels/fused_gdn_kernel_test.py
```diff
@@ -220,6 +241,139 @@ def test_padded_max_num_req(self, decode_N, mixed_seqlens, extra_pad):
...
+                            check_dtypes=False)
+        # Compare the active slots; padding slots are not exercised.
+        self.assertAllClose(state_stale[:N],
+                            state_fresh[:N],
+                            atol=1e-4,
+                            rtol=1e-4,
+                            check_dtypes=False)
+
+    @parameterized.named_parameters(
+        # `all_true` covers the True-preserves-loaded-state path;
```
jrplatin: Don't really follow this either -- what does "True-preserves-loaded-state path" mean? Can you please simplify / make the wording here a bit more clear? Thanks!

tpu_inference/kernels/gdn/fused_gdn_kernel_wrapper.py
```diff
@@ -138,6 +145,14 @@ def fused_gdn(
         distribution: ``i32[2]`` — ``(decode_end, total)``.
         b: Raw betas ``[T, H_v]`` (sigmoid applied inside kernel).
             ``None`` means beta=1 (no beta gating).
+        has_initial_state: Boolean / int32 ``[max_num_req]``. ``True`` /
```
helloworld1: The has_initial_state has always been passing in as Boolean (i8) and also in chunked implementation, but I think kernel needs to convert to int32 for SMEM. Can we just document it as always boolean type so all implementations uses the same types?

helloworld1 (approved):

tpu_inference/kernels/gdn/fused_gdn_kernel_wrapper.py
```diff
@@ -187,6 +202,15 @@ def fused_gdn(
...
             jnp.float32)  # [H_v, num_lanes]
 
+    # Default to all-True (no masking), matching the pre-fix behaviour.
+    # Cast to int32 for SMEM compatibility — the recurrent kernel checks
+    # `has_init == 0` to decide whether to zero h0.
+    max_num_req = state_indices.shape[0]
+    if has_initial_state is None:
+        has_initial_state = jnp.ones((max_num_req, ), dtype=jnp.int32)
+    else:
+        has_initial_state = has_initial_state.astype(jnp.int32)
```
jrplatin, line 213: Why do we need to cast it? Aren't we initing it as INT32?

tpu_inference/kernels/gdn/fused_gdn_recurrent_kernel.py
jrplatin, line 1: @yaochengji Can you PTAL at these changes? Thanks!

tests/kernels/fused_gdn_kernel_test.py
qizzzh: Updated

tpu_inference/kernels/gdn/fused_gdn_kernel_wrapper.py
qizzzh: Updated

tpu_inference/kernels/gdn/fused_gdn_kernel_wrapper.py
qizzzh: As mentioned by @helloworld1 above, the kernel needs to convert it to int32 for SMEM

jrplatin (approved):

tpu_inference/kernels/gdn/fused_gdn_kernel_wrapper.py
helloworld1: We need to cast because kernel's SMEM only support int32 or float32.

yaochengji (approved):
LGTM on the kernel change, thanks!

PR #2434, opened 2026-04-28
refactor the bm vllm server start wait logic

yiw-wang:
Refactor the vllm server startup waiting logic.

This change aims to increase the waiting time for large model like Qwen/Qwen3.5-397B which may need more than 60 mins for turnup. And also directly exit with status 1 if the server cannot be turn up within the MAX_WAIT_SECONDS. During the test, we found out the later steps may fetch and pass dirty(e.g. 0ms) metrics to the Spanner DB if the vllm_bench_serve starts but the server is not up.

b/505124705

# Tests

[First Revision] - https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/211/canvas

[Second Revision] - https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/212/canvas

# Checklist

### DIFF
--- a/.buildkite/benchmark/scripts/run_bm.sh
+++ b/.buildkite/benchmark/scripts/run_bm.sh
@@ -237,21 +237,61 @@ if ! kill -0 "$VLLM_PID" 2>/dev/null; then
     exit 1
 fi
 
-echo "wait for 60 minutes.."
-echo
-for _ in {1..360}; do
-    if grep -Fq "raise RuntimeError" "$VLLM_LOG"; then
-        echo "Detected RuntimeError, exiting."
-        cat "$VLLM_LOG"
+# ---------------------------------------------------------
+# Server startup wait logic
+# ---------------------------------------------------------
+SERVER_WAIT_MINS=${SERVER_WAIT_MINS:-60}
+
+MAX_WAIT_SECONDS=$((SERVER_WAIT_MINS * 60))
+WAIT_START_TIME=$(date +%s)
+ELAPSED=0
+
+echo "Waiting up to ${SERVER_WAIT_MINS} minutes for server to start (PID: ${VLLM_PID})..."
+
+# Initial state set to not started
+SERVER_STARTED="false"
+
+# Loop continues as long as elapsed time is within the maximum allowed
+while (( ELAPSED <= MAX_WAIT_SECONDS )); do
+    
+    # 1. [Fail-Fast Check] Ask the OS if the process is still alive
+    if ! kill -0 "$VLLM_PID" 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -237,22 +237,67 @@ if ! kill -0 "$VLLM_PID" 2>/dev/null; then
...
-echo
-for _ in {1..360}; do
+# ---------------------------------------------------------
+# Server startup wait logic
+# ---------------------------------------------------------
+WAIT_MINS=60
+
+# Dynamically extend timeout for giant model like Qwen/Qwen3.5-397B 
+if [[ "$MODEL" == "Qwen/Qwen3.5-397B"* ]]; then
+    WAIT_MINS=120
```
QiliangCui: is it ok to make this WAIT_MINS a env variable and pass in from the benchmark's json config?

.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -237,22 +237,67 @@ if ! kill -0 "$VLLM_PID" 2>/dev/null; then
...
+    
+    # 1. Check for success flag: Application startup complete
+    if grep -Fq "Application startup complete" "$VLLM_LOG"; then
+        echo "Application started successfully."
+        SERVER_STARTED="true"
+        break
+    fi
+
+    # 2. Check for failure flag: RuntimeError
     if grep -Fq "raise RuntimeError" "$VLLM_LOG"; then
```
QiliangCui: can we use this logic for the wait?

https://github.com/QiliangCui/bm-infra/blob/main/scripts/agent/run_bm.sh#L116-L129

the logic like

```
# Check if vllm process has exited (crashed for any reason)
    if ! kill -0 $VLLM_PID 2>/dev/null; then
        echo "vllm process (PID=$VLLM_PID) has exited unexpectedly. Last 20 lines of vllm log:"
        tail -20 "$VLLM_LOG"
        exit 1
```

It is better than trying to guess the "runtime error".

QiliangCui (changes requested):

.buildkite/benchmark/scripts/run_bm.sh
yiw-wang: this is really a good suggestion, as sometime the long turnup time is not really caused by the model size but also caused by the arguments which may cause the JAX complication time increased. Done

.buildkite/benchmark/scripts/run_bm.sh
yiw-wang: Definitely better than the original error detection logic, also encountered non RuntimeError exceptions during the test which caused the main process wait till the end of WAIT_TIME. Done

QiliangCui (approved):

PR #2450, opened 2026-04-29
Added a KV length threshold to xprof capture trigger and use it for DSv3 multihost decode

m-liu:
# Description

- Added a KV length threshold criteria to xprof capture trigger.
- Added xprof scripts for DSv3 multihost to capture decode only batches where KV context length is 1K and 9K for all requests in batch. 

We use sequence length of 1 / 9K (prefill input seq length of 1) ensure batches will be all decode tokens executing in lockstep of increasing KV seq lengths.


scripts/multihost/benchmarks/torchax/xprof/xprof_deepseek_v3_decode_9k_context_random_routing.sh
```diff
@@ -0,0 +1,40 @@
...
+  --additional-config '{"compilation_sizes": [2560], "sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' \
+  --disable-shared-experts-stream "0" \
+  --generation-config "gs://gpolovets-inference/deepseek/generation_configs/DeepSeek-R1" \
+  --vllm-mla-disable "0" \
+  --moe-requantize-block-size "512" \
+  --moe-requantize-weight-dtype "fp4" \
+  --moe-all-gather-activation-dtype "fp8" \
+  --force-moe-random-routing "1" \
+  --phased-profiling-dir "gs://tpu-commons-ci/xprof/deepseek-r1/torchax/decoode-9k-context-random-routing" \
+  --phased-profiler-decode-only-kv-len-threshold 9100 \
```
gxd3, line 39: nit: at the end of the benchmark, the completed requests are removed from the batch (some requests are slightly ahead of other, due to being prefilled earlier by uLLM), perhaps set this to ~8500 so that we still have a full batch?

gxd3 (approved):

scripts/multihost/benchmarks/torchax/xprof/xprof_deepseek_v3_decode_9k_context_random_routing.sh
gxd3: oh, just realize that you set prefill length to be 1, all requests should mostly at the same length then.

tpu_inference/runner/utils.py
```diff
@@ -383,6 +395,16 @@ def _start_profiling(self, batch_composition_stats: dict) -> None:
                     self.decode_steps_skipped, self.num_decode_steps_to_skip)
                 break
 
+            # Skip decode-only steps until min KV len exceeds threshold
+            if phase == InferencePhase.DECODE_ONLY and \
+                    self.decode_kv_len_threshold >= 0:
+                min_kv_len = batch_composition_stats.get("min_kv_len", 0)
+                if min_kv_len <= self.decode_kv_len_threshold:
```
gpolovets1: nit: Why not use `min_kv_len < ...`?
Otherwise the flag implies all requests need to be greater than this threshold which is a little bit misleading.

gpolovets1 (approved):

PR #2457, opened 2026-04-30
[Kernel] Add sparse MLA attention Pallas kernel (per-query top-k)

NotDevinLe:
# Description

* Add sparse MLA attention Pallas kernel under `tpu_inference/kernels/sparse_mla/v1/`. Does MLA attention restricted to a per-query top-k subset of KV positions, given a `topk_indices: i32[num_tokens, topk]` index list with -1 sentinel for unused slots. KV cache layout, paging, and FlashAttention-2 online softmax are identical to the existing dense MLA kernel.

* Architecture: parameterized MLA kernel. Adds an optional `topk_indices` HBM input + `is_sparse` static flag to the existing `mla_ragged_paged_attention`. Sparse path `pltpu.sync_copy`s a per-Q-block slice of `topk_indices` into a VMEM scratch buffer, builds a membership-test mask, and ORs with the existing causal mask inside `flash_attention` before softmax. Dense path passes a 1×1 sentinel; the `is_sparse=False` branch is elided at JIT time — no behavior change for existing callers.

* `tpu_inference/kernels/sparse_mla/v1/kernel.py` is a thin wrapper that delegates to the parameterized MLA kernel with `topk_indices` set. Pure-JAX reference impl included as the correctness oracle.

* Adds a row to `support_matrices/release/v6e/flax_nnx/kernel_support_matrix.csv` and a feature yaml at `.buildkite/features/Sparse_MLA_Attention.yml` (generated via `add_feature_to_ci.py`).

Motivation: top-k sparse attention is the consumer side of the scheme used in DeepSeek V3.2 (`index_topk`) and inherited by GLM-5.1. This kernel is the TPU implementation of that primitive — model-agnostic, callable by any layer that wants top-k sparse attention on top of an MLA-shaped KV cache. The producer (FP8 indexer) is planned as a follow-up.

Targets v6e and v7x. v4 deferred.

# Tests

All run on v6e-8.

Correctness — `pytest tests/kernels/sparse_mla_v1_test.py` — **15/15 green**:
* 6 reference-impl correctness cases (CPU + TPU compatible), including the load-bearing `topk_full == dense_MLA reference` sanity.
* 5 Pallas-kernel-vs-reference cases on TPU, including `kernel(topk_full) == dense_MLA kernel` — proves the mask injection is correct via two independent code paths.
* 4 perf cases following the timing pattern from `tests/kernels/gather_reduce_test.py`.

No regression — `pytest tests/kernels/mla_v1_test.py` — **14/14 green** (pre-existing dense MLA suite).

Performance (v6e-8, bf16, 5-iter median, after warmup):

| Case | Dense MLA (ms/iter) | Sparse MLA (ms/iter) | Ratio |
|------|---------------------|----------------------|-------|
| `decode_only` `[(1, 256), (1, 1024)]` | 0.274 | 0.286 | 1.04× |
| `mixed_batch` `[(64, 64), (1, 512), (32, 800)]` | 1.168 | 1.181 | 1.01× |
| `prefill_small` `[(64, 64)]` | 0.655 | 0.657 | 1.00× |
| `prefill_uneven` `[(48, 200), (96, 96)]` | 1.579 | 1.580 | 1.00× |

Mask-based v1 still loads all KV pages (same DMA as dense), so parity with dense is the goal — not a speedup. Topk-mask compute overhead ≤4% on the smallest decode case, ≤1% otherwise. Truly-sparse-load (skip non-topk pages, realize the bandwidth win) is a v2 optimization.

# Checklist

### DIFF
--- a/.buildkite/features/Sparse_MLA_Attention.yml
+++ b/.buildkite/features/Sparse_MLA_Attention.yml
@@ -1,0 +1,61 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pipeline-name: Sparse MLA Attention
+# pipeline-type: kernel support matrix
+steps:
+  - label: "${TPU_VERSION:-tpu6e} Correctness tests for Sparse MLA Attention"
+    key: "${TPU_VERSION:-tpu6e}_Sparse_MLA_Attention_CorrectnessTest"
+    soft_fail: true
+    agents:
+      queue: "${TPU_QUEUE_SINGLE:-tpu_v6e_queue}"
+    commands:
+      - .buildkite/scripts/run_in_docker.sh python3 -m pytest -s -v /workspace/tpu_inference/tests/kernels/sparse_mla_v1_test.py
+  - label: "${TPU_VERSION:-tpu6e} Record correctness test result for Sparse MLA Attention"
+    key: "${TPU_VERSION:-tpu6e}_record_Sparse_MLA_Attention_CorrectnessTest"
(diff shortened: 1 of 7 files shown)


kyuyeunk (changes requested):
> No microbenchmark included; perf tuning is a fast-follow PR. The mask-based v1 kernel still loads all KV pages — truly-sparse-load (skip non-topk pages) is a v2 optimization. The PerformanceTest row in the support matrix is marked unverified accordingly.

for kernels, performance measurement is a requirement not an optional. no use in spending time and effort performance tuning if the architecture is fundamentally wrong.

PR #2467, opened 2026-04-30
Improve the BM workflow error handling

yiw-wang:
Enhance the error handling logic for catching the error from bm client and add some metrics validation to fail fast.

The original way to start the `vllm bench serve`, client may end and/or crush for any reason, and then print in `bm_log.txt`. And then the `grep` may grep nothing and `p99_int=$(printf "%.0f" "")` will return `0` as the latency and the throughput which is wrong.

With current change if the `vllm bench serve` is failed, it will return `$client_exit_code` and we will do error check before process and grep the metrics value. And if the metrics value failed the validation, we will end early as well.

### DIFF
--- a/.buildkite/benchmark/scripts/run_bm.sh
+++ b/.buildkite/benchmark/scripts/run_bm.sh
@@ -330,8 +330,15 @@ run_benchmark(){
   fi
 
   echo "[DEBUG] Executing client_cmd: ${CLIENT_CMD_ENVS[*]} ${CLIENT_CMD[*]} > $BM_LOG" >&2
+  set +e
   # Execute the array directly, preserving strict argument boundaries
   env "${CLIENT_CMD_ENVS[@]}" "${CLIENT_CMD[@]}" > "$BM_LOG" 2>&1
+  local client_exit_code=$?
+  set -e
+
+  if [ $client_exit_code -ne 0 ]; then
+      return $client_exit_code
+  fi
 
   throughput=$(grep "Request throughput (req/s):" "$BM_LOG" | sed 's/[^0-9.]//g')
   p99_e2el=$(grep "P99 E2EL (ms):" "$BM_LOG" | awk '{print $NF}')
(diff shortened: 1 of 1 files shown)


.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -348,13 +355,60 @@ printf "[DEBUG] Checking folder structure (Environment: %s)...\n" "$ENV_CONTEXT"
...
+VALID_THROUGHPUT=""
+VALID_P99_E2EL=""
+
+execute_benchmark_safely() {
+    local rate_arg="${1:-}" # Accept the request_rate argument; default to empty string if not provided
+    local output
+    local bm_exit_code
+
+    # 1. Execute the benchmark and intercept the exit code from the subshell pipeline
+    output=$(run_benchmark "$rate_arg" | tail -n 1)
```
QiliangCui2023, line 372: line 16 is `set -euo pipefail`

in this case, if run_benchmark return non zero, it seems the next line can't executed... for example

```
#!/bin/bash
 set -euo pipefail
 inner() { return 17; }
 output=$(inner | tail -n 1)
 echo "still here"   # never prints — script exited at the assignment
```

maybe you need to use the same pattern as above?


```
 set -euo pipefail
 inner() { return 17; }
 set +e
 output=$(inner | tail -n 1)
 bm_exit_code=$?  # local if in unction.
 set -e
 echo "still here"   # now this works.
 echo $bm_exit_code
```

.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -348,13 +355,60 @@ printf "[DEBUG] Checking folder structure (Environment: %s)...\n" "$ENV_CONTEXT"
...
+    bm_exit_code=${PIPESTATUS[0]}
```
QiliangCui2023: if fix as above, is this necessary?

QiliangCui2023 (dismissed):

.buildkite/benchmark/scripts/run_bm.sh
yiw-wang: yes, thats what the newly added line 333 and 337 are needed here. Exactly what you did with 'set +e' then 'set -e' in your example.

.buildkite/benchmark/scripts/run_bm.sh
yiw-wang: yes, you are right, I didn't know this pattern but definitely seems more straightforward.

yiw-wang: updated.

.buildkite/benchmark/scripts/run_bm.sh
QiliangCui: I think you expect when error happens line 374 - line 379 will be executed and you expect to see "--- Dumping BM_LOG for debugging ---" (of line 376).

But the current code i think won't do it. 

when error happens, line 340 returns some non-zero value. 
This non-zero value breaks line 371 and it terminate the execution directly without running line 372 to line 379 and the log won't be kept. 

for example following code won't see the error log :

```
#!/bin/bash
set -euo pipefail

err() {
    return 15
    }

run_benchmark() { 
    echo "hi"
    
    # the line that has high risk of error
    set +e
    err
    local client_exit_code=$?
    set -e

    if [ $client_exit_code -ne 0 ]; then
      echo "return $client_exit_code"
      return $client_exit_code
    fi

    echo "throught 100"
    }

execute_benchmark_safely() { 
    local bm_exit_code
    echo "calling run_benchmark"
    output=$(run_benchmark | tail -n 1)
    bm_exit_code=$?    
    echo "i got $bm_exit_code"
    if [[ "$bm_exit_code" -ne 0 ]]; then
        echo "You should see me if run_benchmark() returns non-zero which is $bm_exit_code".
        return 1
    fi
    }

execute_benchmark_safely

echo "completed..."
```

instead you need to do the "set +e" to execute_benchmark_safely() like below to see the log

```
#!/bin/bash
set -euo pipefail

err() {
    return 15
    }

run_benchmark() { 
    echo "hi"
    
    # the line that has high risk of error
    set +e
    err
    local client_exit_code=$?
    set -e

    if [ $client_exit_code -ne 0 ]; then
      echo "return $client_exit_code"
      return $client_exit_code
    fi

    echo "throught 100"
    }

execute_benchmark_safely() { 
    local bm_exit_code
    set +e    
    output=$(run_benchmark | tail -n 1)
    bm_exit_code=$?    
    set -e
    echo "i got $bm_exit_code"
    if [[ "$bm_exit_code" -ne 0 ]]; then
        echo "You should see me if run_benchmark() returns non-zero which is $bm_exit_code".
        return 1
    fi
    }

execute_benchmark_safely

echo "completed..."
```

do you see the same?

QiliangCui (changes requested):

.buildkite/benchmark/scripts/run_bm.sh
yiw-wang: you are right, I had a wrong understanding for the 'return' rules in bash, updated.

QiliangCui (approved):

PR #2469, opened 2026-05-01
[Bug][Compact mamba] Clear vacated slot ids to null in InputBatch

qizzzh:
`mamba_state_indices_cpu` is a length-`max_num_reqs` array shipped to the GDN op every step. `remove_request` returned the slot to the free pool but did not clear the position; `condense` copied the slot id from `last_req_index` to `empty_index` but did not clear the source. After enough churn the trailing padded positions [num_reqs:] held slot ids that were also held by active positions [:num_reqs]. The GDN op's `recurrent_state.at[state_indices].set(...)` then scatters with duplicate destination indices — undefined on XLA — and the active position's freshly-written state can lose the race to the padded position's write-back-old-state no-op. Symptom: garbage tail responses on bursty traffic and corrupted single follow-ups, mmlu_pro at high concurrency dropping from ~0.83 to noise.

Bench (`--ignore-eos`) didn't catch it because every request finishes on the same step and condense rarely runs mid-batch. Production traffic and lm_eval finish at variable times so condense fires constantly.

Fix: clear `mamba_state_indices_cpu[req_index] = 0` in `remove_request` and `mamba_state_indices_cpu[last_req_index] = 0` in `condense` after the move. Slot 0 is the reserved null block, never assigned to any active request, so the trailing tail now folds harmlessly into slot 0.

Tests: two regressions in `tests/runner/test_input_batch.py` — `test_mamba_state_indices_no_duplicate_in_padded_tail` and `test_mamba_state_indices_remove_clears_position`. Both fail on stock, pass with the fix.

Verified end-to-end on Qwen3.5-397B-A17B-FP8 TP=8 v7x: smoke test (1 → 64 concurrent → 1 follow-up) all sensible; mmlu_pro --limit 50 at num_concurrent=512 → 0.8271 ±0.014, matching the 0.8321 ±0.022 baseline within stderr.


tpu_inference/runner/input_batch.py
```diff
@@ -308,8 +308,17 @@ def remove_request(self, req_id: str) -> Optional[int]:
         # Return the mamba state slot back to the free pool. The slot's
         # contents in the kv cache are stale and will be zeroed by the
         # has_initial_state guard when the next request takes this slot id.
+        # Reset the slot id at this position to 0 (the reserved null block).
+        # The position becomes part of the padded tail until refilled by an
```
jrplatin: I don't really follow this -- can you re-write to be more clear / simpler? Thanks!

jrplatin (approved):

tpu_inference/runner/input_batch.py
qizzzh: Updated with an example

PR #2471, opened 2026-05-01
Implement missing Punica LoRA ops and custom module parsing

bhuvanpkaruturi:
# Description

This PR completes LoRA support in `tpu-inference` by addressing missing Punica implementations, handling custom LoRA module configurations, and fixing `torchax` environment issues. 

Specifically, this change:
- Implements `add_lora_embedding` and `add_lora_logits` in `PunicaWrapperTPU` to fix the `NotImplementedError` raised during inference. The `add_lora_logits` implementation utilizes a down-projection (`bgmv_shrink`) followed by an up-projection (`bgmv_expand_slice`). The `add_lora_embedding` implementation only performs a up-projection (`bgmv_expand_slice`) since the look up in LoRA A is done beforehand in the vLLM embedding layer.
- Adds parsing for the `LORA_MODULE_PATH` environment variable to override the vLLM engine's `target_modules`. This allows users to pass custom paths like `decoder/layers/self_attention/(query|key|value|out)`.
- Introduces `TPULRUCacheWorkerLoRAManager` to ensure dummy LoRA creation happens within the proper `torchax` environment.

FIXES: b/499306719

# Tests

- Validated the fixes on a tpu7-8 node using `examples/features/lora/multilora_offline.py` from vLLM repo.
- Verified that the system correctly parses and targets modules when `LORA_MODULE_PATH` is set.
- Verified model compilation and inference execute successfully, producing the expected text outputs when relying on the new `add_lora_embedding` and `add_lora_logits` methods.
- Added unit tests for environment parsing (`tests/lora/test_env_parsing.py`) and the Punica TPU wrapper (`tests/lora/test_punica_tpu.py`).

# Checklist

- [ ] I have made or will make corresponding changes to any relevant documentation. - N/A

### DIFF
--- a/tpu_inference/lora/torch_punica_tpu.py
+++ b/tpu_inference/lora/torch_punica_tpu.py
@@ -52,6 +52,9 @@ class PunicaWrapperTPU(PunicaWrapperBase):
     def _get_token_lora_indices(self, x: torch.Tensor) -> torch.IntTensor:
         return torch.narrow(self._token_lora_indices, 0, 0, x.size(0))
 
+    def _get_sampler_indices(self, x: torch.Tensor) -> torch.IntTensor:
+        return torch.narrow(self._sampler_indices, 0, 0, x.size(0))
+
     @property
     def embeddings_indices(self) -> torch.Tensor:
         """
@@ -148,8 +151,18 @@ class PunicaWrapperTPU(PunicaWrapperBase):
             lora_b_stacked (torch.Tensor): lora_b's weights.
             add_inputs (bool): Default to True.
         """
-        raise NotImplementedError(
-            "NYI: torch_punica_tpu.PunicaWrapperTPU.add_lora_embedding.")
+        y_orig = y
+        y = y.view(-1, y.shape[-1])
+        x = x.view(-1, x.shape[-1])
+        
+        add_input = kwargs.get('add_input', add_inputs)
+        
+        # Embedding layer only needs the expand operation as lookup in LoRA A is done beforehand.
+        y = bgmv_expand_slice(x, lora_b_stacked, y,
+                              self._get_token_lora_indices(x),
+                              0, y.shape[-1],
+                              add_input)
+        return y.view(y_orig.shape)
 
     def add_lora_linear(self,
                         y: torch.Tensor,
@@ -226,8 +239,20 @@ class PunicaWrapperTPU(PunicaWrapperBase):
             scale (float): Scaling factor.
             buffer (Optional[torch.Tensor]):Default to None.
         """
-        raise NotImplementedError(
-            "NYI: torch_punica_tpu.PunicaWrapperTPU.add_lora_logits.")
+
+        # Step 1: Shrink operation
+        buffer = bgmv_shrink(x, lora_a_stacked, self._get_sampler_indices(x), scale)
+        
+        y_orig = y
+        y = y.view(-1, y.shape[-1])
+        
+        # Step 2: Expand operation
+        y = bgmv_expand_slice(buffer, lora_b_stacked, y,
+                              self._get_sampler_indices(x),
+                              0, y.shape[-1],
+                              True)
+        
+        return y.view(y_orig.shape)
 
     @property
     def token_lora_indices(self) -> torch.Tensor:
--- a/tpu_inference/lora/torch_punica_tpu.py
+++ b/tpu_inference/lora/torch_punica_tpu.py
@@ -155,13 +155,13 @@ class PunicaWrapperTPU(PunicaWrapperBase):
         y = y.view(-1, y.shape[-1])
         x = x.view(-1, x.shape[-1])
         
-        add_input = kwargs.get('add_input', add_inputs)
+        add_inputs = kwargs.get('add_inputs', add_inputs)
         
         # Embedding layer only needs the expand operation as lookup in LoRA A is done beforehand.
         y = bgmv_expand_slice(x, lora_b_stacked, y,
                               self._get_token_lora_indices(x),
                               0, y.shape[-1],
-                              add_input)
+                              add_inputs)
         return y.view(y_orig.shape)
 
     def add_lora_linear(self,
@@ -240,6 +240,9 @@ class PunicaWrapperTPU(PunicaWrapperBase):
             buffer (Optional[torch.Tensor]):Default to None.
         """
 
+        # Note: We use per-token indices here because the TPU bgmv ops expect a tensor
+        # of shape [num_tokens].
+        
         # Step 1: Shrink operation
         buffer = bgmv_shrink(x, lora_a_stacked, self._get_sampler_indices(x), scale)
         
--- a/tpu_inference/lora/torch_punica_tpu.py
+++ b/tpu_inference/lora/torch_punica_tpu.py
@@ -154,13 +154,12 @@ class PunicaWrapperTPU(PunicaWrapperBase):
         y_orig = y
         y = y.view(-1, y.shape[-1])
         x = x.view(-1, x.shape[-1])
-        
+
         add_inputs = kwargs.get('add_inputs', add_inputs)
-        
+
         # Embedding layer only needs the expand operation as lookup in LoRA A is done beforehand.
         y = bgmv_expand_slice(x, lora_b_stacked, y,
-                              self._get_token_lora_indices(x),
-                              0, y.shape[-1],
+                              self._get_token_lora_indices(x), 0, y.shape[-1],
                               add_inputs)
         return y.view(y_orig.shape)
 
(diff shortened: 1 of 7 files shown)


tpu_inference/lora/torch_punica_tpu.py
vanbasten23, line 136: Consider adding a test to tests/lora/test_layers.py for VocabParallelEmbeddingWithLoRA and LogitsProcessorWithLoRA, similar to https://github.com/vllm-project/vllm/blob/main/tests/lora/test_layers.py

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -613,8 +626,15 @@ def load_lora_model(model: torch.nn.Module, vllm_config: VllmConfig,
         logger.warning("Regarding multimodal models, vLLM currently "
                        "only supports adding LoRA to language model.")
 
+    env_modules = _parse_lora_module_path_env()
+    if env_modules is not None and vllm_config.lora_config:
+        vllm_config.lora_config.target_modules = env_modules
```
vanbasten23, line 636: I wonder how the "vllm_config.lora_config.target_modules" is used later.

tpu_inference/lora/torch_punica_tpu.py
```diff
@@ -140,8 +140,18 @@ def add_lora_embedding(self,
             lora_b_stacked (torch.Tensor): lora_b's weights.
             add_inputs (bool): Default to True.
         """
-        raise NotImplementedError(
-            "NYI: torch_punica_tpu.PunicaWrapperTPU.add_lora_embedding.")
+        y_orig = y
+        y = y.view(-1, y.shape[-1])
+        x = x.view(-1, x.shape[-1])
+        
+        add_input = kwargs.get('add_input', add_inputs)
```
vanbasten23: nit: one is singular and the other one is plural. Can you unify them?

tpu_inference/lora/torch_punica_tpu.py
```diff
@@ -218,8 +228,20 @@ def add_lora_logits(self,
             scale (float): Scaling factor.
             buffer (Optional[torch.Tensor]):Default to None.
         """
-        raise NotImplementedError(
-            "NYI: torch_punica_tpu.PunicaWrapperTPU.add_lora_logits.")
+
+        # Step 1: Shrink operation
+        buffer = bgmv_shrink(x, lora_a_stacked, self._get_token_lora_indices(x), scale)
```
vanbasten23: this line and line 240, `_get_token_lora_indices(x)` returns `narrow(self._token_lora_indices, 0, 0, x.size(0)) ` the per-token index tensor.` .
 
But https://github.com/vllm-project/vllm/blob/5a0a8fc1ea7542394ff315138bd5677b7b53bca1/vllm/lora/punica_wrapper/punica_gpu.py#L320 seems to suggest to use per-sample tensor (or you can see from an old commit https://github.com/vllm-project/vllm/blame/d1695758b2f65fd314d1aee71ba2469ceba67a5b/vllm/lora/punica_wrapper/punica_gpu.py#L449).

tpu_inference/lora/torch_punica_tpu.py
bhuvanpkaruturi: Done. I have added corresponding tests to `tpu-inference/tests/lora/test_layers.py` file.

tpu_inference/models/vllm/vllm_model_wrapper.py
bhuvanpkaruturi: Added a comment to clarify how this property is used later. Please let me know if this addresses your question

tpu_inference/lora/torch_punica_tpu.py
bhuvanpkaruturi: Thank you. Fixed the typo.

tpu_inference/lora/torch_punica_tpu.py
bhuvanpkaruturi: In TPU case, the `bgmv_*` ops expect a per-token index tensor of shape [num_tokens]: https://github.com/bhuvanpkaruturi/tpu-inference/blob/bf49385f80b68c24f92975ac68b8838e20346fca/tpu_inference/lora/torch_lora_ops.py#L56-L57

vanbasten23: Yea, I think for "per-sample" num_tokens==num_samples.  Maybe worth double check the vllm implementation.

vanbasten23:
Curious how you test the feature e2e

bhuvanpkaruturi:
> Curious how you test the feature e2e

I tested these changes on TPU7 VM. Ran `examples/features/lora/multilora_offline.py` script. Details are in https://b.corp.google.com/issues/499306719#comment8

tpu_inference/lora/torch_punica_tpu.py
bhuvanpkaruturi: Thanks. Looks like we need to use `num_sample` index tensor for logits layer.  Updated the code to add `_get_sampler_indices` func that returns the num_sample index tensor and used it in `add_lora_logits`.

vanbasten23 (approved):
Thanks Bhuvan for the contribution!

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -649,3 +672,45 @@ def _tpu_reset_lora(self, index: int):
             module.set_lora = _tpu_set_lora.__get__(module, module.__class__)
             module.reset_lora = _tpu_reset_lora.__get__(
                 module, module.__class__)
+
+
+def _parse_lora_module_path_env():
+    """Parses LORA_MODULE_PATH env var into vLLM canonical target_modules list."""
+    env_val = os.environ.get("LORA_MODULE_PATH")
```
kyuyeunk: this env and below name mapping. i'm generally not a fan of hard coded string variables. is there a better way to handle this? for instance, we might leverage some code in maxtext that has the dictionary of this exact mapping and import it to use that as source of truth instead.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -649,3 +672,45 @@ def _tpu_reset_lora(self, index: int):
...
+        "wi_0": "gate_proj",
+        "wi_1": "up_proj",
+        "wo": "down_proj",
+        "embed": "embed_tokens",
+        "lm_head": "lm_head"
+    }
+    target_modules = set()
+
+    # Extract regex groups like (query|key|value|out)
+    if '(' in env_val or ')' in env_val:
```
kyuyeunk: similar to above comment, we really want to avoid any string based logic.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -606,6 +608,17 @@ def compute_pooler_output(
         return compute_pooler_output
 
 
+class TPULRUCacheWorkerLoRAManager(LRUCacheWorkerLoRAManager):
```
kyuyeunk: i don't think this class needs to belong in this file? can you create a separate file for it?

kyuyeunk (changes requested):

tpu_inference/models/vllm/vllm_model_wrapper.py
bhuvanpkaruturi: Moved this class and env parsing logic to a separate file: `tpu-inference/tpu_inference/lora/lora_manager.py`

tpu_inference/models/vllm/vllm_model_wrapper.py
bhuvanpkaruturi: Unfortunately, there is not dictionary exposed by MaxText that has the HF to MaxText layer mapping. Also it would mean we have to add MaxText as a dependency of `tpu-inference`. I have added a comment explaining this. b/499306719 requests that we support specifying target LoRA modules through an env variable, so used `LORA_MODULE_PATH` env here.

tpu_inference/models/vllm/vllm_model_wrapper.py
bhuvanpkaruturi: Moved the `regex` into named variables but we still need some string based logic to validate user specific LoRA module path.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import copy
+import os
```
kyuyeunk: why are these needed? i don't see it being used anywhere in the file?

kyuyeunk (approved):
only one minor comment. otherwise it looks good!

tpu_inference/models/vllm/vllm_model_wrapper.py
bhuvanpkaruturi: Ah! Good catch. This was imported earlier to ready the LORA_MODULE_PATH env var but that logic was moved to a different file now. I have remove the unused import (`os`, `re`) from this file now.

PR #2482, opened 2026-05-02
Enable bf16 SSM state in fused GDN kernels by default on TPU

qizzzh:
Adds end-to-end support for storing the GDN recurrent state at bf16 (or fp16) in HBM on TPU and makes bf16 the default. Compute precision is unchanged: the kernels upcast h_bufs to fp32 inside VMEM (per-token in the decode kernel, per-block in the recurrent kernel), so on-chip math runs at fp32 regardless of HBM storage dtype.

   Relax the fp32-only restriction on initial_state in the fused GDN
   decode and recurrent kernels:
   - fused_gdn_kernel_common.py: validate_gdn_inputs accepts fp32/bf16/fp16.
   - fused_gdn_decode_kernel.py: h_bufs scratch and state out_shape match initial_state.dtype; get_default_block_sizes takes a state_dtype param so VMEM budgeting accounts for the smaller scratch.
   - fused_gdn_recurrent_kernel.py: same pattern as decode.
   - fused_gdn_kernel_wrapper.py: drop the unconditional .astype(jnp.float32) cast on initial_state - that cast would materialize an fp32 copy of a bf16 state in HBM and defeat the storage savings. These patches are inert when callers pass fp32 initial_state.

   Opt-out: TPU_MAMBA_SSM_CACHE_DTYPE=float32 vllm serve ...
   Defer to vLLM/HF config: TPU_MAMBA_SSM_CACHE_DTYPE="" vllm serve ...

On Qwen3.5-397B-A17B-FP8 (60 layers: 15 full attn + 45 GDN) with TP=8, halving the SSM state shrinks the mamba page (4.27 -> 2.17 MB/slot), which under vLLM's hybrid-cache attn_page >= mamba_page constraint also shrinks the attn block_size (1088 -> 576) and grows the attention pool with the freed HBM. Empirical comparison from the kv_cache_manager init log, same hardware/command, only TPU_MAMBA_SSM_CACHE_DTYPE differs (--gpu-memory-utilization=0.95 --max-num-seqs=512 --max-model-len=9216 --kv-cache-dtype=fp8 --enable-expert-parallel, TP=8 on tpu7x-8):

  |                       | mamba_dtype          | mamba HBM | attn HBM  | attn blocks/layer |
  |-----------------------|----------------------|-----------|-----------|-------------------|
  | TPU_MAMBA_SSM=float32 | (bfloat16, float32)  | 91.8 GiB  | 253.6 GiB | 2037              |
  | bf16 (default)        | (bfloat16, bfloat16) | 46.7 GiB  | 298.8 GiB | 4533 (+122%)      |

At the conc=512 production bench (1k in / 8k out, num-prompts=640) this yields +15% total throughput (3510 -> 4029 tok/s).

Eval:
mmlu GPU baseline: 0.8186

|mmlu_pro           |      2|custom-extract|      |exact_match|↑  |0.8286|±  |0.0140|
| - biology         |      3|custom-extract|     5|exact_match|↑  |0.9200|±  |0.0388|
| - business        |      3|custom-extract|     5|exact_match|↑  |0.8600|±  |0.0496|
| - chemistry       |      3|custom-extract|     5|exact_match|↑  |0.9200|±  |0.0388|
| - computer_science|      3|custom-extract|     5|exact_match|↑  |0.8400|±  |0.0524|
| - economics       |      3|custom-extract|     5|exact_match|↑  |0.8800|±  |0.0464|
| - engineering     |      3|custom-extract|     5|exact_match|↑  |0.7200|±  |0.0641|
| - health          |      3|custom-extract|     5|exact_match|↑  |0.7400|±  |0.0627|
| - history         |      3|custom-extract|     5|exact_match|↑  |0.6600|±  |0.0677|
| - law             |      3|custom-extract|     5|exact_match|↑  |0.7600|±  |0.0610|
| - math            |      3|custom-extract|     5|exact_match|↑  |0.9600|±  |0.0280|
| - other           |      3|custom-extract|     5|exact_match|↑  |0.7600|±  |0.0610|
| - philosophy      |      3|custom-extract|     5|exact_match|↑  |0.8200|±  |0.0549|
| - physics         |      3|custom-extract|     5|exact_match|↑  |0.9600|±  |0.0280|
| - psychology      |      3|custom-extract|     5|exact_match|↑  |0.8000|±  |0.0571|

|mmlu_pro|      2|custom-extract|      |exact_match|↑  |0.8286|±  | 0.014|

AIME26 GPU baseline: 0.8

|Tasks |Version|Filter|n-shot|  Metric   |   |Value|   |Stderr|
|------|------:|------|-----:|-----------|---|----:|---|-----:|
|aime26|      0|none  |     0|exact_match|↑  |    1|±  |     0|

GPAQ GPU baseline: 0.8434

|          Tasks          |Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-------------------------|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gpqa_diamond_cot_zeroshot|      1|flexible-extract|     0|exact_match|↑  |  0.9|±  |0.0557|
|                         |       |strict-match    |     0|exact_match|↑  |  0.0|±  |0.0000|

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -265,14 +265,8 @@ environment_variables: dict[str, Callable[[], Any]] = {
                      ]),
     "MOE_ALL_GATHER_ACTIVATION_DTYPE":
     lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
-    # Override for cache_config.mamba_ssm_cache_dtype on TPU. Default
-    # "bfloat16" halves mamba SSM state HBM and grows the attention KV pool
-    # ~2x on hybrid models (compute still runs fp32 internally via per-token
-    # / per-block cast in the GDN kernels). Set to "float32" (or "float16")
-    # to opt out, or to "" to defer to whatever vLLM/HF config decides
-    # (typically "float32" for Qwen3.5). At the pinned vLLM LKG,
-    # --mamba-ssm-cache-dtype's CLI parser does not accept "bfloat16", so
-    # this env var is the way to set bf16 without vendoring vLLM.
+    # Override cache_config.mamba_ssm_cache_dtype on TPU. Default "bfloat16"
+    # halves SSM state HBM; set "float32" to opt out, "" to defer to vLLM.
     "TPU_MAMBA_SSM_CACHE_DTYPE":
     lambda: os.getenv("TPU_MAMBA_SSM_CACHE_DTYPE", "bfloat16"),
     # kv offload to dram: skip pre-compiling swap-related jax functions
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -267,6 +267,8 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
     # Override cache_config.mamba_ssm_cache_dtype on TPU. Default "bfloat16"
     # halves SSM state HBM; set "float32" to opt out, "" to defer to vLLM.
+    # TODO: remove once vLLM MambaDType includes bfloat16
+    # (https://github.com/vllm-project/vllm/pull/41680).
     "TPU_MAMBA_SSM_CACHE_DTYPE":
     lambda: os.getenv("TPU_MAMBA_SSM_CACHE_DTYPE", "bfloat16"),
     # kv offload to dram: skip pre-compiling swap-related jax functions
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
     REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES: list[str] = []
     RAGGED_GATED_DELTA_RULE_IMPL: str = "ragged_gated_delta_rule_chunked"
     MOE_ALL_GATHER_ACTIVATION_DTYPE: str = ""
+    TPU_MAMBA_SSM_CACHE_DTYPE: str = "bfloat16"
     TPU_OFFLOAD_SKIP_JAX_PRECOMPILE: bool = False
     TPU_OFFLOAD_DECODE_SAVE: bool = False
     TPU_OFFLOAD_NUM_CPU_CHUNKS: int = 1024
@@ -264,6 +265,16 @@ environment_variables: dict[str, Callable[[], Any]] = {
                      ]),
     "MOE_ALL_GATHER_ACTIVATION_DTYPE":
     lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
+    # Override for cache_config.mamba_ssm_cache_dtype on TPU. Default
+    # "bfloat16" halves mamba SSM state HBM and grows the attention KV pool
+    # ~2x on hybrid models (compute still runs fp32 internally via per-token
+    # / per-block cast in the GDN kernels). Set to "float32" (or "float16")
+    # to opt out, or to "" to defer to whatever vLLM/HF config decides
+    # (typically "float32" for Qwen3.5). At the pinned vLLM LKG,
+    # --mamba-ssm-cache-dtype's CLI parser does not accept "bfloat16", so
+    # this env var is the way to set bf16 without vendoring vLLM.
+    "TPU_MAMBA_SSM_CACHE_DTYPE":
+    lambda: os.getenv("TPU_MAMBA_SSM_CACHE_DTYPE", "bfloat16"),
     # kv offload to dram: skip pre-compiling swap-related jax functions
     "TPU_OFFLOAD_SKIP_JAX_PRECOMPILE":
     lambda: bool(int(os.getenv("TPU_OFFLOAD_SKIP_JAX_PRECOMPILE", "0"))),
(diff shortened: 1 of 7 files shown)


tpu_inference/kernels/gdn/fused_gdn_kernel_common.py
```diff
@@ -89,9 +89,14 @@ def validate_gdn_inputs(
                          f"k={k.dtype}, v={v.dtype}")
     if g.dtype != jnp.float32:
         raise ValueError(f"g must be float32, got {g.dtype}")
-    if initial_state.dtype != jnp.float32:
+    # Storage may be fp32, bf16, or fp16; compute is always fp32 on-chip via
```
jrplatin: I don't think this comment is needed

tpu_inference/platforms/tpu_platform.py
```diff
@@ -247,6 +247,27 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
                 f"Using KV cache block size: {cache_config.block_size}")
 
+        # Override mamba_ssm_cache_dtype via env var. Defaults to "bfloat16"
+        # (see envs.py): halves SSM state HBM and ~2x's the attention KV
+        # pool on hybrid models. Set TPU_MAMBA_SSM_CACHE_DTYPE=float32 to
+        # opt out, or TPU_MAMBA_SSM_CACHE_DTYPE="" to defer to whatever
+        # vLLM/HF config decided. At the pinned vLLM LKG,
+        # --mamba-ssm-cache-dtype's MambaDType Literal does not include
+        # "bfloat16", but CacheConfig is a pydantic dataclass without
+        # validate_assignment, so programmatic assignment bypasses
```
jrplatin: I think this is probably a bit much as well -- fairly obvious from the code IMO

tpu_inference/envs.py
```diff
@@ -264,6 +265,16 @@ def _get_str_list_env() -> list[str]:
...
     lambda: os.getenv("MOE_ALL_GATHER_ACTIVATION_DTYPE", ""),
+    # Override for cache_config.mamba_ssm_cache_dtype on TPU. Default
+    # "bfloat16" halves mamba SSM state HBM and grows the attention KV pool
+    # ~2x on hybrid models (compute still runs fp32 internally via per-token
+    # / per-block cast in the GDN kernels). Set to "float32" (or "float16")
+    # to opt out, or to "" to defer to whatever vLLM/HF config decides
+    # (typically "float32" for Qwen3.5). At the pinned vLLM LKG,
+    # --mamba-ssm-cache-dtype's CLI parser does not accept "bfloat16", so
+    # this env var is the way to set bf16 without vendoring vLLM.
+    "TPU_MAMBA_SSM_CACHE_DTYPE":
```
helloworld1, line 272: I wonder why do we need to override to bf16 dtype? Can we just use supported fp16 type which is better at avoiding underflow that causes cumulative state errors.

qizzzh: bf16 is better at avoiding underflow than fp16 because of more exponent bits. But I can try fp16 and see what we get.

yaochengji:
Thanks for adding more dtypes support, could you please add unit test in tests/kernels/fused_gdn_kernel_test.py?

tpu_inference/envs.py
helloworld1: More exponents can avoid overflow. The underflow happens when the state updates are small and fp16 has more mantissa bits to capture that. We should test fp16 and if it works, we can remove this override in envs.py and, instead update vLLM upstream to support bf16 instead.

qizzzh: I'm running fp16 evals (will share results when done). One concern: the kernels compute state in fp32 then write back to the state tensor, and JAX already emits a FutureWarning for the fp32→fp16 scatter ("cannot safely cast... will result in an error in future JAX releases"). 
 
 bf16 avoids this since it shares fp32's exponent range — no overflow/underflow risk on the cast. fp16's extra mantissa bits help precision but the narrower dynamic range ([6e-8, 65504]) is a correctness risk for accumulated SSM state. Happy to share the fp16 eval numbers either way, but I think bf16 is the safer default for this reason. We should upstream bf16 support to vLLM's MambaDType so we can drop the env override but I'd like to do it in two steps - land the change in tpu-inference first to unblock perf work; meanwhile make the upstream change which may take some time.

qizzzh: It appears in the 3 evals these two are about the same. However, I think bf16 is the safer default: the kernels compute in fp32 and write back to the state buffer, and fp32→fp16 triggers a JAX FutureWarning ("cannot safely cast... will result in an error in future JAX releases") due to fp16's narrower exponent range ([6e-8, 65504]). bf16 shares fp32's exponent range so the cast is always safe. Agreed we should upstream bf16 to vLLM's MambaDType to remove the env override long-term.

tpu_inference/platforms/tpu_platform.py
qizzzh: Updated

tpu_inference/kernels/gdn/fused_gdn_kernel_common.py
qizzzh: Updated

qizzzh:
> Thanks for adding more dtypes support, could you please add unit test in tests/kernels/fused_gdn_kernel_test.py?

Added

tpu_inference/envs.py
helloworld1: Please add TODO here and link to the GitHub issue / PR to update MambaDType to add BF16 since this change should not be specific to TPU only.

qizzzh: I've created an upstream PR and updated the comment there.

helloworld1 (approved):

PR #2484, opened 2026-05-03
[FIX][TPU Offloading] avoid recompile when `MODEL_IMPL_TYPE=vllm`

juncgu-google:
# Description

The original `update_kv_caches_one` drives kv_cache sharding_spec from input kv_caches, whose sharding_spec may be different from the warmup phase due to buffer donation. It leads to recompile `update_kv_caches` at runtime  (when MODEL_IMPL_TYPE=vllm). 

This PR records kv_cache sharding_spec when registering runner and use it in `update_kv_caches_one`.

# Tests

`MODEL_IMPL_TYPE=vllm pytest -sv tests/offload/tpu_offload_accuracy_test.py`

# Checklist

### DIFF
--- a/.buildkite/features/KV_Cache_Offload.yml
+++ b/.buildkite/features/KV_Cache_Offload.yml
@@ -20,7 +20,6 @@ steps:
     soft_fail: true
     env:
       USE_V6E8_QUEUE: "True"
-      VLLM_LOG_LEVEL: "INFO"
     agents:
       queue: tpu_v6e_8_queue
     commands:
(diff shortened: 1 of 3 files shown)


.buildkite/features/KV_Cache_Offload.yml
```diff
@@ -20,7 +20,6 @@ steps:
     soft_fail: true
     env:
       USE_V6E8_QUEUE: "True"
-      VLLM_LOG_LEVEL: "INFO"
```
richardsliu, line 23: Is this needed?

richardsliu (approved):

.buildkite/features/KV_Cache_Offload.yml
juncgu-google: we do not need to set the logging level.

PR #2486, opened 2026-05-03
Fix performance tests broken by upstream

richardsliu:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tests/e2e/benchmarking/benchmark.sh
+++ b/tests/e2e/benchmarking/benchmark.sh
@@ -225,6 +225,8 @@ else
     exit 1
 fi
 
+# Disable weight tracking
+extra_serve_args+=(--model-loader-extra-config '{"enable_weights_track": false}')
 
 # Spin up the vLLM server
 echo "Spinning up the vLLM server..."


QiliangCui (approved):

tests/e2e/benchmarking/benchmark.sh
```diff
@@ -225,6 +225,8 @@ else
     exit 1
 fi
 
+# Disable weight tracking
+extra_serve_args+=(--model-loader-extra-config '{"enable_weights_track": false}')
```
QiliangCui, line 229: @jrplatin , 

this may fix the test but i'm not sure if this hide some real issue..

QiliangCui: per what i see, it seems impacting only Jax path. @jrplatin , could you help to triage if this fix is safe or we should fix the error of

```
AttributeError: 'Qwen3ForCausalLM' object has no attribute 'named_modules'. Did you mean: 'iter_modules'?

```

richardsliu: This is the original upstream change: https://github.com/vllm-project/vllm/commit/99255f3cb5cec7466bf9fa5310fd310baf87d711

PR #2489, opened 2026-05-03
[Speculative decoding] Jit the whole "propose()" and prepare_inputs() method of the drafter

gxd3:
# Description
jit the whole `propose()`  and `prepare_inputs()` method of the drafter for performance improvement.

Before this PR, there is gap between each drafter's step:
https://screenshot.googleplex.com/9VTrwkgDM8vmoTU
(gap between 2 consecutive target model's step: ~18ms: https://screenshot.googleplex.com/BJnqJpEb2xTAM8W)

After this PR, gamma number of drafter's step is a single jited-fn:
https://screenshot.googleplex.com/ADXRhTviByMUdkV
(gap between 2 consecutive target model's step: ~12ms)

# Tests
Spec decoding acceptance rate: https://paste.googleplex.com/5224862240079872
is on par with previous test record in https://screenshot.googleplex.com/yeoqKrZVYGDEgUY

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -64,6 +64,17 @@ class CompilationManager:
                              dtype: Any,
                              sharding: Optional[NamedSharding] = None) -> Any:
         """Helper to create dummy tensors for precompilation."""
+        if len(shape) > 1:
+            # Use parallel_config as the primary source of truth during initialization
+            # to avoid AttributeError when self.runner.mesh is not yet assigned (common in unit tests).
+            tp_size = 1
+            if self.runner.vllm_config.parallel_config is not None:
+                tp_size = self.runner.vllm_config.parallel_config.tensor_parallel_size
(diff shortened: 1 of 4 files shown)


tpu_inference/runner/compilation_manager.py
```diff
@@ -839,7 +803,8 @@ def _precompile_eagle3_helpers(self) -> None:
                                           dtype),
             ]
 
-            positions = self._create_dummy_tensor((num_tokens, ), jnp.int32)
+            positions = self._create_dummy_tensor((num_tokens, ), jnp.int32,
+                                                  data_parallel_attn_sharding)
```
Lumosis: Why do we need to add dp sharding here?

gxd3: I was trying to fix the precompiling check before.
The sharding change is not necessary, removed.

But to make things simiple for the precompilion, i make `prepare_inputs()` a jited fn as well, PTAL.

PR #2500, opened 2026-05-04
Add Hierarchical Reduce-Scatter kernel 

dawnhan1111:
# Description

The change including the following features:

A base function for accumulation.
Intra-chip accumulation.
Inter-chip accumulation.
Integrate the intra-chip reduction with the inter-chip reduction and scatter steps to form the complete algorithm.

We plan to enable this custom kernel for reduce scatter for DSv3 for now, this feature default to be disabled.

# Tests
Ran e2e test on dsv3 single-host and multi-host devices. The custom kernel with op name `pallas-call` shows up instead of `all-reduce` with the following expression:
```
shard_map.13019 = (bf16[512,7168]{1,0:T(16,128)(2,1)S(1)}, bf16[8192,7168]{1,0:T(8,128)(2,1)}, bf16[8192,7168]{1,0:T(8,128)(2,1)}) custom-call(bf16[8192,7168]{1,0:T(8,128)(2,1)} %fusion.14), custom_call_target="tpu_custom_call", operand_layout_constraints={bf16[8192,7168]{1,0}}, frontend_attributes={kernel_metadata={ "mesh_axes":"[\"attn_dp\", \"attn_dp_expert\", \"dcp\", \"expert\", \"model\"]" }}
```

### DIFF
--- a/tpu_inference/kernels/collectives/hierarchical_reduce_scatter.py
+++ b/tpu_inference/kernels/collectives/hierarchical_reduce_scatter.py
@@ -1,0 +1,448 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Hierarchical Recursive Doubling Reduce-Scatter Implementation."""
+
+import math
+
+import jax
+from jax.experimental import pallas as pl
+from jax.experimental import shard_map
+from jax.experimental.pallas import tpu as pltpu
+
+
+def _next_multiple_of(val, multiple):
+    return ((val + multiple - 1) // multiple) * multiple
+
+
+# Asynchronously accumulates received data into a running sum using VMEM buffers.
+def _accumulate(
+    recv_hbm,
+    run_hbm,
+    out_hbm,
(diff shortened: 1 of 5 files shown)


tpu_inference/kernels/collectives/hierarchical_reduce_scatter.py
```diff
@@ -0,0 +1,445 @@
...
+                pl.ds(mb_start, mb_size_actual),
+            ],
+            vmem_recv=vmem_recv_ref.at[
+                :, pl.ds(0, final_chunk_size), pl.ds(0, mb_size_actual)
+            ],
+            vmem_run=vmem_run_ref.at[
+                :, pl.ds(0, final_chunk_size), pl.ds(0, mb_size_actual)
+            ],
+            sync_sems=sync_sems,
+            vmem_idx=pair_idx % vmem_pipeline_depth,
```
Caslyn: Since `vmem_idx` can wrap around, are we at risk of reusing/overwriting the same `vmem_idx` value in this loop?

Caslyn (approved):

tpu_inference/kernels/collectives/hierarchical_reduce_scatter.py
dawnhan1111: Yes, you are correct. If we scale to a large enough number of chips (num_chips > 4), we will wrap around and risk overwriting VMEM. Since we are currently only testing on topologies with 8 devices or fewer (num_chips <= 4), we aren't hitting this case yet.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -316,7 +362,12 @@ def expert_parallel_gmm(
             data_p_spec,
             data_p_spec,
         ),
-        out_specs=(data_p_spec),
+        out_specs=P(
+            (ShardingAxisName.MLP_DATA, ) +
```
gxd3, line 366: if enable_rs_kernel:
should it be `P(ShardingAxisName.EXPERT)`?

PR #2477, opened 2026-05-02
Add tpu field into ables to track tpu config that kernel tuning uses

patrickji2014:
# Description

In this pr, we open source the ddl which shows the schema of the spanner tables we use to store tuning result/status during the tuning process. At the same time, a field call TPU is used to keep track of the environment where the kernel tuning happens. This will support us to run different kernel tuning in different generation of TPUs. 

# Tests

Validate through local run and BuildKite pipeline build.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2501, opened 2026-05-04
[Dont Review]Support Kernel Tuning to Yield TPU VM to Higher Priority Jobs

patrickji2014:
# Description

Existing Kernel Job are scheduled as a single bucket where a bucket contains many kernel tuning cases, like 1000. But with the current setup, the kernel job will occupied the TPU VM during the measurement of the 1000 tuning cases, which might block other higher priority jobs for a long time if they are waiting. 

This changes will make the kernel tuning job stop after running for x minutes, for example 20 minutes, and schedule another job for the reset of cases in the bucket. 

# Tests

Test using the kernel tuning pipeline and local tuning mode.

# Checklist


tools/kernel/tuner/v1/storage_management/spanner_ddl/case_results.ddl
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE
+  CaseResults ( ID STRING(128) NOT NULL,
```
QiliangCui, line 2: so, the DDL is intentionally uploaded right? 

If so, let's mention this in description.

.buildkite/pipeline_kernel_tuning.yml
QiliangCui, line 83: Did you verify this really works?

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -170,91 +199,143 @@ def _generate_tuning_jobs(
...
+                    f'  --case_set_desc=\"{self.case_set_desc}\" '
+                    f'  --run_locally=False '
+                    f'  --tpu_queue_multi={self.tpu_queue_multi} '
+                    f'  --begin_case_id={case_id_start} --end_case_id={case_id_end}\''
+                ),
+                LiteralString(
+                    f'if [ -f /tmp/kernel_tuning/generated_pipeline.yml ]; then '
+                    f'  buildkite-agent artifact upload /tmp/kernel_tuning/generated_pipeline.yml && '
+                    f'  echo \"Upload generated pipeline YAML to Buildkite artifacts with priority {self.kernel_tuning_job_priority}\" && '
+                    f'  {{ '
```
QiliangCui: did you verify this really works? 

please have a try..

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -427,9 +524,11 @@ def measure_latency(self, caseset_id: str, run_id: str, begin_case_id: int,
...
-                                                   bucket_id,
-                                                   bucket_total_time_us)
+        self.storage_manager.add_bucket_processed_time_us(
+            self.case_set_id, self.run_id, bucket_id, bucket_total_time_us)
+        if bucket_fully_processed:
+            self.storage_manager.mark_bucket_completed(self.case_set_id,
+                                                       self.run_id, bucket_id)
         logger.info(
-            f"Worker [{FLAGS.worker_id}] Completed Bucket {bucket_id} ({begin_case_id}-{end_case_id}) for CaseSetId: {caseset_id}, RunId: {run_id}. Total time: {bucket_total_time_us/1e6:.2f}s."
+            f"Worker [{FLAGS.worker_id}] Completed Bucket {bucket_id} [{begin_case_id}-{last_processed_case_id + 1}) for CaseSetId: {self.case_set_id}, RunId: {self.run_id}. Total time: {bucket_total_time_us/1e6:.2f}s."
```
QiliangCui2023: `last_processed_case_id` is initialized to `None` at line 445 and only assigned to `cid` *after* the time-budget break check (line 462). If the loop breaks on the very first iteration (e.g. on a slow restart where elapsed already exceeds the budget), this format string evaluates `last_processed_case_id + 1` and raises `TypeError: NoneType + int`.

Fix: initialize `last_processed_case_id = begin_case_id - 1`, or guard the f-string with a conditional.

.buildkite/scripts/run_in_docker.sh
```diff
@@ -56,6 +56,8 @@ ENV_VARS=(
   -e KERNEL_TUNING_KERNEL_NAME="${KERNEL_TUNING_KERNEL_NAME:-}"
   -e KERNEL_TUNING_CASE_SET_DESC="${KERNEL_TUNING_CASE_SET_DESC:-}"
   -e KERNEL_TUNING_TPU_VERSION="${KERNEL_TUNING_TPU_VERSION:-}"
+  -e KERNEL_TUNING_TPU_CORES="${KERNEL_TUNING_TPU_CORES:-}"
+  -e KERNEL_TUNING_JOB_PRIORITY="${PRIORITY_KERNEL_TUNING:-10}"
```
QiliangCui2023, line 60: I see in https://github.com/vllm-project/tpu-inference/blob/8fb63ce5a2cf6be377cca811b4027d9ecedb2a84/.buildkite/scripts/configs/pipeline_config.sh#L24

it is "-1" by default. should we align this value?

QiliangCui2023:
**Scope / reviewability**

The 524-line diff bundles three loosely related changes:

1. **Time-yield scheduling** (the actual feature: `measure_latency` budget check, `generate_buildkite_pipeline_subbucket`, `add_bucket_processed_time_us`) — ~150 lines.
2. **New `TPU` column** on `KernelTuningCases` / `WorkBuckets` / `CaseResults` plus 4 new `.ddl` files. Unrelated to scheduling — this is a separate "track which TPU type ran the case" feature.
3. **Constructor signature overhaul** of `KernelTunerBase` (move `case_set_id` / `run_id` / `case_set_desc` / `tpu_*` / `run_locally` / priority into `__init__`). Mechanical churn touching every subclass and every internal call site.

Please split into ≥2 PRs — at minimum, separate the schema/refactor changes from the scheduler change so each can be reviewed line-by-line and bisected independently if a regression shows up.

Also consider squashing the 12 commits before merge — several are pure debug/log fixups (`add logging`, `update log message`, `remove debug condition…`, `fix the command format issue`, `fix kernel tuning group name`).

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -95,20 +106,41 @@ def __init__(self,
...
-
-    def _init_case_set(self, case_set_id: str, desc: str) -> bool:
+        self.case_set_id = case_set_id
+        self.run_id = run_id
+        self.case_set_desc = case_set_desc
+        self.tpu_queue_multi = tpu_queue_multi
+        self.tpu_version = tpu_version
+        self.tpu_cores = tpu_cores
+        self.run_locally = run_locally
+        self.run_at_most_minutes = 20
```
QiliangCui2023, line 147: `self.run_at_most_minutes = 20` is hard-coded here. The PR description says "stop after running for x minutes, for example 20 minutes" — implying it's tunable. As-is, no operator can change the budget without editing the base class.

Promote it to a flag (mirroring `_KERNEL_TUNING_JOB_PRIORITY`) or to a constructor kwarg.

tools/kernel/tuner/v1/storage_management/spanner_database_manager.py
```diff
@@ -202,20 +213,51 @@ def mark_bucket_in_progress(self, cs_id, r_id, b_id):
                 'wid': spanner.param_types.STRING
             }))
 
-    def mark_bucket_completed(self, cs_id, r_id, b_id, tt_us):
+    def mark_bucket_completed(self, cs_id, r_id, b_id, tt_us=None):
+        if tt_us is not None:
```
QiliangCui2023, line 217: After this PR no caller passes `tt_us` to `mark_bucket_completed` — the new `add_bucket_processed_time_us` accumulates time incrementally, and the only call site in `kernel_tuner_base.py` is `mark_bucket_completed(self.case_set_id, self.run_id, bucket_id)` (no `tt_us`). The `if tt_us is not None` branch is dead.

Either drop the branch entirely, or collapse both branches into one `UPDATE` using `COALESCE(@tt, TotalTime)` so the dual signature stays.

.buildkite/scripts/run_in_docker.sh
QiliangCui2023, line 60: The build-level env var is `PRIORITY_KERNEL_TUNING` but it gets re-exported into the container as `KERNEL_TUNING_JOB_PRIORITY`. Same value, two names, no semantic reason for the rename.

Pick one — preferably `KERNEL_TUNING_JOB_PRIORITY` for consistency with the other `KERNEL_TUNING_*` vars in this same block.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -95,20 +106,41 @@ def __init__(self,
...
+        self.kernel_tuning_job_priority = kernel_tuning_job_priority
+
+    def _init_case_set(self) -> bool:
```
QiliangCui2023, line 150: `_init_case_set` no longer takes arguments (it now reads `self.case_set_id` / `self.case_set_desc`), but the docstring still has an `Args:` section listing `case_set_id` and `desc`. Please drop or update.

QiliangCui2023 (changes requested):

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: yes, this works.

patrickji2014: you can see it from https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/69/canvas, the original job yield another job and trigger another upload, leading to another run

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: this should not happen as the first round is always executed. the time it spends so far is just for
"
time_elapsed_minutes = (time.perf_counter() -
                                    bucket_start_perf) / 60
            logger.info(
                f"Worker [{FLAGS.worker_id}] Processing CaseId: {cid} in Bucket {bucket_id}, [{begin_case_id}-{end_case_id}) with elapsed time {time_elapsed_minutes:.2f} minutes."
            )
            if not self.run_locally and (time_elapsed_minutes
                                         > self.max_execution_minutes):

"
which should be always < 20minutes

.buildkite/scripts/run_in_docker.sh
patrickji2014: update to -10 to make it consistent.

tools/kernel/tuner/v1/storage_management/spanner_ddl/case_results.ddl
patrickji2014: these files are uploaded in base PR https://github.com/vllm-project/tpu-inference/pull/2477

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: thanks. make it a flag.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: remove deprecated args in the comment. thanks.

tools/kernel/tuner/v1/storage_management/spanner_database_manager.py
patrickji2014: thanks. tt_us is dropped from the args list now.

.buildkite/pipeline_kernel_tuning.yml
patrickji2014: yes, these are all verified in buildkite. https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/69/canvas

.buildkite/scripts/run_in_docker.sh
patrickji2014: PRIORITY_KERNEL_TUNING is to align the naming with other PRIORITY_xxx_xx in the pieline_config.sh, KERNEL_TUNING_JOB_PRIORITY is to align the naming with other existing KERNEL_TUNING_xxx flags so the inconsistent between the two is hard resolved. But I think it's okay since PRIORITY_KERNEL_TUNING or the priority of the job is not intended to exposed to users.

PR #2503, opened 2026-05-04
NVFP4 dequant-in-VMEM

QiliangCui:
## Summary

Adds NVFP4 support.  Specifically, we add support for both the 1D-Blockwise scale and the (scalar) global scale, which we fold into the 1D-Blockwise scale.  Because we dequantize the weights in the quantized matmul and GMMv2, we can safely ignore the input scale.

### Quantization method: `tpu_inference/layers/vllm/quantization/nvfp4.py` (new)

- `VllmNvfp4Config`: subclasses upstream `ModelOptNvFp4Config` for free config parsing; registers `modelopt_fp4` with the TPU quantization registry; dispatches linear / MoE / attention layers to TPU-specific methods.
- `VllmNvfp4LinearMethod`: extends `VllmUnquantizedLinearMethod`. `create_weights` delegates to upstream `ModelOptNvFp4LinearMethod` for parameter registration; we override `weight_loader` on `input_scale` / `weight_scale_2` with a scalar-fill loader because upstream's `PerTensorScaleParameter` shape `(len(output_partition_sizes),)` mismatches the per-shard scalar in the checkpoint and would trip `linear.py`'s `assert param_data.shape == loaded_weight.shape` in `QKVParallelLinear`. `process_weights_after_loading` unpacks the FP4 weight and combines the two-level scales into a single FP32 blockwise scale; `apply` routes to `sharded_quantized_matmul` for on-the-fly dequant.
- `VllmNvfp4MoEMethod`: `create_weights` delegates to upstream `ModelOptNvFp4FusedMoE` for parameter registration, additionally registers bias (upstream omits), and re-applies `quant_method` attributes to the scale params. `process_weights_after_loading` keeps the FP4 weight as packed E2M1, combines the two-level scales into a single FP32 blockwise scale, and lays out for `process_moe_weights`. `apply_monolithic` routes through `vllm_moe_apply` → `gmm_v2`.

### HumanEval + MBPP results

CMDS:

```
# For MODEL in [Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 , NVFP4/Qwen3-Coder-30B-A3B-Instruct-FP4
DISABLE_WEIGHT_REQUANTIZATION=1 MODEL_IMPL_TYPE=vllm nohup  vllm serve --seed 42 --model MODEL --max-num-batched-tokens 1028 --max-num-seqs 128 --no-enable-prefix-caching --tensor-parallel-size 2 --enable-expert-parallel  &


# For MODEL in [Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 , NVFP4/Qwen3-Coder-30B-A3B-Instruct-FP4
 nohup evalplus.evaluate --model MODEL --dataset humaneval --backend openai -base-url http://localhost:8000/v1 --greedy  &
```

| Benchmark | NVFFP4 Model (pass@1) | FP8 Model (pass@1) |
| :--- | :---: | :---: |
| **humaneval** (base tests) | 0.921 | 0.921 |
| **humaneval+** (base + extra tests) | 0.884 | 0.872 |
| **mbpp** (base tests) | 0.876 | 0.870 |
| **mbpp+** (base + extra tests) | 0.749 | 0.728 |

### Primary: `nvidia/Qwen3-30B-A3B-NVFP4` (TP=8, v7x) — https://buildkite.com/tpu-commons/tpu-inference-dev/builds/380

| Gate | Result | Threshold |
|---|---|---|
| Unit tests (`tests/layers/vllm/test_nvfp4.py`) | PASSED | — |
| lm_eval gsm8k accuracy | 0.8711 | ≥ 0.80 |
| Request throughput | 6.02 req/s | ≥ 5.00 |
| Median TPOT | 19.56 ms | — |

### Generic-support validation on two additional NVFP4 checkpoints

To verify the dispatch is generic across architectures, ran lm_eval gsm8k_cot accuracy + benchmark on two more NVFP4 models (TP=8, v7x):

| Model | Accuracy (flex / strict) | Throughput | Output token tput | Median TPOT |
|---|---|---|---|---|
| `nvidia/Qwen3.5-397B-A17B-NVFP4` | 0.98 / 0.96 | 0.09 req/s | 629.29 tok/s | 94.04 ms |
| `nvidia/Qwen3-Coder-480B-A35B-Instruct-NVFP4` | 0.9052 / 0.8484 | 0.06 req/s | 476.69 tok/s | 128.57 ms |

Accuracy: https://buildkite.com/tpu-commons/tpu-inference-dev/builds/380.
Performance: https://buildkite.com/tpu-commons/tpu-inference-dev/builds/382.

Both extra models load and serve correctly with no code changes — confirms NVFP4 support is generic.

Signed-off-by: Qiliang Cui <derrhein@gmail.com>


.buildkite/scripts/test_nvfp4_serve.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+# Copyright 2026 Google LLC
```
QiliangCui: will be removed later. test only.

.buildkite/pipeline_dev.yml
```diff
@@ -12,34 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
QiliangCui, line 13: will be reverted. this is for testing only.

.buildkite/models/nvidia_Qwen3-30B-A3B-NVFP4.yml
jrplatin, line 1: I actually don't think we need this -- the _model_ itself is already supported, it's just the quantization method, but maybe it would be good to have this in the support matrix?

QiliangCui2023: Before removing this file and declaring support, let me do a test to run this on 

nvidia/Qwen3-Coder-480B-A35B-Instruct-NVFP4
nvidia/Qwen3.5-397B-A17B-NVFP4

QiliangCui2023: Good catch — addressed in commit `debda256f`. The per-model file is removed and replaced with a per-scheme entry under `.buildkite/quantization/NVFP4_W4A16.yml` (mirrors the existing `FP4_W4A16.yml` stub). Static rows added to `support_matrices/release/{v6e,v7x}/vllm/quantization_support_matrix.csv` and the dispatch hardcoded in `generate_support_matrices.sh`'s `get_tpu_generation` / `get_quantization_method` so the README "Quantization Methods" table will pick up an NVFP4 W4A16 row on the next nightly.

.buildkite/scripts/generate_support_matrices.sh
jrplatin, line 1: I think these are filled in automatically? @jcyang43

QiliangCui2023: I see this file is manually updated. my understanding is that this script is used to generate other support matrices in README.md.

jrplatin (approved):

.buildkite/scripts/generate_support_matrices.sh
jcyang43: Thanks Jacob & Qiliang. Yes, this script needs to be updated to include the new quantization method. 

But Jacob you're right that the support matrices csv files don't need to be updated as they will be populated based on results from CI

support_matrices/release/v6e/vllm/quantization_support_matrix.csv
```diff
@@ -1,5 +1,6 @@
 Quantization dtype,Quantization methods,Recommended TPU Generations,CorrectnessTest,PerformanceTest
 "FP4 W4A16",mxfp4,v7,❓ Untested,❓ Untested
+"NVFP4 W4A16",modelopt_fp4,v7,❓ Untested,❓ Untested
```
jcyang43: Qiliang, we don't need to update these csv files manually. Thanks

support_matrices/release/v7x/vllm/quantization_support_matrix.csv
```diff
@@ -1,5 +1,6 @@
 Quantization dtype,Quantization methods,Recommended TPU Generations,CorrectnessTest,PerformanceTest
 "FP4 W4A16",mxfp4,v7,❓ Untested,❓ Untested
+"NVFP4 W4A16",modelopt_fp4,v7,❓ Untested,❓ Untested
```
jcyang43: Qiliang, we don't need to update these csv files manually. Thanks

support_matrices/release/v6e/vllm/quantization_support_matrix.csv
QiliangCui2023: Reverted in f0e4fe4a4. Kept the generate_support_matrices.sh dispatch entry so CI can populate the CSV automatically once the NVFP4_W4A16 stub runs.

support_matrices/release/v7x/vllm/quantization_support_matrix.csv
QiliangCui2023: Reverted in f0e4fe4a4. Kept the generate_support_matrices.sh dispatch entry so CI can populate the CSV automatically once the NVFP4_W4A16 stub runs.

jcyang43 (approved):

tpu_inference/layers/vllm/quantization/nvfp4.py
```diff
@@ -0,0 +1,427 @@
...
+            set_weight_attrs(
+                getattr(layer, name),
+                {"quant_method": FusedMoeWeightScaleSupported.BLOCK.value})
+        for name in ("w13_weight_scale_2", "w2_weight_scale_2"):
+            set_weight_attrs(
+                getattr(layer, name),
+                {"quant_method": FusedMoeWeightScaleSupported.TENSOR.value})
+        # Upstream NVFP4 MoE does not register bias; some models (e.g.
+        # gpt-oss) need it. Keep our own bias registration.
+        if self.moe.has_bias:
```
kyuyeunk, line 320: does it mean that gpt-oss nvfp4 will simply fail in gpu case?

tpu_inference/layers/vllm/quantization/nvfp4.py
```diff
@@ -0,0 +1,427 @@
...
+        self.use_global_sf = False
+        self.extra_backend_kwargs = {}
+        if self.moe_backend == MoEBackend.FUSED_MOE:
+            self.extra_backend_kwargs = dict(ep_axis_name=ep_axis_name)
+
+    @property
+    def is_monolithic(self):
+        return True
+
+    def create_weights(self, layer, num_experts, hidden_size,
```
kyuyeunk, line 295: can you explain the reason for creating a new `create_weights` function instead of re-using the upstream function?

kyuyeunk (commented):
it looks mostly good. only few minor comments.

tpu_inference/layers/vllm/quantization/nvfp4.py
jrplatin: I believe this is addressed [here](https://github.com/vllm-project/tpu-inference/blob/f0e4fe4a48d91233230ea3db5c792158825abd69/tpu_inference/layers/vllm/quantization/nvfp4.py#L304)

@QiliangCui Can you comment more on this since this is your logic?

tpu_inference/layers/vllm/quantization/nvfp4.py
jrplatin: Appears that way to me -- I checked the upstream [code](https://github.com/vllm-project/vllm/blob/10ebb40d62e024116c1a4473f8e357a3e72761ed/vllm/model_executor/layers/quantization/modelopt.py#L1258) and there's no bias registration

tpu_inference/layers/vllm/quantization/nvfp4.py
QiliangCui2023: We do reuse upstream — line 299 delegates to `ModelOptNvFp4FusedMoE.create_weights(self, layer, ...)`. The override only adds two patches that have to come after that call:

1. L304-317: explicitly set `quant_method` on `w*_weight_scale` / `w*_weight_scale_2`. Upstream registers the scale params before populating `extra_weight_attrs["quant_method"]`, so vLLM`s MoE loader rejects the load (`fused_moe/layer.py:1366`: "quant method must be one of ..."). The comment block at L304-307 explains this.
2. L318-333: register `w13_bias` / `w2_bias`. Upstream`s NVFP4 MoE doesn`t register them; models with MoE bias (e.g. gpt-oss) would otherwise silently load with zero bias.

Both are upstream gaps worth fixing in `modelopt.py` separately; for this PR we patch them locally so TPU works against the current vLLM LKG.

tpu_inference/layers/vllm/quantization/nvfp4.py
QiliangCui2023: Yes — upstream `ModelOptNvFp4FusedMoE.create_weights` doesn`t register `w13_bias` / `w2_bias`, so a gpt-oss-style NVFP4 checkpoint with MoE bias would either fail to find the param during loading or load with zero bias on GPU. This PR patches it for TPU. Filing a separate upstream PR to add bias registration in `modelopt.py` is the right follow-up.

kyuyeunk (approved):

PR #2299, opened 2026-04-17
[MultiModal] [Torchax] Wrap mm_embeds in a list for Qwen3-VL and update padding truncation logic

muskansh-google:
# Description

- Refactor the list-wrapping logic for Qwen3-VL multimodal embeddings out of the main wrapper and into a helper function `maybe_wrap_mm_embed_to_list` in `qwen3_vl_patcher.py`.
- Update the padding truncation logic in vllm_model_wrapper. This is needed because for mm_embeds, padding is done at the end of the input, hence is_multimodal mask cannot be used for slicing.

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: #1850

# Tests

Manually tested accuracy and run by serving Qwen3-VL model on Torchax

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2113, opened 2026-04-01
[Multi-Modal] Refactor multimodal_manager to move model specific logic into jax path model

kwang3939:
# Description

Earlier `multimodal_manager` has qwen2.5-vl specific logic to handle `image_grid` and `pixel_values`, causing failure of other models when multimodal data is preprocessed differently.
Refactor by moving the model specific logic to the jax path model code following the design used in `gpu_runner` in vLLM.
Thanks for the suggestions from @wdhongtw !

# Tests

Tested with:
```
python3 -m pytest -s -v tests/models/jax/test_qwen2_5_vl.py
python3 -m pytest -s -v tests/e2e/test_multi_modal_inference.py
pytest -v -s tests/runner/test_multimodal_manager.py
```

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #972, opened 2025-10-29
[Model] Add vision encoder and input embeddings merger warmup for Qwen2.5 VL model

kwang3939:
# Description
Add warm up of Qwen2.5 vl model vision encoder.
Pad the inputs to get_input_embeddings and warmup this embeddings merger.
Pack the multimodal funcs.

# Tests
CI: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/4783

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2504, opened 2026-05-05
[Multimodal] Refactor multimodal_manager and embed_input_ids to align with vLLM

kwang3939:
# Description

In tpu-inference, `MultiModalManager` has `mm_embed` flatten and padding logic causing error in Qwen3-VL torchax path model. This was earlier added to address the recompilation of `embed_input_ids` due to the different shapes of inputs to the function (https://github.com/vllm-project/tpu-inference/pull/972). @muskansh-google created a PR to address it through splitting the merged embeds in torchax path to unblock Qwen3-VL: https://github.com/vllm-project/tpu-inference/pull/2299.

The other way to address is similar to [PR](https://github.com/vllm-project/tpu-inference/pull/2113), this can be refactored to move jax path specific logic from `multimodal_manager` to inside the jax path code, so that `multimodal_manager` code works for both jax and torchax paths.

# Tests

Tested jax path:`python3 -m pytest -s -v tests/e2e/test_multi_modal_inference.py`
Checked with `export JAX_LOG_COMPILES=1` for jax path compilation.

# Checklist


tpu_inference/models/jax/utils/multi_modal_utils.py
```diff
@@ -175,3 +175,22 @@ def merge_multimodal_embeddings(
         (input_ids == placeholder_token_id),
         multimodal_embeddings,
     )
+
+
+def prepare_jax_mm_embeds(mm_embeds: list[jax.Array] | None,
```
wdhongtw: nit: Maybe named `merge_mm_embeds`, or similar name that reflect the what it did. Or add one line of docstring if there is no good name, so it's easier for understanding later.

tpu_inference/models/jax/utils/multi_modal_utils.py
```diff
@@ -175,3 +175,22 @@ def merge_multimodal_embeddings(
...
+                          target_pad_len: int) -> jax.Array | None:
+    if mm_embeds is None or len(mm_embeds) == 0:
+        return None
+
+    flattened_embeds = flatten_embeddings(mm_embeds)
+    if flattened_embeds.shape[0] == 0:
+        return None
```
wdhongtw: Is this check required?

If there is actually a non-empty list of mm_embed and the flatten tensor is empty, should we just return an empty tensor?

wdhongtw (commented):
The overall change looks good to me. But still some questions.

--

Do we really need a separated `jitted_embed_input_ids_fn`?

If we keep using `embed_input_ids_fn`, there is no harm for torchax path.
We just doing some calculation in compilation manager that have no actual benifits.

Also, if we add `jitted_embed_input_ids_fn`, I suppose there is some change in TPUModelRunner that call different embed_input_ids in different branch.

kwang3939:
> The overall change looks good to me. But still some questions.
> 
> --
> 
> Do we really need a separated `jitted_embed_input_ids_fn`?
> 
> If we keep using `embed_input_ids_fn`, there is no harm for torchax path. We just doing some calculation in compilation manager that have no actual benifits.
> 
> Also, if we add `jitted_embed_input_ids_fn`, I suppose there is some change in TPUModelRunner that call different embed_input_ids in different branch.

@wdhongtw Thanks for reviewing! Yeah, makes sense. I also think passing two looks quite duplicate. Just removed the jitted one.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -497,8 +497,6 @@ def embed_input_ids_func(
                         torch_mm_embeds = [torch_view(x) for x in mm_embeds]
                     else:
                         torch_mm_embeds = torch_view(mm_embeds)
-                    assert is_multimodal is not None
-                    torch_mm_embeds = torch_mm_embeds[is_multimodal]
```
lk-chen, line 501: why is this removed? I believe this was for truncation

kwang3939: Yes, it was for truncation. But right now the `mm_embeds` would be list while `is_multimodal` is a jax array.

lk-chen (commented):
LGTM, would like @wdhongtw to take another look

wdhongtw (approved):
Have a look again and test locally.

It looks good and I confirmed that this change is not breaking `google/gemma-4-31B-it`, and it unblocks `Qwen/Qwen3-VL-2B-Thinking`.

PR #2508, opened 2026-05-05
generate the dataset to do decode only benchmark

yiw-wang:
bm workflow will use this script to generate the dataset for decode only benchmark at runtime.

The BM workflow will use the only prompt, generated by this script based on the configurations, to complete the warm up(prefill kv cache) and use the exact same prompt to complete the regular benchmark to mimic the decode only mode. 

### Test

Later PR that call this script within run_bm.sh will show the E2E test logs.

Test this script locally:

```
yiweiw-macbookpro:Desktop yiweiw$ python3 bm.py --input-len 5 --num-prompts 10 --num-distinct 3
[Data Prep] Generating 3 distinct random prompts (Len: 5)...
[Data Prep] Successfully generated Custom JSONL dataset using raw Token IDs.
[Data Prep] Total Lines: 10 | Distinct Request Types: 3
[Data Prep] Saved to /tmp/decode_only_dataset.jsonl
yiweiw-macbookpro:Desktop yiweiw$ cat /tmp/decode_only_dataset.jsonl 
{"prompt_token_ids": [4787, 18609, 43961, 18086, 4788]}
{"prompt_token_ids": [14301, 7103, 28667, 42332, 42913]}
{"prompt_token_ids": [29571, 36332, 34520, 804, 29181]}
{"prompt_token_ids": [4787, 18609, 43961, 18086, 4788]}
{"prompt_token_ids": [14301, 7103, 28667, 42332, 42913]}
{"prompt_token_ids": [29571, 36332, 34520, 804, 29181]}
{"prompt_token_ids": [4787, 18609, 43961, 18086, 4788]}
{"prompt_token_ids": [14301, 7103, 28667, 42332, 42913]}
{"prompt_token_ids": [29571, 36332, 34520, 804, 29181]}
{"prompt_token_ids": [4787, 18609, 43961, 18086, 4788]}
```

The later change within `run_bm.sh` will use [--disable-shuffle](https://docs.vllm.ai/en/stable/cli/bench/serve/#-disable-shuffle) to guarantee the requests within the same batch will be different to maximize the HBM workload but these will be all cached to mimic the decode-only scenario.

### DIFF
--- a/.buildkite/benchmark/scripts/generate_decode_only_bm_dataset.py
+++ b/.buildkite/benchmark/scripts/generate_decode_only_bm_dataset.py
@@ -1,0 +1,79 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import json
+
+import numpy as np
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Generate decode-only JSONL dataset for vLLM benchmark")
+    parser.add_argument("--input-len",
+                        required=True,
+                        type=int,
+                        help="Exact number of input tokens")
+    parser.add_argument(
+        "--num-prompts",
+        type=int,
+        default=5000,
+        help=
+        "Total number of prompts to generate. Must be > benchmark requests to avoid vLLM oversampling.",
+    )
+    parser.add_argument(
+        "--num-distinct",
+        required=True,
+        type=int,
+        help=
+        "Suggested to be equal to --max-num-seqs to maximize the randomness while ensuring all requests can be safely cached without eviction.",
+    )
+    parser.add_argument(
+        "--output-file",
+        type=str,
+        default="/tmp/decode_only_dataset.jsonl",
+        help="Path to save the JSONL",
+    )
+    args = parser.parse_args()
+
+    print(
+        f"[Data Prep] Generating {args.num_distinct} distinct random prompts (Len: {args.input_len})..."
+    )
+
+    # Generate `num_distinct` completely different random token arrays to force realistic hardware utilization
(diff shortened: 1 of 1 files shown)


.buildkite/benchmark/scripts/generate_decode_only_bm_dataset.py
```diff
@@ -0,0 +1,72 @@
...
+        help="Path to save the JSONL",
+    )
+    args = parser.parse_args()
+
+    tokenizer = AutoTokenizer.from_pretrained(args.model,
+                                              trust_remote_code=True)
+
+    # Generate exact sequence length
+    input_tokens = [DUMMY_TOKEN_ID] * args.input_len
+    input_str = tokenizer.decode(input_tokens)
```
mrjunwan-lang: 1. No need to pre-generate the prompts and store them into a file. we can directly generate the request in the fly.

2. we do not need use a tokenizer to generate the random input string as it is pretty slow and waste of compute. :)
and here input_tokens = [DUMMY_TOKEN_ID] * args.input_len is create the same word input_len times 

3. the prompt string generate can be something like:
random_ids = np.random.randint(low=100, high=50000, size=(BS, PROMPT_LEN))
    prompts = [{"prompt_token_ids": row.tolist()} for row in random_ids]

yiw-wang: 1. I think we still need to put the prompts into a file as we want to still leverage the `vllm bench serve` cli for the benchmark which requires the `--dataset-path` and `--dataset-name` to load the data for benchmark.

2. At the very beginning, I think this is not a waste/sever latency as this is part of the setup process, and the bottleneck is the server turnup(1 to 1.5h). But combined with the 3rd point you mentioned above, I think yes, we can form our dataset with  `prompt_token_ids` to void the tokenization. 

3. Could you elaborate why we want to do this? I use `[DUMMY_TOKEN_ID] * args.input_len` is to even simplied the prompt setup within cl/896608428, that form below prompts
```
{"prompt_token_ids": [100, 100, ..., 100]},
{"prompt_token_ids": [101, 101, ...,101]},
...
```
and seed cache first then do real BM use the exact same prompt which have been cached
```
print(f"  Seeding cache for BS={bs}...", end=" ", flush=True)
llm.generate(prompts, SamplingParams(max_tokens=1), use_tqdm=False)
...
llm.generate(prompts, decode_params, use_tqdm=False)
```

mrjunwan-lang (approved):

.buildkite/benchmark/scripts/generate_decode_only_bm_dataset.py
yiw-wang: updated the logic based on the offline sync.

PR #1756, opened 2026-02-19
[MoE] Add ability to use FP4 MoE weights for `fp8` quant method

jrplatin:
# Description

In this PR, I added support to the FP8 quant method to requantize weights to other dtypes, including FP4.  This is triggered via the `REQUANTIZE_WEIGHT_DTYPE` env var (e.g. `REQUANTIZE_WEIGHT_DTYPE=fp4`).  I also add a quick patch to enable GMM for FP4 (otherwise, run into [this](https://paste.googleplex.com/4725244972630016) issue).  Tracking bug for this issue here: b/483425355.

# Tests

FP8 MMLU:

```
# Server
VLLM_MLA_DISABLE=1 MODEL_IMPL_TYPE=vllm  nohup vllm serve \
--model=deepseek-ai/DeepSeek-R1 \
--max-model-len=4096 \
--tensor-parallel-size 8 \
--max-num-seqs 16 \
--max-num-batched-tokens 256 \
--no-enable-prefix-caching \
--disable-log-requests \
--gpu-memory-utilization 0.95 \
--kv-cache-dtype=fp8 \
--enable-expert-parallel \
--async-scheduling &


# Client
python scripts/vllm/benchmarking/benchmark_serving.py --backend vllm --model=deepseek-ai/DeepSeek-R1 --dataset-name mmlu --dataset-path /mnt/disks/jacobplatin/mmlu/data/test/ --num-prompts 1000 --run_eval --temperature 0


# Result
(EngineCore_DP0 pid=3680308) INFO 02-22 00:35:33 [tpu_runner.py:567] Init model | hbm=[(79.38, 94.75), (79.38, 94.75), (79.38, 94.75), (79.38, 94.75), (79.38, 94.75), (79.38, 94.75), (79.38, 94.75), (79.38, 94.75)]GiB


============ Serving Benchmark Result ============
Successful requests:                     1000      
Benchmark duration (s):                  45.16     
Total input tokens:                      162129    
Total generated tokens:                  2000      
Request throughput (req/s):              22.14     
Output token throughput (tok/s):         44.29     
Total token throughput (tok/s):          3634.59   
---------------Time to First Token----------------
Mean TTFT (ms):                          25591.62  
Median TTFT (ms):                        26924.50  
P99 TTFT (ms):                           44528.71  
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          70.03     
Median TPOT (ms):                        70.12     
P99 TPOT (ms):                           71.66     
[...]

PR #2509, opened 2026-05-05
[MoE] Extend MOE_REQUANTIZE_WEIGHT_DTYPE for unquantized checkpoint

lk-chen:
# Description

The env var was introduced in #1756 for FP8, this PR extends to unquantized model, which (if set) quantize MoE weights to target dtype on the fly.

# Tests

1. unit test
2.
```bash
MOE_REQUANTIZE_WEIGHT_DTYPE=float8_e4m3fn USE_BATCHED_RPA_KERNEL=1 MODEL_IMPL_TYPE=flax_nnx vllm \
    serve google/gemma-4-26B-A4B-it \
    --max-model-len=16384 \
    --max-num-seqs=320 \
    --tensor-parallel-size 2 \
    --no-enable-prefix-caching \
    --kv-cache-dtype=fp8
```

# Checklist

### DIFF
--- a/tests/layers/vllm/test_unquantized.py
+++ b/tests/layers/vllm/test_unquantized.py
@@ -521,7 +521,8 @@ def test_fused_moe(use_ep, num_devices, num_tokens, intermediate_size,
                                   vllm_fused_moe.renormalize,
                                   vllm_fused_moe.activation.value)
 
-    with torchax.default_env(), set_forward_context(None, vllm_config):
+    with torchax.default_env(), set_forward_context(
+            None, vllm_config), jax.set_mesh(mesh):
         assert isinstance(vllm_fused_moe.quant_method,
                           VllmUnquantizedFusedMoEMethod)
         if use_ep:
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -231,8 +231,8 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: os.getenv("MOE_REQUANTIZE_WEIGHT_DTYPE", ""),
     # Specify requantization block size for MoE weights
     "MOE_REQUANTIZE_BLOCK_SIZE":
-    lambda: int(block_size) if (block_size := os.getenv(
-        "MOE_REQUANTIZE_BLOCK_SIZE")) is not None else None,
+    lambda: int(block_size)
+    if (block_size := os.getenv("MOE_REQUANTIZE_BLOCK_SIZE")) else None,
     # dictates whether to layout q-proj as NDH (q-heads, model dim, head dim)
     # or DNH (model dim, q-heads, head dim), which is the default (False)
     "LAYOUT_Q_PROJ_AS_NDH":
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
     REQUANTIZE_BLOCK_SIZE: int | None = None
     REQUANTIZE_WEIGHT_DTYPE: str = "float8_e4m3fn"
     MOE_REQUANTIZE_BLOCK_SIZE: int | None = None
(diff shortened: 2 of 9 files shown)


kyuyeunk (approved):

tests/layers/vllm/test_unquantized.py
```diff
@@ -521,7 +521,8 @@ def test_fused_moe(use_ep, num_devices, num_tokens, intermediate_size,
                                   vllm_fused_moe.renormalize,
                                   vllm_fused_moe.activation.value)
 
-    with torchax.default_env(), set_forward_context(None, vllm_config):
+    with torchax.default_env(), set_forward_context(
```
jrplatin, line 524: Nit: is this from pre-comit? Seems easier to read to me as it  was

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -176,13 +190,19 @@ def apply_jax(self, layer: JaxMoE, x: jax.Array, *,
...
             w2_weight = layer.kernel_down_proj_EFD.value
+            w13_scale = getattr(layer, "kernel_gating_upproj_EDF_weight_scale",
+                                None)
+            w2_scale = getattr(layer, "kernel_down_proj_EFD_weight_scale",
+                               None)
             # TODO (jacobplatin/bzgoogle): we should support bias
             weights = FusedMoEWeights(
                 w13_weight=w13_weight,
-                w13_weight_scale=None,
+                w13_weight_scale=w13_scale.value
```
jrplatin: Nit: is this checked needed if `getattr` is returning `None` in the case where the attr doesn't exist?

tpu_inference/envs.py
```diff
@@ -31,7 +31,7 @@
     REQUANTIZE_BLOCK_SIZE: int | None = None
     REQUANTIZE_WEIGHT_DTYPE: str = "float8_e4m3fn"
     MOE_REQUANTIZE_BLOCK_SIZE: int | None = None
-    MOE_REQUANTIZE_WEIGHT_DTYPE: str = "float8_e4m3fn"
+    MOE_REQUANTIZE_WEIGHT_DTYPE: str = ""
```
jrplatin, line 34: I feel like we might want to keep this as the default? I think much of the commands we already have assue this and so this might be disruptive to change

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -561,3 +562,64 @@ def process_fp8_moe_weights(
...
+def process_unquantized_moe_weights(
+    *,
+    moe_backend: MoEBackend,
+    activation: MoEActivation,
+    w13_weight: jax.Array,
+    w13_bias: jax.Array | None,
+    w2_weight: jax.Array,
+    w2_bias: jax.Array | None,
+) -> FusedMoEWeights:
+    """Jit'ed version to process unquantized moe weights. See `process_moe_weights` for details.
```
jrplatin, line 728: Can you add args / return value here? THanks!

tpu_inference/envs.py
lk-chen: "keep this" will break unquantized.py, the expectation is if user doesn't set it, unquantized MoE remains unquantized

tpu_inference/layers/common/process_weights/moe_weights.py
lk-chen: I thought we intended not to duplicate from "process_moe_weights", see above process_fp8_moe_weights

tpu_inference/layers/jax/quantization/unquantized.py
lk-chen: this is checking "obj" instead of "attr", but anyway I updated to use getattr

tests/layers/vllm/test_unquantized.py
lk-chen: yes from pre-commit, because I added another context

PR #2511, opened 2026-05-05
[Spec decoding] Fix flaky test

gxd3:
Use the spec decoding acceptance rate instead of wall time speedup in the test to avoid noise (e.g. increased compilation time)


Lumosis (approved):

Lumosis:
This is great! Thanks Guangxiang!

PR #2517, opened 2026-05-05
Speed up compilation via AoT + threadpool

wenxindongwork:
# Description

This PR introduces parallelization to the precompilation process by utilizing AoT + thread pool. The main goal is to speed up model precompilation by running multiple compilations concurrently. 

To configure the number of worker threads, set the `NUM_PRECOMPILE_WORKERS` environment variable, which is defaulted to 1. Setting `NUM_PRECOMPILE_WORKERS=1` restores the original sequential precompilation behavior. 

**Qwen3 30b fp8**

Baseline: 

`Function 'capture_model' executed in 209.3670 seconds.`

`Function '_precompile_backbone_text_only' executed in 180.2982 seconds.`

With AoT + threadpool (2x speed up)

`Function 'capture_model' executed in 94.0618 seconds.`

`Function '_precompile_backbone_text_only' executed in 59.3216 seconds.`

**Qwen3 235b bf16**

Baseline: 

`Function 'capture_model' executed in 345.0942 seconds.`
`Function '_precompile_backbone_text_only' executed in 311.3094 seconds.`

With AoT + threadpool (2x speed up):

`Function 'capture_model' executed in 156.5563 seconds.`

`Function '_precompile_backbone_text_only' executed in 99.5562 seconds.`

# Tests

commands to reproduce.

# Checklist


tpu_inference/runner/compilation_manager.py
```diff
@@ -95,56 +121,150 @@ def _run_compilation(self,
...
+        warmup_start = time.perf_counter()
+        for name, fn, args, call_kwargs, warmup_handler in tasks:
+            try:
+                if warmup_handler is not None:
+                    out = warmup_handler(fn, args, call_kwargs)
+                else:
+                    out = fn(*args, **call_kwargs)
+                jax.tree.map(lambda r: r.block_until_ready(), out)
+            except Exception as e:
+                logger.warning("Warm-up call for %s failed: %r", name, e)
```
yiw-wang: Just a quick question here to help me understand this better:
If a warm-up fails here, wouldn't it mean that the specific graph might not be fully cached, potentially causing an unexpected JIT compilation (and a huge latency spike) or a crash during the actual runtime serving? I might be missing some context, would it be safer to fail-fast here?

yiw-wang (approved):

tpu_inference/runner/compilation_manager.py
wenxindongwork: good catch, fixed, thank you!

PR #2526, opened 2026-05-06
[benchmark] Fix Total input tokens reporting for chat/multimodal datasets

### LINKED ISSUE #2525: [Bug] benchmark_serving.py reports "Total input tokens: 0" for MMMU-Pro dataset
## Summary

When running `scripts/vllm/benchmarking/benchmark_serving.py` with `--dataset-name mmmu_pro`, the reported `Total input tokens` is always `0`, regardless of `--num-prompts`.

## Repro

```
    --backend vllm-chat \
    --dataset-name mmmu_pro \
    --mmmu-pro-output-len=64 \
    --num-prompts 4 \
    --max-concurrency 4
```

Output:
```
Successful requests:                     4
Total input tokens:                      0
Total generated tokens:                  256
```

## Root cause

`MMMUProDataset.sample()` in `scripts/vllm/benchmarking/benchmark_dataset.py` (around L747) constructs each `SampleRequest` without setting `prompt_len`:

```python
samples.append(
    SampleRequest(
        messages=messages,
        expected_output_len=new_output_len,
        multi_modal_data=mm_content,
        completion=answer,
    ))
```

`SampleRequest.prompt_len` defaults to `0` (`benchmark_core.py`), and `calculate_metrics` in `benchmark_serving.py` aggregates `total_input` by summing `input_requests[i].prompt_len`, so it sums to 0. Other datasets (e.g. `MLPerfDataset`, `MMLUDataset`) correctly set `prompt_len = len(tokenizer(prompt).input_ids)`.

It's also worth noting that the chat backend path in `backend_request_func.py` only consumes `usage.completion_tokens` from the response; `usage.prompt_tokens` (which vLLM does return, both non-streaming and in the trailing SSE chunk when `stream_options.include_usage=true`) is discarded.

## Suggested fix

Two options, not mutually exclusive:

1. In `MMMUProDataset.sample()`, populate `prompt_len` from the tokenized question (and ideally apply the chat template / account for image tokens).
2. In the chat backend in `backend_request_func.py`, also capture `usage.prompt_tokens` onto `RequestFuncOutput`, and prefer it in `calculate_metrics` when available. This would give an accurate count for any multimodal/chat dataset without re-tokenizing client-side.

lk-chen:
## Summary

- Capture `usage.prompt_tokens` from the streamed OpenAI response in `_openai_fetch` (vLLM emits it in the trailing chunk when `stream_options.include_usage=true`).
- Prefer the server-reported value over the client-side `SampleRequest.prompt_len` in `calculate_metrics`, falling back to `prompt_len` when the server doesn't report usage.
- This fixes "Total input tokens: 0" for `MMMUProDataset` (which never set `prompt_len`), and gives a more accurate count for any chat/multimodal dataset since the server count includes chat-template and image tokens that the client tokenizer doesn't see.

Fixes #2525.

## Verification

Before, with `--dataset-name mmmu_pro --num-prompts 4`:
```
Total input tokens:                      0
Total generated tokens:                  256
```

After:
```
Total input tokens:                      1704
Total generated tokens:                  256
```

## Test plan

- [x] Repro with `mmmu_pro` (`standard (10 options)`) — `Total input tokens` is non-zero and matches `usage.prompt_tokens`.
- [ ] Spot-check that an existing completion-API dataset (e.g. `mlperf`) still reports the same `total_input` (server value, when present, equals client `prompt_len`).

### DIFF
--- a/scripts/vllm/benchmarking/benchmark_serving.py
+++ b/scripts/vllm/benchmarking/benchmark_serving.py
@@ -146,7 +146,7 @@ def calculate_metrics(
     tokenizer: PreTrainedTokenizerBase,
     selected_percentiles: list[float],
     goodput_config_dict: dict[str, float],
-) -> tuple[BenchmarkMetrics, list[int]]:
+) -> tuple[BenchmarkMetrics, list[int], list[int]]:
     actual_output_lens: list[int] = []
     actual_input_lens: list[int] = []
     total_input = 0
@@ -349,7 +349,7 @@ async def benchmark(
 
     benchmark_duration = time.perf_counter() - benchmark_start_time
 
-    metrics, actual_output_lens, actual_input_lens = calculate_metrics(
+    metrics, actual_output_lens, input_lens = calculate_metrics(
         input_requests=input_requests,
         outputs=outputs,
         dur_s=benchmark_duration,
@@ -385,7 +385,7 @@ async def benchmark(
         metrics.request_goodput if goodput_config_dict else None,
         "output_throughput": metrics.output_throughput,
         "total_token_throughput": metrics.total_token_throughput,
-        "input_lens": actual_input_lens,
+        "input_lens": input_lens,
         "output_lens": actual_output_lens,
         "ttfts": [output.ttft for output in outputs],
         "itls": [output.itl for output in outputs],
--- a/scripts/vllm/benchmarking/benchmark_serving.py
+++ b/scripts/vllm/benchmarking/benchmark_serving.py
@@ -148,6 +148,7 @@ def calculate_metrics(
     goodput_config_dict: dict[str, float],
 ) -> tuple[BenchmarkMetrics, list[int]]:
     actual_output_lens: list[int] = []
+    actual_input_lens: list[int] = []
     total_input = 0
     completed = 0
     good_completed = 0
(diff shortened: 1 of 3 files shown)


scripts/vllm/benchmarking/benchmark_serving.py
```diff
@@ -148,6 +148,7 @@ def calculate_metrics(
     goodput_config_dict: dict[str, float],
 ) -> tuple[BenchmarkMetrics, list[int]]:
```
wdhongtw: The function signature should be changed.

scripts/vllm/benchmarking/benchmark_serving.py
```diff
@@ -342,7 +349,7 @@ async def limited_request_func(request_func_input, pbar):
 
     benchmark_duration = time.perf_counter() - benchmark_start_time
 
-    metrics, actual_output_lens = calculate_metrics(
+    metrics, actual_output_lens, actual_input_lens = calculate_metrics(
```
wdhongtw: nit: The name here is a little confusing as it's not necessary the actual input lens, but have some fallback values. Same problem in the impl of `calculate_metrics`.

wdhongtw (commented):
Looks good to me. 👍

Just one minor issue.

scripts/vllm/benchmarking/benchmark_serving.py
lk-chen: done, changed

scripts/vllm/benchmarking/benchmark_serving.py
lk-chen: done, renamed to "input_lens"

wdhongtw (approved):
Looks good!

PR #2535, opened 2026-05-07
[Spec decoding] make drafter's prepare_inputs() a jited fn

gxd3:
Turn Eagle3 drafter's prepare_inputs() a jit-fn as well.
The goal is move most of the drafter's work from host cpu to TPU (end goal: no TPU execution bubble when MTP is enabled)

# Tests
Unit test
Acceptance rate: https://paste.googleplex.com/4598859633393664 using the Spec-bench dataset.
xprof: http://xprof/trace_viewer.html?session_id=gxd-1634087802983507300

### DIFF
--- a/tpu_inference/spec_decode/jax/eagle3.py
+++ b/tpu_inference/spec_decode/jax/eagle3.py
@@ -28,7 +28,6 @@ from tpu_inference.layers.common.attention_metadata import AttentionMetadata
 from tpu_inference.logger import init_logger
 from tpu_inference.models.common.model_loader import (
     get_model, resolve_model_architecture)
-from tpu_inference.runner import utils as runner_utils
 from tpu_inference.utils import device_array
 
 logger = init_logger(__name__)
@@ -136,7 +135,6 @@ class Eagle3Proposer:
             else:
                 logger.info("Draft model has its own embed_tokens.")
 
-    @jax.jit(static_argnums=(0, ))
     def _prepare_input_ids(
             self, query_start_loc: jax.Array, target_token_ids: jax.Array,
             next_token_ids: jax.Array,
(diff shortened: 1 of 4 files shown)


tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -259,84 +256,99 @@ def prepare_inputs(
...
+            "EAGLE3 requires auxiliary hidden states from the target model.")
 
         if num_rejected_tokens is None:
-            num_reqs = device_array(self.mesh,
-                                    np.asarray([num_reqs], dtype=jnp.int32))
             # block_tables = device_array(self.mesh, block_tables)
-            attn_metadata = replace(attn_metadata,
-                                    block_tables=device_array(
-                                        self.mesh, block_tables))
+            attn_metadata = replace(attn_metadata, block_tables=block_tables)
```
Lumosis, line 299: Is this something that can be done on TPU?

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -259,84 +256,99 @@ def prepare_inputs(
             draft_kv_cache_group_id].get_cpu_tensor().reshape(-1)
```
Lumosis, line 256: Is it better to do reshape on CPU or on TPU?

tpu_inference/spec_decode/jax/eagle3.py
gxd3: yes, replace() is updating Jax Pytree here, so it can run in TPU.
(if not supported, there should be compilation error during the precompile)

tpu_inference/spec_decode/jax/eagle3.py
gxd3: I kept this unchanged for now, but i can look into moving it to TPU as well in the next PR :)

Lumosis (approved):

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Thanks for explanation

PR #2536, opened 2026-05-07
[MTP] Support speculative tokens in BlockTable and fix shared layer check

Lumosis:
# Description

This PR introduces two main modifications to support Multi-Token Prediction (MTP) and speculative decoding:

* It passes num_speculative_tokens down to the `BlockTable` to ensure sufficient block allocation.
* It relaxes the shared layer assertion in `kv_cache_manager.py` to exclude MTP-specific tensors, which may have different sharing patterns.

## Details
1. Speculative Token Support in Block Table
* In `kv_cache_manager.py`, we now extract num_speculative_tokens from the speculative configuration if it exists. 
* This value is passed to `InputBatch` and subsequently to `MultiGroupBlockTable`
* In MultiGroupBlockTable, the number of blocks is now guaranteed to be at least 1 + num_speculative_tokens, ensuring that we don't under-allocate blocks when running with speculative decoding.

2. MTP-aware Shared Layer Assertion
* Previously, the assertion in `KVCacheManager` required all KV cache tensors to have the same number of shared layers.
* Since MTP modules might not follow the same sharing pattern as the base model (e.g., Qwen3.5 sharing pattern), this PR updates the check to only apply to non-MTP tensors. MTP tensors are identified by checking if "mtp" is present in the names of the layers they are shared by.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/block_table.py
+++ b/tpu_inference/runner/block_table.py
@@ -84,13 +84,19 @@ class BlockTable:
 class MultiGroupBlockTable:
     """The BlockTables for each KV cache group."""
 
-    def __init__(self, max_num_reqs: int, max_model_len: int,
-                 max_num_batched_tokens: int, pin_memory: bool,
-                 block_sizes: list[int]) -> None:
+    def __init__(self,
+                 max_num_reqs: int,
+                 max_model_len: int,
+                 max_num_batched_tokens: int,
+                 pin_memory: bool,
+                 block_sizes: list[int],
+                 num_speculative_tokens: int = 0) -> None:
         self.block_tables = [
-            BlockTable(max_num_reqs, cdiv(max_model_len, block_size),
-                       max_num_batched_tokens, pin_memory)
-            for block_size in block_sizes
+            BlockTable(
+                max_num_reqs,
+                max(cdiv(max_model_len, block_size),
+                    1 + num_speculative_tokens), max_num_batched_tokens,
+                pin_memory) for block_size in block_sizes
         ]
 
     def append_row(self, block_ids: list[list[int]], row_idx: int) -> None:
--- a/tpu_inference/runner/input_batch.py
+++ b/tpu_inference/runner/input_batch.py
@@ -56,6 +56,7 @@ class InputBatch:
         vocab_size: int,
         block_sizes: list[int],
         is_spec_decode: bool = False,
+        num_speculative_tokens: int = 0,
     ):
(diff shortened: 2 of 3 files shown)


gxd3 (approved):

tpu_inference/runner/block_table.py
```diff
@@ -84,13 +84,19 @@ def get_cpu_tensor(self) -> jax.Array:
...
+                 block_sizes: list[int],
+                 num_speculative_tokens: int = 0) -> None:
         self.block_tables = [
-            BlockTable(max_num_reqs, cdiv(max_model_len, block_size),
-                       max_num_batched_tokens, pin_memory)
-            for block_size in block_sizes
+            BlockTable(
+                max_num_reqs,
+                max(cdiv(max_model_len, block_size),
+                    1 + num_speculative_tokens), max_num_batched_tokens,
```
gxd3, line 98: qq:
does this shall be cdiv(max(1 + num_speculative_tokens, max_model_len), block_size)?

gxd3: or maybe cdiv(1 + num_speculative_tokens + max_model_len, block_size)?

Lumosis: ###  Suggestion 1: cdiv(max(1 + num_speculative_tokens, max_model_len), block_size)
Let's plug in the values for a Mamba layer with MTP:

num_speculative_tokens = 2 -> 1 + 2 = 3.
max_model_len = 10240.
max(3, 10240) = 10240.
block_size = 10240 (for Mamba layers).
cdiv(10240, 10240) = 1.
Result: This would allocate 1 column. Why it fails: We need 3 columns to store the 3 block IDs returned by vLLM. This suggestion would not fix the crash.

###  Suggestion 2: cdiv(1 + num_speculative_tokens + max_model_len, block_size)
Let's plug in the values for a Mamba layer with MTP:

1 + 2 + 10240 = 10243.
block_size = 10240.
cdiv(10243, 10240) = 2 (ceiling division).
Result: This would allocate 2 columns. Why it fails: We need 3 columns. This would still result in a broadcast crash when the scheduler returns 3 blocks.

###  Why my formula works better
My formula was: max(cdiv(max_model_len, block_size), 1 + num_speculative_tokens)

For Mamba with MTP: max(1, 3) = 3. It gives exactly the 3 slots we need.

PR #2538, opened 2026-05-07
[xprof] Add env option to allow setting advanced profiling options

wyzhang:
Disabled by default. When enabled, restricted the profiling to one tpu and one sparsecore.

why:
Reduce the xprof size to allow profiling more steps without hitting the
protobuf limit causing trace being dropped.

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
     ONEHOT_MOE_PERMUTE_THRESHOLD: int = 0
+    PROFILE_SINGLE_DEVICE: bool = False
 
 
 def env_with_choices(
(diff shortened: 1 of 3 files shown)


tpu_inference/envs.py
```diff
@@ -21,6 +21,9 @@
     NEW_MODEL_DESIGN: bool = False
     PHASED_PROFILING_DIR: str = ""
     PYTHON_TRACER_LEVEL: int = 1
+    HOST_TRACER_LEVEL: str | None = None
```
helloworld1: I think those are too many knobs to tweak. We should just expose one flag like CONCISE_XPROF_TRACE instead and use a good configuration for 1 TPU chip without  losing too much info.

wyzhang: Done. Just one env now.

helloworld1 (approved):

tpu_inference/envs.py
```diff
@@ -376,6 +378,9 @@ def _get_int_list_env() -> list[int]:
     # is effectively disabled.
     "ONEHOT_MOE_PERMUTE_THRESHOLD":
     lambda: int(os.getenv("ONEHOT_MOE_PERMUTE_THRESHOLD", "0")),
+    # Limit profiling to a single TPU chip + single SparseCore.
+    "PROFILE_SINGLE_DEVICE":
+    env_bool("PROFILE_SINGLE_DEVICE", default=False),
```
bigeye, line 382: I'm open for discussion, but should we actually enable this by default? At least tpu_num_sparse_cores_to_trace (because it will cause event drops in the xprof with default setup anyway). We usually need to look at all chip's profile (e.g., expert imbalance),  but for the sparse core, I believe we wouldn't need to look at one device

wyzhang: Keep it off by default for now to start with

zhangamy-crypto (approved):

vipannalla (approved):

wenxindongwork (approved):

sixiang-google (approved):

PR #2551, opened 2026-05-08
[DSv3] Reduce MLA copy latency with custom call transposes

gpolovets1:
# Description

There are expensive XLA copies caused by transposing the physical tiling before and after the MLA kernel.
By replacing the XLA transposes with transpose kernel implementations, we find that we can reduce the copy latency by ~80%. 
This amounts to about 4% e2e throughput increase on the 1k/8k multi host workload.

# Tests
Confirmed MMLU quality and performance gains manually on a multi host VM. 
Added unit tests to confirm correctness of transpose and no VMEM OOM under different settings 

# Checklist


tpu_inference/kernels/custom_calls/kernel.py
```diff
@@ -0,0 +1,205 @@
...
+import bisect
+from collections.abc import Sequence
+
+import jax
+import jax.experimental.pallas as pl
+import jax.experimental.pallas.tpu as pltpu
+import jax.numpy as jnp
+from sympy import divisors
+
+min_pipeline = 3
```
Xhark: nit: unused?

tpu_inference/kernels/custom_calls/kernel.py
```diff
@@ -0,0 +1,206 @@
+# Copyright 2026 Google LLC
```
gxd3, line 1: could we name the directory "tpu_inference/kernels/transpose" instead for better clarity?

tpu_inference/kernels/custom_calls/kernel.py
```diff
@@ -0,0 +1,206 @@
...
+import jax.numpy as jnp
+from sympy import divisors
+
+min_pipeline = 3
+
+
+@jax.jit(static_argnames=[
+    'transpose_axes',
+])
+def xpose_full(input, *, transpose_axes):
```
gxd3, line 33: is `xpose_full` only used in unit test?
If so, maybe define it in the test file as well?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -500,9 +502,10 @@ def mla_attention(
         sm_scale: softmax scale
     """
     in_specs = (
-        query_tnh_sharding or P(ShardingAxisName.MLP_TENSOR, None, None),  # q
+        query_nth_sharding
+        or P(None, ShardingAxisName.MLP_TENSOR, None),  # q (head-major)
         query_tnh_sharding
-        or P(ShardingAxisName.MLP_TENSOR, None, None),  # q_rope
+        or P(ShardingAxisName.MLP_TENSOR, None, None),  # q_rope (token-major)
```
gxd3, line 508: we didn't change `q_rope`?

tpu_inference/layers/vllm/backends/flash_attn_mla.py
```diff
@@ -146,13 +146,16 @@ def forward(self,
...
-                  jax_view(layer.W_UK_T_scale)).astype(input_dtype)
+        # Einsum selects 'n' as the batch axis and emits it as the major-most physical dimension.
+        q_nope = jnp.einsum(
+            "bnp,npl->nbl",
+            q_nope,
+            jax_view(layer.W_UK_T),  # torch nn param
+            preferred_element_type=jnp.float32)
+        scale = jax_view(layer.W_UK_T_scale)  # torch nn param
+        # (1, N, L) -> (N, 1, L)
+        scale = scale.reshape(scale.shape[1], 1, scale.shape[-1])
```
gxd3: nit: we could adjust the shape `layer.W_UK_T_scale` in weight loading time?

tpu_inference/kernels/custom_calls/kernel.py
gpolovets1: I was considering having this file be reserved for other lower level primitive/operations that we need to replace with kernel calls. But happy to rename to transpose for now and change later as needed.

tpu_inference/kernels/custom_calls/kernel.py
gpolovets1: It was originally proposed by Jaehong but in my experience it was a little bit slower and didn't scale to relatively large tensors. I kept it here for reference/in case we want to try using it again in the future. Do you prefer I remove it altogether?

tpu_inference/layers/common/attention_interface.py
gpolovets1: q_rope does not go through the einsum so it's actually already in token-major and does not need a transpose!

tpu_inference/layers/vllm/backends/flash_attn_mla.py
gpolovets1: Sure, let me look into it!

tpu_inference/kernels/mla/v2/kernel.py
```diff
@@ -23,6 +23,9 @@
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
 
+from tpu_inference.kernels.custom_calls.kernel import (prev_closest_divisor,
```
kyuyeunk: in general, i like to have kernels self-contained as possible. is the newly introduced kernels a generic kernel or mla specific kernel? if it's mla specific, maybe having them in a same kernel might be a better option? or we can rewrite the code in a way that mla kernel is expected to accept pre-transpose argument and the transpose kernel is called at attention_interface.py?

gpolovets1: Thanks for your questions and suggestions! This is a generic transpose kernel for reducing copy latency. So far it's just used to reduce copies introduced by MLA but could be applied more broadly to expensive relayouts for any model.
Also, I think the way that we've structured attention kernels is that there is a `prepare_q_inputs` method in the kernel file that reshapes/pads the q input, so I tried to be consistent in including this kernel call there.

tpu_inference/kernels/custom_calls/kernel.py
gpolovets1: Removed - thanks!

tpu_inference/layers/vllm/backends/flash_attn_mla.py
gpolovets1: Removed reshape and added native loading in correct layout - thanks Guangxiang!

tpu_inference/kernels/custom_calls/kernel.py
gxd3: "custom_calls" appear to be too general that does not tell readers what does the kernel actually do, :)
suggest to rename it to "transpose" :)

tpu_inference/kernels/mla/v2/kernel.py
```diff
@@ -1317,28 +1326,45 @@ def prepare_q_inputs(
...
+        q,
+        (
+            (0, num_q_heads - actual_num_q_heads),
+            (0, 0),
+            (0, head_dim - actual_head_dim),
+        ),
+        constant_values=0,
+    )
+    # Physical transpose: (N, T, D) -> (T, N, D), pipelined over T
+    q = xpose_pipeline(q, transpose_axes=(1, 0, 2), n_tile=128, m_tile=32)[0]
```
gxd3: qq:
why we change the shape so that 
"q_packing" dimension is removed?

tpu_inference/kernels/custom_calls/kernel.py
gxd3: as discussed offline, we plan to move this file under MLA folder.

gxd3 (approved):

tpu_inference/kernels/mla/v2/kernel.py
gpolovets1: It introduces an expensive retiling/copy from (4, 128)(4, 1) to (8, 128)(4, 1) IIRC.

tpu_inference/kernels/mla/v2/kernel.py
```diff
@@ -23,6 +23,12 @@
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
```
gxd3, line 24: a relevant ask:
could we backfill unit test for mla_v2 (maybe just copy `tests/kernels/mla_v1_test.py` and adjust it to test mla-v2)?

tpu_inference/kernels/custom_calls/kernel.py
gpolovets1: Renamed to transpose.py and moved to kernels/mla/v2 directory.

tpu_inference/kernels/mla/v2/kernel.py
gpolovets1: Let me try to work on it as a separate task later this week!

PR #2574, opened 2026-05-11
recurrent_scan_v2: fix has_initial_state and Mosaic tile alignment

qizzzh:
Fix has_initial_state for chunked-prefill continuations: derive has_initial_state from seq_lens - query_lens so brand-new prefills zero stale mamba slot state, matching GPU behavior.

Fix Mosaic tile alignment for DP attention: in DP mode each chip sees qkv_width=12288 (all heads), causing Mosaic to tile as (16,128)(2,1). Dynamic DMA sizes violated the BoundedSlice constraint. Use a static block_size for DMA windows and redirect invalid pipeline steps to a padded sink area beyond valid data.

PR #2577, opened 2026-05-11
Enable correct DP attention for hybrid attention+mamba models

qizzzh:
This PR is in collaboration with @wyzhang 

Key changes:
- InputBatch: accept dp_size, partition mamba slot pools per rank, and route add_request / remove_request through the correct rank's pool
- kv_cache_manager: pass dp_size when recreating InputBatch in maybe_reinitialize_input_batch; track actual_mamba_num_blocks so the runner can re-init pools after KV cache allocation
- tpu_runner: call init_mamba_pools after KV cache init; reorder mamba_state_indices per DP rank (global % local_slots) in _prepare_inputs_dp, matching the block_table reordering
- persistent_batch_manager: forward assigned_dp_rank from scheduler output to InputBatch.add_request
- DPScheduler: replace IPC-based token count queries with local prompt-token tracking; use wait()-based result collection to avoid head-of-line blocking across ranks
- change dp scheduler to hold and dispatch prefill together. Enabled by default. Can be disabled by DP_SCHED_BATCH_PREFILL. This is crucial to achieve better performance by minimizing the mixture of long running prefill with faster decode together in a single dp batch.
- gdn_attention: reduce chunk_size from 64 to 32 for recurrent_scan_v2 prefill path

This change depends on https://github.com/vllm-project/tpu-inference/pull/2574

Validated on Qwen3.5-397B-A17B-FP8 DP=8 (v7x-8): mmlu_pro 0.8129 vs TP=8 baseline 0.8271 (within 1σ noise at limit=50).

DP=8 vs TP=8 benchmark (output tok/s / total tok/s):

server command: NEW_MODEL_DESIGN=1 MODEL_IMPL_TYPE=vllm USE_MOE_EP_KERNEL=0 RAGGED_GATED_DELTA_RULE_IMPL=chunked_kernel_p_recurrent_kernel_d     nohup vllm serve Qwen/Qwen3.5-397B-A17B-FP8     --max-model-len=9216 --max-num-batched-tokens=1024 --enable-chunked-prefill --max-num-seqs=64     --no-enable-prefix-caching --gpu-memory-utilization=0.95     --tensor-parallel-size=8 --async-scheduling --port=8000     --language-model-only --enable-auto-tool-choice --tool-call-parser=qwen3_coder     --reasoning-parser=qwen3 '--limit-mm-per-prompt={"image": 0, "video": 0}'     --kv-cache-dtype=fp8 --enable-expert-parallel     '--additional-config={"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}'

| Config                | TP=8           | DP=8           | DP Advantage |
| --------------------- | -------------- | -------------- | ------------ |
| in=1k, out=8k, c=64   | 1,964 / 2,209  | 1,973 / 2,214  | 0%                  |
| in=1k, out=8k, c=128   | 2,952 / 3,320  | 3,363 / 3,785  | +14%                |
| in=1k, out=8k, c=256   | 4,130 / 4,646  | 5,029 / 5,656  | +22%                |
| in=1k, out=8k, c=512  | 4,266 / 4,798  | 6,914 / 7,777  | +62%       |
| in=8k, out=1k, c=64   | 1,097 / 9,888  | 1,247 / 11,224 | +14%       |
| in=8k, out=1k, c=128   | 1,353 / 12,197  | 1,668 / 15,050 | +23%       |
| in=8k, out=1k, c=256   | 1,496 / 13,483  | 2,033 / 18,326 | +36%       |
| in=8k, out=1k, c=512  | 1,490 / 13,429 | 2,090 / 18,845 | +40%       |


kyuyeunk:
awesome! but out of curiosity, why would dp show any benefit at `c=64`?

wyzhang:
Should this --max-num-batched-tokens=2048  or 1024? Wondering the 2048 would lead to a 16k max number of batched token across all dp, which is larger than the baseline TP with 8k? so the comparison may not be fair ?

qizzzh:
> * buildkite/tpu-inference-ci/pr/tpu6e-tpu-test-notification

DP shows 0% benefit at c=64 for the 1k/8k workload (decode-dominated, small input). The benefit at c=64 shows up in the 8k/1k workload (+24%) — with 8k input per request, prefill takes multiple chunked steps. In DP mode, each rank prefills its own requests independently (8 prefills in parallel across ranks), whereas TP=8 uses all 8 chips to prefill one chunk at a time. Prefill is compute bound so the collective overhead overweights the gain from TP.

qizzzh:
> Should this --max-num-batched-tokens=2048 or 1024? Wondering the 2048 would lead to a 16k max number of batched token across all dp, which is larger than the baseline TP with 8k? so the comparison may not be fair ?

Both TP=8 and DP=8 servers were started with the same --max-num-batched-tokens=2048. You're right that the DPScheduler currently gives each rank the full 2048 budget rather than dividing by dp_size, so the total scheduling capacity across all ranks is higher.

However, max_num_batched_tokens only limits how many tokens can be in a single step's batch. For decode steps (which dominate 1k/8k throughput), batch size = number of concurrent requests, not max_num_batched_tokens. At c=512, TP=8 decodes 512 tokens/step and DP=8 decodes 64 tokens/rank/step — both well under 2048. So this parameter doesn't drive the throughput difference. 

The DP advantage comes from KV cache capacity: TP=8 pads 2 KV heads to 8 (4x waste), giving 4,533 attn blocks. DP=8 eliminates that padding and gets 2,416 blocks × 2176 tokens/block, which can sustain more concurrent requests at high load.

qizzzh:
> Should this --max-num-batched-tokens=2048 or 1024? Wondering the 2048 would lead to a 16k max number of batched token across all dp, which is larger than the baseline TP with 8k? so the comparison may not be fair ?

For prefill heavy it may have an impact. Let me rerun with TP=8 8k and DP=1k and see what I get.

kyuyeunk:
> DP shows 0% benefit at c=64 for the 1k/8k workload (decode-dominated, small input). The benefit at c=64 shows up in the 8k/1k workload (+24%) — with 8k input per request, prefill takes multiple chunked steps. In DP mode, each rank prefills its own requests independently (8 prefills in parallel across ranks), whereas TP=8 uses all 8 chips to prefill one chunk at a time. Prefill is compute bound so the collective overhead overweights the gain from TP.

not sure i follow. even in tp, that should be mitigated by using higher `--max-num-batched-tokens`? for instance, you are currently using `--max-num-batched-tokens=2048`, which means you are effectively using global batch size of 2048 * 8 = 16384. so that should have clear benefit over tp setup using global batch size of 8192?

qizzzh:
> > DP shows 0% benefit at c=64 for the 1k/8k workload (decode-dominated, small input). The benefit at c=64 shows up in the 8k/1k workload (+24%) — with 8k input per request, prefill takes multiple chunked steps. In DP mode, each rank prefills its own requests independently (8 prefills in parallel across ranks), whereas TP=8 uses all 8 chips to prefill one chunk at a time. Prefill is compute bound so the collective overhead overweights the gain from TP.
> 
> not sure i follow. even in tp, that should be mitigated by using higher `--max-num-batched-tokens`? for instance, you are currently using `--max-num-batched-tokens=2048`, which means you are effectively using global batch size of 2048 * 8 = 16384. so that should have clear benefit over tp setup using global batch size of 8192?

Yeah I'm testing --max-num-batched-tokens=1k as mentioned above.

wyzhang:
> > Should this --max-num-batched-tokens=2048 or 1024? Wondering the 2048 would lead to a 16k max number of batched token across all dp, which is larger than the baseline TP with 8k? so the comparison may not be fair ?
> 
> For prefill heavy it may have an impact. Let me rerun with TP=8 8k and DP=1k and see what I get.

@qizzzh and I discussed, investigated and worked together on this. 
1. For fair comparison against the baseline, 1k is the right setting, so total # of tokens across ranks remains the same as 8k like the GPU baseline
2. The other issue we dig into together is the mixing of prefills and decode was causing the overall performance to degrade substantially. After changing the dp scheduler to buffer and dispatch a set of prefills together, this mitigates the issue and we observed substantially better performance ranging from 13% to 63% depending on the traffic pattern and concurrency (except 1k/8k with 64 concurrency which shows on-par performance)

tpu_inference/envs.py
```diff
@@ -340,6 +341,10 @@ def _get_int_list_env() -> list[int]:
     # Enable hierarchical reduce-scatter kernel for MoE
     "ENABLE_RS_KERNEL":
     env_bool("ENABLE_RS_KERNEL", default=False),
+    # DP-scheduler: hold incoming requests (prefills) at the DP layer until
+    # `dp_size` of them have accumulated. The goal is to cluster new prefills.
+    "DP_SCHED_BATCH_PREFILL":
+    env_bool("DP_SCHED_BATCH_PREFILL", default=True),
```
kyuyeunk, line 347: this sounds like it can potentially hurt latency? how much throughput vs. latency impact does this have?

wyzhang: In this case, gain considerably outweigh the potential performance loss. Allowing a prefills to proceed and mix with decode will cause the the next few decode steps to run 10x slower (AFAICR prefill is about 300ms and decode is about 30ms). So the gain of holding onto the prefill for a bit outweighs letting it proceed. When prefill is short and runs as fast as decode, then you are correct, this might hurt. Later we add some heuristic to decide when to buffer and when to not to, based on prefil latency vs decode.

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -238,6 +240,17 @@ def _send_result(result):
                         total_tokens += len(req.all_token_ids)
                     _send_result(total_tokens)
 
+                case SchedulerCommand.GET_PENDING_PREFILL_TOKENS:
+                    # Number of tokens to be prefilled
+                    pending = 0
+                    for req in scheduler.running:
```
wenxindongwork, line 246: should `the scheduler.waiting` queue also be considered here?

wenxindongwork: otherwise new requests could get assigned to a rank with a lot of waiting requests, but few pending running prompt tokens.

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -528,41 +560,65 @@ def _get_result(self,
...
+            for rank in range(self.dp_size):
+                cached_tokens = self._get_result(
+                    rank, SchedulerCommand.PROBE_COMPUTED_BLOCKS)
+                if cached_tokens > best_cache_tokens:
+                    best_cache_tokens = cached_tokens
+                    best_cache_rank = rank
+            if best_cache_tokens > 0:
+                return best_cache_rank
+
+        pending = self._get_rank_pending_prefill_tokens()
```
wenxindongwork, line 620: generally curious what performance benefits you saw by updating the routing strategy to consider pending prefill tokens instead of total tokens? 

and do you think it applies to models other than mamba?

tpu_inference/core/sched/dp_scheduler.py
qizzzh: Good catch. scheduler.waiting isn't counted here. In practice, DP_SCHED_BATCH_PREFILL dispatches prefills one-per-rank in round-robin batches, so waiting queues stay balanced by construction. Requests also move from waiting→running within one schedule() cycle, so the window where they're invisible is very short. That said, it would be straightforward to add scheduler.waiting to the sum for correctness. We can add a TODO/follow-up for this - it's low-risk to defer since the batching policy keeps things balanced today, but it's the right thing to do longer term.

tpu_inference/core/sched/dp_scheduler.py
qizzzh: The switch from total-tokens to pending-prefill-tokens matters because for DP load-balancing, what we really want to equalize is future work, not accumulated state. Total tokens (which is dominated by decode-state tokens for decode heavy workloads) doesn't predict how busy a rank will be in the next few steps — a rank with 10k decode tokens but 0 pending prefill is actually the least loaded for the next step (just 1 decode token/req). Pending prefill tokens directly measures backlog.

The impact was significant: without this change, the old total-token metric would steer new prefills away from ranks that had already accumulated lots of decode state, even though those ranks were idle in terms of prefill work. This led to imbalanced prefill clustering — one rank would get multiple prefills while others sat idle.

For non-mamba (pure attention) models: yes, the same logic applies. The DP scheduler code is model-agnostic - this routing strategy benefits any DP attention setup. The mamba-specific parts of this PR are the state management (mamba slot reordering, has_mamba_layers gating), not the scheduling logic.

wenxindongwork: thanks for the explanation, this makes sense!

tpu_inference/core/sched/dp_scheduler.py
wenxindongwork: i'm mainly concerned that this will impact the scheduling of non-mamba models and cases where DP_SCHED_BATCH_PREFILL is disabled. 
Currently this function gets called for all cases, and I'm not sure what the impact to other models would be if we don't count the waiting requests.

qizzzh: Actually my comment was wrong.. Below already includes scheduler.waiting:

wyzhang: @wenxindongwork I belive the code before this change is incorrectly counting all tokens including those already being prefilled, causing dp scheduler NOT scheduling to rank that has capacity to prefill. 

In short, this change might improve the scheduling for other models. 

The reason is: 

_find_best_rank_for_request 
  -> _get_rank_token_counts 
    ->   self._send_command(rank, SchedulerCommand.GET_TOKEN_COUNT)
      ->  which does 
```
                    total_tokens = 0
                    for req in scheduler.running:
                        total_tokens += len(req.all_token_ids)
                    for req in scheduler.waiting:
                        total_tokens += len(req.all_token_ids)
                    _send_result(total_tokens)
```

This means if a rank has only 1 token to decode after prefiling 8k and decoding 1k - 1 token, the scheduler will treat this rank to be too busy to schedule any request to it, which seems not right.

tpu_inference/core/sched/dp_scheduler.py
wyzhang: @wenxindongwork As mentioned in the other comment, I believe it will, because we are properly counting the prefill-capacity avaliable, while previously the counting including already prefilled tokens. Please correct if you see us misunderstood the code.

tpu_inference/envs.py
kyuyeunk: got it. can you slightly tweak the pr description to make that stands out? because my understanding is that this feature is the crucial to achieve good dp attention performance and by reading the pr description, that doesn't stand out and it sounds more like a general dp attention performance improvement.

wenxindongwork (approved):

tpu_inference/core/sched/dp_scheduler.py
wyzhang: Mark conversation resolved.

tpu_inference/envs.py
wyzhang: Done. Updated to call out why it is necessary for achieving high throughput.

tpu_inference/runner/tpu_runner.py
```diff
@@ -1664,10 +1669,22 @@ def build_block_table_host(kv_cache_gid: int) -> None:
...
         # signature on those models is unchanged).
         if self.kv_cache_config.has_mamba_layers:
-            # Per-request mamba state slot ids; copy to keep the InputBatch's
-            # CPU buffer free for the next step's bookkeeping.
-            mamba_state_indices_cpu = self.input_batch.mamba_state_indices_cpu.copy(
-            )
+            # Reorder mamba_state_indices per DP rank (like block_tables)
+            # and convert global slot ids to rank-local indices so they
+            # index correctly into the per-rank shard of the mamba state.
+            local_slots = self.input_batch._mamba_local_slots
```
kyuyeunk, line 1686: was the mamba state indices just replicated across all dp ranks? and i would assume the changes in gdn layer is to ensure correct indices are fetched for each dp rank?

kyuyeunk (approved):

tpu_inference/runner/tpu_runner.py
qizzzh: Yes to the first part - without this PR, the mamba state pool is sharded along the ATTN_DATA axis (like block_tables), but mamba_state_indices was passed as-is without the per-rank reordering that block_tables already get. So each DP rank would see the global flat indices rather than its own rank-local view, causing it to index into wrong slots.

The fix here mirrors what _prepare_inputs already does for block_tables: reorder per DP rank and convert global slot ids → rank-local indices via global_slots % local_slots. The GDN layer itself (gdn_attention_op.py) didn't need changes - it just reads attn_metadata.mamba_state_indices, which is now correctly reordered before being placed on device.

PR #2587, opened 2026-05-12
Support Yield Kernel Tuning Job to Higher Priority Jobs

patrickji2014:
# Description
In existing kernel tuning infra, when the tuning job(defined by a tuning bucket) is started in a buildkite agent, it will continue to run until it finishes. However, it's hard to select a bucket size to make the tuning job running time predictable which makes it hard to minimize the tuning jobs' impact on CI/CD works. 

To solve the problem above, we set a maximum kernel job running time, by default 20 minutes, to the kernel tuner. When the jobs run more than the threshold it will stop and generate a sub job waiting for buildkite to schedule it again. In this case, when there are higher priority jobs, they will not be starved. 

This PR includes changes:
1. Add kernel_tuning_job_priority and kernel_tuning_max_execution_minutes flags to support yielding the kernel tuning job to higher priority job
2. Inside kernel_tuner_base, check the elapse running time of the job and generate yml file for buildkite to schedule the job again if the running time larger than the kernel_tuning_max_execution_minutes.
3. MISC code refactor to make variables encapsulated in the tuner base class to remove excessive argument passing, which make it hard to manage when generate the yml using python and f-str.

# Tests

Test in BuildKite
https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/80/canvas

# Checklist


.buildkite/pipeline_kernel_tuning.yml
```diff
@@ -75,6 +75,8 @@ steps:
           --case_set_id=\${KERNEL_TUNING_CASE_SET_ID} \
           --run_id=\${KERNEL_TUNING_RUN_ID} \
           --tpu_cores=\${KERNEL_TUNING_TPU_CORES} \
+          --job_priority=\${KERNEL_TUNING_JOB_PRIORITY:-10} \
```
QiliangCui2023: Fallback default here is positive `:-10`, but the configured value in `.buildkite/scripts/configs/pipeline_config.sh:24` is `-10`. If `KERNEL_TUNING_JOB_PRIORITY` is ever unset (bootstrap not sourced), priority silently flips to `+10` — same magnitude, opposite sign. Suggest `${KERNEL_TUNING_JOB_PRIORITY:--10}` here (and the same fix for `PRIORITY_KERNEL_TUNING:-10` in `.buildkite/scripts/run_in_docker.sh:60`) so the default matches the configured value.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -355,34 +412,46 @@ def run(self, tuning_key: TuningKey, tunable_params: TunableParams,
...
-            caseset_id, run_id, begin_case_id, end_case_id)
+            self.case_set_id, self.run_id, begin_case_id, end_case_id)
         all_configs = self.storage_manager.get_bucket_configs(
-            caseset_id, begin_case_id, end_case_id)
+            self.case_set_id, begin_case_id, end_case_id)
 
         bucket_start_perf = time.perf_counter()
         results_buffer = []
+        bucket_fully_processed = True
+        last_processed_case_id = None
```
QiliangCui2023: `last_processed_case_id = None` plus the terminal log `last_processed_case_id + 1` at L525 will raise `TypeError` if the loop never enters (e.g. `begin_case_id == end_case_id`). It works today because the time-budget check at L444 cannot trip on the first iteration, but the invariant is fragile. Cheap fix: `last_processed_case_id = begin_case_id - 1`. Also note that the assignment at L454 happens before the case is actually processed and runs even on the `cid in processed_ids` skip path — the name reads as "last attempted", not "last processed".

tools/kernel/tuner/v1/example_kernel_tuner.py
```diff
@@ -51,13 +51,31 @@ class ExampleKernelTuner(KernelTunerBase):
...
-    def __init__(self, storage_manager, tpu_queue_multi=None):
+    def __init__(self,
+                 storage_manager,
+                 case_set_id=None,
+                 run_id=None,
+                 case_set_desc=None,
+                 tpu_version=None,
+                 tpu_cores=None,
+                 tpu_queue_multi=None,
+                 run_locally=None,
```
QiliangCui2023: Subclass declares `run_locally=None`, but the base in `tools/kernel/tuner/v1/common/kernel_tuner_base.py:116` annotates `run_locally: bool = False`. It works by coincidence (`not None` evaluates the same as `not False`), but the type is wrong and would regress if the base ever switched to `if self.run_locally is False`. Either drop the param and let the base default fire via `super().__init__(...)`, or type it as `bool = False`. Same issue at `tools/kernel/tuner/v1/rpa_v3_kernel_tuner.py:170`.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -96,7 +107,15 @@ def __init__(self,
...
-                 tpu_queue_multi: str = None):
+                 case_set_id: str = None,
+                 run_id: str = None,
+                 case_set_desc: str = None,
+                 tpu_version: str = None,
+                 tpu_cores: int = None,
+                 tpu_queue_multi: str = None,
+                 run_locally: bool = False,
+                 job_priority: int = -10,
+                 max_execution_minutes: int = 20):
```
QiliangCui2023: The constructor now takes 11 kwargs, and both subclasses (`ExampleKernelTuner.__init__`, `RpaV3KernelTuner.__init__`) re-declare all of them only to forward unchanged to `super().__init__(...)`. Adding the next field will mean editing every subclass again. Consider `**kwargs` forwarding, or pulling the run-scoped fields (`case_set_id`, `run_id`, `case_set_desc`, `tpu_version`, `tpu_cores`, `run_locally`, `job_priority`, `max_execution_minutes`) into a small `TunerRunConfig` dataclass so subclasses only declare their own kernel-specific params.

.buildkite/pipeline_kernel_tuning.yml
patrickji2014: thanks. updated

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: i updated the last_processed_case_id to be begin_case_id - 1 but keep the naming since for the following processed_ids check if it's true, then it's still processed. marked as resolved now but feel free to reopen if you have concern. thanks.

tools/kernel/tuner/v1/example_kernel_tuner.py
patrickji2014: Updated to use RunConfig and TunerConfig. Only one place to have default value now.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: updated to use RunConfig and TunerConfig.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -81,196 +115,225 @@ class KernelTunerBase(ABC):
...
-                 tpu_queue_multi: str = None):
-        assert tuning_key_class is not None, "tuning_key_class must be specified"
-        assert tunable_params_class is not None, "tunable_params_class must be specified"
-        assert storage_manager is not None, "storage_manager must be specified"
-        assert kernel_tuner_name is not None, "kernel_tuner_name must be specified, which will be used as the identifier for this kernel tuner in the Buildkite pipeline generation and execution. It should match the key in the KERNEL_TUNER_REGISTRY in kernel_tuner_runner.py to ensure the correct kernel tuner is called during execution."
-        self.tuning_key_class = tuning_key_class
-        self.tunable_params_class = tunable_params_class
-        self.storage_manager = storage_manager
+                 tuner_config: TunerConfig = TunerConfig(),
+                 run_config: RunConfig = RunConfig()):
```
QiliangCui2023: Mutable default arguments: `TunerConfig()` and `RunConfig()` are evaluated once at function-definition time, so every caller relying on the defaults shares the same instance. `_init_case_set` later mutates `self.run_config.case_set_id`, so two tuners constructed with default configs would silently stomp on each other's state. Default these to `None` and instantiate fresh inside the body.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -81,196 +115,225 @@ class KernelTunerBase(ABC):
...
+        assert tuner_config.tuning_key_class is not None, "tuning_key_class must be specified"
+        assert tuner_config.tunable_params_class is not None, "tunable_params_class must be specified"
+        assert tuner_config.kernel_tuner_name is not None, "kernel_tuner_name must be specified, which will be used as the identifier for this kernel tuner in the Buildkite pipeline generation and execution. It should match the key in the KERNEL_TUNER_REGISTRY in kernel_tuner_runner.py to ensure the correct kernel tuner is called during execution."
+        self.storage_manager = LocalDbManager if run_config.run_locally else SpannerStorageManager(
+        )
```
QiliangCui2023: `LocalDbManager` is the class, not an instance — missing `()`. When `run_locally=True`, `self.storage_manager` is bound to the class itself, and every subsequent call (`self.storage_manager.case_set_id_exists(...)`, `add_tuner_case`, ...) will fail at runtime because they expect an instance. Should be `LocalDbManager()`.

QiliangCui2023 (changes requested):
Two correctness bugs in the refactored `KernelTunerBase.__init__` — see inline comments. Both are easy fixes.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: remove the redundant code in the _init_caes_set as we enforce case_set_id to be set through command line flag now.

QiliangCui (approved):

QiliangCui2023 (approved):

PR #2572, opened 2026-05-10
Gemma-4 E2B-it on JAX path (PLE + KV-share + double-wide MLP)

QiliangCui2023:
## Summary

Adds JAX-native support for `Gemma4ForConditionalGeneration` text-only inference (E2B-it tested; the E-family architecture extends to E4B with the same surface).

Architectural extensions on top of the standard Gemma3-style stack:

- **Per-Layer Embeddings (PLE)** — per-layer scaled embeddings projected and combined with text embeds via the vllm-pytorch reference.
- **KV-share** — the last `num_kv_shared_layers` layers reuse the K/V cache slot of the last preceding same-type layer (sliding vs full). Wired via `kv_sharing_target_layer_name` on the Attention module + the kv_cache_manager's `shared_kv_cache_layers` redirect.
- **Double-wide MLP** — KV-shared layers get `2 * intermediate_size` when `text_config.use_double_wide_mlp` is set.
- **Weight loader** — PLE weights (`embed_tokens_per_layer`, `per_layer_model_projection`, `per_layer_projection_norm`); skip `audio_tower` / activation-calibration scalars; full Q/K/V load for shared layers (K/V is computed but the kernel skips the cache write).

### `tpu_inference/models/common/kv_share.py` (new)

`compute_kv_share_map(text_config)` — generic helper that derives `{shared_idx: source_idx}` from HF `text_config.num_kv_shared_layers` + `layer_types`. The implementation operates only on HF text-config attributes (no model-implementation dependencies), so `runner/kv_cache_manager.py` can call it without pulling in any actual model code. Reused inside `gemma4.py` itself for the per-layer `is_kv_shared_layer` decision. Any future JAX model whose HF config uses the same convention picks it up automatically.

## Test plan

- `tests/models/jax/test_gemma4_ple.py` — PLE structural + numerical tests.
- `tests/models/jax/test_gemma4_kv_share.py` — `compute_kv_share_map` edge cases + KV-share layer structure.
- `tests/models/jax/test_gemma4.py` — extended for E2B.
- `.buildkite/models/google_gemma-4-E2B-it.yml` — nightly serving check on v6e (TP=1, chat-templated offline_inference + chat-templated gsm8k + text-only `vllm bench serve`).

Manual coherence verification (offline_inference on v6e single chip, chat-templated):

| Prompt | Generated |
|---|---|
| Hello, my name is | "Hello! It's nice to meet you. What is your name? 😊" |
| The capital of France is | "**Paris**." |
[...]

PR #2601, opened 2026-05-13
[kernel] RPA v3: add update_kv_cache=False for KV-shared layers

QiliangCui:
## Summary

Adds an `update_kv_cache` flag to the v3 ragged-paged-attention kernel so KV-shared layers can read from a previously-written cache slot without overwriting it.

When `update_kv_cache=False`:
- Skip the cache write at the end of the bq loop (`kernel.py:990-997`).
- Read the full `kv_len` from cache instead of splitting `q_len` out from the new K/V (`kernel.py:568-579`).

This mirrors vllm-pytorch's `unified_attention`, which reads only from `key_cache` / `value_cache` regardless of the layer's own k/v projections.

## Why

Models with KV-shared layers (HF text_config's `num_kv_shared_layers > 0`; e.g. Gemma-4 E2B / E4B) need this. For a shared layer, the source layer ran first in the same forward and already populated the cache slot. The shared layer's locally-computed K/V is unused; without this flag the kernel would overwrite the source layer's cache slot.

The kwarg is plumbed through `attention_interface.attention()` and `sharded_ragged_paged_attention()`. The hd64 path doesn't support the flag and isn't needed by any current model — the kwarg is only forwarded to the non-hd64 kernel.

## Tests

3 new unit tests in `tests/kernels/ragged_paged_attention_kernel_v3_test.py` (folded into the existing v3 test file):
- `test_kv_share_prefill_input_kv_is_ignored`
- `test_kv_share_decode_input_kv_is_ignored`
- `test_kv_share_chunked_prefill_input_kv_is_ignored`

Each runs the shared-layer call twice with two different input K/V seeds and asserts the output is identical (because input K/V is ignored) and that the cache is not mutated by the shared layer's call.

Run locally:
\`\`\`
pytest tests/kernels/ragged_paged_attention_kernel_v3_test.py -k kv_share -v
\`\`\`

## Test plan
- [x] Unit tests pass on the new KV-share paths
- [x] No regression on existing v3 RPA tests (full file still passes)
- [ ] Pre-commit CI green

## Stack
This is part 1 of a 3-PR split of #2572. Follow-ups:
- kv_cache_manager None-handling (independent, can land in any order)
- Gemma-4 E2B model (depends on this + the kv_cache_manager PR)

### DIFF
--- a/tests/kernels/ragged_paged_attention_kernel_v3_test.py
+++ b/tests/kernels/ragged_paged_attention_kernel_v3_test.py
@@ -593,10 +593,10 @@ class RaggedPagedAttentionKernelTest(jtu.JaxTestCase):
                              dtype=np.float32)).astype(dtype)
 
         cache_data = jnp.array(
-            rng_cache.random((pages_per_seq * page_size,
-                              num_kv_heads_x2 // kv_packing, kv_packing,
-                              padded_hd),
-                             dtype=np.float32)).astype(dtype)
+            rng_cache.random(
+                (pages_per_seq * page_size, num_kv_heads_x2 // kv_packing,
+                 kv_packing, padded_hd),
+                dtype=np.float32)).astype(dtype)
         cache_pages = cache_data.reshape(pages_per_seq, page_size,
                                          num_kv_heads_x2 // kv_packing,
                                          kv_packing, padded_hd)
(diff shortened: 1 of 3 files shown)


bythew3i (approved):
LGTM for kernel part

PR #2605, opened 2026-05-13
Feat: Enable Qwen3-VL-Embedding-8B Support via meta tensor materialization and Deepstack device patching

anthonsu:
#### **Summary**
This PR introduces full end-to-end support for the **Qwen3-VL-Embedding-8B** model on TPU. It addresses the current **Qwen3-VL** gaps between vLLM V1's lazy loading mechanism (which uses `meta` device tensors) and *Qwen3-VL-Embedding* weight loading and pooling workflow. 
Additionally, it resolves device mismatch errors in the Qwen3-VL multimodal components when running in Torchax mode.

#### **Key Changes**

**1. Universal Weight Materialization in `unquantized.py`**
*   Implemented a robust **"Attribute Promotion"** strategy at the layer level to intercept `meta` tensors before JAX sharding and conversion begin.
*   This resolves the `NotImplementedError: Cannot copy out of meta tensor` issue encountered during initialization.

**2. Deepstack Device Patching in `qwen3_vl_patcher.py`**
*   Resolved a `TypeError` where model hidden states (on TPU) were being added to deepstack buffers.
*   Refactored the patcher to manually force the `deepstack_input_embeds` buffer list to the correct TPU device (via Torchax/JAX) to address the issue caused by ModelForEmbedding adapter.

**3. Functional Scoping (Pooling Guard)**
*   Ensured the materialization hardening logic is scoped to `is_pooling`. This ensures zero logic changes or performance impact for standard generative models.

#### **Validation & Correctness**
*   **E2E Verified on v7 Hardware**: Successfully validated Qwen3-VL-Embedding end-to-end on a `tpu7-8` node.
*   **Chunked Prefill Parity**: Confirmed that `pooling_states` are correctly preserved and accumulated across step boundaries during multi-step multimodal embedding tasks.

- [mm embedding/pooling e2e test result](https://paste.googleplex.com/4553021745922048)
- [qwen3-vl-embedding-8b server log - success](https://paste.googleplex.com/4989306268680192)

NOTE: I will create a follow-up(another) PR to introduce the formal Nightly CI tests for this model.

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -21,6 +21,7 @@ from jax.sharding import Mesh, NamedSharding, PartitionSpec
 from torch.nn import Parameter
 from torchax.interop import jax_view, torch_view
 from torchax.ops.mappings import t2j
+from vllm.config import get_current_vllm_config
 from vllm.model_executor.layers import linear as vllm_linear
 from vllm.model_executor.layers.attention import Attention
 from vllm.model_executor.layers.fused_moe import (FusedMoE, FusedMoEConfig,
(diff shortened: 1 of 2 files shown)


vipannalla:
@kwang3939, @lk-chen, please take a look

lk-chen:
Hi @anthonsu could you explain why `NotImplementedError: Cannot copy out of meta tensor` was observed for this model, while not for others?

anthonsu:
> Hi @anthonsu could you explain why `NotImplementedError: Cannot copy out of meta tensor` was observed for this model, while not for others?

Hi @lk-chen ,

Great question! Here are more details:

- Regarding the `Instruct` model, their weight keys perfectly match the native loader, meaning all tensors are fully materialized well before entering the t2j conversion phase.

- Regarding `Embedding` model, the embedding adapter layer alters the module hierarchy/paths, and vLLM V1's lazy loading mechanism relies on strict key matching to materialize weights. So the standard loader misses certain tensors, leaving them as meta device placeholders. The JAX eager conversion during initialization cannot wait for this lazy loading and strictly requires concrete data, thus throwing the error.

So we need a precise guard (`tensor.device == torch.device("meta")`) to intercept these specific residues. **It safely unblocks the embedding variants without affecting any models that have already loaded their weights correctly.**

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -69,6 +70,33 @@ def _load_weight_for_layer(
     """
     tensor = getattr(layer, param_name)
 
+    if tensor.device == torch.device("meta"):
```
lk-chen, line 73: @kyuyeunk for change in layser/vllm/

lk-chen (approved):
LGTM, would like @kyuyeunk to take a look at torchax change, also cc @muskansh-google for qwen3_vl_patcher change

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -239,3 +242,17 @@ def is_qwen3_vl(vllm_model) -> bool:
 def maybe_apply_qwen3_vl_patches(vllm_model):
     if is_qwen3_vl(vllm_model):
         apply_qwen3_vl_patches(vllm_model)
+
+        if hasattr(vllm_model, "deepstack_input_embeds"):
```
muskansh-google, line 246: Hi @anthonsu, can you please share the error you were getting and why this fix is required?

anthonsu: @muskansh-google , the error was `TypeError: cannot add a JAX array to a list of tensors `(or similar device-level mismatch) when running in Torchax mode.

The standard Qwen3-VL patcher expects `deepstack_input_embeds` to be a buffer of standard tensors. However, the embedding path bypasses some of the standard generative logic, leading to a situation where JAX-backed hidden states are appended to these buffers without proper device conversion.

This hardening is scoped to cases where `deepstack_input_embeds` exists, ensuring no impact on standard generative models while unblocking the embedding workflow.

PR #2608, opened 2026-05-13
[Spec decoding] make spec decoding compatible with async scheduling

gxd3:
Make spec decoding compatible with async scheduling.

The key changes are in `tpu_inference/runner/tpu_runner.py`.

Due to async scheduling, in the n-th step's schedule / execute_model, host does not have the n-1-th step's sampled tokens and draft tokens, we'll need to use `_substitute_placeholder_token` HLO to update `input_ids`, `spec_decode_metadata.draft_token_ids.`

Also, `seq-len` & `position` set by the host in the n-th step is not accurate, because host is not aware of num-accepted-draft-tokens in the n-1-th step yet. Need to use `_subtract_num_rejected_tokens_fn` HLO to update `seq-len` & `position` accordingly before calling the target model.

Note that spec-decoding is NOT yet compatible with attn-dp; defer that support to later PR.

xprof before: https://screenshot.googleplex.com/9a5ehQMMtgw8uBC
xprof with this PR: https://screenshot.googleplex.com/66jVTnShKKpahpA (no TPU bubble now)

# Tests
unit test.
added parameterized test in `tests/e2e/test_speculative_decoding.py` for E2E coverage.
Verified acceptance rate of spec bench dataset: https://paste.googleplex.com/6748805522391040

### DIFF
--- a/tpu_inference/runner/speculative_decoding_manager.py
+++ b/tpu_inference/runner/speculative_decoding_manager.py
@@ -25,7 +25,6 @@ from vllm.v1.spec_decode.ngram_proposer import NgramProposer
 from tpu_inference.runner import utils as runner_utils
 from tpu_inference.runner.utils import SpecDecodeMetadata
 from tpu_inference.spec_decode.jax.eagle3 import Eagle3Proposer
-from tpu_inference.spec_decode.jax.utils import extract_last_sampled_tokens
 from tpu_inference.utils import device_array
 
 if TYPE_CHECKING:
@@ -53,13 +52,18 @@ class SpeculativeDecodingManager:
         self,
         sampled_output: jnp.ndarray,
         logits_indices_selector: np.ndarray,
+        last_sampled_token_id: jnp.ndarray,
+        num_rejected_tokens: jnp.ndarray,
         discard_sampled_tokens_req_indices: list,
         aux_hidden_states: Optional[tuple[jnp.ndarray, ...]],
         attn_metadata: AttentionMetadata,
+        async_scheduling: bool,
         spec_decode_metadata: Optional[SpecDecodeMetadata],
         scheduler_output: Optional[VllmSchedulerOutput] = None,
         input_ids: Optional[jnp.ndarray] = None,
     ) -> None:
+        if async_scheduling:
+            assert self.runner.speculative_config.method == "eagle3", "async scheduling is only supported with eagle3 spec decoding"
         if self.runner.speculative_config.method == "ngram":
             assert isinstance(self.runner.drafter, NgramProposer)
             # For n-gram based proposer, the drafter run on host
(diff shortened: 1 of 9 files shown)


tpu_inference/spec_decode/jax/utils.py
```diff
@@ -82,3 +82,10 @@ def extract_last_sampled_tokens(
                                   (0, max_num_seq - batch_size),
                                   constant_values=0)
     return last_sampled_tokens, num_rejected_tokens
+
+
+@jax.jit
+def concat_last_sampled_tokens_and_draft_tokens(last_sampled_tokens,
```
Lumosis, line 88: We should precompile this as well

tpu_inference/runner/tpu_runner.py
```diff
@@ -1096,31 +1136,48 @@ def _sample_from_logits(
...
             placeholder_req_id_to_index: dict[
                 str, int] = self._update_placeholder(
                     discard_sampled_tokens_req_indices, request_seq_lens,
-                    logits_indices_selector)
+                    spec_decode_metadata, logits_indices_selector)
+
+            spec_decode_next_tokens = None
+            if self.speculative_config:
+                assert spec_decode_last_sampled_token_id is not None
+                spec_decode_next_tokens = concat_last_sampled_tokens_and_draft_tokens(
```
Lumosis: Should we guard it with maybe_forbid_compile?

tpu_inference/runner/speculative_decoding_manager.py
```diff
@@ -139,10 +139,16 @@ def propose_eagle3_draft_token_ids(
             last_token_indices=last_token_indices,
             target_hidden_states=target_hidden_states,
         )
-        draft_token_ids = np.array(draft_token_ids)
-        if draft_token_ids.ndim == 1:
-            draft_token_ids = np.expand_dims(draft_token_ids, axis=-1)
-        return draft_token_ids.tolist()
+
+        if not async_scheduling:
```
Lumosis: nit: use `if async_scheduling` can improve readability. http://go/tott/764

tpu_inference/runner/utils.py
```diff
@@ -558,16 +558,14 @@ class SpecDecodeMetadata:
...
 
-    draft_lengths_cpu: Any = field(init=False)
+    draft_lengths_cpu: Any = field(init=False, default=None)
 
 
 def host_extract_sampled_tokens(
         runner, spec_decode_metadata: Optional[SpecDecodeMetadata],
         sampled_output: jnp.ndarray, logits_indices_selector: np.ndarray,
-        discard_sampled_tokens_req_indices: list):
+        discard_sampled_tokens_req_indices: list, num_reqs):
```
Lumosis: nit: `num_reqs: int`

tpu_inference/runner/speculative_decoding_manager.py
```diff
@@ -53,30 +52,32 @@ def propose_draft_token_ids(
...
         discard_sampled_tokens_req_indices: list,
         aux_hidden_states: Optional[tuple[jnp.ndarray, ...]],
         attn_metadata: AttentionMetadata,
+        async_scheduling: bool,
         spec_decode_metadata: Optional[SpecDecodeMetadata],
         scheduler_output: Optional[VllmSchedulerOutput] = None,
         input_ids: Optional[jnp.ndarray] = None,
     ) -> None:
+        if async_scheduling:
+            assert self.runner.speculative_config.method == "eagle3", "async scheduling is only supported with eagle3 spec decoding"
```
Lumosis: If we want to support MTP as well, we should use `self.runner.speculative_config.use_eagle()`

Lumosis (changes requested):

tpu_inference/runner/speculative_decoding_manager.py
gxd3: Done.

tpu_inference/runner/utils.py
gxd3: Done.

tpu_inference/runner/speculative_decoding_manager.py
gxd3: Done.

tpu_inference/runner/tpu_runner.py
gxd3: added.

tpu_inference/spec_decode/jax/utils.py
gxd3: added.
(this function only have 1 possible input shapes, so i didn't add precompile previously. But just added.)

Lumosis (approved):

PR #2612, opened 2026-05-13
[FP8/CT] Use TPU to do MoE weight re-quantization

jrplatin:
# Description

This PR builds on top of the work of https://github.com/vllm-project/tpu-inference/pull/2028, namely:

Moves FP8 MoE weight requantization from CPU to TPU by default, with a fix for the HBM regression that caused the original feature to be reverted.

**Root cause of the revert:** @jax.jit-compiled requantization programs reserve HBM (bytes_reserved) for XLA program buffers. These reservations persisted after weight loading completed, permanently consuming HBM that would otherwise be allocated to KV cache.

**Fix:** Call jax.clear_caches() in tpu_runner.py after model loading completes. This frees XLA program reservations from the requantization JIT, since those compiled programs are never needed again during inference.

**Update (v2):** Removed the MOE_REQUANTIZE_ON_TPU env flag and CPU fallback path. TPU requantization is now always enabled, no branching logic, no config needed.

## Changes

1. `moe_weights.py`: TPU requantization path using shard_map + lax.scan (replaces the old CPU dequant/quantize path)
2. `jax/quantization/fp8.py`: Shard FP8 weights to TPU before calling process_fp8_moe_weights
3. `vllm/quantization/fp8.py`: Same for vLLM code path
4. `tpu_runner.py`: Unconditionally call jax.clear_caches() after weight loading

# Testing

### Accuracy Validation

```
|Tasks |Version|Filter|n-shot|  Metric   |   |Value |   |Stderr|
|------|------:|------|-----:|-----------|---|-----:|---|-----:|
|aime26|      0|none  |     0|exact_match|↑  |0.9333|±  |0.0463|

```

Matches result [here](https://docs.google.com/spreadsheets/d/1j0mQHW7_xMbWxRHx0YViJw-5Fo_zug39oc8PomHEz3o/edit?resourcekey=0-TLB4B9maYYRMK-M98dRO8w)

### Loading time
| Branch | Run Type | Loading Weights Time (s) | Requant Time (s) |
| :--- | :--- | :--- | :--- |
| **Feature branch** | **Cold** | 519.52 | 48 |
| **Feature branch** | **Hot** | 61.77 | 46 |
| **Main** | **Cold** | 522.38 | 156 |
| **Main** | **Hot** | 62.26 | 162 |

CMD:
```
lm_eval   --model vllm   --model_args '{"pretrained": "Qwen/Qwen3.5-397B-A17B-FP8", "attention_backend": "FLASH_ATTN", "gdn_prefill_backend": "triton", "kv_cache_dtype": "fp8", "tensor_parallel_size": 8, "max_model_len": 35000, "max_num_batched_tokens": 4096, "block_size": 256, "limit_mm_per_prompt": {"image": 0, "video": 0}, "enable_thinking": false}'   --tasks aime26   --apply_chat_template   --log_samples   --output_path eval_logs_397b_aime26    --seed 42  --system_instruction "Put your final answer within \boxed{}."
```

## Skipped Requantization Path

**Summary:** on-par accuracy and ~4x reduction in MoE processing (no requant) speed

### Accuracy Validation
Both `main` and feature branch scored the same:
```
humaneval (base tests)
pass@1:	0.933
humaneval+ (base + extra tests)
pass@1:	0.884
```

### Loading time

| **Branch**       | **MoE Processing Time** |
|------------------|-------------------------|
| `main`           | 21s                      |
| `feature branch` | 4s                     |

CMDs:
```
# Server
DISABLE_WEIGHT_REQUANTIZATION=1 MODEL_IMPL_TYPE=vllm nohup vllm serve --seed 42 --model Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8 --max-num-batched-tokens 1028 --max-num-seqs 128 --no-enable-prefix-caching --tensor-parallel-size 2 --enable-expert-parallel  &

# Client
  evalplus.evaluate --model "Qwen/Qwen3-Coder-30B-A3B-Instruct-FP8" --dataset humaneval --backend openai -base-url http://localhost:8000/v1 --greedy  
```

## Notes

- jax.clear_caches() clears all JIT compilation caches. At the call site (after weight loading, before inference compilation), only loading-related JIT artifacts exist. Inference functions get compiled on first use regardless.
- Large model validation (Qwen3-235B, DeepSeek-R1) requires v7x access. Qwen3-235B OOMs on v6e-8 on main branch (pre-existing).
- 
# Checklist


jrplatin:
@rohan-reddy Does this look good to you? Want to make sure you get the credit you deserve!

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,6 +645,56 @@ def shard_moe_weights(
...
+    Args:
+        weights: FP8 MoE weights (typically on CPU).
+        mesh: The TPU device mesh for inference.
+        source_mesh: The mesh the weights currently reside on (e.g.
+            cpu_mesh()). None when weights are plain CPU arrays.
+
+    Returns:
+        FusedMoEWeights sharded across TPU devices.
+    """
+    expert_axis = ShardingAxisName.EXPERT
```
gxd3: this function only support EP; but the caller seem more general (may request for different sharding?)

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,6 +645,56 @@ def shard_moe_weights(
...
+    if isinstance(expert_axis, str):
+        if expert_axis in mesh.axis_names:
+            shard_axis = expert_axis
+        else:
+            shard_axis = mesh.axis_names[0]
```
gxd3: this should not happen?

kyuyeunk:
does this code work for just fp8 or does it also work for all the other code (like mxfp4, nvfp4 requantization) as well?

jrplatin:
@kyuyeunk I want to start with FP8 and expand to other formats eventually just to limit the blast radius of this PR

kyuyeunk:
> @kyuyeunk I want to start with FP8 and expand to other formats eventually just to limit the blast radius of this PR

sgtm. but wanted to make sure the high level design of the code itself is agonistic to any specific checkpoint dtype so that it can be easily expanded to others.

jrplatin:
@kyuyeunk Actually, changed my mind -- I decided to refactor it to be general, so I added support for CompressedTensorsW8A8 as well, since that and FP8 are the two most common paths right now

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: I think this should be fine -- the real final sharding for execution (which may be TP or EP) is applied later in shard_moe_weights

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Good point, added an assertion to ensure this!

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,18 +645,70 @@ def shard_moe_weights(
     return weights
 
 
+def shard_moe_weights_to_tpu(
```
gxd3, line 661: nit: will `shard_moe_weights_to_tpu` always used along with `process_quantized_moe_weights`?

If so, maybe consider call `shard_moe_weights_to_tpu` within `process_quantized_moe_weights` instead of let the caller call `shard_moe_weights_to_tpu` first then call `process_quantized_moe_weights`

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,18 +645,70 @@ def shard_moe_weights(
...
+    ep_sharding = NamedSharding(mesh, P(shard_axis))
+
+    result_fields = {}
+    for field in fields(FusedMoEWeights):
+        key = field.name
+        weight = getattr(weights, key)
+        if weight is not None:
+            result_fields[key] = general_device_put(weight,
+                                                    ep_sharding,
+                                                    source_mesh=source_mesh)
```
gxd3: nit: lots of exisiting caller of `general_device_put` does not specify `source_mesh` argument, is it necessary?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,18 +645,70 @@ def shard_moe_weights(
...
+    "requant_block_size",
 ))
-def process_fp8_moe_weights(
+def process_quantized_moe_weights(
     weights: FusedMoEWeights,
     moe_backend: MoEBackend,
     mesh: Mesh,
     activation: str,
     weight_block_size: tuple[int, ...] | None = None,
+    desired_quant_dtype: jnp.dtype | None = None,
```
gxd3, line 1047: most callers leave these two arguments as not set, is it intentional?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
...
+    else:
+        w13_block_size = w2_block_size = requant_block_size
+    hidden_size = align_to(orig_hidden_size, w13_block_size)
+    intermediate_size = align_to(orig_intermediate_size, w2_block_size)
+
+    inter_pad = intermediate_size - orig_intermediate_size
+    hidden_pad = hidden_size - orig_hidden_size
+
+    # Determine which mesh axis the expert dim is sharded across.
+    expert_axis = ShardingAxisName.EXPERT
```
gxd3: nit: this code block seem repeated with a part of `shard_moe_weights_to_tpu`, maybe have a shared util function?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
...
+    out = FusedMoEWeights(
+        w13_weight=w13_q,
+        w13_weight_scale=w13_s,
+        w13_bias=w13_b,
+        w2_weight=w2_q,
+        w2_weight_scale=w2_s,
+        w2_bias=w2_b,
+    )
+
+    target_shardings = _get_moe_weight_shardings(out, moe_backend, mesh)
```
gxd3, line 1165: are these repeative of `shard_moe_weights` called subsequently?
https://screenshot.googleplex.com/6352bspthdhZ2A5

gxd3 (approved):

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,18 +645,70 @@ def shard_moe_weights(
...
+        assert expert_axis in mesh.axis_names, f"{expert_axis} not in mesh {mesh}!"
+        shard_axis = expert_axis
+    else:
+        if all(a in mesh.axis_names for a in expert_axis):
+            shard_axis = expert_axis
+        else:
+            shard_axis = mesh.axis_names[0]
+    ep_sharding = NamedSharding(mesh, P(shard_axis))
+
+    result_fields = {}
```
kyuyeunk: hmm doesn't feel like a good software pattern.

can you refer to this instead? https://github.com/vllm-project/tpu-inference/blob/c38cf684a1d02b0e74cded665f6bdd815874c795/tpu_inference/layers/common/process_weights/moe_weights.py#L623-L630

or another approach is to use`dataclasses.replace`: https://docs.python.org/3/library/dataclasses.html#dataclasses.replace

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
             w2_bias=weights.w2_bias,
         )
 
-    else:
+        out = process_moe_weights(
+            weights,
+            moe_backend=moe_backend,
+            w13_reorder_size=w13_reorder_size,
+            w13_interleave=w13_interleave,
+            disable_weight_requantization=envs.DISABLE_WEIGHT_REQUANTIZATION,
```
kyuyeunk: can you have this as an argument of the function?

in generally, it's a good idea to have jax jitted function to be stateless. otherwise, it can easily cause very confusing bugs.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
...
+        )
+
+        target_shardings = _get_moe_weight_shardings(out, moe_backend, mesh)
+        for field in fields(FusedMoEWeights):
+            key = field.name
+            weight = getattr(out, key)
+            if weight is not None:
```
kyuyeunk: nit / very personal preference: i think this approach is cleaner: https://github.com/vllm-project/tpu-inference/blob/c38cf684a1d02b0e74cded665f6bdd815874c795/tpu_inference/layers/common/process_weights/moe_weights.py#L623-L630

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
...
+                sharding = getattr(target_shardings, key)
+                setattr(out, key,
+                        jax.lax.with_sharding_constraint(weight, sharding))
+        return out
+
+    if desired_quant_dtype is None:
         if desired_quant_dtype_from_env := envs.MOE_REQUANTIZE_WEIGHT_DTYPE:
```
kyuyeunk: similar comment as above. let's have this as a function argument.

if you think that would require heavy refactoring, i would be fine with just having a wrapper function like this. benefit of this approach is that we can just take a single glance at wrapper function and figure out which envs can affect the behavior of the function. but in current approach, it's spread out across the function & hard to track.

```
def wrapper(a,b,c,d):
  func(a, b, c, d, envs.MOE_REQUANTIZE_WEIGHT_DTYPE, envs.MOE_REQUANTIZE_BLOCK_SIZE)
```

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -686,58 +752,232 @@ def process_fp8_moe_weights(
...
+        w13_weight=w13_q,
+        w13_weight_scale=w13_s,
+        w13_bias=w13_b,
+        w2_weight=w2_q,
+        w2_weight_scale=w2_s,
+        w2_bias=w2_b,
+    )
+
+    target_shardings = _get_moe_weight_shardings(out, moe_backend, mesh)
+    for field in fields(FusedMoEWeights):
```
kyuyeunk, line 1166: nit: similar comment as above.

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -630,18 +645,70 @@ def shard_moe_weights(
...
 @jax.jit(static_argnames=(
     "moe_backend",
     "mesh",
     "activation",
     "weight_block_size",
+    "desired_quant_dtype",
+    "requant_block_size",
 ))
-def process_fp8_moe_weights(
+def process_quantized_moe_weights(
```
kyuyeunk: it's a bit difficult to pinpoint exactly, but this function looks very very busy with so many components & diverging paths - there are mutiple exit points, nested functions. one way to make it cleaner is to break down this function into smaller one to modularize & have clear division of reponsibility between each modules.

kyuyeunk (changes requested):

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Yep, I think it's good to have the option to do so and doesn't add much overhead -- it's also the existing design pattern so I think it makes sense for commonality

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: I think we need it here because the source tensors are sometimes  created on a specific mesh (like cpu_mesh), which happens in the JAX path (jax/quantization/fp8.py) where weights are processed inside cpu_mesh_context()

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Added!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Updated!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Yep done!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Updated!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Yep, updated thanks!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Good point, will clean up!

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Refactored a bit to make things clearer -- let me know if it looks better!

kyuyeunk (approved):
nice! code looks much cleaner! approved!

PR #2622, opened 2026-05-14
[profiling] enable request ID tracing in the ModelForward pass

rushabh-46:
# Description

This PR introduces searchable request ID tracing directly to the host-side execution loop (`execute_model`) in the JAX `TPUModelRunner`. Previously, trace annotations for TPU execution did not contain any request metadata, making it difficult to correlate profiling timelines with specific user queries inside TensorBoard / XProf.

**Why is this change being made:**
To improve the observability and debuggability of TPU inference. By adding request IDs to traces, developers can trace the lifecycle of a specific request across different nodes and execution steps in XProf.

**Mechanism:**
*   **Request ID Extraction**: We extract the request IDs from **`self.input_batch.req_ids`** for the active requests in the batch. These request IDs are extracted and filtered using **`scheduler_output.num_scheduled_tokens > 0`**.
*   **ID Trimming**: Added a helper `trim_request_id_suffix` in `runner/utils.py` to remove batching and proxy suffixes (e.g., `-0-xxyyzz`) to ensure consistent base IDs across prefill and decode nodes.
*   **Trace Annotation**: We inject request ID metadata directly into the high-level `execute_model` trace.

# Tests

**Manual Verification (Benchmark & Profile):**
Verified the changes by deploying a custom image on GKE using the Qwen3-32B model and running the benchmark serving workload with profiling enabled:
```bash
vllm bench serve \
  --model Qwen/Qwen3-32B \
  --dataset-name random \
  --num-prompts 10 \
  --profile
```

**Results:**
The benchmark completed successfully (10/10 requests) and generated a trace profile (xplane.pb) containing the expected profiling data. Uploaded and visualized it in the Tensorboard XProf UI. The `ModelForward` trace annotations are visible along with the request IDs in the event metadata.

**Xprof UI Screenshot**
AqrK3dXFaPfpp5C

**Xprof snipit**
https://screenshot.googleplex.com/AqrK3dXFaPfpp5C

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -793,8 +793,14 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                                "after execute_model() returns None.")
         reqs = self.input_batch.num_reqs
         toks = scheduler_output.total_num_scheduled_tokens
+
+        req_id_kwargs = {}
+        if jax.profiler.TraceAnnotation.is_enabled():
+            req_id_kwargs = runner_utils.extract_request_ids_for_tracing(
+                self.input_batch, scheduler_output)
+
         with jax.set_mesh(self.mesh), jax.profiler.TraceAnnotation(
-                f"execute_model: {reqs} reqs, {toks} toks"):
+                f"execute_model: {reqs} reqs, {toks} toks", **req_id_kwargs):
             output = self._execute_model(scheduler_output,
                                          intermediate_tensors)
         return output
@@ -996,7 +1002,6 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         # NOTE: right now, mm model will use embeddings as the input,
         # but text-only model will use input_ids
         with self.maybe_forbid_compile:
-
             with set_forward_context(
                     None,
                     self.vllm_config,
--- a/tpu_inference/runner/utils.py
+++ b/tpu_inference/runner/utils.py
@@ -26,6 +26,53 @@ from tpu_inference import envs
 from tpu_inference.logger import init_logger
 from tpu_inference.runner.input_batch import InputBatch
 
+
+def trim_request_id_suffix(request_id: str) -> str:
+    """Trims the suffix from a request ID, keeping only the base ID.
+    
+    Example: cmpl-f0d75fed-c25e-4ccf-b369-9bd0226021b3-0-a9cb3cca 
+          -> cmpl-f0d75fed-c25e-4ccf-b369-9bd0226021b3
+    """
+    parts = request_id.split("-")
+    if len(parts) >= 6 and parts[0] == "cmpl":
+        return "-".join(parts[:6])
+    return request_id
+
+
+def extract_request_ids_for_tracing(
+    input_batch: InputBatch,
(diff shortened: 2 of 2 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -969,6 +969,25 @@ def _execute_model(
         # NOTE: right now, mm model will use embeddings as the input,
         # but text-only model will use input_ids
         with self.maybe_forbid_compile:
+            # Extract request IDs for tracing ONLY if tracing is enabled
+            req_id_kwargs = {}
+            if jax.profiler.TraceAnnotation.is_enabled():
```
bdubey: Are we checking this at other places where TraceAnnotation's are added? If not lets not add this. I dont think this should have any significant overhead.

rushabh-46: sure, removed the condition.

bdubey (approved):

tpu_inference/runner/tpu_runner.py
```diff
@@ -969,6 +969,24 @@ def _execute_model(
         # NOTE: right now, mm model will use embeddings as the input,
         # but text-only model will use input_ids
         with self.maybe_forbid_compile:
+            # Extract request IDs for tracing
+            req_id_kwargs = {}
+            try:
```
jrplatin: Can you move all of this into a utility function? `trim_request_id_suffix` can abstract away all of this IMO

tpu_inference/runner/tpu_runner.py
```diff
@@ -977,20 +995,22 @@ def _execute_model(
...
-                     attn_metadata,
-                     inputs_embeds,
-                     input_positions,
-                     tuple(self.layer_name_to_kvcache_index.items()),
-                     lora_metadata,
-                     intermediate_tensors,
-                     self.is_first_rank,
-                     self.is_last_rank,
-                 )
+                with jax.profiler.TraceAnnotation("ModelForward",
```
jrplatin: @gxd3 @kyuyeunk @gpolovets1 Do we know if adding this annotation will hurt perf?

tpu_inference/runner/tpu_runner.py
rushabh-46: Sure, abstracted into a dedicated `extract_request_ids_for_tracing` helper which internally calls `trim_request_id_suffix`. Kept them separate so `trim_request_id_suffix` is reusable for other tasks.

tpu_inference/runner/tpu_runner.py
gxd3: has same concern, specially if there are O(k) of concurrent requests.
Could we guard this with a environment variable (default off)?

Also consider move similar stats annotation to here as well: https://screenshot.googleplex.com/3y3CGVupvNKiUkZ
instead of multiple annotations spread in multiple places.

rushabh-46: Thanks, updated the existing TraceAnnotation and added the request_id metadata in itself. Also, wrapped this in the [jax.profiler.TraceAnnotation.is_enabled()](https://github.com/openxla/xla/blob/2db58190ae3807b454d209649ffa8205581472ac/xla/python/_profiler.pyi#L88) check which wraps the fast c++ atomic check.

gxd3 (approved):

jrplatin (approved):

PR #2632, opened 2026-05-15
RPA v3 + batched: support update_kv_cache=False (KV-share) on both kernels

QiliangCui2023:
## Summary

Enables `update_kv_cache=False` (KV-share: skip cache write, read full kv_len from the redirected cache slot) on **both** the default v3 RPA kernel and the experimental batched RPA kernel.

- Re-lands the v3-kernel support originally added by #2601 (reverted in #2628 because the wrapper-level kwarg-forwarding crashed the batched path with `TypeError: ragged_paged_attention() got an unexpected keyword argument 'update_kv_cache'`).
- Adds matching `update_kv_cache=False` semantics to the batched RPA kernel so the wrapper can safely forward the kwarg to whichever kernel is bound.

With this PR, KV-share models (Gemma-4 E2B / E4B, etc.) work on `MODEL_IMPL_TYPE=flax_nnx` regardless of whether `USE_BATCHED_RPA_KERNEL=1` is set.

The originally-reported nightly regression (req/s 4.76 → 2.94 on Gemma4-31B with `USE_BATCHED_RPA_KERNEL=1`) was the TypeError killing the server at trace time, misclassified by the harness as throughput. With this PR the server boots cleanly on both kernels.

### v3 default kernel (re-lands #2601)

- `kernels/ragged_paged_attention/v3/kernel.py`: new `update_kv_cache: bool = True` static argument on `ragged_paged_attention`. When `False`, read all `kv_len` from cache and skip the bq-loop's cache-write block.

### Batched RPA kernel (new)

The batched kernel already had the right hooks: its schedule kernel pre-computes a per-`(step, batch)` `do_writeback` flag, and `KVBufferedRef.copy_out` respects it. Adding `update_kv_cache=False` semantics is then just: in `compute_metadata::k_loop`, set `kv_left_frm_cache = kv_len` (instead of `max(kv_len - q_len, 0)`). `kv_left_frm_new` then becomes 0, `new_sz` is 0, and the existing
```python
do_writeback = where((new_sz > 0) & (q_idx == q_wb), 1, 0)
```
line naturally produces 0 — no extra gate needed.

`update_kv_cache: bool = True` is plumbed as a Python-level static through:
- `kernels/experimental/batched_rpa/wrapper.py::ragged_paged_attention` (new `static_argname`)
- `schedule.py::generate_rpa_metadata`
- `schedule.py::rpa_metadata_schedule_kernel`
- `schedule.py::compute_metadata` (gates the `kv_left_frm_cache` branch)

### Wrapper

`layers/common/attention_interface.py`:
- `sharded_ragged_paged_attention` and `attention()` accept `update_kv_cache: bool = True` and forward it to **both** kernels on the non-hd64 path.
- hd64 path raises `NotImplementedError` for `update_kv_cache=False` (no current model uses head_dim=64 + KV-share).

## Tests

- `tests/kernels/ragged_paged_attention_kernel_v3_test.py` — 3 v3-kernel KV-share unit tests (prefill, decode, chunked-prefill) verifying input k/v is unused and the cache is unmutated. (Re-landed from #2601.)
- `tests/layers/common/test_attention_interface.py`:
  - `test_sharded_rpa_forwards_update_kv_cache_when_not_hd64` — wrapper forwards the kwarg.
  - `test_batched_rpa_wrapper_accepts_update_kv_cache` — `inspect.signature` check that directly catches the original regression (batched wrapper missing the kwarg).
  - `test_sharded_rpa_rejects_update_kv_cache_false_on_hd64` — hd64 fails loud.

## E2B end-to-end (Gemma-4 E2B-it offline_inference on v6e)

With #2572 rebased on top of this PR (earlier integration BK [#448](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/448)):

| probe | kernel | errors | generated | sample |
|---|---|---|---|---|
| default RPA | `Using default RPA kernel` | 0 | 35/35 | "...The capital of France is **Paris**." |
| batched RPA | `Using experimental batched RPA kernel` | 0 | 35/35 | "...The capital of France is **Paris**." |

Outputs are byte-identical across the two paths.

## Stack

- This PR unblocks #2572 (Gemma-4 E2B JAX-path) — E2B's KV-shared layers work on both kernels.
- Supersedes #2628 (the full revert). After this PR merges, no further action needed on #2628.

### DIFF
--- a/tests/layers/common/test_attention_interface.py
+++ b/tests/layers/common/test_attention_interface.py
@@ -306,6 +306,126 @@ def test_sharded_ragged_paged_attention_gqa_incompatible_raises_error(
         )
 
 
+def _run_sharded_rpa_capturing_kwargs(monkeypatch, gqa_mesh, update_kv_cache):
+    """Helper: run `sharded_ragged_paged_attention` with a stubbed
+    `ragged_paged_attention` (the module-level binding) and a passthrough
+    `jax.shard_map`. Returns the kwargs forwarded by the closure to the
+    underlying kernel.
+    """
+    head_dim = 128  # non-hd64
+    num_kv_heads = 4
+    q = jnp.ones((TOTAL_TOKENS, NUM_HEADS, head_dim))
+    k = jnp.ones((TOTAL_TOKENS, num_kv_heads, head_dim))
+    v = jnp.ones((TOTAL_TOKENS, num_kv_heads, head_dim))
+    kv_cache = jnp.zeros((num_kv_heads, NUM_BLOCKS, BLOCK_SIZE, head_dim))
+    kv_lens = jnp.zeros((MAX_NUM_SEQS, ), dtype=jnp.int32)
+    page_indices = jnp.zeros((MAX_NUM_SEQS, MAX_BLOCKS_PER_SEQ),
+                             dtype=jnp.int32)
+    cu_q_lens = jnp.zeros((MAX_NUM_SEQS + 1, ), dtype=jnp.int32)
+    distribution = jnp.zeros((3, ), dtype=jnp.int32)
+
+    captured = {}
+
+    def fake_kernel(*_args, **kwargs):
+        captured.update(kwargs)
+        return jnp.ones_like(q), kv_cache
+
+    monkeypatch.setattr(
+        "tpu_inference.layers.common.attention_interface.ragged_paged_attention",
+        fake_kernel,
+    )
+
+    # Passthrough shard_map so the closure actually executes.
+    def passthrough_shard_map(inner_fn, **_):
+        return inner_fn
+
+    monkeypatch.setattr("jax.shard_map", passthrough_shard_map)
+
+    sharded_ragged_paged_attention(
+        mesh=gqa_mesh,
+        q=q,
+        k=k,
+        v=v,
+        kv_cache=kv_cache,
+        kv_lens=kv_lens,
+        page_indices=page_indices,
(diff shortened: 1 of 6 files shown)


tests/layers/common/test_attention_interface.py
```diff
@@ -306,6 +306,128 @@ def test_sharded_ragged_paged_attention_gqa_incompatible_raises_error(
...
+    crashed here with TypeError because the batched kernel didn't accept
+    the kwarg; the kernel now does."""
+    from tpu_inference import envs
+    monkeypatch.setattr(envs, "USE_BATCHED_RPA_KERNEL", True)
+
+    captured = _run_sharded_rpa_capturing_kwargs(monkeypatch,
+                                                 gqa_mesh,
+                                                 update_kv_cache=False)
+
+    assert captured.get("update_kv_cache") is False, (
```
lk-chen: The test is specific to one parameter. I feel we just need to test accuracy at USE_BATCHED_RPA=0/1 case.

cc @kyuyeunk

QiliangCui2023: the "test accuracy" test under .buildkite/models is a "support matrix" test. 

I think it accepts only one setup config - the config that we want to present to customer eventually. 

adding USE_BATCHED_RPA=1/0 seems hard - it will confuse the support matrix generation. 

If we want to do it, we need to create another test for it...

here, i think the unit test is lightweight and fast to verify the change..

lk-chen (approved):

PR #2636, opened 2026-05-15
Disable d2h by default and remove sudo as it not exist in buildkite

mrjunwan-lang:
# Description

remove sudo as it not exist in buildkite run docker

the failure seems a Kernel bug in recent change?? 
It is found by the random failure of "tpu7x E2E test for Single Host DCN-based P/D disaggregation" test in buildkite. 
refer: https://buildkite.com/tpu-commons/tpu-vllm-integration/builds/1170/canvas?sid=019e25ed-f5f6-4175-9dc4-e58e26c9d81b&tab=output

found:
if we set tp=1, it always success: log: https://paste.googleplex.com/6586999145365504
but if we set tp=2, it failed with error: https://paste.googleplex.com/5496803630907392.

We use tp=1 to unblock the others' test and vllm integration.

# Tests

commands to reproduce.

# Checklist


QiliangCui (approved):

tpu_inference/distributed/utils.py
```diff
@@ -75,7 +75,7 @@ def get_transfer_channel_number() -> str:
 
 def get_enable_d2h_transfer() -> bool:
     """Check if device-to-host transfer is enabled via environment variable."""
-    enable_str = os.getenv("TPU_ENABLE_D2H_TRANSFER", "true").lower()
+    enable_str = os.getenv("TPU_ENABLE_D2H_TRANSFER", "0").lower()
```
richardsliu, line 78: Can't we just put TPU_ENABLE_D2H_TRANSFER=0 in the run_disagg script?

mrjunwan-lang: as sometimes we saw d2h has perf issue and do not have bandwith to follow up. And Qwen3.5 KV size is much smaller than Qwn3-480. Disable it by default to reduce noise and we can focus on other important stuff.

PR #2655, opened 2026-05-18
Cherry-pick: Revert "Update JAX to 0.10.0" (#2648)

lk-chen:
# Description

Cherry-pick #2648

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,9 +5,9 @@ pytest-mock
 absl-py
 numpy
 google-cloud-storage
-jax==0.10.0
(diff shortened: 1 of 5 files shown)


QiliangCui (dismissed):
let's hold the reverting for now. 

Now, there is only few test failures in release. i worry if we revert jax, there will be more failures. let's wait for a while to see:

1. if we can fixing forward since this is one test only.
2. if the revertion in main is safe.

QiliangCui2023 (approved):
let's revert in release and debug in main. i think we know what happened.

PR #2660, opened 2026-05-18
Remove forcing --disable-chunked-mm-input for multimodal model

yiw-wang:
Remove the logic to force enabling the --disable-chunked-mm-input for multimodal model. Instead, user will see the warning message and can adding the server argument manually.


mrjunwan-lang (approved):

yiw-wang:
cc @jrplatin for awareness

PR #2662, opened 2026-05-18
Decouple AggregatedStatsLogger from PhaseBasedProfiler

Caslyn:
This commit decouples the aggregated batch statistics logger from the phased-based profiler. While the aggregated stats logger and phase based profiler both collect batched step information, these tools meet different ends: the phase based profiler is used to emit profile traces (eg for xprof) and the aggregated stats logger collects and aggregates the raw stats that can be ingested by user tools.

Previously, you could only use the AggregatedStatsLogger if the PhaseBasedProfiler was also turned on. This PR makes them independent so you can collect raw stats without the overhead of generating full visual traces

Specific changes:
- Replaced `ENABLE_AGGREGATED_STATS_LOGGER` with an independent `AGGREGATED_STATS_DIR` environment variable in `envs.py`.
- Fixed the execution loop in `tpu_runner.py` to explicitly call the logger's `.log()` method on every step, rather than only when the hardware profiler is active.
- Removed legacy coupled environment tests from `test_envs.py`.

This is a follow up from PR# https://github.com/vllm-project/tpu-inference/pull/2283.


tpu_inference/envs.py
```diff
@@ -107,42 +107,30 @@ def _get_validated_env() -> str | None:
...
-             default: bool = False,
-             requires: list[str] | None = None) -> Callable[[], bool]:
+def env_bool(env_name: str, default: bool = False):
     """
     Accepts both numeric strings ("0", "1") and boolean strings
     ("true", "false", "True", "False").
 
     Args:
         env_name: Name of the environment variable
         default: Default boolean value if not set
```
jrplatin, line 118: I think it's ok to keep this actually -- seems like it could have a use moving forward

jrplatin (approved):

tpu_inference/envs.py
Caslyn: Roger that, added back in f83309408d67dbe68b502bb47198319c43c46b3e

PR #2672, opened 2026-05-19
Add mla_kernel_tuner and unittest for kernel tuners

patrickji2014:
…eak when kernel change

# Description

This PR introduces the MlaKernelTuner alongside its corresponding smoke unit tests.

# Changes
Core Tuner Implementation: Added tools/kernel/tuner/v1/mla_kernel_tuner.py containing the MlaKernelTuner class. This includes generating the tuning cases, generate the mla kernel input for each tuning key and run/measure kernel for a caseset. 

Unit Tests: Added smoke tests in tools/kernel/tuner/v1/tests/test_kernel_tuner_runner.py to ensure the tuner runs end-to-end when kernel interface changes.

# Tests

TPU_VERSION=tpu7x TPU_CORES=2 python -m pytest tools/kernel/tuner/v1/tests/test_kernel_tuner_runner.py -v

# Checklist


tools/kernel/tuner/v1/tests/test_kernel_tuner_runner.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Skips the current test if either variable is absent or empty.
+        """
+        tpu_version = os.environ.get("TPU_VERSION", "").strip()
+        tpu_cores_str = os.environ.get("TPU_CORES", "").strip()
+        if not tpu_version or not tpu_cores_str:
+            self.skipTest(
+                "TPU_VERSION and TPU_CORES environment variables must be set "
+                "to run these tests (e.g. TPU_VERSION=tpu6e TPU_CORES=1).")
+        return tpu_version, int(tpu_cores_str)
```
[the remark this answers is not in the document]
patrickji2014: fixed

tools/kernel/tuner/v1/tests/test_kernel_tuner_runner.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _make_run_config(self, kernel_tuner_name: str) -> RunConfig:
+        """Builds a RunConfig with run_locally=True from TPU env vars."""
+        tpu_version, tpu_cores = self._get_tpu_env()
+        tpu_queue_multi = get_tpu_queue_by_version_and_cores(
+            tpu_version, tpu_cores, "")
+        return RunConfig(
```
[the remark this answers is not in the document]
patrickji2014: reviewed

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,333 @@
...
+class TunableParams:
+    decode_batch_size: int  # range from 1 to as high as possible before OOM with steps powers of two
+    # Constraint: batch size % decode_batch_size = 0
+
+    # Below params if passsed in as int, all cases are the same.
+    num_kv_pages_per_block: int  # Number of KV pages to process per block. Range from 1 to as high as possible before OOM,
+    # with steps of powers of two.
+    num_queries_per_block: int  # for batched_decode, this is always 1
+    vmem_limit_bytes: int  # 16MiB(?) to 64MiB, increments of 8MiB.
+    # Select lowest value that gives the highest performance
```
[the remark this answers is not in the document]
patrickji2014: fix reviewed

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,333 @@
...
+    max_num_tokens: int  # Maximum number of tokens in the batch
+    actual_num_q_heads: int  # Actual number of Q heads, <= num_q_heads in the model config, fixed at 128 for now
+    actual_lkv_dim: int  # Actual NOPE head dimension, <= lkv_dim in the model config, fixed at 512 for now
+    actual_r_dim: int  # Actual ROPE head dimension, <= r_dim in the model config, fixed at 64 for now
+    kv_dtype: str  # KV cache and KV input data type, fixed at fp8 for now
+    q_dtype: str  # Q activation dtype, fixed at fp8 for now
+    total_num_pages: int  # Total number of pages in the cache, should be large enough to cover all sequences in the batch
+    page_size_per_kv_packing: int  # Page size per KV packing, should be aligned with the kernel configuration
+    kv_packing: int  # Packing factor for KV, determined by the data type (e.g., 32 for fp8)
+    max_num_seqs: int  # Maximum number of sequences in the batch, should be large enough to cover all sequences in the batch
```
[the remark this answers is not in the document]
patrickji2014: fix reviewed

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,333 @@
...
+            for _ in range(iters):
+                jax.block_until_ready(
+                    mla_ragged_paged_attention(
+                        ql_nope=input_cache['ql_nope'],
+                        q_pe=input_cache['q_pe'],
+                        new_kv_c=input_cache['new_kv_c'],
+                        new_k_pe=input_cache['new_k_pe'],
+                        cache_kv=input_cache['cache_kv'].copy(),  # donated 
+                        kv_lens=input_cache['kv_lens'],
+                        page_indices=input_cache['page_indices'],
```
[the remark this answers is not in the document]
patrickji2014: reviewed

tpu_inference/utils.py
```diff
@@ -102,6 +102,11 @@ def align_to(unpadded_dim, pad_multiple):
     return (unpadded_dim + pad_multiple - 1) // pad_multiple * pad_multiple
 
 
+def cdiv(a, b):
+    assert b != 0
+    return (a + b - 1) // b
```
QiliangCui2023: This is the first new symbol added to `tpu_inference/` production code in this PR, and it has no unit test. The repo convention is that production code under `tpu_inference/` requires a unit test that actually exercises the new behavior. A 3-line test covering a couple of small inputs (and the `b=0` case, see the separate comment on line 106) would suffice. Alternative: drop the `tpu_inference/utils.py` change entirely and `from vllm.utils.math_utils import cdiv` in `mla_kernel_tuner.py` — `cdiv` is already available there and is what `tpu_inference/runner/*` and `layers/vllm/backends/*` already use.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+    def generate_inputs(self, tuning_key: TuningKey):
+        # Generate some mock inputs for the kernel based on the tuning key.
+        if self._TUNING_KEY and tuning_key == self._TUNING_KEY:
+            return self._KERNEL_INPUTS_CACHE
+        self._TUNING_KEY = tuning_key
+        # recover kv_len from tuning_key
+        kv_len = tuning_key.pages_per_seq * tuning_key.page_size_per_kv_packing * tuning_key.kv_packing
+        key = jax.random.PRNGKey(0)
+        seed = int(np.array(key).flatten()[0])
+        rng = np.random.default_rng(seed)
```
QiliangCui2023: The PRNG seeding here is roundabout — `jax.random.PRNGKey(0)` → `np.array(key).flatten()[0]` always evaluates to `0`, so this is equivalent to `rng = np.random.default_rng(0)`. Two suggestions: (a) just write the simpler form, or (b) if the intent was to vary the seed across tuning keys, the current code doesn't do that — every distinct `tuning_key` gets identical random inputs.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+
+        return self._KERNEL_INPUTS_CACHE
+
+    def run(self,
+            tuning_key: TuningKey,
+            tunable_params: TunableParams,
+            iters: int = 1) -> tuple[TuningStatus, float, float]:
+        logger.debug(
+            f"Running mock kernel with tuning_key={tuning_key}, tunable_params={tunable_params}, iters={iters}"
+        )
```
QiliangCui2023, line 289: The debug log says `Running mock kernel with ...` but this isn't a mock — it's the real `mla_ragged_paged_attention`. Rename to `Running mla kernel with ...` so log greps don't mislead readers into thinking a fake path is being exercised.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+        input_cache = self.generate_inputs(tuning_key)
+        start_ns = time.perf_counter_ns()
```
QiliangCui2023: `start_ns = time.perf_counter_ns()` is outside the `try:` block while `end_ns` is inside. The except path doesn't read `start_ns`, so the placement is benign — but it also means line 292's dict lookup is folded into the timed region. Move this assignment to just before the `for _ in range(iters):` loop on line 293 so only the kernel calls are timed.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+                        decode_batch_size=tunable_params.decode_batch_size,
+                        num_kv_pages_per_block=tunable_params.
+                        num_kv_pages_per_block,
+                        num_queries_per_block=tunable_params.
+                        num_queries_per_block,
+                        vmem_limit_bytes=tunable_params.vmem_limit_bytes,
+                    ))
+            end_ns = time.perf_counter_ns()
+            latency_ns = (end_ns - start_ns)
+            return TuningStatus.SUCCESS, latency_ns / iters, latency_ns  # status, average latency, total latency
```
QiliangCui2023: Returns `latency_ns / iters` (float). `KernelTunerBase.run`'s docstring at `common/kernel_tuner_base.py:412` declares `average_latency_ns` as int, and the consumer at `common/kernel_tuner_base.py:486` does `int(warmup_ns // 1000)`. Use `latency_ns // iters` to keep the int contract.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+
+    total_num_pages = max(num_pages, page_count)
+
+    ql_nope = gen_random((num_heads, total_q_len, lkv_dim), q_dtype)
+    q_pe = gen_random((total_q_len, num_heads, r_dim), q_dtype)
+    new_kv_c = gen_random((total_q_len, lkv_dim), kv_dtype)
+    new_k_pe = gen_random((total_q_len, r_dim), kv_dtype)
+
+    cache_kv = gen_random(
+        (total_num_pages, page_size // packing, packing, padded_kv_dim),
```
QiliangCui2023, line 150: `page_size // packing` silently truncates when `page_size % packing != 0`, producing a `cache_kv` with the wrong physical layout. Add `assert page_size % packing == 0, f"page_size ({page_size}) must be a multiple of packing ({packing})"` near the top of `_generate_mla_inputs` (e.g. just after the `packing = get_dtype_packing(...)` line) so the failure mode is explicit.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+            latency_ns = (end_ns - start_ns)
+            return TuningStatus.SUCCESS, latency_ns / iters, latency_ns  # status, average latency, total latency
+        except Exception as err:
+            if "RESOURCE_EXHAUSTED:" in str(err):
+                logger.warning(
+                    f"[Debug] Kernel run failed with OOM for {tuning_key=}, {tunable_params=}"
+                )
+                return TuningStatus.FAILED_OOM, float("inf"), float("inf")
+            logger.warning(
+                f"[Debug] Failed with {tuning_key=}, {tunable_params=}, got error: {err=}"
```
QiliangCui2023: Both `logger.warning(f"[Debug] ...")` lines (at 328 and 332) tag warnings with `[Debug]`, which is misleading — `grep [Debug]` over the logs will surface warnings. Use a subsystem tag like `[mla_kernel_tuner]` (or drop the prefix entirely) so the tag matches the actual log level and component.

tools/kernel/tuner/v1/mla_kernel_tuner.py
```diff
@@ -0,0 +1,334 @@
...
+                                   tunable_params=tunable_params))
+                    decode_batch_size *= 2  # increase decode batch size by powers of two
+        logger.info(f"Generated {len(tuning_cases)} tuning cases.")
+        return tuning_cases
+
+    def generate_inputs(self, tuning_key: TuningKey):
+        # Generate some mock inputs for the kernel based on the tuning key.
+        if self._TUNING_KEY and tuning_key == self._TUNING_KEY:
+            return self._KERNEL_INPUTS_CACHE
+        self._TUNING_KEY = tuning_key
```
QiliangCui2023: Nit: `self._TUNING_KEY` and `self._KERNEL_INPUTS_CACHE` use ALL_CAPS naming, which Python convention reserves for module-level constants — but these are instance attributes that change. Rename to `self._tuning_key` and `self._kernel_inputs_cache`. The attributes are declared in `common/kernel_tuner_base.py:138-139`, so the rename touches there too.

QiliangCui2023 (commented):
Most comments are inline below. One PR-level note: `tools/kernel/tuner/v1/README.md` isn't updated to mention the new `mla_kernel_tuner` entry in `KERNEL_TUNER_REGISTRY` (the checklist in the PR description says docs would be updated). Worth a one-line addition.

QiliangCui (approved):

tpu_inference/utils.py
patrickji2014: use from vllm.utils.math_utils import cdiv instead and drop this file change. thanks

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: i intend to use a fix seed for simpleness. fixed with 0 seed.

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: fixed

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: add assert here. thanks

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: use integer divide instead

tools/kernel/tuner/v1/mla_kernel_tuner.py
patrickji2014: prefix drops as the file name mla_kernel_tuner is printed as in the log already.

PR #2674, opened 2026-05-19
[Qwen3.5] Use onehot+matmul for permute and unpermute when batch size is small.

wyzhang:
Set the threshold to 2k by default, but can be disabled by setting threshold to 0

1k/8k, c=64   — total throughput: 2,211.00 → 2,531.05 (+14.5%)
1k/8k, c=128  — total throughput: 3,302.00 → 3,770.62 (+14.2%)
8k/1k, c=64   — total throughput: 11,269.00 → 12,089.52 (+7.3%)
8k/1k, c=128  — total throughput: 14,961.00 → 15,716.21 (+5.1%)

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -58,6 +58,7 @@ if TYPE_CHECKING:
     ENABLE_RS_KERNEL: bool = False
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
+    ONEHOT_MOE_PERMUTE_THRESHOLD: int = 0
 
 
 def env_with_choices(
@@ -359,6 +360,11 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: int(os.getenv("DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS", "30000")),
     "MLA_XPOSE_N_TILE_SIZE":
     lambda: int(os.getenv("MLA_XPOSE_N_TILE_SIZE", "160")),
+    # Use Onehot+Matmul for permute and unpermute before and after moe
+    # when the batch size <= this threshold. When set to 0, this feature
+    # is effectively disabled.
+    "ONEHOT_MOE_PERMUTE_THRESHOLD":
+    lambda: int(os.getenv("ONEHOT_MOE_PERMUTE_THRESHOLD", "0")),
 }
 
 
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -147,7 +147,8 @@ def moe_gmm_local(x: jax.Array,
                   activation: str,
                   topk: int,
                   parallelism: Literal["tp", "ep"],
-                  enable_rs_kernel: bool = False) -> jax.Array:
+                  enable_rs_kernel: bool = False,
+                  onehot_moe_permute_threshold: int = 0) -> jax.Array:
     """Main MoE logic on a local shard can run in TP or EP mode.
 
     Set parallelism for "tp" or "ep"
@@ -194,9 +195,24 @@ def moe_gmm_local(x: jax.Array,
                       if parallelism == "tp" else ShardingAxisName.EXPERT)
 
     if local_group_size < group_sizes.size:
-        out = ragged_gather_reduce(gmm2_res, topk_argsort_revert_indices,
-                                   topk_weights.reshape(-1), mask.reshape(-1),
-                                   topk)
+        if batch_size <= onehot_moe_permute_threshold:
+            # Use onehot + matmul for unpermutation, which can be faster
+            # for small batch size.
+            assert batch_size % topk == 0, (
+                f"batch_size ({batch_size}) should be a multiple of topk "
+                f"({topk})")
+            num_tokens = batch_size // topk
+            revert_indices = topk_argsort_revert_indices.reshape(
+                num_tokens, topk)
+            onehot = jax.nn.one_hot(revert_indices,
+                                    batch_size,
+                                    dtype=gmm2_res.dtype)
+            combine = (onehot * topk_weights[..., None] * mask).sum(axis=1)
+            out = combine @ gmm2_res
+        else:
+            out = ragged_gather_reduce(gmm2_res, topk_argsort_revert_indices,
+                                       topk_weights.reshape(-1),
+                                       mask.reshape(-1), topk)
     else:
         token_hidden_full = gmm2_res[topk_argsort_revert_indices]
         cur_sorted = token_hidden_full.reshape((-1, topk, gmm2_res.shape[-1]))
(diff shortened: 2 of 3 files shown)


gxd3:
is this possible to attach a xprof before the PR vs xprof after the PR as reference?

tpu_inference/envs.py
```diff
@@ -360,6 +361,10 @@ def _get_int_list_env() -> list[int]:
     lambda: int(os.getenv("DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS", "30000")),
     "MLA_XPOSE_N_TILE_SIZE":
     lambda: int(os.getenv("MLA_XPOSE_N_TILE_SIZE", "160")),
+    # Use Onehot+Matmul for permute and unpermute before and after moe
+    # when the batch size <= this threshold.
+    "ONEHOT_MOE_PERMUTE_THRESHOLD":
```
gxd3: also, would model hidden dimension size affect this threshold?
could we have another flag to guard this feature instead of just one threshold?)

tpu_inference/envs.py
```diff
@@ -360,6 +361,10 @@ def _get_int_list_env() -> list[int]:
...
+    lambda: int(os.getenv("ONEHOT_MOE_PERMUTE_THRESHOLD", "1024")),
```
helloworld1: Should be 2048 here right?

tpu_inference/envs.py
```diff
@@ -360,6 +361,10 @@ def _get_int_list_env() -> list[int]:
     lambda: int(os.getenv("DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS", "30000")),
     "MLA_XPOSE_N_TILE_SIZE":
     lambda: int(os.getenv("MLA_XPOSE_N_TILE_SIZE", "160")),
+    # Use Onehot+Matmul for permute and unpermute before and after moe
```
helloworld1, line 363: How to turn off?

helloworld1:
Great work! Could you validate the lm_eval to make sure it is still quality neutral?

tpu_inference/envs.py
helloworld1: Also, could you show more about the heuristic and why it only applies to the batch dimension (leading dimension? Since permute / unpermute using gather /scatter can be done on sparse core. If hidden dimension is large, it could also be faster.

kyuyeunk: +1. in general, i like using heurstic that is backed up by theoretical proof. example: https://github.com/vllm-project/tpu-inference/blob/188a45912ca8b1d4efa59e0a1a4255ec1065c81a/tpu_inference/kernels/sparse_core/ragged_scatter.py#L395-L401

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -194,9 +194,21 @@ def moe_gmm_local(x: jax.Array,
                       if parallelism == "tp" else ShardingAxisName.EXPERT)
 
     if local_group_size < group_sizes.size:
-        out = ragged_gather_reduce(gmm2_res, topk_argsort_revert_indices,
-                                   topk_weights.reshape(-1), mask.reshape(-1),
-                                   topk)
+        if batch_size <= envs.ONEHOT_MOE_PERMUTE_THRESHOLD:
```
kyuyeunk: please move out envs from this function so that this function can be stateless. example: https://github.com/vllm-project/tpu-inference/blob/188a45912ca8b1d4efa59e0a1a4255ec1065c81a/tpu_inference/layers/common/moe.py#L149

kyuyeunk (changes requested):

wyzhang:
> is this possible to attach a xprof before the PR vs xprof after the PR as reference?

Yes. I will open a bug with xprof. From xprof, decode step latency reduced from 27ms -> 23.4ms.

tpu_inference/envs.py
wyzhang: For the moment, not. But you have a good point. 
How about let me set the default threshold to 0 which is effectively disable the feature. Just try to avoid adding more flags.

tpu_inference/envs.py
wyzhang: Changed the default to 0 to have it disabled by default.

tpu_inference/envs.py
wyzhang: Set to 0 turns this off.  Default now is 0 which effectively turn it off.

tpu_inference/envs.py
wyzhang: Good point. In this case, I am not sure we have a good heurstic. Couple of things contributed to improvement. With the change, I do see xla seems to be able to better fuse ops, while previously it is a sc offload pallas kernel. For now, I have this feature disabled by default. It might need some benchmarking or testing to find the proper threshold. Will open a bug and track some details from analysis

tpu_inference/layers/common/fused_moe_gmm.py
wyzhang: Good catch. Done.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -194,9 +195,24 @@ def moe_gmm_local(x: jax.Array,
                       if parallelism == "tp" else ShardingAxisName.EXPERT)
 
     if local_group_size < group_sizes.size:
-        out = ragged_gather_reduce(gmm2_res, topk_argsort_revert_indices,
-                                   topk_weights.reshape(-1), mask.reshape(-1),
-                                   topk)
+        if batch_size <= onehot_moe_permute_threshold:
```
gxd3, line 198: consider using a different flag for unpermute?
(since `ragged_gather_reduce` do local-sum fusion as well)

tpu_inference/envs.py
gxd3: sg to default to zero.

gxd3 (approved):

tpu_inference/layers/common/fused_moe_gmm.py
helloworld1: Which one is better? I also don't want to add tons of flags if one approach is clearly better than another.

wyzhang: @gxd3 If you don't mind, I will keep as is. We can introduce another flag when there is stronger need

gxd3: ack, as long as default is not turned on.

kyuyeunk (approved):
lgtm!

qizzzh (approved):

tpu_inference/envs.py
```diff
@@ -359,6 +360,11 @@ def _get_int_list_env() -> list[int]:
...
+    # when the batch size <= this threshold. When set to 0, this feature
+    # is effectively disabled.
```
helloworld1, line 365: Good to your previously suggested 1024 / 2048 value here as a starting point. Otherwise, other people don't know what value to set.

helloworld1 (approved):

PR #2675, opened 2026-05-19
Make Kernel Tuner Result Inspector CLI Support Filter and Print as Table Format

patrickji2014:
# Description

As the TuningKey fields and TunableParams fields number grow, it's pretty hard to visualize the tuning result via just printing out the TuningKey and TunableParams together as a string in the console. This PR enhances the usability of the CLI commands via introducing --filter-key and --show options to the commands. Without --show options, the commands will print out all the fields, tuningKey's field and tunableParams fields, in one row. With the repeatable --show options specified, the commands will print out only those fields in order. 

This PR has 4 changes:
1. Add query_case_latency command to flexibly list out the tuning result for the cases that match the optional --filter-key option. This command also support --show options.
2. Add --show repeatable option to query_min_latency command
3. Make the query_case_latency and query_min_latency commands' output as a table in console. 
4. Support scroll back commands in CLI

Note: The ode change in this PR are made AI tools and reviewed/tested by me.

# Tests

I have tested the change via invoking the CLI and inspected the result. 

# Checklist


tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -378,16 +556,96 @@ def _print_table(rows, headers=None):
...
+            print(f'  Warning: unknown field(s) ignored: {", ".join(unknown)}')
+        cols = [f for f in show_fields if f in all_cols]
+        if not cols:
+            print('  (no valid --show fields; showing all columns)')
+            cols = all_cols
+    else:
+        cols = all_cols
+
+    _print_table(flat_rows, headers=cols)
+    print(f'  ({len(rows)} result(s), sorted by latency)')
```
QiliangCui2023: `_print_case_latency` is ~95% the same code as `_print_min_latency` directly above (lines 559-598): identical flat-row construction loop, identical `all_extra`/`all_cols` derivation, identical `show_fields` validation, identical trailing count print. Only the `builtin_cols` list and the per-row dict construction differ. Suggest lifting the show-fields / column logic into a shared helper that takes the row builder + `builtin_cols` as args — otherwise any future fix has to be applied twice.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -356,6 +475,65 @@ def spanner_query_min_latency(db, case_set_id, run_id):
...
+            })
+
+    def _sort_key(row):
+        status = row.get('ProcessedStatus')
+        lat = row.get('Latency')
+        if status == 'SUCCESS' and lat:
+            return (0, lat, row.get('CaseId'))
+        return (1, float('inf'), row.get('CaseId'))
+
+    return sorted(rows, key=_sort_key)
```
QiliangCui2023: The `_sort_key` inner function here is identical to the one in `local_query_case_latency` (lines 290-295), and the row-dict construction at lines 517-525 duplicates lines 280-288. Suggest lifting `_sort_key` to module scope and factoring row assembly into a small `_make_case_row(...)` helper, so the two backends diverge only in the fetch step.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -178,6 +183,120 @@ def local_query_min_latency(db_path, case_set_id, run_id):
...
+    combined.update(kv.get('tunable_params') or {})
+
+    for kv_str in filter_keys:
+        if '=' not in kv_str:
+            continue
+        field, raw = kv_str.split('=', 1)
+        field = field.strip()
+        raw = raw.strip()
+        if field not in combined:
+            return False
```
QiliangCui2023: Silent `return False` on an unknown field means a typo like `--filter-key max_num_token=4` (singular) yields an empty result with no diagnostic. By contrast, `_print_min_latency` warns on unknown `--show` fields (lines 587-589). Suggest at minimum printing a warning the first time an unknown FIELD is seen, or raising — silent acceptance of bad input is the worst CLI failure mode.

QiliangCui2023 (commented):
One note on the PR description: it says "This PR has 3 changes:" but the list below has 4 items (scrollback support is item 4). Worth renumbering.

Three inline comments below — primarily around duplication that this PR introduces between the local/spanner backends and between the two new print helpers.

QiliangCui2023 (commented):
A few more minor / nit items.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -178,6 +183,120 @@ def local_query_min_latency(db_path, case_set_id, run_id):
...
+        stored = combined[field]
+        if stored is None:
+            if raw.lower() not in ('none', 'null', ''):
+                return False
+        elif isinstance(stored, bool):
+            if stored != (raw.lower() in ('true', '1', 'yes')):
+                return False
```
QiliangCui2023: Bool coercion silently treats *any* non-truthy string as `False`. Concretely: with `stored=False`, `--filter-key foo=garbage` matches (since `'garbage'` is not in the truthy set, the RHS becomes `False`, which equals stored). So typos against a stored `False` field produce spurious matches. Suggest validating `raw` against an allowed set (`true/false/1/0/yes/no`) and warning otherwise.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -178,6 +183,120 @@ def local_query_min_latency(db_path, case_set_id, run_id):
...
+    Fields are looked up in both tuning_key and tunable_params sub-dicts.
+    Type coercion is attempted to match the stored value's type.
+    """
+    combined = {}
+    combined.update(kv.get('tuning_key') or {})
+    combined.update(kv.get('tunable_params') or {})
+
+    for kv_str in filter_keys:
+        if '=' not in kv_str:
+            continue
```
QiliangCui2023: Silently swallowing a malformed `--filter-key foo` (no `=`) means the user gets unfiltered results without knowing the flag was ignored. Suggest printing a one-line warning, e.g. `Warning: ignoring malformed --filter-key 'foo' (expected FIELD=VALUE)`.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -178,6 +183,120 @@ def local_query_min_latency(db_path, case_set_id, run_id):
...
+            'Latency': r.get('Latency'),
+            'WarmupTime': r.get('WarmupTime'),
+            'TotalTime': r.get('TotalTime'),
+            'CaseId': r.get('CaseId'),
+        })
+
+    def _sort_key(row):
+        status = row.get('ProcessedStatus')
+        lat = row.get('Latency')
+        if status == 'SUCCESS' and lat:
```
QiliangCui2023: `if status == 'SUCCESS' and lat:` treats `lat == 0` as falsy and pushes the row into the failure bucket. Use `lat is not None` instead. Same issue at line 530 in `spanner_query_case_latency`. Probably can't happen in practice but the check is wrong as written.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -605,21 +944,65 @@ def _run_command(args, source, db_path=None, spanner_db=None):
...
+
+    # Allow overriding location, but default to a clear log-like filename.
+    history_path = os.path.expanduser(
+        os.environ.get('INSPECT_RESULT_CLI_HISTORY_FILE',
+                       '~/.inspect_result_cli_history.log'))
+    try:
+        if os.path.exists(history_path):
+            readline.read_history_file(history_path)
+    except OSError:
+        pass
```
QiliangCui2023: Silent `except OSError: pass` (also at lines 987-988) hides a broken or permission-denied history file. At minimum print a one-line warning like `[history] could not read/write {path}: {e}` so the user knows why scrollback isn't persisting.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -450,6 +708,68 @@ def _build_parser():
...
+         'sub-dicts of the case — field names vary by case set type.\n\n'
+         'Example:\n'
+         '  query_case_latency --case_set_id X --run_id Y \\\n'
+         '    --filter-key max_num_tokens=4 --filter-key q_dtype=fp8'),
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+    p.add_argument('--case_set_id', required=True)
+    p.add_argument('--run_id', required=True)
+    p.add_argument(
+        '--filter-key',
```
QiliangCui2023: `--filter-key` is kebab-case while every other flag in this CLI uses snake (`--case_set_id`, `--run_id`, `--show_all`). For consistency, suggest `--filter_key` (the `dest=` is already `filter_keys`).

tools/kernel/tuner/v1/README.md
```diff
@@ -326,17 +326,38 @@ inspect|cs=testing_tuning_infra_11|run=001> query_run_status
 #### Query minimum latency results
 
 ```
-query_min_latency [--case_set_id ID] [--run_id ID]
+query_min_latency [--case_set_id ID] [--run_id ID]  [--show FIELD ...]
 ```
 
-For each unique `TuningKey`, shows the best measured latency and the corresponding `TunableParam` configuration.
+For each unique `TuningKey`, shows the best measured latency and the corresponding `TunableParam` configuration. If repeatable --show option is specified, only the FIELDs are shown. Without --show option, all the fields in TuningKey and TunableParams are show as a table format in console.
```
QiliangCui2023: Two small grammar fixes:
- Line 332: "are show as a table format" → "are shown as a table"
- Line 351: "--show_all option will enable all the cases even the case tuning failed" → "--show_all includes all cases, even ones where tuning failed"

patrickji2014: done. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: added warning message.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: refactored. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: remove the lat from the if condition. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: modified to raise ValueError instead. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: add check against the supported value and raise error if the input is not in the supported list ('true', 'false', '1', '0', 'yes', 'no'):

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: fixed to raise error when field name is not in the key or value field name list.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: refactored to use a single function. thanks,

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -356,6 +506,58 @@ def spanner_query_min_latency(db, case_set_id, run_id):
...
+                },
+                param_types={
+                    'id': gspanner.param_types.STRING,
+                    'rid': gspanner.param_types.STRING,
+                }):
+            try:
+                kv = json.loads(kv_str)
+            except (json.JSONDecodeError, TypeError):
+                continue
+            if filter_keys and not _matches_filter(kv, filter_keys):
```
QiliangCui2023: **Blocker (correctness):** `_matches_filter` returns a `FilterResult` enum, but this Spanner path still uses the old boolean idiom `not _matches_filter(...)`. Plain `Enum` members are always truthy, so `not <member>` is always `False` — the `continue` never fires and `--filter_key` is **silently ignored on the Spanner backend**, returning every row regardless of the filter. Mirror the local path (lines 312-316): compare against `FilterResult.MATCH` and handle `INVALID_FILTER` explicitly.

QiliangCui2023: **Minor:** even after fixing the truthiness bug, `_matches_filter` is called once per row here, and it `print()`s its warnings internally. A malformed filter would print the warning once per result row instead of aborting once. The local path aborts on `INVALID_FILTER` (lines 313-315); mirror that here.

tools/kernel/tuner/v1/README.md
```diff
@@ -327,17 +327,38 @@ inspect|cs=testing_tuning_infra_11|run=001> query_run_status
...
+...
+64              128                 512             64            16                 1                       2041  
+8               128                 512             64            8                  2                       2035
+```
+
+#### Query case latency
+
+```
+query_case_latency  Query latency for tuning cases with optional field filters
+                        (--case_set_id ID --run_id ID [--filter-key FIELD=VALUE ...] [--show FIELD ...] [--show_all])
```
QiliangCui2023: **Doc bug:** the README uses `--filter-key` (hyphen) here and again at line 355, but argparse registers the option as `--filter_key` (underscore) and does not auto-alias hyphens. Copy-pasting these examples fails with `unrecognized arguments: --filter-key`. Use `--filter_key` to match the implementation.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -605,21 +964,72 @@ def _run_command(args, source, db_path=None, spanner_db=None):
...
+        os.environ.get('INSPECT_RESULT_CLI_HISTORY_FILE',
+                       '~/.inspect_result_cli_history.log'))
+    try:
+        if os.path.exists(history_path):
+            readline.read_history_file(history_path)
+    except OSError:
+        print(
+            f'Warning: could not read history file at {history_path}; starting with empty history.'
+        )
+        pass
```
QiliangCui2023: **Nit:** `print(...)` is immediately followed by a bare `pass` (also at line 1015). The `pass` is dead code here — it can be removed.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -378,16 +580,92 @@ def _print_table(rows, headers=None):
...
-              f"  best_latency_us={r['Latency']}"
-              f"  warmup_us={r['WarmupTime']}"
-              f"  tunable_params={json.dumps(r['tunable_params'])}"
-              f"  case_id={r['CaseId']}")
+        flat = row_builder(r)
+        flat.update(r.get('tuning_key') or {})
+        flat.update(r.get('tunable_params') or {})
+        flat_rows.append(flat)
+
+    all_extra = [k for k in flat_rows[0] if k not in builtin_cols]
```
QiliangCui2023: **Nit:** `all_extra` is derived from `flat_rows[0]` only, so if rows have heterogeneous `tuning_key`/`tunable_params` keys, fields present only in later rows won't get columns. Also, a tuning/tunable field that collides with a built-in name (`case_id`, `latency_us`, ...) silently overwrites the built-in via `flat.update(...)` just above. Edge cases, but worth a guard or a comment.

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -22,13 +22,20 @@
     count_buckets       Count buckets (--case_set_id ID --run_id ID)
     list_bucket_status  Show completed vs pending counts (--case_set_id ID --run_id ID)
     query_run_status    Show timing info for a run (--case_set_id ID --run_id ID)
-    query_min_latency   Show best latency per TuningKey (--case_set_id ID --run_id ID)
+    query_min_latency   Show best latency per TuningKey (--case_set_id ID --run_id ID [--show FIELD ...])
+    query_case_latency  Query latency for tuning cases with optional field filters
```
QiliangCui2023, line 26: **Nit:** this top-of-file usage summary for `query_case_latency` omits `--show` and `--show_all`, which the detailed argparse help and the README both document. Worth adding for consistency.

QiliangCui2023 (commented):
A few inline findings below. The Spanner `--filter_key` path is non-functional (blocker), and the README documents a flag spelling (`--filter-key`) that the CLI doesn't accept. The rest are minor/nits.

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: udpated. thanks

tools/kernel/tuner/v1/inspect_result_cli.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/inspect_result_cli.py
QiliangCui: hm.. this logic doesn't seem right...

tools/kernel/tuner/v1/inspect_result_cli.py
```diff
@@ -178,6 +186,149 @@ def local_query_min_latency(db_path, case_set_id, run_id):
...
+# ---------------------------------------------------------------------------
+# Filtering helpers
+# ---------------------------------------------------------------------------
+class FilterResult(Enum):
+    MATCH = 1
+    NO_MATCH = 2
+    INVALID_FILTER = 3
+
+
+def _matches_filter(kv: dict, filter_keys: list) -> FilterResult:
```
QiliangCui, line 198: should we add a test for this logic?

QiliangCui (approved):

yiw-wang (approved):

QiliangCui (approved):

PR #2679, opened 2026-05-20
Optimize attn DP + MoE EP ReduceScatter using JAX psum_scatter and avoiding any padding

zhangamy-crypto:
…oiding any padding

# Description
Currently under attn DP + MoE EP, the code issues a psum. XLA can sometimes but not always fuse it into all-reduce-scatter. For Qwen 3.5, the specific shape caused XLA to also introduce padding during the fusion, and thus does not allow it to be offloaded to Sparsecore. This change issues a direct psum_scatter and removes the padding, and allows SpraseCore offload.
Note --xla_tpu_use_minor_sharding_for_major_trivial_input=true need to be enabled for this work.

# Tests

Tested on VM and verified the fused all-reduce-scatter is now a native reduce-scatter, the padding is gone, and with --xla_tpu_use_minor_sharding_for_major_trivial_input=true, is able to be offloaded to SparseCore. 

# Checklist


tpu_inference/layers/vllm/interface/moe.py
```diff
@@ -98,6 +98,14 @@ def vllm_moe_apply(layer: FusedMoE, weights: FusedMoEWeights,
             except AssertionError:
                 pass
 
+    is_dp = False
+    if vllm_config is not None:
+        sharding_strategy = vllm_config.additional_config.get("sharding", {}).get("sharding_strategy", {})
```
kyuyeunk: instead of referring to sharding strategy, is it possible to just infer it by reading attention dp axis size?

tpu_inference/layers/vllm/interface/moe.py
```diff
@@ -106,5 +114,5 @@ def vllm_moe_apply(layer: FusedMoE, weights: FusedMoEWeights,
             weights=weights,
             moe_backend=quant_method_instance.moe_backend,
             mesh=quant_method_instance.mesh,
-            extra_backend_kwargs=quant_method_instance.extra_backend_kwargs,
+            extra_backend_kwargs=extra_kwargs,
```
kyuyeunk, line 119: random nit. probably for future clean-up pr, but this extra backend kwargs feels unnecessary considering that we only actually use gmm backend.

wyzhang (approved):

tpu_inference/layers/vllm/custom_ops/fused_moe.py
```diff
@@ -13,12 +13,40 @@
...
+
+    is_dp = False
+    if vllm_config is not None:
+        sharding_strategy = vllm_config.additional_config.get("sharding", {}).get("sharding_strategy", {})
+        is_dp = sharding_strategy.get("enable_dp_attention", False)
+
+    if is_dp:
+        return states[..., :trunc_size]
+
+    return self._original_maybe_reduce_final_output(states, trunc_size)
```
kyuyeunk: can you explain what is the difference between dp path and non-dp path (which invokes the original implementation?)

tpu_inference/layers/vllm/custom_ops/fused_moe.py
```diff
@@ -13,12 +13,40 @@
...
+
 @FusedMoE.register_oot
 class VllmFusedMoE(FusedMoE):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        
+        self.runner._original_maybe_reduce_final_output = self.runner._maybe_reduce_final_output
```
kyuyeunk: instead of overriding the runner function, i feel like the cleaner method is to write custom model runner implementation?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -234,6 +235,11 @@ def moe_gmm_local(x: jax.Array,
                 num_micro_batches=num_mb,
                 axis_name=reduction_axis)
             out = rs_out.astype(x.dtype)
+    elif scatter_results:
+        out = jax.lax.psum_scatter(out,
```
kyuyeunk: so is my understanding correct that without this explicit change, the xla was not performing psum scatter? what happens if you don't make this change and just perform reshard before returning output from `VllmFusedMoE` forward pass? can that hint xla to convert psum into psum scatter?

kyuyeunk:
w.r.t. `--xla_tpu_use_minor_sharding_for_major_trivial_input=true`, do you think it makes sense to just enable it by default in here? https://github.com/vllm-project/tpu-inference/blob/66cf2156ea099766f0a651e496c23b304793c5e6/tpu_inference/models/vllm/vllm_model_wrapper.py#L288-L293

or maybe just enable it globally in here? https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/env_override.py

tpu_inference/layers/common/fused_moe_gmm.py
zhangamy-crypto: Yeah so without this change, XLA sometimes did but other times did not trigger ReduceScatter (e.g. http://xprof.corp.google.com/trace_viewer/wyzhang-7869745658101919264, which was also a DP run) ; when it was triggered, it was done via the TpuAllReduceScatterFusion pass, and thus had the padding issue.
"what happens if you don't make this change and just perform reshard before returning output from VllmFusedMoE forward pass" - that'll still trigger XLA to inject DynamicSlice, which means we still trigger the TpuAllReduceScatterFusion pass, which then introduces the whole padding issue.

tpu_inference/layers/vllm/custom_ops/fused_moe.py
zhangamy-crypto: it does seem conceptually cleaner, but if we define a custom runner, we'd have to either:
1. copy past the entire _init_ of upstream vllm, obviously that causes maintenance issues.
2. dynamically swap the class after init  (self.runner.__class__ = TpuMoERunner). When we dynamically swap the __class__ of an nn.Module in Python, it severs the C++ bindings for the _parameters dictionary and would cause a crash.

tpu_inference/layers/vllm/interface/moe.py
zhangamy-crypto: good suggestion, updated.

tpu_inference/layers/vllm/interface/moe.py
zhangamy-crypto: yeah sg for future cleanup.

tpu_inference/layers/vllm/custom_ops/fused_moe.py
zhangamy-crypto: so my understanding is without this change, it would trigger tensor_model_parallel_all_reduce, and in the non-dp case, it results in a AllReduce(AllReduce(x)), then XLA drops the 2nd AR. in dp case, we stop short of actually triggering that 2nd AR, so ensuring it's just the ReduceScatter we desired.

tpu_inference/layers/vllm/custom_ops/fused_moe.py
kyuyeunk: so it's actually easy to achieve that - you can refer to how `VllmFusedMoE` function handles it.

1. like how `VllmFusedMoE` subclassed orinal `FusedMoE`, just subclass original model runner `MoERunner` so we don't have to write our own init function: https://github.com/vllm-project/vllm/blob/6dc0a71843878ef45e29d4732147290b797b70fd/vllm/model_executor/layers/fused_moe/runner/moe_runner.py#L193
2. like how `VllmFusedMoE` registered its class with `@FusedMoE.register_oot`, doing `@MoERunner.register_oot` will allow any instance creation of the original `MoERunner` will return our own custom implementation.

zhangamy-crypto: Nice, thanks for the suggestion, fixed.

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -234,6 +235,25 @@ def moe_gmm_local(x: jax.Array,
                 num_micro_batches=num_mb,
                 axis_name=reduction_axis)
             out = rs_out.astype(x.dtype)
+    elif scatter_results:
+        dp_axes = ('data', 'attn_dp', 'attn_dp_expert')
```
kyuyeunk: you can just use `ShardingAxisName.ATTN_DATA`. please see https://github.com/vllm-project/tpu-inference/blob/aa42a05cc50a82803056c18ca1dd139e42f6c60b/tpu_inference/layers/common/sharding.py#L34-L61

tpu_inference/layers/vllm/interface/moe.py
```diff
@@ -98,6 +99,12 @@ def vllm_moe_apply(layer: FusedMoE, weights: FusedMoEWeights,
             except AssertionError:
                 pass
 
+    mesh = quant_method_instance.mesh
+    is_dp = get_mesh_shape_product(mesh, ('attn_dp', 'attn_dp_expert')) > 1
```
kyuyeunk: similar comment. you can just use `ShardingAxisName.ATTN_DATA_EXPERT`: https://github.com/vllm-project/tpu-inference/blob/aa42a05cc50a82803056c18ca1dd139e42f6c60b/tpu_inference/layers/common/sharding.py#L42

zhangamy-crypto: i'm using ('attn_dp', 'attn_dp_expert') not ('attn_dp_expert', 'expert')

tpu_inference/layers/common/fused_moe_gmm.py
zhangamy-crypto: done

kyuyeunk (approved):
lgtm!

tpu_inference/layers/vllm/interface/moe.py
kyuyeunk: hmm i see i see. maybe better approach is to have it like this. for context, i'm trying to avoid any hard coded string variables in our code base.

```
# ATTN_DATA = ('data', 'attn_dp', 'attn_dp_expert')
attn_dp_size = get_mesh_shape_product(mesh, ShardingAxisName.ATTN_DATA)
# MLP_DATA = 'data'
dp_size = get_mesh_shape_product(mesh, ShardingAxisName.MLP_DATA)

is_attn_dp = attn_dp_size // dp_size > 1
```

zhangamy-crypto: done! i'll keep that in mind for future PRs too.

PR #2682, opened 2026-05-20
QKVParallelLinear: derive tp size from sharding_config  rather than parallel_config

lenscloth:
# Description

TP size was originally derived from `parallel_config`.  This PR will change it to use `sharding_config` instead.
This is necessary because if attention DP is enabled, TP size will change accordingly. 

https://github.com/vllm-project/tpu-inference/blob/4d54e71116e10e542c257c55916c1fd2dfca7bc1/tpu_inference/platforms/tpu_platform.py#L187

# Tests

Added unit tests on `tests/layers/vllm/test_qkv_kv_replication.py`.

Also conducted a quick smoke test using Qwen/Qwen3.5-35B-A3B-FP8. Specifically,
```
MODEL_IMPL_TYPE=vllm USE_MOE_EP_KERNEL=0 NEW_MODEL_DESIGN=1 vllm serve Qwen/Qwen3.5-35B-A3B-FP8 --max-model-len=9216 --max-num-batched-tokens=1024 --max-num-seqs=64 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --async-scheduling --port=8000 --language-model-only --enable-auto-tool-choice --tool-call-parser=qwen3_coder --reasoning-parser=qwen3 '--limit-mm-per-prompt={"image": 0, "video": 0}' --kv-cache-dtype=fp8 --enable-expert-parallel '--additional_config={"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' --mamba-cache-dtype=bfloat16
```

### DIFF
--- a/tpu_inference/layers/vllm/custom_ops/linear.py
+++ b/tpu_inference/layers/vllm/custom_ops/linear.py
@@ -24,7 +24,8 @@ from vllm.model_executor.layers.linear import (ColumnParallelLinear,
                                                ReplicatedLinear,
                                                RowParallelLinear)
 
-from tpu_inference.layers.common.sharding import ShardingAxisName2D
+from tpu_inference.layers.common.sharding import (ShardingAxisName2D,
+                                                  ShardingConfigManager)
 from tpu_inference.layers.common.utils import (
     reorder_concatenated_tensor_for_sharding,
     slice_sharded_tensor_for_concatenation)
@@ -90,8 +91,10 @@ class VllmQKVParallelLinear(QKVParallelLinear):
                  **kwargs):
         if total_num_kv_heads is None:
             total_num_kv_heads = total_num_heads
-
-        tp = get_current_vllm_config().parallel_config.tensor_parallel_size
+        vllm_config = get_current_vllm_config()
+        sharding_config = (getattr(vllm_config, "sharding_config", None) or
+                           ShardingConfigManager.from_vllm_config(vllm_config))
+        tp = sharding_config.tp_size
         if tp > total_num_kv_heads:
             assert tp % total_num_kv_heads == 0, (
                 f"tp_size ({tp}) must be divisible by total_num_kv_heads "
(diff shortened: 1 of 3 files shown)


tpu_inference/layers/vllm/custom_ops/linear.py
```diff
@@ -24,7 +24,8 @@
                                                ReplicatedLinear,
                                                RowParallelLinear)
 
-from tpu_inference.layers.common.sharding import ShardingAxisName2D
```
kyuyeunk, line 27: instead of `ShardingAxisName2D` which does not work with attention dp, please use more generic version - which is `ShardingAxisName`.

tpu_inference/layers/vllm/custom_ops/linear.py
```diff
@@ -90,8 +91,10 @@ def __init__(self,
                  **kwargs):
         if total_num_kv_heads is None:
             total_num_kv_heads = total_num_heads
-
-        tp = get_current_vllm_config().parallel_config.tensor_parallel_size
```
kyuyeunk, line 94: isn't it possible to just do `tp = get_mesh_shape_product(mesh, ShardingAxisName.ATTN_HEAD)`?

example: https://github.com/vllm-project/tpu-inference/blob/13521831b47b1357dff046d566140d38330014a8/tpu_inference/layers/vllm/quantization/configs.py#L56-L91

tpu_inference/layers/vllm/custom_ops/linear.py
lenscloth: Thanks for the suggestion. Reflected.

tpu_inference/layers/vllm/custom_ops/linear.py
lenscloth: Modified logic to directly use mesh.

kyuyeunk (approved):

PR #2698, opened 2026-05-21
Limit SC tracing to 1 core and 1 tile

bigeye:
With SC profiling enabled, the profiled result got too big, so it fails to capture important events like collectives. This change will make sure we have a proper limit for SC profiling.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/runner/utils.py
+++ b/tpu_inference/runner/utils.py
@@ -514,6 +514,11 @@ class PhasedBasedProfiler:
         }
         self.default_profiling_options = jax.profiler.ProfileOptions()
         self.default_profiling_options.python_tracer_level = envs.PYTHON_TRACER_LEVEL
+        self.default_profiling_options.advanced_configuration = {
+            "tpu_trace_mode": "TRACE_COMPUTE",
+            "tpu_num_sparse_cores_to_trace": 1,
+            "tpu_num_sparse_core_tiles_to_trace": 1,
+        }
 
         self.current_phase: str = ""
 


tpu_inference/runner/utils.py
```diff
@@ -501,12 +502,21 @@ def __init__(self,
         }
         self.default_profiling_options = jax.profiler.ProfileOptions()
         self.default_profiling_options.python_tracer_level = envs.PYTHON_TRACER_LEVEL
+        self.default_profiling_options.advanced_configuration = {
+            "tpu_trace_mode": "TRACE_COMPUTE",
```
kyuyeunk, line 518: nit: is setting trace mode also needed?

kyuyeunk (approved):

tpu_inference/runner/utils.py
bigeye: Good catch! It's actually not for the xprof issue we are facing. I realized that the default tracing mode is TRACE_ONLY_XLA according to (https://docs.jax.dev/en/latest/profiling.html#tpu-options), not TRACE_COMPUTE
Strangly enough, xprof has shown the "compute" ops even without this explicit settings. It could be that the doc is outdated or TRACE_COMPUTE is silently enabled somewhere.. But I think explicit settings based on the documentation would be more robust approach here.  Please let me know if you think otherwise.

tpu_inference/runner/utils.py
```diff
@@ -501,6 +501,11 @@ def __init__(self,
...
+            "tpu_num_sparse_cores_to_trace": 1,
+            "tpu_num_sparse_core_tiles_to_trace": 1,
```
jrplatin, line 520: Can you please add an env var for the number of chips too? E.g.:

"tpu_num_chips_to_profile_per_task" : envs.TPU_NUM....


I think default to `None` should be safe?

bigeye: Thanks for the suggestion! I also wanted to set the environment for tpu trace mode for SYNC_WAIT tracing. Let me do all of these env variable changes in the follow-up cl.

PR #2699, opened 2026-05-21
Clean up GDN implementation and reorganize code

helloworld1:
# Description

* Original recurrent GDN kernels written by @yaochengji has a prefill and decode part. The prefill part is being replaced by recurrent_scan_v2 kernel but the decode is still performing better and used in the decode part. This leads to multiple path of code making it confusing to choose.
* This refactored the code in the following way: Move @yaochengji 's original kernel to gdn.v1 to archive as reference. Copy the @yaochengji 's decode kernel to gdn.v2 together with recurrent_scan_v2 kernel to be the actively used / developed kernels
* Clean up the wrapper part to only have ragged_gated_delta_rule_wrapper.py to orchestrate decode / prefill and choose of different implementations
* Env var has only 3 impls "chunked_jax_pd", "chunked_kernel_pd", "chunked_kernel_p_recurrent_kernel_d" with future plan to leave only jax /  kernel impl to make choices / testing easier.

# Tests
E2E tested, lm_eval shows no regression

# Checklist


helloworld1:
@yaochengji Since this is touching the structure of your original kernel. Kindly request your feedback if you have. Thanks!

yaochengji (approved):
LGTM.

Just want to mention that the recurrent GDN kernel can be used for spec decoding in the future.

helloworld1:
> LGTM.
> 
> Just want to mention that the recurrent GDN kernel can be used for spec decoding in the future.

Ah, thanks. the recurrent prefill kernel won't be removed but left in the "v1" folder in the repo (Alone with its unit test). In the future when working on spec decoding, we can work on it further and move it to the active kernel path when needed.

tpu_inference/kernels/gdn/v1/fused_gdn_kernel_wrapper.py
```diff
@@ -21,9 +21,9 @@
 import jax.numpy as jnp
 from jax.experimental.pallas import tpu as pltpu
 
-from tpu_inference.kernels.gdn.fused_gdn_decode_kernel import \
+from tpu_inference.kernels.gdn.v1.fused_gdn_decode_kernel import \
```
coolkp, line 24: Maybe we should name `v1` and `v2` to be more descriptive of implementation?

coolkp (approved):

tpu_inference/kernels/gdn/v1/fused_gdn_kernel_wrapper.py
helloworld1: Sounds good. I can add more comments in the __init__.

helloworld1: Updated PR to add doc for v1 / v2 impl details.

PR #2700, opened 2026-05-22
Support MPMD DP 

wenxindongwork:
# Description

This pull request introduces support for vLLM-native multi-process data parallelism (DP) on TPU, allowing each data parallel rank to run in its own process, each pinned to a disjoint set of TPU chips. This is enabled via the new `TPU_MULTIPROCESS_DP` environment variable:

MPMD DP will not work for attention DP or Pathways. 

Follow up PR will enable MPMD DP for MoE models in non-attention-dp settings. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/sharding.py
+++ b/tpu_inference/layers/common/sharding.py
@@ -19,6 +19,7 @@ from typing import TYPE_CHECKING, List, Optional
 
 import jax.numpy as jnp
 import numpy as np
+import vllm.envs as vllm_envs
 from jax.sharding import Mesh
 
 from tpu_inference import envs, utils
@@ -177,23 +178,25 @@ class ShardingConfigManager:
         enable_dp_attention = sharding_strategy.get("enable_dp_attention",
                                                     False)
         # vLLM-native multi-process data parallelism: one engine process per
-        # DP rank, fronted by a single load-balanced API endpoint. 
+        # DP rank, fronted by a single load-balanced API endpoint.
         #
         # It is not used (we fall back to single-process SPMD DP) when:
-        #  - the model is MoE 
+        #  - the model is MoE
         #  - attention DP is enabled
+        #  - running on Pathways
         model_config = vllm_config.model_config
         multiprocess_dp = (envs.TPU_MULTIPROCESS_DP and data_parallelism > 1
                            and model_config is not None
                            and not model_config.is_moe
-                           and not enable_dp_attention)
+                           and not enable_dp_attention
+                           and not vllm_envs.VLLM_TPU_USING_PATHWAYS)
         if (envs.TPU_MULTIPROCESS_DP and data_parallelism > 1
                 and not multiprocess_dp):
             logger.warning(
                 "TPU_MULTIPROCESS_DP is set but is not supported for MoE "
-                "models or with attention DP (enable_dp_attention). " 
-                "Falling back to single-process "
-                "SPMD data parallelism.")
+                "models, with attention DP (enable_dp_attention), or on "
+                "Pathways. Falling back to single-process SPMD data "
+                "parallelism.")
         if multiprocess_dp:
             # Each engine process is tensor-parallel only.
             data_parallelism = 1
--- a/tpu_inference/layers/common/sharding.py
+++ b/tpu_inference/layers/common/sharding.py
@@ -22,9 +22,12 @@ import numpy as np
 from jax.sharding import Mesh
 
 from tpu_inference import envs, utils
+from tpu_inference.logger import init_logger
 
 if TYPE_CHECKING:
     from vllm.config import VllmConfig
+
+logger = init_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


tpu_inference/layers/common/sharding.py
```diff
@@ -171,14 +175,39 @@ def from_vllm_config(cls,
...
+            raise ValueError(
+                "TPU_MULTIPROCESS_DP is set but is not supported for MoE "
+                "models, with attention DP (enable_dp_attention), or on "
+                "Pathways. Please disable TPU_MULTIPROCESS_DP.")
+        if multiprocess_dp:
+            data_parallelism = 1
+            logger.warning(
+                "TPU_MULTIPROCESS_DP is enabled: supported for online serving "
+                "only. The offline LLM().generate() API will hang. "
+                "Use `vllm serve` instead.")
```
lk-chen, line 204: what does it mean to "use vllm serve"?

tpu_inference/layers/common/sharding.py
```diff
@@ -171,14 +175,39 @@ def from_vllm_config(cls,
         ss_tensor_parallelsim = sharding_strategy.get("tensor_parallelism",
                                                       None)
         data_parallelism = parallel_config.data_parallel_size
+        enable_dp_attention = sharding_strategy.get("enable_dp_attention",
+                                                    False)
+        # vLLM-native multi-process data parallelism: one engine process per
+        # DP rank, fronted by a single load-balanced API endpoint.
+        #
+        # It is not used (we fall back to single-process SPMD DP) when:
+        #  - the model is MoE
```
lk-chen, line 184: if dp_attention==False, why can't MoE use MPMD?

tpu_inference/layers/common/sharding.py
wenxindongwork: offline inference is not supported. Upstream has the same behavior. For example, if we run offline_inference.py on GPU, we will get this error: ```ValueError: LLM(data_parallel_size=2) is not supported for single-process usage and may hang. Please use the explicit multi-process data-parallel example at 'examples/features/data_parallel/data_parallel_offline.py'.```. 

Offline inference runs on a single process, but DP requires multiple processes.

tpu_inference/layers/common/sharding.py
wenxindongwork: Upstream sometimes defaults to attention_dp when the model is moe and dp is enabled. this check is a bit strict, i will need to follow up with another PR.

tpu_inference/layers/common/sharding.py
lk-chen: Got it, here "vllm serve" meant online serving

tpu_inference/layers/common/sharding.py
lk-chen: Ack, FYI one of Gemma-4 variant is moe, we will want to support MoE shortly

tpu_inference/worker/tpu_worker.py
```diff
@@ -164,11 +164,55 @@ def initialize_cache(self, num_gpu_blocks: int,
...
+    def _setup_dp_chip_isolation(self) -> None:
+        """Sets libtpu env vars so the process initializes JAX on a subset of
+        physical chips. Must run before JAX initializes its TPU backend.
+        """
+        from tpu_inference import tpu_info
+
+        parallel_config = self.vllm_config.parallel_config
+        dp_rank = getattr(parallel_config, "data_parallel_index", 0) or 0
+
+        sharding_config = getattr(self.vllm_config, "sharding_config", None)
```
lk-chen: nit, line 177 doesn't check None, thus line 176 should just be `getattr(..., "sharding_config")` or `self.vllm_config.sharding_config` (require sharding_config exist)

tpu_inference/worker/tpu_worker.py
```diff
@@ -164,11 +164,55 @@ def initialize_cache(self, num_gpu_blocks: int,
...
+        num_devices = sharding_config.total_devices
+
+        cores_per_chip = tpu_info.get_num_cores_per_chip()
+        chips_per_rank = math.ceil(num_devices / cores_per_chip)
```
lk-chen, line 180: `chips_per_rank` should be world_size, see https://github.com/vllm-project/vllm/blob/91f5b92438a568c89e8b9d6c2c55de5a552291f6/vllm/v1/engine/core.py#L2072-L2088

on a v7x-8 VM, if user just use DP=2 TP=2, this chip_per_rank should be 2 instead of 4 (8/2=4)

wenxindongwork: in this example: `DP=2 TP=2` on v7x-8

`chips_per_rank` should already by 2 I believe, because `num_devices=4`, and `cores_per_chip=2`. 

`sharding_config.total_devices` returns the per-dp number of cores.

lk-chen (approved):

PR #2702, opened 2026-05-22
QKVParallelLinear: create 2d mesh of attn_dp and attn_dp

lenscloth:
# Description

Previously we have inferred kv head axis from mesh assuming that `kv_head_axis` is `str`. However this raises an error when `kv_head_axis` is tuple.

https://github.com/vllm-project/tpu-inference/blob/e97028d120b4ee4ccb7c84c63d2f82f822bf684a/tpu_inference/layers/vllm/custom_ops/linear.py#L174-L193

In order to avoid this, we recreate 2d mesh of `ATTN_DATA` and `ATTN_HEAD` using `get_mesh_shape_product` .

# Tests

E2e test with `Qwen/Qwen3.5-397B-A17B-FP8`

### DIFF
--- a/tpu_inference/layers/vllm/custom_ops/linear.py
+++ b/tpu_inference/layers/vllm/custom_ops/linear.py
@@ -171,42 +171,50 @@ class VllmQKVParallelLinear(QKVParallelLinear):
         )
 
         mesh = jax.sharding.get_abstract_mesh()
-
-        # Collapse the N-D mesh into a 2D view: one axis for all mesh axes
-        # backing ATTN_DATA (flattened) and one for all mesh axes backing
-        # ATTN_HEAD (flattened). Then split the head axis into
-        # (kv_head, replica) of sizes (head_size // replicas, replicas).
-        #
-        # The `replica` sub-axis is later marked replicated via `shard_map`
-        # with no data movement, since `_tile_kv` already placed identical
-        # KV-head copies on each replica-group of devices when
-        # TP > total_num_kv_heads.
+        # Split the `kv_head` mesh axis (size kv_size) into a pair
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/vllm/custom_ops/linear.py
```diff
@@ -171,31 +171,42 @@ def forward(
...
+        #
+        # The `replica` sub-axis is later marked replicated via `shard_map`
+        # with no data movement, since `_tile_kv` already placed identical
+        # KV-head copies on each replica-group of devices when
+        # TP > total_num_kv_heads.
         replicas = self.num_kv_head_replicas
-        i = mesh.axis_names.index(kv_head_axis)
-        kv_size = mesh.axis_sizes[i]
-        kv_type = mesh.axis_types[i]
+        attn_data_axis = 'attn_data'
```
kyuyeunk: please avoid using string variables / magic varialbes in the code base. does doing something like this work? https://github.com/vllm-project/tpu-inference/pull/2679#discussion_r3283191377

lenscloth: Done. I've removed the use of string variables.

The appropriate solution would be to add a mesh axis called kv_replicate (as seen in [this code](https://github.com/vllm-project/tpu-inference/blob/7f1e3397f0aab24bd948457b2e983bdd163bd572/tpu_inference/runner/tpu_runner.py#L389)).

However, this would require significant changes, including KV cache replication. How about addressing this in a follow-up PR?

kyuyeunk: okay that makes sense. please create a follow up pr for it!

kyuyeunk (approved):

PR #2717, opened 2026-05-26
Qwen optimizations base changes

h1t35h:
# Description

Adds fixes and optimizations for `Qwen3-VL-235B-A22B-Thinking-FP8` . 

1. Adds accounting for parallelism for identifying remaining HBM available for KV Cache.
2. Fix Sharding configuration for cases where head size is smaller than parallelism.
3. Implement `PositionalEmbeddingGeneration` in pure JAX translated from [original_vllm_impl](https://github.com/vllm-project/vllm/blob/771e1e48b153df0e5e4334061deb1cb2ebde8fab/vllm/model_executor/models/qwen3_vl.py#L278).
   - Similar optimization by adding `embed_input_ids_func`.

# Tests

Tested on TPU v7-8

1. Time to First Token (Mean TTFT)
    Baseline: 335.20 ms
    Optimized: 216.36 ms
    Improvement: 35.45% reduction in time to first token.
2. Time per Output Token (Mean TPOT)
    Baseline: 35.04 ms
    Optimized: 20.03 ms
    Improvement: 42.84% reduction in time per output token.
3. Inter-token Latency (Mean ITL)
    Baseline: 31.54 ms
    Optimized: 18.02 ms
    Improvement: 42.87% reduction in inter-token latency.

Overall, the optimization yielded a ~35% improvement in initial response time and a ~43% improvement in generation speed and smoothness.

# Profiling results (TTFT improvements)
before
after

# Checklist

### DIFF
--- a/tpu_inference/layers/common/sharding.py
+++ b/tpu_inference/layers/common/sharding.py
@@ -109,6 +109,22 @@ class LazyShardingAxisName:
 
     def __getattr__(self, name):
         self._initialize()
+        if name == "KV_CACHE_HEAD" and self._cls is ShardingAxisNameBase:
+            from vllm.config import get_current_vllm_config
+            try:
+                vllm_config = get_current_vllm_config()
+            except Exception:
+                vllm_config = None
+
+            enable_dp_attention = False
+            if vllm_config is not None:
+                enable_dp_attention = vllm_config.additional_config.get(
+                    "sharding", {}).get("sharding_strategy", {}).get(
+                        "enable_dp_attention", False)
+
+            if not enable_dp_attention:
+                return getattr(self._cls, "MODEL", "model")
+
         return getattr(self._cls, name)
 
 
--- a/tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
+++ b/tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
@@ -41,6 +41,9 @@ making sure they go through standard function arguments:
    model's cache just-in-time for the execution, and proceed with the original forward pass.
 """
 
+from functools import partial
+import jax
+import jax.numpy as jnp
 import torch
 import torch.nn as nn
 import vllm.model_executor.models.qwen3_vl as qwen3_vl_mod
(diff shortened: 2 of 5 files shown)


anthonsu:
Thanks @h1t35h for making this change! 

Could you please add `--signature` to your commits so that the PR can pass DCO requirement?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -488,7 +488,7 @@ def wrap_embed_input_ids_func(self):
         if not self.vllm_config.model_config.is_multimodal_model:
             return None
 
-        # The function cannot be JITted directly due to its dynamic implementation
+        @jax.jit
```
yuyanpeng-google, line 491: Does the patch only work for Qwen3-VL, or does it apply to all models in vLLM? I believe this function needs to support all vLLM models, with no guarantee of JIT compilation. cc @wdhongtw

lk-chen: cc @anthonsu

tpu_inference/worker/tpu_worker.py
```diff
@@ -392,6 +392,15 @@ def determine_available_memory(self) -> int:
         total_hbm_limit_cap = total_hbm_limit * gpu_memory_utilization
         total_hbm_avail = int(total_hbm_limit_cap - total_hbm_used)
 
+        # Scale down the total HBM available by the KV Cache replication factor
```
lk-chen, line 395: What do you think we rely on [memory_stats](https://docs.jax.dev/en/latest/_autosummary/jax.Device.html#:~:text=memory-,memory_stats,-Returns%20memory%20statistics) to get available HBM?

tpu_inference/layers/common/sharding.py
```diff
@@ -109,6 +109,22 @@ def _initialize(self):
 
     def __getattr__(self, name):
         self._initialize()
+        if name == "KV_CACHE_HEAD" and self._cls is ShardingAxisNameBase:
+            from vllm.config import get_current_vllm_config
+            try:
+                vllm_config = get_current_vllm_config()
+            except Exception:
+                vllm_config = None
```
anthonsu, line 117: It might be safer to log a warning or catch specific configuration errors to ensure that legitimate issues with vllm_config aren't silently ignored while falling back to the 'model' axis.

lk-chen (commented):
Where's the change for "2. Fix Sharding configuration for cases where head size is smaller than parallelism."? The change in `sharding.py` seems aliasing when DP_ATTN  disabled (which I don't get why necessary).

Also could you update the title to be more descriptive?

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -250,42 +253,210 @@ def _patched_flatten_embeddings(embeddings: NestedTensors) -> torch.Tensor:
...
+    update_values = flattened_padded[gather_indices]
+
+    condition = is_multimodal.unsqueeze(-1)
+    new_embeds = torch.where(condition, update_values, inputs_embeds)
+
+    return new_embeds
+
+
+@partial(jax.jit, static_argnums=(1, 2, 3, 4, 5))
+def pos_embed_interpolate_jax(
```
anthonsu, line 285: Could we add a docstring mapping this JAX implementation to the original PyTorch code? Specifically, defining the (t, h, w) parameters and mapping the jnp operations used for the bilinear interpolation weight logic. It would greatly help in verifying the math and maintenance!

tpu_inference/models/vllm/vllm_model_wrapper.py
anthonsu: @yuyanpeng-google , @lk-chen , I believe the current patch only makes JIT work for Qwen3-VL. For other models like **Gemma4**, the default `vllm.model_executor.models.utils._merge_multimodal_embeddings` will still be triggered, which could cause a crash or a fallback to Eager mode. A similar JIT-friendly implementation is needed for other multi-modal models (including Gemma4) to avoid breaking them.

tpu_inference/worker/tpu_worker.py
h1t35h: Thanks for the suggestion, I was not aware of that. Let me go back and check I believe that should be a better solution but I'll need to deep dive and get back.

tpu_inference/models/vllm/vllm_model_wrapper.py
h1t35h: Yes, I think this will break other implementations (apologies, I'm new to the repo). I have a question though since I'm new to the repo, non JITed methods will always be suboptimal are there any specific cases where they are needed ? Put otherwise if I can try to migrate dependencies for this would it be a preferred solution ?

lk-chen: > any specific cases where they are needed ?

when a new model came and we didn't have time to properly fix it.

That's why there's a `maybe_jit_embed_multimodal_func` in this file

tpu_inference/layers/common/sharding.py
h1t35h: Done

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
h1t35h: Done. There is also a triton kernel implementation for optimizing this but I'll pick it up for moving to pallas in a different change.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -250,42 +253,210 @@ def _patched_flatten_embeddings(embeddings: NestedTensors) -> torch.Tensor:
     return torch.cat(tuple(_patched_flatten_embeddings(t) for t in embeddings))
 
 
+def _jax_compatible_merge_multimodal_embeddings(
```
lk-chen, line 256: @anthonsu is trying to resolve this from vllm side https://github.com/vllm-project/vllm/pull/44474

PR #2723, opened 2026-05-26
[DP] Use rank-0-elected canonical ts for MPMD xprof merge

lk-chen:
# Description

Follow-up to #2705 that hardens the MPMD profile-merge layout against pre-existing data in `PHASED_PROFILING_DIR`.

The previous approach hoisted each rank's capture from `dp_rank_<N>/plugins/profile/<ts>/` up to `<phase>/plugins/profile/<ts>/` and then collapsed timestamp subdirs into the earliest one. That step enumerated *every* timestamp subdir under `<phase>/plugins/profile/`, including ones left behind by an earlier run, and used `dirs[0]` (sorted) as the canonical target — so a stale dir from a prior session could become the destination and the current run's files would be moved into it, producing a misleading layout under a stale timestamp name.

This PR keeps the function (`_merge_profile_directories`) and call site but replaces the internals: instead of discovering a destination via enumeration, rank 0 elects a single canonical destination timestamp up front and every rank moves into it. Each session writes to a fresh ts dir that rank 0 picks itself; we never enumerate `<phase>/plugins/profile/` to discover a destination, so any stale leftovers there are inert.

## Flow

1. **`_start_profiling`** (per phase): rank 0 picks `datetime.now()` as the canonical ts and writes `<phase>/.canonical_ts_<ppid>` atomically (tmp + `os.replace`). Non-zero DP ranks poll the marker (5 s timeout, 50 ms interval) and read it. PPID in the marker name prevents a prior `vllm serve` run's marker from contaminating the current session.
2. **Capture** lands in the per-rank sandbox `<phase>/dp_rank_<N>/plugins/profile/<jax_ts>/<files>` (unchanged from #2705).
3. **`_merge_profile_directories`** (called after `stop_trace`): each rank moves its files from the sandbox into `<phase>/plugins/profile/<canonical_ts>/dp{N}_<file>` (`dp{N}_` prefix only under `TPU_MULTIPROCESS_DP`). The "merge" is implicit — every rank targets the same `<canonical_ts>` so concurrent moves from different ranks converge in one dir without further coordination. The empty `plugins/` subtree inside `dp_rank_<N>/` is removed; the rank dir keeps its `batch_composition_stats_*.json` files.

## Result after all ranks finish

| Case | Layout |
|---|---|
| Non-MPMD, single host | `<phase>/plugins/profile/<canonical_ts>/t1v-X-w-0.xplane.pb` |
| MPMD single host (4 DPs) | `<phase>/plugins/profile/<canonical_ts>/dp{0,1,2,3}_t1v-X-w-0.xplane.pb` (4 coexist) |

`<canonical_ts>` is picked fresh by rank 0 every session, so subsequent runs land in their own ts dir without touching earlier captures.

## Edge cases

- **Non-zero rank arrives at a phase before rank 0**: 5 s poll covers normal timing skew; on true timeout the rank falls back to its own ts and logs a warning (its capture lands in a separate ts dir from rank 0's — visible split, not corrupt).
- **Ray multi-host**: hosts have different ppids → each host's rank 0 elects independently, so cross-host unification is lost. Worth a follow-up that uses a ray-specific shared coordination point (e.g. via Ray's job id or a shared marker on the cluster filesystem).

# Tests

Replaced the two old `_merge_profile_directories_*` tests (which covered the old hoist+collapse internals) with 6 new ones covering the new internals:

- `test_resolve_canonical_dst_ts_rank_zero_writes_marker`
- `test_resolve_canonical_dst_ts_non_zero_rank_reads_marker`
- `test_resolve_canonical_dst_ts_non_zero_rank_falls_back_on_timeout`
- `test_merge_profile_directories_single_rank`
- `test_merge_profile_directories_mpmd`
- `test_merge_profile_directories_ignores_stale_prior_run_data`

`profiler_fixture` updated to stub the new helpers (`_resolve_canonical_dst_ts`, `_merge_profile_directories`) so the existing state-machine tests don't need a real filesystem for the new code paths.

```bash
pytest tests/runner/test_utils.py -k "phased or profiler or canonical or merge_profile"
# 12 passed
```

Verified end-to-end with 31B MPMD text + MM phased-profile runs (`vllm serve google/gemma-4-31B-it` + `PHASED_PROFILING_DIR`): all 4 DP captures coexist in one canonical ts dir per phase with `dp{0..3}_` prefixes; re-running with the same `PHASED_PROFILING_DIR` leaves the earlier session's ts dir untouched and lands in a new one.

# Checklist

### DIFF
--- a/scripts/merge_xprof.py
+++ b/scripts/merge_xprof.py
@@ -1,0 +1,200 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Combine multiple .xplane.pb captures into a single XSpace .xplane.pb.
+
+Phased MPMD profiling writes one xplane.pb per DP rank into
+`<phase>/plugins/profile/<ts>/dp{0,1,...}_*.xplane.pb`. Tools like the
+TensorBoard profile plugin and `xprof` can already read that whole
+directory as one distributed session, but some workflows want a single
+artifact — e.g. attaching to a bug, feeding into a tool that takes
+exactly one file, or sharing via a single URL. This script merges the
+XSpaces from each input into one output proto.
+
+Plane names are taken from each input as-is; xprof identifies devices/
+hosts by plane name, so coinciding names from peer DP ranks (e.g.
+`/device:TPU:0` on a single host) would conflict. Pass `--prefix-planes`
+to namespace each input's planes with its filename stem (e.g.
+`dp0_t1v-..-w-0: /device:TPU:0`) if you need to keep them distinct.
+
+Usage:
+    # Merge specific files
+    python scripts/merge_xprof.py a.xplane.pb b.xplane.pb -o merged.xplane.pb
+
+    # Merge every .xplane.pb under a dir (typically plugins/profile/<ts>/)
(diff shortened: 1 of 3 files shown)


kyuyeunk (approved):
approved!

PR #2753, opened 2026-05-28
Add support for encoder only attention

ap9272:
## Native ModernBERT & BERT Embeddings Serving on GKE TPU

This PR introduces support for TPU native, high-performance serving of bidirectional encoders in ModernBERT and standard BERT like architectures, in both Embeddings and Sequence Classification modes.

---

## Description

This PR makes the following changes to add support for encoder only attention, needed for BERT+ models

### Core Problems Solved:
1.  **JAX Platform Static Task Routing Bottleneck (`tpu_runner.py`)**:
    In standard `tpu-inference`, the JAX platform runner's `get_supported_tasks()` returned `("embed", )` for any pooling models. This meant that even if a model was booted under sequence classification arguments (`--convert=classify`), the API server was forced strictly into embeddings mode. When a classification query entered, the sequence pooler threw `Unsupported task` and crashed.
2.  **vLLM V1 Abstract Interface Violations (`flash_attn.py`)**:
    Under vLLM V1, custom attention backends must register metadata builders. The custom JAX FlashAttention backend lacked this registration, causing the server to crash during abstract interface validation checks on startup.

#### 1. `tpu-inference` Codebase Alignments:
*   **Dynamic Task Routing (`tpu_runner.py`)**: dynamically routing incoming classification/embeddings requests cleanly:
    ```python
    if runner_type == "pooling":
        return (self.model_config.convert_type, )
    ```
* **Encoder-Only Attention Support (`tpu_inference/layers/vllm/backends/flash_attn.py`)**:
   - Enabled `AttentionType.ENCODER_ONLY` support in the Pallas `FLASH_ATTN` backend (`supports_attn_type` and `PallasAttentionBackendImpl.__init__`).
   - Implemented `_jax_encoder_only_attn_func` which uses `sharded_flash_attention` with `causal=False` to execute attention over the full sequence.
   - Added token padding and dynamic segment ID construction (`build_segment_ids`) to correctly handle packed/concatenated batches of variable-length sequences in the encoder.
   - Added support for sliding window attention used in 2 out of 3 layers of ModernBERT. It builds a mask using the `sliding_window` parameter.
   - Branched `PallasAttentionBackendImpl.forward` to skip KV cache operations entirely for `ENCODER_ONLY` models and execute the non-causal self-attention path.
* **Metadata Builder Registration (`tpu_inference/layers/vllm/backends/flash_attn.py`)**:
   - Introduced a dummy metadata builder `PallasAttentionMetadataBuilder` and registered it via `get_builder_cls()`. The builder generates a shell `AttentionMetadata` object since the actual array shapes and values are compiled dynamically prior to JAX execution.
* **Kernel-Level Encapsulation: (`tpu_inference/layers/common/attention_interface.py` and `tpu_inference/kernels/flash_attention/kernel.py`)**
   - Moved all hardware-specific layout preparation (transposition, token padding, dynamic `SegmentIds` construction, and sliding window attention bias generation) to `attention_interface.py` to `kernel.py`. The high-level interface is now purely functional and stateless.
  - New 3D-Specific Interfaces: Introduced `encoder_only_flash_attention` (in `kernel.py`) and `sharded_encoder_only_attention` (in `attention_interface.py`) to cleanly handle the 3D flat token layout used by vLLM.
  - Strict Backward Compatibility: Fully preserved the original 4D signatures of `flash_attention` and `sharded_flash_attention`. This ensures that other multimodal models relying on the 4D pre-padded interface (such as Qwen2.5-VL and Llama 4 Vision) remain completely unaffected, eliminating any regression risk.

---

# Tests

The changes were tested extensively against the standard **BEIR (Benchmarking Information Retrieval) suite** under the `"scifact"` and `"scidocs"` datasets on zonal **GKE TPU v5p** clusters.

### 1. Sequence Classification serving verification:
Verified that sequence classification maps cleanly, compiles dynamically up to max prompt window sizes on the TPU cores, and outputs valid class probabilities over `/classify` for the following models:
* modernBERT-base
* modernBERT-large
* nemocurator-fineweb-nemotron-4-edu-classifier
* nemocurator-fineweb-mixtral-edu-classifier
* Alibaba-NLP/gte-modernbert-base

```json
{
  "id": "classify-988d0676d0ece897",
  "object": "list",
  "model": "answerdotai/ModernBERT-base",
  "data": [{"index": 0, "label": "LABEL_1", "probs": [0.2750, 0.7249], "num_classes": 2}]
}
```
Evaluation of `Alibaba-NLP/gte-modernbert-base` on scifacts
```
BEIR Benchmarking Completed Successfully! 🏆
Dataset:                    SCIFACT
Model:                      Alibaba-NLP/gte-modernbert-base
Execution Backend:          vLLM Serving Endpoint (http://localhost:8000/v1/embeddings)
NDCG@10:                    0.7650
MAP@100:                    0.7243
Recall@100:                 0.9733
Precision@10:               0.1010
```

### 2. Precision Comparative Verification (Embedding Cosine Correlation Test):
Evaluated embeddings precision across GPU and CPU backbones for the exact same long prompt.

* `Cosine Similarity:          0.99996881`
* `Mean Squared Error (MSE):    8.11649447e-08`
* `Max Absolute Difference:    1.17429250e-03`

Across TPU and CPU

* `Cosine Similarity:          0.99992806`
* `Mean Squared Error (MSE):    1.87301648e-07`
* `Max Absolute Difference:    6.05980948e-03`

NOTE: Needs https://github.com/vllm-project/vllm/pull/43896 to actually work as layer aliasing is broken in vllm


tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -280,3 +342,144 @@ def _jax_attn_func(
...
+
+    assert q_len == k_len, "For encoder_only, token lengths should be the same"
+
+    # Convert the shapes from vLLM's convention to what the attention function expects
+    q_thd = q.reshape(q_len, num_heads, head_size)
+    k_thd = k.reshape(k_len, num_kv_heads, head_size)
+    v_thd = v.reshape(k_len, num_kv_heads, head_size)
+
+    # Swap axes to head-first per kernel limit: [num_heads, num_tokens, head_dim]
+    q_htd = q_thd.swapaxes(0, 1)
```
kyuyeunk: all these logic does not belong in this file. tpu_inference/layers/vllm/backends/flash_attn.py should merely act as an interface beteween vllm and tpu-inference. actual hardware specifc logic should reside in https://github.com/vllm-project/tpu-inference/tree/main/tpu_inference/layers/common

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -29,17 +33,55 @@
 TPU_HEAD_SIZE_ALIGNMENT = 128
 
 
+class PallasAttentionMetadataBuilder(AttentionMetadataBuilder[AttentionMetadata]):
```
kyuyeunk: why are all these metadata builder related logic needed at all?

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: I have moved this logic to tpu_inference/layers/common/attention_interface.py::encoder_only_attention

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: vLLM v1 requires backend classes to have `get_builder_cls()` return a meta data builder. This one is a dummy to fulfill that requirement as actual metadata gets built up during execution

docker/Dockerfile
kyuyeunk, line 1: why is this needed? seems unnecessary?

tpu_inference/layers/vllm/backends/flash_attn.py
kyuyeunk: our existing `PallasAttentionBackend` does not have this and still works perfectly fine?

tpu_inference/layers/vllm/backends/flash_attn.py
kyuyeunk, line 1: i think it's better to just create a separate function for it.

tpu_inference/platforms/tpu_platform.py
kyuyeunk, line 1: umm why is this needed? existing logic worked perfectly fine?

kyuyeunk (commented):
in general, i seeing a lot of added logic that doens't make sense? most of them have perfectly working existing counterpart.

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: From my understanding,
For encoder only attention, vLLM builds a new subclass for attention where it makes use of the metadata builder, https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/attention/encoder_only_attention.py#L27

The GPUModelRunner in vLLM uses this metadata class and a builder to build it 
The TPUModelRunner uses the metadata class but does not use a builder to build it

A few GPU Attention mechanisms make use of the builder to build a new backend subclass and they all need a metadatabuilder and can't just rely on the fly metadata building currently in TPUModelRunner
Ex. [static_sink_attention](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/attention/static_sink_attention.py#L46), [chunked_local_attention](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/attention/chunked_local_attention.py#L37), [cross_attention](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/attention/cross_attention.py#L78)

Now, the TPUModelRunner doesn't actually need the builder class because the metadata it builds on the fly is fine. But the design in vLLM forces the need for a builder class. Hence, the builder class has dummy data.

ap9272: All of the attention backends currently supported in tpu-inference don't make use of this codeflow in vLLM

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: Do you mean an encoder_only_flash_attn.py ?

docker/Dockerfile
ap9272: Removed. These were some local changes that shouldn't have been committed.

tpu_inference/platforms/tpu_platform.py
ap9272: Removed

ap9272:
> in general, i seeing a lot of added logic that doens't make sense? most of them have perfectly working existing counterpart.

I've removed some unwanted code changes that shouldn't have been committed. They got rolled in with a rebase but got rid of them.

tpu_inference/layers/vllm/backends/flash_attn.py
kyuyeunk: okay makes sense!

tpu_inference/layers/common/attention_interface.py
```diff
@@ -585,3 +585,122 @@ def _mla_ragged_paged_attention(q, q_rope, k, k_rope, cache, *args):
...
+
+@functools.partial(
+    jax.jit,
+    static_argnames=(
+        "mesh",
+        "sm_scale",
+        "sliding_window",
+    ),
+)
+def encoder_only_attention(
```
kyuyeunk, line 641: vast majority of the code here are implementation details that does not belong to attention_**interface**. please move them to the inside of flash attention kernel function.

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -166,52 +208,71 @@ def forward(
...
-            self.sliding_window,
-        )
-        vllm_model_wrapper_context.kv_caches[kv_cache_index] = new_kv_cache
-
+        # 1. Common JAX View Conversion
+        q_jax = jax_view(query)
+        k_jax = jax_view(key)
+        v_jax = jax_view(value)
+
+        if self.attn_type == AttentionType.ENCODER_ONLY:
```
kyuyeunk: i think match case is more suitted here.

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -280,3 +341,49 @@ def _jax_attn_func(
...
+    static_argnames=(
+        "mesh",
+        "scale",
+        "head_size",
+        "num_heads",
+        "num_kv_heads",
+        "sliding_window",
+    ),
+)
+def _jax_encoder_only_attn_func(
```
kyuyeunk, line 384: i'm seeing a lot of redundant code with this function and existing `_jax_encoder_only_attn_func`

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -280,3 +341,49 @@ def _jax_attn_func(
     outputs = outputs.reshape(q_len, num_heads * head_size)
 
     return new_kv_cache, outputs
+
+
+@functools.partial(
```
kyuyeunk: nit: use `@jax.jit`

tpu_inference/layers/common/attention_interface.py
ap9272: moved to kernels

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: changed

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: I have tried to combine them but the lack of kv cache for attention only decoder makes jit not work

tpu_inference/kernels/flash_attention/kernel.py
```diff
@@ -72,6 +74,121 @@ def get_default(cls, batch_size, num_heads, q_seq_len, kv_len, d_model):
         )
 
 
+@functools.partial(
```
kyuyeunk: nit: use `@jax.jit`

tpu_inference/kernels/flash_attention/kernel.py
```diff
@@ -72,6 +74,121 @@ def get_default(cls, batch_size, num_heads, q_seq_len, kv_len, d_model):
...
+
+    if block_sizes is None:
+        block_sizes = BlockSizes.get_default(1, num_heads, q_len, k_len, head_size)
+    block_q = block_sizes.block_q
+    block_kv = block_sizes.block_k
+
+    padded_len_q = align_to(q_len, block_q)
+    padded_len_kv = align_to(k_len, block_kv)
+
+    q_pad_htd = pad_token(q_htd, padded_len_q - q_len)
```
kyuyeunk, line 124: why are these padding, aligning needed only for encoder only attention?

ap9272: Emulating https://huggingface.co/blog/modernbert#global-and-local-attention
So not limited to encoder only attention but for modernbert style encoder only attention. For BERT, or regular encoder only attention, this will not get triggered.

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -28,6 +31,37 @@
 # TPU requires the head size to be a multiple of 128.
 TPU_HEAD_SIZE_ALIGNMENT = 128
 
+# In recent TPU generations, up to v6e, the SMEM size is 1MB.
+# We limit metadata size to half of SMEM capacity to avoid OOM.
+HALF_SMEM_CAPACITY_BYTES = 512 * 1024
```
kyuyeunk: can you use `pltpu.get_tpu_info()` instead? the same goes for `TPU_HEAD_SIZE_ALIGNMENT`

tpu_inference/layers/common/attention_interface.py
kyuyeunk: thanks!

tpu_inference/kernels/flash_attention/kernel.py
kyuyeunk: okay so this isn't encoder only attention but more about local attention thing? in generally, i really try to maintain first principle thinking - so if this feature isn't actually encoder specific and instead more of global / local attention logic, it makes more sense to have branching logic based on global / local attention.

btw, feel free to correct be since there's a very high possibility that i don't know what i'm talking about.

ap9272: You are correct.
Decoder only attention (stateful attention) 
- global attention (already exists)
- local attention (don't know if it exists)

Encoder only attention (stateless attention ,no KV cache)
- global attention (added in this PR)
- local attention (added in this PR)

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: Changed

tpu_inference/layers/vllm/backends/flash_attn.py
```diff
@@ -25,8 +29,47 @@
 
 logger = init_logger(__name__)
 
-# TPU requires the head size to be a multiple of 128.
-TPU_HEAD_SIZE_ALIGNMENT = 128
+def get_tpu_head_size_alignment() -> int:
+    try:
```
kyuyeunk: can you elaborate when this will happen? in this kind of cases, isn't better to just throw an assertion error?

kyuyeunk (approved):
soft approval.

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: When no tpu, i.e. get_tpu_info() gives blank. I've switched to AssertionErrors instead

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: I have reverted this change. This was not part of the original plan and was causing a lot of failures in the CI with tpu locks

tpu_inference/layers/vllm/backends/flash_attn.py
ap9272: I have reverted this change. This was not part of the original plan and was causing a lot of failures in the CI with tpu locks

Lumosis (approved):

PR #2729, opened 2026-05-27
Remove unused fields from _VllmRunner

wdhongtw:
# Description

As title, remove unused fields from _VllmRunner.

These fields are added two weeks ago by 441316ba.

# Tests

No test required.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2762, opened 2026-05-29
Refactor torchax path to use reparametrize helper

wdhongtw:
# Description

Replaces functional_call with a cleaner reparametrize context manager
using PyTorch's _reparametrize_module. Thus, we can call the required
method as usual and the IDE support can be back to us.

With this change, we simply the code structure by removing unnecessary indirection,
increase the readability and maintainability.
Also, developers can relies on rich IDE support such as auto-complete and pick-definition.

# Tests

A sample workload to cover sufficient potion of changes in this PR.

```sh
MODEL_IMPL_TYPE=vllm python examples/generate/multimodal/vision_language_offline.py --model-type=qwen3_vl --num-prompts=1 --modality=image --time-generate  -tp=4
```

# Checklist

### DIFF
--- a/tests/layers/vllm/process_weights/test_cleanup_sharding.py
+++ b/tests/layers/vllm/process_weights/test_cleanup_sharding.py
@@ -14,10 +14,13 @@
 
 from unittest.mock import MagicMock, patch
 
+import jax
+import jax.sharding
 import torch
+import torchax
 
-from tpu_inference.layers.vllm.process_weights.cleanup_sharding import \
-    shard_model_to_tpu
+from tpu_inference.layers.vllm.process_weights.cleanup_sharding import (
+    reparametrize, shard_model_to_tpu)
 
 
 def test_shard_model_to_tpu_simplification():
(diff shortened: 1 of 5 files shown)


wdhongtw:
Will rebase again once #2729 is merged.

tests/layers/vllm/process_weights/test_cleanup_sharding.py
```diff
@@ -14,10 +14,13 @@
...
 import torch
+import torchax
 
-from tpu_inference.layers.vllm.process_weights.cleanup_sharding import \
-    shard_model_to_tpu
+from tpu_inference.layers.vllm.process_weights.cleanup_sharding import (
+    reparametrize, shard_model_to_tpu)
 
 
 def test_shard_model_to_tpu_simplification():
```
wdhongtw, line 26: The logic in this test is not changed, just tidy up.

yuyanpeng-google:
Just curious if there is a difference between `reparametrize` and `functional_call` . Do they do similar things and incur similar api call overhead?

wdhongtw:
> Just curious if there is a difference between `reparametrize` and `functional_call` . Do they do similar things and incur similar api call overhead?

These two both arrive at `torch.nn.utils.stateless._reparametrize_module`, and the original functional_call has more static checks. The overhead of `reparametrize` in my patch here should be less from this aspect.


```python
@exposed_in("torch.func")
def functional_call(
    module: "torch.nn.Module",
    parameter_and_buffer_dicts: Union[dict[str, Tensor], Sequence[dict[str, Tensor]]],
    args: Optional[Union[Any, tuple]] = None,
    kwargs: Optional[dict[str, Any]] = None,
    *,
    tie_weights: bool = True,
    strict: bool = False,
):
    ... # lots checking here
    return nn.utils.stateless._functional_call(
        module,
        parameters_and_buffers,
        args,
        kwargs,
        tie_weights=tie_weights,
        strict=strict,
    )

def _functional_call(
    module: "torch.nn.Module",
    parameters_and_buffers: dict[str, Tensor],
    args: Any | tuple | None = None,
    kwargs: dict[str, Any] | None = None,
    *,
    tie_weights: bool = True,
    strict: bool = False,
):
    ... # also lots of checking here
    with _reparametrize_module(
        module, parameters_and_buffers, tie_weights=tie_weights, strict=strict
    ):
        return module(*args, **kwargs)
```

lk-chen (approved):
Approve `model_patcher.py`, cc @kyuyeunk for tpu_inference/models/vllm/vllm_model_wrapper.py

wdhongtw:
Hi @kyuyeunk , could you have a look on this change?

tpu_inference/layers/vllm/process_weights/cleanup_sharding.py
```diff
@@ -39,8 +42,32 @@
...
+    """Temporary replace the parameters and buffers for the give model."""
+    # While we are using a "private" interface _reparametrize_module here,
+    # it's quite stable since the first introduction from 2021.
+    # See: https://github.com/pytorch/pytorch/pull/61447
+
+    # Note: Coupled with the implementation of shard_model_to_tpu.
+    # The tie_weights need to be True, since we are using the interfaces,
+    # named_parameters and named_buffers, and the linked parameter/buffer
+    # is not included in the extracted tensors.
+    with torch_stateless._reparametrize_module(
```
kyuyeunk, line 59: the use of private api does seem a bit... scary? at least do you have any examples of other major frameworks using this api? if there is one, it does give me more confidence that this is a stable feature that's unlikely to be broken.

wdhongtw: Yes. JittableModule from torchax is implemented with this interface.
But that's probably not "major" framework. (It's moved from `pytorch/xla` to `google` last year.)
Thus, it's not a proof that the usage won't break in the future.

Still, if the interface is changed in future PyTorch version.
We can check how `torch.func.functional_call` works in the new version and
do corresponding change when we upgrade PyTorch.

@kyuyeunk

wdhongtw: @kyuyeunk Just wondering if you need more time for a proper decision?

We are going to merge #2804. But currently 2804 also depends on this PR. I can adjust both 2804 and 2762 (this one) if you think we need more discussion here.

kyuyeunk: sorry for late reply. approved!

kyuyeunk (approved):

wdhongtw:
Thanks for the review.

Convert to draft, to address the change in 2804.

PR #2768, opened 2026-05-29
[CI] Fix local disagg test

richardsliu:
# Description

Fix a local disagg test failure by cleaning up previous vllm instances.

The test passes on its own, so it is suspected that there are leftover vllm instances holding on to resources.

Similar cleanup was done here: https://github.com/richardsliu/tpu-inference/blob/main/examples/disagg/run_disagg_single_host.sh#L115-L126

# Tests

commands to reproduce.

# Checklist


.buildkite/scripts/cleanup_vllm.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "Cleaning up any running vLLM instances..."
+pkill -f "vllm" || true
+pkill -f "toy_proxy_server" || true
+sleep 5
+pkill -9 -f "vllm" || true
+pkill -9 -f "toy_proxy_server" || true
+fuser -k -9 /dev/vfio/* || true
+fuser -k -9 /dev/accel* || true
+rm -rf /tmp/jax_cache_* || true
```
yiw-wang: this will break the symlink we built to move the persistent cache from /root to /mnt/disk/persist/ and the persistent cache will be loaded to /tmp/ again by other people's PR tests and take down the VMs again by either `No space left on device` error or just get disconnected

yiw-wang (changes requested):

.buildkite/scripts/cleanup_vllm.sh
richardsliu: Changed it to just terminating the vllm processes.

.buildkite/features/Single-Host-P-D-disaggregation.yml
```diff
@@ -20,6 +20,7 @@ steps:
       queue: "${TPU_QUEUE_MULTI:-tpu_v6e_8_queue}"
     commands:
       - |
+        .buildkite/scripts/cleanup_vllm.sh && \
```
mrjunwan-lang, line 23: my recommendation is remove this test (delete this yaml file) as we do not need this test (Pathways solution) any more

QiliangCui (approved):

.buildkite/features/Single-Host-P-D-disaggregation.yml
richardsliu: Thanks, opened https://github.com/vllm-project/tpu-inference/pull/2783 instead.

PR #2771, opened 2026-05-30
[MTP] Gemma4 MTP implementation

Lumosis:
# Gemma4 MTP (Multi-Token Prediction) Implementation & KV-Sharing Support
## Summary
This PR introduces support for **Gemma4 Multi-Token Prediction (MTP)** speculative decoding. The core highlight of this implementation is a novel **cross-model KV-cache sharing mechanism** that allows the draft MTP layers to entirely share and redirect to the target (verifier) model's KV cache, significantly reducing the memory footprint during speculative decoding.

### 1. MTP KV-Cache Sharing
*   **New Utility (`tpu_inference/models/common/kv_share.py`)**: Implemented `compute_mtp_kv_share_map` to dynamically map Gemma4 MTP draft layers to their corresponding target verifier layers of the same type.
*   **Cache Redirection (`tpu_inference/runner/kv_cache_manager.py`)**: Integrated the KV-share map into `KVCacheManager`. When Gemma4 MTP is enabled, draft layers (e.g., `draft_layer.i`) are registered to redirect to the target model's KV cache layers (e.g., `layer.j`) via `self.shared_kv_cache_layers`, avoiding redundant cache allocation.
### 2. Gemma4 MTP Model & Weight Loading
*   **Layer Renaming (`tpu_inference/models/jax/gemma4_mtp.py`)**: Renamed MTP draft layers to use the `draft_layer.` prefix to distinguish them from target layers and enable clean KV-sharing redirection.
*   **LM Head & Embeddings (`tpu_inference/models/jax/gemma4_mtp.py`)**: 
    *   Switched to `JaxLmHead` for the LM head implementation.
    *   Added support for tied word embeddings during weight loading (unconditionally yielding `lm_head.weight` when `model.embed_tokens.weight` is loaded if embeddings are tied).
*   **Weight Permutation (`tpu_inference/models/jax/utils/weight_utils.py`)**: Added `centroids.weight` to the weight loader's permutation logic to support ordered embeddings.
### 3. Speculative Decoding Engine (`tpu_inference/spec_decode/jax/eagle3.py`)
*   **Constant Draft Positions**: Introduced `constant_draft_positions` (enabled when `use_gemma4_mtp` is True). In this mode, positions, sequence lengths, and block tables remain constant during the speculation loop rather than advancing sequentially, matching the MTP verifier cache sharing behavior.
*   **Query Start Location Helper**: Added `_get_loop_query_start_loc` to correctly generate `query_start_loc` inside the speculation loop when positions are constant.
*   **Embedding Sharing**: Enforced that the draft model's embed tokens unconditionally share the target model's embedding values when Gemma4 MTP is active.
### 4. TPU Runner & Async Scheduling (`tpu_inference/runner/tpu_runner.py`)
*   **Token Substitution (`_prepare_async_token_substitution_indices`)**: 
    *   Updated the helper to accept `scheduler_output` to correctly identify which requests are active in speculative decoding.
    *   Skipped token substitution for prefill requests (including chunked prefill).
    *   Enforced that normal decode requests (not doing spec decode) only substitute exactly 1 token.
---
## Verification / Testing
*   Leverages existing E2E speculative decoding integration tests (e.g., MTP + async scheduling + DP tests).
*   Verified that draft layers correctly redirect to target KV caches without extra allocation.

## Manual Verification

* Server command:
```
VLLM_XLA_CHECK_RECOMPILATION=1   MODEL_IMPL_TYPE=flax_nnx DRAFT_MODEL_IMPL_TYPE=flax_nnx vllm     serve google/gemma-4-31B-it     --max-model-len=2048     --max-num-seqs=1024     --tensor-parallel-size 8     --max-num-batched-tokens 16384     --no-enable-prefix-caching  --gpu-memory-utilization=0.8     --async-scheduling     --limit-mm-per-prompt '{"image": 0, "video": 0, "audio": 0}'     --block-size=256 --speculative-config '{"model": "google/gemma-4-31B-it-assistant", "num_speculative_tokens": 4}'
```

* Benchmark results:
```
(vllm_env) ranlihao_google_com@t1v-n-c23b5fbf-w-0:~/tpu-inference$ vllm bench serve --model=google/gemma-4-31B-it  --dataset-name=spec_bench --dataset-path=question.jsonl --spec-bench-category=summarization --num-prompt=100
INFO 06-07 22:52:59 [__init__.py:59] TPU info: node_name=ranlihao-v7e8 | tpu_type=tpu7x-8 | worker_id=0 | num_chips=8 | num_cores_per_chip=2
WARNING:absl:Tensorflow library not found, tensorflow.io.gfile operations will use native shim calls. GCS paths (i.e. 'gs://...') cannot be accessed.
INFO 06-07 22:53:02 [importing.py:46] Triton is installed but 0 active driver(s) found (expected 1). Disabling Triton to prevent runtime errors.
INFO 06-07 22:53:02 [importing.py:81] Triton not installed or not compatible; certain GPU-related functions will not be available.
WARNING 06-07 22:53:03 [tpu_platform.py:354] Pin memory is not supported on TPU.
INFO 06-07 22:53:03 [attention_interface.py:55] Using default RPA kernel
WARNING 06-07 22:53:03 [interface.py:255] Failed to import from vllm._C: ModuleNotFoundError("No module named 'vllm._C'")
WARNING 06-07 22:53:03 [interface.py:255] Failed to import from vllm._C: ModuleNotFoundError("No module named 'vllm._C'")
WARNING 06-07 22:53:03 [interface.py:255] Failed to import from vllm._C: ModuleNotFoundError("No module named 'vllm._C'")
WARNING 06-07 22:53:04 [interface.py:255] Failed to import from vllm._C: ModuleNotFoundError("No module named 'vllm._C'")
INFO 06-07 22:53:04 [__init__.py:115] Registered model loader `<class 'tpu_inference.models.jax.utils.weight_utils.JaxDummyModelLoader'>` with load format `jax_dummy`
INFO 06-07 22:53:04 [__init__.py:115] Registered model loader `<class 'tpu_inference.models.common.pathways_dummy_loader.PathwaysDummyModelLoader'>` with load format `pathways_dummy`
INFO 06-07 22:53:04 [__init__.py:115] Registered model loader `<class 'tpu_inference.models.vllm.vllm_model_loader.IncrementalModelLoader'>` with load format `tpu_streaming_loader`
WARNING 06-07 22:53:04 [__init__.py:104] Load format `runai_streamer` is already registered, and will be overwritten by the new loader class `<class 'tpu_inference.models.vllm.vllm_model_loader.RunaiIncrementalModelLoader'>`.
INFO 06-07 22:53:04 [__init__.py:115] Registered model loader `<class 'tpu_inference.models.vllm.vllm_model_loader.RunaiIncrementalModelLoader'>` with load format `runai_streamer`
WARNING:root:Duplicate op registration for aten.__and__
Namespace(subparser='bench', bench_type='serve', dispatch_function=<function BenchmarkServingSubcommand.cmd at 0x7e698335d8a0>, trust_remote_code=False, seed=0, num_prompts=100, dataset_name='spec_bench', no_stream=False, dataset_path='question.jsonl', no_oversample=False, skip_chat_template=False, enable_multimodal_chat=False, disable_shuffle=False, custom_output_len=256, custom_ensure_client_side_data=False, spec_bench_output_len=256, spec_bench_category='summarization', sonnet_input_len=550, sonnet_output_len=150, sonnet_prefix_len=200, sharegpt_output_len=None, timed_trace_chunk_hash_size=16, timed_trace_sec_multiplier=1, timed_trace_label_timestamp='timestamp', timed_trace_label_input_length='input_length', timed_trace_label_output_length='output_length', timed_trace_label_hash_ids='hash_ids', blazedit_min_distance=0.0, blazedit_max_distance=1.0, asr_max_audio_len_sec=inf, asr_min_audio_len_sec=0.0, random_input_len=1024, random_output_len=128, random_range_ratio='0.0', random_prefix_len=0, random_batch_size=1, no_reranker=False, random_mm_base_items_per_request=1, random_mm_num_mm_items_range_ratio=0.0, random_mm_limit_mm_per_prompt={'image': 255, 'video': 1}, random_mm_bucket_config={(256, 256, 1): 0.5, (720, 1280, 1): 0.5, (720, 1280, 16): 0.0}, hf_subset=None, hf_split=None, hf_name=None, hf_output_len=None, prefix_repetition_prefix_len=256, prefix_repetition_suffix_len=256, prefix_repetition_num_prefixes=10, prefix_repetition_output_len=128, speed_bench_dataset_subset='qualitative', speed_bench_output_len=4096, speed_bench_category=None, label=None, backend='openai', base_url=None, host='127.0.0.1', port=8000, endpoint='/v1/completions', header=None, max_concurrency=None, model='google/gemma-4-31B-it', input_len=None, output_len=None, tokenizer=None, tokenizer_mode='auto', use_beam_search=False, logprobs=None, request_rate=inf, burstiness=1.0, disable_tqdm=False, num_warmups=0, profile=False, save_result=False, save_detailed=False, append_result=False, metadata=None, result_dir=None, result_filename=None, ignore_eos=False, self_timed=None, percentile_metrics=None, metric_percentiles='99', goodput=None, request_id_prefix='bench-0504d521-', top_p=None, top_k=None, min_p=None, temperature=None, frequency_penalty=None, presence_penalty=None, repetition_penalty=None, served_model_name=None, lora_modules=None, lora_assignment='random', ramp_up_strategy=None, ramp_up_start_rps=None, ramp_up_end_rps=None, ready_check_timeout_sec=0, chat_template_kwargs=None, extra_body=None, skip_tokenizer_init=False, insecure=False, plot_timeline=False, timeline_itl_thresholds='25,50', plot_dataset_stats=False)
INFO 06-07 22:53:06 [datasets.py:315] Oversampled requests to reach 100 total samples.
WARNING: vllm bench serve no longer sets temperature==0 (greedy) in requests by default. The default will be determined on the server side and can be model/API specific. For the old behavior, include --temperature=0.
Starting initial single prompt test run...
Skipping endpoint ready check.
Starting main benchmark run...
Traffic request rate: inf
Burstiness factor: 1.0 (Poisson process)
Maximum request concurrency: None
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:06<00:00, 15.17it/s]
tip: install termplotlib and gnuplot to plot the metrics
Successful requests:                     100       
Benchmark duration (s):                  6.59      
Total input tokens:                      72692     
Total generated tokens:                  13108     
Request throughput (req/s):              15.17     
Output token throughput (tok/s):         1988.01   
Peak output token throughput (tok/s):    1285.00   
Peak concurrent requests:                100.00    
Total token throughput (tok/s):          13012.73  
Mean TTFT (ms):                          1462.14   
Median TTFT (ms):                        1621.97   
P99 TTFT (ms):                           2497.73   
Mean TPOT (ms):                          30.93     
Median TPOT (ms):                        30.03     
P99 TPOT (ms):                           49.30     
Mean ITL (ms):                           86.66     
Median ITL (ms):                         74.30     
P99 ITL (ms):                            543.01    
---------------Speculative Decoding---------------
Acceptance rate (%):                     50.62     
Acceptance length:                       3.02      
Drafts:                                  4345      
Draft tokens:                            17380     
Accepted tokens:                         8797      
Per-position acceptance (%):
  Position 0:                            75.49     
  Position 1:                            55.56     
  Position 2:                            41.31     
  Position 3:                            30.10     
```

# Checklist

### DIFF
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -909,8 +909,6 @@ class JaxAutoWeightsLoader(AutoWeightsLoader):
                     permute_dims = (1, 2, 0)
                 elif "embed_tokens.weight" in name:
                     permute_dims = (0, 1)
-                elif "centroids.weight" in name:
-                    permute_dims = (0, 1)
                 elif "lm_head" in name:
                     permute_dims = (1, 0)
 
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -909,6 +909,8 @@ class JaxAutoWeightsLoader(AutoWeightsLoader):
                     permute_dims = (1, 2, 0)
                 elif "embed_tokens.weight" in name:
                     permute_dims = (0, 1)
+                elif "centroids.weight" in name:
+                    permute_dims = (0, 1)
                 elif "lm_head" in name:
                     permute_dims = (1, 0)
 
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -1741,16 +1741,12 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
 
     def _prepare_async_token_substitution_indices(
             self, req_ids_dp, scheduled_tokens_per_dp_rank,
-            padded_num_scheduled_tokens_per_dp_rank, dp_size,
-            scheduler_output: "VllmSchedulerOutput"):
+            padded_num_scheduled_tokens_per_dp_rank, dp_size):
         """Prepare token substitution indices for async scheduling."""
         # For input_ids substitution.
         token_in_tpu_cur_input_indices_dp = {}
         token_in_tpu_pre_next_tokens_indices_dp = {}
         spec_decode_enabled = (self.speculative_config is not None)
-
-        spec_decode_req_ids = set(scheduler_output.scheduled_spec_decode_tokens
-                                  .keys()) if spec_decode_enabled else set()
 
         for dp_rank in range(dp_size):
             token_in_tpu_cur_input_indices_dp[dp_rank] = []
@@ -1768,20 +1764,25 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
 
             for i, req_id in enumerate(req_ids_dp[dp_rank]):
                 acc_cur_len += num_scheduled_tokens_per_req[i]
-                if req_id not in self._pre_async_results.placeholder_req_id_to_index:
-                    continue
 
                 req_idx = self.input_batch.req_id_to_index[req_id]
                 is_prefill = self.input_batch.num_computed_tokens_cpu[
                     req_idx] < self.input_batch.num_prompt_tokens[req_idx]
 
+                # We need an explicit `is_prefill` check here because of preemption.
+                # If a request is preempted and immediately resumed, it goes back to
+                # the prefill stage. However, its `req_id` might still be in
+                # `_pre_async_results.placeholder_req_id_to_index` from the previous step.
+                # Without this check, we would incorrectly perform token substitution
+                # for a resumed prefill request.
                 if is_prefill:
                     # Skip substitution for prefill requests (including chunked prefill)
                     continue
 
-                req_doing_spec_decode = req_id in spec_decode_req_ids
+                if req_id not in self._pre_async_results.placeholder_req_id_to_index:
+                    continue
 
-                if not req_doing_spec_decode:
+                if not spec_decode_enabled:
                     # Treat as normal decode: substitute 1 token
                     assert num_scheduled_tokens_per_req[i] == 1, (
                         f"Expected 1 token for normal decode request {req_id}, "
@@ -1866,6 +1867,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
             scheduled_tokens_cur_rank = scheduled_tokens_per_dp_rank[rank]
             for i, req_id in enumerate(req_ids_dp[rank]):
                 acc_cur_len += scheduled_tokens_cur_rank[i]
+
                 if req_id not in self._pre_async_results.placeholder_req_id_to_index:
                     continue
 
@@ -1930,8 +1932,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                 token_in_tpu_pre_next_tokens_indices_dp,
             ) = self._prepare_async_token_substitution_indices(
                 req_ids_dp, scheduled_tokens_per_dp_rank,
-                padded_num_scheduled_tokens_per_dp_rank, dp_size,
-                scheduler_output)
+                padded_num_scheduled_tokens_per_dp_rank, dp_size)
 
         self.device_buffer.reset()
 
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -954,6 +954,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         self,
         grammar_output: "GrammarOutput | None",
     ) -> ModelRunnerOutput | AsyncTPUModelRunnerOutput:
+
         if self.execute_model_state is None:
             # This can happen in pipeline parallel case.
             return EMPTY_MODEL_RUNNER_OUTPUT
@@ -1175,6 +1176,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
         # TODO: make _get_input_ids_embeds within this context
         # NOTE: right now, mm model will use embeddings as the input,
         # but text-only model will use input_ids
+
         with self.maybe_forbid_compile:
             with set_forward_context(
                     None,
@@ -1183,6 +1185,8 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                     scheduler_output) as kv_connector_output:
                 # NOTE(Wenlong): It takes both `input_ids` and `inputs_embeds`,
                 # but one of them would be `None`
+
+
                 (self.kv_caches, hidden_states, aux_hidden_states,
                  expert_indices) = self.model_fn(
                      self.state_leaves,
@@ -1197,6 +1201,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                      self.is_first_rank,
                      self.is_last_rank,
                  )
+
             if not self.is_last_rank:
                 assert isinstance(hidden_states, JaxIntermediateTensors)
                 hidden_states.kv_connector_output = kv_connector_output
@@ -1257,6 +1262,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                 lora_metadata,
             )
 
+
         self.execute_model_state = ExecuteModelState(
             scheduler_output=scheduler_output,
             attn_metadata=attn_metadata,
@@ -1309,6 +1315,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
 
         processed_bonus_logits = None
         if spec_decode_metadata is None:
+
             logits = logits.astype(jnp.float32)
             with self.maybe_forbid_compile:
                 next_tokens, processed_logits = sample(
@@ -1317,6 +1324,7 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                     logits,
                     tpu_sampling_metadata,
                 )
+
         else:
             if tpu_sampling_metadata.do_sampling:
                 bonus_rng, rejection_rng = jax.random.split(step_rng)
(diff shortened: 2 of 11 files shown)


tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -909,6 +909,8 @@ def __init__(self,
                     permute_dims = (1, 2, 0)
                 elif "embed_tokens.weight" in name:
                     permute_dims = (0, 1)
+                elif "centroids.weight" in name:
```
lk-chen: plz check comment at line 884, for modules "beyond standard transformers, please consider setting weight_loader."

lk-chen (approved):

jrplatin (approved):

tpu_inference/runner/tpu_runner.py
```diff
@@ -1768,7 +1771,21 @@ def _prepare_async_token_substitution_indices(
                 if req_id not in self._pre_async_results.placeholder_req_id_to_index:
                     continue
 
-                if not spec_decode_enabled:
+                req_idx = self.input_batch.req_id_to_index[req_id]
+                is_prefill = self.input_batch.num_computed_tokens_cpu[
+                    req_idx] < self.input_batch.num_prompt_tokens[req_idx]
+
+                if is_prefill:
```
gxd3, line 1778: in line 1771 - 1772, 

```
if req_id not in self._pre_async_results.placeholder_req_id_to_index:
                    continue
```
should be doing the exact same thing, i.e. skip substitute for prefill tokens segment, 
if we don't change `_prepare_async_token_substitution_indices()`, will some test fail?

tpu_inference/spec_decode/jax/eagle3.py
gxd3, line 310: since Gemma drafter share KV cache with Target model,
This means Gemma drafter model should not run "prefilling" at all (prefill is just to populate the KV cache), therefore we probably need to update the logic in `_prepare_inputs` to skip prefill tokens.

tpu_inference/spec_decode/jax/eagle3.py
```diff
@@ -66,6 +66,8 @@ def __init__(
         self.rng_key = jax.random.key(self.vllm_config.model_config.seed)
```
gxd3, line 66: Just curious:
w.r.t. "Acceptance rate (%):                     50.62  " in the E2E test, wonder if we have run GPU vLLM using the same benchmark to confirm the accept rate is on par between vLLM GPU vs TPU-inference to verify things are correct?

tpu_inference/spec_decode/jax/eagle3.py
gxd3, line 632: qq: where in the code to tell the drafter's layer not to perform kv cache update within the attention kernel? (drafter only read target's kv cache, should not perform write?)

tpu_inference/runner/tpu_runner.py
Lumosis: You are right. Taking a second look, line 1771-1772 does intend to solve the same issue. But it missed the `preemption` case. If a request got preempted&resumed, its req_id will be in placeholder_req_id_to_index while it's in prefill stage. I'll delete the first check.

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: I did some initial testing and it looks good. Extensive testing is in progress

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Please check gemma4_mtp.py line 285

gxd3: saw `update_kv_cache=False` in `Gemma4MTPAttention`.
Thanks for the pointer!

Lumosis: Good point. Updated. In fact, no Eagle3-like draft model (even non-KV-sharing ones like Llama-Eagle3) ever needs to run prefilling over the prompt. The draft model feeds on the target model's output hidden states (which already encode the entire prompt context), the speculator only ever needs 1 token of input (the newly sampled token) to begin drafting.

tpu_inference/runner/tpu_runner.py
```diff
@@ -1765,10 +1768,25 @@ def _prepare_async_token_substitution_indices(
...
+                    # Skip substitution for prefill requests (including chunked prefill)
+                    continue
+
                 if req_id not in self._pre_async_results.placeholder_req_id_to_index:
                     continue
 
-                if not spec_decode_enabled:
+                req_doing_spec_decode = req_id in spec_decode_req_ids
+
+                if not req_doing_spec_decode:
```
gxd3: if "spec_decode_enabled" is true (even if the current req's draft-len=0),
we should take the else code path starting from line 1794,
because the previous sampling output array format is conditioned on whether "spec_decode_enabled":
https://screenshot.googleplex.com/7CSDZC6izKHh7Mw

tpu_inference/spec_decode/jax/eagle3.py
gxd3: for normal/classic Eagle3-like draft model, it does need to run prefill on the prompt tokens (left shifted by 1) to populate the drafter's own KV cache?

tpu_inference/runner/tpu_runner.py
Lumosis: Yea, you are right. Sorry for missing that part. Thanks for pointing out.

tpu_inference/spec_decode/jax/eagle3.py
Lumosis: Sorry for the confusion, I double checked with experiments: it does still need to run prefill on prompt tokens in the first iteration.

gxd3, line 360: discussed offline, let's revert the changes in `prepare_inputs` to make things simpler and less error prone.
Perhaps leave a comment about Gemma4 MTP (doing prefill for Gemma4 drafter is basically no-op, wasting a bit of work (as drafter won't write kv cache), but won't affect correctness)

tpu_inference/runner/tpu_runner.py
gxd3: discussed offline, preemption may affect / trigger bugs elsewhere in the tpu-runner (as preemption affect other things as well beside spec decoding ), perhaps leave a comment about it.

gxd3 (approved):

PR #2777, opened 2026-05-31
[Qwen3.5/perf] SC allreduce and allgather offload threshold.

wyzhang:
Change:
`SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES` to allow overriding SC offload threshold. By defeault set to `auto` which will use vmem size as the threshold. To disable, set it to `0` which will use xla's default threshold.

Why:
When data size is small and can fit in vmem, sc offload can be slower. 
Keeping small collectives on the Tensor Core (TC) avoids sc overhead

Benchmark: 
| Workload | Concurrency | Before | After | Improvement |
| :--- | ---: | ---: | ---: | :--- |
| `8k/1k` | `64` | `12,757` | `13,105` | **+2.7%** |
| `8k/1k` | `128` | `16,374` | `17,305` | **+5.7%** |
| `1k/8k` | `64` | `2,557` | `2,649` | **+3.6%** |
| `1k/8k` | `128` | `4,146` | `4,212` | **+1.6%** |

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -62,6 +62,7 @@ if TYPE_CHECKING:
     ONEHOT_MOE_PERMUTE_THRESHOLD: int = 0
     PROFILE_SINGLE_DEVICE: bool = False
     LORA_MODULE_PATH: str = ""
+    SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES: str = "auto"
 
 
 def env_with_choices(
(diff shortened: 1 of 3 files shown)


kyuyeunk (commented):
offload threshold can be a function of vmem capacity & activation size. see similar pr: https://github.com/vllm-project/tpu-inference/pull/2303

instead of introducing a threshold, can you see if you can formulate this into a equation?

wyzhang:
> offload threshold can be a function of vmem capacity & activation size. see similar pr: #2303
> 
> instead of introducing a threshold, can you see if you can formulate this into a equation?

The default is auto mode that derives the threshold automatically from vmem size.  So this PR already does the same as  #2303. In addition, it allows users to set an explicit value that could be much smaller than vmem  size, or disable this to use the default setting in xla.

helloworld1 (approved):

kyuyeunk (approved):

PR #2779, opened 2026-06-01
Fix mamba/GDN recurrent-state slot lifecycle leaks

yaochengji:
Hybrid linear-attention models (GDN + short conv) keep per-request recurrent state in a physical mamba slot managed by a hand-rolled per-DP-rank free list, separate from the KV-cache block manager. The state lives in the slot, not in the persistent-batch row that condense/swap moves, so the slot must be preserved, freed, or reset at the right lifecycle transition. Several transitions did the wrong thing, letting one request read or overwrite another's recurrent state (silent eval regressions).

- Preserve the slot for temporarily-unscheduled (not finished) requests via a durable CachedRequestState.mamba_state_slot and remove_request( free_mamba_slot=False); reuse it on re-add.
- Reset (free + clear) the slot for preempted/resumed requests so a recomputed request cannot read stale state, including when it already left the batch.
- Free the slot for finished requests, including when not in the active batch.
- Add release_mamba_slot() to return a slot to the correct rank's pool while ignoring null slots and avoiding double-free.
- Add assert_mamba_state_invariants() run after update_states as a safety net.
- Guard assigned_dp_rank to a dict before use.

# Checklist


tpu_inference/runner/input_batch.py
```diff
@@ -190,6 +191,88 @@ def init_mamba_pools(self, mamba_num_blocks: int) -> None:
...
+                        f"Request {req_id} on DP rank {expected_rank} has "
+                        f"mamba slot {slot} from rank {slot_rank}")
+            if requests is not None:
+                req_state = requests.get(req_id)
+                if req_state is not None and req_state.mamba_state_slot != slot:
+                    raise AssertionError(
+                        f"Request {req_id} active slot {slot} does not match "
+                        f"cached slot {req_state.mamba_state_slot}")
+
+        duplicate_active = {
```
qizzzh: list.count() per element is O(N²). For Qwen3.5 production scale (max_num_reqs=512), worst-case fully-occupied batches do ~260K compares per update_states call.

The tail check is also slow: iterates the numpy array in Python with int(slot) casts per element, rather than a vectorized (tail != 0).any().

It runs every step in real workloads — the only gate is isinstance(self.input_batch, InputBatch), which is always true outside unit tests.

Two suggestions, in priority order:

a. Cheap rewrite — preserves the safety net at trivial cost:
```
  if len(set(active_slots)) != len(active_slots):
      from collections import Counter
      dups = sorted(s for s, c in Counter(active_slots).items() if c > 1)
      raise AssertionError(f"Duplicate active mamba slots: {dups}")
  ...
  tail = self.mamba_state_indices_cpu[self.num_reqs:]
  if tail.any():
      raise AssertionError(
          f"Non-zero mamba slots in padded tail: "
          f"{sorted(set(tail[tail != 0].tolist()))}")
```

b. Or gate behind a debug env knob — add a VLLM_DEBUG_MAMBA_INVARIANTS env to tpu_inference/envs.py and skip when off. An always-on invariant is fine if cheap; the current implementation is a measurable per-step tax for a check that only pays off on regression.

qizzzh (approved):

tpu_inference/runner/input_batch.py
yaochengji: Thanks Qi for the careful review! I used a for the fix.

qizzzh (approved):

yaochengji:
@qizzzh thanks for your approval. Seems like it also needs the code owner's approval

cc @jrplatin @kyuyeunk

kyuyeunk (approved):
Approved!

yaochengji:
Thanks @kyuyeunk !

Seems we also need @jrplatin or @wenxindongwork 's approval. The code-owner system in tpu-inference repo is a bit complex :)

sixiang-google (approved):

jrplatin (approved):

PR #2780, opened 2026-06-01
[Qwen3.5/dp-scheduler] Route reqs by considering inflight count and   completed tokens

wyzhang:
Why:
For DP attention, consider inflight count for scheduling which will
improve balance across dp ranks.
Adding rank's max completed-output tokens as a tertiary key biases
admission toward the rank closest to finishing a req -- the rank most
likely to free a slot soon, so any momentary bump resolves quickly.

Benchmark results
* Concurrency 256 variance much smaller ~1.2% (down from 5.3%)
* Concurrency 256 total token throughput improved by ~13.27% to 7,727 (from average 6822 before)
* Concurrency 64 and 128 also gained 4-5% 

| workload | concurrency | total (baseline → +PR) | output (baseline → +PR) | Δ |
  |---|---|---|---|---|
  | 1k/8k | 64  | 2,566 → 2,691 | 2,281 → 2,392 | **+4.88%** |
  | 8k/1k | 64  | 12,832 → 12,861 | 1,424 → 1,427 | +0.20% |
  | 1k/8k | 128 | 4,160 → 4,351 | 3,698 → 3,868 | **+4.60%** |
  | 8k/1k | 128 | 17,074 → 17,264 | 1,894 → 1,915 | +1.13% |
  | 1k/8k | 256 | 6,822 → 7,727 | 6,064 → 6,869 | **+13.27%** |
  | 8k/1k | 256 | 18,686 → 18,713 | 2,073 → 2,076 | +0.15% |
  | 1k/8k | 512 | 8,244 → 8,286 | 7,329 → 7,367 | +0.51% |
  | 8k/1k | 512 | 19,948 → 20,048 | 2,213 → 2,224 | +0.50% |

### DIFF
--- a/tpu_inference/core/sched/dp_scheduler.py
+++ b/tpu_inference/core/sched/dp_scheduler.py
@@ -68,6 +68,7 @@ class SchedulerCommand(Enum):
     GET_REQUEST_COUNTS = "get_request_counts"
     GET_TOKEN_COUNT = "get_token_count"
     GET_PENDING_PREFILL_TOKENS = "get_pending_prefill_tokens"
+    GET_MIN_REMAINING_OUTPUT = "get_min_remaining_output"
     PROBE_COMPUTED_BLOCKS = "probe_computed_blocks"
     RESET_ENCODER_CACHE = "reset_encoder_cache"
     SET_PAUSE_STATE = "set_pause_state"
@@ -252,6 +253,19 @@ def _scheduler_worker_process(
                     for req in scheduler.waiting:
                         pending += req.num_prompt_tokens
                     _send_result(pending)
+
+                case SchedulerCommand.GET_MIN_REMAINING_OUTPUT:
+                    # Smallest (max_tokens - len(output_token_ids)) across
+                    # this rank's running reqs. Used as an admission
+                    # tiebreaker: lower = a running req closer to its
+                    # max_tokens, so this rank will likely free a slot
+                    # soonest.
+                    if not scheduler.running:
+                        _send_result(2**31 - 1)
+                    else:
+                        _send_result(
+                            min(r.max_tokens - len(r.output_token_ids)
+                                for r in scheduler.running))
 
                 case SchedulerCommand.PROBE_COMPUTED_BLOCKS:
                     # Probe for cached blocks without recording prefix cache stats.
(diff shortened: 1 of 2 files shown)


sixiang-google (approved):

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -622,8 +636,10 @@ def _find_best_rank_for_request(self, request: Request) -> int:
             if best_cache_tokens > 0:
                 return best_cache_rank
 
+        inflight = self._get_rank_inflight_reqs()
         pending = self._get_rank_pending_prefill_tokens()
-        return min(pending, key=pending.get)
+        return min(range(self.dp_size),
```
wenxindongwork, line 699: currently this routing strategy handles decode imbalance, but might trigger prefill imbalance right? one rank could get shorter requests while another rank get longer reqests, because `_get_rank_inflight
_reqs` doesn't take into account the number of tokens in each request. It is difficult to balance these two...  

this is upstream's load balancing logic, perhaps also worth a try?

https://github.com/vllm-project/vllm/blob/193ce8812eb4b43fc48039757072c71ac2eb0a28/vllm/v1/engine/core_client.py#L1350-L1378

wyzhang: Updated the code. We want to pick rank first by lowest pending prefill token count. If all dp ranks are doing decode, we use the pending + inflight reqs to balance decode across ranks.  PTAL

wenxindongwork: thanks, this makes sense! 

do you mind adding some unit tests to test for the intended behaviour?

wenxindongwork (approved):

tpu_inference/core/sched/dp_scheduler.py
wyzhang: Will do in a follow PR.

helloworld1 (approved):

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -622,8 +670,13 @@ def _find_best_rank_for_request(self, request: Request) -> int:
             if best_cache_tokens > 0:
                 return best_cache_rank
 
+        inflight = self._get_rank_inflight_reqs()
         pending = self._get_rank_pending_prefill_tokens()
-        return min(pending, key=pending.get)
+        max_out = self._get_rank_max_output_tokens()
```
qizzzh: Each helper sends to all ranks then awaits all replies before returning, so the three waves serialize. With prefix caching off, this triples the per-request routing latency vs. baseline; with prefix caching on but no hit, it goes from 2 → 4 waves.

The workers process commands one-at-a-time off their input pipe, so the three commands per rank can pipeline if you interleave sends:
```
  for rank in range(self.dp_size):
      self._send_command(rank, SchedulerCommand.GET_REQUEST_COUNTS)
      self._send_command(rank, SchedulerCommand.GET_PENDING_PREFILL_TOKENS)
      self._send_command(rank, SchedulerCommand.GET_MAX_OUTPUT_TOKENS)
  # then collect 3*dp_size results
```

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -252,6 +253,19 @@ def _send_result(result):
                         pending += req.num_prompt_tokens
                     _send_result(pending)
 
+                case SchedulerCommand.GET_MAX_OUTPUT_TOKENS:
+                    # Largest len(output_token_ids) across this rank's
```
qizzzh: The heuristic is len(output_token_ids), not max_tokens - len(output_token_ids). These are only equivalent when all running reqs on a rank share max_tokens. For uniform-workload benchmarks (1k/8k, 8k/1k) this is true; for production mixes with heterogeneous max_tokens or early-EOS termination, the "closest to finishing" claim weakens to "has been generating the longest."

Probably use the true distance `min(req.max_tokens - len(req.output_token_ids)` for r in scheduler.running) (and flip the negation)?

wyzhang: Done

tpu_inference/core/sched/dp_scheduler.py
wyzhang: Done.

qizzzh (approved):

PR #2781, opened 2026-06-01
Cap down proj weight block size under GMM_TP

BirdsOfAFthr:
# Description

This PR dynamically caps the down-projection ($w_2$) block-wise requantization size to the device-local dimension size under `GMM_TP` tensor parallelism. This is a change from the past where `MOE_REQUANTIZE_BLOCK_SIZE=512` was applied, which failed with shape errors when the TP mesh was larger than the block size.

* **Why this change is being made**: To prevent JAX sharding indivisibility errors when using block-wise quantization on larger tensor parallel meshes (such as `TP=16`).
* **The problem being solved**: When using a static block size (e.g., `512`), the scale only has `4096 // 512 = 8` elements. On a model partitioned across 16 chips under TP, it will cause indivislbe error.
* **Why this is a good solution**: By dynamically capping the $w_2$ block size to the local partitioned dimension size on the mesh, we eliminate the sharding error completely while preserving the maximum possible block-wise quantization precision on any TP mesh configuration.
* **Specific implementation details**:
  - Modified `tpu_inference/layers/common/process_weights/moe_weights.py` to compute the local dimension size on the target mesh and cap `w2_block_size` accordingly.

# Tests
* I launched Mistral 3 Large on a multi-host setup with TP=16 and hidden dim=4096
* I added a new, fully self-adaptive unit test `test_process_weights_requant_512` in `tests/layers/common/test_process_weights.py` which dynamically adjusts weight shapes and mesh sizes to the physical devices available (`tp_size = len(devices)`), asserting global shapes, sharding specs, and addressable local shard shapes.

### Commands to Reproduce:

* **Test standard CPU fallback (TP=16)**:
``` 
JAX_PLATFORMS='cpu' XLA_FLAGS='--xla_force_host_platform_device_count=16' python3 -m unittest tests.layers.common.test_process_weights.TestProcessWeights.test_process_weights_requant_512
```
# Checklist

### DIFF
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -1092,11 +1092,12 @@ def _process_quantized_moe_weights_impl(
     else:
         w13_block_size = w2_block_size = requant_block_size
 
-    tp_size = get_mesh_shape_product(mesh, ShardingAxisName.MLP_TENSOR)
-    max_w2_block_size = orig_intermediate_size // tp_size
+    if requant_block_size is not None and moe_backend == MoEBackend.GMM_TP:
+        tp_size = get_mesh_shape_product(mesh, ShardingAxisName.MLP_TENSOR)
+        max_w2_block_size = orig_intermediate_size // tp_size
 
-    # Cap the block size to avoid sharding indivisible errors
-    w2_block_size = min(w2_block_size, max_w2_block_size)
+        # Cap the block size to avoid sharding indivisible errors
+        w2_block_size = min(w2_block_size, max_w2_block_size)
     hidden_size = align_to(orig_hidden_size, w13_block_size)
     intermediate_size = align_to(orig_intermediate_size, w2_block_size)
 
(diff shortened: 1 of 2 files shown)


tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -1082,8 +1087,17 @@ def _process_quantized_moe_weights_impl(
     if requant_block_size is None:
         w13_block_size = w13_weight.shape[-1]
         w2_block_size = w2_weight.shape[-1]
+    elif isinstance(requant_block_size, tuple):
+        w13_block_size, w2_block_size = requant_block_size
     else:
         w13_block_size = w2_block_size = requant_block_size
+
+    if requant_block_size is not None and moe_backend == MoEBackend.GMM_TP:
```
kyuyeunk, line 1095: why is a separate code path for tp is needed? if my understanding is correct, tp path only changes dimension size, so the actually overriding logic can be shared across ep & tp - and we can just tweak the dim size based on tp vs. ep right?

BirdsOfAFthr: Right. Initially, I didn't limit this to MoEBackend.GMM_TP, but that caused several EP tests to fail. Since we don't actually shard the contracting dimension in EP, and it's generally large enough anyway—I restricted this to TP to maintain backward compatibility.

kyuyeunk (approved):

PR #2792, opened 2026-06-02
[DeepSeek V4] initial bring up skeleton PR

gxd3:
[DS V4] initial bring up skeleton PR

Main change:
* add VllmDeepseekV4Fp8Config for "deepseek_v4_fp8" format, it dispatch to VllmMxfp4MoEMethod or VllmFp8MoEMethod base on expert dtype, it mirrors the behavior vLLM upstream's DeepseekV4FP8Config.

* Register pytorch based implementation for mHC ops.

* Added VllmDeepseekV4MLA, VllmDeepseekV4MLAAttention as empty pass-through skeleton.

TESTED=can start a vLLM server with DS V4 Flash: https://paste.googleplex.com/6479272792424448


tpu_inference/layers/common/quant_methods.py
```diff
@@ -18,3 +18,4 @@
 COMPRESSED_TENSORS = "compressed-tensors"
 FP8 = "fp8"
 NVFP4 = "modelopt_fp4"
+DSV4_FP8 = "deepseek_v4_fp8"
```
jrplatin, line 21: Why do we need a new quant method? The chkpt looks like default FP8?

Image

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -150,7 +150,11 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
         weight = jnp.transpose(weight)
         delattr(layer, "weight")
 
-        weight_scale = t2j(layer.weight_scale_inv, use_dlpack=False)
+        weight_scale_inv = layer.weight_scale_inv
+        # Float8_e8m0fnu (ue8m0) scales cannot be converted via numpy in t2j.
+        if weight_scale_inv.dtype == torch.float8_e8m0fnu:
+            weight_scale_inv = weight_scale_inv.to(torch.float32)
```
jrplatin, line 157: I think this means the scales will be loaded in as FP32 obviously right? For now, I don't think this an issue because our kernels / quantization code expects/will convert to FP32 anyways, but this might not be optimal in the sense that your increasing the memory footprint by 4x for your scales

tpu_inference/layers/common/quant_methods.py
gxd3: `deepseek_v4_fp` is a new format,
vLLM upstream also has this: https://screenshot.googleplex.com/99JoGYQ3HnHTWHZ, https://screenshot.googleplex.com/AEnuSoAjkmaciyf

We just mirror it: dispatch to VllmMxfp4MoEMethod or VllmFp8MoEMethod base on expert dtype.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
```diff
@@ -0,0 +1,152 @@
...
+from vllm.v1.kv_cache_interface import KVCacheSpec, MLAAttentionSpec
+
+from tpu_inference.layers.vllm.backends.flash_attn_mla import \
+    PallasMLAttentionBackend
+from tpu_inference.logger import init_logger
+
+logger = init_logger(__name__)
+
+
+# TODO: implement VllmDeepseekV4MLAAttention.
```
jrplatin: Maybe raise a `NotImplementedError` or something of the equivalent?

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
```diff
@@ -0,0 +1,152 @@
...
+            self,
+            q: torch.Tensor,  # [T, num_heads, head_dim]
+            kv: torch.Tensor,  # [T, 1, head_dim]
+            positions: torch.Tensor,
+            output: torch.Tensor,  # [T, num_heads, head_dim]
+    ) -> None:
+        return q
+
+
+# TODO: Implement VllmDeepseekV4MLA.
```
jrplatin: Same comment as above

tpu_inference/layers/vllm/custom_ops/fused_moe.py
```diff
@@ -53,5 +53,6 @@ def forward(
         self,
         hidden_states: torch.Tensor,
         router_logits: torch.Tensor,
+        input_ids: torch.Tensor | None = None,
```
jrplatin: Why is this change needed?

tpu_inference/layers/vllm/custom_ops/mhc.py
jrplatin, line 1: Is this code hooked up anywhere or just dead right now? If the former, then might be good to raise `NotImplementedErrors`

tpu_inference/layers/vllm/quantization/deepseek_v4_fp8.py
jrplatin, line 1: Can you help me understand why this is needed?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -69,6 +70,83 @@
 logger = init_logger(__name__)
 
 
+@contextmanager
+def _maybe_patch_for_deepseek_v4(vllm_config: VllmConfig):
+    architectures = getattr(vllm_config.model_config.hf_config,
+                            "architectures", None) or []
+    is_ds_v4 = any("DeepseekV4ForCausalLM" in arch for arch in architectures)
+    if not is_ds_v4:
+        yield
```
jrplatin, line 99: Is this a typo or is the yield, return combo on purpose?

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -69,6 +70,83 @@
...
+    # DeepseekV4MLA is a plain nn.Module instantiated directly in
+    # amd/model.py (not CustomOp/register_oot hook).
+    # Swap the class symbol for the TPU subclass before the
+    # model is built.
+    from tpu_inference.layers.vllm.custom_ops.deepseek_v4_attention import \
+        patch_deepseek_v4_mla_cls
+    patch_deepseek_v4_mla_cls()
+
+    # DeepSeek-V4 builds CUDA streams (``torch.cuda.Stream``) at construction
+    # which is unavailable on TPU. Mock it to return None.
```
jrplatin, line 128: If we are using the AMD version of the code, why is this an issue?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -531,7 +540,8 @@ def get_kv_cache_spec(self):
         else:
             # Else propagate attention modules from compilation config.
             layers = get_layers_from_vllm_config(
-                self.runner.vllm_config, (Attention, MLAAttention, MambaBase))
+                self.runner.vllm_config,
```
jrplatin, line 550: Why is this change needed?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -573,6 +584,20 @@ def get_kv_cache_spec(self):
                         kv_cache_spec[layer_name] = spec
                     continue
 
+                cache_for_ds_v4 = isinstance(
```
jrplatin: Might be good to move this logic into `tpu_inference/runner/utils.py`

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -824,18 +849,9 @@ def _allocate_mamba(c_shape=cache_shape,
                     # We should only init a new kv cache for the first layer in shared_by
                     # if duplicate_shared_layers is False.  Otherwise, if duplicate_shared_layers
                     # is True, we should init a new kv cache for each layer in shared_by
```
jrplatin, line 853: Why are you removing all of this logic? Maybe I am reading this wrong but it seems like the new logic isn't backwards compatible / equivalent to the old logic (fro DSv3)?

tpu_inference/layers/vllm/quantization/deepseek_v4_fp8.py
gxd3: this is same as above:

deepseek_v4_fp is a new format,
vLLM upstream also has this: https://screenshot.googleplex.com/99JoGYQ3HnHTWHZ, https://screenshot.googleplex.com/AEnuSoAjkmaciyf
We just mirror it: dispatch to VllmMxfp4MoEMethod or VllmFp8MoEMethod base on expert dtype.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
gxd3: Added log error for the "not implemented" status.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
gxd3: I want to be able to start the vLLM server (to expose failure if any), so i use pass-through for these new module.
But added log error for the "not implemented" status.

tpu_inference/layers/vllm/custom_ops/fused_moe.py
gxd3: this is needed for https://github.com/vllm-project/tpu-inference/pull/2800#pullrequestreview-4422272979.

Revert it here.

tpu_inference/runner/kv_cache_manager.py
gxd3: because we need to match against DeepseekV4IndexerCache, DeepseekV4SWACache, DeepseekV4MLAAttention, CompressorStateCache. As below.

tpu_inference/runner/kv_cache_manager.py
gxd3: I think the original code (line 833-835) is not actually needed for DS V3, it's just a small cleanup.
(We have E2E test for DS V3 in the buildkite, it verifies the DS V3 is not broken by this)

tpu_inference/models/vllm/vllm_model_wrapper.py
gxd3: it's intended,
basically it means: ""for the non-DS-V4 case, just run the block as-is and do nothing else.""

tpu_inference/models/vllm/vllm_model_wrapper.py
gxd3: without this, we'll hit error like: https://paste.googleplex.com/6473464587354112, this is needed.
(the amd model.py call cuda.Stream() also)

tpu_inference/runner/kv_cache_manager.py
gxd3: extracted this few line into a helper function put at the top of this file.

tpu_inference/layers/vllm/quantization/fp8.py
gxd3: Make sense, thanks for pointing it out!
(Will optimize it later after initial bringup)

tpu_inference/layers/vllm/custom_ops/mhc.py
gxd3: They are hooked already.
Added log error for the "not implemented" status.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
jrplatin: Sg!

tpu_inference/layers/vllm/quantization/deepseek_v4_fp8.py
```diff
@@ -0,0 +1,84 @@
...
+                return VllmFp8LinearMethod(self, linear_config)
+            case FusedMoE():
+                if is_layer_skipped(
+                        prefix=prefix,
+                        ignored_layers=self.ignored_layers,
+                        fused_mapping=self.packed_modules_mapping,
+                ):
+                    return VllmUnquantizedFusedMoEMethod(layer.moe_config)
+
+                if self.expert_dtype == "fp4":
```
jrplatin, line 76: Are there cases when the expert_dtype isn't fp4? Where is `expert_dtype` coming from?

tpu_inference/runner/kv_cache_manager.py
jrplatin: Ok I see, so `layer_spec.head_size` now respects the `head_size = text_config.kv_lora_rank + \ text_config.qk_rope_head_dim` logic, got it

tpu_inference/layers/vllm/quantization/deepseek_v4_fp8.py
gxd3: expert_dtype inherit from vLLM's DeepseekV4FP8Config base class:
https://screenshot.googleplex.com/AfGvRjpEKL8DboH

expert type can only be fp4 or fp8: https://screenshot.googleplex.com/7cMnwgjtpFhV4Zf

tpu_inference/layers/vllm/quantization/fp8.py
gxd3: Left a TODO for future optimization

jrplatin (approved):

sixiang-google (approved):

vanbasten23:
I wonder how you can check if the implementation is correct.

vanbasten23:
Suggest to add some tests.

vanbasten23 (approved):

PR #2799, opened 2026-06-02
[Spec Decoding] Support Logprobs with Speculative Decoding

Lumosis:
### Summary

This PR enables correct logprob extraction and materialization during speculative decoding in the JAX/TPU runner, resolving an `IndexError: list index out of range` crash that occurs under HTTP workloads when `logprobs` is enabled. It also fixes compilation shape mismatches and embedding-sharing shape mismatches when target and draft models have different dimension configurations (e.g. LLaMA-3.3-70B with EAGLE3-70B).

#### 1. Speculative Logprobs Materialization (`tpu_inference/runner/tpu_runner.py`)
- **Resolved Async Request Mismatch**: In async scheduling, device output for step $N-1$ is resolved during step $N$'s execution. Modified the materialization pipeline to pass the snapshotted request count `num_reqs` (from step $N-1$) instead of reading the active runner's `runner.input_batch.num_reqs` (which would be 0 during empty cycles, leading to empty logprobs and length mismatch crashes in vLLM server).
- **Logprobs Parsing for Multi-Request Batches**: Implemented sequence-specific parsing in `_jax_logprobs_materialize`. It now calculates sequence slice offsets using the `padded_tokens_per_seq` logic, filtering out placeholder tokens and integrating the bonus token logprobs per request.
- **Non-Speculative Path Safety**: Added a fallback in `_jax_logprobs_materialize` to populate `cu_num_generated_tokens` in non-speculative execution paths (such as prefill or non-speculative decode steps) to ensure the API server always receives formatting metadata, preventing formatting crashes.
#### 2. Rejection Sampler Alignment (`tpu_inference/layers/jax/sample/rejection_sampler.py`)
- **Aligned Token Slicing**: Replaced the sequential `start_idx` calculation with block-aligned slicing based on `padded_tokens_per_seq`. This matches the JAX padded output layout per request in multi-request batches.
#### 3. Logits Padding & Sharding Shape Alignment
- **`tpu_inference/runner/tpu_runner.py`**: Corrected `padded_logits_length` calculation under speculative decoding to base padding size on `padded_num_reqs_per_dp_rank + total_active_draft_tokens` rather than `total_sampled_tokens`, preventing shape mismatches.
- **`tpu_inference/runner/speculative_decoding_manager.py`**: Aligned target logits padding to `padded_logits_length_dp_rank - padded_num_reqs_per_dp_rank`.
- **`tpu_inference/runner/compilation_manager.py`**: Added parameter padding variations to the precompilation manager to precompile targets for all possible scheduled request configurations.
- **`tpu_inference/runner/utils.py`**: Updated `host_extract_sampled_tokens` to use `target_logits_indices` shape instead of `final_logits_indices` shape, and bypass speculative token extraction entirely if no draft tokens are scheduled.
#### 4. Robust Embedding Parameter Sharing (`tpu_inference/models/vllm/vllm_model_wrapper.py`)
- **Dimension Shape Protection**: Added checks to skip embedding parameter sharing (e.g. `embed_tokens.weight`) between draft and target models if they have mismatched hidden shapes. This enables models with differing configurations (like LLaMA-3.3-70B with 8192 hidden size and EAGLE3-70B with 6144 hidden size) to run speculative decoding by loading weights independently.
- **Layout Sharding Adjustments**: Adjusted JIT sharding configurations for hidden states combinations to prevent layout compilation conflicts.
#### 5. Tests
- Added `test_eagle3_logprobs_correctness` to verify speculative decoding with logprobs under EAGLE3 model settings.
- Added logprobs length verification assertions to `_test_correctness_helper` in E2E tests.
---
### Verification Plan
Verification was completed using python test suites.
To run the speculative decoding e2e correctness tests:
```bash
python3 -m pytest tests/e2e/test_speculative_decoding.py -k test_eagle3_logprobs_correctness
```

# Checklist

### DIFF
--- a/tpu_inference/platforms/tpu_platform.py
+++ b/tpu_inference/platforms/tpu_platform.py
@@ -7,6 +7,21 @@ import jax.numpy as jnp
 import numpy
 import torch
 import vllm.envs as vllm_envs
+
+# Monkeypatch torch.accelerator.empty_cache to ignore device_allocator error on TPU.
+if hasattr(torch, "accelerator") and hasattr(torch.accelerator, "empty_cache"):
+    _orig_empty_cache = torch.accelerator.empty_cache
+
+    def _patched_empty_cache(*args, **kwargs):
+        try:
+            _orig_empty_cache(*args, **kwargs)
+        except RuntimeError as e:
+            if "Allocator for jax is not a DeviceAllocator" in str(e):
+                pass
+            else:
+                raise e
+
+    torch.accelerator.empty_cache = _patched_empty_cache
 from vllm.platforms.interface import Platform, PlatformEnum
 
 from tpu_inference import envs
--- a/tpu_inference/runner/speculative_decoding_manager.py
+++ b/tpu_inference/runner/speculative_decoding_manager.py
@@ -294,7 +294,9 @@ class SpeculativeDecodingManager:
             out_logits_indices.append(
                 _pad(logits_indices, padded_logits_length_dp_rank))
             out_target_logits_indices.append(
-                _pad(target_logits_indices, padded_logits_length_dp_rank))
+                _pad(
+                    target_logits_indices, padded_logits_length_dp_rank -
+                    padded_num_reqs_per_dp_rank))
             out_bonus_logits_indices.append(
                 _pad(bonus_logits_indices, padded_num_reqs_per_dp_rank))
             out_draft_lengths.append(
--- a/tpu_inference/runner/speculative_decoding_manager.py
+++ b/tpu_inference/runner/speculative_decoding_manager.py
@@ -294,9 +294,7 @@ class SpeculativeDecodingManager:
             out_logits_indices.append(
                 _pad(logits_indices, padded_logits_length_dp_rank))
             out_target_logits_indices.append(
-                _pad(
-                    target_logits_indices, padded_logits_length_dp_rank -
-                    padded_num_reqs_per_dp_rank))
+                _pad(target_logits_indices, padded_logits_length_dp_rank))
             out_bonus_logits_indices.append(
                 _pad(bonus_logits_indices, padded_num_reqs_per_dp_rank))
             out_draft_lengths.append(
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -335,17 +335,18 @@ def _jax_logprobs_materialize(
                 per_rank_output_size]
 
             padded_num_seqs_per_rank = bonus_token_ids.shape[0]
+            padded_tokens_per_seq = padded_tokens_length // padded_num_seqs_per_rank
             cur_rank_num_draft_tokens = spec_decode_metadata.draft_lengths_cpu[
                 rank * padded_num_seqs_per_rank:(rank + 1) *
                 padded_num_seqs_per_rank]
 
-            start_idx = 0
             req_indices = spec_decode_metadata.req_indices_dp[rank]
             for i, req_idx in enumerate(req_indices):
                 if req_idx >= num_reqs:
                     continue
 
                 seq_length = int(cur_rank_num_draft_tokens[i])
+                start_idx = i * padded_tokens_per_seq
                 end_idx = start_idx + seq_length
 
                 # Get draft logprobs for this sequence
@@ -375,8 +376,6 @@ def _jax_logprobs_materialize(
                 req_logprob_token_ids[req_idx] = valid_token_ids
                 req_logprobs[req_idx] = valid_logprobs
                 req_sampled_token_ranks[req_idx] = valid_ranks
-
-                start_idx = end_idx
 
         # Flatten the collected lists back to 2D/1D arrays
         flat_token_ids = []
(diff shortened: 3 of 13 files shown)


tpu_inference/runner/speculative_decoding_manager.py
```diff
@@ -294,7 +294,9 @@ def _pad(arr: np.ndarray, target_len: int) -> np.ndarray:
             out_logits_indices.append(
                 _pad(logits_indices, padded_logits_length_dp_rank))
             out_target_logits_indices.append(
-                _pad(target_logits_indices, padded_logits_length_dp_rank))
+                _pad(
```
gxd3: qq: is this change necessary?
(wonder if we could constraint the diff within `_jax_logprobs_materialize`, to reduce risk of subtle error)

tpu_inference/runner/utils.py
```diff
@@ -904,8 +905,8 @@ def host_extract_sampled_tokens(
         valid_sampled_token_ids = runner.rejection_sampler.parse_output(
             next_tokens, runner.input_batch.vocab_size,
             spec_decode_metadata.draft_lengths_cpu, num_reqs,
-            spec_decode_metadata.final_logits_indices.shape[0], runner.dp_size,
-            spec_decode_metadata.req_indices_dp)
+            spec_decode_metadata.target_logits_indices.shape[0],
```
gxd3: same as above:
 is this change necessary?
(wonder if we could constraint the diff within _jax_logprobs_materialize, to reduce risk of subtle error)

tpu_inference/platforms/tpu_platform.py
```diff
@@ -7,6 +7,21 @@
 import numpy
 import torch
 import vllm.envs as vllm_envs
+
+# Monkeypatch torch.accelerator.empty_cache to ignore device_allocator error on TPU.
+if hasattr(torch, "accelerator") and hasattr(torch.accelerator, "empty_cache"):
```
gxd3, line 12: supporting Logprobs + spec decoding don't require changes in tpu_inference/platforms/tpu_platform.py?

Lumosis: Yes, this one is for fixing a device_allocator error during e2e testing

tpu_inference/runner/tpu_runner.py
```diff
@@ -281,18 +286,140 @@ def _subtract_num_rejected_tokens_fn(seq_lens: jax.Array, positions: jax.Array,
...
     selected_token_ranks = np.asarray(
         jax.device_get(logprobs_tensors.selected_token_ranks))
 
-    if logits_indices_selector is not None:
-        log_token_ids = log_token_ids[logits_indices_selector]
-        logprobs_arr = logprobs_arr[logits_indices_selector]
-        selected_token_ranks = selected_token_ranks[logits_indices_selector]
+    if (spec_decode_metadata is not None
+            and np.sum(spec_decode_metadata.draft_lengths_cpu) > 0):
+        assert runner is not None
```
gxd3, line 309: could we extract this part as a util function, put it in spec_decode/jax/util.py?
(so that we don't make the runner.py even larger)

Also add some comment on what this block is trying to do?

tpu_inference/runner/tpu_runner.py
```diff
@@ -1512,6 +1660,74 @@ def _sample_from_logits(
 
         return model_runner_output
 
+    @jax.jit(static_argnums=(0, ))
```
gxd3: similar as above, move it in spec_decode/jax/util.py?

tpu_inference/runner/tpu_runner.py
```diff
@@ -1309,9 +1437,25 @@ def _sample_from_logits(
         with self.maybe_forbid_compile:
 
             if tpu_sampling_metadata.logprobs:
-                logits = processed_logits if self.model_config.logprobs_mode == "processed_logprobs" else logits
+                if spec_decode_metadata is not None:
+                    if (self.model_config.logprobs_mode == "processed_logprobs"
```
gxd3: perhaps move this to move it in spec_decode/jax/util.py too?

gxd3 (approved):

tpu_inference/spec_decode/jax/utils.py
```diff
@@ -97,3 +100,218 @@ def concat_last_sampled_tokens_and_draft_tokens(last_sampled_tokens,
                                                 draft_tokens):
     return jnp.concat([last_sampled_tokens[:, None], draft_tokens],
                       axis=1).reshape(-1)
+
+
+def filter_speculative_logprobs(
```
gxd3, line 106: if not too hard, perhaps can add unit test for this function as well, as it look non-trivial,
Thanks!

tpu_inference/spec_decode/jax/utils.py
```diff
@@ -97,3 +100,218 @@ def concat_last_sampled_tokens_and_draft_tokens(last_sampled_tokens,
...
+            cu_num_generated_tokens)
+
+
+@jax.jit
+def extend_logits_simple(
+    target_logits: jax.Array,
+    bonus_logits: jax.Array,
+) -> jax.Array:
+    """Concatenates target and bonus logits along the first axis."""
+    return jnp.concatenate(
```
gxd3: may need to wrap this within jax.shard_map to be compatible with attn-dp?

jrplatin (approved):

mrjunwan-lang (approved):

PR #2807, opened 2026-06-03
xla autotune: OFAT sweep over LIBTPU_INIT_ARGS on Buildkite

guowei-dev:
One-factor-at-a-time autotune that stands up vllm serve on a TPU pod, benchmarks it with benchmark_serving for each candidate XLA / libtpu flag, and ships per-trial results to Buildkite artifacts.

Under .buildkite/xla_autotune/:
  - autotuner.py            OFAT planner + per-trial runner.  Sharded via
                            --slice-{index,count}; --skip-candidates N
                            resumes mid-sweep (baselines always re-run,
                            trial indices stay stable).
  - vllm_test_framework.py  vllm serve + benchmark_serving runner with a
                            single MODELS registry (ModelSpec per model).
  - flags.txt               854 candidate --flag=value lines.
  - pipeline.yml            Buildkite pipeline (single shard; matrix-scalable).
  - README.md               Layout, sharding, trigger, knobs, JAX cache and
                            result-handoff mechanics.

.buildkite/scripts/run_xla_autotune_shard.sh is the host wrapper: per-build artifact dir, a watcher that uploads per-trial JSON + log bundles as they land, a git clone of the shared bench_serving harness, then autotuner.py.

Routing avoids hijacking pipeline_dev.yml: bootstrap_dev.sh now uploads the file named by the DEV_PIPELINE_FILE build env var (default pipeline_dev.yml, so the demo and every GitHub-check path are unchanged).  Launch the sweep on tpu-inference-dev with env DEV_PIPELINE_FILE=.buildkite/xla_autotune/pipeline.yml.

JAX persistent compile cache: the key hashes the whole LIBTPU_INIT_ARGS string, so each distinct flag set is one cold compile then warm on re-run; baselines (empty args) reuse the production cache.  Verified: a code/kernel change only re-keys the affected HLO modules, not the whole cache.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist


.buildkite/xla_autotune/vllm_test_framework.py
```diff
@@ -0,0 +1,371 @@
...
+        self._server: Optional[subprocess.Popen] = None
+        atexit.register(self._stop_server)
+
+        base = param.base_log_dir or os.path.join(os.getcwd(), "logs")
+        ts = datetime.now().strftime("%Y%m%d_%H%M%S")
+        folder = f"{param.tag}_EXP_{ts}" if param.tag else f"EXP_{ts}"
+        self.exp_dir = os.path.join(base, folder)
+        os.makedirs(self.exp_dir, exist_ok=True)
+
+        logging.basicConfig(
```
QiliangCui2023: **Bug: `logging.basicConfig()` is a no-op after the first trial — all subsequent trials log to trial 1's file**

Python's `logging.basicConfig()` does nothing if the root logger already has handlers configured (see [docs](https://docs.python.org/3/library/logging.html#logging.basicConfig)). Since `VLLMTestFramework.__init__` calls it once per trial, only the first trial's `FileHandler` (pointing at trial 1's `exp_dir`) is ever registered. Every subsequent trial calls `basicConfig` with a new `FileHandler` pointing at its own `exp_dir`, but that call is silently ignored.

`self.log = logging.getLogger(__name__)` returns the same shared logger singleton on every instantiation (loggers are global, keyed by name). So every `self.log.info(...)` in trials 2, 3, … N all route through trial 1's `FileHandler` — writing to `trial_1_dir/framework_main.log`. The per-trial log files for trials 2+ are never created.

**Fix:** don't use `basicConfig`; add handlers directly to a per-trial logger:
```python
self.log = logging.getLogger(f"{__name__}.{param.tag}")
self.log.setLevel(logging.INFO)
self.log.addHandler(logging.FileHandler(os.path.join(self.exp_dir, "framework_main.log")))
self.log.addHandler(logging.StreamHandler())
```

.buildkite/xla_autotune/vllm_test_framework.py
```diff
@@ -0,0 +1,371 @@
...
+                for line in iter(self._server.stdout.readline, ""):
+                    f.write(line)
+                    f.flush()
+
+        threading.Thread(target=tail, daemon=True).start()
+        self._wait_for_port()
+
+    def _wait_for_port(self) -> None:
+        t0 = time.time()
+        while True:
```
QiliangCui2023, line 242: **Bug: `_wait_for_port` has no timeout — hangs forever if the server stalls without crashing**

The loop exits only if `poll()` returns non-None (process died) or the port opens. If the server hangs in TPU init — process alive, port never opens — the loop spins indefinitely. `t0` is captured on line 227 but never checked.

On a misconfigured `LIBTPU_INIT_ARGS` flag this would block the entire shard until the Buildkite job-level timeout fires (~60–90 min), burning a queue slot with no actionable error message.

**Fix:** add an upper-bound check and raise with the server log tail on expiry:
```python
STARTUP_TIMEOUT_S = 600

def _wait_for_port(self) -> None:
    t0 = time.time()
    while True:
        if time.time() - t0 > STARTUP_TIMEOUT_S:
            raise RuntimeError(
                f"vllm serve did not open port {self.param.port} within {STARTUP_TIMEOUT_S}s"
            )
        if self._server.poll() is not None:
            raise RuntimeError(
                f"vllm serve exited during startup (rc={self._server.poll()})"
            )
        try:
            with socket.create_connection(("127.0.0.1", self.param.port), timeout=1):
                self.log.info("Server ready (%ds)", int(time.time() - t0))
                return
        except OSError:
            time.sleep(10)
```

.buildkite/xla_autotune/pipeline.yml
```diff
@@ -0,0 +1,49 @@
...
+      TPU_VERSION: "tpu7x"
+      AUTOTUNE_MODEL: "Qwen/Qwen3.5-397B-A17B-FP8"
+      AUTOTUNE_TARGET_METRIC: "total_token_throughput"
+      AUTOTUNE_BASELINE_RUNS: "2"
+      AUTOTUNE_FLAGS: ".buildkite/xla_autotune/flags.txt"
+    matrix: ["1"]
+    agents:
+      queue: tpu_v7x_8_queue
+    commands:
+      - .buildkite/scripts/run_xla_autotune_shard.sh {{matrix}} 1
```
QiliangCui2023: **Fragile two-place invariant: `matrix` size and the hardcoded `1` (slice_count) must be kept in sync manually**

The script signature is `run_xla_autotune_shard.sh <slice_index> <slice_count>`. Here `{{matrix}}` is `slice_index` and the trailing `1` is `slice_count`. The README documents scaling to N shards by changing `matrix` to `["1","2","3","4"]` and passing `4` as the second arg — but those are in two different places with nothing tying them together.

If you add shards to `matrix` but forget to update the argument, `slice_flags` raises immediately:
```
ValueError: slice_index 2 out of [1, 1]
```
Shards 2–N crash before doing any work. The label also has the hardcoded `1`:
```yaml
label: "tpu7x [xla-autotune] shard {{matrix}}/1"
```
so the Buildkite UI would show "shard 2/1", "shard 3/1" — visibly wrong but only after the jobs are already queued.

**Fix:** single-source the shard count via an env var so the matrix is the only thing to change:

```yaml
    env:
      AUTOTUNE_TOTAL_SHARDS: "1"   # change this and matrix together
      ...
    matrix: ["1"]
    commands:
      - .buildkite/scripts/run_xla_autotune_shard.sh {{matrix}} $AUTOTUNE_TOTAL_SHARDS
```

Or have the script derive `SLICE_COUNT` from the env var directly (`SLICE_COUNT="${AUTOTUNE_TOTAL_SHARDS:-1}"`) so the command argument is eliminated entirely.

.buildkite/xla_autotune/vllm_test_framework.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+
+            for i in range(passes):
+                is_measured = i == passes - 1
+                name = base if is_measured else f"{base}_warmup{i+1}"
+                fname = f"{name}_result_{datetime.now().strftime('%Y%m%d_%H%M%S')}.json"
+
+                cmd = ["python3", self.param.benchmark_script_path]
+                for k, v in self.param.benchmark_args.items():
+                    cmd.append(k if isinstance(v, bool) and v else f"{k}={v}")
```
QiliangCui2023: **Bug: boolean `False` values serialize as `--flag=False` instead of being omitted, causing argparse to error out**

There are three cases for `v`:

- **String** (e.g. `"--dataset-name": "random"`): `isinstance(v, bool)` is False → else branch → appends `--dataset-name=random`. ✓
- **`True`** (e.g. `"--ignore-eos": True`): `isinstance(v, bool) and v` is True → appends `--ignore-eos` with no value. ✓ Correct for `store_true` flags.
- **`False`** (e.g. `"--ignore-eos": False`): `isinstance(v, bool) and v` is False → else branch → appends `--ignore-eos=False`. ✗

Passing `--ignore-eos=False` to a `store_true` flag causes argparse to error:
```
argument --ignore-eos: ignored explicit argument 'False'
```
The benchmark subprocess exits nonzero and the trial fails.

This doesn't trigger with the current defaults (all booleans in `DEFAULT_BENCHMARK_ARGS` are `True`), but it's reachable: `benchmark_args` can be overridden via `apply_overrides` when `AUTOTUNE_CONFIG` is set. Passing `{"benchmark_args": {"--ignore-eos": False}}` hits the bug.

**Fix:** omit the flag entirely when `v` is `False`:
```python
for k, v in self.param.benchmark_args.items():
    if isinstance(v, bool):
        if v:
            cmd.append(k)        # --ignore-eos
        # False → omit entirely
    else:
        cmd.append(f"{k}={v}")  # --dataset-name=random
```

.buildkite/scripts/run_xla_autotune_shard.sh
```diff
@@ -0,0 +1,104 @@
...
+
+set +e
+.buildkite/scripts/run_in_docker.sh bash -c "
+  set -euo pipefail
+  cd /workspace/tpu_inference
+  # Shared benchmark_serving harness — same source as tests/e2e/benchmarking.
+  if [ ! -e bench_serving ]; then
+    git clone https://github.com/kimbochen/bench_serving.git
+  fi
+  echo \"bench_serving commit: \$(git -C bench_serving rev-parse HEAD)\"
```
QiliangCui2023, line 89: **bench_serving is cloned at HEAD with no commit pin — benchmark semantics can silently change between builds**

The `if [ ! -e bench_serving ]` guard means a fresh container always clones whatever is at HEAD of `kimbochen/bench_serving` at that moment. Two builds on different days can clone different commits and produce non-comparable throughput numbers — making the OFAT comparison invalid without any indication that the harness changed.

Pin to a specific commit so any change to the benchmark harness is a deliberate, reviewable edit in this repo:

```suggestion
  BENCH_SERVING_SHA="<pin-to-specific-commit-hash>"
  if [ ! -e bench_serving ]; then
    git clone https://github.com/kimbochen/bench_serving.git
  fi
  git -C bench_serving checkout "${BENCH_SERVING_SHA}"
  echo "bench_serving commit: $(git -C bench_serving rev-parse HEAD)"
```

Fill in the SHA with the current HEAD of `bench_serving` at merge time. The `checkout` also handles the stale-cache case (if the workdir somehow persists across builds) — the clone is a no-op but the checkout ensures the pinned version is used.

QiliangCui (approved):

.buildkite/xla_autotune/vllm_test_framework.py
guowei-dev: Added the bound in cbb8f643. One deliberate deviation: the default is **7200 s, not 600 s**, and it's overridable via `AUTOTUNE_SERVER_STARTUP_TIMEOUT_S`. The point of the OFAT sweep is that a new flag set cold-compiles the full graph on first run — on the 397B model that legitimately takes well over an hour (we've observed ~90 min compiles that then succeed), so a 10-min cap would false-fail real trials. 7200 s still catches a genuine hang (process alive, port never opens) without blocking on the job-level timeout. Verified locally — a hung server now raises `did not open port … within Ns`.

.buildkite/xla_autotune/vllm_test_framework.py
guowei-dev: Fixed exactly as suggested in cbb8f643: booleans emit the bare flag when `True` and are **omitted** when `False`; non-bools serialize as `k=v`. Verified locally — `{"--keep-me": True, "--drop-me": False, "--name": "random"}` produces `--keep-me --name=random` with no `--drop-me`.

.buildkite/xla_autotune/pipeline.yml
guowei-dev: Fixed in cbb8f643 via your second suggestion. Added `AUTOTUNE_TOTAL_SHARDS` (next to `matrix`, with a sync comment); the shard script reads it (`SLICE_COUNT="${2:-${AUTOTUNE_TOTAL_SHARDS:-1}}"`), so the command no longer carries a magic trailing number and the label dropped the misleading `/1` → now `shard {{matrix}}`. A positional 2nd arg still overrides for ad-hoc runs. README/Sharding updated to match.

.buildkite/scripts/run_xla_autotune_shard.sh
guowei-dev: Pinned in cbb8f643 to `ee867231de0b268e2810a6e31751b23cf5903fc5` (current `kimbochen/bench_serving` HEAD): clone-if-absent then `git checkout --quiet "${BENCH_SERVING_SHA}"`, with the resolved commit still echoed into the build log. Bumping the harness is now a deliberate, reviewable edit. Verified the expanded in-container script syntax-checks clean.

.buildkite/xla_autotune/vllm_test_framework.py
guowei-dev: Confirmed that every `framework_main.log` in the sweep was 0 B because `basicConfig` no-op'd after trial 1. Fixed in cbb8f643: each trial now uses its own logger keyed on the tag (`getLogger(f"{__name__}.{param.tag}")`) with fresh `FileHandler` + `StreamHandler` and `propagate=False`, so there's no cross-trial leakage. Verified locally — two trials write `MARKER_A`/`MARKER_B` only into their own files (previously all 0 B).

PR #2808, opened 2026-06-03
Set MPMD DP as the default for non Pathways environments

wenxindongwork:
# Description

The changes make the MPMD DP mode selection automatic. MPMD DP will be automatically enabled for online serving except for when running on Pathways or when attention dp is enabled. If user wants to use SPMD DP instead, they can manually set `TPU_MULTIPROCESS_DP=False`. 

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
     TPU_NAME: str | None = None
     TPU_WORKER_ID: str | None = None
     TPU_MULTIHOST_BACKEND: str = ""
-    TPU_MULTIPROCESS_DP: bool = False
+    TPU_MULTIPROCESS_DP: bool | None = None
     PREFILL_SLICES: str = ""
     DECODE_SLICES: str = ""
     SKIP_JAX_PRECOMPILE: bool = False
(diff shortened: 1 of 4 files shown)


tpu_inference/envs.py
```diff
@@ -211,9 +210,12 @@ def _get_int_list_env() -> list[int]:
     # Use vLLM-native multi-process data parallelism (one engine process per
     # DP rank, single load-balanced API endpoint) instead of tpu-inference's
     # single-process SPMD data parallelism. Each DP rank is pinned to a
-    # disjoint set of TPU chips. Dense (non-MoE) models only.
+    # disjoint set of TPU chips. Defaults to True, except on Pathways
+    # (JAX_PLATFORMS=proxy) which defaults to False.
     "TPU_MULTIPROCESS_DP":
-    env_bool("TPU_MULTIPROCESS_DP", default=False),
+    lambda: env_bool("TPU_MULTIPROCESS_DP",
```
lk-chen: does `env_bool` support `default=None`? I feel it will break lots of things if we flip here.

If None is acceptable, we can (in sharding.py) treat it as False if (pathway or dp_attn or offline) else True. And  raise only when user explicitly enable this flag in (pathway or dp_attn or offline) case

wenxindongwork: great idea!

lk-chen (approved):

sixiang-google (approved):

kyuyeunk (approved):

PR #2811, opened 2026-06-03
[DeepSeek V4] Add Lightening Indexer

inucool12:
# DeepSeek-V4 Lightning Indexer: TPU StreamIndex Top-K Kernel & Indexer

## Overview
  This design implements the **streamindex_topk** custom kernel and its PyTorch/vLLM integration module **VllmDeepseekV4Indexer** for the **DeepSeek-V4 Lightning Indexer**.
  
  The integration operates across three main components:
  1. **PyTorch Indexer Layer (`deepseek_v4_indexer.py`)**: A vLLM-compatible indexer module that intercepts forward passes, structures and buckets metadata (batch, token, block dimensions) to avoid compilation overhead, and performs JAX interop using TorchAX.
  2. **JAX Pallas Kernel (`streamindex_topk.py`)**: A hardware-optimized custom kernel on TPUs mapping sequences directly to hardware program grids (`pl.program_id`). It performs running Top-K computations fully on-chip in Vector Memory scratchpads.
  3. **Mocks & Tests (`test_streamindex_topk.py`)**: Validates shape constraints and verifies batched numerical correctness against a naive NumPy implementation.

### DIFF
--- a/tests/kernels/deepseek_v4/test_streamindex_topk.py
+++ b/tests/kernels/deepseek_v4/test_streamindex_topk.py
@@ -1,0 +1,313 @@
+import jax
+import jax.numpy as jnp
+import numpy as np
+import pytest
+
+try:
+  from google3.experimental.users.hwanginho.deepseek_v4.streamindex_topk.streamindex_topk import streamindex_topk
+except ImportError:
+  from tpu_inference.layers.vllm.custom_ops.streamindex_topk import streamindex_topk
+
+pytestmark = pytest.mark.filterwarnings("ignore::DeprecationWarning")
+
+
+@pytest.mark.parametrize(
+    "B, num_tokens, page_size, max_blocks, H_I, H_KV, D, k, compression_ratio,"
+    " bq_sz, bkv_p, seq_lens_list, cu_q_lens_list",
+    [
+        # Small standard case
+        (2, 6, 16, 8, 4, 1, 64, 512, 2, 32, 2, [4, 6], [0, 3, 6]),
+        # # Single token batch (Decode Phase)
+        (1, 1, 16, 4, 2, 1, 32, 512, 1, 16, 1, [10], [0, 1]),
+        # Large batch, multiple tokens per sequence (Prefill Phase)
+        (
+            4,
+            20,
+            32,
+            16,
+            8,
+            1,
+            128,
+            512,
+            4,
+            64,
+            4,
+            [10, 20, 30, 40],
+            [0, 5, 10, 15, 20],
+        ),
+        # Odd chunk sizes
+        (2, 5, 8, 6, 4, 1, 16, 512, 1, 16, 3, [8, 12], [0, 2, 5]),
+        # Single head for KV but multiple heads for Queries (MQA/GQA pattern)
+        (2, 10, 16, 8, 8, 1, 64, 512, 2, 32, 4, [16, 24], [0, 4, 10]),
+        # # High D, High K
+        (1, 8, 16, 4, 2, 1, 256, 512, 1, 32, 4, [60], [0, 8]),
+        # Mixed batch: sequence 0 has 1 token (decode), sequence 1 has 5 tokens
+        # (prefill)
+        (2, 6, 16, 8, 4, 1, 64, 512, 2, 32, 2, [4, 6], [0, 1, 6]),
+    ],
+)
+def test_streamindex_topk_shape(
+    B,
+    num_tokens,
+    page_size,
+    max_blocks,
+    H_I,
+    H_KV,
+    D,
+    k,
+    compression_ratio,
+    bq_sz,
+    bkv_p,
+    seq_lens_list,
+    cu_q_lens_list,
+):
+  """Tests the shape and basic execution bounds of streamindex_topk."""
+  print(f"\n{'-'*60}")
+  print(f"SHAPE TEST: B={B}, Tokens={num_tokens}, k={k}")
+  print(f"{'-'*60}")
+
+  query_projection = jnp.zeros((num_tokens, H_I, D), dtype=jnp.float32)
+  indexer_weights = jnp.zeros((num_tokens, H_I), dtype=jnp.float32)
+
+  num_pages = max_blocks * B
+  kv_cache = jnp.zeros((num_pages, page_size, H_KV, D), dtype=jnp.float32)
+
+  block_table = jnp.zeros((B, max_blocks), dtype=jnp.int32)
+  page_indices = block_table.flatten()
+  cu_q_lens = jnp.array(cu_q_lens_list, dtype=jnp.int32)
+
+  print("Inputs:")
+  print(f"  - query_projection: {query_projection.shape}")
+  print(f"  - kv_cache:         {kv_cache.shape}")
+  print(f"  - seq_lens:         {seq_lens_list}")
+  print(f"  - cu_q_lens:        {cu_q_lens_list}")
+
+  # Count number of decode sequences (T == 1) at the beginning of the batch
+  num_decodes = 0
+  while (
+      num_decodes < B
+      and (cu_q_lens_list[num_decodes + 1] - cu_q_lens_list[num_decodes]) == 1
+  ):
+    num_decodes += 1
+  distribution = (num_decodes, num_decodes, B)
+  expected_shape = (num_tokens, k)
+
+  seq_lens = jnp.array(seq_lens_list, dtype=jnp.int32)
+
+  out_shape_idxs = jax.eval_shape(
+      streamindex_topk,
+      query_projection,
+      indexer_weights,
+      kv_cache,
+      seq_lens,
+      page_indices,
+      cu_q_lens,
+      distribution,
+      k=k,
+      compression_ratio=compression_ratio,
+      num_kv_pages_per_block=bkv_p,
+      num_queries_per_block=bq_sz,
+  )
+
+  print("\nOutputs:")
+  print(f"  - Expected shape: {expected_shape}, dtype: int32")
+  print(
+      f"  - Actual shape:   {out_shape_idxs.shape}, dtype:"
+      f" {out_shape_idxs.dtype}"
+  )
+
+  assert out_shape_idxs.shape == expected_shape
+  assert out_shape_idxs.dtype == jnp.int32
+  print("Result: PASS")
+
+
+@pytest.mark.parametrize(
+    "B, T_list, S_list, page_size, H_I, H_KV, D, k, comp_ratio, bq_sz, bkv_p,"
+    " block_table_list",
+    [
+        # 1. Single sequence, highly fragmented block table
+        (1, [4], [16], 8, 4, 1, 16, 1024, 2, 8, 2, [[2, 0]]),
+        # 2. Batched Decode (T=1, B=2)
+        (2, [1, 1], [12, 16], 8, 4, 1, 16, 1024, 1, 8, 1, [[1, 3], [0, 2]]),
+        # 3. High GQA (8 Query Heads, 2 KV Heads)
+        (1, [6], [20], 4, 8, 1, 16, 1024, 1, 8, 2, [[4, 1, 3, 0, 2]]),
+        # 4. Multi-Batch Prefill with variable query/sequence lengths
+        (2, [5, 3], [16, 16], 8, 4, 1, 16, 1024, 1, 8, 4, [[3, 1], [2, 4]]),
+        # 5. Dummy sequences / Padding (B=3, but sequence 1 has 0 tokens)
+        (
+            3,
+            [2, 0, 3],
+            [16, 0, 8],
+            8,
+            4,
+            1,
+            16,
+            1024,
+            1,
+            8,
+            2,
+            [[1, 2], [0, 0], [4, 3]],
+        ),
+        # 6. Mixed batch: sequence 0 has 1 token, sequence 1 has 5 tokens
+        (2, [1, 5], [12, 16], 8, 4, 1, 16, 1024, 2, 8, 2, [[1, 3], [2, 0]]),
+    ],
+)
+def test_streamindex_topk_numerical_correctness(
+    B,
+    T_list,
+    S_list,
+    page_size,
+    H_I,
+    H_KV,
+    D,
+    k,
+    comp_ratio,
+    bq_sz,
+    bkv_p,
+    block_table_list,
+):
+  """Executes randomized input data against a naive NumPy ground truth."""
+  print(f"\n{'='*60}")
+  print(f"BATCHED NUMERICAL TEST (B={B})")
+  print(f"{'='*60}")
+
+  np.random.seed(42)
+
+  # 1. Setup Random Tensors
+  num_tokens = sum(T_list)
+  q = np.random.randn(num_tokens, H_I, D).astype(np.float32)
+  weights = np.random.uniform(-1.5, 1.5, size=(num_tokens, H_I)).astype(
+      np.float32
+  )
+
+  # Create a unified physical KV Cache pool large enough for all block indices
+  max_physical_page = np.max(block_table_list)
+  kv = np.random.randn(max_physical_page + 1, page_size, H_KV, D).astype(
+      np.float32
+  )
+
+  block_table = np.array(block_table_list, dtype=np.int32)
+  page_indices = block_table.flatten()
+  seq_lens = np.array(S_list, dtype=np.int32)
+  cu_q_lens = np.concatenate([[0], np.cumsum(T_list)]).astype(np.int32)
+
+  print("Configuration:")
+  print(f"  - Tokens total: {num_tokens}, Sequences(B): {B}, K: {k}")
+  print(f"  - GQA: {H_I} Query Heads -> {H_KV} KV Heads")
+  print(f"  - Compression Ratio: {comp_ratio}")
+
+  print("\nInputs Generated:")
+  print(f"  - Query Tensor: {q.shape}")
+  print(f"  - KV Cache:     {kv.shape}")
+  print(f"  - Block Table:  {block_table.tolist()}")
+
+  # =====================================================================
+  # 2. NAIVE COMPUTATION (The Ground Truth)
+  # =====================================================================
+  print("\n[1/3] Computing exact ground-truth using naive NumPy loops...")
+  expected_topk = np.full((num_tokens, k), -1, dtype=np.int32)
+
+  for b in range(B):
+    T_seq = T_list[b]
+    S_total = S_list[b]
+    q_start = cu_q_lens[b]
+
+    if T_seq == 0:
+      continue  # Skip dummy padded sequences
+
+    S_valid = S_total // comp_ratio
+
+    # Manually reconstruct this sequence's physical contiguous cache
+    seq_blocks = block_table[b]
+    seq_kv = np.concatenate([kv[p] for p in seq_blocks], axis=0)
+
+    # Ensure naive_scores has at least K columns for sorting bounds
+    naive_scores = np.full((T_seq, max(S_valid, k)), -np.inf, dtype=np.float32)
+
+    for t_idx in range(T_seq):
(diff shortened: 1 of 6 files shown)


tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
```diff
@@ -44,6 +49,60 @@
 logger = init_logger(__name__)
 
 
+from tpu_inference.layers.vllm.custom_ops.fused_indexer_topk import streamindex_chunked_topk
+
+class TpuDeepseekV4Indexer(DeepseekV4Indexer):
```
gxd3: let's name it `VllmDeepseekV4Indexer` (following existing naming convention),

and also name it to its own standalone file?

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
```diff
@@ -44,6 +49,60 @@
 logger = init_logger(__name__)
 
 
+from tpu_inference.layers.vllm.custom_ops.fused_indexer_topk import streamindex_chunked_topk
```
gxd3: also would be good to have some basic unit test for the new component to verify the new component is functionally correct.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
inucool12: Ok, I will change the class name. Just to clarify, do you mean I should move it to a separate file named vllm_deepseek_v4_indexer.py?

gxd3: "move it to a separate file named vllm_deepseek_v4_indexer.py?", yes, move it to a new file "deepseek_v4_indexer.py"

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
inucool12: I actually ran some functional tests to verify the code here: with [UnitTest](https://paste.googleplex.com/6306161686085632). Do you want me to add and run more test cases ?

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
inucool12: Done! I made the change to reflect your suggestion.

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+
+import jax
+import jax.numpy as jnp
+
+
+@jax.jit(static_argnames=["k", "compression_ratio", "c_s", "c_t"])
+def streamindex_chunked_topk(
+    query_projection: jax.Array,  # Shape: [B, T, H_I, D]
+    compressed_keys: jax.Array,   # Shape: [B, S, H_I, D]
```
gxd3: K is stored in the "paged" format,
that is, k's shape will be [num-pages, page-size, 1, head-dim]
we need a kv-cache as input: https://screenshot.googleplex.com/5WpVCRM5uu96o2L

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,117 @@
...
+    indexer_weights: jax.Array,   # Shape: [B, T, H_I]
+    k: int,
+    compression_ratio: int,
+    c_s: int = 8196,
```
gxd3: curious: what is the meaning of these 2 arguments?
and how the default value is chosen?

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,73 @@
...
+        compressed_kv_score: torch.Tensor,
+        indexer_weights: torch.Tensor,
+        positions: torch.Tensor,
+        rotary_emb: nn.Module,
+    ) -> torch.Tensor:
+        q, _ = self.wq_b(qr)
+        q = q.view(-1, self.n_head, self.head_dim)
+
+        k = self.compressor(compressed_kv_score, positions, rotary_emb)
+
```
gxd3: does this line run in TPU?
(as we didn't implement the compressor yet)

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,73 @@
...
+        from vllm.v1.attention.ops.deepseek_v4_ops import fused_indexer_q_rope_quant
+        q_quant, weights = fused_indexer_q_rope_quant(
```
gxd3: can TPU run `fused_indexer_q_rope_quant`? 
vLLM's fused_indexer_q_rope_quant() seem to call some Triton code

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
gxd3: Thank you!

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
gxd3: before merging, we do want to add some unit tests in https://github.com/vllm-project/tpu-inference/tree/main/tests/layers/vllm.

I left a few comment below, let's focus on those first, thanks!

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
gxd3: since the k is stored in a paged format,

we'll need metadata as input for the computation: https://screenshot.googleplex.com/9zgnKEe6UopLSsh, to tell: 1) how long is each seq; 2) for each seq what are its page indices in the kv-cache for the K etc

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
inucool12: You are right. I find I just mocked the compressor when I tested it.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
inucool12: Correct. Triton kernels don't map to TPU/XLA hardware. We should reimplement this  natively using Torch ops.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
inucool12: Okay, I will focus on addressing your comments first. Thanks!

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
gxd3: maybe comment this line out for now, then implement q-rope-quant in later PR?

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
gxd3: if so, maybe comment this line out?
(and add logging mentioning: VllmDeepseekV4Indexer.forward is not fully implemented)

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
inucool12: c_s and c_t are chunk sizes for the Context (S) and Query (T) dimensions. Computing the full [B, T, H_I, S] attention score matrix all at once would require massive HBM allocation (easily 50-100+ GB for long contexts), causing an instant out-of-memory error.

- Prevents HBM OOM: We run the top_k reduction inside the loop, meaning the only thing permanently written back to main HBM is the tiny [B, T, H_I, k] tensor.
- 
- Maximizes Speed: The defaults (8196 and 2048) are heuristically chosen so the intermediate computation fits entirely inside the TPU VMEM (data type size x c_s x c_t), avoiding slow memory spills into HBM.

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
inucool12: Got it. This implementation assumed a contiguous KV cache, not considering paged attention. I will update the code to accept the paged KV cache and its corresponding metadata. Thanks!

gxd3: Thanks Inho!
(In page attention, Q is flattened into 1 dimensional array, with different seq may have different number of 1 tokens in a forward step (i.e. some requests are in prefill phase, some requests are in decode phase), can take a look at https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/mla/v2/kernel.py for reference)

inucool12: Hi Guangxiang, 
I have updated the implementation to address your comments.

If the current commit looks good to you, I will proceed with writing the unit test under tests/layers/vllm to properly validate this behavior of this kernel.

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,178 @@
...
+def streamindex_chunked_topk(
+    query_projection: jax.Array,  # Shape: [num_tokens, H_I, D] (Flattened Q)
+    kv_cache: jax.Array,          # Shape: [num_pages, page_size, H_KV, D]
+    block_table: jax.Array,       # Shape: [B, max_blocks]
+    seq_lens: jax.Array,          # Shape: [B] (Number of tokens currently in context)
+    cu_seq_lens: jax.Array,       # Shape: [B+1] (Cumulative sum of query lengths)
+    indexer_weights: jax.Array,   # Shape: [num_tokens, H_I]
+    k: int,
+    compression_ratio: int,
+    num_tokens: int,
```
gxd3: this can be inferred from the shape of `query_projection`?

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,178 @@
...
+    c_s: int = 8192,
+    c_t: int = 2048,
+    page_size: int = 64,
```
gxd3: similarly, this can be inferred from the shape of `kv_cache`?

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
gxd3: Thanks Inho!
The interface of streamindex_chunked_topk() LGTM (i need more time to review and understand the details of the implementation), please go ahead to backfill unit test for it.
Thanks!

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
inucool12: Agreed. Added a TODO for the follow-up PR. PTAL

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
inucool12: Makes sense. I commented it out and left a TODO block explaining the missing pieces we need for the TPU implementation.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_attention.py
inucool12: Added unit test and its result.

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,180 @@
...
+            new_val_acc_global = jax.lax.dynamic_update_slice_in_dim(
+                val_acc_global, final_local_vals, t_idx * c_t, axis=0
+            )
+            new_idx_acc_global = jax.lax.dynamic_update_slice_in_dim(
+                idx_acc_global, final_local_ids, t_idx * c_t, axis=0
+            )
+
+            return new_val_acc_global, new_idx_acc_global
+
+        _, final_batch_indices = jax.lax.fori_loop(
```
gxd3: I think there is some issue:

each invocation of  `batch_step` should be handling 1 seq.

in line 160: why we use `num_t_chunks_global` ((num_tokens + c_t - 1) // c_t)? instead of ( cu_seq_lens[b_idx + 1] - cu_seq_lens[b_idx]) / chunk-size?

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
```diff
@@ -0,0 +1,180 @@
...
+    cu_seq_lens = cu_seq_lens.astype(jnp.int32)
+    block_table = block_table.astype(jnp.int32)
+
+    assert c_s % page_size == 0, (
+        f"c_s ({c_s}) must be a multiple of page_size ({page_size}) for bulk KV"
+        " gathering"
+    )
+    B = seq_lens.shape[0]
+    max_blocks = block_table.shape[1]
+    max_s_elements = max_blocks * page_size
```
gxd3: we don't need to loop over max-blocks too. For a seq, we know its' valid length in `seq_lens[seq_id]`, we only need to loop over its own valid chunks

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,186 @@
...
+                cu_seq_lens=csl_jax,
+                indexer_weights=w_jax,
+                k=self.topk_tokens,
+                compression_ratio=self.compress_ratio,
+                c_s=c_s,
+                c_t=c_t
+            )
+            return torch_view(topk_jax)[:n_tokens, :]
+
+        if not isinstance(attn_metadata_dict, dict):
```
gxd3: we probably don't need this if branch, which is for "dummy_run" in the GPU Model Runner in vLLM upstream, in tpu-inference's tpu-runner, we don't have such dummy run

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,186 @@
...
+
+                topk_decode = run_jax_kernel(
+                    q_slice=q_quant[:num_decode_tokens],
+                    w_slice=weights[:num_decode_tokens],
+                    bt_slice=decode_meta.block_table,
+                    sl_slice=decode_meta.seq_lens.to(q.device),
+                    csl_slice=decode_cu_seq_lens,
+                    bucket_tokens=16
+                )
+                self.topk_indices_buffer[:num_decode_tokens, :self.topk_tokens].copy_(topk_decode)
```
gxd3: we may not need to invoke `streamindex_chunked_topk` twice, once for prefill tokens segment, another one for the decode tokens segment.

we should be able to handle "raggness" in once call of `streamindex_chunked_topk` (that is, streamindex_chunked_topk can handle the case where different sequences can have different number of tokens in the current forward pass)

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,186 @@
...
+
+                    topk_prefill = run_jax_kernel(
+                        q_slice=q_quant[q_start:q_end],
+                        w_slice=weights[q_start:q_end],
+                        bt_slice=chunk.block_table,
+                        sl_slice=chunk_seq_lens,
+                        csl_slice=chunk_cu_seq_lens,
+                        bucket_tokens=256
+                    )
+                    self.topk_indices_buffer[q_start:q_end, :self.topk_tokens].copy_(topk_prefill)
```
gxd3: no need to copy over the result of `streamindex_chunked_topk` into " self.topk_indices_buffer"? (we probably don't need to use ` self.topk_indices_buffer` as all, just return the return value of streamindex_chunked_topk?)

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
inucool12: Fxed.
I've updated the logic inside batch_step so the jax.lax.fori_loop over query chunks now uses num_t_chunks_local = (T_seq + c_t - 1) // c_t, bounded correctly for each sequence's length.

tpu_inference/layers/vllm/custom_ops/fused_indexer_topk.py
inucool12: Thanks for point out this. I've switched the inner s_loop to use jax.lax.fori_loop bounded dynamically by num_s_chunks_local = (S_valid + c_s - 1) // c_s. It no longer loops up to max_blocks.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
inucool12: Yes! I've removed the separate topk_prefill and topk_decode runs. I am now extracting the decode logic and the prefill chunk loops into q_lens, seq_lens_list, and block_tables_list upfront, padding the ragged block table across the entire batch, and calling streamindex_chunked_topk exactly once.

tpu_inference/layers/vllm/custom_ops/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,182 @@
+# Copyright 2026 Google LLC
```
gxd3: let's put `VllmDeepseekV4Indexer` in "/tpu_inference/layers/vllm/custom_ops/experimental/deepseek_v4/"

and copy the indexer's attend kernel from g3 to "/tpu_inference/kernels/experimental/deepseek_v4/ ".

Also, we want to merge to "main" branch instead: https://screenshot.googleplex.com/4kYSEj24vkV2K2J

tpu_inference/layers/vllm/custom_ops/experimental/deepseek_v4/deepseek_v4_indexer.py
```diff
@@ -0,0 +1,252 @@
...
+        # whether the scales is a separate jax.Array, or packed into the same
+        # jax.Array as `cache_kv`.
+        # (The exact attribute name depends on vLLM's FP8 implementation, e.g., `kv_scale`)
+        # kv_scales_tensor = self.k_cache.kv_scale
+        # assert kv_scales_tensor.is_contiguous(), "Scales must be contiguous"
+        # kv_scales_jax = jax_view(kv_scales_tensor)
+
+        actual_num_tokens = hidden_states.shape[0]
+
+        c_s = getattr(self, "c_s", 8192)
```
gxd3: tpu_inference/layers/vllm/custom_ops/experimental/deepseek_v4/deepseek_v4_indexer.py need updates?

`c_s`, `c_t` etc no longer needed?

tests/kernels/deepseek_v4/test_streamindex_topk.py
```diff
@@ -0,0 +1,317 @@
+import jax
```
gxd3, line 1: move the test to tests/kernels/deepseek_v4/..?

tpu_inference/kernels/experimental/deepseek_v4/streamindex_topk.py
```diff
@@ -0,0 +1,806 @@
...
+
+  seq_len = seq_lens_ref[seq_idx]
+  kv_len = seq_len // compression_ratio
+
+  def compute_topk(
+      q,
+      kv,
+      *,
+      bkv_idx,
+      actual_bq_sz,
```
gxd3: nit: could we update this file after cl/933175705?
thanks!

PR #2829, opened 2026-06-05
Fix DeepSeek-V4 imports for vLLM DeepseekV4Attention refactor

guowei-dev:
vllm-project/vllm#44569 ("[DSV4] Refactor DeepseekV4Attention") merged the DeepseekV4MLA wrapper and DeepseekV4MLAAttention impl into a single DeepseekV4Attention class, and the AMD decoder now instantiates DeepseekV4ROCMAiterMLAAttention (a DeepseekV4Attention subclass) directly.

The removed symbols broke tpu_inference imports against vLLM main, which stalled the LKG nightly verify (no new vLLM commit could be promoted):

  ImportError: cannot import name 'DeepseekV4MLA' from
    'vllm.models.deepseek_v4.attention'
  ImportError: cannot import name 'DeepseekV4MLAAttention' from
    'vllm.models.deepseek_v4.attention'

Changes:
- kv_cache_manager: import/isinstance DeepseekV4Attention (the AttentionLayerBase the concrete MLA layer now subclasses) instead of DeepseekV4MLAAttention.
- deepseek_v4_attention: collapse the old wrapper/impl TPU stubs into one VllmDeepseekV4MLAAttention(DeepseekV4Attention) providing the abstract platform hooks; retarget the monkey-patch to DeepseekV4ROCMAiterMLAAttention.
- vllm_model_wrapper: update the swap-site comment to the new class name.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist


gxd3 (approved):

kyuyeunk (commented):
wait, i'm slightly confused: why is `VllmDeepseekV4MLAAttention` needed at all? it's not even doing anything right? also, this `patch_deepseek_v4_mla_cls` seems very hacky.

gxd3:
> wait, i'm slightly confused: why is `VllmDeepseekV4MLAAttention` needed at all? it's not even doing anything right? also, this `patch_deepseek_v4_mla_cls` seems very hacky.

we're in the process of implement it. Currently it's just a skeleton (Plan to merge the new components step by step instead of having a huge PR in one shot).

"patch_deepseek_v4_mla_cls" is due to vLLM upstream has not expose this as `CustomOp` or `PlugableLayer`, we need other way to inject our subclass.

guowei-dev:
Thanks for @gxd3 's help in reply.  Yep, and also I agreed the patch is hacky, but vLLM instantiates this class directly in amd/model.py with no register_oot hook, so rebinding the symbol is the only seam. It also predates this PR; I just retargeted it to the renamed class.

kyuyeunk:
https://github.com/vllm-project/vllm-ascend/blob/e58dd4558f9ff77850399a1f31339f1473cbed01/vllm_ascend/models/__init__.py#L4-L7

it looks like this is more proper approach that is taken by other plugins.

kyuyeunk:
anyway, this is not the scope of this pr, so will approve for now, but suggest looking into following other plugin's methods.

kyuyeunk (approved):

gxd3:
> anyway, this is not the scope of this pr, so will approve for now, but suggest looking into following other plugin's methods.

Thanks Kyuyeun for the pointer!

https://github.com/vllm-project/vllm-ascend/blob/e58dd4558f9ff77850399a1f31339f1473cbed01/vllm_ascend/models/deepseek_v4.py#L1075
seem to fork and modify the vLLM upstream's model.py (which has some code duplicate)

DS V4 has been constantly refactored by vLLM upstream. That said, we'll revisit different ways to inject TPU specific override as the bringup progress.

mrjunwan-lang (approved):

QiliangCui (approved):

PR #2830, opened 2026-06-05
[Multimodal] Add support for mm-encoder-tp-mode for ViT

kwang3939:
# Description

Earlier we noticed that the vision encoder is more performant using DP only based on roofline due to the small size. 
Add the support of `mm-encoder-tp-mode` referring to vllm: https://github.com/vllm-project/vllm/issues/22743.

* Directly using flag `--mm-encoder-tp-mode data` replies on if this feature is enabled for that specific model in vLLM. Thus, currently using `--additional_config='{"mm_encoder_tp_mode": "data"}'` instead.
* When set as `data`, not using TP in vision encoder and use that for DP, while when it is unset or set as `weights`, running as the initial TP setting. 
* This should work for both SPMD and MPMD. When using MPMD, using `DP=4, TP=2` with `data` would be using MPMD for DP=4 while SPMD for DP=2. 

# Tests

Tested with gemma4 with SPMD and MPMD. 

# Checklist


tpu_inference/layers/common/sharding.py
```diff
@@ -266,8 +287,12 @@ def from_vllm_config(cls,
             vllm_config.parallel_config.data_parallel_rank = 0
             vllm_config.parallel_config.data_parallel_size_local = 1
 
+        # To enable data-parallel vision encoding, pass:
+        # --additional_config='{"mm_encoder_tp_mode": "data"}'
```
lk-chen: I thought it's [--mm-encoder-tp-mode](https://docs.vllm.ai/en/v0.22.1/configuration/engine_args/#-mm-encoder-tp-mode)

tpu_inference/models/jax/gemma4_mm.py
```diff
@@ -727,6 +759,10 @@ def get_single_image_embedding(self, pixel_values: jax.Array,
             jnp.expand_dims(sort_idx, axis=-1),
             axis=1)
 
+        projected_vision_features = jax.lax.with_sharding_constraint(
```
lk-chen, line 763: We talked about this offline, but could you remind me what's the issue if we don't have this?

kwang3939: I was wondering for current code if explicit duplicate performs better or not. Just tested and found current should be good.

tpu_inference/layers/common/sharding.py
wenxindongwork, line 37: I think this could be cleaner if we introduce a `vit_model` mesh axis instead of overriding `VIT_BATCH` at runtime.

the batch axis names would looks like 

```
VIT_BATCH = ('data', 'attn_dp', 'attn_dp_expert', 'vit_model')

VIT_MODEL = 'model'

ATTN_HEAD = ('model', 'vit_model', 'expert', 'dcp')

...

```

in the default mode, vit_model=1, and everything follows the usual tp mode. In the non default mode, vit_model>1, and model=1, so we should use vit_model everywhere model is used, and VIT_MODEL and VIT_BATCH would behave as you intended.

kwang3939: Thanks for the comment! 
I think in non default mode, if we set 'vit_model' as tp_size and 'model' = 1 this would break the text model weight annotations: https://github.com/vllm-project/tpu-inference/blob/cb2fb67af91e81bfa4e62d3ecf0708acc058b7a7/tpu_inference/models/jax/gemma4.py#L77, which will cause no TP in text model part while we want to keep the DP and TP size unchanged for the text model. I also noticed this 'model' is used in many other jax path models. So this was why I didn't add it as 'vit_model'. What do you think? 

Alternatively, we may set both 'vit_model' and 'model' as tp_size. However, I think this would require TP² times as many devices, since mesh axes multiply.

kwang3939: Chat with @wenxindongwork, mark this comment as resolved for now. We'll further discuss the possible improvements later.

tpu_inference/models/jax/gemma4_mm.py
```diff
@@ -548,15 +580,15 @@ def __init__(self, vllm_config: VllmConfig, rng_key: jax.Array,
         rng = nnx.Rngs(rng_key)
         self.mesh = mesh
 
-        from tpu_inference.models.jax.gemma4 import Gemma4Model
+        vllm_config.sharding_config.apply_vision_sharding()
```
lk-chen, line 584: not sure why runtime overriding is necessary.. is it an option to
```
self.vit_batch_axis_name = "batch" if ... else ("batch", "model")
self.vit_model_axis_name = "model" if ... else None
...
# pass self.vit_batch_axis_name to get_mesh_shape_product or attention_interface
```
inside gemma4_mm.py? Or put into `SupportsMultiModal` interface like [vllm does](https://github.com/vllm-project/vllm/blob/e0871ad2259768add6dc43e2972bd364d0d13086/vllm/model_executor/models/interfaces.py#L113-L117)

wenxindongwork (approved):

lk-chen:
@copilot could you show the per-file owner? (don't explicitly at each user)

tpu_inference/models/jax/gemma4_mm.py
kwang3939: Thanks for the comment! I switched to not use `self.vit_batch_axis` was trying to avoid passing it layer by layer in the jax model code, which I found setting is as `VIT_BATCH` could make it cleaner.

lk-chen: You mean in `sharded_flash_attention` right? which is anyway passing layer by layer 

Will approve and let you decide

lk-chen (approved):

tpu_inference/models/jax/gemma4_mm.py
kwang3939: I mainly refer to passing them from `Gemma4ForConditionalGeneration` to `Gemma4VisionModel` all the way down to `Gemma4VisionFlashAttention`. For `sharded_flash_attention`, I found we indeed need to pass it from the model.

PR #2833, opened 2026-06-05
Flush pending async TPU outputs on empty schedules

YJYJLee:
## Summary
When TPU async scheduling reaches a step with no scheduled tokens, the runner currently returns an empty output immediately. If a previous async TPU result is still pending, that result is never materialized and vLLM can keep a request in RUNNING with no schedulable work, causing repeated `Should not schedule a request that does nothing!` warnings.

This drains the pending async TPU result before returning from a no-work step and clears the pending async state. That lets the scheduler consume the real sampled tokens and finish or advance the request instead of spinning on empty schedules.

## Test Plan
- `python -m py_compile tpu_inference/runner/tpu_runner.py tests/runner/test_tpu_runner.py`
- `git diff --check`
- `python -m pytest tests/runner/test_tpu_runner.py::test_flush_prev_async_results_returns_pending_output_and_clears_state -q` could not be run locally because pytest is not installed in this checkout's default Python environment.

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -1017,6 +1017,26 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
                 req_state.output_token_ids.pop()
             req_state.output_token_ids.extend(sampled_ids)
 
+        return valid_sampled_token_ids
+
+    def _flush_prev_async_results(
+        self,
+        kv_connector_output: Optional[KVConnectorOutput] = None,
+    ) -> ModelRunnerOutput:
+        assert self._pre_async_results is not None
+        pre_req_ids = self._pre_async_results.req_ids
+        valid_sampled_token_ids = self._modify_prev_results()
+        self._pre_async_results = None
+        return ModelRunnerOutput(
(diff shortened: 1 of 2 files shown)


kyuyeunk:
@yuyanpeng-google can you take a look since you worked on async scheduler?

yuyanpeng-google:
Can you point me to the code where vLLM schedules a step without tokens, and explain why it needs the ModelOutput from the previous step?

tpu_inference/runner/tpu_runner.py
```diff
@@ -1061,6 +1017,26 @@ def _modify_prev_results(self):
...
+        self,
+        kv_connector_output: Optional[KVConnectorOutput] = None,
+    ) -> ModelRunnerOutput:
+        assert self._pre_async_results is not None
+        pre_req_ids = self._pre_async_results.req_ids
+        valid_sampled_token_ids = self._modify_prev_results()
+        self._pre_async_results = None
+        return ModelRunnerOutput(
+            req_ids=pre_req_ids,
+            req_id_to_index=self.input_batch.req_id_to_index.copy(),
```
yuyanpeng-google, line 1032: why we need a copy here?

PR #2834, opened 2026-06-06
Add Support for Kernel Tuner Specific Flags

patrickji2014:
Existing tuning infra doesn't support kernel customization via using env vars. This change make adding kernel specific env vars easier by flag naming convention. Adding kernel specific flag should follow this process:
1. Add the kernel specific flags in the kernel_tuning.py with kernel_name as a prefix.
   For example:  your_kernel_name_flag_name
2. Using the flag when run the kernel tuner:
  2.1 When invoke the kernel pipeline through BuildKite UI, append KERNEL_TUNING_ to the kernel name. For example: KERNEL_TUNING_YOUR_KERNEL_NAME_FLAG_NAME
  2.2 When invoke the kernel pipeline through local run, just use the original flag name, for example, your_kernel_name_flag_name

This naming convention helps avoid naming conflict and reduce code change when authoring the kernel tuner.

# Tests

Test in the BuildKite https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/117/canvas?sid=019ef792-2855-4eb3-ac98-8a7a45c0b7c0&open=false
built with the following env vars
KERNEL_TUNING_CASE_SET_DESC="test_tuning_infra_change"
KERNEL_TUNING_CASE_SET_ID="test_tuning_infra_change_0"
KERNEL_TUNING_KERNEL_TUNER_NAME="mla_kernel_tuner"
KERNEL_TUNING_RUN_ID="001"
KERNEL_TUNING_TPU_CORES="2"
KERNEL_TUNING_TPU_VERSION="tpu7x"

# Checklist


.buildkite/scripts/run_in_docker.sh
```diff
@@ -59,19 +59,14 @@ ENV_VARS=(
...
-  -e KERNEL_TUNING_TPU_VERSION="${KERNEL_TUNING_TPU_VERSION:-}"
-  -e KERNEL_TUNING_TPU_CORES="${KERNEL_TUNING_TPU_CORES:-}"
-  -e KERNEL_TUNING_JOB_PRIORITY="${PRIORITY_KERNEL_TUNING:--10}"
-  -e KERNEL_TUNING_MAX_EXECUTION_MINUTES="${KERNEL_TUNING_MAX_EXECUTION_MINUTES:-20}"
   -e HOST_NAME="${HOST_NAME:-}"
   -e GCS_BUCKET="${GCS_BUCKET:-}"
 )
 
+# shellcheck disable=SC1091
+source "$SCRIPT_DIR/kernel_tuning_envs.sh"
```
QiliangCui2023: `$SCRIPT_DIR` is used here, but it is not defined until line 82 (`SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)`). With `set -euo pipefail` set at the top of this file, `-u` turns this use-before-definition into an immediate `SCRIPT_DIR: unbound variable` abort.

Because this `source` is unconditional and `run_in_docker.sh` is the shared entrypoint for every suite (accuracy, throughput, benchmarks, kernel tuning), it breaks **all** pipelines, not only kernel tuning.

Fix: move the `SCRIPT_DIR=...` definition (currently line 82) above this line.

.buildkite/scripts/kernel_tuning_envs.sh
```diff
@@ -0,0 +1,68 @@
...
+)
+
+# Validation logic
+for entry in "${KERNEL_TUNING_ENV_VARS[@]}"; do
+    # Extract the variable name from the '-e NAME=VALUE' string
+    # This pattern removes everything before '=' and the '-e ' prefix
+    var_name=$(echo "$entry" | sed 's/^-e //;s/=.*//')
+  
+    # Check if the variable is defined and not empty
+    if [[ -z "${!var_name}" ]]; then
```
QiliangCui2023: This validation loop iterates over `KERNEL_TUNING_ENV_VARS`, but each `-e NAME=VALUE` entry is **two** array elements: `-e` and `NAME=VALUE`. On the bare `-e` element, `var_name` becomes empty (`echo "-e"` is consumed as an `echo` option and prints nothing), so `${!var_name}` is an indirect expansion of an empty name and bash raises `: invalid variable name`. Under the inherited `set -euo pipefail`, the script exits 1 on the very first element — even when all required variables are set, so the happy path never passes.

Fix: drive the loop off an explicit list of variable names rather than parsing the `-e`-interleaved docker-arg array, e.g.

```bash
REQUIRED_VARS=(KERNEL_TUNING_CASE_SET_ID KERNEL_TUNING_RUN_ID KERNEL_TUNING_KERNEL_NAME KERNEL_TUNING_TPU_VERSION KERNEL_TUNING_TPU_CORES)
for var_name in "${REQUIRED_VARS[@]}"; do
  if [[ -z "${!var_name:-}" ]]; then ... ; fi
done
```

.buildkite/pipeline_kernel_tuning.yml
```diff
@@ -48,25 +48,33 @@ steps:
...
+            "--tpu_version=${TPU_VERSION}"
+          )
+          
+          # Dynamically gather all KERNEL_TUNING_ variables
+          for key in "${!KERNEL_TUNING_@}"; do
+            # ${!key} gets the value of the variable name stored in $key (indirect expansion)
+            value="${!key}"
+            
+            # Remove "KERNEL_TUNING_" prefix and convert to lowercase
+            arg_name=$(echo "${key#KERNEL_TUNING_}" | tr "[:upper:]" "[:lower:]")
```
QiliangCui2023: The convention here strips `KERNEL_TUNING_` and lowercases, so `KERNEL_TUNING_KERNEL_NAME` maps to `--kernel_name`. But the runner defines the flag as `--kernel_tuner_name` (`tools/kernel/tuner/v1/kernel_tuner_runner.py:33`) and there is no `--kernel_name`, so absl exits FATAL with `Unknown command line flag 'kernel_name'`. The kernel name never reaches the runner — a regression vs. the previous explicit `--kernel_tuner_name=...`.

Fix: rename the env var to `KERNEL_TUNING_KERNEL_TUNER_NAME` (and update `kernel_tuning_envs.sh` and the README), or special-case this single mapping.

.buildkite/pipeline_kernel_tuning.yml
```diff
@@ -48,25 +48,33 @@ steps:
...
+        .buildkite/scripts/run_in_docker.sh bash -c '
+          set -e
+          
+          # Combine pip installs to save time
+          pip install --upgrade google-cloud-spanner google-api-core google-auth absl-py
+          
+          # Start with the arguments that do not follow the KERNEL_TUNING_ pattern
+          PYTHON_ARGS=( 
+            "--generate_buildkite_pipeline=True"
+            "--tpu_version=${TPU_VERSION}"
```
QiliangCui2023: `--tpu_version` is also produced by the dynamic loop below, from the required `KERNEL_TUNING_TPU_VERSION` var, so it lands on the command line twice — once from `$TPU_VERSION` here and once from `$KERNEL_TUNING_TPU_VERSION`. absl does not error on a duplicate flag; it keeps the last occurrence, so the loop's `KERNEL_TUNING_TPU_VERSION` silently overrides this explicit value. If the two ever differ, the value set here is silently ignored.

Fix: use a single source — drop this explicit line and rely on the loop, or exclude `KERNEL_TUNING_TPU_VERSION` from the propagated set.

.buildkite/scripts/kernel_tuning_envs.sh
```diff
@@ -0,0 +1,68 @@
...
+    -e KERNEL_TUNING_MAX_EXECUTION_MINUTES="${KERNEL_TUNING_MAX_EXECUTION_MINUTES:-20}"
+)
+
+# Creates a string of just the variable names (e.g., "KERNEL_TUNING_CASE_SET_ID KERNEL_TUNING_RUN_ID")
+EXISTING_VARS=$(echo "${KERNEL_TUNING_ENV_VARS[@]}" | grep -oP 'KERNEL_TUNING_[A-Z_]+')
+
+# Gather and append only if not already present, these are kernel specific environment variables
+for var in "${!KERNEL_TUNING_@}"; do
+    # Skip if the variable name is already in the EXISTING_VARS list
+    if [[ "$EXISTING_VARS" =~ $var ]]; then
```
QiliangCui2023: `[[ "$EXISTING_VARS" =~ $var ]]` is an unanchored regex/substring match against the space-joined list of already-known vars. A kernel-specific var whose name is a substring of an existing one — e.g. `KERNEL_TUNING_TPU` against `KERNEL_TUNING_TPU_VERSION` / `KERNEL_TUNING_TPU_CORES` — matches and is silently skipped, so it is never propagated. That defeats the purpose of this pass-through loop.

Fix: match whole names — iterate the known-names array and compare with `==`, or anchor the pattern, e.g. `=~ (^|[[:space:]])$var([[:space:]]|$)`.

.buildkite/scripts/run_in_docker.sh
QiliangCui2023: Separate concern from the `$SCRIPT_DIR` ordering issue above: this `source` is **unconditional**, and `kernel_tuning_envs.sh` hard-requires `KERNEL_TUNING_CASE_SET_ID`, `KERNEL_TUNING_RUN_ID`, `KERNEL_TUNING_KERNEL_NAME`, `KERNEL_TUNING_TPU_VERSION`, and `KERNEL_TUNING_TPU_CORES` — it `exit 1`s if any of them is empty (see kernel_tuning_envs.sh:30-45).

But `run_in_docker.sh` is the shared entrypoint for every suite (accuracy, throughput, benchmarks, …), and none of those set the `KERNEL_TUNING_*` vars. So even after the `$SCRIPT_DIR` ordering and the `-e`-array iteration issues are addressed, every non-kernel-tuning job will fail this validation with `Required environment variable KERNEL_TUNING_CASE_SET_ID is not defined or is empty` and exit 1. The hard requirement on these vars should not apply to callers that aren't the kernel-tuning pipeline.

patrickji2014: fixed.

.buildkite/pipeline_kernel_tuning.yml
patrickji2014: change the flag name to KERNEL_TUNING_KERNEL_TUNER_NAME which better represent the flag intend. update readme and kernel_tuning_envs.sh also.

.buildkite/scripts/run_in_docker.sh
patrickji2014: fixed.

.buildkite/scripts/kernel_tuning_envs.sh
patrickji2014: fix by iterating through the list and skip the KERNEL_TUNING_ENV_VARS since it also starts with KERNEL_TUNING_.

.buildkite/pipeline_kernel_tuning.yml
patrickji2014: fix by dropping this explicit

.buildkite/scripts/kernel_tuning_envs.sh
patrickji2014: fix by skipping '-e' when iterating through the list

.buildkite/scripts/run_in_docker.sh
```diff
@@ -90,6 +81,12 @@ SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
 source "$SCRIPT_DIR/setup_docker_env.sh"
 setup_environment $IMAGE_NAME
 
+# shellcheck disable=SC1091
+if [[ "$BUILDKITE_PIPELINE_NAME" =~ (kernel-tuning|kernel-autotune) ]]; then
```
QiliangCui2023: `$BUILDKITE_PIPELINE_NAME` is referenced unguarded here, but this file runs under `set -euo pipefail` (line 21). Buildkite always sets this variable, so CI is unaffected — but any invocation where it is unset (e.g. running `run_in_docker.sh` locally / outside a Buildkite agent) now aborts with `BUILDKITE_PIPELINE_NAME: unbound variable` before reaching the test command. That is a regression introduced by gating the kernel-tuning `source` on the pipeline name.

Fix: default it — `if [[ "${BUILDKITE_PIPELINE_NAME:-}" =~ (kernel-tuning|kernel-autotune) ]]; then`.

.buildkite/scripts/kernel_tuning_envs.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    # Check if the variable is defined and not empty
+    if [[ -z "${!var_name}" ]]; then
+        echo "Error: Required environment variable $var_name is not defined or is empty."
+        exit 1
+    fi
+          if [ -z "${KERNEL_TUNING_RUN_ID:-}" ] || [ -z "${KERNEL_TUNING_CASE_SET_ID:-}" ]; then
+            echo "Error: KERNEL_TUNING_RUN_ID or KERNEL_TUNING_CASE_SET_ID not set"
+            exit 1
+          fi
```
QiliangCui2023: This nested block is dead/redundant. The loop above already validates every entry of `KERNEL_TUNING_ENV_VARS`, and both `KERNEL_TUNING_RUN_ID` and `KERNEL_TUNING_CASE_SET_ID` are members of that array (lines 23-24), so they are already checked by the generic check above. This re-checks the same two vars on every iteration — and the 10-space indent gives it away as leftover paste from the old inline pipeline command. It can be deleted:

```suggestion
```

QiliangCui (approved):

.buildkite/scripts/run_in_docker.sh
patrickji2014: this should be fine as we never run run_in_docker.sh locally. but for completeness, i adopt the Fix. thanks.

.buildkite/scripts/kernel_tuning_envs.sh
patrickji2014: fixed

PR #2838, opened 2026-06-07
Get the right rope_scaling values from config. 

lc5211:
# Description
I've observed the logprobs regression from transformers=5.5.0 comparing to transformers=4.57.1 for qwen3.py. The changes fixes this regression.

# Tests

Tested with my RL framework(Tunix) and confirmed the regression is gone after the fix.

# Checklist

Signed-off-by: lc5211 (linchai@google.com)


tpu_inference/models/jax/qwen3.py
```diff
@@ -62,7 +62,20 @@ def __init__(self,
         self.num_heads = config.num_attention_heads
         self.num_kv_heads = config.num_key_value_heads
         self.rope_theta = config.rope_parameters["rope_theta"]
-        self.rope_scaling = getattr(config, "rope_scaling", None)
+        rope_scaling = getattr(config, "rope_parameters", None) or getattr(
```
lk-chen: Thanks, let's unify the logic into `get_rope()` simlar to what vLLM [does](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/qwen3.py#L115-L118)

lc5211: Good suggestion. Updated accordingly.

lk-chen (approved):

docker/Dockerfile
```diff
@@ -61,7 +61,7 @@ RUN git clone $VLLM_REPO . && \
         git checkout $VLLM_COMMIT_HASH; \
     fi
 
-RUN --mount=type=cache,target=/root/.cache/pip pip install --extra-index-url https://download.pytorch.org/whl/cpu -r requirements/tpu.txt --retries 3
+RUN --mount=type=cache,target=/root/.cache/pip pip install --extra-index-url https://download.pytorch.org/whl/cpu torch -r requirements/tpu.txt --retries 3
```
QiliangCui: can you put this inside requirements/tpu.txt?

QiliangCui (dismissed):

docker/Dockerfile
lc5211: Do you mean the requirements/tpu.txt in vllm repo?

kwang3939 (approved):

PR #2842, opened 2026-06-08
Customized a experimental RPA kernel for DCP usecase

weiyu0824:
# Description
Add a experimental RPA kernel for Context Parallelism. 

- Dynamic CP Rank/Static CP Size: Add this 2 parameters for DCP usecase, each dcp rank would only store partial kv tokens to caches.

- Two-Phase Attention Control: Added skip_cache_attn and skip_current_attn variables to enable the two-phase attention calculation required by DCP (followed upstream vllm).

- Internal Local Cache Length Calculation: In RPAv3, kv_len - q_len represent global_cache_len, and the kernel need to calculate its local_cache_len internally. 

- LSE Calculation: Added a parameter to return the lse value.

- KV Cache Update Logic: Previously, a block KV (bkv) and an offset were sufficient to update all values to the cache. Since a bkv can now contain a mix of partial cache and full current tokens, a new fn `udpate_kv_cache_partial` is added, and I renamed previous `update_kv_cache` to `update_kv_cache_full`. 
  - Note: in `udpate_kv_cache_partial` `src_start_base`  represent the index of the first new KV token within the bkv buffer. The offset continues to represent the global index of the first new KV token in this buffer.

# Tests
### Correctness: 
Verify this fix won't hurt existing RPA function by running 
```
NEW_MODEL_DESIGN=1 python3 examples/offline_inference.py --max-model-len=9216 --max-num-batched-tokens=256 --max-num-seqs=128 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --decode-context-parallel-size 1 --model Qwen/Qwen3-235B-A22B-Instruct-2507
```

# Checklist


wenxindongwork (dismissed):

tpu_inference/kernels/experimental/context_parallelism_rpa/kernel.py
```diff
@@ -0,0 +1,2208 @@
+"""TPU-Friendly Ragged Paged Attention kernel.
```
wenxindongwork: why not just update the v3 kernel?

weiyu0824: I think we need a lots more experiments to make sure it doesn't degrade other use case.

wenxindongwork: would it be possible to run some benchmarks on non DCP inputs and verify that performance does not degrade? 

the default logic has remained the same, but you introduced more scalar arithmetic, is that understanding correct?

weiyu0824: ok.

kyuyeunk: > I think we need a lots more experiments to make sure it doesn't degrade other use case.

you don't have to do any guessing game if you have a low level compiled code that shows introduction of this feature have zero impact on existing code.

weiyu0824: I integrate the CP into existing RPA, and seems like this change would change the LLO in my local experiments.  Will figure out the llo diff and run some perf benchmark on the kernel.

wenxindongwork (approved):

kyuyeunk (approved):

PR #2846, opened 2026-06-08
[MTP] Add gemma4 MTP e2e test

Lumosis:
# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,5 +21,4 @@ gcsfs==2026.1.0
 hypothesis==6.151.9
 sortedcontainers==2.4.0
 # Upstream pins to 5.5.3 https://github.com/vllm-project/vllm/blob/de3fe8dc62f3d77eb8dab8125ca90436f606bccb/requirements/test/nightly-torch.txt#L32
-# TODO: unpin once transformers release with the fix https://github.com/huggingface/transformers/pull/45680
-transformers==5.5.3
+transformers>=5.8.0
--- a/tests/e2e/test_speculative_decoding.py
+++ b/tests/e2e/test_speculative_decoding.py
@@ -688,3 +688,127 @@ def test_eagle3_logprobs_correctness(monkeypatch: pytest.MonkeyPatch, ):
                              max_num_seqs=2,
                              async_scheduling=True,
                              enable_dp_attention=False)
+
+
+@pytest.fixture(scope="module")
+def gemma4_mtp_baseline():
+    '''
+    Compute the gemma4 mtp reference prompts, baseline outputs and the LLM extra_kwargs
+    once and share them across all parameterized test_gemma4_mtp_correctness cases.
+    '''
+    model_name = "google/gemma-4-E2B-it"
+    sampling_config = SamplingParams(temperature=0,
+                                     max_tokens=32,
+                                     ignore_eos=True,
+                                     repetition_penalty=1,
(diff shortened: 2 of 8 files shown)


gxd3 (approved):

requirements.txt
```diff
@@ -21,5 +21,4 @@ gcsfs==2026.1.0
 hypothesis==6.151.9
 sortedcontainers==2.4.0
 # Upstream pins to 5.5.3 https://github.com/vllm-project/vllm/blob/de3fe8dc62f3d77eb8dab8125ca90436f606bccb/requirements/test/nightly-torch.txt#L32
-# TODO: unpin once transformers release with the fix https://github.com/huggingface/transformers/pull/45680
-transformers==5.5.3
+transformers
```
QiliangCui: do we still need this considering https://github.com/vllm-project/vllm/blob/main/requirements/common.txt#L10?

tests/e2e/test_speculative_decoding.py
```diff
@@ -688,3 +688,127 @@ def test_eagle3_logprobs_correctness(monkeypatch: pytest.MonkeyPatch, ):
...
+    Compare the outputs of an original LLM and a speculative LLM;
+    they should be the same when using Gemma-4 MTP speculative decoding.
+    '''
+    model_name = "google/gemma-4-E2B-it"
+    monkeypatch.setenv("MODEL_IMPL_TYPE", "flax_nnx")
+    monkeypatch.setenv("DRAFT_MODEL_IMPL_TYPE", "flax_nnx")
+
+    speculative_config = {
+        "model": "google/gemma-4-E2B-it-assistant",
+        "num_speculative_tokens": 4,
```
QiliangCui, line 755: in line 806, i see you also have ` "method": "mtp",`. 

missed here?

tests/e2e/test_speculative_decoding.py
```diff
@@ -688,3 +688,127 @@ def test_eagle3_logprobs_correctness(monkeypatch: pytest.MonkeyPatch, ):
...
+                                     max_tokens=32,
+                                     ignore_eos=True,
+                                     repetition_penalty=1,
+                                     frequency_penalty=0,
+                                     presence_penalty=0,
+                                     min_p=0,
+                                     logprobs=None)
+    extra_kwargs = {
+        "seed": 42,
+        "max_model_len": 2048,
```
QiliangCui: nit: from get_eagle3_test_prompts() i see the prompt is actually very small... do we need max_model_len: 2048? if shorter, does it make the test faster?

requirements.txt
Lumosis: Thanks! Updated. `5.8.0` is the oldest version that supports gemma4 mtp

tests/e2e/test_speculative_decoding.py
Lumosis: Gemma4 MTP is kinda special. We don't need to specifically set method to mtp. Ref: https://docs.vllm.ai/projects/recipes/en/latest/Google/Gemma4.html#online-serving

tests/e2e/test_speculative_decoding.py
Lumosis: Nice catch! Thanks

QiliangCui (approved):

PR #2858, opened 2026-06-09
[DSV4] Add KV compressor module

a1yssan13:
Bringup of some very basic JAX implementation for DeepSeek-V4 KV Compressor.


tpu_inference/kernels/experimental/deepseek_v4/compressor.py
```diff
@@ -0,0 +1,153 @@
...
+        score=score,
+        ape=ape,
+        positions=positions,
+        state_cache=state_cache,
+        slot_mapping=slot_mapping,
+        compress_ratio=compress_ratio,
+    )
+
+
+def compressor_forward(
```
gxd3, line 67: nit: we probably need a class similar to `DeepseekCompressor` in vLLM repo (and placed in "tpu_inference/layers/vllm/custom_ops/experimental/deepseek_v4"), to hold the projection weights param etc (similar example: https://screenshot.googleplex.com/6BeGf6RdJS5uUxC for DS V3), whose forward() call into methods here :)

Don't need to do that in this PR, just FYI.

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
```diff
@@ -0,0 +1,153 @@
...
+    ape: jax.Array,                 # [compress_ratio, coff * head_dim] fp32
+    norm_weight: jax.Array,         # [head_dim] fp32
+    cos_sin_cache: jax.Array,       # [max_pos, rope_head_dim] fp32
+    positions: jax.Array,           # [num_tokens] int
+    slot_mapping: jax.Array,        # [num_tokens] int
+    block_table: jax.Array,         # [num_reqs, max_blocks] int
+    token_to_req_indices: jax.Array,  # [num_tokens] int
+    kv_slot_mapping: jax.Array,     # [num_tokens] int
+    state_cache: jax.Array,         # [num_blocks, block_size, 2*coff*head_dim] fp32
+    nope_cache: jax.Array,          # [kv_blocks, kv_block_size, nope_dim] fp8
```
gxd3: is this possible to combine `nope_cache`, `rope_cache`, `scale_cache` into one cache (a jax.Array) (bit pack at the most minor dim), to make the cache allocation more consistent with vLLM upstream? This would make the integration easier.

Or having separate caches for (nope, rope, scale) will have big performance advantage?

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
```diff
@@ -0,0 +1,153 @@
...
+    wkv_wgate: jax.Array,           # [2 * coff * head_dim, hidden_size] fp32
+    ape: jax.Array,                 # [compress_ratio, coff * head_dim] fp32
+    norm_weight: jax.Array,         # [head_dim] fp32
+    cos_sin_cache: jax.Array,       # [max_pos, rope_head_dim] fp32
+    positions: jax.Array,           # [num_tokens] int
+    slot_mapping: jax.Array,        # [num_tokens] int
+    block_table: jax.Array,         # [num_reqs, max_blocks] int
+    token_to_req_indices: jax.Array,  # [num_tokens] int
+    kv_slot_mapping: jax.Array,     # [num_tokens] int
+    state_cache: jax.Array,         # [num_blocks, block_size, 2*coff*head_dim] fp32
```
gxd3: in vLLM, different caches may share the same underlying buffer (i.e. jax.Array in our case): https://screenshot.googleplex.com/C4D2UW2hUgNXMLE

In our case, the `state_cache`'s shape, dtype will be the same as the compressed kv Cache (they are pointing to the same jax.Array), we'll need to do bitcast to reinterpret it as f32 buffer; 

This part may be a bit complicated, we could leave a TODO and adjust that in a later PR.

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
```diff
@@ -0,0 +1,153 @@
...
+
+def compressor_forward(
+    hidden_states: jax.Array,       # [num_tokens, hidden_size] fp32
+    wkv_wgate: jax.Array,           # [2 * coff * head_dim, hidden_size] fp32
+    ape: jax.Array,                 # [compress_ratio, coff * head_dim] fp32
+    norm_weight: jax.Array,         # [head_dim] fp32
+    cos_sin_cache: jax.Array,       # [max_pos, rope_head_dim] fp32
+    positions: jax.Array,           # [num_tokens] int
+    slot_mapping: jax.Array,        # [num_tokens] int
+    block_table: jax.Array,         # [num_reqs, max_blocks] int
```
gxd3: nit: for clarity, maybe document `block_table` is for the state_cache or the compressed kv cache?

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
a1yssan13: Ya, makes sense, will leave a TODO!

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
a1yssan13: Ack'd, thanks!

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
gxd3: thanks!

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
```diff
@@ -0,0 +1,147 @@
...
+        score=score,
+        ape=ape,
+        positions=positions,
+        state_cache=state_cache,
+        slot_mapping=slot_mapping,
+        compress_ratio=compress_ratio,
+    )
+
+
+def compressor_forward(
```
gxd3, line 67: Wonder if it's possible to add basic unit test for correctness?
 
For example, have a very naive and simple reference implementation (e.g. eager execution, with python for / if, and don't need to respect static shape requirement etc) to compare result against this.

It's easier to catch issues (if any) at small scale / granularity than having to debug when E2E went wrong.

Otherwises, LGTMed.

gxd3 (approved):

tpu_inference/kernels/experimental/deepseek_v4/compress_norm_rope.py
```diff
@@ -0,0 +1,318 @@
...
+        [_to_byte_lane(q), _to_byte_lane(scale)], axis=-1)
+
+    num_blocks, blk, packed_width = kv_cache.shape
+    num_slots = num_blocks * blk
+    dest = _boundary_dest(positions, slot_mapping, kv_slot_mapping,
+                          compress_ratio, num_slots)
+
+    flat = kv_cache.reshape(num_slots, packed_width)
+    flat = flat.at[dest].set(packed, mode="drop")
+    return flat.reshape(num_blocks, blk, packed_width)
```
gxd3: for the compressed kv cache, wonder if we could use the shape
[num-pages, page-size // kv-packing, kv-packing, align(packed-width, 128)]?

MLA kernel has been using this shape for quite long time:
https://screenshot.googleplex.com/XY9kEtJi3NdutkH


v4 attn kernel (which consume the compressed kv cache) is a fork and modified version from v3's MLA kernel, and assume the same shape: https://screenshot.googleplex.com/BNpiM44ogaAgAqU.

related question: is there downside of using [num-pages, page-size // kv-packing, kv-packing, align(packed-width, 128)] kv cache shape?

tpu_inference/kernels/experimental/deepseek_v4/compress_norm_rope.py
```diff
@@ -0,0 +1,469 @@
...
+def _from_byte_lane(b: jax.Array, dtype) -> jax.Array:
+    """Inverse of ``_to_byte_lane``: read trailing bytes back as ``dtype``."""
+    itemsize = jnp.dtype(dtype).itemsize
+    if itemsize == 1:
+        return jax.lax.bitcast_convert_type(b, dtype)
+    grouped = b.reshape(*b.shape[:-1], b.shape[-1] // itemsize, itemsize)
+    return jax.lax.bitcast_convert_type(grouped, dtype)
+
+
+def quantize_fp8_ue8m0(x: jax.Array, block_size: int):
```
a1yssan13, line 67: this is to match upstream vLLM, which does something like this: 

```
# ── FP8 UE8M0 quant: single block, flat reduction ────────────────
    tl.static_assert(
        TRITON_BLOCK_SIZE == QUANT_BLOCK,
        "Indexer expects one quant block (QUANT_BLOCK == TRITON_BLOCK_SIZE)",
    )
    INV_FP8_MAX: tl.constexpr = 1.0 / FP8_MAX

    result_bf16 = result.to(tl.bfloat16).to(tl.float32)
    absmax = tl.max(tl.abs(result_bf16), axis=0)  # scalar
    absmax = tl.maximum(absmax, 1e-4)
    raw_scale = absmax * INV_FP8_MAX
    exponent = tl.ceil(tl.log2(raw_scale))
    inv_scale = tl.exp2(-exponent)

    x_scaled = result_bf16 * inv_scale
    x_clamped = tl.clamp(x_scaled, -FP8_MAX, FP8_MAX)
    x_fp8 = x_clamped.to(tl.float8e4nv)
    x_uint8 = x_fp8.to(tl.uint8, bitcast=True)

    tl.store(fp8_ptr + block, x_uint8, mask=mask)

    # Single float32 scale
    scale_val = tl.exp2(exponent)
    tl.store(scale_ptr.to(tl.pointer_type(tl.float32)), scale_val)
```

tpu_inference/kernels/experimental/deepseek_v4/compressor.py
a1yssan13: done!

gxd3: Thank you Alyssa!

tpu_inference/kernels/experimental/deepseek_v4/compress_norm_rope.py
gxd3: resolved

tpu_inference/kernels/experimental/deepseek_v4/compress_norm_rope.py
gxd3: ack!

gxd3 (approved):

PR #2860, opened 2026-06-10
Fix DP sampling padding bug by mapping correctly to JAX sharding slices

huybro:
**The Problem:** In `TPUSupportedSamplingMetadata.from_input_batch`, sampling parameters like temperature, top_k, and top_p were previously populated by placing all active requests contiguously at the front of the array, followed by a contiguous block of -1.0 padding at the tail.

When this global array was sharded across dp_size, the tail-end DP ranks received slices composed entirely of the -1.0 padding values. This caused real, active requests to silently execute using the -1.0 padding temperature (or other incorrect request parameters).

**Why it destroyed Speculative Decoding:** When rejection_sampler.py evaluates Draft model tokens, the Target model calls `_compute_probs()`, which applies `logits /= temperature `and then immediately calls softmax. Because there is no `is_greedy `safeguard check in `_compute_probs()`, receiving the -1.0 padding temperature causes it to divide by -1.0. This completely flips the logits upside down. 

**Why Standard Decoding Survived:** Standard decoding (sampling.py) was also suffering from the scrambled parameters, but it survived without producing gibberish because it contains a hardcoded safety net (is_greedy = temperature < 1e-5). 

**Why the E2E tests missed it:** The existing e2e test in test_speculative_decoding.py hardcodes` temperature=0`. When `temperature=0`, vLLM flags the batch as` all_greedy=True`, which triggers the `do_sampling=False` flag. This early-exits the sampler into `_greedy_rejection_sample_with_segment`, entirely bypassing the flawed `_compute_probs() `math where the division actually occurs.

The Solution: This PR modifies fill_slice in sampling_metadata.py and updates tpu_runner.py to pass the req_indices_dp layout mapping. Instead of a single contiguous block of data at the front, we now explicitly compute the chunk boundaries for each DP rank (dp_rank * padded_num_reqs_per_dp_rank) and place the assigned requests accurately within their respective shards.

# Tests

- Verified locally using `offline_inference.py` with `eagle3` (speculative decoding) on 4 ranks (`dp=4`, `tp=2`). 
- Ran generation with 5 active requests to reproduce the scenario where `num_reqs < padded_num_reqs`.
- Confirmed that the global padded arrays are now spaced correctly into chunks, and all DP ranks successfully generate valid text instead of corrupted artifacts.

# Checklist


kyuyeunk:
Oh this is a good find! 
@wenxindongwork @wyzhang can you review this pr?

tpu_inference/models/common/model_loader.py
```diff
@@ -93,6 +93,7 @@ def _get_model_architecture(config: PretrainedConfig) -> nnx.Module:
     _MODEL_REGISTRY[
         "Qwen2_5_VLForConditionalGeneration"] = Qwen2_5_VLForConditionalGeneration
     _MODEL_REGISTRY["Eagle3LlamaForCausalLM"] = EagleLlama3ForCausalLM
+    _MODEL_REGISTRY["LlamaForCausalLMEagle3"] = EagleLlama3ForCausalLM
```
lk-chen: What's the difference between these two?

huybro: My bad, this PR was accidentally branched from my local eagle3 work! I just force-pushed to remove those unrelated changes.

tpu_inference/layers/jax/sample/sampling_metadata.py
gxd3, line 54: Thanks Huy for the fix!

2 questions:
* does the spec-decoding-disabled code path also hit by this bug? If so, will this fix work for spec-decoding-disabled code path?

* we have e2e test for spec-decoding + dp> 1 in tests/e2e/test_speculative_decoding.py, why the existing test does not catch this bug? could we update the existing test so that it will fail before this fix?

tpu_inference/layers/jax/sample/sampling_metadata.py
```diff
@@ -76,11 +76,26 @@ def from_input_batch(
...
         num_reqs = input_batch.num_reqs
 
-        def fill_slice(cpu_torch_tensor: torch.Tensor,
-                       fill_val: float) -> torch.Tensor:
-            # Pad value is the default one.
-            cpu_torch_tensor[num_reqs:padded_num_reqs] = fill_val
-            return cpu_torch_tensor
+        def fill_slice(cpu_tensor_np: np.ndarray,
+                       fill_val: float) -> np.ndarray:
+            out_tensor = np.full((padded_num_reqs, ),
```
gxd3, line 80: we may not be able to use `padded_num_reqs`, which is a padding-up of `num_req` in Input_batch, which may not be divisible of DP-rank?

tpu_inference/layers/jax/sample/sampling_metadata.py
```diff
@@ -76,11 +76,26 @@ def from_input_batch(
...
+                                 fill_val,
+                                 dtype=cpu_tensor_np.dtype)
+
+            if req_indices_dp is not None:
```
gxd3: nit: `req_indices_dp` should be always non-None,

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: 1) Yes, the standard decoding path is affected by the parameter sharding bug, but it doesn't crash or output gibberish. sampling.py has a hardcoded `is_greedy` check that catches the -1.0 padding temperature and safely defaults to greedy sampling. So while standard decoding hasn't been catastrophically failing, it has been silently using the wrong sampling parameters. 

2) The e2e test uses a hardcoded `temp=0`. When `temp=0`, the batch is flagged as `all_greedy=True`, which triggers the `do_sampling=False` flag. This completely bypasses the `_compute_probs` math where the bug actually occurs, so it works perfectly.

To catch this before the fix, we just need to add a parameterized test case running `temp > 0` combined with `dp > 1`. That will immediately trigger the bug.

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: In tpu_runner.py, it is explicitly calculated as padded_num_reqs = padded_num_reqs_per_dp_rank * dp_size. So It should be divisible.

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: Done! Thank you for pointing out!

tpu_inference/layers/jax/sample/sampling_metadata.py
gxd3: I see, sorry for missing it, could we add assert padded_num_reqs % dp_size ==0 just to be safe?

tpu_inference/layers/jax/sample/sampling_metadata.py
gxd3: > "sampling.py has a hardcoded is_greedy check that catches the -1.0 padding temperature and safely defaults to greedy sampling. So while standard decoding hasn't been catastrophically failing, it has been silently using the wrong sampling parameters."

great catch! let's fix that as well?

> To catch this before the fix, we just need to add a parameterized test case running temp > 0 combined with dp > 1. That will immediately trigger the bug

let's add / adjust that test for coverage of this PR :)

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: Done!

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: I've added and adjusted those tests. Thanks for the suggestions. Please let me know if anything else needs tweaking!

gxd3: Thanks Huy!

do you want to update "tests/e2e/test_speculative_decoding.py" as well?

> "o while standard decoding hasn't been catastrophically failing, it has been silently using the wrong sampling parameters"

We could fix that in a later PR :)

gxd3 (approved):

tpu_inference/layers/jax/sample/sampling_metadata.py
```diff
@@ -74,13 +74,24 @@ def from_input_batch(
...
-            cpu_torch_tensor[num_reqs:padded_num_reqs] = fill_val
-            return cpu_torch_tensor
+        def fill_slice(cpu_tensor_np: np.ndarray,
+                       fill_val: float) -> np.ndarray:
+            out_tensor = np.full((padded_num_reqs, ),
+                                 fill_val,
+                                 dtype=cpu_tensor_np.dtype)
+
+            dp_size = len(req_indices_dp)
+            assert padded_num_reqs % dp_size == 0, "padded_num_reqs must be divisible by dp_size"
```
lk-chen: nit, show `padded_num_reqs` value in assertion message

huybro: Done! Thank you for pointing out!

tpu_inference/layers/jax/sample/sampling_metadata.py
huybro: I just add another test in "tests/e2e/test_speculative_decoding.py".

> We could fix that in a later PR :)

I believe this work will fix the error I mentioned!

gxd3: I see, thanks for confirming!

Lumosis (approved):

jrplatin (approved):

lk-chen (approved):

PR #2866, opened 2026-06-10
Add Batched RPA Kernel Tuner

patrickji2014:
# Description

1. Add tuner for batched rpa kernel
2. MISC kernel tuner infra enhancement
  2.1 Support OOM early termination, this requires each TunableParams dataclass to override the ``__ge__`` function so if smaller tunable_param is OOM, larger tunable_param will automatically OOM.
  2.2 Use xprof to measure kernel run latency, this requires each kernel tuner to provide a jit_kernel_name which will be used to search the module in the xprof.
  2.3 Default job bucket size to 500 which is more reasonable so each buildkite step will not execute too short.
  2.4 Use kernel name as each local run's DB file path's prefix
  2.5 Update README.md according
  2.6 Extracted common datatypes (TuningCase, TunableParams, RunConfig, etc.) from `kernel_tuner_base.py` into a dedicated `tuner_datatypes.py`.
  2.7 Introduced `TuningCaseLogger` to handle logging and retrieval of tuning cases.
  2.8 Centralize tuning cases storage into tuning_cases folder
3. Add palceholder for getting the tuned block size in batched_rpa kernel wrapper

# Tests

Validate locally via command line: 
python -m tools.kernel.tuner.v1.kernel_tuner_runner --run_locally --kernel_tuner_name=batched_rpa_kernel_tuner --case_set_desc=batched_rpa_kernel_tuning_setup --case_set_id=batched_rpa_0 --run_id=0 --tpu_version=tpu7x --tpu_cores=2

Validate using BuildKite: 
https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/111/canvas 
Build Parameters:
KERNEL_TUNING_KERNEL_NAME=batched_rpa_kernel_tuner
KERNEL_TUNING_CASE_SET_ID=batched_rpa_kernel_tuner_setup_0
KERNEL_TUNING_RUN_ID=001
KERNEL_TUNING_CASE_SET_DESC=batched_rpa_kernel_tuner
KERNEL_TUNING_TPU_VERSION=tpu7x
KERNEL_TUNING_TPU_CORES=2

# Checklist


tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -418,6 +427,15 @@ def run(self, tuning_key: TuningKey, tunable_params: TunableParams,
...
         )
 
+    def _cleanup_xprof_dir(self):
+        """Clean up the xprof directory to avoid interference from previous runs."""
+        if os.path.exists(self.xprof_dir):
+            try:
+                os.system(f'rm -rf {self.xprof_dir}/*')
+            except Exception as e:
+                logger.warning(
+                    f"Failed to clean up xprof dir {self.xprof_dir}: {e}")
```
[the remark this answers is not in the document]
patrickji2014: fixed.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -139,6 +144,9 @@ def __init__(self,
         self._tuning_key = None
         self.tuner_config = tuner_config
         self.run_config = run_config
+        self.xprof_dir = os.path.join("/tmp/kernel_tuning",
+                                      self.tuner_config.kernel_tuner_name,
+                                      "xprof")
```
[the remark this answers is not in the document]
patrickji2014: concurrent run not applied here. won't fix.

tools/kernel/tuner/v1/common/utils.py
```diff
@@ -0,0 +1,78 @@
...
+            line_name = line.name if line.name else f"ID: {line.id}"
+
+            for event in line.events:
+                name = event_names.get(event.metadata_id, "Unknown Event")
+
+                # 4. Check if the event name matches our regex pattern
+                if pattern.search(name):
+                    duration_us = event.duration_ps // 1e6
+                    start_us = event.offset_ps // 1e6
+
```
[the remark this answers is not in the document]
patrickji2014: fixed.

tools/kernel/tuner/v1/common/utils.py
```diff
@@ -0,0 +1,78 @@
...
+            for file in files:
+                if file.endswith(".xplane.pb"):
+                    pb_files.append(os.path.join(root, file))
+        if not pb_files:
+            raise FileNotFoundError(
+                f"No .xplane.pb files found in directory: {pb_file_path}")
+        assert len(
+            pb_files
+        ) == 1, f"Multiple .xplane.pb files found in directory: {pb_file_path}. Please specify the file path directly."
+        pb_file_path = pb_files[0]
```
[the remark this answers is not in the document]
patrickji2014: multi-host or multi-rank not applied here. Won't fix.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
```diff
@@ -0,0 +1,112 @@
...
+        return configs.BlockSizes(bq_sz=self.bq_sz,
+                                  bq_c_sz=self.bq_c_sz,
+                                  bkv_sz=self.bkv_sz,
+                                  batch_size=self.batch_size,
+                                  n_buffer=self.n_buffer)
+
+    # Define comparison operators for skipping tuning case when smaller block sizes hit OOM already
+    def __ge__(self, other) -> bool:
+        return self.bq_sz >= other.bq_sz and self.bq_c_sz >= other.bq_c_sz and self.bkv_sz >= other.bkv_sz and self.batch_size >= other.batch_size
+
```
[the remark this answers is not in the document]
patrickji2014: it's okay to be conservative to not skip.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
```diff
@@ -491,18 +496,17 @@ def ragged_paged_attention(
...
         kv_cache: jax.Array,
     ):
         if mode == configs.RpaCase.DECODE:
-            effective_blocks = decode_block_sizes or default_decode
+            effective_blocks = decode_block_sizes or get_tuned_params(
+                model_cfgs, serve_cfgs, case='decode')
         else:
-            effective_blocks = prefill_block_sizes or default_prefill
+            effective_blocks = prefill_block_sizes or get_tuned_params(
+                model_cfgs, serve_cfgs, case='prefill')
```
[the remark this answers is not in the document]
patrickji2014: fixed.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,302 @@
...
+
+    pages_per_seq = num_page_indices // num_seqs
+    kv_lens = [max_input_len
+               ] * max_prefill_seqs + [0] * (num_seqs - max_prefill_seqs)
+
+    cu_q_lens = [1] * (num_seqs + 1)
+    cu_q_lens[0] = 0
+    for i in range(1, max_prefill_seqs + 1):
+        cu_q_lens[i] = cu_q_lens[i - 1] + max_input_len
+
```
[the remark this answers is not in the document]
patrickji2014: This is intended. Won't fix.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
```diff
@@ -373,7 +375,10 @@ def find_best_block_sizes(
         "use_causal_mask",
         "update_kv_cache",
     ),
-    donate_argnames=("queries", "keys", "values", "kv_cache"),
```
QiliangCui2023, line 385: I have concern of doing this. 

Are we sure the impact of removing "donation" from here? 

cc: @kyuyeunk and @a1yssan13

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
```diff
@@ -0,0 +1,112 @@
...
+
+def get_tuned_params(
```
QiliangCui2023, line 286: This is new production code under `tpu_inference/**` (`get_tuned_params`, `TunableParams.__ge__`, `TuningKey.from_config`) but it has no unit test exercising it — the only test added in this PR is the tuner smoke test under `tools/`. Per repo policy, production code in `tpu_inference/` needs a test that actually covers the new behavior. Please add one that hits:

- `get_tuned_params` empty-mapping fallback to `calculate_block_sizes` (the current default), and a populated-mapping hit returning the tuned `BlockSizes`,
- the `TunableParams.__ge__` partial order (including the `n_buffer` dimension).

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
```diff
@@ -0,0 +1,112 @@
...
+    def to_block_sizes(self) -> configs.BlockSizes:
+        return configs.BlockSizes(bq_sz=self.bq_sz,
+                                  bq_c_sz=self.bq_c_sz,
+                                  bkv_sz=self.bkv_sz,
+                                  batch_size=self.batch_size,
+                                  n_buffer=self.n_buffer)
+
+    # Define comparison operators for skipping tuning case when smaller block sizes hit OOM already
+    def __ge__(self, other) -> bool:
+        return self.bq_sz >= other.bq_sz and self.bq_c_sz >= other.bq_c_sz and self.bkv_sz >= other.bkv_sz and self.batch_size >= other.batch_size
```
QiliangCui2023: intentionally no n_buffer?

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -468,46 +498,97 @@ def measure_latency(self, begin_case_id: int, end_case_id: int):
...
 
-            begin_case_id_time = time.perf_counter_ns()
-            # status can be SUCCESS, FAILED_OOM, UNKNOWN_ERROR.
-            status, warmup_ns, _ = self.run(tuning_key,
-                                            tunable_params,
-                                            iters=1)
-            if status != TuningStatus.SUCCESS:
+            # check whether tuning_key is same as last one and if last one is OOM, then we can skip
+            if any(tuning_key == k and s == TuningStatus.FAILED_OOM
+                   and p <= tunable_params
```
QiliangCui2023, line 544: This OOM-skip uses `p <= tunable_params`, but `TunableParams` only defines `__ge__` (no `__le__`). It happens to work because Python falls back to the reflected operator `tunable_params.__ge__(p)`, but that dependency is non-obvious and fragile — anyone adding `__le__` later (or a different `TunableParams` that does define it) silently changes the meaning here. Please write it explicitly as `tunable_params >= p` (or define `__le__` alongside `__ge__`).

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -468,46 +498,97 @@ def measure_latency(self, begin_case_id: int, end_case_id: int):
...
                 continue
 
-            average_latency_us = int(average_latency_ns // 1000)
-            total_time_us = int(total_time // 1000)
+            if self.tuner_config.jit_kernel_pattern is not None:
+                matching_events, average_latency_us = find_events_by_pattern(
+                    self.xprof_dir, self.tuner_config.jit_kernel_pattern)
+                assert len(
+                    matching_events
+                ) == measurement_iters, f"The number of matching events {len(matching_events)} is different from the measurement iters {measurement_iters} for Case {cid}."
```
QiliangCui2023: This `assert len(matching_events) == measurement_iters` runs at the top level of `measure_latency` (not inside `run()`'s try/except), so a single xprof capture that yields the wrong event count raises and aborts the entire bucket/worker mid-run. That defeats the point of bucketed tuning + OOM early-termination, which is to survive long unattended runs. Please demote this to a logged `UNKNOWN_ERROR` for the affected case and `continue` to the next, rather than killing the whole bucket on one flaky measurement.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
patrickji2014: yes, i originally intentionally left the n_buffer behind to be conservative but it should be find to add it back since it's just more buffer in the vmem. when more buffer allocated in vmem, hit oom earlier.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: `__le__` added

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
patrickji2014: i added it back.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: demote it to a XPROF_MEASUREMENT_ERROR rather than assert

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
patrickji2014: I got UserWarning: Some donated buffers were not usable: float8_e4m3fn[8192,1,256], float8_e4m3fn[8192,1,256]. which indicates the key/value buffer are not used. Remove the key/value from the list should resolve the warning.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
patrickji2014: added.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
```diff
@@ -0,0 +1,115 @@
...
+def get_tuned_params(
+        model_config: configs.ModelConfigs,
+        serve_config: configs.ServingConfigs,
+        vmem_limit_bytes: int | None = None,
+        case: Literal['decode', 'prefill'] = 'decode') -> configs.BlockSizes:
+    if vmem_limit_bytes is None:
+        vmem_limit_bytes = pltpu.get_tpu_info().vmem_capacity_bytes
+    tuning_key = TuningKey.from_config(model_config, serve_config, case=case)
+    if tuning_key not in tuned_params_mapping:
+        from tpu_inference.kernels.experimental.batched_rpa.wrapper import \
```
a1yssan13: Would it make more sense to just return None if tuning_key doesn't exist vs. importing wrapper into get_tuned_params and then importing get_tuned_params into wrapper? 

And then we can fall back on calculate_block_sizes in wrapper.py?

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
a1yssan13: I think we sync'd on this offline, did using .copy() for your test cases fix it? would prefer not removing this if possible

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,303 @@
...
+            serve_config = log_entry.serve
+            prefill_tuned_block_size = log_entry.prefill_block_sizes
+            prefill_tuning_key = TuningKey.from_config(model_config,
+                                                       serve_config,
+                                                       case='prefill')
+            prefill_tunable_params = TunableParams(**asdict(
+                prefill_tuned_block_size),
+                                                   is_baseline=True)
+            # We setup this for tuning Gemme4's rpa block sizes. When total_q_tokens is smaller than sequence input length,
+            # it cannot be used for scheduling prefill case. We only tuned for prefill case at this moment.
```
a1yssan13: thanks for doing this! curios if is adding decode support a future PR / i'm wondering why we prioritized prefill?  (is adding decode support much harder to do)?

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,303 @@
...
+                continue
+            tuning_cases.append(
+                TuningCase(tuning_key=prefill_tuning_key,
+                           tunable_params=prefill_tunable_params))
+
+            bq_c_sz = prefill_tunable_params.bq_c_sz
+            bkv_sz = prefill_tunable_params.bkv_sz
+            n_buffer = prefill_tunable_params.n_buffer
+
+            for prefill_batch_size in [1, 2]:
```
a1yssan13: I would make this [1, 2, 3] since historically some kernels performed best with batch_size = 3

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,303 @@
...
+                for bq_sz in range(512, 8193, 512):
```
a1yssan13: i don't think i've ever seen bq_sz to be necessary up to 8k? 

My intuition is that 2k is probably large enough 

let me know if your sweeps indicate differently

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,303 @@
...
+                    for bq_c_sz in range(128, bq_sz + 1, 128):
```
a1yssan13: from experience i think optimal bq_c_sz can go as low as 16 

i think something like (8, 16, 32, 64, 128) is a good set 

at least for some of the other models i've handled i would actually probably say this might be the best parameter to tune that the heuristic isn't doing great on?

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,303 @@
...
+                        if bq_sz % bq_c_sz != 0:
+                            continue
+                        for bkv_sz in range(512, 8193, 512):
```
a1yssan13: i think it's probably ok to keep bkv_sz below 2k 

maybe like 

(256, 512, 768, 1024) 

is sufficient to cover most cases i've seen?

a1yssan13 (commented):
Thanks for making this! super helpful :) 

I left some comments on what i think some good prefill search spaces are -- but i've only had experience tuning like a few models so my suggestions may not cover some other optimal sizes. 

totally optional suggestion -- but would it be possible to do a highly finegrained 1-time search of a few models you care about (maybe 2 -- gemma / qwen 3.5?) and see what the optimal sizes are? And once you develop a sense of where the optimal configs for both fall then we can narrow the search space a bit more...

kyuyeunk:
> totally optional suggestion -- but would it be possible to do a highly finegrained 1-time search of a few models you care about (maybe 2 -- gemma / qwen 3.5?) and see what the optimal sizes are? And once you develop a sense of where the optimal configs for both fall then we can narrow the search space a bit more...

+1 on this. we really want kernel tuning to be incorporated into our optimization feed back loop. instead of tile tuning being seen as a completely separate process for kernel development, there might be some interesting insights that tile tuner provides - for instance, tile tuner might figure out that a tile size that we initially thought was really suboptimal actually performs really good. and we can leverage this information to figure out why our initial hypothesis was wrong and most importantly - create a optimization technique that can leverage this new insight.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
patrickji2014: using .copy() doesn't fix this.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
patrickji2014: The imports here should be fine and will not introduce recursive import error here. But actually i would like to move calculate_block_sizes into this tuned_params.py as we are building the e2e auto kernel tuning pipeline. we need to have the baseline so it's prefer for the get_tuned_params returns a TunedParams here rather than None.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: Decode should be supported later in following PR. This PR is initiated from tuning batched rpa for gemma4 prefill.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: thanks. Added 3 into the list

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: thanks for the input. updated

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: updated. thanks.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: update to cover from 256 to 2048 with 256 step size.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
patrickji2014: Add them back as they are not critical to this PR.

tpu_inference/kernels/experimental/batched_rpa/tuned_params.py
patrickji2014: move the function from wrapper.py to this file.

patrickji2014:
> > totally optional suggestion -- but would it be possible to do a highly finegrained 1-time search of a few models you care about (maybe 2 -- gemma / qwen 3.5?) and see what the optimal sizes are? And once you develop a sense of where the optimal configs for both fall then we can narrow the search space a bit more...
> 
> +1 on this. we really want kernel tuning to be incorporated into our optimization feed back loop. instead of tile tuning being seen as a completely separate process for kernel development, there might be some interesting insights that tile tuner provides - for instance, tile tuner might figure out that a tile size that we initially thought was really suboptimal actually performs really good. and we can leverage this information to figure out why our initial hypothesis was wrong and most importantly - create a optimization technique that can leverage this new insight.

we have setup the e2e autotune pipeline and so it can search for all the inputs shapes/dtypes we used in our benchmarks. With this we can get all the cases with a broader search spaces and hopefully from there we can use some tools to provide more insight. https://github.com/vllm-project/tpu-inference/pull/2984

patrickji2014:
> Thanks for making this! super helpful :)
> 
> I left some comments on what i think some good prefill search spaces are -- but i've only had experience tuning like a few models so my suggestions may not cover some other optimal sizes.
> 
> totally optional suggestion -- but would it be possible to do a highly finegrained 1-time search of a few models you care about (maybe 2 -- gemma / qwen 3.5?) and see what the optimal sizes are? And once you develop a sense of where the optimal configs for both fall then we can narrow the search space a bit more...

definitely, at the same time, as we have some optimization technique (like early termination or using [bayesian optimizer](https://github.com/vllm-project/tpu-inference/pull/3111)) in place in the kernel tuner, we can always use broader tuning for all the cases. Especially using bayesian opt, we can intelligently reduce the search steps even the space is defined huge.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -591,46 +539,107 @@ def measure_latency(self, begin_case_id: int, end_case_id: int):
...
+                        iters=measurement_iters,
+                        warmup_us=warmup_us)
+            else:
+                status, average_latency_ns = run_and_record_failure(
+                    tuning_key,
+                    tunable_params,
+                    iters=measurement_iters,
+                    warmup_us=warmup_us)
             if status != TuningStatus.SUCCESS:
                 results_buffer.append(
```
QiliangCui2023: **Duplicate result write on main-run failure.** The new `run_and_record_failure` helper added just above already appends the failed case to `results_buffer` when `status != SUCCESS`, so this pre-existing append records the same case a second time. The warmup-failure branch was updated to only `logger.warning` + `continue` — this main-run branch should be too.

The two writes aren't even consistent: the helper writes the tuple as `(…, 0, warmup_us, 0)` (`Latency=0, WarmupTime=warmup_us`), while this one writes `(…, warmup_us, 0, 0)` (`Latency=warmup_us, WarmupTime=0`). Since both rows share the same `(ID, RunId, CaseId)` and `save_results_batch` uses `insert_or_update`, the second clobbers the first, so a failed case is persisted with `warmup_us` mislabeled in the `Latency` column — and issuing two mutations for the same key in one Spanner batch is itself fragile.

Suggest deleting this `results_buffer.append(...)` and keeping only the `logger.warning` + `continue`.

.buildkite/pipeline_jax.yml
```diff
@@ -627,6 +627,7 @@ steps:
         commands:
           - |
             .buildkite/scripts/run_in_docker.sh bash -c "
+              pip install --upgrade -r tools/kernel/tuner/v1/storage_management/requirements.txt tensorflow &&
```
QiliangCui2023: **Pin `tensorflow` and avoid `--upgrade`-ing it into the jax[tpu] image.** This pulls the full, unpinned `tensorflow` with `--upgrade` into the libtpu container solely so `tools/kernel/tuner/v1/common/utils.py` can `import tensorflow.tsl.profiler.protobuf.xplane_pb2`. Two concerns:

1. `--upgrade tensorflow` can bump shared deps (`ml-dtypes`, `numpy`, `protobuf`) beyond what the pinned jax/libtpu support and break the runtime at import.
2. `tensorflow.tsl.profiler.protobuf` is a TF-internal module path that has moved between TF releases, so an unpinned install can silently break this step on a future `tensorflow`.

Recommend pinning a known-good version and using `tensorflow-cpu`, or vendoring just the xplane proto to drop the TF dependency. (The same `--upgrade … tensorflow` is also added in `.buildkite/pipeline_kernel_tuning.yml` and `KernelTunerBase._build_step`.)

tpu_inference/kernels/mla/v2/tuned_params.py
```diff
@@ -52,8 +52,15 @@ class TunableParams:
     # with steps of powers of two.
     num_queries_per_block: int  # for batched_decode, this is always 1
     vmem_limit_bytes: int  # 16MiB(?) to 64MiB, increments of 8MiB.
+
     # Select lowest value that gives the highest performance
 
+    def __ge__(self, other) -> bool:
```
QiliangCui2023, line 58: **Please add tests for these operators and the OOM-skip path.** These `__ge__`/`__le__` encode a non-obvious inversion: `vmem_limit_bytes` is compared in the opposite direction to the other three fields (a higher limit is treated as "smaller"/less OOM-prone). The direction looks correct, but it's exactly the kind of ordering that regresses silently and it has no test — the batched-RPA equivalents are covered in `tests/kernels/experimental/batched_rpa/test_tuned_params.py`, but MLA's are not.

More importantly, the consumer of these operators — the OOM early-termination skip in `KernelTunerBase.measure_latency` (`p <= tunable_params` ⇒ mark `SKIPPED`) — has no unit test at all, though it's the headline feature of this PR. It's easily testable by mocking `self.run` to return `FAILED_OOM` for a small params set and asserting a larger set is skipped. Such a test would also have caught the duplicate-write issue flagged in `kernel_tuner_base.py`.

QiliangCui2023 (commented):
A few review notes inline (result double-write, tensorflow pinning, and test coverage for the new comparison operators / OOM-skip).

QiliangCui (approved):

.buildkite/pipeline_jax.yml
patrickji2014: remove --upgrade option and pip tensorflow to 2.21.0. update other yml file also.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: fixed by removing this second append.

tpu_inference/kernels/mla/v2/tuned_params.py
patrickji2014: Added unittest for this tuned_params.py for testing the le/ge. Added unittest for covering the early oom termination.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,322 @@
...
+        """Returns independent lists of candidate values for each tunable param.
+
+        For batched RPA, all bq_sz values are multiples of 256 and all bq_c_sz
+        values are powers of 2 ≤ 128, so the constraint bq_sz % bq_c_sz == 0
+        is always satisfied and no filtering is required for those two params.
+        bkv_sz values are filtered to multiples of page_size as required by the
+        scheduler.
+        """
+        bkv_sz_list = [
+            v for v in range(256, 2049, 256) if v % tuning_key.page_size == 0
```
a1yssan13, line 259: Can you leave a comment here that these values are specific for PREFILL case and there's a TODO for decode specific search spaces in the future? Thanks!

a1yssan13 (approved):

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: Done. Thanks

gxd3 (approved):

PR #2869, opened 2026-06-11
Add MoE unpermute pipelining with collectives via VLLM_MOE_CHUNK_SIZE

zhangamy-crypto:
# Description

4 Way pipeline with MoE combine-reduce using SC kernel + psum_scatter on TC, this is on prefill only. I created a new configurable VLLM_MOE_CHUNK_SIZE, which is set to 256 in our case. I have tested 2 way as well, but 4 way performed better. I'm running against 6/2's commit, so
baseline: http://xprof.corp.google.com/trace_viewer/hning-9120565002754914291. After: https://xprof.corp.google.com/trace_viewer/zhangamy-5070159141590755285?view_start=809.465&view_end=811.451

Improvement: 17-18% reduction in latency for 8k prefill 4 Layer (3 GDN+1GQA) E2E latency
baseline: average 20ms;    median 19.4ms.
after:       average 16.2ms; median 16.1ms.
This is based on 34*4 complete layers.

Post MoE ragged_gather_reduce + ReduceScatter + post reduce f32->bf16 convert_select_fusion total time: baseline median: 1409us.
after median: 1008us. only a small portion of the 1st ragged_gather_reduce is exposed, followed by 4 ReduceScatters. Using median over average here since the impact of expert imbalance on 1 vs. 4 gather_reduce/RS per layer makes average not really apples to apples comparison.

Key notes:
xla_tpu_enable_sparse_core_collective_offload_reduce_scatter=false --xla_tpu_ars_combiner_threshold_in_bytes=0 --xla_tpu_enable_async_collective_merger=false are required to make the pipelined ReduceScatter work on TC.

the past.

Quality: ran mmlu_pro eval, got 0.8343 stderr 0.0138, on par with latest values

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -69,6 +69,7 @@ if TYPE_CHECKING:
     NUM_PRECOMPILE_WORKERS: int = 1
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
+    VLLM_MOE_CHUNK_SIZE: int = 0
     ONEHOT_MOE_PERMUTE_THRESHOLD: int = 0
     PROFILE_SINGLE_DEVICE: bool = False
     LORA_MODULE_PATH: str = ""
@@ -405,6 +406,8 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: int(os.getenv("DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS", "30000")),
     "MLA_XPOSE_N_TILE_SIZE":
     lambda: int(os.getenv("MLA_XPOSE_N_TILE_SIZE", "160")),
+    "VLLM_MOE_CHUNK_SIZE":
+    lambda: int(os.getenv("VLLM_MOE_CHUNK_SIZE", "0")),
     # Use Onehot+Matmul for permute and unpermute before and after moe
     # when the batch size <= this threshold. When set to 0, this feature
     # is effectively disabled.
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
     NUM_PRECOMPILE_WORKERS: int = 1
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
+    VLLM_MOE_CHUNK_SIZE: int = 0
     ONEHOT_MOE_PERMUTE_THRESHOLD: int = 0
     PROFILE_SINGLE_DEVICE: bool = False
     LORA_MODULE_PATH: str = ""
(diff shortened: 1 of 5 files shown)


kyuyeunk:
awesome change! @guowei-dev @gxd3 can you take a look at this pr since you both worked on this specific part of moe layer before?

tpu_inference/envs.py
```diff
@@ -60,6 +60,7 @@
     NUM_PRECOMPILE_WORKERS: int = 1
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
+    VLLM_MOE_CHUNK_SIZE: int = 0
```
kyuyeunk, line 72: so in general, i really really prefer having a good default option instead of having flag guarded option. and i feel like this threshold can be formulate as a function of dma latency vs. bandwidth requirement of each chunk. is it possible to derive such default value?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -263,20 +209,104 @@ def moe_gmm_local(x: jax.Array,
...
+                    num_devices=scatter_axis_size,
+                    num_micro_batches=num_mb,
+                    axis_name=reduction_axis)
+                out = rs_out.astype(x.dtype)
+        elif scatter_results:
+            if reduce_axes:
+                chunk_hidden = jax.lax.psum(chunk_hidden,
+                                            axis_name=reduce_axes)
+            if scatter_axes:
+                out = jax.lax.psum_scatter(chunk_hidden,
```
gxd3, line 352: Not sure if i understand correctly:
chunking and reduce-scatter will reorder rows incorrectly?
for example, there is 2 rank in the system only, each rank has 2 rows, before all-gather of MOE,
rank0: r0, r1
rank1: r2, r3

if the chunk-size is 2,
after post-gmm chunking-and-reduce-scatter:
rank0 will get r0, r2
rank1 will get r1, r3

guowei-dev: +1 on the correctness issue. The per-chunk reduce-scatter + concat effectively transposes tokens across ranks, so each rank ends up with the wrong tokens. Can we add a correctness test that compares chunked vs non-chunked output on a DP/EP path with more than one chunk at the fused moE layer level (e.g. tests/layers/jax/moe/test_moe.py)? The current kernel test only calls ragged_gather_reduce directly with x.shape[0] == indices.size, so it can't actually cover the newly added host-side chunk + collective path.

zhangamy-crypto: Thanks for identifying this. I have the fix for the numerical issue already but am trying to get the overlap to work again on latest commit/jax version. Will ping for review when I have both working.

zhangamy-crypto: Okay PR #2887 fixed the overlap issue and the overlap is working again. xprof: https://screenshot.googleplex.com/3T5zUUEjcNbJT93
also ran mmlu_pro eval, got 0.8343 stderr 0.0138, on par with latest values.

tpu_inference/envs.py
zhangamy-crypto: the tradeoff in reality can differ from theoretical roofline. for example, the RS could be slower than expected (i actually ran into this during my testing), so i think it's best to not set a default. at least for now.

gxd3 (approved):

kyuyeunk (approved):

PR #2887, opened 2026-06-15
Lower single-mesh SparseCore kernels through core_map to fix Jax 0.10.1 regression

guowei-dev:
Jax 0.10.1 lowers single-mesh pl.kernel via mpmd_map instead of core_map and led to a perf regression.

Route the SparseCore ragged kernels through a small core_map helper that keeps the fast single-mesh lowering, restoring the overlap without an upstream jax change.

Qwen3.5 - 8k/1k conc=512 Total tok/s:
baseline on HEAD : 22410.4 (Avg of 22224, 22546, 22462)
After fix: 24259.9 (Avg of 24200, 24199, 24380)


tpu_inference/kernels/sparse_core/core_map_helper.py
kyuyeunk, line 1: can you elaborate why this helper function is needed? is it not possible to just use core_map as-is? is there any api difference?

guowei-dev: Yeah, mostly because their APIs are different.

pl.kernel is value-in/value-out: it automatically handles ref allocation, wraps operands, and reads back results. core_map is the lower-level primitive underneath it: it takes no inputs and returns nothing, only mutating closed-over refs.

When we used core_map directly, we'd have to write manual ref-allocation and read-back boilerplate at every single call site. The helper centralizes this. It essentially restores the pre-0.10.1 behavior of pl.kernel (reusing most of the official code in [jax/_src/pallas/helpers.py from before JAX commit 18597c6](https://github.com/jax-ml/jax/commit/18597c6032#diff-093b5e2302b2d718e92c5fa66034a789c24a1551cf546559698e65ca0fcff947)). This lets us keep the exact same signature while skipping the new mpmd_map ref aliasing that was serializing the two kernels.

kyuyeunk: got! a nit comment but maybe renaming the file name to contain "helper" might be a good idea? just a suggestion and either is fine.

kyuyeunk (approved):

tpu_inference/kernels/sparse_core/core_map_helper.py
guowei-dev: Sure, updated

tpu_inference/kernels/sparse_core/core_map_helper.py
```diff
@@ -0,0 +1,73 @@
...
+
+``pl.kernel`` lowers through ``mpmd_map``, which is slower than the ``core_map``
+path for these single-mesh kernels on TPU. ``kernel`` here is a drop-in
+replacement for ``pl.kernel`` that uses ``core_map`` directly, so the SparseCore
+kernels keep that lowering without depending on the jax-side ``pl.kernel``
+implementation.
+"""
+
+from jax._src import api
+from jax._src import core as jax_core
```
helloworld1, line 24: Since this workaround involves jax private APIs and this can break after jax updates. Do we expect the Jax to update soon so we can remove this workaround?

guowei-dev: Yes, we have raised this issue in the jax-dev oncall thread and reached out to the active maintainers. Will keep following up with them.

PR #2894, opened 2026-06-15
Add MOE_REQUANTIZE_CLIP_PERCENTILE for FP4 requantization

AahilA:
When requantizing FP8 MoE weights to FP4 with large block sizes (e.g. 512), extreme weight outliers inflate the per-block scale and waste dynamic range for the majority of values.

This adds an optional MOE_REQUANTIZE_CLIP_PERCENTILE env var that clips outlier weights before quantization.

Usage:
  MOE_REQUANTIZE_CLIP_PERCENTILE=99.9 \
  MOE_REQUANTIZE_WEIGHT_DTYPE=float4_e2m1fn \
  MOE_REQUANTIZE_BLOCK_SIZE=512 \
  vllm serve ...


tpu_inference/envs.py
```diff
@@ -286,6 +287,12 @@ def _get_int_list_env() -> list[int]:
     "MOE_REQUANTIZE_BLOCK_SIZE":
     lambda: int(block_size)
     if (block_size := os.getenv("MOE_REQUANTIZE_BLOCK_SIZE")) else None,
+    # Clip outlier weights before requantization at the given percentile
+    # (e.g. 99.9). Reduces quantization error for large block sizes by
+    # preventing extreme outliers from inflating the per-block scale.
+    "MOE_REQUANTIZE_CLIP_PERCENTILE":
```
jrplatin, line 293: Does it make more sense just to give it as a decimal (e.g. 0.999) off the bat instead of 99.9?

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -913,6 +913,18 @@ def _requant_expert_batch_fn(
     w2_pad_widths = ((0, 0), (0, hidden_pad), (0, inter_pad))
     w2_fp32 = jnp.pad(w2_fp32, w2_pad_widths)
 
+    if envs.MOE_REQUANTIZE_CLIP_PERCENTILE is not None:
```
jrplatin, line 916: Is this cleaner?

```
if envs.MOE_REQUANTIZE_CLIP_PERCENTILE is not None:
    percentile_val = envs.MOE_REQUANTIZE_CLIP_PERCENTILE 
    
    for arr_name in ('w13', 'w2'):
        arr = w13_fp32 if arr_name == 'w13' else w2_fp32
        
        clip_val = jnp.percentile(jnp.abs(arr), percentile_val)
        
        if arr_name == 'w13':
            w13_fp32 = jnp.clip(w13_fp32, -clip_val, clip_val)
        else:
            w2_fp32 = jnp.clip(w2_fp32, -clip_val, clip_val)
```

AahilA: I agree this looks cleaner. However then I think we should keep the value to 99.9 not decimal 0.999 since jnp.percentile requires a value from 1-100.

tpu_inference/envs.py
AahilA: If we do as you said below and and use jnp.percentile then we need to keep it as 99.9.

lk-chen:
what bug/feature are you working on?

cc @lxhfirenking

tpu_inference/layers/common/process_weights/moe_weights.py
jrplatin: Ok yep sg!

jrplatin (approved):

AahilA:
I am working on TPU FP4 quality benchmarking. These changes are required for better quality on the Qwen3.5 model wrt to TPU FP4 quantization.

lk-chen:
Have you tried applying these clip offline? we are trying to prevent unnecessary env vars

QiliangCui (approved):

AahilA:
Its definitely possible to clip offline, however it would mean baking into a single percentile choice while online would let us tune it without regenerating a large checkpoint files. This env var is effectively a noop when unset and the clipping code itself is only a few lines of code.

PR #2902, opened 2026-06-16
Short circuit step on Buildkite level so that we don't block on waiting for TPUs

theminghuang:
Move the short circuit logics out of the script and do that at buildkite level.
If we do the check in the script, we still need to wait for a tpu to run it. Doing it at the buildkite level allow us to bypass it entirely.

For per pr ci run, this reduces the number of jobs performed from 64 to 48, saving 16 unnecessary dispatches to tpu.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -173,23 +173,17 @@ steps:
           TPU_VERSION: "${TPU_VERSION:-tpu6e}"
         agents:
           queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_queue}
+        # Skips this step without provisioning TPUs unless NIGHTLY=1 or kernel-related files are modified.
+        # Check bootstrap.sh for the exact file regex.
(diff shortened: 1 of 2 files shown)


theminghuang:

- #2902 ◀

.buildkite/pipeline_jax.yml
QiliangCui, line 683: this may fail.

QiliangCui, line 686: when skipped, this step may fail i remember. take a look at the check_results.sh logic, if some step doesn't exists what happens.

theminghuang: Look like they are correctly marked as skipped as opposed to missing https://screenshot.googleplex.com/33dVWSFP7PQ96Gd.  and we don't fail the step if it's skip https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/scripts/check_results.sh#L36

QiliangCui (approved):

PR #2903, opened 2026-06-16
[DeepSeek V4] initial version of main attention kernels

gxd3:
[DSv4] initial version of main attention kernel (copied from g3)

break the main attention into 2 subkernel for simplicity

* mla_swa.py: it writes kv to sliding window cache & do attention on the kvs in the sliding window cache

* mla.py: it takes the output from sliding window attn above and combine the attention with compressed kv cache to produce the final output. It doe NOT write to compressed KV cache (in current implementation, the compressed kv cache is written by token compressor)

Test coverage: unit tests


inucool12 (approved):
LGTM

PR #2909, opened 2026-06-17
[MoE] Fuse the MoE token permute into GMM1's gmm_v2 kernel

guowei-dev:
## [MoE] Fuse the MoE token permute (gather) into GMM1's gmm_v2 kernel

Adds `MOE_FUSE_PERMUTE`. Instead of materializing the expert-grouped activations with a standalone permute (ragged_gather / onehot matmul) before GMM1, `gmm_v2` gathers its LHS rows by index directly:

    gmm_v2(src, w1, gs, gather_indices=idx)   ==   gmm_v2(src[idx], w1, gs)

The permute is folded into the matmul's LHS read, as a result the grouped activations don't get written to / read back from HBM.

Correctness tested in `test_gather_matches_unfused`

Perf tested in `test_fused_permute_perf`:

| size_m | k | n | E | fused (us) | gather (us) | onehot (us) | fused/onehot |
|-------:|----:|----:|---:|------:|-------:|-------:|:---:|
|  4096 | 2048 | 1536 |  8 | 321.9 | 295.3 | 290.5 | 1.11 |
|  8192 | 4096 | 2048 |  8 | 544.5 | 743.3 | 632.4 | 0.86 |
| 16384 | 2048 |  768 | 16 | 469.8 | 461.6 | 655.3 | 0.72 |

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -92,6 +92,10 @@ def align_to(x, a):
     return pl.cdiv(x, a) * a
 
 
+def idx_window_len(tile_m: int) -> int:
+    return align_to(127 + tile_m, 128)
+
+
 # Define data classes.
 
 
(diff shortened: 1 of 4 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -892,6 +1023,96 @@ def kernel_main(
...
+    body = functools.partial(
+        gathered_inner_kernel,
+        cfgs=cfgs,
+        lhs_hbm=lhs_ref,
+        gather_idx_hbm=gather_idx_hbm,
+    )
+
+    pipeline_fn = pltpu.emit_pipeline(
+        body,
+        grid=(num_n, num_gm, num_k),
```
BirdsOfAFthr, line 1089: Is the idea of using `n` as the outermost dim in `emit_pipeline` to prefetch weights along the middle dimension `gm` of the grid?

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -616,6 +620,133 @@ def matmul_last():
...
+                lhs_hbm.at[row, pl.ds(k_base, tile_k)],
+                buf_slot.at[r],
+                sem,
+            ).start()
+
+    def wait_gather(buf_slot, sem):
+        for r in range(tile_m):
+            # Only the dst shape + sem matter for the wait; addresses are ignored.
+            pltpu.make_async_copy(
+                lhs_hbm.at[0, pl.ds(0, tile_k)],
```
BirdsOfAFthr, line 703: Is this just a dummy reference to just meet the api requirement?

tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -616,6 +620,133 @@ def matmul_last():
...
+        # the tile) and use `delta` to re-index the per-row reads.
+        aligned_start = (m_offset // 128) * 128
+        delta = m_offset - aligned_start
+        pltpu.make_async_copy(gather_idx_hbm.at[pl.ds(aligned_start, win_len)],
+                              gather_idx_smem, gather_idx_sem.at[0]).start()
+        pltpu.make_async_copy(gather_idx_hbm.at[pl.ds(aligned_start, win_len)],
+                              gather_idx_smem, gather_idx_sem.at[0]).wait()
+        k_base = k * tile_k
+        for r in range(tile_m):
+            pos = jnp.clip(delta + r, 0, win_len - 1)
```
BirdsOfAFthr, line 691: How to make sure we are reading the correct row if the `pos` is derived from clipping?

tpu_inference/kernels/megablox/gmm_v2.py
guowei-dev: The grid is the existing kernel_main layout and unchanged here. Outermost n means that the LHS is re-gathered per n-tile, which is also why the gather is only taken at num_n == 1 (else fallback to explicit permute).

tpu_inference/kernels/megablox/gmm_v2.py
guowei-dev: Yes, that's just a placeholder to satisfy the API. .wait() matches on dst-shape + semaphore and source is ignored.

tpu_inference/kernels/megablox/gmm_v2.py
guowei-dev: The clamp is purely defensive and never actually triggers for valid rows. Since we load 128-aligned windows, the exact index we need is always delta + r. Also the win_len is constructed to be large enough (align_to(127 + tile_m, 128)) to guarantee delta + r (≤ 126 + tile_m) is always < win_len. For edge cases at the end of the tensor, the HBM array is padded to prevent out-of-bounds DMA reads, and any invalid rows computed are masked out downstream anyway.

PR #2919, opened 2026-06-17
Refactor DFlash to support cross-request paged KV caching and dynamic batching

huybro:
**Description**
This PR refactors the DFlash speculator/proposer and model layers to use the JAX/TPU framework's global paged kv_caches and block_tables instead of a local dense cache. It also vectorizes input preparation and model forward calls to support dynamic multi-request batching (num_reqs > 1).

**Why is this change being made / Context**
Previously, DFlash maintained a local, contiguous array for its KV cache, allocated as (1, num_heads, max_kv_len, head_dim). This design limited DFlash to a batch size of 1 (num_reqs == 1) and prevented cache sharing or persistence across different requests. When the inference runner swapped requests (e.g., due to continuous batching swaps or preemption), DFlash had to wipe its local state (resetting _ctx_len and _cache_len to 0) and reconstruct the context from scratch on the next forward pass.

**Why this is a good solution**
By refactoring DFlash to use the global paged memory structure already managed by the runner:

Multi-request dynamic batching is fully enabled, allowing the speculator to run concurrently across batched sequences.
Cross-request cache persistence is achieved naturally, as the framework's page-level memory manager handles swapping and page mapping across request IDs.
Stateless proposer: It eliminates fragile manual state tracking (_ctx_len, _cache_len, _prev_seq_len) in the proposer, reducing edge-case bugs around cache cropping and recompute.

**Implementation Details**
Proposer (`dflash.py` (spec_decode)): Removed all local state-tracking variables. prepare_inputs now handles dynamic batch sizes, computing target hidden states, and vectorizing noise block construction (_build_noise_block). It propagates the global block_tables as part of the attention metadata.

Model & Attention (`dflash.py` (models)): Modified DFlashForCausalLM and layers to accept the paged kv_caches and target token metadata, passing them through to the attention layers. Registered DFlashForCausalLM in model_loader.py.

Attention Kernel (`DFlashAttention`): Replaced the dense lax.dynamic_update_slice and flash_attention logic with a two-step execution using the optimized ragged_paged_attention kernel:

1. Writes the target accepted tokens to the paged KV cache (using a causal mask to prevent attention leakage).
2. Writes the draft noise tokens to the cache, performs a non-causal attention pass over the full context, and computes logits.

**Shortcomings & Future Improvements**
Currently, the two-step paged attention requires two successive invocations of ragged_paged_attention per forward pass because the first step writes the target tokens under a causal mask, and the second step writes and attends to the draft tokens non-causally. Combining these steps into a single unified pass if/when supported by the kernel could yield further speedups.

**Tests**
I have verified the correctness and batched behavior using the following:

Unit & Correctness Tests

1. Updated test_dflash.py to match the new batched tensor layouts.
2. Removed the obsolete stateful flow test (test_dflash_proposer_multi_iteration_state_flow) since DFlash is now stateless.
3. Added test_build_noise_block_batched to verify vectorized noise block construction across multiple requests.

Checklist

### DIFF
--- a/tests/models/jax/test_dflash.py
+++ b/tests/models/jax/test_dflash.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Unit tests for the DFlash speculative decoding draft model on JAX/TPU."""
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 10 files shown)


Lumosis:
This PR is relatively large. Could you break it down?

PR #2939, opened 2026-06-19
[CI] Split spec-decoding e2e into per-push BVT subset + nightly full matrix

QiliangCui2023:
## Problem
The `E2E speculative decoding test` step runs the whole `tests/e2e/test_speculative_decoding.py` (25 cases) on **every push**. The suite has grown ~4x over the past month as models/cases were added (gemma-4 MTP, eagle3 logprobs, async/DP variants), so a single run now takes ~60 min on the single-chip queue and is a major consumer of `tpu_v7x_2_queue` capacity during the day.

## Change
Introduce a minimal per-push **BVT** (smoke) subset that still covers each method (ngram, eagle3, mtp) for both **correctness** and **performance**, and spreads the `(async_scheduling, enable_dp_attention)` combinations across those cases so each flag value is still exercised. The full matrix continues to run **nightly**.

Selection is driven by the `BVT_ONLY` env var (opt-in: filtering happens **only** when it is set, so the default everywhere is "run all"):
- `tests/conftest.py` registers a `bvt` marker and a `pytest_collection_modifyitems` hook. When `BVT_ONLY=1`, for any module that declares at least one `@pytest.mark.bvt` item it keeps only the bvt-marked items from that module (modules without the marker are left untouched, so it can never empty an unrelated suite). When `BVT_ONLY` is unset, nothing is filtered. The partition logic is a pure helper (`partition_bvt_only`) so it can be unit-tested directly.
- Only the 8 chosen cases are marked — no test bodies change.
- `run_in_docker.sh` forwards `BVT_ONLY` into the container.
- `pipeline_jax.yml`: a single spec-decoding step computes `BVT_ONLY=1` iff this is **not** a nightly build (`if NIGHTLY != 1`), so PR builds run the 8-case subset and nightly (`NIGHTLY=1`) runs the full 25 — one step, one `test_6` key. No other build sets `BVT_ONLY`, so feature support-matrix and release-tag builds run the full set unchanged.
- `tests/test_nightly_gate.py` unit-tests the gate (default = nothing filtered; `BVT_ONLY=1` narrows a marked module without touching others).

## BVT subset (8 cases, ~21 min vs ~60 min full)
| case | covers |
|---|---|
| `ngram_correctness_greedy` | ngram correctness, (F,F) |
| `ngram_performance_greedy` | ngram perf, batch=1 |
| `eagle3_correctness[F,F]` | eagle3 correctness + Llama-3-8B baseline |
| `eagle3_correctness[T,T]` | eagle3 correctness, async=T + dp=T |
| `eagle3_performance[20,T,T]` | eagle3 perf, async=T + dp=T, batch=20 |
| `mtp_correctness[F,F]` | mtp correctness + Qwen3.5-4B baseline |
| `mtp_correctness[T,T]` | mtp correctness, async=T + dp=T |
| `mtp_performance[20,T,F]` | mtp perf, async=T, dp=F |

Flag coverage across the subset: `(F,F)`, `async=T`, `dp=T`, `(T,T)`, `(T,F)`, batch sizes 1 and 20. The exotic `(async=F, dp=T)` combo and all gemma-4 cases run nightly.

## Verification
`pytest --collect-only` on the test file:
- default (no `BVT_ONLY`): **25** cases selected (full matrix)
- `BVT_ONLY=1`: **8** cases selected (the table above)
- collecting another test file alongside it is unaffected (the hook only narrows the bvt-declaring module, and only when `BVT_ONLY=1`).

`tests/test_nightly_gate.py` passes against the real conftest.

---

## Test evidence — nightly runs the full matrix

Verified on a nightly build of this branch (`NIGHTLY=1`):
**https://buildkite.com/tpu-commons/tpu-inference-ci/builds/20561**

The spec-decoding step run with `NIGHTLY=1` (`tpu7x E2E speculative decoding test`) **passed with all 25 cases** (`25 passed ... in 3504s`, `collected 25`) — i.e. with `BVT_ONLY` unset the gate runs the whole matrix, not the 8-case subset:

| family | count |
|---|---|
| ngram | 4 (correctness greedy/random, performance greedy/random) |
| eagle3 | 9 (correctness ×4, performance ×3, logprobs ×2) |
| mtp | 6 (correctness ×3, performance ×3) |
| gemma4_mtp | 6 (correctness ×3, performance ×3) |

The feature support-matrix steps also ran the full set and passed on both TPU generations:
`tpu7x`/`tpu6e` `Speculative_Decoding-_Eagle3` (correctness + performance) and `Speculative_Decoding-_Ngram` (correctness + performance) — confirming non-`BVT_ONLY` consumers of this file are unaffected.


QiliangCui2023:
### Heads-up: the marker gates more than the `pipeline_jax.yml` step — and a proposal to make the default fail-safe

While validating this change I found that the `pytest_collection_modifyitems` hook in `tests/conftest.py` keys off the `bvt` marker **module-wide**, so it narrows *any* collection run that pulls from `tests/e2e/test_speculative_decoding.py` while `NIGHTLY != 1` — not just the intended `- bvt` step in `pipeline_jax.yml`.

#### Why
The hook builds `bvt_modules` from the collected items of the current invocation; if those items include ≥1 `bvt`-marked item, every non-`bvt` item in that module is deselected. Two existing pipelines invoke specific functions/params of this file and therefore get narrowed too:

- `.buildkite/features/Speculative_Decoding-_Eagle3.yml` → `pytest ...::test_eagle3_correctness` (4 params) and `::test_eagle3_performance` (3 params)
- `.buildkite/features/Speculative_Decoding-_Ngram.yml` → `::test_ngram_correctness_greedy` + `::test_ngram_correctness_random` (and the perf pair)

Minimal repro of the hook against a module with the same marker layout (no `NIGHTLY`):

```
$ pytest test_spec.py::test_eagle3_correctness --collect-only -q
test_spec.py::test_eagle3_correctness[False-False]
1/4 tests collected (3 deselected)

$ pytest test_spec.py::test_ngram_correctness_greedy test_spec.py::test_ngram_correctness_random --collect-only -q
test_spec.py::test_ngram_correctness_greedy
1/2 tests collected (1 deselected)
```

So the eagle3 correctness feature step collapses 4→1, ngram 2→1, etc.

#### Scope of impact (limited)
Those feature pipelines are uploaded only by `upload_models_and_features.sh`, which runs only inside `nightly_verify.yml`, gated `if: build.env("NIGHTLY") == "1" || build.tag =~ /<release tag>/`:

- **Scheduled nightly** — `NIGHTLY=1` is set at the build level (it's what the `if` reads) and is present in the step env, so `run_in_docker.sh` forwards it and the hook returns early → full matrix. ✅ unaffected
- **PR CI** — does not run these pipelines at all. ✅ unaffected
- **Release-tag builds** — the group runs via the `|| build.tag` arm, but `NIGHTLY` is unset, so the hook filters → the committed support matrix under-reports spec-decode coverage (eagle3 4→1, ngram 2→1). ⚠️ doc artifact only

#### Proposal: invert the marker `bvt` → `nightly_only`
Drop the marked (heavy) cases per-push instead of keeping only the marked (smoke) ones:

```python
def pytest_collection_modifyitems(config, items):
    """On nightly run everything; otherwise drop @pytest.mark.nightly_only cases.
    Unmarked tests always run, so new tests are never silently skipped on PRs."""
    if os.environ.get("NIGHTLY") == "1":
        return
    drop = [it for it in items if it.get_closest_marker("nightly_only")]
    if not drop:
        return
    config.hook.pytest_deselected(items=drop)
    items[:] = [it for it in items if not it.get_closest_marker("nightly_only")]
```

Then tag the ~19 heavy cases `@pytest.mark.nightly_only` — function-level where every param is nightly (gemma-4 corr/perf, both eagle3 logprobs, ngram random), per-`pytest.param` for the mixed eagle3/mtp lists. Same per-push savings (~16 vs ~60 min).

**Why it's better — fail-safe default.** With `bvt` (keep-marked), a newly added or unmarked spec-decode test is silently deselected on every PR and easy to miss. With `nightly_only` (drop-marked), unmarked tests always run; only explicitly tagged heavy cases are gated. Forgetting a mark costs a little time, never coverage. It also keeps local `pytest <file>` intuitive — the only things deselected are the ones literally tagged `nightly_only`.

**Caveat:** the inversion keeps the same `NIGHTLY` trigger, so it does **not** by itself change the release-tag residual above — a tag build still drops the marked cases. That residual is minor (support-matrix doc on releases only). To close it fully without adding a new env var, confirm release-tag builds set `NIGHTLY=1`; otherwise it's reasonable to accept.

#### Also
Worth adding a small unit test for the collection hook (synthetic items in → asserted selection out) — it's the riskiest logic in this PR and is currently only verified manually via `--collect-only`.

.buildkite/pipeline_jax.yml
```diff
@@ -80,8 +80,29 @@ steps:
...
 
-      - label: "${TPU_VERSION:-tpu6e} E2E speculative decoding test"
-        key: ${TPU_VERSION:-tpu6e}_test_6
+      # Per-push smoke subset (BVT). Sets BVT_ONLY=1 so the conftest collection
+      # gate keeps only @pytest.mark.bvt cases. The full matrix runs on nightly
+      # via the "- nightly" step below, which leaves BVT_ONLY unset (default =
+      # run all). No other build sets BVT_ONLY, so feature/release-tag builds run
+      # the full set unchanged.
+      - label: "${TPU_VERSION:-tpu6e} E2E speculative decoding test - bvt"
+        key: ${TPU_VERSION:-tpu6e}_test_6_bvt
```
QiliangCui: if you change the key here, update - label: "${TPU_VERSION:-tpu6e} TPU Test Notification" as well.

.buildkite/pipeline_jax.yml
```diff
@@ -80,8 +80,29 @@ steps:
...
+          BVT_ONLY: "1"
+        agents:
+          queue: ${TPU_QUEUE_SINGLE:-tpu_v6e_queue}
+        commands:
+          - |
+              .buildkite/scripts/run_in_docker.sh \
+                bash -c 'python3 -m pytest -s -v -x /workspace/tpu_inference/tests/e2e/test_speculative_decoding.py'
+
+      - label: "${TPU_VERSION:-tpu6e} E2E speculative decoding test - nightly"
+        key: ${TPU_VERSION:-tpu6e}_test_6_nightly
```
QiliangCui: if you change the key here, update - label: "${TPU_VERSION:-tpu6e} TPU Test Notification" as well.

QiliangCui2023:
### Follow-up: the `BVT_ONLY` opt-in resolves the gating concern ✅

Re-reviewed at `d1f9a62`. Switching the gate from "filter unless `NIGHTLY=1`" to "filter only when `BVT_ONLY=1`" fully addresses the earlier concern — only the `- bvt` step sets `BVT_ONLY`, so every other consumer of `tests/e2e/test_speculative_decoding.py` runs the full set.

Verified end-to-end against the shipped `tests/conftest.py`:

| invocation | env | result |
|---|---|---|
| `::test_eagle3_correctness` (feature pipeline / release-tag path) | no `BVT_ONLY` | 4/4 ✅ |
| ngram greedy + random (feature pipeline) | no `BVT_ONLY` | 2/2 ✅ |
| whole file (nightly / local) | no `BVT_ONLY` | 25 ✅ |
| per-push step | `BVT_ONLY=1` | 6/25 (smoke subset) ✅ |

Feature support-matrix pipelines, release-tag builds, and local runs are all unaffected now; only the per-push step narrows.

`tests/test_nightly_gate.py` is a good add — extracting `partition_bvt_only` as a pure helper and asserting (a) default = nothing deselected and (b) `BVT_ONLY=1` narrows a marked module without emptying others directly encodes the two invariants that matter. The 4 tests pass against the real conftest.

#### One thing to fix before merge — the PR description is stale
The description still documents the previous `NIGHTLY`-driven mechanism, which is now inverted:
- "Selection is driven by the `NIGHTLY` env var" → it's now `BVT_ONLY`. `NIGHTLY` only selects *which step* runs; `BVT_ONLY` (set only by the `- bvt` step) selects *which cases*.
- "`run_in_docker.sh` forwards `NIGHTLY`" → it forwards `BVT_ONLY`.
- "default (no `NIGHTLY`): 6 cases / `NIGHTLY=1`: 25" → the default is reversed: **no `BVT_ONLY` → 25 (full)**, **`BVT_ONLY=1` → 6**.

#### Nits (non-blocking)
- Keep-marked polarity means a newly added *unmarked* case in `test_speculative_decoding.py` won't run on the per-push step (nightly only). Reasonable as a curated allowlist — worth a one-line note by the `bvt` cases so contributors remember to mark new smoke-worthy ones.
- `test_nightly_gate.py` docstring says "no TPU/jax execution involved" — accurate for execution, but importing the conftest does `import jax`, so it needs jax importable (fine in the jax CI step).

.buildkite/pipeline_jax.yml
QiliangCui2023: Done — the TPU Test Notification step now references both new keys (`test_6_bvt` and `test_6_nightly`) in its `depends_on` and in the `check_results.sh` argument list. Only one of the two runs per build; the skipped one returns `skipped` from `buildkite-agent step get`, which `check_results.sh` already treats as passed.

.buildkite/pipeline_jax.yml
QiliangCui2023: Done — the TPU Test Notification step now references both new keys (`test_6_bvt` and `test_6_nightly`) in its `depends_on` and in the `check_results.sh` argument list. Only one of the two runs per build; the skipped one returns `skipped` from `buildkite-agent step get`, which `check_results.sh` already treats as passed.

gxd3 (approved):

tests/e2e/test_speculative_decoding.py
```diff
@@ -430,7 +435,7 @@ def eagle3_baseline():
 @pytest.mark.parametrize(
     "async_scheduling, enable_dp_attention",
     [
-        (False, False),
+        pytest.param(False, False, marks=pytest.mark.bvt),
         (False, True),
         (True, False),
         (True, True),
```
gxd3: put (True, True) in presubmit test too?

tests/e2e/test_speculative_decoding.py
```diff
@@ -551,7 +557,7 @@ def mtp_baseline():
 @pytest.mark.parametrize(
     "async_scheduling, enable_dp_attention",
     [
-        (False, False),
+        pytest.param(False, False, marks=pytest.mark.bvt),
         (True, False),
         (True, True),
```
gxd3: put ((True, True) in presubmit test too?

tests/e2e/test_speculative_decoding.py
QiliangCui2023: Done — added `(True, True)` (async + dp) to the per-push BVT set as `pytest.param(True, True, marks=pytest.mark.bvt)`. Per-push now collects 8 cases (was 6); the baseline is already built by the `[F,F]` case and shared via the module-scoped fixture, so this adds only one extra spec-engine build (~2–3 min).

tests/e2e/test_speculative_decoding.py
QiliangCui2023: Done — added `(True, True)` (async + dp) to the per-push BVT set as `pytest.param(True, True, marks=pytest.mark.bvt)`. Per-push now collects 8 cases (was 6); the baseline is already built by the `[F,F]` case and shared via the module-scoped fixture, so this adds only one extra spec-engine build (~2–3 min).

QiliangCui (approved):

PR #2951, opened 2026-06-21
fix vLLM integration error and proceed lkg version

gxd3:
fix vLLM integration error and proceed lkg version

The error is due to method renaming from vLLM upstream.

For the lora changes:
 `async_tensor_h2d` no longer takes a `pin_memory` argument; vLLM now gates pinning on its module-level `PIN_MEMORY` constant, which already resolves to False on TPU, so no patching is needed.


QiliangCui (approved):

tpu_inference/lora/torch_punica_tpu.py
```diff
@@ -20,10 +17,9 @@
 
 from tpu_inference.lora.torch_lora_ops import bgmv_expand_slice, bgmv_shrink
 
-# Pin memory is unsupported on TPU; force `pin_memory` off for
-# `async_tensor_h2d` calls.
-_punica_utils.async_tensor_h2d = functools.partial(
-    _punica_utils.async_tensor_h2d, pin_memory=is_pin_memory_available())
+# NOTE: `async_tensor_h2d` no longer takes a `pin_memory` argument; vLLM now
+# gates pinning on its module-level `PIN_MEMORY` constant, which already
+# resolves to False on TPU, so no patching is needed here.
```
QiliangCui: maybe this comment is not needed if the code is deleted.

QiliangCui: Maybe moving this comment to PR description is good enough for future reference.

gxd3: Updated.

vanbasten23 (approved):

mrjunwan-lang (approved):

PR #2958, opened 2026-06-22
Profile per concurrency per phase

BirdsOfAFthr:
# Description

This PR updates `PhasedBasedProfiler` to support profiling and capturing separate traces for the same phase across multiple concurrency sizes during benchmark sweeps, while maintaining full backward compatibility.

### Why is this change being made?
Previously, `PhasedBasedProfiler` tracked profiled phases globally via a simple boolean map (`self.inference_phase_seen`). Once a phase (like `DECODE_ONLY`) was profiled once (typically during warm-up or the first sweep iteration), the profiler would never trigger trace collection for that phase again. During concurrency sweeps (e.g., testing batch sizes of 1, 2, 4, 8, 16, 32, 64 and 128), we want to profile the compile shapes and runtime traces for every concurrency size individually. Without this change, we had to start the server again for each concurrency.

### Specific Implementation:
1.  **Concurrency-Aware Phase Tracking**: Refactored the tracking map from a global dictionary of boolean flags to a set `self.inference_phases_profiled` containing `(phase, concurrency)` tuples.
2.  **Dynamic Directory Suffixes**: Suffixes the phase output directory with `_concur_{concurrency}` (e.g., `decode_only_concur_64/`) to prevent concurrency runs from overwriting each other's trace files.
3.  **Strict Backward Compatibility**: If the concurrency size `num_reqs` is absent or `0` (e.g. when executing standard, non-sweep profiling workloads):
    *   It falls back to tracking by `phase` only.
    *   It keeps the original directory name format (no `_concur` suffix).
    *   It re-enables the original "all phases seen" shutdown gate to terminate profiling after seeing every phase once.
---

# Tests

Tested this change in two scenarios on a 16-chip `v7x-16` cluster running Mistral 3 Large (FP8):

1.  **Multi-concurrency Sweep Run**:
    *   Ran a sweep with `CONCURRENCIES=[1, 2, 4, 8, 16, 32, 64, 128]`.
    *   Verified that distinct trace outputs were successfully written to separate folders (e.g., `decode_only_concur_16`, `decode_only_concur_32`, `decode_only_concur_64`).
2.  **Backward Compatibility Check**:
    *   Ran a benchmark without the `max_concurrency` dictionary key.
    *   Verified that traces compiled directly into `decode_only` and `prefill_heavy` without suffixes, and that the profiler shut down correctly once all phases were profiled.
    * list all profiles under default mode:
```
amandaliang@amandaliang:~/tpu-inference$ gcloud compute tpus tpu-vm ssh amandaliang-v7x-8 \
  --zone=us-central1-c \
  --project=cloud-tpu-inference-test \
  --command="find /mnt/pd/vllm-xprof/ -name '*.xplane.pb' | sort"
Using ssh batch size of 1. Attempting to SSH into 1 nodes with a total of 1 workers.
SSH: Attempting to connect to worker 0...
/mnt/pd/vllm-xprof/decode_only/plugins/profile/2026_07_22_00_49_43/t1v-n-9eb622fc-w-0.xplane.pb
/mnt/pd/vllm-xprof/prefill_heavy/plugins/profile/2026_07_22_00_48_23/t1v-n-9eb622fc-w-0.xplane.pb
/mnt/pd/vllm-xprof/prefill_only/plugins/profile/2026_07_22_00_45_14/t1v-n-9eb622fc-w-0.xplane.pb
```

---

# Checklist

### DIFF
--- a/tests/runner/test_utils.py
+++ b/tests/runner/test_utils.py
@@ -528,7 +528,8 @@ def test_phased_profiler_full_cycle(profiler_fixture):
     mock_start.assert_called_once()
     assert profiler.profiling_n_steps_left == PHASED_PROFILER_NUM_STEPS_TO_PROFILE_FOR
     assert profiler.current_phase == "prefill_heavy_concur_2"
-    assert (InferencePhase.PREFILL_HEAVY, 2) in profiler.inference_phases_profiled
+    assert (InferencePhase.PREFILL_HEAVY,
+            2) in profiler.inference_phases_profiled
     assert mock_file().write.call_count == 1  # Wrote stats on start
 
     # 2. Step profiling (N-1 steps)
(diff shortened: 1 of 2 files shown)


jrplatin:
Synced offline, but I don't believe we should have tracing for each concurrency be the default -- it will produce too many artifacts

jrplatin (approved):

PR #2965, opened 2026-06-23
[Bugfix] Fix routed-experts slot mapping for prompt tokens

QiliangCui2023:
## Summary

When MoE routed experts are returned (`enable_return_routed_experts=True`), per-token expert IDs are stored into the scheduler-side slot buffer (`RoutedExpertsManager`) keyed by physical KV-cache slot and read back block-relative from position 0. `_reconstruct_routed_experts` must place each request's tokens at their true absolute positions, i.e. the chunk's start = the request's **pre-step** `num_computed_tokens`.

It derived that start from `runner.requests[req_id].num_computed_tokens` — a per-request field that `update_states` **rewrites in place every step**. Reconstruction runs at a path-dependent time, so it read different values:

- **async path** (the default): reconstruction is deferred to `AsyncTPUModelRunnerOutput.get_output`, which runs *after* the next step's `update_states` has already advanced the field → it reads the **post-step** value. The prompt's experts were therefore stored `num_tokens` slots too far forward, and the scheduler read those slots back as zeros → **routed experts for prompt tokens came back zero-filled**.
- **sync path** reads the same field pre-step, so it was correct on main.

Root cause: deferred/async-reachable code read a **per-step-mutated field** (`runner.requests[...]`) instead of a per-step snapshot.

The bug was masked because the standalone and batched runs produced the same (wrong) prompt experts; it surfaced via `test_moe_expert_ids_batch_isolation` after the vLLM block-allocation change vllm-project/vllm#42656 changed the batched block layout and broke that symmetry.

## Fix

Source the chunk start from the per-step, immutable **`scheduler_output`** (`scheduled_new_reqs` / `scheduled_cached_reqs` `num_computed_tokens` — always the pre-step value) instead of the live `runner.requests` field. A fresh `scheduler_output` is created each step and never mutated, so it yields the correct start **regardless of when reconstruction runs** — making the async path (the reported bug) and the sync path agree. This follows the snapshot-don't-read-live-state pattern the rest of `AsyncTPUModelRunnerOutput` already uses (it captures `scheduler_output`, `next_tokens`, etc.). A non-negativity assert guards against a future regression producing silently-wrong slots.

Note: the sync path is required by `enable_continue_decode` (`async_scheduling=False`). A naive `num_computed_tokens - num_tokens` correction would fix async but underflow to a **negative** start on the sync path's first prefill (engine crash), so sourcing the pre-step value directly is what makes both paths correct at once.

For a 5-token prompt in block 1 (block_size 16), the slots are now `[16..20]` (positions 0–4) on both paths — matching the scheduler-side read — instead of `[21..25]` (async, zero-filled on read).

## Testing

- **Unit** (`tests/runner/test_routed_experts_slots.py`): `_reconstruct_slots_for_request` slot math (block-relative, multi-block, decode offset, empty, negative-start guard) and `_reconstruct_routed_experts` driven from `scheduler_output` for new-request and cached-request chunk starts (a sentinel `num_computed_tokens` asserts that field is no longer used). 7 passed.
- **TPU e2e** (`tests/runner/test_moe_expert_ids.py`): `test_moe_expert_ids_batch_isolation` (async) passes; new `test_moe_expert_ids_sync_path_prompt_populated` (`async_scheduling=False` + continue_decode) covers the sync path. 2 passed on tpu7x.
- **e2e** (`tests/e2e/test_continue_decode.py`): routed-experts value check against the single-step baseline, scoped to the deterministic prompt slice (avoids false-fails on generation divergence).
- Verified on tpu7x: the sync repro no longer crashes and returns populated prompt experts.

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -367,20 +367,29 @@ def _reconstruct_slots_for_request(
     req_state: CachedRequestState,
     num_tokens: int,
     block_size: int,
+    start_pos: int,
 ) -> np.ndarray:
-    """Reconstructs physical slot mappings for a request using vectorized NumPy."""
-    # num_computed_tokens has already been advanced past this step's tokens by
-    # the time routed experts are reconstructed, so this step's tokens occupy
-    # absolute positions [num_computed_tokens - num_tokens, num_computed_tokens).
-    # Using num_computed_tokens directly shifts the slots forward by num_tokens,
-    # which misaligns with the scheduler-side slot read (RoutedExpertsManager.get,
-    # block-relative from position 0) and yields zero-filled routed experts for
-    # the prompt tokens.
-    start_pos = req_state.num_computed_tokens - num_tokens
-    block_ids = req_state.block_ids[0] if req_state.block_ids else []
+    """Reconstructs physical KV-cache slots for ``num_tokens`` tokens of a
+    request using vectorized NumPy.
 
+    ``start_pos`` is the absolute sequence position of the first of these
+    tokens and is supplied by the caller, because the two call sites have
+    different token contracts: the routed-experts reconstruction passes a
+    single scheduler chunk (whose tokens end at ``num_computed_tokens``, so it
+    passes ``num_computed_tokens - num_tokens``), while the continue_decode
+    path passes an on-device decode burst (not reflected in
+    ``num_computed_tokens``, so it passes ``num_computed_tokens``). The slots
+    must match the scheduler-side read (``RoutedExpertsManager.get``, which is
+    block-relative from position 0).
+    """
     if num_tokens <= 0:
         return np.array([], dtype=np.int32)
+
+    assert start_pos >= 0, (
+        f"[routed-experts] start_pos must be non-negative, got {start_pos} "
+        f"(num_tokens={num_tokens}); slots would be wrong via numpy negative "
+        f"indexing")
+    block_ids = req_state.block_ids[0] if req_state.block_ids else []
 
     pos = np.arange(start_pos, start_pos + num_tokens, dtype=np.int32)
     block_idx = pos // block_size
@@ -446,12 +455,20 @@ def _reconstruct_routed_experts(
                                                              dp_end,
                                                              dtype=np.int32)
 
-            # Reconstruct slots for this request using vectorized NumPy
+            # Reconstruct slots for this request using vectorized NumPy.
+            # num_computed_tokens has already been advanced past this chunk, so
+            # the chunk's tokens start at num_computed_tokens - n. Passing
+            # num_computed_tokens directly would shift slots forward by n and
+            # misalign with the scheduler-side block-relative read, yielding
+            # zero-filled routed experts for the prompt tokens.
             req_state = runner.requests[req_id]
             if n > 0:
                 global_slots[
                     global_start:global_end] = _reconstruct_slots_for_request(
-                        req_state, n, block_size)
+                        req_state,
+                        n,
+                        block_size,
+                        start_pos=req_state.num_computed_tokens - n)
 
     # 3. Perform global rank reordering and transpose in a single fancy indexing sweep!
     expert_indices_reordered = expert_indices_cpu[:, indices_map, :].transpose(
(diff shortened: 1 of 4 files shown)


pv97 (approved):

tpu_inference/runner/tpu_runner.py
```diff
@@ -369,7 +369,14 @@ def _reconstruct_slots_for_request(
...
     """Reconstructs physical slot mappings for a request using vectorized NumPy."""
-    start_pos = req_state.num_computed_tokens
+    # num_computed_tokens has already been advanced past this step's tokens by
+    # the time routed experts are reconstructed, so this step's tokens occupy
+    # absolute positions [num_computed_tokens - num_tokens, num_computed_tokens).
+    # Using num_computed_tokens directly shifts the slots forward by num_tokens,
+    # which misaligns with the scheduler-side slot read (RoutedExpertsManager.get,
+    # block-relative from position 0) and yields zero-filled routed experts for
+    # the prompt tokens.
+    start_pos = req_state.num_computed_tokens - num_tokens
```
QiliangCui2023: **`_reconstruct_slots_for_request` is shared by two callers with different `num_tokens` contracts — this fix is correct for one, but likely wrong for the other.**

The fix is right for the `_reconstruct_routed_experts` callers (async `get_output` and the sync path), where `num_tokens = num_scheduled_tokens[req_id]` is a single scheduler chunk and `num_computed_tokens` has indeed been advanced past it. I confirmed the direction against the consumer `RoutedExpertsManager.get`, which reads block-relative from position 0, so the first prefill must store at `start_pos = num_computed_tokens - num_tokens = 0`. 👍

But the same helper is also called from `_execute_continue_decode` (`tpu_inference/runner/tpu_runner.py:1490`) with `num_tokens = actual_len` — the number of tokens produced by the multi-step on-device decode burst, whose length is decided **on-device via EOS early-exit**. The scheduler cannot have advanced `num_computed_tokens` by `actual_len`, so the invariant this comment encodes ("`num_computed_tokens` advanced past this step's tokens") does not hold there:
- for a burst longer than one token, `start_pos` is shifted by `actual_len - 1`;
- for a short prompt, `num_computed_tokens - actual_len` can go **negative**, which silently produces wrong slots via numpy negative indexing (no guard).

Suggest decoupling the two semantics by passing an explicit `start_pos` from each caller instead of deriving it inside the helper — `_execute_continue_decode` already has the burst's initial `input_positions`. (Note: I could not run this on TPU, so treat the continue_decode case as a risk to verify rather than a confirmed break; for `actual_len == 1` the fix is fine.)

QiliangCui2023 (commented):
Thanks for the fix — the root-cause and the default-path correctness check out, and I verified the store/read alignment against vLLM's `RoutedExpertsManager.get` (reads block-relative from position 0). One inline comment on a shared-helper concern, plus a test-coverage gap below.

---

**The continue_decode routed-experts test only checks shape, not values — it cannot catch a slot-mapping regression.**

`tests/e2e/test_continue_decode.py:281-299` is the only test that runs `enable_continue_decode=True` together with `enable_return_routed_experts=True`, but its MoE assertions only verify `routed_experts` is non-None and that `shape[0] == prompt + gen - 1`. That token dimension equals `num_tokens` regardless of `start_pos`, and `_check_correctness` (around L69) compares only generated text + logprobs — so the actual routed-expert **IDs are never validated**. A wrong `start_pos` in the `_execute_continue_decode` call (see the inline comment on `tpu_runner.py`) would keep this test green.

Suggest adding a value-level check there: compare the continue_decode routed experts against the single-step `baseline_outputs` the test already produces — the same standalone-vs-optimized comparison that `tests/runner/test_moe_expert_ids.py::TestMoEExpertIds::test_moe_expert_ids_batch_isolation` does.

QiliangCui2023:
Addressed both review points in e67b75ca9.

**Inline comment (shared helper / continue_decode contract).** You're right — `_reconstruct_slots_for_request` is shared by two callers with different `num_tokens` contracts, and deriving `start_pos` inside the helper was correct for the routed-experts reconstruction but wrong for `_execute_continue_decode` (where `actual_len` is an on-device EOS-trimmed burst not reflected in `num_computed_tokens`, so `num_computed_tokens - actual_len` could shift the slots or go negative). Decoupled it: the helper now takes an explicit `start_pos`.
- routed-experts reconstruction passes `num_computed_tokens - num_tokens` (the fix);
- `_execute_continue_decode` passes `num_computed_tokens` (its original, unchanged behavior).
Also added `assert start_pos >= 0` so the negative-index case fails loudly instead of silently producing wrong slots.

This was empirically confirmed: the existing unit test `tests/runner/test_tpu_runner.py::TestTPUJaxRunner::test_execute_continue_decode_with_routed_experts` (which asserts continue_decode `slot_mapping` values) failed with the in-helper version and passes again after the decouple. `test_moe_expert_ids_batch_isolation` still passes on TPU.

**Test-coverage comment.** 
- Unit: `tests/runner/test_routed_experts_slots.py` now covers the helper with explicit `start_pos` (block-relative, multi-block, decode offset, empty, and a negative-`start_pos` guard) plus an integration test of `_reconstruct_routed_experts` asserting the prefill `slot_mapping` is block-relative from position 0 (the off-by-`num_tokens` site).
- E2E: `tests/e2e/test_continue_decode.py` now compares continue_decode routed experts against the single-step baseline at value level (not shape-only), per your suggestion.

Note: the continue_decode slot path already had value-level coverage via `test_execute_continue_decode_with_routed_experts` — that's the test that caught the original regression.

kyuyeunk (approved):

PR #2967, opened 2026-06-23
Add split-input fused FFT benchmark

csgoogle:
## Summary

Add an experimental `fft2d` TPU kernel using Pallas.

XLA’s complex FFT path splits work into real and imaginary components, then repacks them at the complex API boundary. This conversion adds latency. The new kernel provides:

- `fft2d(complex64)`: convenience complex-to-complex API.
- `fft2d_split(real, imag)`: accepts and returns separate float32 real/imaginary planes, avoiding complex packing and unpacking for callers that already use split representations.

The kernel evaluates the fused dense DFT `W_col @ X @ W_row`, keeping the intermediate in VMEM.

## TPU7x results

Measured on 8× TPU7x, input shape `100 x 1024 x 1024`, `complex64`.

Tiles: `col_tile=512`, `row_k_tile=1024`, `col_k_tile=1024`

```bash
LIBTPU_INIT_ARGS='--xla_tpu_dvfs_p_state=7 --xla_tpu_scoped_vmem_limit_kib=65536'
```

| Implementation | Latency | Improvement vs JAX |
| --- | ---: | ---: |
| `jax.numpy.fft2` | 23.23 ms | baseline |
| `fft2d(complex)` | 20.93 ms | ~10% lower latency |
| `fft2d_split(real, imag)` | 7.36 ms | ~68% lower latency / 3.16× faster |

The split-input/output path is also 2.85× faster than the complex-to-complex wrapper.

## Validation

- Relative error vs `jax.numpy.fft2`: `5.25e-7`

### DIFF
--- a/tpu_inference/kernels/experimental/fft/fft2d.py
+++ b/tpu_inference/kernels/experimental/fft/fft2d.py
@@ -1,0 +1,466 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


tpu_inference/kernels/experimental/fft/fft2d.py
kyuyeunk, line 1: i think this can be a single file without a folder?

PR #2974, opened 2026-06-23
Remove jax.clear_caches() from assign_and_shard_param to avoid re-compilation and slow weight loading

Lumosis:
# Description

jax.clear_caches() clear all compilation and staging caches. This will trigger re-compilation and slow down weight loading process.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/models/jax/utils/weight_utils.py
+++ b/tpu_inference/models/jax/utils/weight_utils.py
@@ -788,7 +788,6 @@ def assign_and_shard_param(jax_param: nnx.Param,
         jax_param.set_value(shard_put(jax_weight, spec, mesh=param_mesh))
         jax_param.set_metadata("_is_loaded", True)
         del jax_weight
-        jax.clear_caches()
     except Exception as e:
         raise RuntimeError(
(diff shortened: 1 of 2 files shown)


tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -788,7 +788,6 @@ def assign_and_shard_param(jax_param: nnx.Param,
         jax_param.set_value(shard_put(jax_weight, spec, mesh=param_mesh))
         jax_param.set_metadata("_is_loaded", True)
         del jax_weight
-        jax.clear_caches()
```
lk-chen, line 791: Does it make sense to move to between these 2 lines https://github.com/vllm-project/tpu-inference/blob/5038df73ff2db27133b755be97fa314a4c0180c0/tpu_inference/models/jax/utils/weight_utils.py#L983-L984

Context of this line: we used to observe OOM during weight loading due to compile cache, so decided to clear cache here

Lumosis: Done

lk-chen (approved):

PR #2980, opened 2026-06-24
[DSv4] implement DeepseekV4Attention to connect pieces together

gxd3:
# Description

* implement `VllmDeepseekV4MLAAttention`, that instantiates and calls compressor, indexer, and the core-attention. The logic should mirror vLLM upstream's behavior.

* Add `VllmDeepseekCompressor`, which calls into the jax code we had. Move the projection part into `VllmDeepseekV4MLAAttention`'s attn_gemm() to keep it same as vLLM upstream. Inline save_partial_states() into call site, delete "tpu_inference/kernels/experimental/deepseek_v4/compress_store.py".

* For `VllmDeepseekV4Indexer`, its call to `streamindex_topk` is wrapped within jax.sharded-map so that it work with attn-dp. It calls VllmDeepseekCompressor.forward() instead of directly calling code in tpu_inference/kernels/experimental/deepseek_v4/compressor.py.

* sync Pallas kernel code from g3 for a few small fixes.


inucool12 (approved):
All the changes look good to me.

a1yssan13 (approved):

PR #2981, opened 2026-06-24
[DSv4] KV cache creation logic for DSv4

gxd3:
Add logics to handle hybrid KV cache creation for DeepSeek V4.

### DIFF
--- a/tpu_inference/runner/kv_cache_manager.py
+++ b/tpu_inference/runner/kv_cache_manager.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import dataclasses
+import os
 from typing import TYPE_CHECKING, List
 
 import jax
@@ -64,6 +65,10 @@ def is_cache_for_ds_v4(attn_module: AttentionLayerBase) -> bool:
         attn_module, DeepseekV4SWACache) or isinstance(
             attn_module, DeepseekV4Attention) or isinstance(
                 attn_module, CompressorStateCache)
+
+
+def is_ds_v4(vllm_config):
+    return "DeepseekV4ForCausalLM" in (vllm_config.model_config.architectures)
 
 
 class KVCacheManager:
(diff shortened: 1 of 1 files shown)


tpu_inference/runner/kv_cache_manager.py
```diff
@@ -872,13 +882,55 @@ def _allocate_mamba(c_shape=cache_shape,
...
+                            # DSV4 FP8 format is packed as unit8
+                            os.environ["MLA_KV_PACKING_SIZE"] = "4"
+
+                            contains_indexer_cache = any(
+                                "indexer" in layer
+                                for layer in kv_cache_tensor.shared_by)
+                            if contains_indexer_cache:
+                                # Indexer's compressor state cache kv cache must be overlay on
+                                # Indexer's compressed kv cache.
+                                # Main attn's compressor state cache kv cache and sliding windor
```
inucool12: nit: typo of window

gxd3: good catch! Fixed

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -452,20 +457,15 @@ def get_kv_cache_spec(self):
...
         if self.use_mla:
             # Individually pad the RopE and latents
-            if "DeepseekV4ForCausalLM" in (
-                    self.runner.vllm_config.model_config.architectures):
-                mla_head_size = common_utils.align_to(text_config.head_dim,
-                                                      128)
+            qk_rope_head_dim = getattr(text_config, "qk_rope_head_dim", 0)
+            padded_kv_lora_rank = common_utils.align_to(
+                getattr(text_config, "kv_lora_rank", 0), 128)
+            if tpu_envs.MLA_TRANSPOSE_KV_CACHE:
```
inucool12, line 463: I am wondering if it is safe to read `tpu_envs.MLA_TRANSPOSE_KV_CACHE` at this point if the environment variable is only being configured later during the allocation sequence. 
Shouldn't we move the configuration check to the top of `get_kv_cache_spec` to prevent a structural shape mismatch.

gxd3: Actually, this section of code https://screenshot.googleplex.com/8oDiJQwHaPX2Ty6 only take effect for DSv3.

For DSv4, here is where the spec is generated: https://screenshot.googleplex.com/8vNRDNBNwErPDxJ, mla_qk_rope_head_dim / mla_head_size are never used.

inucool12: Okay, I see.

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -872,13 +882,55 @@ def _allocate_mamba(c_shape=cache_shape,
...
+                            if mla_spec is not None:
+                                layer_spec = mla_spec
+                                block_size = layer_spec.storage_block_size
+                            if mla_spec is None:
+                                # Edge case handling: for DSV4 Flash,
+                                # There are 21 CSA layers, 43 SWA caches,
+                                # since 43 % 21 !=0, there is one SWA cache
+                                # not sharing a KV tensor with other caches.
+                                assert "swa_cache" in layer_name
+                                assert len(kv_cache_tensor.shared_by) == 1
```
inucool12, line 926: Just a minor scoping concern: seems like layer_name leaks from the outer loop. 

Wondering if it would be better to check shared_by[0] instead? Casting to a string avoids potential TypeErrors:

assert "swa_cache" in str(kv_cache_tensor.shared_by[0])

inucool12 (approved):
LGTM

mrjunwan-lang (approved):

tpu_inference/runner/kv_cache_manager.py
gxd3: sg, done.

PR #2984, opened 2026-06-24
Kernel Autotune E2E pipeline

patrickji2014:
# Description

This PR creates a kernel auto tuning pipeline, which does the following steps:
1. Invoke benchmark pipeline and collect kernel tuning case, save to spanner
   - Note that the way use to intercept the tuning cases is intended after evaluating other options like using sitecustomize.py, usercustomize.py or create an custom entrypoint etc. The code used to patch the tuned_params.py is small enough to be managable and we also validate the patched tuned_params.py to make sure it's not corrupted. Using others options might conflict with system wide python library and not obvious to other engineers. 
3. Invoke kernel tuning pipeline and tune the tuning case from step 1
4. Patch the kernel tuning result which is the optimal tunable params into different kernel's tuned_params.py. Create a new branch after patching
5. Invoke the benchmark pipeline on the step 3 created branch
6. Compare step 1 and step 4's benchmark metrics and prepare PR.

Context:
Existing kernel tuning pipeline is only suitable for manual tuning, where engineers collect the tuning cases, hardcode it and launch the tuning pipeline, collect tuning result, run benchmark, compare result and prepare PR. Create a auto tuning pipeline can save engineering effort and continuous discover better kernel tiling size as a long term solution.

This pipeline try to reuse the existing pipeline as much as possible by triggering benchmark and kernel tuning pipeline. Using Bayesian optimization for reducing searching space is will be supported in following PR.

Notes:
Some of the change are not optimal and TODO is put alongside just to make this PR manageable before it grows too large. Here is a list of them:
1. the get_search_space function inside the kernel_tuner_base is a placeholder and will be implemented/override by subclass in following pr.

# Tests

Test in BuildKite
Example Pipeline https://buildkite.com/tpu-commons/wip-tpu-inference-kernel-autotune/builds/118/canvas
Example PR Created https://github.com/vllm-project/tpu-inference/pull/3031
Example artifacts when PR not created https://buildkite.com/tpu-commons/wip-tpu-inference-kernel-autotune/builds/128/canvas?sid=019f07b7-c7c1-4132-9b74-aea92ccae5cb&tab=artifacts
# Checklist


tools/kernel/tuner/v1/autotune/autotune_result_processing.py
```diff
@@ -0,0 +1,688 @@
...
+                    verdict = 'BASELINE_ZERO'
+                else:
+                    if metric in lower_is_better_metrics:
+                        # Positive means better (lower latency after tuning).
+                        delta_pct = (baseline - tuned) / baseline
+                    else:
+                        # Positive means better (higher throughput after tuning).
+                        delta_pct = (tuned - baseline) / baseline
+
+                    improved = delta_pct > 0
```
QiliangCui2023: **Improvement gate ignores `--metric_improvement_threshold` → auto-PRs on benchmark noise**

`improved = delta_pct > 0` uses no threshold, while the regression check on the next line uses `delta_pct < -threshold`. So the `--metric_improvement_threshold` flag (default 0.4%) — named an *improvement* threshold — is only ever applied to regressions, never to improvements.

Consequence: `should_create_pr = monitor_improved and not has_regression and not hard_blocker` fires whenever a monitored metric (Throughput / MedianITL) is better by *any* margin — including run-to-run benchmark noise — as long as nothing regresses by more than 0.4%. Since noise on throughput/ITL is routinely larger than the improvements this pipeline is trying to detect, this will open PRs that encode noise rather than real wins.

Suggested fix — gate the improvement side with the same threshold (or a dedicated one):

```python
improved = delta_pct > threshold
```

If the intent really is "any positive delta with no >0.4% regression," then the flag name/help is misleading and should be renamed (e.g. `regression_threshold`) with a separate, explicit improvement threshold.

tools/kernel/tuner/v1/autotune/autotune_result_processing.py
```diff
@@ -0,0 +1,688 @@
...
+            f"Updating {existing_params_updated} existing keys with new values."
+        )
+        new_keys_to_add = new_keys - existing_keys
+        logger.info(f"Adding {len(new_keys_to_add)} new keys to the mapping.")
+        # Update the in-memory mapping first.
+        module.tuned_params_mapping.update(tuning_key_to_params)
+        # Persist only tuned_params_mapping back to the original file.
+        self._write_tuned_params_mapping(Path(file_path), module)
+
+    def _write_tuned_params_mapping(self, file_path: Path, module):
```
QiliangCui2023, line 202: **No test covers the source-rewrite or the PR-gate logic — the two riskiest pieces**

This function (and `_build_mapping_text` just below it) rewrites a real `tuned_params.py`, and the result is auto-committed and force-pushed to an evaluation branch / PR. `evaluate_benchmark_metrics_and_create_pr_body` then decides whether to open that PR. Neither has a unit test.

`test_tuned_params_structure.py` and the new `from_string` test are good, but they don't exercise the rewrite itself. A round-trip test would be high-value and cheap:

1. Build a small module with `TuningKey`/`TunableParams` dataclasses + a sample `tuned_params_mapping`.
2. Run it through `_build_mapping_text` / `_write_tuned_params_mapping`.
3. `ast.parse` (or `compile`) the output, re-import it, and assert the resulting `tuned_params_mapping` equals the expected mapping (existing entries preserved + new ones applied).

That test would also catch the class of bug where a field doesn't round-trip through `repr(asdict(...))` — e.g. an enum/tuple/nested-dataclass field on a future registered kernel would emit invalid or unhashable source, which is currently committed with no guard (the shell path runs `py_compile`, but this Python rewrite path does not). It pairs naturally with adding an `ast.parse`/compile self-check inside `_write_tuned_params_mapping` before it writes the file.

Separately, the pure comparison logic in `evaluate_benchmark_metrics_and_create_pr_body` (delta / verdict / `should_create_pr`) can be factored out and table-tested without Spanner.

QiliangCui (approved):

tools/kernel/tuner/v1/autotune/autotune_result_processing.py
patrickji2014: thanks. apply same threshold to the improved condition here.

tools/kernel/tuner/v1/autotune/autotune_result_processing.py
patrickji2014: applied ast check before applying change to the tuned_params.py.

patrickji2014: tests are added into test_tuned_paramms_structure.py

yiw-wang (approved):

PR #2985, opened 2026-06-24
Refine cost model for reduce gather reduce kernel tiling and partitioning

bigeye:
# Description
Current partitioning algorithm doesn't consider the overhead of
pipeline and the cost of the prologue. This change adds more limits/caps
to accomodate those impacts based on emperical data points. With this
consideration, the performance of kernel improves up to 1.33x (at
num_tokens=8192)

# Microbenchmark Results (topk=10, hidden_dim=4096)

| # of tokens | # of valid rows | Baseline Latency (μs) | Optimized Latency (μs) | Speedup |
| --- | --- | --- | --- | --- |
| 8192 | 10240 | 519 | 392 | **1.32x** |
| 4096 | 5120 | 287 | 217 | **1.32x** |
| 2048 | 2560 | 158 | 140 | **1.12x** |
| 1024 | 1280 | 89 | 90 | **0.98x** |
| 512 | 640 | 62 | 62 | **1.00x** |
| 256 | 320 | 54 | 54 | **1.00x** |
| 128 | 160 | 50 | 51 | **0.98x** |

# QWEN3.5 E2E Benchmark Results

| Workload | Num Prompts | Max Concurrency | Baseline TPS | Experiment TPS | Speedup |
| --- | --- | --- | --- | --- | --- |
| 1k/8k | 640 | 64 | 2706.43 | 2693.39 | **1.00x** |
| 1k/8k | 640 | 128 | 4465.75 | 4548.37 | **1.02x** |
| 1k/8k | 640 | 256 | 6421.02 | 6768.09 | **1.05x** |
| 1k/8k | 640 | 512 | 8720.81 | 8770.91 | **1.01x** |
| 8k/1k | 640 | 64 | 14619.66 | 14929.13 | **1.02x** |
| 8k/1k | 640 | 128 | 20769.10 | 21506.32 | **1.04x** |
| 8k/1k | 640 | 256 | 23988.56 | 25023.29 | **1.04x** |
| 8k/1k | 640 | 512 | 27102.13 | 28389.88 | **1.05x** |

### DIFF
--- a/tpu_inference/kernels/sparse_core/ragged_gather_reduce_v2.py
+++ b/tpu_inference/kernels/sparse_core/ragged_gather_reduce_v2.py
@@ -88,6 +88,20 @@ class _Scratch:
         return getattr(self, dataclasses.fields(self)[index].name)
 
 
+class _CostModelConstants:
+    # Limit on the number of outer loop pipeline iterations. Too many iterations
+    # cause high cumulative pipeline overhead (e.g., from frequent pipeline
+    # startup/teardown bubbles). We try to find partitioning that does not exceed
+    # this limit on iterations.
+    MAX_ITERATIONS: int = 40
+
+    # Upper cap on the column chunk size processed per inner pipeline step.
+    # While larger chunk sizes help utilize bandwidth better, excessively large
+    # chunk sizes cause large pipeline bubbles. We cap it here to balance
+    # efficiency and bubble sizes.
+    MAX_COL_CHUNK_SIZE: int = 1024
+
+
 # ceil up to the nearest multiple of b.
 def _align_to(a, b):
     return pl.cdiv(a, b) * b
@@ -107,8 +121,9 @@ def _fallback_implementation(
     return out
 
 
-def _calculate_num_column_partitions(hidden_size: int, num_cores: int,
-                                     num_lanes: int) -> int:
+def _calculate_num_column_partitions(hidden_size: int, input_size: int,
+                                     num_cores: int, num_lanes: int,
+                                     num_simd_lanes: int) -> int:
     """Calculates the number of row partitions."""
     # Each column partition should be multiple of 128 (number of lanes) due to
     # DMA requirements.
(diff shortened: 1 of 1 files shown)


bigeye:
CC'ing @guowei-dev @kyuyeunk @gxd3 for reviews. I'm not sure how to add reviewers here. probably I don't have access.

kyuyeunk:
> CC'ing @guowei-dev @kyuyeunk @gxd3 for reviews. I'm not sure how to add reviewers here. probably I don't have access.

added as revieweres!

tpu_inference/kernels/sparse_core/ragged_gather_reduce_v2.py
```diff
@@ -123,7 +124,27 @@ def _calculate_num_column_partitions(hidden_size: int, num_cores: int,
...
            and hidden_size //
            (num_column_partitions * 2 * num_lanes) >= preferred_num_stages):
-        num_column_partitions *= 2
+        next_candidate = num_column_partitions * 2
+        next_row_partitions = num_cores // next_candidate
+
+        # Calculate exactly how many pipeline invocations (outer loop)
+        base_block_size = num_simd_lanes * next_row_partitions
+        num_row_subchunks = max(1, min(4, (input_size + base_block_size - 1) // base_block_size))
+        row_chunk_size = num_simd_lanes * num_row_subchunks
```
guowei-dev: Looks like this is replicating the call-site row tiling around L608 to predict the iteration count? If so, the two would always need to stay in sync, and we may want to pull them into a shared helper to keep them from diverging.

guowei-dev (approved):
Thanks for the great work!

tpu_inference/kernels/sparse_core/ragged_gather_reduce_v2.py
bigeye: Good catch! That makes sense. Done, thanks!

kyuyeunk (commented):
i feel like there's a lot of magic numbers in here. can you try to use a value from `get_tpu_info()` or try to have all empirical values in a one consolidated place?

bigeye:
> i feel like there's a lot of magic numbers in here. can you try to use a value from `get_tpu_info()` or try to have all empirical values in a one consolidated place?

Thanks @kyuyeunk for the comment. Yes, I agree that these are all empirical based magic numbers, which isn't indeed ideal. My first attempt (locally) was to use the spec BW to build this cost model, but I realized that we don't seem to be utilizing BW fully, even with big enough DMA size. This caused the cost model with another set of magic numbers (e.g., % of BW and startup latency) which I find it hard to make them accurate. I ended up having simpler heuristic.

I've added these magic constants into a single class. As a next step, I'll find a way to maximize DMA BW first, which will help us to find even more robust cost model.

kyuyeunk (approved):
nice! looks much better!

PR #3001, opened 2026-06-25
Support advanced JAX Profiler options passed via profile_prefix in vLLM

muditgokhale2:
Signed-off-by: Mudit Gokhale <muditgokhale@google.com>

Support advanced JAX Profiler options passed via profile_prefix in vLLM

### Description
Currently, vLLM on TPU only supports starting and stopping profiler traces without custom options, or relies entirely on static environment variables. This limits user control over profiling detail (e.g., host vs device levels) and prevents passing advanced accelerator flags. This PR implements dynamic parsing and mapping of advanced profiler options passed via the `profile_prefix` string in `TPUWorker.profile()`.

**Why do we need this?**
Profiling options are needed to enable advanced profiling knobs.

**Why is this a good solution?**
It provides an interface consistent with existing PyTorch profiling patterns while remaining flexible enough to pass arbitrary experimental JAX/compiler config variables directly to `libtpu`. It avoids hardcoding option keys, enabling users to immediately leverage new compiler options.

**Implementation Details:**
1. Parsing Algorithm:
- The parser splits the semicolon-separated (`;`) prefix string into individual configuration parts to avoid delimiter collisions with options containing inner commas or colons (e.g. lists).
- Each part is stripped of surrounding whitespace and matched against a strict regex pattern: `^([a-zA-Z_][a-zA-Z0-9_]*)\s*:\s*(.+)$`
- Keys matching the pattern are categorized as standard options if they are present in the stable keys set `_STANDARD_KEYS` (`host_tracer_level`, `device_tracer_level`, `python_tracer_level`) and are mapped directly as attributes on JAX's `ProfileOptions` class.
- All other keys are categorized as experimental options and are added directly to JAX's `advanced_configuration` dictionary.
- The parser infers value types: `true`/`false` (case-insensitive) are mapped to Python booleans, integers are parsed as `int`s, and all other strings are retained. String values are strictly validated against a safe whitelist pattern (`^[a-zA-Z0-9_./,:-]*$`) to prevent injection vulnerabilities.

2. Rules for prefix strings:
- Must contain only ASCII characters.
- Must be formatted as semicolon-separated key-value pairs where key and value are separated by a colon, e.g., `host_tracer_level:3;tpu_num_chips_to_profile_per_task:2;my_custom_option:true`
- Option values can contain inner commas or colons (e.g., lists of configs), which are preserved.
- Refer to the OpenXLA XProf documentation for the list of supported advanced profiling options: https://github.com/openxla/xprof/blob/master/docs/advanced_profiler_options.md

3. Invalid Case Handling and Errors:
- Non-ASCII inputs: Raises a `ValueError` if any non-ASCII characters are detected.
- Syntax / Colon-less strings (e.g., missing delimiter or using `=`): Raises a `ValueError` stating the invalid format.
- Unsafe characters: Raises a `ValueError` if an option value contains characters outside the safe whitelist.
- Unknown / Unrecognized Experimental Flags: Set `check_experimental_options` to `True` by default in the advanced configuration. This causes JAX/libtpu backend to fail with a runtime initialization error (rather than silently ignoring them) if a key is not recognized. Users can opt-out of this strict check by explicitly passing `check_experimental_options:false`.

4. Test Setup Refactoring:
- Factored out the initialization of `TPUWorker` test instances into a shared private helper method `_create_worker()` in `tests/worker/tpu_worker_test.py` to reduce code duplication across the unit test suite, addressing code review feedback.

### Tests
Tested this change using two approaches:
1. Unit Tests: Added a comprehensive test suite in `tests/worker/tpu_worker_test.py` covering:
   - Default tracer startup/shutdown behaviour.
   - Parsing of standard `ProfileOptions`.
   - Parsing of advanced custom options (with commas and colons).
   - Malformed/invalid configuration handling (invalid symbols, non-ASCII inputs, colon-less segments).
   Run commands:
   `pytest tests/worker/tpu_worker_test.py`

2. Physical Hardware Validation: Successfully ran JAX-native Qwen2.5-0.5B model on physical TPU v6e-4 VM (tensor_parallel_size=2) and collected traces using custom JAX options. Verified that host metadata (`*.trace.json.gz`) and accelerator hardware traces (`*.xplane.pb`) were generated successfully.
   - Specifically verified that passing custom firmware throttling and power events dynamically via `profile_prefix`:
     `e2e_enable_fw_throttle_event:true;e2e_enable_fw_power_level_event:true;e2e_enable_fw_thermal_event:true`
     successfully captures the HLO/timeline execution trace.

   **Xprof snipit:**
   https://screenshot.googleplex.com/8iSboAJzFWoXY7y

Signed-off-by: Mudit Gokhale <muditgokhale@google.com>

### DIFF
--- a/tests/worker/tpu_worker_test.py
+++ b/tests/worker/tpu_worker_test.py
@@ -420,16 +420,21 @@ class TestTPUWorker:
     # --- Profiling and Health Check Tests ---
     #
 
-    @patch('tpu_inference.worker.tpu_worker.jax')
-    @patch.dict('os.environ', {"PYTHON_TRACER_LEVEL": "1"}, clear=True)
-    def test_profile_start(self, mock_jax, mock_vllm_config):
-        """Tests starting the JAX profiler."""
+    def _create_worker(self,
+                       mock_vllm_config,
+                       profile_dir="/tmp/profile_dir") -> TPUWorker:
         worker = TPUWorker(vllm_config=mock_vllm_config,
                            local_rank=0,
                            rank=0,
                            distributed_init_method="test")
-        worker.profile_dir = "/tmp/profile_dir"
+        worker.profile_dir = profile_dir
+        return worker
 
+    @patch('tpu_inference.worker.tpu_worker.jax')
+    @patch.dict('os.environ', {"PYTHON_TRACER_LEVEL": "1"}, clear=True)
+    def test_profile_start(self, mock_jax, mock_vllm_config):
+        """Tests starting the JAX profiler."""
+        worker = self._create_worker(mock_vllm_config)
         worker.profile(is_start=True)
 
         mock_jax.profiler.ProfileOptions.assert_called_once()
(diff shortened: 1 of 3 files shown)


tests/worker/tpu_worker_test.py
```diff
@@ -366,6 +380,102 @@ def test_profile_start(self, mock_jax, mock_vllm_config):
         # Verify options from env var were used
         assert kwargs['profiler_options'].python_tracer_level == 1
 
+    @patch('tpu_inference.worker.tpu_worker.jax')
+    def test_profile_with_prefix_options(self, mock_jax, mock_vllm_config):
+        """Tests that profile_prefix options are correctly parsed and applied."""
+        worker = TPUWorker(vllm_config=mock_vllm_config,
+                           local_rank=0,
+                           rank=0,
+                           distributed_init_method="test")
```
rushabh-46: considering factoring this setup into a shared helper across all the unit tests

tpu_inference/worker/tpu_worker.py
```diff
@@ -89,6 +90,89 @@ def __post_init__(self):
...
+    Returns:
+        A tuple containing two dictionaries:
+          1. standard_opts: Dictionaries of parsed option keys and their typed
+             values (e.g. bool, int, or str) that match standard JAX options
+             (host_tracer_level, device_tracer_level, python_tracer_level).
+          2. advanced_opts: Dictionaries of parsed custom option keys and their
+             typed values to be passed to JAX's advanced_configuration dictionary.
+          E.g., ({"host_tracer_level": 3}, {"tpu_num_chips_to_profile_per_task": 2, "my_custom_option": True})
+    """
+    if not profile_prefix:
```
rushabh-46: Handle whitespace only:
```
if not profile_prefix or not profile_prefix.strip():
    # Fall back to default profiling behavior or skip custom parsing
    ...
```

rushabh-46 (approved):

tpu_inference/worker/tpu_worker.py
muditgokhale2: Done

tests/worker/tpu_worker_test.py
muditgokhale2: Done

tpu_inference/worker/tpu_worker.py
```diff
@@ -89,6 +90,89 @@ def __post_init__(self):
...
+    """
+    val_lower = val.lower()
+    if val_lower == "true":
+        return True
+    if val_lower == "false":
+        return False
+    try:
+        return int(val)
+    except ValueError:
+        if not re.match(r'^[a-zA-Z0-9_./,:-]*$', val):
```
aniket20: Should we compile the regex once and reuse it here?

At module level: `_ALLOWED_VALUE_CHARS = re.compile(r'^[a-zA-Z0-9_./,:-]*$')`

And then here: `_ALLOWED_VALUE_CHARS.match(val)`

tests/worker/tpu_worker_test.py
```diff
@@ -366,22 +381,92 @@ def test_profile_start(self, mock_jax, mock_vllm_config):
...
+                                                   mock_vllm_config):
+        """Tests that malformed options (without colons) throw ValueError."""
+        worker = self._create_worker(mock_vllm_config)
+
+        prefix = "host_tracer_level=5;another_option"
+        with pytest.raises(ValueError, match="Expected 'key:value' format"):
+            worker.profile(is_start=True, profile_prefix=prefix)
+
+    @patch('tpu_inference.worker.tpu_worker.jax')
+    def test_profile_with_prefix_options_non_ascii(self, mock_jax,
```
aniket20, line 512: Consider adding a test case that ensures empty parts or extra semicolons are gracefully ignored (e.g., `prefix = "host_tracer_level:3;;tpu_power_trace_level:2;"`). The current implementation handles it correctly via `if not part: continue`, but explicit test coverage ensures it doesn't regress.

tpu_inference/worker/tpu_worker.py
```diff
@@ -89,6 +90,89 @@ def __post_init__(self):
...
+    standard_opts = {}
+    advanced_opts = {}
+
+    parts = profile_prefix.split(';')
+
+    for part in parts:
+        part = part.strip()
+        if not part:
+            continue
+        match = re.match(r'^([a-zA-Z_][a-zA-Z0-9_]*)\s*:\s*(.+)$', part)
```
aniket20: Same comment here, should we compile the regex once and reuse?

tpu_inference/worker/tpu_worker.py
```diff
@@ -534,15 +607,34 @@ def profile(self,
...
+                advanced_config.update({
                     "tpu_num_chips_to_profile_per_task": 1,
                     "tpu_num_sparse_cores_to_trace": 1,
                     "tpu_num_sparse_core_tiles_to_trace": 1,
-                }
+                })
+
+            # Override with parsed options
+            for key, val in standard_opts.items():
+                setattr(options, key, val)
```
aniket20: Should we check that the key exists in options first?

```
if hasattr(options, key):
   setattr(options, key, val)
else:
        # Optional: log a warning
  pass
```

muditgokhale2: Done.

tpu_inference/worker/tpu_worker.py
muditgokhale2: Done.

tests/worker/tpu_worker_test.py
muditgokhale2: Added a test case.

tpu_inference/worker/tpu_worker.py
muditgokhale2: That's a good suggestion. Done!

aniket20 (approved):

mrjunwan-lang (approved):

bvrockwell (approved):

PR #3004, opened 2026-06-25
fix the retrieving of sharding config to avoid weights replication during model loading.

lc5211:
# Description
fix the retrieving of sharding config to avoid weights replication during model loading. With gemma4-e2b-it as an example, 
Before the fix, when loading the model on a v5p-4 TPU host, each tpu device HBM consumption is around 9GiB, total at 36GiB. Basically the model is fully replicated on each TPU device.
With the fix, each tpu device HBM consumption is down to around 2.2GiB, total at 8.8GiB.

# Tests

Verified manually on a v5p-4 TPU host.

# Checklist

Signed-off-by: lc5211 ([linchai@google.com](mailto:linchai@google.com))

---
<a href="https://app.blacksmith.sh/vllm-project/codesmith/tpu-inference/pr/3004"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785008399&installation_id=142609222&pr_number=3004&repository=vllm-project%2Ftpu-inference&return_to=https%3A%2F%2Fgithub.com%2Fvllm-project%2Ftpu-inference%2Fpull%2F3004&signature=d5c5bb75501cc82a13d05392df786238e34c4ff8735ad7c014b8785173da740f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


tpu_inference/layers/jax/norm.py
```diff
@@ -41,8 +41,13 @@ def __init__(self,
         # apply transpose here to match HF weight layout.
         self.weight = self.scale
         delattr(self, 'scale')
-        if hasattr(self.weight, 'out_sharding'):
-            self.weight.set_metadata('sharding', self.weight.out_sharding)
+        if self.weight is not None:
```
lk-chen: not sure if I follow the logic

1. we use [with_partitioning](https://github.com/vllm-project/tpu-inference/blob/ea832c1b735e5b497f9e6c86c4da312121517b04/tpu_inference/models/jax/gemma4.py#L80) to provide sharding
2. which will assign [out_sharding](https://github.com/google/flax/blob/07560a7a9d0c04b1a168055ed7ce0de1170782ce/flax/nnx/spmd.py#L128) in varaible/param
3. here in norm/linear/embed etc. we put both "out_sharding" and "sharding" for compatibility

Your change is checking both "out_sharding" and "sharding", I'm not sure why that helps

lc5211: The main issue was fixed in weight_utils.py where originally we only check for `sharding` attr when loading dummy weights. But here to make it more robust, it's more safe to check for both attributes so we won't miss the sharding in either case.

lk-chen: But as I explained above, we shouldn't expect "sharding" missing.

A better fix is to remove re-direct logic totally
```
        if hasattr(self.weight, 'out_sharding'):
            self.weight.set_metadata('sharding', self.weight.out_sharding)
```
and in weight_utils we always check "out_sharding"

lc5211: Done, that's cleaner fix.

lk-chen:
thx, waiting for CI finish and will take another look

lc5211:
> thx, waiting for CI finish and will take another look

Thanks! The CI is done and clear!

lk-chen (approved):

PR #2804, opened 2026-06-03
Use newer mm-data encoder interface when possible

wdhongtw:
# Description

If user explicitly specifies using "Encoder CUDA Graph" feature (and the model does support the interface),
use the new interface to achieve better performance and avoid runtime re-compilation that slows down online-inference services.

- Use `SupportsEncoderCudaGraph.encoder_cudagraph_forward` instead of `SupportsMultiModal.embed_multimodal`
- Implement pre-compilation for supported token budgets
- Split mm-data into multiple batches when necessary
- Move the encoder CUDA graph logic into `MultiModalManager` so it can also be leveraged by the JAX path in the future
- Extract mm encoder manager into a separate file (`tpu_inference/models/vllm/mm_encoder_manager.py`) with unit tests
- Patch encoder forward and corresponding pre-compile function in `TPURunner` when appropriate

Corp internal doc: <a href="https://docs.google.com/document/d/10cYjQllhOhRPhtzuKLxS4G6HbpHz9lu56P6KA4zZve8/edit?usp=sharing">link</a>

### Performance

| Metric | Before | After |
|--------|--------|-------|
| Throughput (req/s) | 0.78 | 3.05 |

# Tests

A command to trigger corresponding logic.

```shell
MODEL_IMPL_TYPE=vllm python examples/generate/multimodal/vision_language_offline.py --model-type=qwen3_vl --num-prompts=1 --modality=image --time-generate  -tp=4 --enable-vit-cuda-graph
```

Also, for benchmarking.

```shell
MODEL_IMPL_TYPE=vllm vllm serve Qwen/Qwen3-VL-2B-Instruct \
  --max-model-len 8192 --disable-uvicorn-access-log \
  --max-num-batched-tokens 4096 --max-num-seqs 256 \
  --tensor-parallel-size 8 --trust-remote-code \
  --limit-mm-per-prompt '{"image": 8}' \
  --no-enable-prefix-caching --mm-processor-cache-gb=0 --gpu-memory-utilization 0.9 \
  --disable-chunked-mm-input \
  --mm-processor-kwargs '{"max_pixels": 512000}' \
  --compilation-config '{"cudagraph_mm_encoder": true}'

client:
time python scripts/vllm/benchmarking/benchmark_serving.py \
  --backend vllm-chat \
  --endpoint /v1/chat/completions \
  --model Qwen/Qwen3-VL-2B-Instruct \
  --dataset-name mmmu_pro \
  --mmmu-pro-subset 'standard (10 options)' \
[...]

PR #2802, opened 2026-06-03
[Torchax] Re-implement MMEncoderJITManager

lk-chen:
# Description

The component is needed to adopt vLLM [ViT CUDA Graph](https://docs.vllm.ai/en/v0.22.0/design/cuda_graphs_multimodal/). This PR is refactoring of #2804 

* This PR pulls encoder manager into a standalone class (instead of nested function) which is more testable, [test coverage (91%)](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/20369#019ed455-62f1-419a-9346-5c3802faf5b4/L233-L234) is better than [HEAD (51%)](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/20361#019ed433-7236-47ca-8ab2-e5b285a6b6d5/L225)
* Performance is better, with this refactoring we get [3.46 req/s](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/20369#019ed455-8869-4679-a9b5-90e80fbc0ed3/L2198) compare to [2.82 req/s on nightly](https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/1498#019ed032-39d2-4260-8a1d-95e833762972/L4509)

image

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3008, opened 2026-06-26
[Gemma4] Refactor MMEncoderJITManager and add support to use it in Jax path Gemma4

kwang3939:
# Description

Based on the enable of using [Vit Cuda Graph](https://docs.vllm.ai/en/v0.22.0/design/cuda_graphs_multimodal/) on torchax path: 
PR: https://github.com/vllm-project/tpu-inference/pull/2804
Refactor PR: https://github.com/vllm-project/tpu-inference/pull/2802

Refactor the MMEncoderJITManager to be shared by both paths. Add adapter and funcs needed for jax path Gemma4 to use it.

# Tests

Verified the correctness of jax path Gemma4: `{'accuracy': 0.7116, 'correct': 1231, 'total': 1730, 'gen_num': 1730}`
Verified the correctness of torchax path `Qwen/Qwen3.5-397B-A17B-FP8`.
Tested with `python3 -m pytest -s -v tests/runner/test_mm_encoder_jit_manager.py`.

# Checklist

### DIFF
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -41,6 +41,8 @@ from tpu_inference.models.jax.utils.qwix.qwix_utils import (
     update_vllm_config_for_qwix_quantization)
 from tpu_inference.models.jax.utils.weight_utils import (BaseWeightLoader,
                                                          LoadableWithIterator)
+from tpu_inference.runner.mm_encoder_jit_manager import \
+    maybe_create_mm_encoder_jit_manager
 from tpu_inference.utils import to_jax_dtype, to_torch_dtype
 
 logger = init_logger(__name__)
@@ -462,8 +464,26 @@ def get_flax_model(
         return jitted_model_fn(*args, **kwargs)
 
     compute_logits_fn = run_compute_logits
-    embed_multimodal_fn = run_embed_multimodal
     embed_input_ids_fn = run_embed_input_ids
+
+    _mm_jit = maybe_create_mm_encoder_jit_manager(
+        vllm_config=vllm_config,
+        vllm_model=model,
+        vllm_runner=None,
+        params_and_buffers=None,
+    )
+    if _mm_jit is not None:
+        precompile_vision_encoder_fn = _mm_jit.precompile_vision_encoder
+
+        def embed_multimodal_fn(state_leaves, modality=None, **kwargs):
+            if modality is not None and _mm_jit.supports_modality(modality):
+                return _mm_jit.execute(kwargs)
+            return run_embed_multimodal(state_leaves,
+                                        modality=modality,
+                                        **kwargs)
+    else:
+        embed_multimodal_fn = run_embed_multimodal
+
     lora_manager, model = None, None
     combine_hidden_states_fn = combine_hidden_states
 
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -482,6 +482,8 @@ def get_flax_model(
                                         modality=modality,
                                         **kwargs)
     else:
+        # TODO(kwang3939): implement SupportsEncoderCudaGraph for Qwen2.5-VL and enforce
+        # all JAX multimodal models to go through MMEncoderJITManager.
         embed_multimodal_fn = run_embed_multimodal
 
     lora_manager, model = None, None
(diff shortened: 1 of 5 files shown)


tpu_inference/models/common/model_loader.py
```diff
@@ -462,8 +464,26 @@ def wrapped_model_fn(*args, **kwargs):
         return jitted_model_fn(*args, **kwargs)
 
     compute_logits_fn = run_compute_logits
-    embed_multimodal_fn = run_embed_multimodal
     embed_input_ids_fn = run_embed_input_ids
+
+    _mm_jit = maybe_create_mm_encoder_jit_manager(
```
lk-chen, line 469: for Gemma4, how does it create manager? in `mm_encoder_jit_manager.py:451` it checks [support_encoder_cudagraph](https://github.com/vllm-project/vllm/blob/3d3b96488f5f7d94b1ab63919e0f1e7922a9ded6/vllm/model_executor/models/interfaces.py#L1703) which only checks if model is instance of `SupportsEncoderCudaGraph`. Gemma-4 is not yet inheriting `SupportsEncoderCudaGraph`?

kwang3939: For jax path I passed the jax model to check on `support_encoder_cudagraph`, so I added `supports_encoder_cudagraph = True` in our gemma4_mm.py.

lk-chen: how's this boolean used? vllm's `supports_encoder_cudagraph` method only checks `isinstance(..`

kwang3939: I had the same concern and tested it and found it was initialized correctly. I looked up a bit and found SupportsEncoderCudaGraph is a @runtime_checkable Protocol, so isinstance() just checks if supports_encoder_cudagraph exists as a class attribute, not actual inheritance.

tpu_inference/models/jax/gemma4_mm.py
```diff
@@ -808,6 +812,218 @@ def embed_multimodal(self, **kwargs) -> List[jax.Array]:
...
+            modalities=["image"],
+            buffer_keys=["pixel_values", "pixel_position_ids"],
+            out_hidden_size=text_hidden,
+            max_frames_per_video=1,
+            padding_logics={"pixel_position_ids": pad_pixel_position_ids},
+        )
+        return config
+
+    def get_max_frames_per_video(self) -> int:
+        return 1  # images only
```
lk-chen: shall we raise unimplemented exception

tpu_inference/models/common/model_loader.py
lk-chen: Got it, I was confused because vLLM models all do inheritance e.g. https://github.com/vllm-project/vllm/blob/dccb412e2c72a0c147166c14b25c01f045a74163/vllm/model_executor/models/deepseek_ocr.py#L366 If it's technically possible maybe we should do the same (inherit vllm class) for better readability.

tpu_inference/models/common/model_loader.py
```diff
@@ -462,8 +464,26 @@ def wrapped_model_fn(*args, **kwargs):
...
+        precompile_vision_encoder_fn = _mm_jit.precompile_vision_encoder
+
+        def embed_multimodal_fn(state_leaves, modality=None, **kwargs):
+            if modality is not None and _mm_jit.supports_modality(modality):
+                return _mm_jit.execute(kwargs)
+            return run_embed_multimodal(state_leaves,
+                                        modality=modality,
+                                        **kwargs)
+    else:
+        embed_multimodal_fn = run_embed_multimodal
```
lk-chen, line 487: let's put a TODO here to implement this protocol for qwe2.5-vl and enforce all jax model to go this approach

lk-chen (approved):

tpu_inference/models/jax/gemma4_mm.py
kwang3939: Done.

tpu_inference/models/common/model_loader.py
kwang3939: Done.

tpu_inference/models/common/model_loader.py
kwang3939: Just tried using explicit inheritance and found it causes a metaclass conflict. I'll keep this as it is then.

lk-chen:
@copilot could you list per-file owner? don't at them, just put username

kyuyeunk (approved):

PR #3012, opened 2026-06-26
[Model] Add NemotronH (hybrid Mamba2 + non-gated MoE) TPU support

Junyi-99:
## Summary

Adds TPU support for NVIDIA Nemotron-H, a hybrid model interleaving Mamba2, GQA attention, and non-gated squared-ReLU MoE layers. This PR validates `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` end-to-end on TPU v6e-4 with correct output.

The implementation follows the existing GDN integration pattern: unsupported PyTorch/vLLM ops are traced through torchax where possible and replaced with explicit JAX/OOT implementations where torchax has no lowering.

This depends on the companion vLLM-core change that allows `is_act_and_mul=False` on TPU. (vllm-project/vllm#46795)

## Changes

* Adds non-gated MoE support on the TPU GMM path: non-gated experts use `w1/w2`, skip the `w3` split, run plain GMM, and apply squared-ReLU in JAX.
* Supports non-uniform hybrid KV-cache groups produced by irregular attention spacing in Nemotron-H.
* Adds a TPU/JAX lowering for `torch.ops.vllm.mamba_mixer2`, including conv1d, SSD recurrence, cache read/write, gated RMSNorm, and output projection.
* Adds a jagged chunked SSD fresh-prefill kernel, with decode, mixed batches, prefix caching, and continuation prefill falling back to the token-scan path.
* Fixes `warning_once` on non-uniform KV-cache groups by passing a hashable tuple instead of a list.

## Review focus

The most important correctness point is TP sharding of `x/B/C`.

`in_proj` emits `[gate | x | B | C | dt]`, but `x` is head-sharded while `B/C` are group-sharded. Passing concatenated `[x|B|C]` into one `shard_map P(ATTN_HEAD)` slices across tensor boundaries and mismatches `x` heads with `B/C` groups, producing incorrect SSD output.

The fix splits `x`, `B`, and `C` into separate `shard_map` arguments, shards each by its own logical dimension, and re-concatenates them inside the per-shard core.

## Validation

Validated against an HF transformers CPU reference and end-to-end TPU generation:

* Per-shard post-conv `x/B/C` align with the corresponding HF shard after the TP sharding fix.
* Jagged SSD prefill matches the token-scan path for both `y` and final `ssm_state`.
* End-to-end generation on TPU v6e-4 (`tp=4`, bf16, `enforce_eager`) produces expected outputs, including:

  * `The capital of France is` → ` Paris`
  * `The largest planet in our solar system is` → ` Jupiter.`
  * `The opposite of hot is` → ` cold.`

The jagged SSD kernel speeds up isolated fresh prefill substantially versus token-scan; non-fresh-prefill cases remain on the conservative scan path.

## AI assistance

Developed with AI assistance. Every changed line has been reviewed by the submitter.

## Acknowledgements

This work is part of the effort from the [medusa-compute](https://github.com/medusa-compute) team.

Special thanks to @Raphael-Hao for his substantial help and guidance throughout this work.

### DIFF
--- a/tpu_inference/kernels/mamba2/__init__.py
+++ b/tpu_inference/kernels/mamba2/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/tpu_inference/kernels/mamba2/jagged_ssd.py
+++ b/tpu_inference/kernels/mamba2/jagged_ssd.py
@@ -1,0 +1,343 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Single Pallas kernel for jagged (variable-length) Mamba2 SSD prefill on TPU.
+
+`ssd_candidate(x, dt, A_log, B, C, cu_seqlens) -> y` computes the chunked
+state-space-dual (SSD) scan for a packed batch of sequences in one pallas_call,
+fusing the three SSD phases (intra-chunk y, cross-chunk state carry, inter-chunk
+y from the carried state) so no (NumChunks, H, P, N) intermediates hit HBM.
+
+Algebra inside chunk c (h_init carries from the previous chunk):
+    M[i,j,h]   = exp(cum_log_a[i,h]-cum_log_a[j,h]) * dt[j,h]
+                 * 1[same sub-seq(i,j)] * 1[j<=i]
+    score[i,j] = sum_n C[i,h,n] * B[j,h,n]
+    y_intra    = sum_j M * score * x[j]
+    h_local    = sum_j M[CHUNK-1,j,h] * x[j] (x) B[j]
+    y_inter[t] = exp(cum_log_a[t]) * inter_valid[t] * sum_n C[t,h,n]*h_init[h,n]
+    h_init    <- carry_factor[c] * h_init + h_local
+
+Shapes: x (T,H,P) bf16, dt (T,H) bf16, A_log (H,) f32 (log(a)=A_log*dt),
+B/C (T,H,N) bf16, cu_seqlens (num_seqs+1,) i32; y (T,H,P) bf16. Hardcoded
+H=128, CHUNK=256, I_TILE=64; T is padded to a CHUNK multiple internally.
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import pallas as pl
+from jax.experimental.pallas import tpu as pltpu
+
+
+CHUNK = 256
+I_TILE = 64
+NUM_I_TILES = CHUNK // I_TILE   # 4
+HBG_SIZE = 128
+
+# Bound exp arg in masked / outside-causal positions so exp doesn't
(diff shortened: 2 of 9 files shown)


Junyi-99:
### What was validated

Tested on TPU v6e-4 with `tp=4`, bf16, and `enforce_eager`.

#### 1. MoE Nemotron-H variant

`nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`

Basic generation produces expected outputs:

```text
'The capital of France is'                  -> ' Paris'
'The largest planet in our solar system is' -> ' Jupiter.'
'The opposite of hot is'                    -> ' cold.'
```

#### 2. Dense Nemotron-H variant

`nvidia/NVIDIA-Nemotron-3-Nano-4B-BF16`

This variant has no MoE, so it separately validates the Mamba2 lowering and non-uniform hybrid KV-cache path without relying on the MoE implementation.

Basic generation also produces expected outputs:

```text
'The capital of France is'                  -> ' Paris'
'The largest planet in our solar system is' -> ' Jupiter, which is a gas giant.'
'The opposite of hot is'                    -> ' cold.'
```

### Prefill sanity check

I also ran `vllm bench serve` sweeps on v6e-4 (`tp=4`, bf16, eager, `DEC_OUT=64`) to check the fresh-prefill path and fallback behavior.

Prefill throughput with concurrency 1:

| ctx  | 4B tok/s | 30B tok/s |
| ---- | -------- | --------- |
| 512  | 7,080    | 7,903     |
| 1024 | 12,456   | 12,337    |
| 2048 | 19,651   | 17,040    |
| 4096 | 24,781   | 20,924    |
| 8192 | 92.7     | 162       |

Fresh single-chunk prefill scales cleanly up to 4096. The 8192 cliff is expected: with `max_num_batched_tokens=4096`, `ctx > 4096` goes through chunked/continuation prefill and falls back to the token-scan path instead of the jagged SSD kernel.

### Why this vLLM change is safe

Today, vLLM rejects `FusedMoEConfig(is_act_and_mul=False)` on TPU during config validation before the TPU backend can run.

The companion backend PR shows that the TPU path can handle this configuration. This PR only adds TPU to the existing allow-list for `is_act_and_mul=False`; CUDA, XPU, and ROCm behavior is unchanged, and unsupported platforms remain blocked.

### Known backend limitations

The jagged SSD fresh-prefill kernel currently covers fresh prefill only. Continuation prefill, including chunked prefill beyond `max_num_batched_tokens`, still falls back to the token-scan path.

The jagged SSD kernel also pads per-shard heads to `H=128`, which increases scoped VMEM usage. The dense 4B variant exposed that the previous 96 MB VMEM cap was too tight for some valid configs; increasing the cap avoids the immediate compile-time OOM, while specializing the kernel to the real `H` would be the cleaner follow-up.

Decode benchmarking was only used as a smoke test under eager mode. The observed TPOT degrades sharply at larger batch sizes without graph capture, and one `bs=64` 30B run OOMed the serve engine. I would not treat those eager-mode decode numbers as representative performance results for this PR.

Junyi-99:
## Nemotron-3-Nano-30B accuracy benchmark on TPU v6e-4

Purpose: verify that the TPU fork stack (`Junyi-99/vllm@nemotron` + `Junyi-99/tpu-inference@nemotron`) preserves model quality. This is a small-sample correctness check, not an official NVIDIA benchmark reproduction.

### Setup

* Hardware: TPU v6e-4, `tensor_parallel_size=4`, bf16, `enforce_eager=True`.
* Stack: vLLM fork `0.1.dev17769+gcbbbe41af`, tpu-inference fork, JAX 0.10.1, libtpu 0.0.41, torch 2.10.0.
* Model: `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`, restored from GCS `model-cache/`.
* Eval: `lm-eval 0.4.12` for completion/chat runs; direct vLLM script for true-thinking.
* Thinking mode is controlled by the chat template flag `enable_thinking`.


### GSM8K results

Small-sample GSM8K runs only, N=20–30.

| Mode         | Setup                                          | Result                                        |
| ------------ | ---------------------------------------------- | --------------------------------------------- |
| Non-thinking | completion 5-shot, lm-eval gsm8k               | strict **0.85** (17/20), flexible 0.50        |
| Non-thinking | chat zero-shot, empty `<think></think>`        | flexible **0.667** (20/30), strict 0.0        |
| Thinking     | chat zero-shot, open `<think>` via direct vLLM | **18/20 truly correct**; 16/20 auto-extracted |

### Conclusion

The TPU fork stack appears numerically correct. GSM8K outputs are coherent in both thinking and non-thinking modes, with no visible quality degradation from Mamba2 lowering, bf16, hybrid KV-cache, or jagged kernels.

### Caveats

Small samples only. MMLU-Pro was not run. Longer thinking runs should use `max_tokens >= 4096`.

Junyi-99:
[tpu-inference Nemotron gsm8k logs.zip](https://github.com/user-attachments/files/29368269/tpu-inference.Nemotron.gsm8k.logs.zip)

The raw logs of evaluation has been attached.

PR #3025, opened 2026-06-26
[Fix][Raiden] Load tpu raiden before jax to avoid XLA symbol crash

juncgu-google:
# Description

Make`tpu_inference/__init__.py` import the Raiden engine extension before the first `import jax`, so its embedded XLA runtime comes up before jaxlib's libjax_common.so and the two XLA copies don't collide in static initializers. harmless when tpu_raiden is not installed (import is skipped quietly).

# Tests

Verified by running `examples/offline_inference.py`. tpu-inference does not crash with or without tpu_raiden installed.

# Checklist

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -74,6 +74,7 @@ if TYPE_CHECKING:
     PROFILE_SINGLE_DEVICE: bool = False
     LORA_MODULE_PATH: str = ""
     SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES: str = "auto"
+    TPU_RAIDEN_IMPORT_FIRST: bool = True
 
 
 def env_with_choices(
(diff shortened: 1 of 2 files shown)


yiw-wang (approved):

tpu_inference/envs.py
```diff
@@ -74,6 +74,7 @@
     PROFILE_SINGLE_DEVICE: bool = False
     LORA_MODULE_PATH: str = ""
     SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES: str = "auto"
+    TPU_RAIDEN_IMPORT_FIRST: bool = True
```
QiliangCui2023: do you mean default to "True" or false?

QiliangCui2023: from description "true" is on purse. then, why do we need this flag?

QiliangCui2023: if we default the value to True and import is safe, I think it is not necessary to have the flag.

juncgu-google: Thanks for the suggestions. Updated the code.

QiliangCui2023 (approved):

QiliangCui2023 (approved):

QiliangCui (approved):

PR #3028, opened 2026-06-26
Add q_split logic for mixed mode for Kimi 8k/1k prefill.

Xhark:
# Description

Add q_split logic for mixed mode, and add tuned params for Kimi prefill.

# Tests

It makes ~2% improvement on `Output token throughput (tok/s)` for 8k/1k Kimi task.

### DIFF
--- a/tests/kernels/mla_tuned_vs_baseline_test.py
+++ b/tests/kernels/mla_tuned_vs_baseline_test.py
@@ -29,6 +29,9 @@ from tpu_inference.kernels.mla.v2.tuned_params import (TunableParams,
 def _get_tuned_test_cases():
     test_cases = []
     for key in tuned_params_mapping.keys():
+        # Current performance test only for batch_decode.
+        if key.case != "batched_decode":
+            continue
         name = (f"tokens_{key.max_num_tokens}_"
                 f"seqs_{key.max_num_seqs}_"
                 f"pagesperseq_{key.pages_per_seq}")
--- a/tests/layers/common/test_attention_interface.py
+++ b/tests/layers/common/test_attention_interface.py
@@ -489,4 +489,5 @@ def test_mla_attention(monkeypatch, mesh):
     _, kernel_kwargs = mock_mla_kernel.call_args
     assert kernel_kwargs["num_kv_pages_per_block"] == (3, 1, 1)
     assert kernel_kwargs["num_queries_per_block"] == (1, 16, 16)
+    assert kernel_kwargs["mixed_q_split"] == 1
     assert kernel_kwargs["sm_scale"] == 0.1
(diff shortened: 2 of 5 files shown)


gxd3 (approved):

Xhark:
@kyuyeunk It passes all the tests, could you review this PR?

tpu_inference/envs.py
```diff
@@ -424,6 +424,10 @@ def _get_int_list_env() -> list[int]:
     env_bool("SLICE_ROPE_CACHE", default=False),
     "MLA_TRANSPOSE_KV_CACHE":
     env_bool("MLA_TRANSPOSE_KV_CACHE", default=False),
+    "MIXED_Q_SPLIT":
```
kyuyeunk: hmm i'm generally really against adding more manually tuned parameters. it's a really bad user experience and prevents our optimization to scale across different use cases or different model architecture that leverages the same kernel. similar to how batched rpa or gmm has cost model based value, is it possible to calculate a good default value?

Xhark: With current MLA param tuning design, I could add them into tpu_inference/kernels/mla/v2/tuned_params.py .  Do you think it would be better to move this params to there? or could you point out any reference code pointer?

Xhark: I'm moving envs to tuned params. testing is WIP, PTAL code shapes better or not.

kyuyeunk (approved):

QiliangCui2023:
Heads-up: this appears to have broken the kernels CI job on `main` on **both tpu6e and tpu7x**.

First failing build: [tpu-inference-ci #22385](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22385) (commit 472353848, this PR's merge commit — the immediately preceding main builds 22338/22345 were green on the same jobs). Later main builds show the kernels job as `broken` (gated), so it may not have surfaced since.

```
FAILED tests/kernels/mla_v2_test.py::MlaRaggedPagedAttentionKernelV2Test::test_ragged_paged_attention_basic
jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: E1001: CompileTimeScopedVmemOom:
Ran out of memory in memory space vmem while allocating on stack for
%xpose_pipeline_shape_384x128x512_... = bf16[128,384,512]... custom-call(%pallas_call.20)
Scoped allocation with size 33.56M and limit 32.00M exceeded scoped vmem limit by 1.56M.
```

Two observations:
- The transpose pipeline inside `mla_ragged_paged_attention` now allocates 33.56M of scoped vmem — just over the 32M default — on the *basic* test case.
- The error reports the **32M default** limit, i.e. the 64M budget that #3159 scopes to the MLA kernel tests via module-top `LIBTPU_INIT_ARGS` was not in effect for this compile (measured fact from the log; I haven't root-caused why — possibly the transpose `pallas_call` carries its own `vmem_limit_bytes`).

Deterministic (fails identically across independent builds/platforms), so retries don't help. cc MLA owners.

Xhark:
I saw similar testing error while I've run some tests with several version of containers in this PR, but latest tests right before merge this PR have no issue. 

This PR doesn't contain any changes for `xpose_pipeline` which is direct root cause of VMEM OOM.

I guess some tests vmem limitation may related to jax version (compiler version?), but may need to specify why that happened. I was trying to reproduce VMEM OOM on my GCP machine before, but test was just passed.

PR #620, opened 2025-09-02
[Torchax][Kernel] Support FP8 Dynamic And Static Quantization

kyuyeunk:
# Description

- Following PR adds support for running both FP8 static and FP8 dynamic models in torchax
	- Please note that this requires torchax version to be updated to include fp8 support: https://github.com/pytorch/xla/pull/9573 
- Using model [RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic](https://huggingface.co/RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic) in Ironwood, I have verified that it's running correctly.
	- Numeric validation
		- Command: `TPU_BACKEND_TYPE=jax MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=RedHatAI/Meta-Llama-3.1-8B-Instruct-FP8-dynamic --tensor_parallel_size=2 --task=generate --max_model_len=1024 --download_dir=/mnt/disks/persist`
		- Result: http://gpaste/6602029312770048
	- Performance validation
		- Baseline ([meta-llama/Llama-3.1-70B-Instruct](https://huggingface.co/meta-llama/Llama-3.1-70B-Instruct)): 6.52 reqs/s
		- FP8 ([RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic](https://huggingface.co/RedHatAI/Meta-Llama-3.1-70B-Instruct-FP8-dynamic)): 9.55 reqs/s (+46%)
- Limitation / Future work
	- This adds support of two types of FP8 models: FP8 static (where input scales are loaded from checkpoint) and FP8 dynamic (where input scales are computed during inference time). However, only FP8 dynamic uses the kernel and FP8 static uses XLA. Meaning, even though FP8 static should theoretically be faster, due to the lack of kernel support, currently FP8 dynamic is faster.

# Tests

```
pytest -v tests/models/vllm/layers/test_compressed_tensors_w8a8_fp8.py
```

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #618, opened 2025-09-02
[Torchax] Support Running AWQ models

kyuyeunk:
# Description

- Following PR adds support for running loading & running AWQ models in torchax
  - Please note that this requires a fix in vLLM repo to be submitted as well: https://github.com/vllm-project/vllm/pull/23902
- Using model [Qwen/Qwen2.5-32B-Instruct-AWQ](https://huggingface.co/Qwen/Qwen2.5-32B-Instruct-AWQ), I have verified that it'ss running correctly.
	- Numeric validation
		- Command `TPU_BACKEND_TYPE=jax MODEL_IMPL_TYPE=vllm python3 examples/offline_inference.py --model=Qwen/Qwen2.5-32B-Instruct-AWQ --tensor_parallel_size=8 --task=generate --max_model_len=1024 --download_dir=/mnt/disks/persist`
		- Result: http://gpaste/6205721003425792
	- Performance validation (degradation over baseline is expected)
		- Baseline ([Qwen/Qwen2.5-32B-Instruct](https://huggingface.co/Qwen/Qwen2.5-32B-Instruct)): 12.61 reqs/s
		- AWQ ([Qwen/Qwen2.5-32B-Instruct-AWQ](https://huggingface.co/Qwen/Qwen2.5-32B-Instruct-AWQ)): 11.34 reqs/s
- Limitation / Future work
	- AWQ uses w4a16 subchannel quantization, and for initial implementation, I have utilized XLA instead of kernel. However, there is known performance issue with subchannel quantization in XLA. Therefore, we need to implement w4a16 subchannel support into the kernel for optimal performance.

# Tests

```
pytest -v tests/models/vllm/layers/test_awq.py
```

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3036, opened 2026-06-28
[Bugfix] SC dense_gather_reduce: fall back when output block has zero rows (fixes Qwen3-30B-A3B accuracy)

QiliangCui2023:
## Problem
The nightly `tpu6e Accuracy for Qwen/Qwen3-30B-A3B` has failed on **main** for many nightlies. The engine fails to initialize during AOT lowering with:
```
ValueError: Invalid shape for `swap`. Ref shape: (2, 0, 2048). Expected shape: (0, 32). Value shape: (1, 32).
```

## Root cause
#2979 wired the SparseCore `dense_gather_reduce` kernel into the MoE GMM path (`fused_moe_gmm.py:moe_gmm_local`). The kernel's output `BlockSpec` row dimension is `(num_lanes // reduce_group_size) // packing` (see `out_rows_per_step` / `out_specs` in `_sc_gather_reduce`), where `packing = 32 // bits(dtype)` (bf16 -> 2, f32 -> 1). When `num_lanes // reduce_group_size < packing` this floors to **0**, producing a zero-height output block that the kernel body (which still writes one row) cannot store into.

Qwen3-30B-A3B has topk=8; on v6e SparseCore `num_lanes = 8`, so bf16 gives `8 // 8 // 2 = 0` -> crash. v7x has `num_lanes = 16` (`16 // 8 // 2 = 1`) and is unaffected (matches the tpu6e-only failure). Both `num_lanes` values were confirmed empirically (see Test evidence).

`is_compatible()` already gates a JAX fallback (`_jax_fallback`) but did not check this packing constraint.

Bisected to #2979 with a commit-vs-parent A/B (same vLLM LKG): the Qwen3-30B-A3B tpu6e accuracy test fails at `8f92c3098` and passes at its parent `455f99882`.

## Fix
Gate on the **arithmetic** condition `(num_lanes // reduce_group_size) // packing >= 1` rather than the TPU generation. This fixes the crash (the degenerate config falls back to `_jax_fallback`, the pre-#2979 path) while keeping the kernel for configs that leave >= 1 packed output row -- including valid v6e configs (e.g. f32 with packing=1, or smaller `reduce_group_size`). An earlier revision excluded all of v6e; that was broader than the defect (per review) and needlessly disabled the kernel where it works.

## Test evidence
All on the dev pipeline (full URLs):

**Empirical num_lanes** — https://buildkite.com/tpu-commons/tpu-inference-dev/builds/616
- v6e: `num_lanes=8` -> `8 // 8 // 2 = 0` (degenerate); v7x: `num_lanes=16` -> `16 // 8 // 2 = 1` (ok).

**v6e**
- https://buildkite.com/tpu-commons/tpu-inference-dev/builds/618 — `dense_gather_reduce` unit tests + `Qwen3-30B-A3B` accuracy (TP=8): green (model now falls back, no `swap` crash).
- https://buildkite.com/tpu-commons/tpu-inference-dev/builds/617 — `dense_gather_reduce` kernel correctness test run **on v6e** (guard-permitted configs exercise the real kernel and match the reference) + Qwen3-30B-A3B accuracy: green. Confirms the kernel is correct on v6e for permitted configs, so the precise guard (not a blanket v6e block) is the right boundary.

**v7x**
- https://buildkite.com/tpu-commons/tpu-inference-dev/builds/620 — `dense_gather_reduce` kernel correctness + `Qwen3-30B-A3B` accuracy (kernel path): green. Confirms the guard does not regress the kernel where it is valid.

## Tests
- `IsCompatibleTest` (hardware-independent, mocks `get_tpu_info`) covers the packing boundary: v6e-lanes+bf16+topk8 -> False; v6e-lanes+f32+topk8 -> True; v6e-lanes+bf16+topk4 -> True; v7x-lanes+bf16+topk8 -> True. Passes on CPU.

## Backport
Targets `main`. The same failure is on `releases/v0.24.0`; will cherry-pick there after this merges.

### DIFF
--- a/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
+++ b/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
@@ -41,24 +41,26 @@ def is_compatible(
     if op.shape[0] % reduce_group_size != 0:
         return False
 
-    tpu_info = pltpu.get_tpu_info()
-    # dense_gather_reduce is not supported on TPUv6e (matches the skip in
-    # tests/kernels/dense_gather_reduce_test.py). Concretely the kernel's output
-    # BlockSpec row dimension is (num_lanes // reduce_group_size) // packing (see
-    # out_rows_per_step / out_specs in _sc_gather_reduce); for sub-32-bit dtypes
-    # packing > 1 (bf16 -> 2), so v6e SparseCore with reduce_group_size close to
-    # num_lanes (e.g. Qwen3-30B-A3B topk=8) floors this to 0, yielding a
-    # zero-height output block and crashing AOT lowering with "Invalid shape for
-    # `swap`". Excluding v6e here routes those models through _jax_fallback (the
-    # pre-#2979 path) instead of the MoE GMM path tripping the crash.
-    if tpu_info.generation == 6:
-        return False
-
-    sc_info = tpu_info.sparse_core
+    sc_info = pltpu.get_tpu_info().sparse_core
     if sc_info is None:
         return False
 
     if sc_info.num_lanes % reduce_group_size != 0:
+        return False
+
+    # The kernel's output BlockSpec row dimension is
+    # (num_lanes // reduce_group_size) // packing (see out_rows_per_step and the
+    # out_specs in _sc_gather_reduce). For sub-32-bit dtypes packing > 1 (bf16 ->
+    # 2), so when num_lanes // reduce_group_size < packing this floors to 0,
+    # producing a zero-height output block that the kernel body (which still
+    # writes one row) cannot store into -> AOT lowering crashes with "Invalid
+    # shape for `swap`". This is exactly what regressed Qwen3-30B-A3B (topk=8,
+    # bf16) on v6e SparseCore (num_lanes=8 -> 8//8//2 = 0). Gate on the arithmetic
+    # rather than the TPU generation: configs that leave >=1 packed output row
+    # (larger num_lanes, smaller reduce_group_size, or f32 with packing=1) keep
+    # using the kernel; only the degenerate ones fall back to _jax_fallback.
+    packing = 32 // jax.dtypes.itemsize_bits(op.dtype)
+    if (sc_info.num_lanes // reduce_group_size) // packing < 1:
         return False
 
     num_cores = 1 if single_sc else sc_info.num_cores
--- a/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
+++ b/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
@@ -48,17 +48,8 @@ def is_compatible(
     if sc_info.num_lanes % reduce_group_size != 0:
         return False
 
-    # The kernel's output BlockSpec row dimension is
-    # (num_lanes // reduce_group_size) // packing (see out_rows_per_step and the
-    # out_specs in _sc_gather_reduce). For sub-32-bit dtypes packing > 1 (bf16 ->
-    # 2), so when num_lanes // reduce_group_size < packing this floors to 0,
-    # producing a zero-height output block that the kernel body (which still
-    # writes one row) cannot store into -> AOT lowering crashes with "Invalid
-    # shape for `swap`". This is exactly what regressed Qwen3-30B-A3B (topk=8,
-    # bf16) on v6e SparseCore (num_lanes=8 -> 8//8//2 = 0). Gate on the arithmetic
-    # rather than the TPU generation: configs that leave >=1 packed output row
-    # (larger num_lanes, smaller reduce_group_size, or f32 with packing=1) keep
-    # using the kernel; only the degenerate ones fall back to _jax_fallback.
+    # The output block has (num_lanes // reduce_group_size) // packing rows;
+    # fall back to JAX when that is 0 (the kernel can't emit a zero-row block).
     packing = 32 // jax.dtypes.itemsize_bits(op.dtype)
     if (sc_info.num_lanes // reduce_group_size) // packing < 1:
         return False
--- a/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
+++ b/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
@@ -41,7 +41,20 @@ def is_compatible(
     if op.shape[0] % reduce_group_size != 0:
         return False
 
-    sc_info = pltpu.get_tpu_info().sparse_core
+    tpu_info = pltpu.get_tpu_info()
+    # dense_gather_reduce is not supported on TPUv6e (matches the skip in
+    # tests/kernels/dense_gather_reduce_test.py). Concretely the kernel's output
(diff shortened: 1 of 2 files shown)


QiliangCui2023:
Validated on the dev pipeline: [tpu-inference-dev #605](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/605) — green.
- `tpu6e dense_gather_reduce unit test` (incl. new `IsCompatibleTest`): passed
- `tpu6e Qwen3-30B-A3B accuracy` (TP=8, the previously-failing config): passed

The v6e exclusion routes Qwen3-30B-A3B through `_jax_fallback` and the AOT `swap` crash is gone. Still DO NOT MERGE pending review.

tpu_inference/kernels/sparse_core/dense_gather_reduce.py
```diff
@@ -41,7 +41,20 @@ def is_compatible(
...
+    # dense_gather_reduce is not supported on TPUv6e (matches the skip in
+    # tests/kernels/dense_gather_reduce_test.py). Concretely the kernel's output
+    # BlockSpec row dimension is (num_lanes // reduce_group_size) // packing (see
+    # out_rows_per_step / out_specs in _sc_gather_reduce); for sub-32-bit dtypes
+    # packing > 1 (bf16 -> 2), so v6e SparseCore with reduce_group_size close to
+    # num_lanes (e.g. Qwen3-30B-A3B topk=8) floors this to 0, yielding a
+    # zero-height output block and crashing AOT lowering with "Invalid shape for
+    # `swap`". Excluding v6e here routes those models through _jax_fallback (the
+    # pre-#2979 path) instead of the MoE GMM path tripping the crash.
+    if tpu_info.generation == 6:
```
QiliangCui2023: **nit (boundary choice).** `generation == 6` is coarser than the precise crash condition `(num_lanes // reduce_group_size) // packing >= 1`. For this release fix that's the right call — it matches the kernel's own `dense_gather_reduce is not supported on TPUv6e` skip in `tests/kernels/dense_gather_reduce_test.py` and the crash is v6e-only. If we later want to keep the kernel on any future *valid* v6e shape, the precise packing guard could be added on `main` as a follow-up. No change needed here.

tpu_inference/kernels/sparse_core/dense_gather_reduce.py
```diff
@@ -41,7 +41,20 @@ def is_compatible(
...
+        return False
```
QiliangCui2023: **nit.** This also routes f32 on v6e (packing=1 -> no zero-height block, so it never trips the `swap` crash) through `_jax_fallback` — a perf-only fallback with no correctness change, and consistent with treating the kernel as unsupported on v6e. Worth a one-line mention in the PR description so the f32-on-v6e perf implication is explicit.

QiliangCui2023 (commented):
Self-review of the v6e exclusion fix.

**Verdict: approve.** Excluding generation 6 is the correct, conservative boundary for a release branch — it honors the kernel's own `dense_gather_reduce is not supported on TPUv6e` skip and routes the crashing MoE shapes to the verified `_jax_fallback`.

Verified (no blocking issues):
- Guard placement is correct (after the dtype / `shape[0] % reduce_group_size` checks, before `sc_info is None`) and folds the previous double `get_tpu_info()` into a single call.
- `_jax_fallback` is a correct reference path (gather -> reshape `(-1, reduce_group_size, hidden)` -> fp32 weighted sum -> cast back); its fp32-accumulation comment is accurate.
- `IsCompatibleTest` exercises the new gate, not just import: `test_v6e_falls_back` uses `num_lanes=16, rgs=8` — a config that passes every *other* gate (16//8=2, 2//2=1), so `False` is caused solely by the gen-6 check; `test_v7x_compatible` is the matching gen-7 control.

Two non-blocking notes inline.

tpu_inference/kernels/sparse_core/dense_gather_reduce.py
QiliangCui2023: Acknowledged — keeping the generation-6 exclusion for this release fix (matches the kernel's own v6e test skip; crash is v6e-only). Noted the precise packing guard `(num_lanes // reduce_group_size) // packing >= 1` as a possible `main` follow-up. No code change.

tpu_inference/kernels/sparse_core/dense_gather_reduce.py
QiliangCui2023: Good point — added an explicit note to the PR description: excluding v6e also routes f32-on-v6e (packing=1, which never trips the zero-height `swap` crash) through `_jax_fallback`; that is a perf-only change with no correctness impact, consistent with treating the kernel as unsupported on v6e.

QiliangCui2023:
Retargeted this PR from `releases/v0.24.0` to **`main`** — the failure reproduces on `main` (it is the source of the release-branch failure), so it should be reviewed/tested on `main` and then cherry-picked to `releases/v0.24.0`. The branch was rebased onto the latest `main` (`f28fce6e`); the diff is unchanged. Re-validating on internal CI against `main`'s current vLLM LKG.

QiliangCui2023:
Updated per review (the `generation == 6` boundary was too broad): switched to the precise arithmetic gate `(num_lanes // reduce_group_size) // packing >= 1`. This still fixes the Qwen3-30B-A3B crash (v6e bf16 topk=8 -> 8//8//2 = 0 -> fallback) but keeps the kernel on valid v6e configs (e.g. f32 packing=1, or smaller topk). Unit tests now cover the packing boundary. Re-validating on internal CI.

QiliangCui2023:
Verification complete on **both v6e and v7x** (precise packing guard). v6e: [dev #618](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/618) (Qwen3-30B-A3B accuracy + unit) and [dev #617](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/617) (kernel correctness on v6e + fallback); v7x: [dev #620](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/620) (kernel correctness + accuracy). num_lanes confirmed v6e=8 / v7x=16 via [dev #616](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/616). Shortened the inline comment; the change is comment-only so the validations above still apply.

BirdsOfAFthr (approved):
Maybe update the PR title since we are not excluding v6e?

QiliangCui2023:
Thanks for the review! Updated the title to drop "exclude TPUv6e" (the gate is now the precise packing arithmetic, not a v6e exclusion). Also fixed the failing `pre-commit` check (yapf formatting in the unit test) — pushed; the change is formatting-only.

guowei-dev (approved):
LGTM as an immediate fix. Moreover, instead of skipping the kernel, we could probably keep it when `num_lanes // reduce_group_size < packing` by buffering the output across packing row-subchunk steps and storing one whole int32 row per write. Worth exploring later if the SC kernel beats the fallback a lot for that config. cc @BirdsOfAFthr

kyuyeunk (approved):

PR #2630, opened 2026-05-14
[CI] Fix `test_multi_modal_inference.py`

ShobhitBehl:
# Description

Fix `pass_config.enable_qk_norm_rope_fusion` pydantic ValidationError (None not valid bool) occuring on v6 and v7  when running `test_multi_modal_inference.py`

# Tests

 - Locally on v7: `python3 -m pytest -s -v tests/e2e/test_multi_modal_inference.py`
 - CI 

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #611, opened 2025-08-29
[Kernel] Improve logic for finding TPU version number

kyuyeunk:
# Description

Use regex pattern matching to find TPU version number.

As a future TODO, consider using same `get_tpu_version` function for both quantized matmul and RPA kernel.

# Tests

https://buildkite.com/tpu-commons/tpu-commons-ci/builds/2486

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3037, opened 2026-06-28
[CI] Multimodal test: release TPU between parametrizations

QiliangCui2023:
## Problem
`tpu6e/tpu7x Correctness tests for Multimodal Inputs` have failed on **main** since ~Jun 25. The `enable_dynamic_image_sizes=True` case fails at engine init:
```
jax.errors.JaxRuntimeError: ABORTED: The TPU is already in use by process with pid <N>.
RuntimeError: Engine core initialization failed.
FAILED tests/e2e/test_multi_modal_inference.py::test_multi_modal_inference[True]
```

## Root cause
`test_multi_modal_inference` is parametrized over `enable_dynamic_image_sizes=[False, True]` and pytest runs both in one process. The test never tore down the first `LLM`, relying on GC/process-exit to free the TPU. A vLLM change altered engine-shutdown timing so the first engine's worker keeps the TPU when the second `LLM(...)` starts -> "TPU already in use".

Bisected on the dev pipeline by holding tpu-inference code fixed and moving only the pinned vLLM LKG: the multimodal test passes at LKG `0bc479e6` and fails at LKG `e7df2322` (the bump in `[skip ci] Update vLLM LKG` commit `678e023ed`). The regression is therefore in vLLM range `0bc479e6..e7df2322` (15 commits); a separate vLLM-side follow-up can pin the exact commit. The prior multimodal test fixes (#2782, #2630, #2443, #2371) were all about expected-text/similarity, not teardown -- this is a new failure mode.

## Fix
Wrap the run in `try/finally` and explicitly release the TPU after each parametrization via `llm.llm_engine.engine_core.shutdown()` + `gc.collect()`, mirroring `tests/e2e/test_continue_decode.py`. The shutdown is itself wrapped in `try/except` (+ traceback) so a teardown failure cannot mask the body's similarity assertion.

## Test evidence
Full multimodal test (both `enable_dynamic_image_sizes` parametrizations) on the dev pipeline against `main`'s current vLLM LKG (which includes the upstream regression), current PR HEAD:
- **v6e**: https://buildkite.com/tpu-commons/tpu-inference-dev/builds/615 — green.
- **v7x** (tp=2): https://buildkite.com/tpu-commons/tpu-inference-dev/builds/619 — green.

## Backport
Targets `main`. The same failure is on `releases/v0.24.0`; will cherry-pick there after this merges.

### DIFF
--- a/tests/e2e/test_multi_modal_inference.py
+++ b/tests/e2e/test_multi_modal_inference.py
@@ -120,8 +120,8 @@ def test_multi_modal_inference(monkeypatch, enable_dynamic_image_sizes):
 
         # Check output against the closest known-good caption.
         similarity_score = max(
-            difflib.SequenceMatcher(None, generated_text, expected,
-                                    autojunk=False).ratio()
+            difflib.SequenceMatcher(
+                None, generated_text, expected, autojunk=False).ratio()
             for expected in EXPECTED_TEXTS)
         print(f"Similarity Score: {similarity_score:.4f}")
         assert similarity_score >= 0.85, (
--- a/tests/e2e/test_multi_modal_inference.py
+++ b/tests/e2e/test_multi_modal_inference.py
@@ -5,6 +5,7 @@
 # compares the output to a known-good output.
 
 import difflib
+import gc
 import os
 from dataclasses import asdict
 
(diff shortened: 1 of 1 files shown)


QiliangCui2023:
tpu7x cross-check also green: [tpu-inference-dev #611](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/611) — `tpu7x Multimodal Inputs` (tp=2, both parametrizations) passed. The teardown fix is validated on both tpu6e ([#610](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/610)) and tpu7x. Still DO NOT MERGE pending review.

tests/e2e/test_multi_modal_inference.py
```diff
@@ -91,37 +92,47 @@ def test_multi_modal_inference(monkeypatch, enable_dynamic_image_sizes):
...
+        similarity_score = max(
+            difflib.SequenceMatcher(None, generated_text, expected,
+                                    autojunk=False).ratio()
+            for expected in EXPECTED_TEXTS)
+        print(f"Similarity Score: {similarity_score:.4f}")
+        assert similarity_score >= 0.85, (
+            f"Text similarity too low ({similarity_score:.2f}).\n"
+            f"Expected one of: {EXPECTED_TEXTS}\n"
+            f"Actual:   {generated_text}")
+    finally:
```
QiliangCui2023, line 131: **minor (reuse).** This teardown block is now duplicated from `tests/e2e/test_continue_decode.py`. Fine for a targeted release fix; as a `main` follow-up, consider lifting it into a shared autouse fixture or a `managed_llm(**engine_args)` context manager so future e2e tests get TPU release for free.

tests/e2e/test_multi_modal_inference.py
```diff
@@ -91,37 +92,47 @@ def test_multi_modal_inference(monkeypatch, enable_dynamic_image_sizes):
...
+        # Release the TPU before the next parametrization builds a new engine.
+        # This test is parametrized over enable_dynamic_image_sizes=[False, True]
+        # and pytest runs both in one process; without an explicit engine
+        # shutdown the first engine's worker keeps the TPU and the second
+        # LLM(...) fails with "TPU already in use by pid <first>". Mirrors the
+        # teardown in tests/e2e/test_continue_decode.py.
+        llm.llm_engine.engine_core.shutdown()
```
QiliangCui2023: **minor.** If `shutdown()` raises here it will mask a body `AssertionError` (the similarity check) — the standard `try/finally` trade-off, inherited from the mirrored pattern. Optional hardening: wrap the shutdown in its own `try/except` (log + swallow) so the original assertion always surfaces as the test failure.

tests/e2e/test_multi_modal_inference.py
```diff
@@ -91,37 +92,47 @@ def test_multi_modal_inference(monkeypatch, enable_dynamic_image_sizes):
...
+        gc.collect()
```
QiliangCui2023, line 148: **nit.** After this explicit shutdown, the `llm` finalizer may call shutdown again at GC / interpreter exit. Low risk — vLLM's client shutdown is idempotent and this was validated green — so no action; noting for completeness.

QiliangCui2023 (commented):
Self-review of the multimodal teardown fix.

**Verdict: approve-with-nits.** `try/finally` placement is correct — `llm = LLM(**engine_args)` is constructed *outside* the `try`, so a construction failure won't enter `finally` with an unbound `llm` — and `shutdown()` + `gc.collect()` empirically releases the TPU for both parametrizations. The verification/assert stays inside `try`, so the test is still meaningful. Only soft items remain (inline).

One description-level note: the Validation link points to the `tpu-commons` Buildkite org, which external reviewers of this public PR can't open. Consider summarizing the result inline (e.g. "both parametrizations green on internal CI") the way #3036's description does, rather than relying on the link.

tests/e2e/test_multi_modal_inference.py
QiliangCui2023: Acknowledged — for this targeted release fix I'm keeping the inline teardown that mirrors `test_continue_decode.py`. Lifting it into a shared autouse fixture / `managed_llm(...)` context manager is a good `main` follow-up so future e2e tests get TPU release for free.

tests/e2e/test_multi_modal_inference.py
QiliangCui2023: Done in 72e0f1c07 — wrapped `engine_core.shutdown()` in `try/except` + `traceback.print_exc()` so a teardown failure surfaces loudly instead of masking the body's similarity `AssertionError`. Re-validated on the dev pipeline (tpu6e, both parametrizations).

tests/e2e/test_multi_modal_inference.py
QiliangCui2023: Acknowledged — no action. vLLM's client shutdown is idempotent so a finalizer re-shutdown at GC/exit is harmless; validated green. Noting for completeness.

QiliangCui2023:
Retargeted this PR from `releases/v0.24.0` to **`main`** — the failure reproduces on `main` (it is the source of the release-branch failure), so it should be reviewed/tested on `main` and then cherry-picked to `releases/v0.24.0`. The branch was rebased onto the latest `main` (`f28fce6e`); the diff is unchanged. Re-validating on internal CI against `main`'s current vLLM LKG.

QiliangCui2023:
Verification complete on **both v6e and v7x** on main at current HEAD (full multimodal test, both parametrizations): v6e [dev #615](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/615), v7x (tp=2) [dev #619](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/619).

kwang3939 (approved):

pv97 (approved):

PR #3057, opened 2026-07-01
[MTP] Mamba State Rollback for Speculative Decoding

Lumosis:
# Description

This PR implements and optimizes the Mamba State Rollback mechanism required to support Speculative Decoding in Mamba-based models.

Unlike Transformer KV caches that can be truncated by adjusting token pointers, Mamba models maintain a continuous internal recurrent and convolutional state. During speculative decoding verification, the target model evaluates multiple draft tokens in parallel. If a draft token is rejected, the target model's internal state must be cleanly rolled back to exactly match the last accepted token.

This PR introduces the infrastructure to track intermediate state histories across draft tokens, execute device-side state rollback upon rejection, and correctly route execution between optimized parallel implementations and reference recurrent implementations.

# Key Changes

1. State History Materialization & Implementation Routing
* 2D State Indices Support: Modified the decode_only gated delta rule wrapper to support 2D state_indices.
* Target Model Routing: Speculative verification steps are explicitly routed to the ref_recurrent_impl (which runs token-by-token), enabling the layer to continuously build and materialize intermediate state history.

2. Device-Side Rollback Mechanism
* Rollback Kernel (_rollback_mamba_layer_states_fn): Implemented a device-side JAX function that securely copies the accepted state back to the context slot.
* Slot Architecture: Expanded Mamba slot allocation for speculative requests to num_spec + 2. This guarantees enough space for the initial context, all speculative steps, and evaluation.
* Rollback Logic: Unconditionally evaluates the target state rollback index as num_accepted_tokens + 1, successfully restoring the final accepted state to global Slot 0.

3. Prompt Prefill Optimizations
* Chunked Implementation Routing: Speculative prompt prefill is now explicitly routed to the optimized chunked implementation, completely bypassing the slower token-by-token recurrent implementation.
* Seamless Rollback: During prefill (where num_accepted = 0), the rollback safely executes a copy from Slot 1 back to Slot 0 without introducing no-op branches or corrupting the prefilled state.

# Tests

Tested with num_spec_tokens=4:
```
Successful requests:                     100       
Benchmark duration (s):                  150.55    
Total input tokens:                      72560     
Total generated tokens:                  25600     
Request throughput (req/s):              0.66      
Output token throughput (tok/s):         170.04    
Peak output token throughput (tok/s):    54.00     
Peak concurrent requests:                100.00    
Total token throughput (tok/s):          652.01    
Mean TTFT (ms):                          75189.89  
Median TTFT (ms):                        75636.58  
P99 TTFT (ms):                           148036.82 
Mean TPOT (ms):                          5.31      
Median TPOT (ms):                        5.36      
P99 TPOT (ms):                           6.80      
Mean ITL (ms):                           19.25     
Median ITL (ms):                         19.31     
P99 ITL (ms):                            20.48     
---------------Speculative Decoding---------------
Acceptance rate (%):                     66.24     
Acceptance length:                       3.65      
Drafts:                                  7034      
Draft tokens:                            28136     
Accepted tokens:                         18636     
Per-position acceptance (%):
  Position 0:                            82.12     
  Position 1:                            70.29     
  Position 2:                            60.12     
  Position 3:                            52.42     
```

# Checklist

### DIFF
--- a/tests/kernels/causal_conv1d_test.py
+++ b/tests/kernels/causal_conv1d_test.py
@@ -51,17 +51,8 @@ def reference_causal_conv1d(
         start_row = row - kernel_size + 1
         s_idx = int(row_to_seq_idx[row])
         seq_start = int(query_start_loc[s_idx])
-        token_step = row - seq_start
+        state_idx = int(state_indices[s_idx])
         has_init = bool(has_initial_state[s_idx])
-
-        # In speculative decoding (2D state_indices), read from the initial state (Slot 0)
-        # and write to the corresponding history slot for the current token step.
-        if state_indices.ndim == 2:
-            state_idx_read = int(state_indices[s_idx, 0])
-            state_idx_write = int(state_indices[s_idx, token_step + 1])
-        else:
-            state_idx_read = int(state_indices[s_idx])
-            state_idx_write = int(state_indices[s_idx])
 
         row_out = jnp.zeros_like(x[0], dtype=jnp.float32)
         window_vals = []
@@ -72,7 +63,7 @@ def reference_causal_conv1d(
             if idx < seq_start:
                 state_offset = idx - (seq_start - kernel_size + 1)
                 if has_init:
-                    val = conv_state[state_idx_read, state_offset]
+                    val = conv_state[state_idx, state_offset]
                 else:
                     val = jnp.zeros_like(x[0])
             else:
@@ -87,9 +78,9 @@ def reference_causal_conv1d(
 
         out_list.append(row_out.astype(x.dtype))
 
-        # Update the state at every token step to build the history
-        update = jnp.stack(window_vals[1:], axis=0)
-        new_conv_state = new_conv_state.at[state_idx_write].set(update)
+        if row == int(query_start_loc[s_idx + 1]) - 1:
+            update = jnp.stack(window_vals[1:], axis=0)
+            new_conv_state = new_conv_state.at[state_idx].set(update)
 
     out = jnp.stack(out_list, axis=0)
     return out, new_conv_state
--- a/tests/kernels/causal_conv1d_test.py
+++ b/tests/kernels/causal_conv1d_test.py
@@ -51,8 +51,15 @@ def reference_causal_conv1d(
         start_row = row - kernel_size + 1
         s_idx = int(row_to_seq_idx[row])
         seq_start = int(query_start_loc[s_idx])
-        state_idx = int(state_indices[s_idx])
+        token_step = row - seq_start
         has_init = bool(has_initial_state[s_idx])
+
+        if state_indices.ndim == 2:
+            state_idx_read = int(state_indices[s_idx, 0])
+            state_idx_write = int(state_indices[s_idx, token_step + 1])
+        else:
+            state_idx_read = int(state_indices[s_idx])
+            state_idx_write = int(state_indices[s_idx])
 
         row_out = jnp.zeros_like(x[0], dtype=jnp.float32)
         window_vals = []
@@ -63,7 +70,7 @@ def reference_causal_conv1d(
             if idx < seq_start:
                 state_offset = idx - (seq_start - kernel_size + 1)
                 if has_init:
-                    val = conv_state[state_idx, state_offset]
+                    val = conv_state[state_idx_read, state_offset]
                 else:
                     val = jnp.zeros_like(x[0])
             else:
@@ -78,9 +85,9 @@ def reference_causal_conv1d(
 
         out_list.append(row_out.astype(x.dtype))
 
-        if row == int(query_start_loc[s_idx + 1]) - 1:
-            update = jnp.stack(window_vals[1:], axis=0)
-            new_conv_state = new_conv_state.at[state_idx].set(update)
+        # Update the state at every token step to build the history
+        update = jnp.stack(window_vals[1:], axis=0)
+        new_conv_state = new_conv_state.at[state_idx_write].set(update)
 
     out = jnp.stack(out_list, axis=0)
     return out, new_conv_state
--- a/tests/kernels/causal_conv1d_test.py
+++ b/tests/kernels/causal_conv1d_test.py
@@ -54,6 +54,8 @@ def reference_causal_conv1d(
         token_step = row - seq_start
         has_init = bool(has_initial_state[s_idx])
 
+        # In speculative decoding (2D state_indices), read from the initial state (Slot 0)
+        # and write to the corresponding history slot for the current token step.
         if state_indices.ndim == 2:
             state_idx_read = int(state_indices[s_idx, 0])
             state_idx_write = int(state_indices[s_idx, token_step + 1])
--- a/tpu_inference/kernels/gdn/reference/ragged_gated_delta_rule_wrapper.py
+++ b/tpu_inference/kernels/gdn/reference/ragged_gated_delta_rule_wrapper.py
@@ -25,6 +25,8 @@ from tpu_inference.kernels.gdn.reference import \
 from tpu_inference.kernels.gdn.v2.gdn_decode_kernel import \
     ragged_gated_delta_rule_decode_only
 from tpu_inference.kernels.gdn.v2.recurrent_scan_v2 import recurrent_scan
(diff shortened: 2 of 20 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -1221,6 +1258,29 @@ def _update_placeholder(
 
         return placeholder_req_id_to_index
 
+    def _device_rollback_mamba_states(
```
gxd3, line 1269: nit: move this (and other mamba rollback related) functions to spec-decoding related src files instead of putting them in runner.py?

tpu_inference/runner/tpu_runner.py
```diff
@@ -328,7 +330,41 @@ def _subtract_num_rejected_tokens_fn(seq_lens: jax.Array, positions: jax.Array,
...
+        rollback_valid, num_rejected_tokens[rollback_subtract_indices], 0)
+    num_accepted_tokens_dev = rollback_draft_lengths - mapped_num_rejected
+
+    return seq_lens, positions, num_accepted_tokens_dev
+
+
+@jax.jit
+def _rollback_mamba_layer_states_fn(
+    conv_state: jax.Array,
+    recurrent_state: jax.Array,
```
gxd3, line 347: "donate" these 2 arguments for in-place update?

tpu_inference/runner/tpu_runner.py
```diff
@@ -2518,18 +2623,55 @@ def build_block_table_host(kv_cache_gid: int) -> None:
...
+                # we pass a 1D array pointing to `Slot 1` (since prefill's final state needs
+                # to end up in Slot 1 for the verification step's rollback to correctly copy it).
+                # This 1D array allows `ragged_gated_delta_rule_wrapper` to dynamically route
+                # to the highly optimized chunked implementation!
+                cached_reqs = getattr(scheduler_output,
+                                      "scheduled_cached_reqs", None)
+                num_cached = len(cached_reqs.req_ids
+                                 ) if cached_reqs is not None and hasattr(
+                                     cached_reqs, "req_ids") else 0
+                is_prompt_prefill = (len(scheduler_output.scheduled_new_reqs)
```
gxd3: qq: it means the current step contains at least 1 new (potentially chunked) prompt ?

tpu_inference/kernels/gdn/reference/ragged_gated_delta_rule_wrapper.py
```diff
@@ -145,6 +147,28 @@ def ragged_gated_delta_rule_wrapper(
     is_decode_only = distribution[0] == distribution[2]
 
     def decode_only_branch(_):
+        # For speculative decoding (2D state_indices), we force-route to the reference
+        # recurrent JAX implementation. While the chunked JAX kernel supports 2D indices,
+        # the fused decode kernel does not yet support reading from and writing to separate
+        # history slots, so we uniformly fallback here.
+        if state_indices.ndim == 2:
+            return ref_recurrent_impl(
```
gxd3: qq: will the jax based (instead of Pallas) implementation be too slow?

tpu_inference/runner/kv_cache_manager.py
```diff
@@ -384,7 +384,13 @@ def _maybe_set_compact_mamba_num_blocks_override(
         # block, rounded up to the sharding divisor. `runner.max_num_reqs`
         # already includes the DP multiplier
         # (= `dp_size × scheduler_config.max_num_seqs`).
-        mamba_num_blocks = self.runner.max_num_reqs + 1
+        # For speculative decoding, we allocate (num_spec + 1) blocks per request
+        # to store the state history, plus 1 block for the null slot.
+        if self.runner.speculative_config:
+            num_spec = self.runner.speculative_config.num_speculative_tokens
+            mamba_num_blocks = self.runner.max_num_reqs * (num_spec + 1) + 1
```
QiliangCui2023: **Block-count vs. stride mismatch → `IndexError` under high concurrency**

This budgets `max_num_reqs * (num_spec + 1) + 1` mamba blocks, but the slot pool in `input_batch.py` (both `__init__` and `init_mamba_pools`) is built with **stride `num_spec + 2`** — one physical slot per history column (Slot 0 for the initial state + Slots `1..num_spec+1` for the post-token states), i.e. `num_spec + 2` slots per request.

With only `num_spec + 1` blocks budgeted per request but a `num_spec + 2` stride, the free-list construction

```python
range(base + local_slots - stride, base, -stride)   # stride = num_spec + 2
```

yields ≈ `max_num_reqs * (num_spec + 1) / (num_spec + 2)` base slots — **fewer than `max_num_reqs`**. Once the number of concurrent mamba requests exceeds that, `add_request` calls `self._free_mamba_slots_per_rank[dp_rank].pop()` on an empty list → `IndexError` (no guard at the pop site).

Concretely for `max_num_reqs = 128, num_spec = 4`: the pool holds ≈ 106 request chunks. The benchmark in the PR description ran at peak concurrency **100 (< 106)**, which is why it didn't surface — it's latent above that. This only bites when compact-mamba sizing actually runs (real TPU with HBM); when the override is skipped the pool is sized from the larger uniform `num_blocks` and masks the deficit.

Suggested fix — make the budget match the stride:

```python
mamba_num_blocks = self.runner.max_num_reqs * (num_spec + 2) + 1
```

A unit test that feeds `init_mamba_pools` this same formula and then allocates `max_num_reqs` requests would keep the two sites from drifting (there's already spec-aware coverage in `tests/runner/test_input_batch.py`).

tests/kernels/causal_conv1d_test.py
```diff
@@ -51,9 +51,18 @@ def reference_causal_conv1d(
...
         seq_start = int(query_start_loc[s_idx])
-        state_idx = int(state_indices[s_idx])
+        token_step = row - seq_start
         has_init = bool(has_initial_state[s_idx])
 
+        # In speculative decoding (2D state_indices), read from the initial state (Slot 0)
+        # and write to the corresponding history slot for the current token step.
+        if state_indices.ndim == 2:
+            state_idx_read = int(state_indices[s_idx, 0])
+            state_idx_write = int(state_indices[s_idx, token_step + 1])
```
QiliangCui2023: **No test exercises the new 2D (speculative) path — this branch is never run**

This `ndim == 2` branch (and the per-step history-slot write just below it) is the reference model for the speculative path, but the test that drives `reference_causal_conv1d` only ever passes a **1D** `state_indices` (`jnp.arange(1, max_reqs + 1)`, ~line 113). So this branch — and the corresponding 2D kernel path in `causal_conv1d.py` (`read_state_indices`/`write_state_indices`, `write_history`, the prefill-anchor `b_idx_should_write` logic) — is never validated by any test.

More broadly, the PR adds the speculative Mamba-rollback machinery across several production files with no unit coverage for any of it:

- `runner/tpu_runner.py` — `_rollback_mamba_layer_states_fn`: the `source = num_accepted + 1` slot arithmetic and the prefill `Slot 1 → Slot 0` copy.
- `ragged_gated_delta_rule_ref.py` — 2D history materialization (read Slot `i` / write Slot `i+1`) and the prefill scratch-slot routing.
- `ragged_gated_delta_rule_chunked.py` — the 2D `decode_only` read/write split.
- `runner/input_batch.py` — the stride-based mamba slot allocation for speculative decode.
- `layers/common/utils.py` — the new `axis` parameter on `truncate_sharded_tensor`.

Per the repo's testing expectation for `tpu_inference/**`, each of these needs a unit test that actually exercises the new behavior. Concretely:

1. Extend this test with a 2D `state_indices` case and assert the kernel output/state matches this reference.
2. Add a numeric test for `_rollback_mamba_layer_states_fn` (a few requests, varying `num_accepted`; assert Slot 0 ends up equal to Slot `num_accepted + 1`).
3. Add an `input_batch` allocation test that feeds `init_mamba_pools` the `kv_cache_manager` block formula and then allocates `max_num_reqs` requests — this would also catch the `num_spec + 1` vs `num_spec + 2` mismatch flagged separately. `tests/runner/test_input_batch.py` already has spec-aware slot tests to build on.

kyuyeunk:
based on the code changes, is my understanding correct that this pr is still in prototyping stage? if so, can you convert it into draft to make it clear?

PR #3065, opened 2026-07-01
[DeepSeekv4 bingup] Have a dedicate model.py override for DSv4 in TPU

gxd3:
Have a dedicate model.py override for DSv4 in TPU: tpu_inference/models/vllm/experimental/deepseek_v4.py

This avoid the hacking packing in tpu_inference/models/vllm/vllm_model_wrapper.py to override AMD's DSv4 model.py

I copied and adjusted amd's model.py, and put it in tpu_inference/models/vllm/experimental/deepseek_v4.py:
key adjustments:
* we remove cudu.Stream etc

* set `apply_routed_scale_to_output` to True for `FusedMoE`, this is a MUST have for correctness in TPU code path

* inject `gate.tid2eid` to `experts.routed_experts.hash_indices_table` so that TPU's MOE override can pick up `hash_indices_table` from vLLM's `RoutedExperts` object in `tpu_inference/layers/vllm/interface/moe.py` . This is a MUST have for DSv4, as its bottom layer used token-id-hash based routing

### DIFF
--- a/tpu_inference/models/vllm/experimental/deepseek_v4.py
+++ b/tpu_inference/models/vllm/experimental/deepseek_v4.py
@@ -1,0 +1,801 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+import typing
+from collections.abc import Callable, Iterable
+from itertools import islice
+
+import regex as re
+import torch
+import torch.nn as nn
+from vllm.config import VllmConfig
+from vllm.distributed import (get_pp_group, get_tensor_model_parallel_rank,
+                              get_tensor_model_parallel_world_size)
+from vllm.model_executor.layers.activation import (SiluAndMul,
+                                                   SiluAndMulWithClamp)
+from vllm.model_executor.layers.fused_moe import (
+    FusedMoE, GateLinear, fused_moe_make_expert_params_mapping)
+from vllm.model_executor.layers.layernorm import RMSNorm
+from vllm.model_executor.layers.linear import (MergedColumnParallelLinear,
+                                               RowParallelLinear)
+from vllm.model_executor.layers.logits_processor import LogitsProcessor
+from vllm.model_executor.layers.mhc import (HCHeadOp, MHCFusedPostPreOp,
+                                            MHCPostOp, MHCPreOp)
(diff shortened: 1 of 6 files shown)


tpu_inference/models/vllm/experimental/deepseek_v4.py
```diff
@@ -0,0 +1,801 @@
...
+                            loaded_weight,
+                            name_mapped,
+                            shard_id=expert_shard_id,
+                            expert_id=expert_id,
+                            return_success=True,
+                        )
+                        if success:
+                            name = name_mapped
+                            break
+                    loaded_params.add(name_mapped)
```
inucool12, line 650: name_mapped is defined inside the loop but used outside it. If expert_mapping is ever empty or skips every iteration, I think this will result in a crash.

gxd3: this part of the code is copied from vLLM upstream directly: https://screenshot.googleplex.com/6xCk9FjjCJqhbkW
(i only added https://screenshot.googleplex.com/bhTduWGjWzKjTMB for loading subset of layers.)

> If expert_mapping is ever empty or skips every iteration, I think this will result in a crash.

probably it won't happen in practice i think

tpu_inference/models/vllm/experimental/deepseek_v4.py
```diff
@@ -0,0 +1,801 @@
...
+        x: torch.Tensor,
+        positions: torch.Tensor,
+        input_ids: torch.Tensor | None,
+        post_mix: torch.Tensor | None = None,
+        res_mix: torch.Tensor | None = None,
+        residual: torch.Tensor | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor | None, torch.Tensor | None,
+               torch.Tensor | None]:
+        if not self.has_tilelang:
+            return self._forward_unfused_post_pre(x, positions, input_ids,
```
inucool12: Since self.has_tilelang is hardcoded to False, do we still need this?

gxd3: make senses, removed death code associated with has_tilelang=True

inucool12 (approved):
LGTM

jrplatin (approved):

QiliangCui (approved):

PR #3067, opened 2026-07-01
[Torchax] Add compressed_tensor w4a4_nvfp4 support

lxhfirenking:
# Description

Add compressed tensor w4a4_nvfp4 support to the torchax path. 

The rationale is mainly to enable tpu-inference to load fp4 quantization checkpoints produced by LLM Compressor. 

In the original VLLM implementation, `compressed_tensors_w4a4_nvfp4.py` supports two cases: 
 - both weights and activation being nvfp4
 - weights are in nvfp4, but activations are unquantized bf16 (so it is really w4a16)

In this initial implementation, we do not try to quantize the activation, given how native fp4 x fp4 matmul is not yet supported on TPUs. A warning is produced if a nvfp4 compressed tensor contains config for activation quantization, which will then be ignored. 

Unit tests are added for the new method, along with utils method refactoring to avoid code duplication with existing test_nvfp4.py

# Tests

Unit test
```
pytest tests/layers/vllm/test_compressed_tensors_w4a4_nvfp4.py
```

Loading nvfp4 compressed_tensor checkpoint
```
USE_BATCHED_RPA_KERNEL=1 \
JITTED_MM_MODULE_KEYS="model.vision_tower.encoder" \
REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES="transformers.modeling_outputs.BaseModelOutputWithPast" \
MODEL_IMPL_TYPE="vllm" \
vllm serve \
    --max-num-seqs=127 \
    --max-model-len=4096 \
    --no-enable-prefix-caching  \
    --model=RedHatAI/gemma-4-31B-it-NVFP4 \
    --limit-mm-per-prompt '{"image": 8, "video": 0}' \
    --tensor-parallel-size=2 \
    --kv-cache-dtype=fp8
```
Benchmark was run against the launched model instance to verify that accuracy was as expected
```
    --backend vllm-chat \
    --model RedHatAI/gemma-4-31B-it-NVFP4 \
    --dataset-name mmmu_pro \
    --mmmu-pro-output-len=2000 \
    --run-eval \
    --num-prompts 1730 \
    --max-concurrency 32
```

# Checklist

### DIFF
--- a/tests/layers/vllm/nvfp4_utils.py
+++ b/tests/layers/vllm/nvfp4_utils.py
@@ -1,0 +1,98 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+import torch
+from torchax.ops.mappings import t2j
+
+# NVFP4 block size (elements per scale group).
+NVFP4_GROUP_SIZE = 16
+
+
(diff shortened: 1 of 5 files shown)


lk-chen (approved):

jrplatin:
Do we not need MoE support as well?

lxhfirenking:
> Do we not need MoE support as well?
Yes. That will be a separate PR.

jrplatin (approved):

PR #3070, opened 2026-07-01
skip multimodal if configuration disables it.

lc5211:
# Description

Still respect the `limit_mm_per_prompt` configuration from users when using jax native multimodal implementation.

# Tests

commands to reproduce.

# Checklist

Signed-off-by: lc5211 (linchai@google.com)


lk-chen (commented):
why is it necessary? for gemma4 , vision tower is ~600M but text backbone is ~31B. So we didn't care about vision weights, only disable mm [here](https://github.com/vllm-project/tpu-inference/blob/40b2b709c587f7ecad84efb00480ba7f487acca3/tpu_inference/runner/tpu_runner.py#L938)

An easier way is to set `--hf_overrides '{"architectures": ["Gemma4ForCausalLM"]}'` so we use the text-only model from `gemma4.py`

lc5211:
> why is it necessary? for gemma4 , vision tower is ~600M but text backbone is ~31B. So we didn't care about vision weights, only disable mm [here](https://github.com/vllm-project/tpu-inference/blob/40b2b709c587f7ecad84efb00480ba7f487acca3/tpu_inference/runner/tpu_runner.py#L938)
> 
> An easier way is to set `--hf_overrides '{"architectures": ["Gemma4ForCausalLM"]}'` so we use the text-only model from `gemma4.py`

With vision tower, we will need to respect the shape for its parameters when sharding the model. For example, previously I can place the gemma4-e2b-it model on to v6e-8 with model_shards=8. but enabling vision tower, I hit error of `RuntimeError: Failed to load weight 'vision_tower.layers.0.self_attn.k_proj.weight' with shape (768, 12, 64) into param with shape (768, 12, 64)`. I think we should provide the option to only enable text given that's the major usecase. 
But the suggestion of  `--hf_overrides '{"architectures": ["Gemma4ForCausalLM"]}'`  could work with some other changes as well. I updated the PR to make `Gemma4ForCausalLM` functionally correct. PTAL. Thanks!

tpu_inference/models/jax/utils/weight_utils.py
```diff
@@ -1073,7 +1073,8 @@ def load_weights(self, weights: Iterable[tuple[str,
             self,
             pytorch_pooler=pytorch_pooler,
             skip_prefixes=(["lm_head"]
-                           if not hasattr(self, 'lm_head') else None))
+                           if not hasattr(self, 'lm_head') else None),
+            ignore_unexpected_prefixes=getattr(self, "ignore_unexpected_prefixes", None))
```
lk-chen: Prefer follow [vllm](https://github.com/vllm-project/vllm/blob/d891b9bd51ce726a1bff24a72d03090057718789/vllm/model_executor/models/gemma4_unified.py#L452), specifically don't modify weight_utils.py, instead override `load_weights` in gemma4.py and pass `ignore_prefixes` there

lk-chen (commented):
LGTM mostly

tpu_inference/models/jax/utils/weight_utils.py
lc5211: Good suggestion. Skip the vision_tower with stripped_weights in Gemma4ForCausalLM then.

lk-chen (approved):

PR #3071, opened 2026-07-02
update docs for v0.24.0 release

CienetStingLin:
# Description

Update support matrices for v0.24.0 release.

# Tests

Docs update

# Checklist

### DIFF
--- a/docs/includes/nightly_parallelism.md
+++ b/docs/includes/nightly_parallelism.md
@@ -38,7 +38,7 @@
       <td>TP</td>
       <td><span title="✅&nbsp;Passing">✅</span></td>
       <td><span title="❓&nbsp;Untested">❓</span></td>
-      <td><span title="✅&nbsp;Passing">✅</span></td>
+      <td><span title="❌&nbsp;Failing">❌</span></td>
       <td><span title="❓&nbsp;Untested">❓</span></td>
     </tr>
     <tr>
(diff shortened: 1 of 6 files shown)


QiliangCui (approved):

docs/includes/nightly_parallelism.md
```diff
@@ -38,7 +38,7 @@
       <td>TP</td>
       <td><span title="✅&nbsp;Passing">✅</span></td>
       <td><span title="❓&nbsp;Untested">❓</span></td>
-      <td><span title="✅&nbsp;Passing">✅</span></td>
+      <td><span title="❌&nbsp;Failing">❌</span></td>
```
QiliangCui, line 41: @CienetStingLin , can you see why this become s failing?

CienetStingLin: This is because at the commit where the branch was created from `main`, the [nightly test result](https://github.com/vllm-project/tpu-inference/blob/update-docs-for-v0.24.0/support_matrices/nightly/v6e/vllm/parallelism_support_matrix.csv) for that item was failing. When `update_readme.py` is executed, it updates the entire README all at once.

Previously, the documentation was updated all at once. However, after giving it some thought, perhaps we only need to update the content related to the `release`, or we could just wait for the `nightly` job to automatically update the support matrices and it will return to normal. What do you think?

QiliangCui: If this is part of the release, it should reflect the release status ideally. let's create a bug to address it.

QiliangCui: Can you check the status of the release branch and see if we can manual change this?

cc: @bvrockwell .

PR #3072, opened 2026-07-03
[Feature] Support compressed-tensors quantization for the JAX path

### LINKED ISSUE #2261: [Feature]: Support compress-tensor quantization method for jax-path

Similar to #1623 , add "compressed-tensor" [here](https://github.com/vllm-project/tpu-inference/blob/a9c6227f5559b85275343633c28a924eecb86bc0/tpu_inference/layers/jax/quantization/__init__.py#L30-L37)

First target model could be one of 
- https://huggingface.co/RedHatAI/gemma-4-31B-it-FP8-block
- https://huggingface.co/RedHatAI/gemma-4-31B-it-FP8-Dynamic

### Alternatives

_No response_

### Additional context

_No response_

jerviscz:
# Description

Registers a JAX-native `compressed-tensors` quantization config so that fp8 compressed-tensors checkpoints (e.g. `RedHatAI/gemma-4-31B-it-FP8-block` / `-FP8-Dynamic`) can be loaded through the JAX path instead of a hand-written Qwix patch.

Previously the JAX quantization delegation map (`layers/jax/quantization/__init__.py`) only knew about `fp8` and unquantized checkpoints; a compressed-tensors checkpoint had no config to dispatch to. This PR adds that config and wires it into the map.

The new `CompressedTensorsConfig` holds a parsed upstream vLLM `CompressedTensorsConfig` and reuses its `config_groups` parsing (`target_scheme_map` / `ignore`) and scheme detection. `get_quant_method` then does a JAX-native per-layer dispatch: matched fp8 w8a8 layers are routed to the existing `Fp8BlockwiseLinearMethod` / `Fp8TensorwiseLinearMethod`, which already run fp8×fp8 on the MXU, so no new build/quantize/apply code is introduced.

Notable details:

- `Fp8BlockwiseLinearMethod` gains a `weight_scale_name` parameter, following the existing `Fp8FusedMoEMethod.weight_scale_name` pattern. The default (`"weight_scale_inv"`) keeps DeepSeek-style fp8 checkpoints unchanged; the compressed-tensors path passes `"weight_scale"`, which is the name those checkpoints serialize (verified against the gemma-4 FP8-block safetensors index). Without this, the block scales find no matching param and weight loading never completes. Both names are the same quantity (the dequant multiplier), so this is a pure rename — no reciprocal involved.
- Upstream `find_matched_target` needs a torch module (it matches on `module.__class__.__name__`), so per-layer matching is reimplemented with the upstream pure-string regex helpers plus a `JaxEinsum -> "Linear"` mapping.
- A small shim supplies `weight_block_size` to `Fp8BlockwiseLinearMethod` instead of fabricating a full `Fp8Config`.

**Scope / shortcomings (follow-ups):**

- Only fp8 w8a8 is supported (block-wise and per-tensor/channel weights, both with dynamic activations — matching the two target checkpoints). w4a8 / wNa16 schemes raise `NotImplementedError` and are left to a follow-up.
- Dense models only. Both named target checkpoints are dense (`enable_moe_block: false`, `num_experts: null` in their configs); MoE dispatch is intentionally deferred and will land with its own tests when there is a compressed-tensors MoE target.
- End-to-end loading/serving of the named checkpoints is not yet validated (the 31B models need a multi-chip slice). Two known unknowns remain for e2e: the per-channel scale shape on the FP8-Dynamic variant (`[out, 1]` in the checkpoint vs `(out,)` param), and fused-layer (gate_up/qkv) ignore semantics — `packed_modules_mapping` is not yet wired for the JAX path.
- Scheme detection reuses the private upstream `_is_fp8_w8a8` helper (deliberate coupling, kept identical to the torch path's behavior).

Related issue: #2261

# Tests

Added `tests/layers/jax/quantization/test_compressed_tensors.py` (mirrors `test_fp8.py::TestFp8Config`), asserting per-layer routing from a compressed-tensors `quantization_config`:

- fp8 block group -> `Fp8BlockwiseLinearMethod`
- fp8 per-tensor group -> `Fp8TensorwiseLinearMethod`
- `ignore` regex -> `UnquantizedLinearMethod` (per-layer: one layer skipped while its sibling stays quantized)
- blockwise scale param is created under the compressed-tensors name (`weight_scale`, not `weight_scale_inv`)
- upstream config parsing smoke test

Run:

```
pytest tests/layers/jax/quantization/test_compressed_tensors.py -v
```

The first four tests were validated on a v5e TPU VM (4 passed). The scale-name test was added afterwards together with the `weight_scale_name` change and has only been checked locally (syntax/lint); it will be exercised by CI. Existing `test_fp8.py` coverage guards the default-name (DeepSeek-style) path.

# Checklist

- [ ] I have made or will make corresponding changes to any relevant documentation. (No user-facing docs affected; internal quantization dispatch only.)

### DIFF
--- a/tpu_inference/layers/jax/quantization/compressed_tensors.py
+++ b/tpu_inference/layers/jax/quantization/compressed_tensors.py
@@ -1,0 +1,124 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""JAX-native ``compressed-tensors`` quantization config (issue #2261).
+
+Composes (does not subclass) the upstream vLLM ``CompressedTensorsConfig`` to
+reuse its config-group parsing and scheme detection, then dispatches each layer
+to the existing JAX fp8 quant methods.
+"""
+
+from types import MappingProxyType
+from typing import Optional
+
+from vllm.model_executor.layers.quantization.compressed_tensors.compressed_tensors import \
+    CompressedTensorsConfig as VllmUpstreamCTConfig
+from vllm.model_executor.layers.quantization.compressed_tensors.utils import (
+    check_equal_or_regex_match, should_ignore_layer)
+
+from tpu_inference.layers.jax import JaxModule
+from tpu_inference.layers.jax.linear import JaxEinsum
+from tpu_inference.layers.jax.quantization import QuantizeMethodBase
+from tpu_inference.layers.jax.quantization.configs import (QuantizationConfig,
+                                                           QuantLinearConfig)
+from tpu_inference.layers.jax.quantization.fp8 import (
+    Fp8BlockwiseLinearMethod, Fp8TensorwiseLinearMethod)
+from tpu_inference.layers.jax.quantization.unquantized import \
+    UnquantizedLinearMethod
+
+
+class _Fp8BlockConfigShim:
+    """Stand-in for Fp8Config passed to Fp8BlockwiseLinearMethod.
+
+    That method only reads ``quant_config.weight_block_size``, so exposing that
+    single attribute avoids fabricating a full Fp8Config.
+    """
+
+    def __init__(self, weight_block_size):
+        self.weight_block_size = weight_block_size
+
+
+def _weight_block_size(weight_quant) -> Optional[list[int]]:
+    """Return [block_n, block_k], or None if the weights are not block-quantized."""
+    block = getattr(weight_quant, "block_structure", None)
+    return list(block) if block is not None else None
+
+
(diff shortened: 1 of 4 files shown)


lk-chen (approved):

lk-chen:
can I load `RedHatAI/gemma-4-31B-it-FP8-block / -FP8-Dynamic` after this PR?

cc @lxhfirenking

lxhfirenking:
> can I load `RedHatAI/gemma-4-31B-it-FP8-block / -FP8-Dynamic` after this PR?
> 
> cc @lxhfirenking

Btw that checkpoint already works on the torchax path so we should be able to do a comparison across the two once the jax path is lit up

tpu_inference/layers/jax/quantization/compressed_tensors.py
```diff
@@ -0,0 +1,124 @@
...
+        # layers are named differently, so map JaxEinsum onto "Linear".
+        if isinstance(layer,
+                      JaxEinsum) and "Linear" in self._target_scheme_map:
+            return self._target_scheme_map["Linear"]
+        return None
+
+    def get_quant_method(self, layer: JaxModule,
+                         prefix: str) -> Optional[QuantizeMethodBase]:
+        if not isinstance(layer, JaxEinsum):
+            return None
```
lxhfirenking, line 90: We should ideally fail fast and crash, or at least produce a warning if the layer is quantized in the compressed tensor checkpoint but not yet supported by this class yet.

Otherwise the logic would be running the model in a quantizaiton state that is different from what is actually in the checkpoint, and the user would not know unless they read the actual implementation details.

Issue #967, opened 2025-10-29
[New Model]: Qwen3-Next Support


mokeddembillel:

mokeddembillel:
Hi guys, I'm working on this currently.

py4:
@mokeddembillel Thank you! We can help with the review once it's ready

bvrockwell:
Thanks @mokeddembillel ! cc @yaochengji (we chatted about this one earlier today, seems feasible + good first issue).

Akasxh:
@bvrockwell may I know more about this issue if it is still open?

PR #3073, opened 2026-07-04
[New Model] Add JAX-native Qwen3-Next support

bzantium:
Implements JAX-native (flax nnx) Qwen3-Next, closing #967.

Qwen3-Next is a hybrid model: three gated-delta-net (GDN) linear-attention layers per one gated full-attention layer, with a sparse MoE (512 experts, top-10) plus a sigmoid-gated shared expert on every layer. This adds the flax_nnx implementation; the torchax path already serves this model, so this follows CONTRIBUTING.md's preferred torchax then JAX-native progression.

## What's included

- `models/jax/qwen3_next.py`: GDN linear attention (reuses the existing v3 fused conv1d+GDN Pallas kernel via `run_jax_gdn_attention`), gated full attention with partial RoPE, sparse MoE with a shared expert.
- `runner/kv_cache_manager.py`: the JAX-native `get_kv_cache_spec` branch now emits `MambaSpec` for `linear_attention` layers, activating the existing backend-agnostic hybrid mamba runtime (tuple `(conv_state, ssm_state)` caches, `mamba_state_indices`, compact sizing). The module-independent core of the hybrid page-size reconciliation is split out of `update_mamba_page_size_padded` so both the torchax and JAX paths share it.
- HF checkpoint loading: zero-centered RMSNorm +1 folding, both fused-3D and per-expert MoE tensor layouts, `A_log`/`dt_bias` kept in float32, MTP tower skipped.
- Model registration plus per-layer jit out-shardings for the hybrid cache.
- Shared `models/common/layer_types.py` constants so the kv-cache spec, the jit out-sharding selection, and the decoder layer cannot silently desync on the HF layer-type strings.

## Testing

- CPU unit tests (`tests/models/jax/test_qwen3_next.py`) compare the GDN block, gated full attention, and the MoE block against HF transformers, and check end-to-end logits through a tiny checkpoint round-trip. Hybrid kv-cache spec tests live in `tests/runner/test_kv_cache_manager_jax_hybrid.py`.
- A TPU forward smoke test exercising the real GDN and ragged-paged-attention kernels is added and gated to run under the new `.buildkite/models` v6e CI entry.
- I have not run 80B-A3B accuracy or performance on real TPU hardware. The CI entry currently covers unit tests only; the accuracy and benchmark stages are left for follow-up once hardware thresholds are measured.

## Known follow-ups (not blocking)

- Promote the fused-3D expert handling into shared `JaxMoE` if other MoE models adopt that checkpoint layout.
- Derive the hybrid jit out-shardings from the kv-cache spec rather than re-scanning `layer_types`.
- Pipeline parallelism is disabled for this model for now (`_PP_DISABLED_MODELS`).

### DIFF
--- a/tests/models/jax/test_qwen3_next.py
+++ b/tests/models/jax/test_qwen3_next.py
@@ -361,7 +361,7 @@ class TestSparseMoeBlock:
                                                       prefix="mlp")
 
     def _copy_weights(self, block, hf):
-        block.gate.kernel_DE.value = jnp.asarray(
+        block.gate.weight.value = jnp.asarray(
             hf.gate.weight.detach().numpy().T)
         # HF stores fused 3D expert tensors: gate_up_proj (E, 2F, D) with
         # the gate rows first, down_proj (E, D, F).
@@ -373,11 +373,11 @@ class TestSparseMoeBlock:
             gate_up[:, intermediate:, :].transpose(0, 2, 1))
         block.experts.kernel_down_proj_EFD.value = jnp.asarray(
             hf.experts.down_proj.detach().numpy().transpose(0, 2, 1))
-        block.shared_expert_gate_proj.weight.value = jnp.asarray(
+        block.shared_expert.gate_proj.weight.value = jnp.asarray(
             hf.shared_expert.gate_proj.weight.detach().numpy().T)
-        block.shared_expert_up_proj.weight.value = jnp.asarray(
+        block.shared_expert.up_proj.weight.value = jnp.asarray(
             hf.shared_expert.up_proj.weight.detach().numpy().T)
-        block.shared_expert_down_proj.weight.value = jnp.asarray(
+        block.shared_expert.down_proj.weight.value = jnp.asarray(
             hf.shared_expert.down_proj.weight.detach().numpy().T)
         block.shared_expert_gate.weight.value = jnp.asarray(
             hf.shared_expert_gate.weight.detach().numpy().T)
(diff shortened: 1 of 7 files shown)


tests/models/jax/test_qwen3_next.py
```diff
@@ -0,0 +1,648 @@
...
+        np.testing.assert_allclose(np.asarray(
+            experts.kernel_down_proj_EFD.value),
+                                   down.transpose(0, 2, 1),
+                                   rtol=1e-6,
+                                   atol=1e-6)
+
+
+class TestCheckpointRoundTrip:
+
+    def test_tiny_checkpoint_load_and_forward(self, tmp_path, monkeypatch, rng,
```
lk-chen, line 465: other jax-model test usually just load ckpt and does a forward run.

does it make sense to load [Next-80B](https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct) but just first 4 layers like [this](https://github.com/vllm-project/tpu-inference/blob/62465576ed7c364384944c4e4b651eac4bb76aed/tests/models/jax/test_qwen3.py#L266-L267) (because full_attention_interval = 4, so loading 4 layers will cover GDN + normal attention). And do a [forward path](https://github.com/vllm-project/tpu-inference/blob/62465576ed7c364384944c4e4b651eac4bb76aed/tests/models/jax/test_qwen3.py#L318-L330) for smoke test

tpu_inference/models/jax/qwen3_next.py
```diff
@@ -0,0 +1,941 @@
...
+                    rngs=rng,
+                    prefix="lm_head",
+                )
+            else:
+                self.lm_head = PPMissingLayer()
+        else:
+            self.lm_head = PPMissingLayer()
+
+        for name, param in self.named_parameters():
+            if (name.endswith(_ZERO_CENTERED_NORM_SUFFIXES)
```
lk-chen: it's more readable to create new `JaxZeroCenteredRmsNorm` class inherit `JaxRmsNorm` here, and set "weight_loader" inside the class.

Then `input_layernorm` etc. can be created from new class.

tpu_inference/models/common/model_loader.py
lk-chen, line 59: plz add `Qwen3NextForCausalLM` here as well

bzantium: Added `Qwen3NextForCausalLM` to `_VLLM_PREFERRED_ARCHITECTURES` in fd3d69a9f. The torchax path stays the default until this JAX-native implementation is validated on hardware.

tpu_inference/models/jax/qwen3_next.py
bzantium: Done in fd3d69a9f. Added a `JaxZeroCenteredRmsNorm(JaxRmsNorm)` subclass that sets its own `weight_loader` to fold the `+1`, and `input_layernorm`, `post_attention_layernorm`, `q_norm`, `k_norm` and the final `norm` now use it. This drops the model level suffix list and the `named_parameters` scan.

tests/models/jax/test_qwen3_next.py
bzantium: Reworked the smoke test in fd3d69a9f to follow that convention. It now loads the real Qwen/Qwen3-Next-80B-A3B-Instruct config's first 4 layers (which cover 3 GDN plus 1 full attention, since full_attention_interval=4) via `skip_layers_model_loader_for_test` and runs a prefill forward.

One constraint worth flagging: the first layer is GDN, whose fused conv1d+delta-rule Pallas kernel only runs in interpret mode on CPU, so this forward test is gated to TPU (`skipif backend != tpu`). The block level numerics tests in this file cover CPU by substituting a pure JAX reference for that kernel.

tpu_inference/models/jax/qwen3_next.py
```diff
@@ -0,0 +1,942 @@
...
+
+
+class Conv1dWeight(JaxModule):
+    """Holds the depthwise causal conv weight under a `weight` param so its
+    parameter name lines up with the HF checkpoint (`...conv1d.weight`)."""
+
+    def __init__(self, conv_dim: int, kernel_size: int, dtype: jnp.dtype,
+                 rng: nnx.Rngs):
+        self.weight = create_param(rng, (conv_dim, 1, kernel_size),
+                                   ("model", None, None), dtype)
```
lk-chen, line 102: Similarly, we could set `_load_reordered_conv_weight` as weight_loader here. Plz also make _load_reordered_conv_weight a private method.

As convention, Plz put `Jax` prefix to class name

PR #3080, opened 2026-07-06
[Qwen3.5/InferenceX] Benchmark and Optimizations (2/N) 

wyzhang:
# Why 
A set of changes to optimize performance for comparing Qwen3.5 TPU against published GPU numbers in InferenceX.

# What 
(see details in each individual commit message) 
- **[Qwen3.5/InferenceX] R6: shrink rotary cos_sin_cache to max_model_len at load**
- **[Qwen3.5/InferenceX] R7: size max-num-seqs per DP rank + add the concurrency 4 attn bucket**

There is an R5 optimization that reduces host overhead—primarily for low concurrency workloads (4 and 8 clients). 
However, the change is somewhat involved and non-trivial to gate behind an environment variable. Since the host is not yet a bottleneck in this PR, deferring the change to a later PR.

# Results (before vs after)
  | isl | osl | concurrency | before | after | Δ |
  |---:|---:|---:|---:|---:|---:|
  | 8192 | 1024 | 4 | 2717.4 | 3079.3 | +13.3% |
  | 8192 | 1024 | 8 | 4860.2 | 5401.2 | +11.1% |
  | 8192 | 1024 | 16 | 8348.7 | 9036.7 | +8.2% |
  | 8192 | 1024 | 32 | 12504.3 | 12983.1 | +3.8% |
  | 8192 | 1024 | 64 | 17825.2 | 18322.8 | +2.8% |
  | 8192 | 1024 | 128 | 24904.6 | 25437.0 | +2.1% |
  | 8192 | 1024 | 256 | 32584.8 | 32643.9 | +0.2% |
  | 1024 | 1024 | 4 | 666.3 | 769.9 | +15.5% |
  | 1024 | 1024 | 8 | 1213.1 | 1388.7 | +14.5% |
  | 1024 | 1024 | 16 | 2229.0 | 2424.2 | +8.8% |
  | 1024 | 1024 | 32 | 3659.5 | 3881.5 | +6.1% |
  | 1024 | 1024 | 64 | 5988.5 | 5959.5 | -0.5% |
  | 1024 | 1024 | 128 | 9146.1 | 9283.0 | +1.5% |
  | 1024 | 1024 | 256 | 13523.5 | 13186.0 | -2.5% |

See individual commit messages for the performance delta and before/after metrics of each specific change.

Baseline is based on
- tpu-inference `53d20239446c6dfe04cc0428d4bda8906886af37` 
- vllm `8ad4a01825ef941e785b2bc305ac7a6b5ca9c530`

### DIFF
--- a/tpu_inference/models/vllm/vllm_model_wrapper.py
+++ b/tpu_inference/models/vllm/vllm_model_wrapper.py
@@ -352,6 +352,25 @@ class VllmModelWrapper:
         self.model = _VllmRunner(vllm_model)
         params_and_buffers = shard_model_to_tpu(self.model, self.mesh)
 
+        # Slice rotary cos_sin_cache buffers to max_model_len to save HBM usage
+        # and reduce the per-step overhead of the XLA layout copy. Assumes rope
+        # positions are 1-D and bounded by max_model_len (standard text RoPE);
+        # MRoPE video positions are structural and can exceed max_model_len, so
+        # skip the slice there to avoid an out-of-bounds cos_sin_cache gather.
+        if envs.SLICE_ROPE_CACHE and \
+                not self.vllm_config.model_config.uses_mrope:
+            max_len = self.vllm_config.model_config.max_model_len
+            for key, val in list(params_and_buffers.items()):
+                if key.endswith("rotary_emb.cos_sin_cache"):
+                    arr = jax_view(val)
+                    if arr.shape[0] > max_len:
+                        params_and_buffers[key] = torch_view(arr[:max_len])
+                        logger.info(
+                            "Sliced rope cache %s rows %d -> %d. Assumes "
+                            "positions are 1-D and bounded by max_model_len "
+                            "(%d); MRoPE (video) can exceed it and is excluded",
+                            key, arr.shape[0], max_len, max_len)
+
         self._pooler: Pooler | None = self.model.pooler
 
         if self.vllm_config.model_config.is_multimodal_model:
--- a/tests/e2e/benchmarking/inferencex/qwen3.5/server.sh
+++ b/tests/e2e/benchmarking/inferencex/qwen3.5/server.sh
@@ -57,6 +57,8 @@ export ONEHOT_MOE_PERMUTE_THRESHOLD=32768
 export VLLM_MOE_CHUNK_SIZE=256
 export RAGGED_GATED_DELTA_RULE_IMPL=chunked_kernel_p_recurrent_kernel_d
 export NEW_MODEL_DESIGN=1
+# Slice rope cache to max_model_len (envs.SLICE_ROPE_CACHE is off by default).
+export SLICE_ROPE_CACHE=1
 # Disable DP-scheduler batched prefill, which is for variable isl and osl.
 export DP_SCHED_BATCH_PREFILL=false
 export LIBTPU_INIT_ARGS=' --xla_tpu_use_minor_sharding_for_major_trivial_input=true --xla_tpu_enable_sparse_core_collective_offload_reduce_scatter=false --xla_tpu_ars_combiner_threshold_in_bytes=0 --xla_tpu_enable_async_collective_merger=false'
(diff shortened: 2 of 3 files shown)


tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -352,6 +352,21 @@ def load_weights(self,
         self.model = _VllmRunner(vllm_model)
         params_and_buffers = shard_model_to_tpu(self.model, self.mesh)
 
+        # Slice rotary cos_sin_cache buffers to max_model_len to save HBM usage
+        # and reduce the per-step overhead of the XLA layout copy.
+        if envs.SLICE_ROPE_CACHE:
```
qizzzh: The suffix match `key.endswith("rotary_emb.cos_sin_cache")` catches the language `self.rotary_emb` buffer in every torchax-loaded model, including MRoPE-backed ones. `get_rope(...)` returns `MRotaryEmbedding` when `rope_parameters` contains `mrope_section`, and that class **intentionally over-provisions the cache to `max_position_embeddings * 4`**:

```python
# vllm/model_executor/layers/rotary_embedding/mrope.py:236-238
# In Qwen2.5-VL, the maximum index value is related to the duration of
# the input video. We enlarge max_position_embeddings to 4 times to get
# a larger the cos and sin cache.
self.cache_max_position_num = max_position_embeddings * 4
```

The forward then gathers by raw position value: `cos_sin = cos_sin_cache[positions]` (`mrope.py:284, 336`). Video/image position IDs can exceed `max_model_len`, so slicing at `max_model_len` puts those gathers OOB → NaNs/garbage rather than a clean error.

The perf numbers were collected with `--language-model-only` + `--limit-mm-per-prompt='{"image":0,"video":0}'`, so this doesn't fire in the benchmark. But `SLICE_ROPE_CACHE=True` is a **global default in `envs.py`**, so any production Qwen3.5 / Qwen3-VL / Qwen2.5-VL served via `MODEL_IMPL_TYPE=vllm` with real MM inputs inherits this footgun.

Please pick one:
1. Skip slicing when `self.vllm_config.model_config.uses_mrope` is True (or `is_multimodal_model and not language_model_only`).
2. Resolve the actual rope module for each matched key and skip when `isinstance(rope, MRotaryEmbedding)` or `getattr(rope, "cache_max_position_num", 0) > max_len`.
3. Default `SLICE_ROPE_CACHE=False` and opt-in from `server.sh` (the benchmark already sets a pile of env vars).

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -352,6 +352,21 @@ def load_weights(self,
...
+            max_len = self.vllm_config.model_config.max_model_len
+            for key, val in list(params_and_buffers.items()):
+                if key.endswith("rotary_emb.cos_sin_cache"):
+                    arr = jax_view(val)
+                    if arr.shape[0] > max_len:
+                        params_and_buffers[key] = torch_view(arr[:max_len])
+                        logger.info(
+                            "Sliced rope cache %s rows %d -> %d, because "
+                            "max_model_len is %d so decode never gathers "
```
qizzzh: The claim "decode never gathers positions beyond it (the tail is dead weight)" is only true for 1-D text positions with standard RoPE. For MRoPE with video, positions can (and by design do) exceed `max_model_len`. Please rephrase to make this an assumption you're relying on, not a fact — e.g.:

> "Assumes positions are 1-D and bounded by max_model_len; MRoPE video positions may violate this."

That way anyone tripping over an OOB gather can grep the log and understand why.

tpu_inference/models/vllm/vllm_model_wrapper.py
wyzhang: Done. Disabled by default

tpu_inference/models/vllm/vllm_model_wrapper.py
wyzhang: Good catch. Updated the comment and warning that this is applicable to text only. Would break multi-modal.

qizzzh (approved):

gxd3 (approved):

Lumosis (approved):

wenxindongwork (approved):

vipannalla (approved):
lgtm

PR #3084, opened 2026-07-06
Fix allowed-token mask cleanup on request removal

### LINKED ISSUE #3083: [Bug]: allowed_token_ids request removal can crash on NumPy 2
### Describe the bug
Removing a request that uses `allowed_token_ids` can fail in `InputBatch.remove_request()`.

The cleanup path resets the allowed-token mask with a NumPy-style array, but the current implementation uses a PyTorch-style `fill_` call and initializes the mask with deprecated / invalid NumPy dtype usage. On NumPy 2 this path is brittle and can crash during request removal.

### Expected behavior
Request removal should clear the mask row without error.

### Reproduction
1. Create a request with `SamplingParams(allowed_token_ids=[...])`.
2. Add it to `InputBatch`.
3. Call `remove_request()`.

### Impact
This affects any path that uses `allowed_token_ids` and later removes the request from the batch.

harivilasp:
Closes #3083.

Fix request cleanup for `allowed_token_ids`:
- reset the NumPy mask row with `fill(False)` instead of `fill_`
- use valid `(rows, cols)` shape tuples for mask allocation
- switch deprecated `np.bool` / `jnp.bool` usage to the non-deprecated boolean dtypes

Verification:
- `python3 -m py_compile tpu_inference/runner/input_batch.py tests/runner/test_input_batch.py`
- stubbed runtime check that adds a request with `allowed_token_ids`, then removes it and confirms the mask row is cleared

### DIFF
--- a/tests/runner/test_input_batch.py
+++ b/tests/runner/test_input_batch.py
@@ -623,3 +623,22 @@ def test_dp_mamba_global_to_local_conversion():
     assert (7 * local_slots + 1) % local_slots == 1
     # Boundary: rank 1 last slot = 595 → local 297
     assert 595 % local_slots == 297
+def test_remove_request_with_allowed_token_ids_clears_mask(
+    input_batch: InputBatch,
+):
+    """Removing a request with allowed tokens should not crash."""
+    req = create_dummy_request(
+        "req-1",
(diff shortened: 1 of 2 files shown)


harivilasp:
i think team need to add label of ready to this PR to run these automated checks

Issue #2952, opened 2026-06-21
[Feature]: Support native JAX GPT-OSS MXFP4 quantization loading


Leo7Deng:
### 🚀 The feature, motivation and pitch

Similar to #1623 and #2261, add an mxfp4 method to the jax quantization registry [here](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/jax/quantization/__init__.py#L30-L37) so GPT-OSS can load natively. Currently GPT-OSS MXFP4 loads its decoded weights into Qwix QArray fields (gpt_oss.py _load_mxfp4), which is the path #1680 wants to deprecate.

Scoped to TPUv7+ since fp4 isn't supported on v6e (shown from #1425).

### Additional context

This should follow the existing Fp8FusedMoEMethod. The MXFP4 decode helpers should exist, little kernel work needed.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

Leo7Deng:
@lk-chen can you help assign this to me?

lk-chen:
Hi @Leo7Deng I don't feel #2954 needs to be merged as prerequisite.

The first step is to add MXFP4 [here](https://github.com/vllm-project/tpu-inference/tree/main/tpu_inference/layers/jax/quantization), then re-write gpt-oss similar to what #1623 did with Qwen models.

Leo7Deng:
@lk-chen Sounds good, thanks! I've been converting GptOssMoE over to JaxMoE with an Mxfp4FusedMoEMethod, since gpt-oss has a custom MoE that Qwen didn't.

lk-chen:
What exact difference are you seeing?

For reference, gemma4 has custom router, so we added [Gemma4Moe](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/models/jax/gemma4.py#L164), but recently I noticed #3041 and started #3043

Leo7Deng:
@lk-chen Cool, I've been following Gemma4MoE. The difference is that GptOssMoE is currently its own nnx.Module that reimplements the expert computation, rather than subclassing JaxMoE like Gemma4MoE. So I'm rebuilding it on JaxMoE to bring it onto the native quantization path like the other models.
[...]

PR #3086, opened 2026-07-06
[Jax] Add native GPT-OSS MXFP4 quantization method (1/2)

Leo7Deng:
# Description

Part 1 of 2 for #2952: adds the MXFP4 quant method, config, and registry entry under `layers/jax/quantization`. Part 2 is ready and will follow after this one. It rewrites GPT-OSS onto `JaxRoutedExperts` the same way #3052 did for Gemma4, and moves its expert loading off Qwix. GPT-OSS doesn't actually use this method until then, so this PR has no user-visible effect on its own.

Right now the flax_nnx path can't select MXFP4 at all: `get_tpu_quantization_config` raises `NotImplementedError` for `gpt_oss_mxfp4`.

### Key changes

1. New `Mxfp4FusedMoEMethod` (`layers/jax/quantization/mxfp4.py`). It follows the `Fp8FusedMoEMethod` lifecycle (`create_weights_jax`, `load_weights`, `process_weights_after_loading`, `apply_jax`) and reuses the same dequantize, requantize to TPU fp4, process, and shard sequence that the torchax `VllmMxfp4MoEMethod` already runs, so there's no new quantize or apply math here.
2. New `Mxfp4Config`, registered for `gpt_oss_mxfp4`. Modeled on `CompressedTensorsConfig`: routed-experts layers get the method, everything else gets `None`, so attention and the other linears stay bf16 (same as vLLM's GPT-OSS MXFP4 behavior) and nothing else is affected.
3. Moved `MXFP4_REQUANTIZED_BLOCK_SIZE` to `layers/common/quantization` next to `MXFP4_BLOCK_SIZE`, so the JAX and torchax frontends share one definition instead of both declaring 512.

Some finer details:

- `load_weights` handles all six expert tensors itself instead of delegating like FP8 does. `JaxRoutedExperts` has no generic loader to delegate to (its hook just passes a stub).
- The packed checkpoint MXFP4 gets dequantized and then requantized to TPU `float4_e2m1fn` (block size 512) instead of consuming the checkpoint scales directly. This matches what the torchax path does.
- The checkpoint `*_blocks` tensors are 4-D, so staging flattens them into the 3-D layout the shared dequant helper expects (with a shape check on every staged tensor).
- Staging params are pinned to the CPU mesh so the big packed tensors stay off device until sharding.
- Expert biases ride through `FusedMoEWeights`. FP8 passes `None` there, but GPT-OSS checkpoints ship real expert biases.
- Scale attrs are just `_weight_scale`, since MXFP4 scales aren't inverse scales like FP8's `weight_scale_inv`.

# Tests

CPU unit tests against the repo's pinned vLLM version:

```
pytest --continue-on-collection-errors tests/layers/jax/quantization/ tests/layers/jax/moe/
```

51 passed, 4 skipped. (`test_fp8.py` fails collection on my macOS CPU host because `jaxlib._jax` is missing `TransferConnection`, unrelated to this change.)

- New tests in `test_mxfp4.py`: config dispatch (`TestMxfp4Config`, same shape as `test_compressed_tensors.py`) and the method lifecycle (`TestMxfp4FusedMoEMethod`). That includes an unmocked end-to-end run from `load_weights` through `process_weights_after_loading` on CPU, partial loads across multiple checkpoint files, and a regression test for the checkpoint-shaped 4-D blocks vs the flattened 3-D form.
- `pre-commit run` passes on the changed files.
- I haven't run this on real TPU hardware, since the requantized `float4_e2m1fn` path needs v7+. Nothing calls this method until part 2 so I think CPU coverage is enough here, but part 2 will exercise it with real GPT-OSS checkpoints. How should I handle TPU validation for that one?

# Checklist

- [x] I have made or will make corresponding changes to any relevant documentation. (Internal quantization dispatch only; no user-facing docs affected.)


lk-chen (approved):

lk-chen:
Thank you , LGTM

lk-chen:
@copilot show owner in username

jrplatin (approved):

PR #3091, opened 2026-07-07
Revert vLLM LKG to fa4321de3d (vllm #47070 breaks set_forward_context)

QiliangCui2023:
## Summary
Roll the pinned vLLM LKG back from `ba22152096` to `fa4321de3d` to restore green CI.

## Why
vLLM `ba22152096` includes vllm-project/vllm#47070 (commit `b6cc46ec`), which restructured `set_forward_context` in `vllm/forward_context.py` so the `num_tokens_across_dp is None` path now runs `assert num_tokens is not None`. tpu-inference's `VllmModelWrapper.step_fun_impl` (`tpu_inference/models/vllm/vllm_model_wrapper.py`) calls `set_forward_context(attn_metadata=..., vllm_config=...)` without `num_tokens`, so the assert fires during backbone warm-up compilation:

```
compilation_manager.py model_fn_warmup
  -> vllm_model_wrapper.py step_fun_impl
  -> vllm/forward_context.py set_forward_context
  -> assert num_tokens is not None
```

This turns ~9 tpu7x jobs red (EP recompilation, lm_eval accuracy for qwen3-coder / gpt-oss / Qwen3.5-397B, perf regressions, DeepSeek MMLU).

## Proof (single-commit A/B, tpu-inference held fixed on latest main)
- vLLM `b6cc46ec^` (== `fa4321de`) -> PASS (generates output)
- vLLM `b6cc46ec` -> FAIL (`assert num_tokens is not None`)

## Follow-up
The proper fix is to thread `num_tokens` into the `set_forward_context` calls in `vllm_model_wrapper.py`; this revert only unblocks CI in the meantime.


CienetStingLin (approved):
LGTM

yiw-wang (approved):

PR #3093, opened 2026-07-07
Include vLLM commit in CI image tag + verify image vLLM (fix LKG mis-promotion)

QiliangCui2023:
## Problem
The CI cache image is tagged `vllm-tpu:<tpu_inference_sha>-<tpu_version>` (`setup_docker_env.sh`), which **omits the vLLM commit**. Two pipelines push that same tag:
- `tpu-inference-ci` builds with the **pinned LKG** vLLM (`.buildkite/vllm_lkg.version`)
- `tpu-vllm-integration` builds with **vLLM HEAD**

When both run on the same tpu-inference commit at the same time, they build the **same tag** and overwrite each other's image. A test then pulls whichever image landed last — potentially a different vLLM than intended.

## This actually happened (with timestamps)
On 2026-07-06, both fired on tpu-inference commit `27688e35` at `10:00:04`:

| time (UTC) | pipeline | action | vLLM |
|---|---|---|---|
| 10:08:30 | integration #1794 build | push `vllm-tpu:27688e35-tpu7x` | `ba221520` (HEAD) |
| 10:09:39 | CI #21628 build | **push (overwrite)** same tag | `fa4321de` (LKG) |
| ≥10:10:12 | integration #1794 test | **pull** same tag → got | `fa4321de` |

So integration validated `fa4321de` but its promote step wrote **`ba221520`** to LKG — shipping a vLLM its tests never ran. That vLLM (`ba221520`, via vllm-project/vllm#47070) then turned `main` red with `assert num_tokens is not None` during backbone compile. Confirmed by pulling the current tag: `vllm-tpu:27688e35…-tpu7x` today contains yet a *third* vLLM (`40cc2e83`), proving the tag is mutable and does not pin the vLLM.

## Fix
1. **Include `VLLM_COMMIT_HASH` in the cache tag** → `vllm-tpu:<tpu_inference_sha>-<vllm_sha>-<tpu_version>`. CI-LKG and integration-HEAD images no longer share a tag, so they can't clobber each other. (Falls back to the old format when the hash is empty, e.g. local/dev builds.)
2. **`verify_image_vllm()`** — after pulling and after building, assert the image's actual `/workspace/vllm` HEAD equals the expected `VLLM_COMMIT_HASH`; abort on mismatch. No-op for empty hash (local/dev) or PyPI builds.

Precedent: `nightly_releases.yml` already includes `$VLLM_SHORT_COMMIT` in its image tag.

## Notes / follow-ups (out of scope here)
- The affected integration tests are `soft_fail: true`, so the promote gate (`wait: ~`) treats a red test as a pass. Even with correct images, a genuinely-failing integration test can still promote. Recommend a separate change to make the promote gate respect real test results (or drop `soft_fail` on the gating jobs).
- A promote-time image check could be added to `integration_promote.yml` for defense in depth; with fix #1 the promoted vLLM now equals the tested vLLM by construction.

## Testing
- `shellcheck` + `bash -n` clean; all pre-commit hooks pass.
- The tag change is self-contained: every build/pull path goes through `setup_environment`, which computes `CACHE_TAG` once, so build and pull stay consistent.

### DIFF
--- a/.buildkite/scripts/setup_docker_env.sh
+++ b/.buildkite/scripts/setup_docker_env.sh
@@ -78,6 +78,34 @@ cleanup_docker_resource() {
   echo "Cleanup complete."
 }
 
+# Verify a built or pulled image actually contains the expected vLLM commit.
+# Guards against tag/image mismatches -- e.g. when the CI (LKG vLLM) and
+# integration (HEAD vLLM) pipelines build the same tpu-inference commit
+# concurrently, they previously shared a vLLM-agnostic tag and could clobber
+# each other's image in the registry. No-op when the expected hash is empty
+# (local/dev builds that clone vLLM HEAD) or for PyPI builds (no
+# /workspace/vllm git checkout to inspect).
+verify_image_vllm() {
+  local image_ref="$1"
+  local expected_vllm="${2:-}"
+  if [[ -z "${expected_vllm}" ]]; then
+    return 0
+  fi
+  if [[ "${RUN_WITH_PYPI:-false}" == "true" ]]; then
+    echo "[verify-vllm] RUN_WITH_PYPI=true; skipping vLLM commit verification."
+    return 0
+  fi
+  local actual_vllm
+  actual_vllm=$(docker run --rm --entrypoint git "${image_ref}" \
+    -C /workspace/vllm rev-parse HEAD 2>/dev/null || echo "unknown")
+  if [[ "${actual_vllm}" != "${expected_vllm}" ]]; then
+    echo "[FATAL][verify-vllm] Image ${image_ref} contains vLLM ${actual_vllm}, but expected ${expected_vllm}."
+    echo "[FATAL][verify-vllm] Aborting to avoid testing or promoting the wrong vLLM."
+    exit 1
+  fi
+  echo "[verify-vllm] OK: ${image_ref} contains expected vLLM ${expected_vllm}."
+}
+
 setup_environment() {
   local image_name_param=${1:-"vllm-tpu"}
   local should_push=${2:-"false"}
@@ -135,7 +163,15 @@ setup_environment() {
       TPU_INFERENCE_HASH="$BUILDKITE_COMMIT"
   fi
  
+  # Include the vLLM commit in the cache tag so an image is uniquely identified
+  # by BOTH its tpu-inference commit and its vLLM commit. Without this, the CI
+  # pipeline (LKG vLLM) and the integration pipeline (HEAD vLLM) produce the same
+  # tag for the same tpu-inference commit and can overwrite each other's image in
+  # the registry -- a test could then silently run a different vLLM than intended.
   local CACHE_TAG="${TPU_INFERENCE_HASH}-${LOCAL_TPU_VERSION}"
+  if [[ -n "${VLLM_COMMIT_HASH}" ]]; then
+    CACHE_TAG="${TPU_INFERENCE_HASH}-${VLLM_COMMIT_HASH}-${LOCAL_TPU_VERSION}"
+  fi
 
   # ==========================================
   # Pull-Only Mode for TPU execution nodes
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/setup_docker_env.sh
```diff
@@ -78,6 +78,34 @@ cleanup_docker_resource() {
...
+  if [[ -z "${expected_vllm}" ]]; then
+    return 0
+  fi
+  if [[ "${RUN_WITH_PYPI:-false}" == "true" ]]; then
+    echo "[verify-vllm] RUN_WITH_PYPI=true; skipping vLLM commit verification."
+    return 0
+  fi
+  local actual_vllm
+  actual_vllm=$(docker run --rm --entrypoint git "${image_ref}" \
+    -C /workspace/vllm rev-parse HEAD 2>/dev/null || echo "unknown")
```
QiliangCui2023: **Distinguish "couldn't read the image's vLLM" from a real mismatch — don't let an infra failure masquerade as a vLLM mismatch**

`2>/dev/null || echo "unknown"` collapses every failure mode of this `git rev-parse` — git not on `PATH`, `/workspace/vllm` missing, or git's `safe.directory` "dubious ownership" guard tripping when the container runs git as a non-owner user — into `actual_vllm="unknown"`. That then hits the `[FATAL] … contains vLLM unknown, but expected …` branch and aborts, pointing the on-call at a *vLLM mismatch* when the real cause is an infra/tooling problem.

Suggested: capture stderr and branch on the exit code so a broken check reads as a broken check:

```bash
local actual_vllm rc
actual_vllm=$(docker run --rm --entrypoint git "${image_ref}" \
  -C /workspace/vllm rev-parse HEAD 2>/tmp/verify_vllm.err); rc=$?
if [[ $rc -ne 0 ]]; then
  echo "[FATAL][verify-vllm] Could not read vLLM commit from ${image_ref}: $(cat /tmp/verify_vllm.err)"
  exit 1
fi
```

Also worth confirming the CI images run git as the same user that owns `/workspace/vllm`, otherwise `safe.directory` will trip this on every build.

.buildkite/scripts/setup_docker_env.sh
```diff
@@ -135,14 +163,23 @@ setup_environment() {
       TPU_INFERENCE_HASH="$BUILDKITE_COMMIT"
   fi
  
+  # Include the vLLM commit in the cache tag so an image is uniquely identified
+  # by BOTH its tpu-inference commit and its vLLM commit. Without this, the CI
+  # pipeline (LKG vLLM) and the integration pipeline (HEAD vLLM) produce the same
+  # tag for the same tpu-inference commit and can overwrite each other's image in
+  # the registry -- a test could then silently run a different vLLM than intended.
   local CACHE_TAG="${TPU_INFERENCE_HASH}-${LOCAL_TPU_VERSION}"
+  if [[ -n "${VLLM_COMMIT_HASH}" ]]; then
```
QiliangCui2023, line 185: **Log (or hard-fail) when a real pipeline build falls back to the empty-hash tag — otherwise the fix silently disables itself**

When `VLLM_COMMIT_HASH` is empty, `CACHE_TAG` reverts to the old `${TPU_INFERENCE_HASH}-${LOCAL_TPU_VERSION}` format *and* `verify_image_vllm` becomes a no-op — i.e. the exact CI-LKG vs integration-HEAD clobber this PR fixes silently reappears, with no signal. That's the right default for local/dev and PyPI builds, but inside a real pipeline an empty hash means something is wrong (e.g. `buildkite-agent meta-data get "VLLM_COMMIT_HASH"` returned empty, or bootstrap didn't set it).

Suggested: when `BUILDKITE=true` and not `RUN_WITH_PYPI`, emit a loud warning (or fail) if the hash is empty, so a mis-set/missing hash is visible rather than quietly degrading back to the shared-tag behavior this PR is meant to eliminate.

.buildkite/scripts/setup_docker_env.sh
QiliangCui2023: Good catch — fixed in 24268943a. `verify_image_vllm` now captures git's exit code separately (`... 2>"${err}") || rc=$?`, which is `set -e`-safe) and, on a non-zero exit, reports `[FATAL] Could not read the vLLM commit ... (git exit N)` with git's stderr, instead of collapsing every failure into `unknown` and hitting the mismatch branch. Also added `-c safe.directory=/workspace/vllm` to pre-empt the dubious-ownership trip you flagged (the image builds vLLM as root, which owns /workspace/vllm, so it also works in practice — this is belt-and-suspenders).

.buildkite/scripts/setup_docker_env.sh
QiliangCui2023: Done in 24268943a. When `VLLM_COMMIT_HASH` is empty and `BUILDKITE` is set and it's not a PyPI build, it now hard-fails with a `[FATAL]` explaining that bootstrap likely didn't set the metadata — rather than quietly degrading to the vLLM-agnostic tag. Local/dev and PyPI builds still fall back to the old tag silently (the legitimate empty-hash cases). Went with hard-fail over warn since a silent revert is exactly the shared-tag behavior this PR removes.

CienetStingLin (approved):
LGTM

PR #3096, opened 2026-07-07
Support PCP E2E for single request on dense and MoE GQA models

wenxindongwork:
# Description

This pull request introduces end to end support for PCP (prefill context parallel). This PR supports both dense and MoE models, but does not support GDN and MLA models. PCP currently only processes one request at a time.

**Model Quality**

lm_eval 
```
SKIP_JAX_PRECOMPILE=1 NEW_MODEL_DESIGN=1 vllm serve  Qwen/Qwen3-8B-Base  --max-model-len=4096   --max-num-batched-tokens=4096 --max_num_seqs=1  --gpu-memory-utilization=0.
9   --tensor-parallel-size=4   --prefill-context-parallel-size 2

lm_eval --model local-chat-completions     --model_args base_url="http://localhost:8000/v1/chat/completions",model="Qwen/Qwen3-8B-Base",eos_string="<|im_end|>",num_concurrent=256     --tasks gsm8k     --apply_chat_template    --fewshot_as_multiturn --limit 200

```

```
|Tasks|Version|     Filter     |n-shot|  Metric   |Value|   |Stderr|
|gsm8k|      3|flexible-extract|     5|exact_match|0.815|±  |0.0275|
|     |       |strict-match    |     5|exact_match|0.815|±  |0.0275|
```

**PCP Performance**

Kernel performance: 

v7x-8, NQ=16, NKV=4 (GQA), HD=256, bf16

```
┌─────────┬─────────┬─────────┬──────────┬──────────┬──────────┬──────────┬─────────┐
│ context │     tp4 │     tp8 │ pcp2_tp4 │ pcp8_tp1 │ tp4/pcp2 │ tp4/pcp8 │ tp4/tp8 │
├─────────┼─────────┼─────────┼──────────┼──────────┼──────────┼──────────┼─────────┤
│     16k │    2.88 │    2.14 │     3.03 │     4.73 │    0.95× │    0.61× │   1.35× │
│     64k │   25.61 │   15.52 │    18.59 │    24.55 │    1.38× │    1.04× │   1.65× │
│    128k │   90.60 │   50.23 │    56.19 │    67.03 │    1.61× │    1.35× │   1.80× │
│    512k │ 1266.96 │  659.94 │   669.36 │   666.64 │    1.89× │    1.90× │   1.92× │
│      1M │ 4924.83 │ 2519.91 │  2473.31 │  2346.60 │    1.99× │    2.10× │   1.95× │
│      2M │   19386 │ 9822.31 │  9442.76 │  8714.75 │    2.05× │    2.22× │   1.97× │
└─────────┴─────────┴─────────┴──────────┴──────────┴──────────┴──────────┴─────────┘
```

E2E performance: 

- Model: Qwen/Qwen3-235B-A22B-Instruct-2507-FP8
- Input length: 256k 
- Output length: 1
- Chunk prefill chunk size: 16k 

```python tpu-inference/scripts/vllm/benchmarking/benchmark_serving.py --backend=vllm --model=Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 --dataset-name=random --num-prompts=10 --random-input-len=262134 --random-output-len=1 ```

TP=4 
---

```
VLLM_TPU_BUCKET_PADDING_GAP=2048 PHASED_PROFILING_DIR=gs://wenxindong-vm/trace/pcp/e2e/tp NEW_MODEL_DESIGN=1 vllm serve Qwen/Qwen3-235B-A22B-Instruct-2507-FP8  --max-model-len=262144 --max-num-seqs=1 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=4  --download_dir=/mnt/disks/persist --max-num-batched-tokens=16384
```

```
Successful requests:                     10        
Benchmark duration (s):                  1372.31   
Total input tokens:                      2621340   
Total generated tokens:                  10        
Request throughput (req/s):              0.01      
Output token throughput (tok/s):         0.01      
Total token throughput (tok/s):          1910.18   
Mean TTFT (ms):                          755085.15 
Median TTFT (ms):                        755093.03 
P99 TTFT (ms):                           1359917.71
Mean TPOT (ms):                          0.00      
Median TPOT (ms):                        0.00      
P99 TPOT (ms):                           0.00      
Mean ITL (ms):                           0.00      
Median ITL (ms):                         0.00      
P99 ITL (ms):                            0.00      
Mean E2EL (ms):                          755085.15 
Median E2EL (ms):                        755093.03 
P99 E2EL (ms):                           1359917.71
```
TTFT: 137s

PCP=2 TP=4 
----------

```VLLM_TPU_BUCKET_PADDING_GAP=2048 PHASED_PROFILING_DIR=gs://wenxindong-vm/trace/pcp/e2e/tp NEW_MODEL_DESIGN=1 vllm serve Qwen/Qwen3-235B-A22B-Instruct-2507-FP8  --max-model-len=262144 --max-num-seqs=1 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=4  --download_dir=/mnt/disks/persist --max-num-batched-tokens=16384 --prefill-context-parallel-size 2```

```
Successful requests:                     2         
Benchmark duration (s):                  131.18    
Total input tokens:                      524268    
Total generated tokens:                  2         
Request throughput (req/s):              0.02      
Output token throughput (tok/s):         0.02      
Total token throughput (tok/s):          3996.45   
Mean TTFT (ms):                          98553.62  
Median TTFT (ms):                        98553.62  
P99 TTFT (ms):                           130525.77 
Mean TPOT (ms):                          0.00      
Median TPOT (ms):                        0.00      
P99 TPOT (ms):                           0.00      
Mean ITL (ms):                           0.00      
Median ITL (ms):                         0.00      
P99 ITL (ms):                            0.00      
Mean E2EL (ms):                          98553.62  
Median E2EL (ms):                        98553.62  
P99 E2EL (ms):                           130525.77 
```

TTFT: 65s

# Tests

commands to reproduce.

# Checklist

### CODE REVIEW

File: tpu_inference/layers/common/sharding.py [hunk 1]
```diff
@@ -78,15 +81,17 @@ class ShardingAxisName2D:
...
     MLP_TENSOR = 'model'
     MOE_TENSOR = 'model'
     EXPERT = 'model'
     EXPERT_DATA = ('data', 'model')
     VOCAB = ('data', 'model')
     BATCH = 'data'
-    CONTEXT = None
-    KV_CACHE_HEAD = 'model'
+    KV_CONTEXT = None
+    PREFILL_CONTEXT = None
```
weiyu0824, line 93: Could you elaborate on why we need this PREFILL_CONTEXT sharding name?

wenxindongwork: we need PREFILL_CONTEXT to shard the input ids over the pcp axis. I think we could reuse the DATA axis as well, and it might be a better choice. let me update the PR.

tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
```diff
@@ -732,6 +734,14 @@ def _fetch_bkv(seq_idx, bkv_idx, bkv_sem_idx, *, wait=False):
             # Fetch new kvs.
             if not skip_current_attn:
                 new_kv_len_start = _seq_kv_new_end - kv_left_frm_new
+                if pcp_chunk_size is not None:
```
weiyu0824, line 737: nit: Can we avoid using C = pcp_chunk_size? I think pcp_chunk_size is more readable and c could be something like chunk_idx. wdyt?

tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
```diff
@@ -2340,7 +2365,7 @@ def run(scalar_prefetches, hbms):
 
     def _prepare_block_sizes(block_sizes, case):
         if block_sizes is None:
-            return get_default_block_sizes(
```
weiyu0824, line 2343: Do we have to make change on this part? Seems like there is no difference.

tpu_inference/layers/common/sharding.py
```diff
@@ -60,8 +61,10 @@ class ShardingAxisNameBase:
 
     # These axes are used in KV caches management.
     BATCH = ('data', 'attn_dp', 'attn_dp_expert')
-    CONTEXT = 'dcp'
-    KV_CACHE_HEAD = ('model', 'expert')
+    KV_CONTEXT = ('pcp', 'dcp')
```
weiyu0824, line 64: I think we have to add note to remind ourself that KV_CONTEXT should be ('dcp') only in the future.

tpu_inference/layers/common/attention_interface.py
```diff
@@ -58,6 +59,7 @@
     logger.info_once("Using default RPA kernel")
 
 ragged_paged_attention = rpa.ragged_paged_attention
+pcp_ragged_paged_attention_kernel = rpa_v3_cp.ragged_paged_attention
```
weiyu0824: nit: Can we just use rpa_v3_cp.ragged_paged_attention instead of add another pcp_ragged_paged_attention_kernel var?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -770,6 +773,161 @@ def forward_with_dcp(
...
+            skip_current_attn=True,
+            use_causal_mask=False,
+            update_kv_cache=False,
+            **common)
+        o1, l1 = _lse_reduce_scatter(o1, l1, pcp_axis, pcp_size)
+
+        # ---- current phase -----
+        # `cu_q_lens[0]` = [0, C, C+tail_real] and `pcp_qpos` = [head_offset, tail_offset]
+        page_size_local = kvc.shape[1]
+        remap_kv = (C >= page_size_local) and (C % page_size_local == 0)
```
weiyu0824: why if we remap kv depend on chunk_size >= page_size_local and chunk_size % page_size_local == 0?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -770,6 +773,161 @@ def forward_with_dcp(
...
+                      v_scale=v_scale)
+
+        # ---- cache phase ----
+        ag_q = ag(q_l)  # [pcp*2C]
+        cu_cache = jnp.zeros_like(cu_q_lens[0]).at[1:].set(padded_q_len)
+        dist_cache = jnp.array([0, 0, 1], jnp.int32)
+        o1, kvc1, l1 = pcp_ragged_paged_attention_kernel(
+            ag_q,
+            k_l,
+            v_l,
```
weiyu0824: nit: I think a better way to pass k, v is to just pass None. however our current cp kernel doesn't support this yet, should we track this in a another bug?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -660,14 +660,17 @@ def merge_attn_states(context_out: jax.Array, context_lse: jax.Array,
         query_lse = [seq, local_heads]
     """
     max_lse = jnp.maximum(context_lse, query_lse)
```
weiyu0824, line 661: Do we update this max_lse_safe because pcp will trigger a case where cache_phase and current_phase both contain -inf lse?

tpu_inference/layers/common/attention_interface.py
weiyu0824: Can you explain a bit on why we need remap_kv flag? it seems like we don't need chunk reordering when remap_kv=True

weiyu0824 (commented):
Hi wenxin, thanks for the work, I have several questions that need some explanation.

tpu_inference/layers/common/attention_interface.py
wenxindongwork: yes that is a great idea.

tpu_inference/layers/common/attention_interface.py
wenxindongwork: yes, we need to all-gather kv regardless, but we don't need to reorder kv since we can just provide some offsets to the kernel and index the kv blocks directly at the right offset. however, this only applies when the page size divides the local chunk size.

wenxindongwork: answered below .

tpu_inference/layers/common/sharding.py
wenxindongwork: why should KV_CONTEXT only be dcp in the future?

tpu_inference/layers/common/sharding.py
wenxindongwork: in fact it is not possible to just use the DATA axis, we need prefill_context for lse merge witin the pcp group.

tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
wenxindongwork: we have to make sure that bkv_sz ≤ pcp_chunk_size and pcp_chunk_size % bkv_sz == 0 (kv blocks never straddle a chunk.)

tpu_inference/layers/common/attention_interface.py
wenxindongwork: yes, exactly

tpu_inference/layers/common/sharding.py
weiyu0824: Got it. Thanks.

weiyu0824 (approved):

bhuvanpkaruturi (approved):

tpu_inference/layers/common/sharding.py
weiyu0824: I believe upstream decided to use dcp as the KV context sharding axis and pcp as the input sharding axis. Therefore, while specifying pcp=2 currently shards both the input and the KV context, in the future pcp=2 will only shard the input (KV context ends up having duplication). If we want to maintain the current behavior then, we will need to specify both pcp=2 and dcp=2.

wenxindongwork: got it, thanks! let's follow up with dcp + pcp approach. currently we only support dcp-only and pcp-only.

jrplatin (approved):

lk-chen (approved):

PR #3097, opened 2026-07-07
Pass num_tokens to set_forward_context (fix vLLM #47070); bump LKG to b6cc46ec

QiliangCui2023:
## Problem
vLLM `b6cc46ec` (vllm-project/vllm#47070, *"Support sequence parallel without the need for DP"*) changed `set_forward_context` so its DP-metadata block now runs for MoE models whenever `parallel_config.use_sequence_parallel_moe` is true — i.e. `all2all_backend ∈ {allgather_reducescatter, deepep_*, …}` **AND `enable_expert_parallel` AND `tensor_parallel_size > 1`** — not only when `data_parallel_size > 1`. It also added:

```python
elif num_tokens_across_dp is None:
    assert num_tokens is not None
    num_tokens_across_dp = torch.tensor([num_tokens], dtype=torch.int32)
```

`VllmModelWrapper.step_fun_impl` / `draft_step_fun_impl` call `set_forward_context(attn_metadata=..., vllm_config=...)` **without `num_tokens`**, so for every MoE + expert-parallel model (TP>1) that `assert num_tokens is not None` fires during backbone warm-up compilation → `EngineCore failed to start`. This turned ~9 tpu7x jobs red (EP recompilation, lm_eval qwen3-coder / gpt-oss / Qwen3.5-397B, perf regressions, DeepSeek MMLU).

## Fix
Pass `num_tokens` — the padded token count, `input_positions.shape[-1]` (handles both `(N,)` and mRoPE `(3, N)`) — to both `set_forward_context` calls, matching vLLM's own runner which passes `num_tokens=num_tokens_padded`.

Also **bump the vLLM LKG to `b6cc46ec`** (the commit that introduced the breakage, = `fa4321de` + #47070) so CI validates this fix against it. LKG had been reverted to `fa4321de` in #3091 to unblock main; this moves it forward past #47070.

## Root-cause / bisect
Proven by single-commit A/B (tpu-inference held fixed): vLLM `b6cc46ec^` (== `fa4321de`) passes; `b6cc46ec` fails with the `num_tokens` assert. The entry condition flip is exactly #47070's change to the `set_forward_context` DP-metadata guard.

## Testing
- **Local:** the EP-recompilation offline_inference run (`Qwen1.5-MoE-A2.7B`, `--tensor-parallel-size 4 --enable-expert-parallel`, `SKIP_JAX_PRECOMPILE=0 VLLM_XLA_CHECK_RECOMPILATION=1`) against vLLM `b6cc46ec` now completes with **coherent output** instead of aborting on the assert.
- **CI:** the LKG bump runs the full tpu7x MoE/EP model + accuracy suite (the jobs that were red) against `b6cc46ec`, exercising this path end-to-end.


gxd3 (approved):

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -442,13 +442,23 @@ def step_fun_impl(
...
             lora_metadata = torch_view(lora_metadata)
+            # Pass num_tokens (padded) to set_forward_context. vLLM's
+            # set_forward_context requires it whenever the DP-metadata path runs
+            # -- which for MoE models is now gated on
+            # parallel_config.use_sequence_parallel_moe (enable_expert_parallel +
+            # tensor_parallel_size > 1), not just data_parallel_size > 1. Without
+            # it, that path asserts `num_tokens is not None`. input_positions is
+            # (num_tokens,) or (3, num_tokens) for mRoPE, so shape[-1] is the
+            # padded token count.
+            num_tokens = input_positions.shape[-1]
```
QiliangCui2023, line 453: **Same latent assert lives in the native runner path — confirm it's unreachable or fix it too**

This fixes the two `set_forward_context` calls in the vLLM-wrapper path, but the JAX-native runner has two more with no `num_tokens`:

- `tpu_inference/runner/tpu_runner.py:1306` — `set_forward_context(None, self.vllm_config)`
- `tpu_inference/runner/tpu_runner.py:1464` — `set_forward_context(None, self.vllm_config)`

Both hit the identical `assert num_tokens is not None` from #47070 if a native MoE model runs with `use_sequence_parallel_moe` (i.e. `enable_expert_parallel` + `tensor_parallel_size > 1` + a sequence-parallel `VLLM_ALL2ALL_BACKEND`). It's almost certainly latent today — the default `naive` all2all backend makes `use_sequence_parallel_moe` false, which is why only the vLLM-impl jobs went red — but it's the same footgun. `input_positions` is already in scope at 1306. Worth either confirming these can't be reached with `use_sequence_parallel_moe=true`, or applying the same `num_tokens=input_positions.shape[-1]` fix defensively so a future `deepep`/`allgather_reducescatter` native run doesn't reintroduce this.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -503,12 +513,16 @@ def draft_step_fun_impl(
         ) -> Tuple[List[jax.Array], jax.Array, List[jax.Array],
                    Optional[jax.Array]]:
             layer_name_to_kvcache_index = dict(layer_name_to_kvcache_index)
+            # See step_fun_impl above: vLLM's set_forward_context requires
+            # num_tokens for the MoE sequence-parallel DP-metadata path.
+            num_tokens = attn_metadata.input_positions.shape[-1]
```
QiliangCui2023, line 518: **No unit test exercises the new `num_tokens` path**

No test file covers `vllm_model_wrapper.py`, so nothing pins the `num_tokens` computation or that it reaches `set_forward_context`. A regression here would only surface as a full-CI red again, not a fast unit failure.

I know this is genuinely hard to unit-test — the path is a nested closure inside `jit_step_func` wrapped in torchax/jax context — and the LKG bump does give real end-to-end coverage of exactly this path, so I wouldn't block on it. But a focused test that mocks `set_forward_context` and asserts it receives `num_tokens == input_positions.shape[-1]` for both a `(N,)` and a `(3, N)` (mRoPE) positions tensor would be cheap and would guard the `shape[-1]` choice directly.

tpu_inference/models/vllm/vllm_model_wrapper.py
QiliangCui2023: Confirmed unreachable, and added clarifying comments at both sites (951d62db). Those calls pass `attn_metadata=None` and no `num_tokens`, so vLLM's DP-metadata guard `(attn_metadata is not None or num_tokens is not None)` (forward_context.py:288) is False and the whole block — including `assert num_tokens is not None` — is skipped, regardless of `use_sequence_parallel_moe`. That's the key difference from the VllmModelWrapper path, which passes a non-None `attn_metadata`. Note I did **not** add `num_tokens` here: doing so would flip that guard to True and force-enter the DP path the native runner deliberately avoids. The added comments warn against exactly that future change.

tpu_inference/models/vllm/vllm_model_wrapper.py
QiliangCui2023: Agreed it's worth guarding, and thanks for not blocking. The path is a nested closure inside `jit_step_func` under torchax/jax context, so a meaningful test has to reconstruct the model + mesh + torchax env — which is exactly what the EP e2e jobs do. With the LKG bump to b6cc46ec, this PR's CI runs that whole MoE+EP suite against the breaking commit, giving direct end-to-end coverage of the num_tokens path (the non-mRoPE `(N,)` shape via the qwen3-coder / gpt-oss / Qwen3.5-397B jobs; mRoPE `(3,N)` is exercised by the multimodal accuracy jobs). I locally reproduced the fix on the EP-recompilation config too. Happy to add the mock-based closure test if you'd still like it despite the scaffolding cost — just say the word.

QiliangCui (approved):

jrplatin (approved):

PR #3100, opened 2026-07-07
feat(sample): replay-exact sharded candidate sampler (indexed threefry)

lokic233:
## Summary
Adds a replay-exact candidate sampler that avoids the full-vocabulary logits all-gather on tensor-parallel decode. Instead of gathering [B, V], takes exact global top-k then regenerates Gumbel noise for only those k candidate positions via indexed partitionable threefry.

## Key Properties
- **Replay-exact (Level A):** same token as `jax.random.categorical`, bit-for-bit
- **Communication:** V→k reduction (~3000× for V=151936, k=50)
- **Validated:** 260/260 rows bit-exact on CPU (jax 0.10.2) AND real TPU v7x (jax 0.10.1)
- **Version-robust:** import handles jax 0.10.1 (`_src.prng`) vs 0.10.2 (`_src.random.threefry2x32`)

## Scope
Gated behind `TPU_SAMPLING_REPLAY_EXACT_CANDIDATE` (default off). Ships module + device test. `sample()` wiring deferred to follow-up pending end-to-end jitted-path measurement.

### DIFF
--- a/tpu_inference/layers/jax/sample/candidate_sampler.py
+++ b/tpu_inference/layers/jax/sample/candidate_sampler.py
@@ -1,0 +1,193 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Replay-exact distributed candidate sampler (v3): avoids full-vocab all-gather
+while preserving EXACT bit-for-bit token parity with the original full-gather sampler.
+
+Key insight: jax.random.categorical implements the Gumbel-max trick:
+  selected_token = argmax(logits + Gumbel_noise)
+where Gumbel_noise[b, v] is derived from a counter-based PRNG:
+  Gumbel_noise[b, v] = -log(-log(uniform(key, flat_index=b*V + v)))
+
+For the top-k candidates (the only positions that CAN win the argmax, since non-top-k
+have logits=-1e12 → score≈-1e12 regardless of noise), we can reproduce their EXACT
+Gumbel values using just their global vocab indices. This gives REPLAY-EXACT parity:
+the same token is selected as the full-vocabulary categorical, using the same key.
+
+Communication: gather ~k*TP candidate logits + global IDs (200 for k=50, TP=4) instead of 151,936.
+Correctness: Level A replay-exact (same Gumbel values → same argmax → same token, same logprob).
+"""
+from typing import Optional, Tuple
+
+import jax
+import jax.numpy as jnp
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
+
+from tpu_inference.layers.common.binary_search import topp_mask
+
+# Low-level threefry primitive + counter layout, used for true position-addressable
+# (indexed) RNG. This is the SAME primitive jax.random.bits uses under the default
+# jax_threefry_partitionable=True path, so values match bit-for-bit.
+#
+# The primitive's module location moved between jaxlib versions (verified on-device):
+#   jax 0.10.1 (TPU prod jaxlib): jax._src.prng.threefry2x32_p
+#   jax 0.10.2 (CPU):             jax._src.random.threefry2x32.threefry2x32_p
+# Import version-robustly so the sampler is portable across the cluster's jaxlib.
+try:
+    from jax._src.prng import threefry2x32_p as _threefry2x32_p
+except ImportError:  # pragma: no cover - older/newer layout
+    from jax._src.random.threefry2x32 import threefry2x32_p as _threefry2x32_p
+
+_NEG_INF = -1e12
+
+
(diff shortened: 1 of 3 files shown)


vipannalla:
@jrplatin, can you please take a look?

tpu_inference/layers/jax/sample/candidate_sampler.py
```diff
@@ -0,0 +1,178 @@
...
+      and match the full generation exactly. See _indexed_uniform.
+    """
+    B, k = indices.shape
+    # Flat counter for position (b, v) in a row-major [B, V] tensor = b * V + v.
+    batch_offsets = jnp.arange(
+        B, dtype=jnp.uint32)[:, None] * jnp.uint32(vocab_size)
+    flat_indices = (batch_offsets + indices.astype(jnp.uint32)).reshape(
+        -1)  # [B*k]
+    u = _indexed_uniform(key, flat_indices).reshape(B, k)
+    gumbel_candidates = -jnp.log(-jnp.log(u))  # [B, k]
```
pv97, line 127: Would this be inf if u=1?

lokic233: Good catch to check — but `u` can never be exactly 1.0, so it stays finite. `_indexed_uniform` reproduces `jax.random.uniform`'s bound exactly: the mantissa trick yields u01 in [0,1) that tops out at 1 − 2⁻²⁴ = 0.99999994, then it's floored at `tiny`. So u ∈ [tiny, 1) — never 0.0 or 1.0 — and `-log(-log(u))` is finite (worst case |value| ≈ 16.6 in float32). This is identical to `jax.random.gumbel`, which does the same `-log(-log(uniform()))` and relies on the same open-interval bound (that's exactly why this path stays replay-exact against jax). Added a comment documenting the guarantee.

tpu_inference/layers/jax/sample/candidate_sampler.py
```diff
@@ -0,0 +1,182 @@
...
+# Low-level threefry primitive + counter layout, used for true position-addressable
+# (indexed) RNG. This is the SAME primitive jax.random.bits uses under the default
+# jax_threefry_partitionable=True path, so values match bit-for-bit.
+#
+# The primitive's module location moved between jaxlib versions (verified on-device):
+#   jax 0.10.1 (TPU prod jaxlib): jax._src.prng.threefry2x32_p
+#   jax 0.10.2 (CPU):             jax._src.random.threefry2x32.threefry2x32_p
+# Import version-robustly so the sampler is portable across the cluster's jaxlib.
+try:
+    from jax._src.prng import threefry2x32_p as _threefry2x32_p
```
jrplatin, line 48: Is this not exposed in a more public-facing API?

tpu_inference/layers/jax/sample/candidate_sampler.py
```diff
@@ -0,0 +1,182 @@
...
+
+    Replay-exact mechanism (proven on CPU, jax 0.10.2, partitionable threefry):
+      Position (b, v) in a row-major [B, V] tensor has flat counter b*V + v.
+      threefry is a keyed permutation of counters, so the value at counter c
+      depends ONLY on c — we can regenerate any subset of positions independently
+      and match the full generation exactly. See _indexed_uniform.
+    """
+    B, k = indices.shape
+    # Flat counter for position (b, v) in a row-major [B, V] tensor = b * V + v.
+    batch_offsets = jnp.arange(
```
jrplatin, line 118: Do we need to worry about overflow here?

jrplatin:
Is this wired up right now or is this code dead?

tpu_inference/layers/jax/sample/candidate_sampler.py
lokic233: Good catch. The counter is b*V + v, max ≈ B*V. For realistic decode (B ≤ a few thousand, V ≤ ~256k) that's ~10-130M, well under uint32 max (4.29B). But you're right that the uint32 multiply would *silently wrap* past that. I added a trace-time guard that raises if B*V ≥ 2³² (so it fails loudly rather than corrupting indices), and documented why we stay uint32: this path runs without jax_enable_x64, so a uint64 request just truncates back to uint32 anyway (including _indexed_uniform's hi/lo counter split) — genuine >2³² support needs x64 end-to-end. Verified the uint32 path is still bit-exact vs jax.random.uniform (max|diff|=0.0).

tpu_inference/layers/jax/sample/candidate_sampler.py
lokic233: Not that I could find — there's no public jax API to evaluate threefry at *arbitrary counter positions* without materializing the full [B,V] array. jax.random.uniform/categorical always generate the whole shape. Binding threefry2x32_p directly is what lets us regenerate just the top-k positions and still match the full generation bit-for-bit (the whole point of 'replay-exact'). It's a private primitive, so I import it version-robustly (both the 0.10.1 and 0.10.2 module layouts) and have a bit-exactness test guarding against a jax-side change. If a public counter-indexed RNG lands upstream we can switch to it, but today this is the only way to get the exact-replay property. Open to alternatives if you know a public entry point.

lokic233:
@jrplatin on 'is this wired up' — it's opt-in and not on the default path yet. `candidate_sample` returns `None` when inapplicable and the caller falls back to the existing sampler, so it's safe to land without changing default behavior. The intent is to enable it for the replay-exact/speculative path where regenerating gumbel noise at gathered positions (instead of over the full vocab) is the win. If you'd prefer, I can either (a) add the explicit call-site that selects it, or (b) keep this PR as the standalone, tested primitive and wire it up in a follow-up. Happy to go either way — what's your preference?

PR #2338, opened 2026-04-20
Add env variable for overriding rpa block sizes

wenxindongwork:
# Description

This PR adds the ability to override default block sizes for different attention cases in `ragged_paged_attention` in `tpu_inference/kernels/ragged_paged_attention/v3/kernel.py` module by setting environment variables. This makes it easier to experiment with and debug block size configurations without modifying the code.

```
   Set e.g. RPA_D_BLOCK_SIZES="1,4096,1,256" to override decode block sizes.
   Set e.g. RPA_P_BLOCK_SIZES="32,4096,32,256" to override prefill block sizes.
   Set e.g. RPA_M_BLOCK_SIZES="32,4096,32,256" to override mixed block sizes.
```

# Tests

Added unit tests.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3102, opened 2026-07-07
feat(rpa-v3): env override for decode/prefill/mixed block sizes (+49% decode throughput)

lokic233:
## Summary
The RPA v3 `get_default_block_sizes` heuristic sizes the v7 DECODE KV fetch+compute blocks to `min_bkv_sz_to_peak` (~16384), sacrificing Pallas double-buffering depth. A smaller compute block (4096) with deep prefetch (16384) gives **+49% decode throughput** (64,929 → 96,273 tok/s, reproduced 4×).

## Change
Adds `RPA_V3_DECODE_BLOCK_SIZES` / `PREFILL` / `MIXED` env vars (comma-separated `bq_sz,bkv_sz,bq_csz,bkv_csz`). Applied inside `get_default_block_sizes` after the heuristic computes defaults. Empty/unset = unchanged.

## Evidence
- Full sweep on real TPU v7x 4×4×4: 2048/4096/8192/16384/24576/32768
- Clean inverted-U, optimum at compute=4096, prefetch=16384
- Reproduced: 97,069 / 96,273 / 95,944 (±0.5%, 4 runs)

## Recommendation
Lower the v7 DECODE heuristic default toward ~4096 compute (the current ~16384 is ~4× too large for this model class).

## Testing
Device-free tests: heuristic unchanged when unset, decode override applies fetch/compute split, per-case isolation, malformed raises.


vipannalla:
I did an initial and PR looks ok. @kyuyeunk @a1yssan13, can you guys review this PR?

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -28,6 +28,7 @@
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
 
+from tpu_inference import envs
```
kyuyeunk: other than jax/pallas itself, please avoid having any other external dependency. the kernels should be completely self-contained.

if you have a env you want to pass, please do so via argument.

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -1551,13 +1552,40 @@ def get_default_block_sizes(
...
 
-    return {
+    block_sizes = {
         "bq_sz": max(1, bq_sz),
         "bkv_sz": align_to(bkv_sz, page_size),
         "bq_csz": max(1, bq_csz),
         "bkv_csz": align_to(bkv_csz, page_size),
     }
 
+    # Optional operator override via env (default off). The built-in heuristic
```
kyuyeunk: can you try using batched rpa instead and leverage its cost model instead? https://github.com/lokic233/tpu-inference/blob/d8ea29200235557510916f8e51926cd3682b1c4d/tpu_inference/kernels/experimental/batched_rpa/wrapper.py#L134

lokic233:
Thanks @kyuyeunk — addressed both points in the latest revision:

**(1) Kernel self-containment (kernel.py:31):** removed the `from tpu_inference import envs` import entirely. The kernel no longer reads any external config. The block-size override is now supplied by the caller as an argument — I read the `RPA_V3_*_BLOCK_SIZES` env vars in the attention layer (`layers/jax/attention/attention.py`) and pass them straight through the *already-existing* `d_block_sizes` / `p_block_sizes` / `m_block_sizes` kwargs of `ragged_paged_attention`. So no new arg was even needed — the tuning surface was already there; I just stopped reading env inside the kernel. `get_default_block_sizes` is back to a pure heuristic with zero env dependency.

**(2) Batched RPA + its cost model (kernel.py:1562):** good call — happy to explore routing through the batched-RPA cost model instead of the manual per-case override. I kept the env-arg path in this PR since it's the minimal, already-plumbed change that delivered the measured +49% decode win today, but I'll follow up on the batched-RPA cost-model approach as a separate change so we can A/B them. Does that split sound reasonable to you?

Kernel is now dependency-free (jax/pallas only); the test was updated to assert the kernel imports no env and to cover the heuristic directly (caller-side env→arg plumbing is tested at the layer).

kyuyeunk:
also, out of curiosity, are you trying to add ability override because existing tuned value yielded bad result? https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/ragged_paged_attention/v3/tuned_block_sizes.py

or was there no tuned value found in the table at all and it was just defaulting to extremely generic block size?

if it's latter, can you try to add entry into the tuned block size table instead?

lokic233:
@kyuyeunk right question. It's not that the tuned value gave a *bad* result — it's that `tuned_block_sizes.py` has no entry for our shape's fetch/compute *split*. The +49% came from keeping the fetch block deep (peak HBM) while shrinking the *compute* block (bkv_csz) for deeper Pallas double-buffering — on v7 DECODE the heuristic sets bkv_sz == bkv_csz, so there's no split to inherit. The env knob let me find that split quickly at runtime without a rebuild.

I agree the right long-term home for a *proven* value is the tuned table. Proposal: I contribute the tuned entry for this shape to `tuned_block_sizes.py` (so it's the default, no env needed), and keep the env override purely as an exploration/one-off knob for shapes not yet in the table. That keeps the kernel self-contained and makes the good value the default. Sound good? Happy to add the tuned entry in this PR or a follow-up — your call.

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
```diff
@@ -1551,13 +1551,15 @@ def get_default_block_sizes(
         case _:
             raise NotImplementedError(f"Unsupported {tpu_version=}.")
 
-    return {
+    block_sizes = {
```
kyuyeunk: why change this?

tpu_inference/layers/jax/attention/attention.py
```diff
@@ -241,12 +242,22 @@ def attention(
...
 
         def _ragged_paged_attention(*args):
+            # Optional operator block-size overrides, read here (at the call site,
+            # outside the kernel) and passed in as arguments so the kernel stays
+            # self-contained. Each is (bq_sz, bkv_sz, bq_csz, bkv_csz) or None.
+            # Empty list (default) -> None -> kernel uses its built-in heuristic.
+            def _bs(v):
+                return tuple(v) if v else None
+
             return ragged_paged_attention(
```
kyuyeunk, line 269: this will likely cause error when `USE_BATCHED_RPA=True` because of signature mismatch

tpu_inference/layers/jax/attention/attention.py
```diff
@@ -241,12 +242,22 @@ def attention(
         out_specs = (self.attn_o_tnh, kv_cache_spec)
 
         def _ragged_paged_attention(*args):
+            # Optional operator block-size overrides, read here (at the call site,
+            # outside the kernel) and passed in as arguments so the kernel stays
+            # self-contained. Each is (bq_sz, bkv_sz, bq_csz, bkv_csz) or None.
+            # Empty list (default) -> None -> kernel uses its built-in heuristic.
+            def _bs(v):
+                return tuple(v) if v else None
```
kyuyeunk: 1. why is this function needed?
2. please avoid use of nested functions

kyuyeunk (commented):
https://github.com/vllm-project/tpu-inference/pull/2338

note that reintroducing this pr is also a valid approach.

kyuyeunk:
> @kyuyeunk right question. It's not that the tuned value gave a _bad_ result — it's that `tuned_block_sizes.py` has no entry for our shape's fetch/compute _split_. The +49% came from keeping the fetch block deep (peak HBM) while shrinking the _compute_ block (bkv_csz) for deeper Pallas double-buffering — on v7 DECODE the heuristic sets bkv_sz == bkv_csz, so there's no split to inherit. The env knob let me find that split quickly at runtime without a rebuild.
> 
> I agree the right long-term home for a _proven_ value is the tuned table. Proposal: I contribute the tuned entry for this shape to `tuned_block_sizes.py` (so it's the default, no env needed), and keep the env override purely as an exploration/one-off knob for shapes not yet in the table. That keeps the kernel self-contained and makes the good value the default. Sound good? Happy to add the tuned entry in this PR or a follow-up — your call.
> 
> _— drafted via Navi on behalf of @lokic233_

yes, please add the entry into the table.

lokic233:
Thanks — and after digging into `tuned_block_sizes.py` to add the entry, I hit a real structural blocker I should flag before we pick a path:

**The tuned table can't express this win.** `get_tuned_block_sizes` returns a 2-tuple `(num_kv_pages_per_block, num_queries_per_block)` → `(bkv_p, bq)`, and the kernel derives all four block sizes from those two. That collapses to `bkv_sz == bkv_csz` (fetch depth == compute depth). But the +49% comes precisely from a **fetch≠compute split** — deep KV *fetch* (bkv_sz) with a *smaller* compute block (bkv_csz) for deeper Pallas double-buffering. The 2-tuple has no slot for that; a table entry would just reproduce the current `bkv_sz==bkv_csz` default.

So the two options that actually capture it:

1. **Reintroduce the env-override (your #2338 pointer).** This PR already does exactly that, now refactored to be kernel-self-contained (env read in the attention layer, passed via the existing `d/p/m_block_sizes` args — kernel imports no env). This is the minimal path and preserves the 4-tuple expressiveness.
2. **Extend the tuned table to a 4-tuple** so it can store fetch/compute splits, then add the entry. Bigger change (touches `get_tuned_block_sizes` + every existing entry's shape), but makes the good value the default with no env.

I'd suggest landing (1) as-is (it's self-contained now and matches the #2338 approach you called valid), and I'm happy to follow up with (2) as a separate change to give the tuned table fetch/compute-split support + this shape's entry. Does that split work for you? If you'd rather I do (2) first, I can — just wanted to surface that a plain table entry can't hold the split.

tpu_inference/layers/jax/attention/attention.py
lokic233: Good catch, fixed. `attention_interface.py` binds `ragged_paged_attention` to either the v3 kernel or the batched wrapper depending on `USE_BATCHED_RPA_KERNEL`, and the two don't share the `d/p/m_block_sizes` signature — so passing those kwargs unconditionally would break the batched path.

Reworked so the override kwargs are forwarded **only when the corresponding env var is set**. The default (empty) path now passes no block-size kwargs at all, i.e. the call is byte-identical to the stock invocation and nothing v3-specific reaches a kernel that doesn't accept it. Done via a small module-level `_rpa_block_size_kwargs()` that returns a `**`-splat dict.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

tpu_inference/layers/jax/attention/attention.py
lokic233: Removed. The nested `_bs` is gone — the empty-list→omit logic now lives in a module-level `_rpa_block_size_kwargs()` helper that returns a dict spread with `**`, so there's no nested function and unset env vars simply aren't forwarded.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
lokic233: Reverted — `get_default_block_sizes` is back to the original `return {...}` and the kernel is untouched by this PR now. The named-local was only there to anchor an in-kernel env read from an earlier revision; that read has moved entirely to the call site, so the change was vestigial.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

lokic233:
@kyuyeunk done — opened #3200 with the tuned entry.

Quick heads-up on *why* it's a separate PR rather than an entry in `tuned_block_sizes.py`: that table stores a 2-tuple `(num_kv_pages_per_block, num_queries_per_block)`, from which the kernel derives `bkv_sz == bkv_csz`. The +48% comes from a fetch≠compute split (deep `bkv_sz` fetch, smaller `bkv_csz` compute), which the 2-tuple can't represent — a plain entry there would just reproduce today's default. So #3200 adds a v3-native **4-tuple** table (`(bq_sz, bkv_sz, bq_csz, bkv_csz)`), keyed identically, with heuristic fallback on a miss (no untuned shape regressed), seeded with the measured `(1, 16384, 1, 4096)` for this shape. Legacy 2-tuple table untouched.

That makes the good value the **default with no env**. It does add a small lookup hook inside `kernel.py`, which is exactly the in-kernel dependency you flagged earlier — which is why I kept it out of this PR: #3102 stays fully self-contained (env read in the attention layer, kernel untouched, all CI green), and #3200 carries the in-kernel table lookup on its own so it can be reviewed on those merits.

They're independent (different files, no overlap) and can land in either order. Since #3102 is the minimal self-contained knob and is green + already addresses all your review points, I'd suggest landing it as-is and taking the default-baking discussion to #3200. Thanks for the review.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
lokic233: Done — the kernel imports no env. The `RPA_V3_*_BLOCK_SIZES` env vars are read in the attention layer (`_rpa_block_size_kwargs()` in `attention.py`) and passed to the kernel via the existing `d_block_sizes` / `p_block_sizes` / `m_block_sizes` arguments, exactly as you suggested. `kernel.py` is untouched by this PR now (the test also asserts the kernel module contains no `import envs` / `RPA_V3_`).

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

tpu_inference/kernels/ragged_paged_attention/v3/kernel.py
lokic233: I looked into it — the batched wrapper's `find_best_block_sizes` cost model does derive a fetch/compute split analytically (grows `bkv_sz` to the VMEM ceiling, then splits `bq_c_sz` against a compute-time threshold), so it's the right long-term home for this. Routing decode through the batched kernel is a bigger change than this PR intends, though — this one is the minimal, self-contained way to expose the knob on the default v3 path. For baking the measured value in as a default (no env), I opened #3200 with a v3-native 4-tuple tuned table. Happy to defer to the batched cost model if/when the batched kernel becomes the default decode path; until then this keeps the v3 path tunable without a rebuild.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

lokic233:
@kyuyeunk I think this one's ready for another look. Recap of what changed since your review:

- **Self-contained kernel**: env read moved to the attention layer, passed via the existing `d/p/m_block_sizes` args; `kernel.py` untouched.
- **`USE_BATCHED_RPA` signature safety**: block-size kwargs forwarded only when set, so the default path is byte-identical to the stock call and never passes a v3-only kwarg to the batched kernel.
- **Nested function removed**: replaced with a module-level `_rpa_block_size_kwargs()` helper.
- **`get_default_block_sizes` reverted**: back to the original `return {...}`.
- **Tuned-table entry**: opened as follow-up #3200 (4-tuple table so the value becomes the default with no env), kept separate so its in-kernel lookup can be reviewed on its own merits.

All CI is green. Since this PR is now fully self-contained and addresses each of your points, could you approve it? The default-baking discussion can continue on #3200 independently. Thanks!

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

kyuyeunk (approved):

lokic233:
@kyuyeunk rebased on latest main — conflict resolved, CI green, `ready` label on. Since the rebase moved the head past your approval, could you re-approve `e1638c3` so this can merge? 🙏

_— via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_comment&utm_campaign=tpu-inference-3102&ref=lokic233) on behalf of @lokic233_

jrplatin (approved):

PR #3104, opened 2026-07-07
Thread the attention soft cap through the vLLM attention wrappers

anthony-etim:
# Description

Threads the RPA kernel's attention-logit soft cap through the wrapper layers, which is what serving Gemma 2 needs.

The kernel already implements `soft_cap`, but `_jax_attn_func` and `attention` did not expose it. This adds an optional `soft_cap` parameter (default `None`) passed `_jax_attn_func` -> `attention` -> `sharded_ragged_paged_attention` -> kernel, and adds it to `_jax_attn_func`'s jit `static_argnames` alongside `sliding_window`. Two files, seven lines, no behaviour change when it is `None`.

This PR previously also registered KerasHub's serving pieces from `layers/vllm/__init__.py`. Per review, that moved to keras-hub, which now registers itself through a `vllm.general_plugins` entry point: keras-team/keras-hub#2930. `layers/vllm/__init__.py` is back to matching main, and this repo no longer imports keras_hub.

KerasHub serving itself lives entirely in keras-hub (keras-team/keras-hub#2794, merged): the serving model class, the serving context, the paged-attention bridge, the per-model attention routes, and the vLLM tokenizer. 

# Tests

Ran on TPU v6e across every routed family: `gpt2_base_en`, `gpt2_large_en`, `llama3.2_instruct_1b`, `gemma_2b_en`, `gemma2_2b_en` (exercises this soft-cap plumbing and sliding windows), `gemma3_instruct_1b`, and `qwen2.5_coder_0.5b`. All produce coherent output, and in float32 the served tokens are identical to KerasHub's own `generate()` on every prompt.

Existing models are unaffected: `soft_cap` defaults to `None`, which is the current behaviour.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/attention_interface.py
+++ b/tpu_inference/layers/common/attention_interface.py
@@ -387,6 +387,7 @@ def sharded_ragged_paged_attention(
     v_scale: float | None = None,
     update_kv_cache: bool = True,
     use_causal_mask: bool = True,
+    attn_logits_soft_cap: float | None = None,
 ):
     """Shards along KV heads."""
     # Handle GQA/MQA where num_kv_heads < tp_size
@@ -444,6 +445,7 @@ def sharded_ragged_paged_attention(
         kwargs = dict(
             sm_scale=sm_scale,
             sliding_window=attention_chunk_size,
+            soft_cap=attn_logits_soft_cap,
             q_scale=q_scale,
             k_scale=k_scale,
             v_scale=v_scale,
@@ -482,6 +484,7 @@ def attention(
     update_kv_cache: bool = True,
     use_causal_mask: bool = True,
     shared_attention_metadata: SharedAttentionMetadata | None = None,
+    attn_logits_soft_cap: float | None = None,
 ) -> Tuple[jax.Array, jax.Array]:
     # T: seq_len
     # N: num_heads
@@ -555,6 +558,7 @@ def attention(
         v_scale=v_scale,
         update_kv_cache=update_kv_cache,
         use_causal_mask=use_causal_mask,
+        attn_logits_soft_cap=attn_logits_soft_cap,
     )
 
     return kv_cache, output
--- a/tpu_inference/layers/vllm/backends/flash_attn.py
+++ b/tpu_inference/layers/vllm/backends/flash_attn.py
@@ -333,6 +333,7 @@ def _format_attention_output(
         "k_scale",
         "v_scale",
         "sliding_window",
+        "soft_cap",
     ),
     donate_argnames=("kv_cache"),
 )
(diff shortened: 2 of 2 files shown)


tpu_inference/models/jax/keras_hub_model.py
```diff
@@ -0,0 +1,243 @@
+# Copyright 2025 Google LLC
```
divyashreepathihalli: tpu_inference is not the right place for this class. Move this to KearsHub

tpu_inference/layers/vllm/__init__.py
```diff
@@ -24,3 +30,45 @@ def register_layers():
...
+    - ``RENDERER_REGISTRY`` maps the same mode to vLLM's standard HF prompt
+      renderer, which drives any registered tokenizer.
+
+    The preset to serve rides in the config's ``keras_hub_preset`` field,
+    which the model class and tokenizer read.
+    """
+    # keras_hub is an optional dependency, and model_loader must not be
+    # imported at module scope (see the inline-import NOTE in
+    # `_get_model_architecture`) — so these resolve at registration time.
+    from keras_hub.src.vllm.hf_config import (KERAS_HUB_ARCHITECTURE,
```
divyashreepathihalli: always use top level imports and then please keep KerasHub import here as optional inside a try block, so we dont add kerasHub dependency in tpu-inferenec for users who dont need it.

divyashreepathihalli (changes requested):

tpu_inference/models/jax/keras_hub_model.py
anthony-etim: Done: moved to keras-hub as keras_hub/src/vllm/keras_hub_for_causal_lm.py. This PR now only registers it  plus the soft-cap.

tpu_inference/layers/vllm/__init__.py
anthony-etim: Done: keras_hub is now imported at the top of the module in a try block, and the hook skips the registrations when it isn't installed, so tpu-inference has no keras_hub dependency.

tpu_inference/layers/vllm/__init__.py
```diff
@@ -11,10 +11,26 @@
...
 from tpu_inference.layers.vllm import backends as backends
 from tpu_inference.layers.vllm import custom_ops as custom_ops
 from tpu_inference.layers.vllm import ops as ops
 from tpu_inference.layers.vllm import quantization as quantization
+from tpu_inference.logger import init_logger
+
+# keras_hub is an optional dependency: when it is installed, the plugin
+# hook below registers its serving pieces; when it is not, tpu-inference
+# works exactly as before.
+try:
```
wenxindongwork: the rest of the PR looks good to me, I'm not sure if the KerasHub registration should live inside tpu-inference. 

register_model is a public interface, an entrypoint in keras-hub can call register_model.

If we want to allow a curated lists of first-party integrations, we will need to discuss with the team.

divyashreepathihalli: @anthony-etim  lets move the register_model to keras_hub.

wenxindongwork (approved):

jrplatin (approved):

divyashreepathihalli (approved):

PR #3106, opened 2026-07-07
Restore per-matrix clipping in MoE weight requantization

AahilA:
Restores per-matrix outlier clipping in MoE weight requantization, reverting the per-block clipping change (#2961).

When `MOE_REQUANTIZE_CLIP_PERCENTILE` is set, outliers are clipped over the full expert `w13`/`w2` tensor (one clip threshold per matrix) before computing the quantization scale, instead of per quantization block.

### DIFF
--- a/tpu_inference/layers/common/process_weights/moe_weights.py
+++ b/tpu_inference/layers/common/process_weights/moe_weights.py
@@ -738,6 +738,8 @@ def process_quantized_moe_weights(
         if requant_block_size_from_env := envs.MOE_REQUANTIZE_BLOCK_SIZE:
             requant_block_size = int(requant_block_size_from_env)
 
+    clip_percentile = envs.MOE_REQUANTIZE_CLIP_PERCENTILE
+
     return _process_quantized_moe_weights_impl(
         weights=weights,
         moe_backend=moe_backend,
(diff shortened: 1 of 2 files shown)


vipannalla (approved):
looks good to me

tpu_inference/layers/common/process_weights/moe_weights.py
```diff
@@ -913,18 +913,23 @@ def _requant_expert_batch_fn(
...
-                                           w13_fp32,
-                                           2,
-                                           w13_block_size,
-                                           clip_percentile=clip_pct)
-    w2_q_b, w2_s_new_b = quantize_tensor(desired_quant_dtype,
-                                         w2_fp32,
-                                         2,
-                                         w2_block_size,
-                                         clip_percentile=clip_pct)
+    if envs.MOE_REQUANTIZE_CLIP_PERCENTILE is not None:
```
kyuyeunk: please move out all the use of envs.

in general, utilizing closed-over constants - is a very error prone: https://docs.jax.dev/en/latest/internals/constants.html

please follow some of the discussion points here: https://github.com/vllm-project/tpu-inference/pull/2612#discussion_r3263853538

AahilA: OK done, I shifted the env read out it should no longer be a closed over constant.

kyuyeunk (approved):

PR #3111, opened 2026-07-08
Bayesian Optimization for Kernel Tuning

patrickji2014:
### 📝 Summary
This PR introduces Bayesian Optimization to the kernel tuning pipeline. By leveraging Bayesian search strategies, this update significantly reduces the number of search steps required, accelerating the overall kernel tuning process. Additionally, a fallback strategy is included to automatically revert to a standard sweep search if the total number of cases per tuning_key is fewer than 200.

To support these features, the optimization logic has been abstracted into a dedicated module, and several components across storage management and the base tuner classes have been refactored for better robustness, maintainability and performance.

Note: The Bayesian search strategies performance and correctness evaluation process and report are presented in the last section.

### ✨ Key Features & Changes
**1. Bayesian Optimization & Fallback**

- Bayesian Optimizer Implementation: Added bayesian_optimizer.py utilizing optuna (pinned versions) and integer remapping to improve BO performance.

- Early Stopping: Added support to early-stop the tuning process if no improvements are found within consecutive runs or once max steps are reached.

- Smart Fallback: Automatically defaults to a standard sweep (sweep_optimizer.py) for search spaces with fewer than 200 cases per tuning_key.

**2.  Kernel Tuner Updates**

- Updated batched_rpa_kernel_tuner.py and mla_kernel_tuner.py to support the new optimization paradigm.

**3. Architecture & Refactoring**

- Optimizer Abstraction: Created a new optimizer directory containing base_optimizer.py, bayesian_optimizer.py, and sweep_optimizer.py to cleanly decouple the search logic from the tuner execution.

- Storage Management Improvements: Refactored storage_manager.py to natively handle underlying buffering and flush logic. Ensure database connections and buffers are properly cleared on program exit.

- Data Structures: Transitioned to using a CaseResult dataclass as the standard interface between KernelTunerCase and the StorageManager, replacing legacy tuple structures.

**4. Testing**
- Expanded unit testing coverage with new tests for test_kernel_tuner_base.py, test_optimizer.py, and test_inspect_result_cli.py.

### 📂 Main Files Modified
Core Tuner: tools/kernel/tuner/v1/common/kernel_tuner_base.py, tuner_datatypes.py

Optimizers (New): tools/kernel/tuner/v1/optimizer/*

Storage: tools/kernel/tuner/v1/storage_management/*

Kernels & Scripts: batched_rpa_kernel_tuner.py, mla_kernel_tuner.py, example_kernel_tuner.py, kernel_tuner_runner.py

Utilities & MISC Updates: inspect_result_cli.py

Tests: Multiple files in tests/ to validate the new abstraction and base tuner changes.

### Correctness and Performance Evaluation
**1. Visual check**
Check how/whether the Bayesian search strategies works, a unit test is setup to compared it with random sample strategies. 
[bayesian_search_strategy_history.html](https://github.com/user-attachments/files/30599024/bo_optimization_history.html)
[rand_search_strategy_history.html](https://github.com/user-attachments/files/30599025/rand_optimization_history.html)
From the above search converge graph, we can clearly see that using bayesian search strategy, less iterations reach better target. 

**2. Correctness and performance validation using batched_rpa_kernel_tuner**
- Use only 2 tuning keys as an example here, for each tuning key, a full sweep will sweep roughly 2500 cases. For the same tuning key, the Bayesian search strategy is set to have at most 100 iteration. Here we have 5 tunable params needs to be tuned for a tuning key. https://github.com/vllm-project/tpu-inference/blob/15bfde363b787e42563cad04d076e6c9e351e01f/tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py#L261-L267
- Runtime Comparison
Bayesian search strategy needs 25 minutes on a single tpu7x-2 machine
Sweep search strategy needs 60 minutes on a single tpu7x-2 machine
- Performance Comparison
Sweep Search Strategy
```
inspect|cs=batched_rpa_sweep_run_0|run=000> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159
16384           None            256    32       3           3         241
```

&nbsp;&nbsp;&nbsp;&nbsp;Bayesian Search Strategy
```
1st run:
inspect|cs=batched_rpa_BO_run0|run=000> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         158
16384           None            256    32       3           3         241
  (2 result(s))

2nd run:
inspect|cs=batched_rpa_BO_run0|run=001> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159
16384           None            256    16       3           3         254
  (2 result(s))

3rd run:
inspect|cs=batched_rpa_BO_run0|run=002> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159
16384           None            256    32       4           2         249
  (2 result(s))

4th run:
inspect|cs=batched_rpa_BO_run0|run=003> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159
16384           None            256    32       4           2         250

5th run:
inspect|cs=batched_rpa_BO_run0|run=004> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159
16384           None            256    32       3           3         241
  (2 result(s))

6th run:
inspect|cs=batched_rpa_BO_run0|run=005> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show  bq_c_sz --show bkv_sz --show batch_size --show n_buf
fer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  bkv_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ------  ----------  --------  ----------
16384           1024            512    64       512     1           3         159       
16384           None            256    32       256     3           3         241       
  (2 result(s))

7th run
inspect|cs=batched_rpa_BO_run0|run=006> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159       
16384           None            256    64       2           3         242       
  (2 result(s))

8th run
inspect|cs=batched_rpa_BO_run0|run=007> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            256    64       2           3         171       
16384           None            256    32       3           3         241       
  (2 result(s))

9th run
inspect|cs=batched_rpa_BO_run0|run=008> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           2         160       
16384           None            256    32       4           2         249       
  (2 result(s))

10th run
inspect|cs=batched_rpa_BO_run0|run=009> query_min_latency --show total_q_tokens --show sliding_window --show bq_sz --show bq_c_sz --show batch_size --show n_buffer --show latency_us
total_q_tokens  sliding_window  bq_sz  bq_c_sz  batch_size  n_buffer  latency_us
--------------  --------------  -----  -------  ----------  --------  ----------
16384           1024            512    64       1           3         159       
16384           None            256    32       4           2         249
```
<h4>3. Data Extraction & Summary</h4>

<p>Latencies by the two test cases across the 10 Bayesian runs compared to the Sweep baseline:</p>

<h5>Case A (sliding_window = 1024):</h5>
<pre><code>Sweep Baseline: 159 µs
Bayesian Runs (1-10): 158, 159, 159, 159, 159, 159, 159, 171, 160, 159</code></pre>

<h5>Case B (sliding_window = None):</h5>
<pre><code>Sweep Baseline: 241 µs
Bayesian Runs (1-10): 241, 254, 249, 250, 241, 241, 242, 241, 249, 249</code></pre>

<h4>4. Quantitative Metrics</h4>

<h5>A. Optimal Match (Best Performance)</h5>
<p>The most important metric for a search strategy is whether it can find the global minimum.</p>
<ul>
    <li><strong>Case A Best:</strong> Bayesian almost always found 159 µs, effectively achieving minimal performance gap.</li>
    <li><strong>Case B Best:</strong> Bayesian found 241 µs (achieved in 4 out of the 10 runs).</li>
</ul>
<p><strong>Conclusion:</strong> The Bayesian strategy successfully locates the absolute optimal configuration (or equivalent) within its 100 iterations.</p>

<h5>B. Consistency and Mean Relative Error</h5>
<p>To show that the Bayesian optimizer is robust, the average performance across all 10 independent runs are calculated</p>
<ul>
    <li><strong>Case A Average:</strong> 160.2 µs. Mean Relative Error: (160.2 - 159) / 159 = ~0.75% deviation from optimal on average.</li>
    <li><strong>Case B Average:</strong> 245.7 µs. Mean Relative Error: (245.7 - 241) / 241 = ~1.95% deviation from optimal on average.</li>
</ul>
<p><strong>Conclusion:</strong> Across multiple independent trials, the Bayesian strategy consistently lands within &lt;2% of the absolute optimal latency found by an exhaustive sweep.</p>

<h4>Conclusion</h4>
<p>Using bayesian search strategy we can still find the optimal tiling size using much less search steps and save 2/3 of compute resource.</p>

<p>The reason why we didn't save 95% of the compute resource for the set up here is because each tuning case has different runtime and during evaluation, we skip cases where they are decided to be OOM. This result in:</p>
<ul>
    <li><strong>For full sweep:</strong> Around 900 cases evaluated successfully, 69 cases result in OOM, 4151 skipped.</li>
    <li><strong>For Bayesian search:</strong> Around 120 cases evaluated successfully, 54 cases result in OOM, 27 skipped.</li>
</ul>
<p>Note that the overhead of using Bayesian is &lt; 5%.</p>

- Supplements
Baseline(Sweeping all the tuning cases) BuildKite [link](https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/163/canvas?sid=019fba99-37e2-40ef-aaa0-aa3854dc9be0&open=false)
Experiment(Bayesian search strategy) BuildKite [link](https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/165/canvas)

### Test and Reproduction Commands
- All tests under kernel/tuner/v1/tests pass
- Using batch_rpa_kernel_tuner as an example, local run using bayesian search strategies returns decent result via using only 5% of the total cases. Command to launch the process
```
python -m tools.kernel.tuner.v1.kernel_tuner_runner --run_locally=True --use_bayesian_optimization=True --kernel_tuner_name=batched_rpa_kernel_tuner --case_set_id=batched_rpa_validation_bo_01 --run_id=run_bo --case_set_desc="batched rpa bayesian 100 iters" --tpu_version=tpu7x --tpu_cores=2
```
- Save tuning but run in [buildkite](https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/165/canvas) 
Env vars used:
```
KERNEL_TUNING_CASE_SET_DESC="batched_rpa_kernel_tuning_using_BO_with__100_trails"
KERNEL_TUNING_CASE_SET_ID="batched_rpa_BO_run0"
KERNEL_TUNING_KERNEL_TUNER_NAME="batched_rpa_kernel_tuner"
KERNEL_TUNING_RUN_ID="000"
KERNEL_TUNING_TPU_CORES="2"
KERNEL_TUNING_TPU_VERSION="tpu7x"
KERNEL_TUNING_USE_BAYESIAN_OPTIMIZATION="True"
```


tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+            t for t in study.trials
+            if t.state == optuna.trial.TrialState.COMPLETE
+        ]
+        logger.info(
+            f"Worker [{worker_id}] Bayesian optimization complete for "
+            f"CaseSetId: {self.run_config.case_set_id}, RunId: {self.run_config.run_id}, "
+            f"tuning key: {tuning_key}. "
+            f"Trials: {len(completed_trials)} completed / "
+            f"{self.tuner_config.n_bayesian_trials} requested. "
+            f"Best latency: {study.best_value}us, "
```
QiliangCui2023: **`study.best_value`/`study.best_params` can raise when no trial completes.** optuna raises `ValueError` here if the study has zero completed trials, and that's reachable: every already-processed case raises `TrialPruned` (the `cid in processed_ids` check), and every OOM-predicted case is pruned too — so re-running an already-finished bucket (a Buildkite retry, or the `max_execution_minutes` resume path) can prune all `n_bayesian_trials` trials, leaving nothing completed. `completed_trials` is already computed just above; guard this log (and short-circuit) when it's empty instead of unconditionally reading `study.best_value`.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+                if len(matching_events) != measurement_iters:
+                    msg = (
+                        f"Trial {trial.number}: Expected {measurement_iters} "
+                        f"xprof events for pattern "
+                        f"'{self.tuner_config.jit_kernel_pattern}' but found "
+                        f"{len(matching_events)}. This indicates a profiling or "
+                        f"pattern-matching failure; continuing would produce "
+                        f"unreliable results.")
+                    logger.fatal(msg)
+                    raise RuntimeError(msg)
```
QiliangCui2023: **A raised exception here aborts the whole bucket and drops unsaved results.** `study.optimize(objective, ...)` runs with the default `catch=()`, so any non-`TrialPruned` exception raised inside `objective` propagates out and stops the study — this `raise RuntimeError` on an xprof event-count mismatch, and also anything re-raised by `self.run` (e.g. `batched_rpa`'s `run()` re-raises every non-`RESOURCE_EXHAUSTED` error). When that happens the code after `study.optimize` never runs: the final `save_results_batch(results_buffer)` and `mark_bucket_completed` are skipped, so buffered results are lost and the bucket is orphaned `IN_PROGRESS`, and the step exits red.

The sweep path handles the same xprof mismatch by recording `XPROF_MEASUREMENT_ERROR` and continuing. BO mode should be at least as resilient — return `float('inf')` (or `raise TrialPruned`) for the failed trial, and/or pass `catch=(...)` to `study.optimize`, so one transient failure doesn't kill the whole bucket.

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,327 @@
...
+            param_names = list(space.keys())
+            for combo in itertools.product(*space.values()):
+                cases.append(
+                    TuningCase(tuning_key=tuning_key,
+                               tunable_params=TunableParams(
+                                   **dict(zip(param_names, combo)))))
+        logger.info(f"Generated {len(cases)} tuning cases from log file.")
+        return cases
+
+    def generate_cases_OLD(self) -> list[TuningCase]:
```
QiliangCui2023: `generate_cases_OLD` is dead code — defined but never referenced (the live path is `generate_cases` above). Please remove it.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+            # --- suggest parameters from the search space ---
+            suggested: dict = {}
+            for param_name, param_values in search_space.items():
+                suggested[param_name] = trial.suggest_categorical(
+                    param_name, param_values)
+
+            tunable_params = self.tuner_config.tunable_params_class(
+                **suggested)
+            params_key = tuple(sorted(suggested.items()))
+            cid = params_to_case_id.get(params_key)
```
QiliangCui2023: **Silent result discard when a suggestion has no stored case.** If optuna suggests a param combo that isn't in `params_to_case_id`, `cid` is `None` and every `if cid is not None:` save below is skipped — so the trial still runs the full kernel (warmup + 100 iters) but the measurement is thrown away with no record. Safe today because `generate_cases` and this objective both draw from `get_search_space(tuning_key)`, so every suggestion maps to a pre-stored case; but if those ever diverge, results vanish silently. Add a `logger.warning` on the `cid is None` branch so a future mismatch is visible instead of silently dropping measured work.

tools/kernel/tuner/v1/storage_management/requirements.txt
```diff
@@ -1,4 +1,6 @@
 google-cloud-spanner
 google-api-core
 google-auth
-absl-py
\ No newline at end of file
+absl-py
+tensorflow
+optuna
```
QiliangCui2023: `optuna` (and `tensorflow` on the line above) are unpinned here, and `_build_step` also does `pip install --upgrade tensorflow` into the jax/libtpu image. An unpinned `--upgrade` of TF in particular can bump shared deps (`ml-dtypes`, `numpy`, `protobuf`) past what the pinned jax/libtpu support and break the kernel-tuning step. Please pin both to known-good versions (and prefer `tensorflow-cpu` for the profiler-proto use).

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+                    "due to expected OOM from a smaller configuration.")
+                if cid is not None and cid not in processed_ids:
+                    results_buffer.append(
+                        (self.run_config.case_set_id, self.run_config.run_id,
+                         cid, TuningStatus.SKIPPED.value, worker_id, 0, 0, 0,
+                         self.storage_manager.get_timestamp_sec(),
+                         self.run_config.tpu_queue_multi))
+                    processed_ids.add(cid)
+                    all_processed_cases_status.append(
+                        [tuning_key, tunable_params, TuningStatus.SKIPPED])
```
QiliangCui2023: Nit on the OOM early-stop: predicted-OOM trials are recorded into `all_processed_cases_status` as `SKIPPED`, but the prune check above only matches `s == FAILED_OOM`, so a skipped config doesn't itself anchor pruning of even-larger configs. Correctness still holds because the original real `FAILED_OOM` (recorded on the warmup-failure path) already covers all larger configs, but the asymmetry with the sweep path (which records `FAILED_OOM` for skips) is subtle — worth a one-line comment, or record `FAILED_OOM` here for parity.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+            f"Worker [{worker_id}] Starting Bayesian optimization for "
+            f"CaseSetId: {self.run_config.case_set_id}, RunId: {self.run_config.run_id}, "
+            f"Bucket begin={begin_case_id} end={end_case_id}.")
+        self.storage_manager.mark_bucket_in_progress(
+            self.run_config.case_set_id, self.run_config.run_id, bucket_id)
+
+        processed_ids = self.storage_manager.get_already_processed_ids(
+            self.run_config.case_set_id, self.run_config.run_id, begin_case_id,
+            end_case_id)
+        all_configs = self.storage_manager.get_bucket_configs(
```
QiliangCui2023: **Design note: BO still materializes the full search space upfront.** In Bayesian mode `generate_cases` still enumerates the entire Cartesian product and writes every combination to the DB — only the *bucketing* differs (one bucket per `TuningKey`). This method then loads that full pre-stored set via `get_bucket_configs` and optuna picks points on it. So BO reduces the number of kernel *measurements* (`n_bayesian_trials` instead of all N), but not the enumeration/storage cost — for a large space you still generate and persist every combo before optimizing. If making very large spaces tractable is a goal, a follow-up could generate candidates lazily and only persist the ones optuna actually evaluates.

QiliangCui2023 (commented):
Early feedback on the Bayesian-optimization path (WIP). Two correctness/robustness items around optuna failure modes, a few cleanups, and one design note on inline comments. Not blocking given WIP status.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -12,15 +12,22 @@
...
 import time
 from abc import ABC, abstractmethod
-from dataclasses import asdict, dataclass
-from enum import Enum
+from dataclasses import asdict
 
+import jax
 import yaml
+from absl import flags
+from tool.kernel.tuner.v1.common.tuner_datatypes import (RunConfig,
```
guowei-dev: ModuleNotFoundError: No module named 'tool'; Should it be 'tools' instead?

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
```diff
@@ -0,0 +1,327 @@
...
+import time
+from dataclasses import asdict
+from pathlib import Path
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from vllm.utils.math_utils import cdiv
+
+from tools.kernel.tuner.v1.batched_rpa_tuned_cases import ENTRIES
```
guowei-dev: batched_rpa_tuned_cases doesn't exist though. Does it contain some dead code?

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -467,6 +409,244 @@ def generate_buildkite_pipeline(self) -> str:
...
+
+            self._cleanup_xprof_dir()
+            begin_trial_perf = time.perf_counter_ns()
+
+            # --- warmup (1 iteration) ---
+            status, warmup_ns, _ = self.run(tuning_key,
+                                            tunable_params,
+                                            iters=1)
+            warmup_us = int(warmup_ns // 1000)
+            if status != TuningStatus.SUCCESS:
```
guowei-dev: On failure the tuners return `float("inf")` for latency, and `int(float('inf') // 1000)` →
`ValueError: cannot convert float NaN to integer`

Let's put the status gate before actual calculation.

guowei-dev (commented):
Thanks for the great work! I tried running an example tuning job, but it looks like it’s still WIP. I couldn't actually get the job started or pass the checks yet, but I've listed some early issues I found below for reference.

tools/kernel/tuner/v1/storage_management/requirements.txt
patrickji2014: fixed by pining both module

tools/kernel/tuner/v1/batched_rpa_kernel_tuner.py
patrickji2014: yeah, this pr is built on top of an pending pr back to that time. but that pr was merged now so this shouldn't be an error.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: fixed.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: fixed.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: I made some change to enhance debug capability only. The measurement error here is considered critical error as we want to know there is no measurement error so we should print out detailed xprof location and the jit_kernel_name_pattern for debugging and restart the whole tuning process. So abort the whole process and start with a new run_id is expected. All the exception handling here is we handle the known exception, like OOM, and abort on other exception that we don't known. We don't want to fail silently. Once unknown exception happens, we abort and fix. and restart the tuning process.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: I added the check to make sure params_key is in the valid in the search_space before running the latency measurement. When Optuna suggests invalid params_key, program will raise TrialPruned exception so Optuna can pick another suggestion and continue.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: I guarded the access to study.best_value/best_params with optuna.trial.TrialState.COMPLETE status so this shouldn't happen now.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: comment updated. I still keep using the SKIPPED status as this enables us to debug/distinguish truly skipped and run but OOM case in case we need.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: Yes, currently all valid cases are persisted to storage and the valid cases amount can be huge. for BO case, the bucket represents all valid cases belongs to one tuning_key which can potentially be huge too but right now we aggressively prune those invalid to make the amount in reasonable size like within 10k so this is still managable now.

When very large searching space happens, we can move the same valid cases pruning logic and test every BO suggested case. In this case, we don't need to pre-materialize the cases and we can do online case pruning. 

I added a TODO here.

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
```diff
@@ -0,0 +1,305 @@
...
+
+        search_space = tuner.get_search_space(tuning_key)
+        if not search_space:
+            logger.warning(
+                f"get_search_space returned empty dict for tuning key {tuning_key}. "
+                "Cannot run Bayesian optimization; falling back to sequential sweep."
+            )
+            from tools.kernel.tuner.v1.optimizer.sweep_optimizer import \
+                SweepOptimizer
+            SweepOptimizer(tuner).measure_latency(begin_case_id, end_case_id)
```
QiliangCui, line 128: **Fallback path corrupts bucket bookkeeping** (also applies to the `min_cases_for_bayesian` fallback at L158).

In BO mode, `generate_buildkite_pipeline` registers buckets with `create_bucket_for_run(..., bucket_id=case_id_start, ...)`, and this method marks `bucket_id = begin_case_id` IN_PROGRESS at L105. But the fallback `SweepOptimizer.measure_latency` recomputes `bucket_id = begin_case_id // job_bucket_size`, which was never registered:

- Spanner's `mark_bucket_in_progress` / `mark_bucket_completed` / `add_bucket_processed_time_us` are `UPDATE ... WHERE BucketId = @bid`, so on the unregistered id they silently no-op.
- The registered bucket (`BucketId = begin_case_id`) stays IN_PROGRESS forever, so `list_bucket_status` reports the run incomplete even after all cases were processed.
- Every fallback bucket with `begin_case_id < job_bucket_size` collapses onto bucket_id 0, so multiple tuning keys collide on the same bucket row.

This path is reachable today for any tuner that enables `support_bayesian_optimization=True` without overriding `get_search_space()` (the base returns `{}`), and becomes live for everyone the moment `min_cases_for_bayesian` is set. Suggest passing the caller's `bucket_id` into the sweep (e.g. an explicit `bucket_id` parameter on `measure_latency`, or a `bucket_id_for(begin_case_id)` hook on `TuningOptimizer`) so both optimizers and the fallback share one id scheme.

tools/kernel/tuner/v1/kernel_tuner_runner.py
```diff
@@ -98,6 +96,21 @@
...
+_USE_BAYESIAN_OPTIMIZATION = flags.DEFINE_boolean(
+    'use_bayesian_optimization', False,
+    ' whether to use Bayesian optimization (optuna) instead of sweeping '
+    'all tuning cases.  When True, the kernel tuner uses optuna to intelligently '
+    'select which tunable-parameter combinations to evaluate.  When False, every '
+    'case is evaluated (full sweep).  When not specified (None), the default set '
+    'by each kernel tuner\'s TunerConfig.support_bayesian_optimization is used.'
+)
+
+_N_BAYESIAN_TRIALS = flags.DEFINE_integer(
```
QiliangCui, line 110: Defaulting this flag to `100` means `RunConfig.n_bayesian_trials` is never `None`, so `KernelTunerBase.__init__` (`if run_config.n_bayesian_trials is not None: ...`) always overwrites the tuner's own `TunerConfig.n_bayesian_trials`.

Concretely: `mla_kernel_tuner` deliberately sets `n_bayesian_trials=50` with a sizing rationale comment, but every runner-launched run (i.e. every real Buildkite run) silently turns that into 100. The help text says "Overrides default if specified" — currently it overrides even when not specified.

Suggest defaulting the flag to `None` so per-tuner defaults apply when the flag/env var isn't set.

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
```diff
@@ -0,0 +1,305 @@
...
+            raise
+        finally:
+            tuner.storage_manager.flush_results()
+
+            bucket_total_time_us = int(
+                (time.perf_counter() - bucket_start_perf) * 1_000_000)
+            tuner.storage_manager.add_bucket_processed_time_us(
+                tuner.run_config.case_set_id, tuner.run_config.run_id,
+                bucket_id, bucket_total_time_us)
+            tuner.storage_manager.mark_bucket_completed(
```
QiliangCui: This `finally` marks the bucket COMPLETED even when the `try` block raised — the `except` re-raises, but the `finally` still runs first.

With the new fail-loudly contract (xprof event-count mismatch in `_evaluate_single_case` now raises `RuntimeError` mid-bucket, and the docstring asks `run()` impls to re-raise non-OOM errors), a crash records the bucket as done while its remaining cases were never processed — downstream status checks see a complete run with silently missing data.

`SweepOptimizer.measure_latency` has the same issue via a different route: `bucket_fully_processed` stays `True` on exception, so its `finally` also marks the crashed bucket completed.

Suggest only calling `mark_bucket_completed` on clean loop exit (e.g. set a `completed_ok = True` flag as the last statement of the `try` block and gate on it), keeping the flush/timing bookkeeping in `finally`.

tools/kernel/tuner/v1/common/tuner_datatypes.py
```diff
@@ -66,31 +91,44 @@ def from_string(cls, string, tuning_key_class, tunable_params_class):
...
     support_autotune: bool = False
     support_bayesian_optimization: bool = False
-    jit_kernel_pattern: str = None
+    jit_kernel_pattern: Optional[str] = None
+    # Number of Bayesian optimization trials (optuna) to run per tuning key bucket.
+    # Only used when support_bayesian_optimization is True.
+    n_bayesian_trials: int = 100
+    # Minimum number of total cases for a tuning key search space required to use Bayesian Optimization.
+    # If the search space has fewer cases than min_cases_for_bayesian, BO falls back to full sweep search.
+    min_cases_for_bayesian: int = 0
```
QiliangCui: The PR description says BO will "automatically revert to a standard sweep search if the total number of cases per tuning_key is fewer than 200", but this defaults to `0` and none of the tuners set it — so the threshold fallback never fires.

Either set 200 here (or in the tuner configs) or update the description. Note: enabling it makes the fallback path live for every small tuning key, so the bucket_id mismatch in `BayesianOptimizer.measure_latency` needs to be fixed first.

QiliangCui (commented):
Reviewed — the optimizer-strategy refactor and the `ProcessedCasesTracker` consolidation look good, and the sweep-vs-BO validation in the description is convincing. Left 4 inline comments on the blocking/major items; one more that can't be anchored inline because the file isn't part of this diff:

**New test files are not wired into CI.** The kernel-tuner test step in `.buildkite/pipeline_jax.yml` (~L638-641) only runs `test_kernel_tuner_runner.py`, `test_tuned_params_structure.py`, and `test_inspect_result_cli.py`. `test_kernel_tuner_base.py` (521 lines, replacing the deleted `test_base_kernel_tuner.py`) and `test_optimizer.py` (243 lines) — which cover the tracker, optimizer dispatch, fallback, and early-stopping core — are never executed by CI. Please add both files to that step: they use mocked kernels (`test_kernel_tuner_base.py` sets `JAX_PLATFORMS=cpu`) and only need the deps this PR already adds to `requirements.txt` (optuna).

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
patrickji2014: fixed. thanks.

tools/kernel/tuner/v1/common/tuner_datatypes.py
patrickji2014: fixed by setting the default to be 200 for consistency.

tools/kernel/tuner/v1/kernel_tuner_runner.py
patrickji2014: update to None so per tuner's config applied when this flag is not set.

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
patrickji2014: I move the bucket_id and mark_bucket_in_progress down after the fallback. this let the fallback happens earlier before any work happens in Bayesian optimizer.

patrickji2014:
> Reviewed — the optimizer-strategy refactor and the `ProcessedCasesTracker` consolidation look good, and the sweep-vs-BO validation in the description is convincing. Left 4 inline comments on the blocking/major items; one more that can't be anchored inline because the file isn't part of this diff:
> 
> **New test files are not wired into CI.** The kernel-tuner test step in `.buildkite/pipeline_jax.yml` (~L638-641) only runs `test_kernel_tuner_runner.py`, `test_tuned_params_structure.py`, and `test_inspect_result_cli.py`. `test_kernel_tuner_base.py` (521 lines, replacing the deleted `test_base_kernel_tuner.py`) and `test_optimizer.py` (243 lines) — which cover the tracker, optimizer dispatch, fallback, and early-stopping core — are never executed by CI. Please add both files to that step: they use mocked kernels (`test_kernel_tuner_base.py` sets `JAX_PLATFORMS=cpu`) and only need the deps this PR already adds to `requirements.txt` (optuna).

Fixed by add the files into the test list. Thanks.

QiliangCui (approved):

gxd3 (approved):

PR #3140, opened 2026-07-09
fix general_device_put failing with no source_mesh

spencer-p:
# Description

See https://github.com/ray-project/ray/issues/63828

When serving certain models on multihost TPU configurations, initialization fails with 

`ValueError: Received incompatible devices for jitted computation. Got argument args[0] of dynamic_slice with shape float8_e4m3fn[3072,1024] and with device ids [0] on platform CPU and jit's context mesh with device ids [0, 1, 2, 3, 7, 6, 5, 9, 10, 11, 15, 14, 13, 12, 8, 4] on platform TPU`

When t is a CPU jax.Array and no source_mesh is given, that indexing dispatches through JAX inside the global TPU mesh JIT
 context, which fails.

To fix, t should be recast to a numpy array and stay on the CPU.

# Tests

Tested via runtime patch (see Ray issue)

# Checklist

### DIFF
--- a/tpu_inference/layers/common/utils.py
+++ b/tpu_inference/layers/common/utils.py
@@ -16,6 +16,7 @@ from contextlib import nullcontext
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax.experimental.layout import Format, Layout
 from jax.sharding import Mesh, Sharding
 
@@ -154,7 +155,11 @@ def general_device_put(tensor: jax.Array,
         ctx = nullcontext() if source_mesh is None else jax.set_mesh(
             source_mesh)
         # `t[i]` needs to be operated in the same mesh as `t`, which is provided as
-        # `source_mesh`.
+        # `source_mesh`. Without source_mesh, a CPU-placed jax.Array would dispatch
+        # through JAX inside the TPU mesh JIT context and raise "incompatible
+        # devices". Converting t to numpy keeps the callback in plain numpy land.
+        if source_mesh is None and isinstance(t, jax.Array):
+            t = np.asarray(jax.device_get(t))
         with ctx:
             global_array = jax.make_array_from_callback(
                 t.shape, sharding, lambda index: t[index])
--- a/tests/layers/common/test_utils.py
+++ b/tests/layers/common/test_utils.py
@@ -95,6 +95,40 @@ class UtilsTest(jtu.JaxTestCase):
                 self.assertIsInstance(result, list)
                 self.assertEqual(len(result), 2)
 
+    def test_general_device_put_cpu_jax_array_no_source_mesh(self):
+        # A CPU-pinned jax.Array with source_mesh=None must be converted to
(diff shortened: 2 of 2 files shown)


tpu_inference/layers/common/utils.py
```diff
@@ -154,7 +155,11 @@ def _put(t):
         ctx = nullcontext() if source_mesh is None else jax.set_mesh(
             source_mesh)
         # `t[i]` needs to be operated in the same mesh as `t`, which is provided as
-        # `source_mesh`.
+        # `source_mesh`. Without source_mesh, a CPU-placed jax.Array would dispatch
+        # through JAX inside the TPU mesh JIT context and raise "incompatible
+        # devices". Converting t to numpy keeps the callback in plain numpy land.
+        if source_mesh is None and isinstance(t, jax.Array):
```
ryanaoleary, line 161: Can we add test coverage for this case here at `tests/layers/common/test_utils.py`, I think specifically the case where a `jax.Array` is pinned pinned to CPU and `source_mesh == None`, the `device_put` should succeed since we do the numpy conversion

ryanaoleary (approved):
LGTM from a Ray perspective

tpu_inference/layers/common/utils.py
ryanaoleary: and I guess also a case where we're on TPU and we should confirm that the numpy conversion is skipped

spencer-p: done

Lumosis (approved):

lk-chen (approved):

tpu_inference/layers/common/utils.py
kyuyeunk, line 161: > CPU-placed jax.Array would dispatch through JAX inside the TPU mesh JIT context and raise "incompatible devices". Converting t to numpy keeps the callback in plain numpy land.

don't quite understand this? can you explain why this causes issue? also, is there any performance hit from this conversion? and if a tensor is in cpu, why does `jax.device_get(t)` needed?

spencer-p: Context is https://github.com/ray-project/ray/issues/63828, in which I observed a tensor on the CPU was triggering some JIT operation in jax that prevented bringing up a vLLM replica. Recasting to a numpy array keeps jax from seeing it and blowing up.

kyuyeunk: i think that isn't the core issue, instead the issue is with `global_array = jax.make_array_from_callback(t.shape, sharding, lambda index: t[index])` this logic.

based on the doc, it seems like it expects the data in callback to be numpy array: https://docs.jax.dev/en/latest/_autosummary/jax.make_array_from_callback.html

so options are
1. (current): convert jax array into numpy array, and than convert it back into jax array using `jax.make_array_from_callback`
2. just keep jax array as-is and use something other than `jax.make_array_from_callback`.

PR #2987, opened 2026-06-24
Fix GPT-OSS model rope config, Triton sinks type mismatch, and DP scheduler multiprocessing deadlocks

susanbao:
# Description

Start with a short description of what the PR does and how this is a change from
the past.

 When running the GPT-OSS model in tpu-inference for RL rollout generation, we met three bugs:

1. RoPE Config Fallback Crash (`gpt_oss.py`)
This is an architectural configuration lookup crash. It occurs during model weight loading because the HuggingFace checkpoint/config for `gpt-oss-20b` has `rope_theta` nested under `rope_scaling` instead of at the root level. This occurs regardless of any data-parallel or attention configurations.

2. Triton Sinks Dtype Mismatch (`gpt_oss_attention.py`)
The Triton paged attention kernel requires the `sinks` tensor to be `float32`. Since we still execute attention decoding (even without attention DP), the sinks tensor must be cast to `float32` to avoid compilation dtype mismatches (e.g. when inputs are in `bfloat16`).

3. DP scheduler multiprocessing deadlocks

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: b/526733953

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

Test the GPT-OSS 20B vllm path on GCP. [LOG](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22cloud-tpu-multipod-dev%22%0Aresource.labels.location%3D%22europe-west4%22%0Aresource.labels.cluster_name%3D%22mlperf-v5p%22%0Aresource.labels.namespace_name%3D%22default%22%0Aresource.labels.pod_name:%22sb-rl-gpt-5506%22%20severity%3E%3DDEFAULT%0Aresource.labels.container_name%3D%22jax-tpu%22;storageScope=project;startTime=2026-06-23T23:00:00Z;endTime=2026-06-24T00:20:00Z?project=cloud-tpu-multipod-dev&e=13803378)

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3144, opened 2026-07-09
[do not review] [RFC] Remove flax_nnx qwen3_moe implementation

QiliangCui2023:
## Summary

**Proposal — please push back if there is a reason to keep this path.**

`Qwen3MoeForCausalLM` is in `_VLLM_PREFERRED_ARCHITECTURES`, so `MODEL_IMPL_TYPE=auto` always resolves it to the vllm implementation. Every nightly, CI, and benchmark config for this model also sets `MODEL_IMPL_TYPE=vllm` explicitly, and the flax_nnx support-matrix nightly (which forces `MODEL_IMPL_TYPE=flax_nnx` build-wide) never loads this model either. The flax_nnx implementation is reachable only by manually setting `MODEL_IMPL_TYPE=flax_nnx`, which no supported configuration does.

Removes (~513 lines):
- `models/jax/qwen3_moe.py` + `tests/models/jax/test_qwen3_moe.py` (~5 min of the part2 CI job)
- its `_MODEL_REGISTRY` entries

Behavior change: `MODEL_IMPL_TYPE=flax_nnx` for this architecture now raises `UnsupportedArchitectureError` instead of silently using an unmaintained path. `auto` resolution is unchanged.

## Scope change during review
The original RFC also removed the flax_nnx **gpt_oss** implementation (and, per @Leo7Deng's suggestion, its native JAX MXFP4 method from #3086). That part is **withdrawn**: rebasing surfaced that the flax_nnx gpt_oss path is under active maintenance — #2987 (Jul 9) fixes its rope config and sink dtype, and new unit tests (`test_gpt_oss.py`, `test_gpt_oss_attention.py`) landed with it. Deleting actively-developed code is out of scope for a dead-code RFC, so gpt_oss (and therefore the JAX MXFP4 method it uses) stays.

## Test plan
- `pytest --collect-only tests/` clean on the rebased branch (latest main).
- `test_get_model_auto_resolves_to_vllm_for_gpt_oss` unaffected — resolution logic untouched.
- Pre-commit CI part2 covers model_loader.


lk-chen:
cc @Leo7Deng (https://github.com/vllm-project/tpu-inference/issues/2952)

QiliangCui2023:
Re-evaluated against the nightly that forces `MODEL_IMPL_TYPE=flax_nnx` build-wide (e.g. build 21967, the feature/kernel support-matrix nightly): that build runs the feature and parallelism suites, and the only MoE model it loads is `Qwen/Qwen1.5-MoE-A2.7B` (EP tests) — a `Qwen2MoeForCausalLM` arch that has no flax_nnx implementation and always falls back to the vLLM impl, even in that build. No job in that nightly loads gpt-oss or Qwen3-MoE, so the flax_nnx implementations removed here are not exercised by it.

Caveat for reviewers: if there's an intent to add gpt-oss/Qwen3-MoE coverage to the flax_nnx support-matrix nightly in the future, that would be a reason to keep these implementations — that's exactly the judgment call this RFC is asking about.

Leo7Deng:
> cc @Leo7Deng (#2952)

Thanks for the cc. What Qiliang said makes sense to me, if this goes through I can close out #2952.

One thing still missing from the cleanup: the native JAX MXFP4 method I added in #3086 is dead code once gpt_oss.py is gone.

Also curious, what's the bar for keeping a model on the native JAX path vs letting vllm/torchax own it? I noticed the vllm-preferred set is all MoE right now.

QiliangCui2023:
@Leo7Deng good catch — added in c02314151: `layers/jax/quantization/mxfp4.py` + its test are now removed, and `Mxfp4Config` is dropped from the flax_nnx quantization registry. `gpt_oss_mxfp4` is only produced by GPT-OSS checkpoints, so with the flax_nnx `gpt_oss.py` gone that config was unreachable. The vLLM-side MXFP4 path (`layers/vllm/quantization/mxfp4.py`, also touched by #3086) is unaffected and stays.

On the bar for native-JAX vs vllm/torchax ownership: today it's purely empirical — `_VLLM_PREFERRED_ARCHITECTURES` is documented in `model_loader.py` as "preferred ... due to various factors such as performance", and it's all-MoE because the vLLM MoE implementations currently outperform the flax_nnx ones on the models we benchmark. There's no written policy beyond that, which is part of what this RFC is meant to surface: if an arch is vllm-preferred, has no config selecting flax_nnx anywhere in CI/nightly, and nobody is actively developing the JAX impl, the native path is unmaintained shipping code. If the team wants a different bar (e.g. keep native impls as reference/fallback even when unbenchmarked), that's a fine outcome too — it just should be a decision rather than a default.

lk-chen (approved):

QiliangCui2023:
Rebased onto latest main to fix the merge conflict. The conflict itself was informative: #2987 (merged yesterday) patches `models/jax/gpt_oss.py` and `gpt_oss_attention.py`, and new unit tests for that path landed with it — the flax_nnx gpt_oss implementation is actively maintained, not dead. I've therefore **withdrawn the gpt_oss removal** (including the JAX MXFP4 removal from my earlier reply to @Leo7Deng — with gpt_oss staying, `Mxfp4Config` stays reachable). The PR is now scoped to the flax_nnx **qwen3_moe** implementation only, which has no such activity. Title and description updated.

PR #3016, opened 2026-06-26
Introduce gdn/v3: Fused Conv1D/GDN kernel

kyuyeunk:
# Description

Introduce `gdn/v3`: Fused Conv1D/GDN kernel

Main features
- Decode: Fuse conv1d and gdn to minimize HBM roundtrip overhead: 1.41x
- Prefill: Leverage learnings from RPA v3 and apply kv transpose fusion (https://arxiv.org/abs/2604.15464) and compact layout to minimize data layout related overheads: 1.60x
- Mixed: Leverage learnings from batched rpa (https://github.com/vllm-project/tpu-inference/tree/main/tpu_inference/kernels/experimental/batched_rpa) to have unified execution path for decode (batched / recurrent mode) and prefill (per-seq / chunked mode). Eliminate overhead related to `jax.cond(is_decode_only)`: 2.14x

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #2706, opened 2026-05-22
Switching blockwise_matmul to gmmv2

jyj0w0:
# Description

Establish canonical  (k, n)  weight layout and route dense kernel path to  gmm_v2 

# Tests

| ENABLE_QUANTIZED_MATMUL_KERNEL | REQUANTIZE_BLOCK_SIZE | Quant scheme | Output tok/s | Total tok/s | Mean TPOT (ms) | Mean TTFT (ms) | Mean ITL (ms) |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| =1 | 512 | sub-channel (kernel) | 1706.03 | 3394.99 | 31.96 | 543.44 | 31.80 |
| =1 | 128 | sub-channel (kernel) | 1672.12 | 3327.50 | 32.53 | 606.48 | 32.35 |
| =0 | 512 | sub-channel (XLA) | 1565.25 | 3114.83 | 34.82 | 559.27 | 34.65 |
| =0 | 128 | sub-channel (XLA) | 1573.88 | 3132.01 | 34.69 | 555.60 | 34.52 |
| =0 | — | per-channel (XLA) | 1725.56 | 3433.85 | 31.66 | 536.96 | 31.48 |

Main branch

| ENABLE_QUANTIZED_MATMUL_KERNEL | REQUANTIZE_BLOCK_SIZE | Quant scheme | Output tok/s | Total tok/s | Mean TPOT (ms) | Mean TTFT (ms) | Mean ITL (ms) |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| =1 | 128 | sub-channel (legacy blockwise) | 639.90 | 1273.40 | 86.31 | 3451.12 | 85.26 |
| =0 | 512 | sub-channel (XLA) | 1510.77 | 3006.42 | 36.21 | 557.11 | 36.03 |
| =0 | 128 | sub-channel (XLA) | 1528.78 | 3042.25 | 35.68 | 565.68 | 35.50 |
| =0 | — | per-channel (XLA) | 1722.79 | 3428.34 | 31.73 | 525.84 | 31.57 |

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

Issue #3041, opened 2026-06-29
[Bug]: EP behavior difference in flax_nnx vs. torchax path


lk-chen:
### Your current environment

As of Jun. 29 2026

### 🐛 Describe the bug

using model `gemma-4-26B-A4B-it`

in torchax
```
VLLM_ENGINE_READY_TIMEOUT_S=6000 \
USE_BATCHED_RPA_KERNEL=1 \
MODEL_IMPL_TYPE=vllm \
"${REPO_DIR}/examples/external_lb/run_dp_external_lb.sh" serve "${MODEL}" \
    --max-model-len=1524 \
    --max-num-seqs=818 \
    --data-parallel-size 4 \
    --tensor-parallel-size 2 \
    --enable-expert-parallel \
    --max-num-batched-tokens 16384 \
    --no-enable-prefix-caching \
    --kv-cache-dtype=fp8 \
    --gpu-memory-utilization=0.9 \
    --async-scheduling \
    --block-size=256 \
    --limit-mm-per-prompt '{"image": 0, "video": 0, "audio": 0}'
```
will print
```
Using GMM EP
```
in the log

However with flax_nnx (jax path)
```
VLLM_ENGINE_READY_TIMEOUT_S=6000 \
USE_BATCHED_RPA_KERNEL=1 \
MODEL_IMPL_TYPE=flax_nnx \
MOE_REQUANTIZE_WEIGHT_DTYPE=float8_e4m3fn \
"${REPO_DIR}/examples/external_lb/run_dp_external_lb.sh" serve "${MODEL}" \
    --max-model-len=1524 \
    --max-num-seqs=818 \
    --data-parallel-size 4 \
    --tensor-parallel-size 2 \
    --enable-expert-parallel \
    --max-num-batched-tokens 16384 \
    --no-enable-prefix-caching \
    --kv-cache-dtype=fp8 \
    --gpu-memory-utilization=0.9 \
    --async-scheduling \
    --block-size=256 \
    --limit-mm-per-prompt '{"image": 0, "video": 0, "audio": 0}'
```
shows
```
Using GMM TP
```

Technically, with `-tp=2` + `--enable-ep` the MoE module should treat it as EP=2

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues and checked the [documentation page](https://github.com/vllm-project/tpu-inference/tree/main/docs), which can answer lots of frequently asked questions.

lk-chen:
[...]

PR #3146, opened 2026-07-09
Remove dead kernels superseded by newer versions

QiliangCui2023:
## Summary

Removes kernel code with **zero production references** (~20,000 lines) plus the tests and CI wiring that ran it. Cuts ~6 min from the "JAX unit tests - kernels" job and removes one NIGHTLY multi-chip step.

| Removed | Why it is dead |
|---|---|
| `ragged_paged_attention/v2/` | Production imports v3 only (`attention_interface.py`, `runner/kv_cache.py`). Its two tests were already `--ignore`'d in the kernels CI job; the ignore flags are dropped too. |
| `mla/v1/` | Production uses `mla/v2` only. The pure-JAX reference (`ref_mla_ragged_paged_attention` + kv-cache update) moves into `mla_v2_test.py`, which used it as the numerical baseline. `test_deepseek_v3.py` switches to `v2.get_kv_cache_shape` (drop-in superset of the v1 signature). |
| `quantized_matmul/` (entire directory) | The Pallas kernels are superseded by `gmm_v2` since #2706 — the `enable_quantized_matmul_kernel` path in `layers/common/linear.py` calls `gmm_v2`. The four live pure-JAX helpers from `util.py` move to `layers/common/quantized_matmul_util.py`. |
| `gdn/v1`, `gdn/v2`, `gdn/reference`, `triangle_solver.py` | Superseded by `gdn/v3` (#3016); `gdn_attention.py` imports v3 only. |
| `sparse_core/gather_reduce.py`, `ragged_scatter.py` | Production uses `dense_gather_reduce` + the `*_v2` kernels (follow-up to #3136). |
| `structured_sparse_matmul/` | Never wired in (#1938); its test was already `--ignore`'d. |
| `collectives/all_gather_matmul*`, `collectives/util.py` | Test-only; `hierarchical_reduce_scatter` (live) does not use them. The "collective kernels" CI step, its bootstrap.sh trigger, the feature YAML, and the step's `depends_on`/notify references are removed too. |
| **`MEGABLX_GMM` MoE backend + megablox `gmm` v1** (per review) | Reachable only via `USE_UNFUSED_MEGABLOCKS=1`, set nowhere in CI/nightly/benchmarks; sole consumer of gmm v1 and of the sparse-MoE permute/all-to-all helpers (`sparse_moe.py`, `utils.py` family). `gmm_v2` (the default GMM_TP/GMM_EP backend + quantized linear path) is self-contained and stays. |

Per review (@kyuyeunk):
- `causal_conv1d/` is **kept** (a non-GDN layer may need a standalone causal conv1d kernel).
- `quantized_matmul/` removed entirely (helpers relocated).
- `DENSE_MAT` is **kept** as the pure-JAX ground-truth backend that `test_fused_moe.py`/`test_moe.py` verify production backends against.

## Test plan
- Full `tests/kernels` suite passed on **both tpu6e and tpu7x** with the kernel removals (dev builds + CI build 22118 fully green).
- MoE backend removal: `tests/layers/jax/moe/` + `tests/layers/vllm/test_fused_moe.py`/`test_moe.py` pass on real tpu7x.
- `pytest --collect-only tests/` clean; repo-wide grep for every removed module/symbol: no remaining references.


kyuyeunk:
- suggest maintaining it just in case https://github.com/vllm-project/tpu-inference/tree/main/tpu_inference/kernels/causal_conv1d
  - there might be non-gdn layer that requires separate causal conv1d kernel
- remove entire directory? https://github.com/vllm-project/tpu-inference/tree/main/tpu_inference/kernels/quantized_matmul
- remove gmm v1 as well?
  - https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/common.py
  - https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/gmm.py
  - https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/kernels/megablox/tuned_block_sizes.py

QiliangCui2023:
Thanks for the review — addressed in 1f860b641:

1. **causal_conv1d: restored** (kernel + test), per your suggestion that a non-GDN layer may need a standalone causal conv1d kernel.

2. **quantized_matmul: entire directory now removed.** The only live content was four pure-JAX helpers in `util.py` (`quantize_tensor`, `quantize_block`, `get_max_min`, `xla_quantized_batched_matmul`) used by `layers/common/linear.py` and `models/jax/deepseek_v3.py` — they moved to `layers/common/quantized_matmul_util.py` next to their main consumer (they aren't Pallas kernels, so `layers/common` seems like the right home; happy to rename/move if you prefer another spot).

3. **gmm v1: left in place for now** — it's still reachable from production: `layers/jax/moe/utils.py:24` imports it, `gmm_fn` (utils.py:205) calls it, and that's invoked via `sparse_moe_func`, which `layers/common/moe.py:204` dispatches to; `deepseek_v3.py` and the JaxMoE path also import from the same utils module. It looks like it can only go once the JaxMoE→JaxRoutedExperts migration (#3041) lands and the sparse_moe path is retired. If you know that dispatch branch is already unreachable, I'm glad to include it here — otherwise I'd defer it to the #3041 cleanup.

QiliangCui2023:
Good questions — answered both, and implemented the removable half in 2dcdf9c82:

**Where gmm v2 is used:** it's the workhorse of the default MoE path and the quantized linear path:
- `layers/common/fused_moe_gmm.py` — backs the **GMM_TP** and **GMM_EP** backends, which `select_moe_backend()` returns by default (GMM_TP for non-EP, GMM_EP for EP unless `USE_MOE_EP_KERNEL=1` picks the fused kernel). This is what Qwen3-MoE / GPT-OSS / DeepSeek MoE layers run on in CI and nightly today.
- `layers/common/linear.py` — the blockwise-quantized matmul path calls `gmm_v2` directly (that's what let this PR delete the old Pallas quantized_matmul kernels).
So gmm v2 (and `gmm_test.py`) must stay.

**The DENSE_MAT / MEGABLX_GMM paths:**
- **MEGABLX_GMM: removed** in 2dcdf9c82. It was reachable only via `USE_UNFUSED_MEGABLOCKS=1`, which no CI/nightly/benchmark config sets, and it was the sole consumer of megablox **gmm v1** — so this also completes the gmm v1 removal you asked about earlier. Gone with it: `sparse_moe.py`, the sparse-MoE permute/all-to-all helpers in `layers/jax/moe/utils.py`, `kernels/megablox/{gmm,common,tuned_block_sizes}.py`, the env var, the benchmark-script flag, and its test cases. `gmm_v2.py` is self-contained and unaffected.
- **DENSE_MAT: kept** — it's the pure-JAX ground-truth backend that `test_fused_moe.py` and `test_moe.py` verify the production backends against (FUSED_MOE and GMM outputs are asserted `allclose` to it). Removing it would leave those tests without a reference implementation. It's env-gated (`USE_DENSE_MOE`, logged as "naive, not intended for production") and costs ~150 lines. If you'd rather drop even that and switch the tests to a hand-rolled einsum reference, happy to do it — but keeping one unfused reference backend seemed like the better trade.

Validated: all of `tests/layers/jax/moe/` and `tests/layers/vllm/test_fused_moe.py`/`test_moe.py` pass on a real tpu7x with the removal; repo-wide grep for every removed symbol is clean. Branch also rebased onto today's main. Net PR size is now ~20k lines removed.

kyuyeunk (approved):

Issue #1604, opened 2026-02-03
[Bug]: test_qwen3.py:test_model_loading failure from transformer pylib


lk-chen:
### Your current environment

```
Collecting environment information...
==============================
        System Info
==============================
OS                           : Ubuntu 24.04.3 LTS (x86_64)
GCC version                  : (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Clang version                : Could not collect
CMake version                : version 4.2.1
Libc version                 : glibc-2.39

==============================
       PyTorch Info
==============================
PyTorch version              : 2.9.0+cu128
Is debug build               : False
CUDA used to build PyTorch   : 12.8
ROCM used to build PyTorch   : N/A

==============================
      Python Environment
==============================
Python version               : 3.12.3 (main, Jan  8 2026, 11:30:50) [GCC 13.3.0] (64-bit runtime)
Python platform              : Linux-6.14.0-1014-gcp-x86_64-with-glibc2.39

==============================
       CUDA / GPU Info
==============================
Is CUDA available            : False
CUDA runtime version         : No CUDA
CUDA_MODULE_LOADING set to   : N/A
GPU models and configuration : No CUDA
Nvidia driver version        : No CUDA
cuDNN version                : No CUDA
HIP runtime version          : N/A
MIOpen runtime version       : N/A
Is XNNPACK available         : True

==============================
          CPU Info
==============================
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           52 bits physical, 57 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  224
On-line CPU(s) list:                     0-223
Vendor ID:                               GenuineIntel
Model name:                              INTEL(R) XEON(R) PLATINUM 8581C
CPU family:                              6
Model:                                   207
Thread(s) per core:                      2
Core(s) per socket:                      56
Socket(s):                               2
Stepping:                                2
BogoMIPS:                                1999.99
[...]

PR #3148, opened 2026-07-10
Fix silent no-op qwix axis in test_qwen3 and trim duplicate runs

QiliangCui2023:
## Summary

The qwix parametrization in `test_qwen3_600M` wrote `additional_config["quanntization"]` (typo, double-n), while `apply_qwix_quantization` reads `additional_config.get("quantization")` (`qwix_utils.py`). Every qwix-parametrized run therefore silently skipped quantization and was an exact duplicate of the corresponding non-qwix run — **8 of the 16 runs of the most expensive test in the part2 CI job tested nothing**.

Changes:
- **Fix the config key** so qwix quantization actually applies.
- Fixing it unmasked two mock gaps that the typo had been hiding since the test was written (the qwix path had never executed here): `MockVllmConfig` lacked `sharding_config.total_dp_size` and a real `cache_config.block_size` (a bare `MagicMock` coerces to `int() == 1`, which RPA v3 rejects as `page_size=1`). Both are now provided.
- **Trim the qwix axis to the single-stage PP config** (quantization is orthogonal to pipeline parallelism): 16 runs → 10, cutting several minutes of duplicate real-checkpoint loads while adding the qwix coverage the test always intended.
- Remove the HF-comparison block that was permanently disabled behind `if True: return` (TODO #1604 kept as a comment) and its now-unused `torch`/`AutoModelForCausalLM` imports.

## Test plan
- Full `tests/models/jax/test_qwen3.py` passed on tpu7x (dev build, prebuilt image from this branch) — including the qwix runs executing real fp8 quantization for the first time.

### DIFF
--- a/tests/models/jax/test_qwen3.py
+++ b/tests/models/jax/test_qwen3.py
@@ -19,7 +19,8 @@ import numpy as np
 import pytest
 from flax.typing import PRNGKey
 from jax.sharding import Mesh
-from vllm.config import ModelConfig, set_current_vllm_config
+from qwix._src.core.qarray import QArray
+from vllm.config import set_current_vllm_config
 from vllm.model_executor.model_loader import LoadConfig, get_model_loader
 
 from tpu_inference.distributed.jax_parallel_state import \
@@ -32,25 +33,6 @@ from tpu_inference.models.jax.qwen3 import Qwen3ForCausalLM
 from tpu_inference.models.jax.utils.qwix.qwix_utils import \
     apply_qwix_quantization
 from tpu_inference.runner.kv_cache import create_kv_caches
-
-
-class MockVllmConfig:
-
-    def __init__(self, model: str, kv_cache_dtype: str):
-        self.model_config = ModelConfig(model=model)
-        self.model_config.dtype = jnp.bfloat16
-        self.load_config = MagicMock()
-        self.load_config.download_dir = None
-        # block_size must be a real int: apply_qwix_quantization builds
-        # kv caches from cache_config.block_size for its dummy forward pass.
-        self.cache_config = MagicMock(cache_dtype=kv_cache_dtype,
-                                      block_size=32)
-        self.quant_config = None
-        self.additional_config = {}
-        self.parallel_config = None
-        # apply_qwix_quantization reads sharding_config.total_dp_size when
-        # building the dummy inputs for the quantization forward pass.
-        self.sharding_config = MagicMock(total_dp_size=1)
 
 
 @pytest.fixture(scope="module")
@@ -133,7 +115,8 @@ class TestQwen3ForCausalLM:
         (QWIX_FP8_RULES, 0, 1),
     ])
     def test_qwen3_600M(self, model_name, kv_cache_type, qwix_rules, rng, mesh,
-                        mock_model_inputs, pp_rank, pp_world_size):
+                        mock_model_inputs, pp_rank, pp_world_size,
+                        mock_vllm_config):
         """Tests model init and model forward for the 0.6B model variant."""
         init_pp_distributed_environment(
             ip="",
@@ -142,7 +125,7 @@ class TestQwen3ForCausalLM:
             device=jax.devices()[0],
             need_pp=False,
         )
-        mock_vllm_config = MockVllmConfig(model_name, kv_cache_type)
+        mock_vllm_config = mock_vllm_config(model_name, kv_cache_type)
         if qwix_rules:
             mock_vllm_config.additional_config["quantization"] = dict(
                 qwix=dict(rules=qwix_rules))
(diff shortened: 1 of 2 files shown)


tests/models/jax/test_qwen3.py
```diff
@@ -134,7 +144,7 @@ def test_qwen3_600M(self, model_name, kv_cache_type, qwix_rules, rng, mesh,
         )
         mock_vllm_config = MockVllmConfig(model_name, kv_cache_type)
         if qwix_rules:
-            mock_vllm_config.additional_config["quanntization"] = dict(
+            mock_vllm_config.additional_config["quantization"] = dict(
```
QiliangCui2023, line 130: Now that the qwix axis actually applies, consider adding a post-condition after the `apply_qwix_quantization` call so this can't silently regress to a no-op again (the failure mode this PR fixes): `apply_qwix_quantization` returns the model unchanged, without logging, when the `quantization`/`qwix`/`rules` lookup misses, and every existing assert in this test passes identically for an unquantized model.

```python
from qwix._src.core.qarray import QArray
weight = model.model.layers[model.model.start_layer].mlp.gate_proj.weight.value
assert isinstance(weight, QArray) == (qwix_rules is not None)
```

tests/models/jax/test_qwen3.py
```diff
@@ -43,10 +41,16 @@ def __init__(self, model: str, kv_cache_dtype: str):
...
+        # block_size must be a real int: apply_qwix_quantization builds
+        # kv caches from cache_config.block_size for its dummy forward pass.
+        self.cache_config = MagicMock(cache_dtype=kv_cache_dtype,
+                                      block_size=32)
         self.quant_config = None
         self.additional_config = {}
         self.parallel_config = None
+        # apply_qwix_quantization reads sharding_config.total_dp_size when
+        # building the dummy inputs for the quantization forward pass.
+        self.sharding_config = MagicMock(total_dp_size=1)
```
QiliangCui2023: `conftest.py`'s `mock_vllm_config` fixture has the same `(model, kv_cache_dtype)` signature (this file already uses it in `test_expected_error_with_tight_threshold` and `test_model_loading`) and builds a real `ShardingConfigManager(ShardingStrategy())`, so `total_dp_size` comes for free and any future attribute read the mock doesn't cover fails loudly instead of silently auto-Mocking. Adding `block_size=32` to that fixture's `cache_config` (additive for all consumers) would let this test use the shared fixture and delete the local class.

Related: the forward pass below hardcodes a second `block_size=32` in the `create_kv_caches` call — reading `mock_vllm_config.cache_config.block_size` there would keep the two in sync.

QiliangCui2023 (commented):
A few notes from a deeper pass over the diff.

nit: the comment at L217 (`# 1 seq with 16 tokens`) predates this PR but contradicts the fixture (`num_tokens=8`; the next assert checks `(8, hidden_size)`) — worth fixing while in the file.

tests/models/jax/test_qwen3.py
```diff
@@ -134,7 +144,7 @@ def test_qwen3_600M(self, model_name, kv_cache_type, qwix_rules, rng, mesh,
         )
         mock_vllm_config = MockVllmConfig(model_name, kv_cache_type)
         if qwix_rules:
-            mock_vllm_config.additional_config["quanntization"] = dict(
+            mock_vllm_config.additional_config["quantization"] = dict(
                 qwix=dict(rules=qwix_rules))
```
QiliangCui2023, line 131: Follow-up beyond this PR: this nested dict is hand-spelled here and `"quantization"` appears at 7 more sites in `qwix_utils.py`; a missing or misspelled key is treated as "quantization disabled" with no warning (and `{"quantization": {...}}` without a `"qwix"` key raises `AttributeError` on `.get("qwix").get("rules")`). Exporting a key constant or a small `make_qwix_config(rules)` helper from `qwix_utils`, and/or warning on unrecognized `additional_config` keys, would make the next near-miss loud instead of silent.

tests/models/jax/test_qwen3.py
```diff
@@ -328,32 +338,8 @@ def test_model_loading(self, model_name, pp_rank, pp_world_size,
...
-            hf_model.config.num_hidden_layers = 1
-            hf_output = hf_model(
-                inputs_embeds=input_tensor_hf, ).float().numpy()
-            np.testing.assert_allclose(
-                jax_output,
-                hf_output,
-                rtol=1e-2,
-                atol=1e-2,
-            )
+        # TODO(#1604): Compare jax_output against the HF reference model once
```
QiliangCui2023, line 336: The removal leaves a few danglers in this function: the comments at L297 (`# Create random input for comparison`) and L320 (`# Forward pass only the 1st layer for comparison`) refer to the deleted HF comparison, and with it gone `assert jax_output is not None` is near-vacuous — the layer call already returned. Consider rewording the comments and asserting something real, e.g. `jax_output.shape == (seq_len, model_dim)` and `jnp.isfinite(jax_output).all()`.

QiliangCui2023:
Review comments addressed in 787814817 (validated by running the full `tests/models/jax/test_qwen3.py` on a real TPU — 28/28 pass):

1. **qwix no-op guard (L147): added.** One adjustment to the suggested snippet: qwix wraps the quantized param as `WithAux(array=QArray(...))`, not a bare `QArray`, so the assert unwraps `.array` first. With rules set the weight must be a `QArray`; without rules it must not be.

2. **Shared fixture (L53): done.** The local `MockVllmConfig` class is deleted; `test_qwen3_600M` and `test_vocab_padding_for_sharding` now use the conftest fixture (real `ShardingConfigManager`, so future uncovered attribute reads fail loudly). The fixture's `cache_config` gains a real `block_size=32`, and the `create_kv_caches` call reads `cache_config.block_size` instead of hardcoding a second 32.

3. **Config-helper follow-up (L148): agreed, deferring** — a `make_qwix_config(rules)` helper / key constant in `qwix_utils` plus a warning on unrecognized `additional_config` keys is a good hardening, but it touches 7+ production sites so it deserves its own PR.

4. **HF-removal danglers (L341): fixed** — and the suggested shape assert caught a real pre-existing bug: the test unpacked the layer output as `(output, _)`, but the layer returns `(new_kv_cache, hidden_states)`, so the old `is not None` assert was inspecting the KV cache. The unpacking is fixed and the test now asserts `shape == (seq_len, model_dim)` and finiteness of the actual hidden states.

5. **L217 token-count nit: fixed** (comment now says 8 tokens and points at the fixture).

PR #2840, opened 2026-06-08
[KV Offload] Enable KV offloading for attention data parallelism

amitkumar307d:
# Description

This PR enables KV Cache Offloading support when Attention Data Parallelism (`enable_dp_attention: true`) is enabled.

**Context & Problem:**
Previously, when KV offloading was used alongside Attention DP, the system relied on custom Pallas ops (`stack_kv_cache_cross_layers` and `update_kv_caches_one`) that did not support the `attn_dp` sharding layout. This led to errors during the save/load chunk operations. 

**Error:**
```bash
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433]     - Failed to pre-compile for 1 blocks: shard_map applied to the function '_start_chunked_copy_kernel' was given argument arrays with axis sizes that are not evenly divisible by the corresponding mesh axis sizes:
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] 
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] The mesh given has shape (1, 8, 1, 1, 1, 1) with corresponding axis names ('data', 'attn_dp', 'attn_dp_expert', 'expert', 'model', 'dcp').
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] 
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] the passed args[4] of shape float8_e4m3fn[1,256,4,640], where args[4] is bound to _start_chunked_copy_kernel's parameter 'src_array: jax.Array', corresponds to in_specs[4] of value P(('data', 'attn_dp', 'attn_dp_expert'), 'dcp'), which maps array axis 0 (of size 1) to mesh axes ('data', 'attn_dp', 'attn_dp_expert') (of total size 8), but 8 does not evenly divide 1
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] 
(EngineCore pid=533) WARNING 06-04 15:37:02 [tpu_offload_connector.py:1433] Array arguments' axis sizes must be evenly divisible by the mesh axis or axes indicated by the corresponding elements of the argument's in_specs entry. Consider checking that in_specs are correct, and if so consider changing the mesh axis sizes or else padding the input and adapting '_start_chunked_copy_kernel' appropriately.
```

# Solution
[...]

PR #3152, opened 2026-07-10
Implement KVCache Offloading multi-host

dannawang0221:
# Description

Implement KVCache Offloading multi-host. The aggregate function has been already implemented with https://github.com/vllm-project/tpu-inference/pull/2840
This PR is focusing on tracking multi-host completion counts for KV cache offloading. In a multi-host TPU environment under Tensor Parallelism (TP), KV cache chunks are partitioned across multiple TP workers. Previously, the offload driver scheduler would free staging buffers or mark CPU chunks ready upon receiving the first completion report, resulting in race conditions and data corruption as other TP workers were still processing their partitions.

To resolve this:

- Dynamically calculate the target number of TP workers in the last pipeline stage based on the Ray cluster node count and pipeline parallel size
- Introduce per-chunk completion counters (_save_chunk_completion_counts and _load_chunk_completion_counts) in the TPU offload scheduler.
- Ensure that staging buffers are only freed and operations are marked complete once all expected TP workers have successfully reported completion.
- ThreadPoolExecutor-based D2H transfers can run out-of-order across
different TPU hosts, leading to execution sequence mismatches and hangs during
multihost offloading. Move the transfer (`jax.device_put`) initiation to the main thread
- Support multi-host conditions in accuracy/unit tests
- Adding a LeaderWorkerSet deployment for multi-host verification

# Tests

examples/offload/gke/pod_tpu_commons_cpu_offload_verification_multihost.yaml
examples/offload/gke/pod_tpu_host_offload_multihost_unit_tests.yaml

# Checklist


tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -631,6 +650,48 @@ def __init__(self, vllm_config: "VllmConfig"):
...
+                 the number of expected completion stats per chunk).
+        """
+        num_nodes = 1
+        if ray.is_initialized():
+            device_str = current_platform.ray_device_key
+            num_nodes = len([
+                n for n in ray.nodes() if device_str in n.get("Resources", {})
+            ])
+        pp_size = self.vllm_config.parallel_config.pipeline_parallel_size
+        num_tp_workers = num_nodes // pp_size
```
juncgu-google, line 828: add a note that DP is not supported yet.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -127,6 +127,9 @@
     from vllm.v1.request import Request
     from vllm.forward_context import ForwardContext
 
+import ray
```
juncgu-google: can we put "import ray" into try-except? just in case ray is not installed.

tests/offload/tpu_offload_connector_scheduler_test.py
```diff
@@ -42,6 +43,11 @@ class Cache:
         def __init__(self, block_size):
             self.block_size = block_size
 
+    class Parallel:
+
+        def __init__(self, pipeline_parallel_size=1):
```
juncgu-google: can we add test case where PP > 1 if PP is supported. Otherwise, add a note in the tpu_offload_connector.py that PP is not supported yet.

tests/offload/tpu_offload_connector_worker_test.py
```diff
@@ -74,6 +75,11 @@ class KVTransferConfig:
         ip = "ip"
         port = 1234
 
+    class Parallel:
+
+        def __init__(self, pipeline_parallel_size=1):
```
juncgu-google: same here. can you add test with PP > 1?

juncgu-google (approved):
LGTM.

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Added the note and TODO to remove the note once PP is supported

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Done

tests/offload/tpu_offload_connector_scheduler_test.py
dannawang0221: Done

tests/offload/tpu_offload_connector_worker_test.py
dannawang0221: Done

dannawang0221:
@Lumosis @kyuyeunk can you help review and approve this PR? Thanks

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -631,6 +673,62 @@ def __init__(self, vllm_config: "VllmConfig"):
...
+        Terminology & Architecture Notes:
+        1. Ray Actor (Worker Actor): In vLLM, distributed workers are instantiated
+           as remote Ray Actors (processes running `RayWorkerWrapper`).
+        2. TPU Worker: In a multi-host TPU environment executing in SPMD mode, there is
+           exactly one worker actor process (Ray Actor) per host node. Thus, the total
+           number of worker actors in the cluster equals `num_nodes`.
+        3. TP (Tensor Parallel) Worker: Under Pipeline Parallelism (PP), the worker actors
+           are divided into `pp_size` stages. Within each stage, they run in a tensor-parallel
+           fashion. Therefore, each stage contains `num_nodes // pp_size` worker actors,
+           representing the TP size per stage.
```
saikat-royc: please double check the comment. "each stage contains `num_nodes // pp_size` worker actors, representing the TP size per stage." When we say TP size per stag, it should have a device count associated.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -631,6 +673,62 @@ def __init__(self, vllm_config: "VllmConfig"):
...
+
+        In multi-host TPU setups, the KV cache tensors are partitioned across these TP
```
saikat-royc: You can make the descriptive more illustrative with an example . See internal discussion thread

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -1979,8 +2105,21 @@ def _async_batch_transfer_task(*args, **kwargs):
         # Note: We use manifest for the pending future tracking.
         # record_save will be handled in the main thread by _process_completed_saves.
 
+        chunks_on_cpu = []
```
saikat-royc, line 2222: please put a comment explaining why the data movement from HBM to host RAM is moved to main thread.

Lumosis:
This PR is too large. Could you break it into several smaller PRs?

examples/offload/gke/benchmarks/single-host/deploy-cpu-offload.yaml
```diff
@@ -32,7 +32,11 @@ spec:
       #   gke-gcsfuse/cpu-limit: "2"
       #   gke-gcsfuse/ephemeral-storage-limit: "200Gi"
     spec:
+      # Added: terminationGracePeriodSeconds
+      # This gives the TPU and vLLM time to shut down gracefully.
+      terminationGracePeriodSeconds: 1200
       nodeSelector:
+        kubernetes.io/hostname: gke-tpu-8f4b85ba-jfps  # FORCES scheduling to this specific node
```
saikat-royc: we can remove this

saikat-royc:
lgtm with minor comments

dannawang0221:
> This PR is too large. Could you break it into several smaller PRs?

@Lumosis It's difficult to break this PR down into smaller commits without breaking functionality. Core feature changes are concentrated in tpu_inference/offload/tpu_offload_connector.py and tpu_inference/worker/tpu_worker.py, while the remaining files are just unit tests and benchmark setups.

tpu_inference/worker/tpu_worker.py
```diff
@@ -335,7 +335,8 @@ def init_device(self,
         self.parallel_config.data_parallel_size = 1
         try:
             with set_current_vllm_config(self.vllm_config):
-                temp_file = tempfile.mkstemp()[1]
+                fd, temp_file = tempfile.mkstemp()
+                os.close(fd)
```
Lumosis, line 432: Please add comments

tpu_inference/offload/tpu_offload_connector.py
Lumosis: If DP/PP is not supported, we should fail fast and raise exception

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Updated

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Updated

examples/offload/gke/benchmarks/single-host/deploy-cpu-offload.yaml
dannawang0221: Done

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Done

tpu_inference/worker/tpu_worker.py
dannawang0221: Done

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -455,8 +455,19 @@ def __init__(
     ):
         super().__init__(vllm_config, role, kv_cache_config)
         logger.info("TPUOffloadConnector: Entering __init__")
-        # TODO(dannawang): Pipeline Parallelism (PP) is not supported/tested for KV cache offloading yet.
-        # Move or remove this note once PP mode is fully supported and verified.
+        # Check DP/PP support
```
dannawang0221, line 451: @Lumosis Added fast fail check for PP and DP

Lumosis (approved):

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -123,9 +123,16 @@
...
 if TYPE_CHECKING:
+    from vllm.forward_context import ForwardContext
     from vllm.v1.core.kv_cache_manager import KVCacheBlocks
     from vllm.v1.request import Request
-    from vllm.forward_context import ForwardContext
+
+try:
+    import ray
+except ImportError:
+    ray = None
```
kyuyeunk: see https://en.wikipedia.org/wiki/Fail-fast_system

apply the same for system for all the `try/except` statements in this file.

dannawang0221: I understand the concept of fail-fast, but I'm not sure how you'd like it applied to the try/except blocks in this file. Are you suggesting that we raise error in all the except blocks? Could you provide an example or clarify what you'd like changed?

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -426,6 +455,20 @@ def __init__(
     ):
         super().__init__(vllm_config, role, kv_cache_config)
         logger.info("TPUOffloadConnector: Entering __init__")
+        # Check DP/PP support
+        parallel_config = vllm_config.parallel_config
+        pipeline_parallel_size = getattr(parallel_config,
```
dannawang0221: @kyuyeunk Could you please leave a comment here detailing the updates we discussed during the meeting?

kyuyeunk: `getattr(parallel_config, "pipeline_parallel_size", 1)` might silently fail if upstream decides to change the varilable name into something like `pipeline_parallel_sizes`. a better way is to
- programatically confirm if `parallel_config` is the same instance as `ParallelConfig` class: https://github.com/vllm-project/vllm/blob/6c7e679f048dc6123caecc3985150766e455ff22/vllm/config/parallel.py#L119
- maybe add something like `assert isinstance(parallel_config, ParallelConfig`
- and access each member variables normally like `pipeline_parallel_size = parallel_config.pipeline_parallel_size`

dannawang0221: updated

examples/offload/gke/pod_tpu_host_offload_multihost_unit_tests.yaml
```diff
@@ -0,0 +1,149 @@
...
+    protocol: TCP
+    targetPort: 8000
+  - name: ray-gcs
+    port: 6379
+    protocol: TCP
+    targetPort: 6379
+  selector:
+    leaderworkerset.sigs.k8s.io/name: tpu-job-host-offload-multi-host-unit-tests
+  sessionAffinity: None
+  type: ClusterIP
```
kyuyeunk: new line?

examples/offload/offline_inference_kv_cache_verification.py
```diff
@@ -137,7 +149,18 @@ def main(args: dict):
                                            num_invocations=1)
         baseline_output = baseline_outputs[0]
         print(f"Baseline Generated Text: {baseline_output!r}")
+        # Explicitly shut down the background EngineCore process and Ray actors
+        # to release TPU device allocations and placement groups immediately.
+        # Otherwise, they remain locked until garbage collection, which blocks
+        # the subsequent test runs from allocating TPU resources.
+        if hasattr(baseline_llm,
```
kyuyeunk: reason for using `hasattr`?

tests/offload/prompt_files/small_prompt.txt
kyuyeunk, line 1: nit: out of curiosity, what is the reason for this change?

examples/offload/offline_inference_kv_cache_verification.py
```diff
@@ -149,7 +172,12 @@ def main(args: dict):
                                        test_params,
                                        prompts=prompts,
                                        num_invocations=2)
+        if hasattr(test_llm, "llm_engine") and test_llm.llm_engine is not None:
+            if hasattr(test_llm.llm_engine, "engine_core"
+                       ) and test_llm.llm_engine.engine_core is not None:
+                test_llm.llm_engine.engine_core.shutdown()
         del test_llm
+        gc.collect()
```
kyuyeunk, line 175: in many cases the need for running `gc.collect()` means something was not working correctly and could indicate it's a band-aid fix - not a proper fix. but i can be completely wrong in this case, so i'm happy to hear counterargument.

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -319,28 +322,43 @@ class KVOffloadConnectorStats(KVConnectorStats):
...
         }
 
+    @property
+    def process_index(self) -> int:
+        if getattr(self, "_process_index", None) is None:
+            try:
+                # Under JAX's multi-process SPMD model, each host node runs its own Python process.
+                # Since Ray launches exactly one worker process (RayWorkerWrapper) per node,
+                # jax.process_index() acts as a unique ID mapping 1:1 to each Ray worker actor.
+                self._process_index = jax.process_index()
```
kyuyeunk: i don't think this is a good approach? i understand that this part
>  jax.process_index() acts as a unique ID mapping 1:1 to each Ray worker actor.

but why does it ned to be wrapped under exception handling?

tpu_inference/offload/tpu_offload_connector.py
```diff
@@ -631,6 +682,162 @@ def __init__(self, vllm_config: "VllmConfig"):
...
+        num_nodes = 1
+        try:
+            is_multihost = jax.process_count() > 1
+        except Exception:
+            is_multihost = False
+
+        if is_multihost:
+            try:
+                import ray
+            except ImportError as e:
```
kyuyeunk: what happens if you remove this?

kyuyeunk (commented):
looks much better than before!

tests/offload/prompt_files/small_prompt.txt
dannawang0221: The original prompt is too short to trigger CPU offloading. To ensure the prefix matches at least one block, the prompt length must exceed the HBM block size (256 tokens for Llama-3.1-8B).

examples/offload/offline_inference_kv_cache_verification.py
dannawang0221: Adding gc.collect() is to force cleanup between sequential tests, preventing slow vllm engine shutdowns from crash the next test.

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Update with checking TPU_MULTIHOST_BACKEND value

tpu_inference/offload/tpu_offload_connector.py
dannawang0221: Removed try/except

examples/offload/offline_inference_kv_cache_verification.py
dannawang0221: That check was originally added for version compatibility. Since KV cache offloading is only supported in vLLM v1, I remove the hasattr check.

examples/offload/gke/pod_tpu_host_offload_multihost_unit_tests.yaml
dannawang0221: Updated

kyuyeunk (approved):

PR #3157, opened 2026-07-11
Fix Llama-Guard-4 JAX path: five breakages exposed by re-enabled nightly tests (#3142)

QiliangCui2023:
## What
Fixes the Llama-Guard-4-12B engine-init crash on `releases/v0.25.0`:

```
AttributeError: 'Llama4VisionConfig' object has no attribute 'rope_theta'
  tpu_inference/models/jax/llama_guard_4.py:405
```

Seen in the v0.25.0 release nightly ([build 22159](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22159), `tpu7x Accuracy for meta-llama/Llama-Guard-4-12B` text-only + multimodal).

## Why
transformers >= 5.6 moved `rope_theta` from a flat config attribute into the `rope_parameters` dict. The vLLM requirement is a floating `transformers >= 5.5.3`, and the image built for the LKG bump to `e5588e49` picked up the new transformers, so `vision_config.rope_theta` no longer exists. `main` will hit the same failure on its next nightly (its last nightly, build 22076, was built with the older LKG `f1a5addd` image).

## Fix
Use the existing `get_rope_theta()` helper (`tpu_inference/layers/jax/rope_interface.py`) for both the text and vision configs — it reads `rope_parameters` first and falls back to the legacy attribute, so both transformers layouts work.

## Test
- `tests/models/jax/test_llama_guard_4.py` mocks updated to the new config layout (`rope_parameters` dict, no `rope_theta` attribute) — the previous code fails these tests with the exact CI AttributeError; the fixed code passes.
- 10/10 pass locally on a TPU host with transformers 5.12.1.
- Dev-pipeline validation of the nightly safety-model accuracy job on this branch: running, results will be posted here.

Note: the same patch is applicable to `main` (identical code there); this PR targets the release branch per the release-blocker process.

### DIFF
--- a/tests/models/jax/test_llama_guard_4.py
+++ b/tests/models/jax/test_llama_guard_4.py
@@ -86,12 +86,26 @@ class MockVllmConfig:
         text_config_mock.num_attention_heads = 4
         text_config_mock.num_key_value_heads = 2
         text_config_mock.head_dim = 32
-        # transformers >= 5.6 layout: rope_theta lives in rope_parameters and
-        # the flat rope_theta attribute no longer exists on the config.
-        text_config_mock.rope_parameters = {
+        # Explicit numeric values: MagicMock auto-attrs are Mock objects and
+        # blow up only at forward time inside jitted math.
+        text_config_mock.num_layers = 2
+        text_config_mock.rms_norm_eps = 1e-5
+        text_config_mock.intermediate_size = 256
+        text_config_mock.hidden_act = "silu"
+        # transformers >= 5.6 layout: rope_theta lives in rope_parameters,
+        # the flat rope_theta attribute no longer exists, and rope_scaling
+        # is a property alias of rope_parameters. Llama-Guard-4 ships
+        # llama3-style scaling keys, so mirror that faithfully.
+        llama3_rope_parameters = {
+            "rope_type": "llama3",
             "rope_theta": 500000.0,
-            "rope_type": "default"
+            "factor": 8.0,
+            "low_freq_factor": 1.0,
+            "high_freq_factor": 4.0,
+            "original_max_position_embeddings": 8192,
         }
+        text_config_mock.rope_parameters = llama3_rope_parameters
+        text_config_mock.rope_scaling = llama3_rope_parameters
         del text_config_mock.rope_theta
 
         hf_config_mock = MagicMock()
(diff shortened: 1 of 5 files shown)


QiliangCui2023:
Validation complete: [tpu-inference-dev build 667](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/667) ran both nightly accuracy jobs (`safety_model_benchmark.sh --mode accuracy`, text-only + multimodal, tpu7x TP=2) on this branch — **both PASSED**.

This ended up being **five stacked breakages** on a path that was dormant while the Llama-Guard-4 nightly tests were disabled, and got re-exposed when #3142 (Jul 10) re-enabled the Llama-4 matrix tests:

1. `Llama4VisionConfig.rope_theta` removed by transformers >= 5.6 (rope_parameters layout) → engine-init AttributeError. Fixed via the existing `get_rope_theta()` helper.
2. 2-value unpack of `TransformerBlock`, which returns 3 values since #2343 (May) → `ValueError: too many values to unpack` at first warmup forward.
3. Model forward must return a 4-tuple (kv, hidden, aux_hidden, expert_ids) per `run_model`'s `out_shardings` → pjit pytree-structure mismatch. Mirrors llama4.py.
4. torch **bfloat16** `pixel_values` cannot pass through numpy (`TypeError: Got unsupported ScalarType BFloat16`) → int16 bit-view conversion, same as gemma4_mm.
5. `Llama4VisionRotaryEmbedding.__call__` returned the `nnx.Param` object; newer JAX removed implicit `__jax_array__` conversion → return `.value`. (This one lives in shared `rope.py`, so it also pre-fixes plain Llama-4 vision models.)

Each layer was found by iterating dev validation on the release base (builds 654 → 659 → 664/665 → 667). Unit tests updated so the transformers-5.6 config layout is what the mocks exercise (old code fails them).

All five fixes apply to `main` identically — main's next nightly will hit layer 1 (it now builds with the new-transformers image after the LKG bump to `e5588e49`).

Do not merge until the release owner approves.

QiliangCui2023:
Retargeted to `main` per the release owner: fix lands on `main` first, then gets cherry-picked to `releases/v0.25.0`. Validation already done on the release base (see above); the code is identical on both branches for these files. Once this merges, I will open the cherry-pick PR to `releases/v0.25.0`.

lk-chen (approved):

PR #3162, opened 2026-07-13
[Jax] Rewrite GPT-OSS onto JaxRoutedExperts, migrate MXFP4 loading off Qwix (2/2)

### LINKED ISSUE #2952: [Feature]: Support native JAX GPT-OSS MXFP4 quantization loading

Leo7Deng:
# Description

Part 2 of 2 for #2952 (part 1: #3086 added `Mxfp4FusedMoEMethod` / `Mxfp4Config` under `layers/jax/quantization`). This rewrites GPT-OSS onto `JaxRoutedExperts` so its experts load MXFP4 natively through the quant method lifecycle, and removes the Qwix QArray loading path.

### Key changes

1. `GptOssMoE` is now a `JaxRoutedExperts` subclass (`layers/jax/moe/gpt_oss_moe.py`), same as what #3052 did for `Gemma4MoE`. The router (with bias) stays in the model and hands raw logits to the experts, so expert compute runs on the shared `moe_apply` GMM path with the same `softmax` + `renormalize` routing that `Gemma4MoE` uses. I deleted the old expert MLP implementation.
2. GPT-OSS weight loading moves off Qwix (`models/jax/gpt_oss.py`). The six expert tensors per layer go through the quant method contract, and the Qwix pool loader and default config are deleted. Non-expert weights load the same way as before.  GPT-OSS keeps its own loader rather than JaxAutoWeightsLoader, so it calls process_weights_after_loading itself.

Details:
- Official GPT-OSS checkpoints are MXFP4-only, and the old loader's BF16 expert path was never used or tested, so loading non-MXFP4 expert weights now raises a ValueError.
- The GMM kernel hardcodes the swigluoai clamp at 7.0 (which all official GPT-OSS checkpoints use), so a config with a different `swiglu_limit` now raises NotImplementedError at init rather than silently clipping at the wrong value.

FIXES: #2952

# Tests

CPU unit tests against the repo's pinned vLLM version:

```
pytest --continue-on-collection-errors tests/layers/jax/quantization/ tests/layers/jax/moe/ tests/layers/jax/test_qwix.py tests/models/jax/test_gpt_oss.py
```

92 passed, 4 skipped. (`test_fp8.py` fails collection on this macOS CPU host because `jaxlib._jax` lacks `TransferConnection`, unrelated to this change.)

- New tests in `test_gpt_oss_moe.py` cover the router contract for the fused and dense backends, the softmax/top-k routing equivalence, and the `GptOssMoE` wiring.
- `pre-commit run` passed on the changed files.

# Checklist

- [x] I have made or will make corresponding changes to any relevant documentation. (Internal quantization dispatch only; no user-facing docs affected.)


lk-chen (approved):

Leo7Deng:
Rebased on main, tests still pass. lk-chen approved a while back. @kyuyeunk @jrplatin could you take a look when you get a chance?

PR #3163, opened 2026-07-13
[Qwen3-VL] End-to-end M-RoPE + spec-decode + multimodal + DP-attention serving

qizzzh:
End-to-end fix chain that lets Qwen3-VL + eagle3 (or MTP) + multimodal
serve on TPU under DP-attention: unblocks startup against the current
vLLM LKG, fixes a warmup HBM OOM caused by a fully-replicated dummy in
spec-decode precompile, unblocks the first serving request (text and
multimodal) under M-RoPE + spec-decode, eliminates the classes of
runtime JIT that would otherwise re-fragment HBM mid-serve, and
restores the opt-in encoder-cache offload to `pinned_host` to reduce
serving-time fragmentation from variable-shape encoder outputs.

## 1. vLLM API-drift compat (kv_cache, awq, mm_encoder)

Restores the tpu-inference torchax path against the current vLLM LKG
so tpu-inference can start at all against upstream. Two rename
follow-ups and one signature refactor:

- `runner/kv_cache_manager.py`: rename `DeepseekV4Attention` →
  `DeepseekV4MLAAttention` to match vLLM after the flat-name revert;
  make `KVCacheTensor.block_stride` access via `getattr(..., None)`
  so the DeepseekV4 packed-layout branch is skipped (rather than
  crashing) on builds without that PR.
- `layers/vllm/quantization/awq.py`:
  `vllm.model_executor.layers.quantization.auto_awq` is gone in
  current vLLM; import `AWQConfig` / `AWQLinearMethod` from `.awq`
  directly (drop the `AutoAWQConfig as AWQConfig` alias).
- `runner/mm_encoder_jit_manager.py`: adopt vLLM PR #41234 (simplify
  ViT CUDA graph interfaces). `EncoderCudaGraphCaptureInputs` no
  longer has `.values`; it exposes `mm_kwargs` (pixel_values +
  grid_thw) and `buffers` (pos_embeds / rotary / cu_seqlens /
  max_seqlen / sequence_lengths) separately.
  `model.encoder_cudagraph_forward(mm_kwargs, buffers)` is now the
  two-arg signature. Templates are stored as two dicts
  (`mm_kwargs_templates` / `buffers_templates`) so the JIT cache
  keys correctly on both shape signatures.
- `tests/runner/test_mm_encoder_jit_manager.py`: update to match the
  refactored API — `_pad_dict_to_template(replay_values, template)`,
  the new attribute names, and the new
  `EncoderCudaGraphCaptureInputs(mm_kwargs=..., buffers=...)` /
  `EncoderCudaGraphReplayBuffers(buffers=...)` constructors.

## 2. Qwen3-VL Moe `__init__` workarounds

`Qwen3VLMoeForConditionalGeneration.__init__` calls
`super(Qwen3VLForConditionalGeneration, self).__init__()` which
SKIPS `Qwen3VLForConditionalGeneration.__init__` (where
`self.model_config = vllm_config.model_config` would be set). The
resulting `AttributeError` kills the engine during
`MMEncoderJITManager` init when vLLM calls
`get_encoder_cudagraph_config` (which reaches
`get_max_frames_per_video` → `self.model_config`). Patch three
methods in `models/vllm/experimental/qwen3_vl_patcher.py` to route
through `get_current_vllm_config()` when `self.model_config` is
missing: `get_encoder_cudagraph_config` (also filters `video` from
the modalities list when the user set `limit_per_prompt.video == 0`),
`get_max_frames_per_video`, and `get_encoder_cudagraph_budget_range`.

## 3. `DISABLE_QWEN3_VL_DEEPSTACK` env-var escape hatch

The Qwen3-VL deepstack add
(`hidden_states + deepstack_input_embeds[key]` at
`qwen3_vl_moe.py:120` / `qwen3_vl.py:1547`) trips
`TypeError: unsupported operand type(s) for +: 'View' and 'Tensor'`
during the "backbone with embeds" precompile under the torchax +
torch.compile dispatch. Attempted alternative materializations
(`.contiguous()`, `.clone()`, compilation-config mode NONE) all hit
the same failure — root cause is in the torchax / torch.compile
dispatch layer, not in the patcher.

Wire an opt-in bypass via `DISABLE_QWEN3_VL_DEEPSTACK=1` in three
places: `_patched_get_deepstack` (returns None so the LM skips the
add), `_patched_forward` (clears deepstack state pre-forward —
necessary because torch.compile/dynamo may inline the ORIGINAL bound
method rather than our monkey-patched lambda), and
`_patched_embed_input_ids` (short-circuits packing so `inputs_embeds`
keeps its `visual_dim` width). Vision quality degrades (deepstack
residual dropped) but the server comes up.

## 4. HBM: batch-shard dummy logits in `_precompile_process_and_extend_logits`

Fixes the DP-attention warmup OOM at ~92.95 GB/chip on Qwen3-VL-30B-A3B
+ eagle3 + `enable_dp_attention` (config: TP=2, attn_dp=4,
`max_num_seqs=128` → `max_num_reqs=512`).

`runner/compilation_manager.py`: allocate the dummy `target_logits`
and `processed_bonus_logits` in
`_precompile_process_and_extend_logits` with
`PartitionSpec(ShardingAxisName.ATTN_DATA)` instead of the empty
`PartitionSpec()` (fully replicated). This matches production —
the real `process_and_extend_logits` at
`spec_decode/jax/utils.py:304` wraps a `shard_map` with
`in_specs = PartitionSpec(ShardingAxisName.ATTN_DATA)` for both
target and bonus logits. The old dummy allocation replicated the
largest tensor in the runner (float32 × num_logits × vocab, up to
~2.5 GiB per shape at num_logits=4096 / vocab≈152064) on every
chip, so under DP=4 the per-chip footprint was 4× larger than
production, and the final dummy iteration in this precompile loop
was the direct cause of the eagle3 warmup OOM
("Attempting to allocate 2.32G. That was not possible. There are
1.80G free."). With the dummy sharded to match production, per-chip
peak drops from 92.95 GB (OOM) to 89.15 GB (server reaches
"Available routes" and `/health` = 200).

## 5. Eagle3 `_select_inputs_for_loop_speculation` handles 2-D M-RoPE positions

M-RoPE targets (Qwen3-VL etc.) feed 2-D positions of shape
`(3, num_tokens)` with `PartitionSpec(None, ATTN_DATA)`; non-M-RoPE
targets feed 1-D `(num_tokens,)` on `ATTN_DATA`.
`Eagle3Proposer._select_inputs_for_loop_speculation._select`
(spec_decode/jax/eagle3.py:548) unconditionally did
`positions[last_token_indices]`. For 2-D positions this indexes axis 0
with a length-B index array and produces `(B, N)` per shard instead of
`(3, B)`. The mis-shaped positions then fed into
`_update_inputs_for_loop_speculation`, where `jnp.any(axis=0)` reduced
the leading axis, and downstream
`jnp.where(exceeds_max_model_len_reduced, 1, new_seq_lens)` crashed
with:

    ValueError: Incompatible shapes for broadcasting:
        shapes=[(16,), (), (128,)]

where 16 = padded_total_num_scheduled_tokens/dp_size and
128 = max_num_reqs/dp_size at DP=4.

Fix: drop the method-based split; always index the token axis for 2-D
positions, mirroring the existing 2-D handling in
`_filter_token_and_prepare_initial_inputs`.

## 6. `_precompile_eagle3_helpers` primes 2-D M-RoPE positions

`compilation_manager._precompile_eagle3_helpers` unconditionally built
1-D positions `(num_tokens,)`. Runtime feeds 2-D for M-RoPE, so the
`@jax.jit`-decorated `_prepare_inputs` / `_propose` missed the cache
on the first serving request and re-lowered inside `ForbidCompile`
(about 25 s). Match runtime shape when `self.runner.uses_mrope`. The
MTP precompile helper already had the 2-D branch; only Eagle3 was
missing it.

## 7. Preserve pre-embed input_ids for multimodal + spec-decode

`TPUModelRunner._get_input_ids_embeds` returns `(None, inputs_embeds)`
on the multimodal path (the model consumes merged image+text
embeddings, not raw token ids). That `None` was stored into
`ExecuteModelState.input_ids` and reached `_sample_from_logits`, which
requires token ids for `_extract_draft_token_ids` when
`spec_decode_metadata is not None`. First multimodal request with
spec-decode enabled crashed with:

    AssertionError: assert input_ids is not None

Multimodal + spec-decode has no test coverage and appears to have
never worked end-to-end. Fix: keep the pre-embed `input_ids` under a
separate name when spec-decode is active and hand that to
`ExecuteModelState`. For text-only requests the drafter still sees
ordinary token ids; for multimodal requests it sees image placeholder
ids for the image span (the encoder replaces those with visual
features), and ordinary text ids for the generated tail — which is
exactly what the rejection sampler compares against.

## 8. Restore `OFFLOAD_ENCODER_CACHE_TO_HOST=1`

vLLM PR #2802 re-implemented `MMEncoderJITManager` and inadvertently
removed the earlier `OFFLOAD_ENCODER_CACHE_TO_HOST=1` opt-in. Under
multimodal workloads at concurrency, `runner.encoder_cache: dict[str,
jax.Array]` accumulates variable-shape entries (one per encoded image,
sized by per-image patch count). Combined with the compile-cache
persistent buffers and KV-pool layout, this contributes to HBM
external fragmentation — the "free > needed > largest_contiguous;
defrag-fails" pattern that libtpu's BFC can't recover from because it
cannot relocate live tensors.

Re-added `_offload_encoder_cache_enabled` +
`_encoder_cache_to_pinned_host` / `_encoder_cache_to_device` helpers
in `multimodal_manager.py`, wired into `execute_mm_encoder` (producer)
and `gather_mm_embeddings` (consumer). Sharding (mesh + PartitionSpec)
is preserved; only `memory_kind` changes. A lazy
`_maybe_unwrap_torchax` helper handles both plain `jax.Array` and
torchax-wrapped tensors without a hard torchax import at module scope.

Off by default; opt in with `OFFLOAD_ENCODER_CACHE_TO_HOST=1`.
Observational verification via `logger.info` sentinel on first encoded
image; A/B on MMMU-Pro conc=64 shows +3 GB `largest_free_block_bytes`
with offload on at the same peak HBM.

## 9. `_precompile_continue_decode`: prime 2-D M-RoPE positions

`_precompile_continue_decode` (compilation_manager.py:1690, 1702)
passed `init_tokens` (shape `(num_reqs,)`) as
`AttentionMetadata.input_positions` for both single-KV-group and
multi-KV-group branches, with no `uses_mrope` gate. Runtime
`_execute_continue_decode` threads 2-D M-RoPE positions of shape
`(3, num_tokens)` into the jitted `_decode_core` and
`_update_loop_state` (decode_loop.py). First M-RoPE decode-only
request re-lowered under `maybe_forbid_compile` and raised.

Fix: gate on `uses_mrope`, build a `(3, num_reqs)` positions tensor
with `PartitionSpec(None, ATTN_DATA)`, and thread it through both
`AttentionMetadata` constructions.

## 10. `_precompile_backbone_with_inputs_embeds`: fix `num_reqs` coverage

`_precompile_backbone_with_inputs_embeds` had an indentation bug: the
`if uses_mrope` block and the `_precompile_backbone_helper(...)` call
were one dedent short, sitting inside `for num_tokens` but OUTSIDE
`for num_reqs`. Because `padded_num_reqs` is a `meta_field` on
`AttentionMetadata`, every distinct `num_reqs` bucket is its own jit
cache key — the shallow indentation silently primed only the LAST
`num_reqs` bucket per `(h_size, num_tokens)`, so every other bucket
cache-missed at runtime. Affects all models (M-RoPE and non-M-RoPE)
that hit the multimodal-embed backbone.

Fix: re-indent the block one level so it runs inside `for num_reqs`.

## 11. `_precompile_extend_logits_simple`: match runtime `PartitionSpec`

`_precompile_extend_logits_simple` special-cased `attn_data_size == 1`
to `PartitionSpec()`, but runtime `extend_logits_simple`
(spec_decode/jax/utils.py:260) always uses
`PartitionSpec(ATTN_DATA, None)` for its `shard_map` in/out specs. On
single-DP meshes this produced a different `NamedSharding` cache key
and cache-missed on the first spec-decode logprobs call — which raises
under `VLLM_XLA_CHECK_RECOMPILATION=1` since the caller is inside
`maybe_forbid_compile`.

Fix: drop the special case and always prime with
`PartitionSpec(ATTN_DATA, None)`.

## 12. Regression test for the M-RoPE broadcast bug

`tests/spec_decode/test_eagle3.py::test_select_inputs_for_loop_speculation_mrope`
exercises `_select_inputs_for_loop_speculation` with 2-D positions for
both `method="eagle3"` and `method="mtp"` and asserts the output stays
`(3, num_reqs)`. Pre-fix the eagle3 parametrisation produced
`(num_reqs, total_tokens)` and the assertion would have caught the bug
at unit-test time. Note
`SpeculativeConfig(model=<eagle3_dir>, method="mtp")` auto-normalises
`.method` to `"eagle3"` internally, so the test overrides
`proposer.method` after construction to actually exercise both
branches genuinely.

## Verification

End-to-end on Qwen/Qwen3-VL-30B-A3B-Instruct + eagle3, DP=4/TP=2,
max_num_seqs=128, max_num_batched_tokens=8192, kv_cache_dtype=fp8,
`enable_dp_attention=true`, `cudagraph_mm_encoder=true`,
`num_speculative_tokens=4`:

1. Warmup completes (peak HBM 89.15 GB/chip; pre-fix warmup OOM at
   92.95 GB/chip).
2. `/health` → 200; first `/v1/chat/completions` → 200 with real
   completion.
3. First multimodal request (image + text) → 200 with correct
   caption.
4. `VLLM_XLA_CHECK_RECOMPILATION=1` end-to-end through single +
   concurrency=8 text + multimodal requests: zero post-startup
   recompile detections. Pre-fix the first request always triggered a
   ~25 s runtime compile with signature
   `jit(_prepare_inputs) in_avals=[(3, 64):int32, ...]`.
5. MMMU-Pro at concurrency=64, 512 prompts, output_len=128: 512/512
   success; 0.89 req/s, 466 tok/s total, mean E2EL 70 s, peak HBM
   ~90 GB/chip; 0 serving-time fragmentation OOMs.

## Not done in this commit

1. Does NOT solve the torchax/dynamo `View + Tensor` error at
   `qwen3_vl_moe.py:120` under `@support_torch_compile`. Deepstack
   remains dropped by the `DISABLE_QWEN3_VL_DEEPSTACK=1` escape hatch
   when the operator sets it; vision-quality trade-off is unchanged.
2. No unit-test coverage for the new precompile shape branches (6,
   9, 10, 11), the multimodal + spec-decode input_ids preservation
   (7), or the offload helpers (8) — those paths need a full runner
   spin-up to exercise which is heavier than the fixes warrant.

### DIFF
--- a/tests/runner/test_mm_encoder_jit_manager.py
+++ b/tests/runner/test_mm_encoder_jit_manager.py
@@ -69,7 +69,8 @@ class _DummyModel:
     def get_encoder_cudagraph_config(self):
         return EncoderCudaGraphConfig(
             modalities=["image"],
-            buffer_keys=["pixel_values"],
+            input_key_by_modality={"image": "pixel_values"},
+            buffer_keys=[],
             out_hidden_size=64,
         )
 
@@ -83,8 +84,11 @@ class _DummyModel:
                                                  max_batch_size,
                                                  max_frames_per_batch, device,
                                                  dtype):
-        return EncoderCudaGraphCaptureInputs(
-            values={"pixel_values": torch.zeros(token_budget, 4, dtype=dtype)})
+        return EncoderCudaGraphCaptureInputs(mm_kwargs={
+            "pixel_values":
+            torch.zeros(token_budget, 4, dtype=dtype)
+        },
+                                             buffers={})
 
     def get_encoder_cudagraph_item_specs(self, mm_kwargs):
         return []
@@ -95,7 +99,7 @@ class _DummyModel:
     def prepare_encoder_cudagraph_replay_buffers(self, mm_kwargs,
                                                  max_batch_size,
                                                  max_frames_per_batch):
-        return EncoderCudaGraphReplayBuffers(values={})
+        return EncoderCudaGraphReplayBuffers(buffers={})
 
     def postprocess_encoder_output(self,
                                    output,
@@ -181,22 +185,22 @@ class TestAdapterPostprocessEncoderOutput:
 
 
 class TestPadToTemplate:
-    """Test MMEncoderJITManager._pad_to_template — zero-padding, slicing, and dtype-casting.
+    """Test MMEncoderJITManager._pad_dict_to_template — zero-padding,
+    slicing, and dtype-casting.
 
     Pure torch logic — tested without a full manager init by constructing a
-    bare instance via object.__new__ and setting only budget_templates.
+    bare instance via object.__new__.
     """
 
-    def _manager(self, template: dict) -> MMEncoderJITManager:
-        m = object.__new__(MMEncoderJITManager)
-        m.budget_templates = {256: template}
-        return m
+    def _manager(self) -> MMEncoderJITManager:
+        return object.__new__(MMEncoderJITManager)
 
     def test_general_case_zeros_then_copies(self):
         """src smaller than template: zero the buffer and slice-copy src."""
         tmpl = torch.zeros(10, 4)
         src = torch.ones(4, 4)
-        result = self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
+        result = self._manager()._pad_dict_to_template({"pv": src},
+                                                       {"pv": tmpl})
         assert result["pv"].shape == (10, 4)
         assert torch.all(result["pv"][:4] == 1.0)
         assert torch.all(result["pv"][4:] == 0.0)
@@ -205,29 +209,30 @@ class TestPadToTemplate:
         """src already has the template shape: returned as-is (no allocation)."""
         tmpl = torch.zeros(8, 4)
         src = torch.ones(8, 4)
-        result = self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
+        result = self._manager()._pad_dict_to_template({"pv": src},
+                                                       {"pv": tmpl})
         assert result["pv"] is src
 
     def test_scalar_uses_template_value(self):
         """0-dim scalars (e.g. max_seqlen): always use the budget-fixed template."""
         tmpl = torch.tensor(256)
         src = torch.tensor(64)
-        result = self._manager({
-            "max_seqlen": tmpl
-        })._pad_to_template({"max_seqlen": src}, 256)
+        result = self._manager()._pad_dict_to_template({"max_seqlen": src},
+                                                       {"max_seqlen": tmpl})
         assert result["max_seqlen"] is tmpl
 
     def test_none_src_uses_template(self):
         """Key absent from replay_values (or explicitly None): use template."""
         tmpl = torch.zeros(8, 4)
-        result = self._manager({"optional": tmpl})._pad_to_template({}, 256)
+        result = self._manager()._pad_dict_to_template({}, {"optional": tmpl})
         assert result["optional"] is tmpl
 
     def test_dtype_cast_on_copy(self):
         """src dtype differs from template: cast to template dtype during copy."""
         tmpl = torch.zeros(10, 4, dtype=torch.bfloat16)
         src = torch.ones(4, 4, dtype=torch.float32)
-        result = self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
+        result = self._manager()._pad_dict_to_template({"pv": src},
+                                                       {"pv": tmpl})
         assert result["pv"].dtype == torch.bfloat16
         assert torch.all(result["pv"][:4] == 1.0)
 
@@ -253,9 +258,11 @@ class TestMMEncoderJITManagerInit:
     def test_budget_templates_keyed_by_budget(self):
         """One template entry per budget, shape matches the budget size."""
         manager = _make_manager()
-        assert set(manager.budget_templates.keys()) == set(
+        assert set(manager.mm_kwargs_templates.keys()) == set(
             manager.token_budgets)
-        for budget, tmpl in manager.budget_templates.items():
+        assert set(manager.buffers_templates.keys()) == set(
+            manager.token_budgets)
+        for budget, tmpl in manager.mm_kwargs_templates.items():
             assert "pixel_values" in tmpl
             assert tmpl["pixel_values"].shape[0] == budget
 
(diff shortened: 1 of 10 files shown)


lk-chen:
https://github.com/vllm-project/vllm/pull/44863 resolves `2. Qwen3-VL Moe __init__ workarounds`

PR #3171, opened 2026-07-13
docs(dflash): add Speculative Decoding: DFlash to feature support matrix

huybro:
## Summary
This PR adds **`Speculative Decoding: DFlash`** to the feature support matrices and documentation


Lumosis:
Have we tested it on v6e as well? If not, let's remove that for now.

huybro:
> Have we tested it on v6e as well? If not, let's remove that for now.

We have tested it in buildkite tests.

bvrockwell (approved):

PR #3174, opened 2026-07-13
feat(continue_decode): env-gated EOS-check interval for the fused decode loop

lokic233:
## Summary
Adds a `CONTINUE_DECODE_EOS_CHECK_INTERVAL` env var that controls how often the fused `continue_decode` loop checks the any-sequence-hit-EOS early-exit condition, so it can be retuned per model family (or on a serving fleet) without a rebuild.

## Relationship to #3068
#3068 (merged) added the `continue_decode_eos_check_interval` control and the gating logic in `decode_loop.py` (`cond_fn`: check the early exit only on steps where `i % interval == 0`).

Rebased on top of #3068, this PR **reuses that exact gating path** — `decode_loop.py` is untouched. Per review, the value is sourced **solely from the `CONTINUE_DECODE_EOS_CHECK_INTERVAL` env var** (the `additional_config` key has been consolidated away):
1. registers the env var (`envs.py`), and
2. reads it into `self.continue_decode_eos_check_interval` (`tpu_runner.py`), which flows into #3068's gating.

Default / unset = `1` = stock every-step check (behavior unchanged).

## Why
Every sequence still stops at its own EOS — the loss masks the `≤ N-1` extra decoded-then-discarded tokens exactly like a normal stop — so the sampled distribution is unchanged; the loop just amortizes the per-step early-exit dispatch (dispatch-bound on v7x). Measured on TPU v7x DAPO rollout: interval 8 is a clear win (~20%); benefit saturates by 8.


pv97:
This looks similar to https://github.com/vllm-project/tpu-inference/pull/3068

lokic233:
@pv97 thanks — they're related but solve different things:

- **#3068 (yours):** a flag to *not exit* the decode loop on EOS (keep generating past EOS). That's about **whether** to stop.
- **#3174 (this):** an env-gated *interval* for **how often** the fused continue-decode loop runs the EOS check. Checking every token adds a sync per step; checking every N tokens amortizes it (the +20% comes from that reduced sync overhead), at the cost of overrunning EOS by up to N-1 tokens.

They're complementary and compose: #3068 decides the stop policy, #3174 tunes the check cadence. Happy to rebase on top of #3068 if it lands first, or coordinate so the two don't conflict on the loop structure. Do you want me to stack this on #3068, or keep them independent?

pv97 (approved):

tpu_inference/runner/tpu_runner.py
```diff
@@ -856,8 +857,19 @@ def __init__(
...
+        # key and the CONTINUE_DECODE_EOS_CHECK_INTERVAL env var feed this single
+        # knob and share decode_loop's gating logic. The env var takes precedence
+        # when explicitly set, so operators can retune it per model family (or on
+        # a serving fleet) without editing additional_config; otherwise the
+        # additional_config value (default 1 = stock every-step check) is used.
         self.continue_decode_eos_check_interval = self.vllm_config.additional_config.get(
             "continue_decode_eos_check_interval", 1)
+        if "CONTINUE_DECODE_EOS_CHECK_INTERVAL" in os.environ:
+            self.continue_decode_eos_check_interval = (
+                envs.CONTINUE_DECODE_EOS_CHECK_INTERVAL)
```
wenxindongwork: can we use tpu_inference.envs here?

```
import tpu_inference.envs as envs
envs.CONTINUE_DECODE_EOS_CHECK_INTERVAL
```

tpu_inference/runner/tpu_runner.py
```diff
@@ -856,8 +857,19 @@ def __init__(
...
+        # continue_decode EOS-check interval: how often the fused decode loop
+        # observes the any-sequence-hit-EOS early exit (see decode_loop.py and
+        # envs.CONTINUE_DECODE_EOS_CHECK_INTERVAL). Both the additional_config
+        # key and the CONTINUE_DECODE_EOS_CHECK_INTERVAL env var feed this single
+        # knob and share decode_loop's gating logic. The env var takes precedence
+        # when explicitly set, so operators can retune it per model family (or on
+        # a serving fleet) without editing additional_config; otherwise the
+        # additional_config value (default 1 = stock every-step check) is used.
         self.continue_decode_eos_check_interval = self.vllm_config.additional_config.get(
             "continue_decode_eos_check_interval", 1)
```
wenxindongwork: if we could pass this in as env var, do we still need the additional config path?

tpu_inference/envs.py
```diff
@@ -319,6 +320,18 @@ def _get_int_list_env() -> list[int]:
     env_bool("USE_JAX_PROFILER_SERVER"),
     "JAX_PROFILER_SERVER_PORT":
     lambda: int(os.getenv("JAX_PROFILER_SERVER_PORT") or "9999"),
+    # continue_decode: how often the fused multi-step decode loop checks the
```
wenxindongwork: nit: this comment could be shortened

tpu_inference/runner/tpu_runner.py
lokic233: Done — added an `envs.is_set(name)` helper and switched to `envs.is_set("CONTINUE_DECODE_EOS_CHECK_INTERVAL")`, so the runner no longer reads `os.environ` directly (and dropped the now-unused `import os`). Added `is_set` coverage to `test_integer_env_vars` as well. (20acaf80)

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3174&ref=lokic233) on behalf of @lokic233_

tpu_inference/envs.py
lokic233: Done — trimmed from 10 lines to 4. (20acaf80)

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3174&ref=lokic233) on behalf of @lokic233_

tpu_inference/runner/tpu_runner.py
lokic233: Kept it for now: that path is what #3068 added and tested (`test_execute_continue_decode_eos_check_interval_config` asserts `additional_config={"continue_decode_eos_check_interval": 5}`), so dropping it would remove that surface and break the test. The env var layers on top purely as a runtime override that wins when explicitly set.

Happy to consolidate to env-only if you and @pv97 prefer that — just didn't want to unilaterally remove the config path you both landed in #3068. What's your preference?

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3174&ref=lokic233) on behalf of @lokic233_

wenxindongwork: yes let's consolidate to env only, thank you!

lokic233: Done — consolidated to env-only in `3afcaf47`. `continue_decode_eos_check_interval` is now sourced solely from the `CONTINUE_DECODE_EOS_CHECK_INTERVAL` env var (default 1 = stock); dropped the `additional_config` read and the `is_set()` precedence helper. #3068's decode_loop gating is unchanged — only the value source changed. Updated the affected tests accordingly.

_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_review_comment&utm_campaign=tpu-inference-3174&ref=lokic233) on behalf of @lokic233_

wenxindongwork (approved):

QiliangCui (approved):

PR #3175, opened 2026-07-13
Harden convert_torch_tensor_to_jax and LG4 forward test (post-#3157 review follow-up)

QiliangCui2023:
Follow-up from the #3157 review (round 2), three non-blocking hardening items:

1. **`convert_torch_tensor_to_jax` integer guard** — the shared helper routed every non-bf16 dtype through `.float()` and cast to the bf16 default, which silently corrupts integer payloads beyond the bf16 mantissa. Integer/bool tensors now keep their own dtype. This matters before `gemma4_mm.py` / `mm_encoder_jit_manager.py` migrate onto the helper (their inline copies convert int tensors like `pixel_position_ids`). Test with >2^16 values added.
2. **LG4 forward test skip-guard** — `if not jax.devices()` can never fire (it raises instead of returning empty); check `platform != "tpu"` so the test skips instead of failing on CPU-only hosts. Verified: skips under `JAX_PLATFORMS=cpu`, passes on TPU.
3. **Realistic `query_start_loc`** — `[0, num_tokens]` cumsum instead of degenerate all-ones metadata.

Remaining tracked follow-up: migrate the two inline bf16 bit-view copies (`gemma4_mm.py`, `mm_encoder_jit_manager.py`) onto the shared helper.

### DIFF
--- a/tests/models/jax/test_llama_guard_4.py
+++ b/tests/models/jax/test_llama_guard_4.py
@@ -309,8 +309,10 @@ class TestLlamaGuard4ForwardContract:
 
         # Single-device mesh: the ragged-attention shard_map requires the
         # tiny request metadata to divide the data axis evenly.
-        if not jax.devices():
-            pytest.skip("No JAX devices available for mesh creation.")
+        # (jax.devices() raises rather than returning [] when no backend
+        # exists, so check the platform explicitly.)
+        if jax.devices()[0].platform != "tpu":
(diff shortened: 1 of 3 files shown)


lk-chen (approved):

QiliangCui2023:
Closing: #3183 reverts #3157 (and #3142) per the release owner's decision to drop the Llama-4 matrix tests, which removes the code this PR hardens.

PR #3178, opened 2026-07-14
[MTP] Mamba state rollback for speculative decoding

Lumosis:
## Summary

Wires the GDN v3 windowed kernel (merged in #3189) into the runner and attention layers so hybrid mamba models can undo the linear-attention state advance of rejected draft tokens during MTP speculative decoding. Previously the kernel only produced the state after the last token of a verify window, so a sequence's mamba state silently kept the effect of rejected draft tokens. This PR keeps one state checkpoint per window position and resumes each request from the checkpoint of its last accepted token — rollback becomes pure slot selection, with no state copies.

This is the runner/layer integration half of the feature; the kernel half landed separately as #3189. It sits directly on top of `main` now that #3189 is merged.

## How it works

The data flow after each verify step is: rejection sampler → per-request read offset → attention metadata → persistent-batch ordering → GDN kernel.

- **Read offset.** `extract_last_sampled_tokens` now also returns `mamba_read_offsets = num_accepted - 1` per sequence. This is scattered into a per-*physical-slot* buffer on device so the value survives requests being rescheduled, condensed, or skipped across steps. The kernel reads a request's initial state from `state_indices[s] + read_offset[s]` (last accepted token's checkpoint) and writes fresh checkpoints starting at `state_indices[s]`.
- **Batch ordering.** The persistent batch is reordered into three contiguous segments `[1-token decodes][spec verify windows][prefill/mixed]`. Ragged paged attention keeps using its 1-token decode front segment via `request_distribution`, while the GDN kernel runs its windowed mode over the first two segments via the new `mamba_request_distribution` (same 3-counters-per-rank format, but the first segment covers all windowed sequences). Both segmentations hold at once, so RPA and GDN read the same batch without conflict.
- **Kernel call.** The GDN op gathers the per-slot offsets (`slot_read_offsets[state_indices]`) inside the shard and passes `read_offsets` / `num_spec_tokens` to `fused_conv1d_gdn`. `num_spec_tokens` is recovered statically from the vLLM MambaSpec conv-state width, so no new plumbing is needed to thread it through.

## Metadata additions

Two fields on `AttentionMetadata`, both `None` unless the model has mamba layers *and* speculative decoding is enabled:

| Field | Shape | Meaning |
|---|---|---|
| `mamba_slot_read_offsets` | `(mamba_num_blocks,)` | per-slot `num_accepted - 1` from the last verify step |
| `mamba_request_distribution` | `(3 * dp_size,)` | GDN request distribution whose first segment covers windowed sequences (decodes + verify windows) |

## Files

- `spec_decode/jax/utils.py` — third return value `mamba_read_offsets`.
- `layers/common/attention_metadata.py` — the two fields above.
- `runner/persistent_batch_manager.py` — three-segment `[decode][verify][prefill]` reorder.
- `layers/common/gdn_attention.py`, `layers/vllm/custom_ops/gdn_attention_op.py` — gather offsets, pass `read_offsets` / `num_spec_tokens` to the kernel.
- `runner/tpu_runner.py`, `runner/compilation_manager.py` — scatter offsets after rejection sampling, build `mamba_request_distribution`, precompile the new op.

## Tests

`tests/spec_decode/test_utils.py` — `extract_last_sampled_tokens` now checks the `mamba_read_offsets` output against a reference (6/6 pass).

### End-to-end

Verified serving `Qwen/Qwen3.5-4B` on v7x, TP=8, MTP `num_speculative_tokens=4`, with `VLLM_XLA_CHECK_RECOMPILATION=1` (so any un-precompiled shape hard-errors):

- Server starts with **no recompilation** — every runtime shape including the mamba scatter and windowed GDN precompiled cleanly.
- Counting prompt returns correct, unbroken output (`14 … 109`); single-prompt spec metrics 90.4% acceptance, per-position `1.0 / 1.0 / 0.938 / 0.677`.
- `spec_bench` summarization, 100 prompts: **100/100 successful, 0 failed, 0 errors**. Acceptance length 2.90 with ~18.5k of 35.3k draft tokens rejected — the rollback path ran on nearly every step and never corrupted output.

### DIFF
--- a/tests/spec_decode/test_utils.py
+++ b/tests/spec_decode/test_utils.py
@@ -136,14 +136,19 @@ def test_extract_last_sampled_tokens_matches_parse_output(
 
     mesh = _make_mesh()
     with jax.set_mesh(mesh):
-        last_sampled, num_rejected = extract_last_sampled_tokens(
-            metadata, sampled, num_speculative, VOCAB_SIZE, max_num_seq, mesh)
+        (last_sampled, num_rejected,
+         mamba_read_offsets) = extract_last_sampled_tokens(
+             metadata, sampled, num_speculative, VOCAB_SIZE, max_num_seq, mesh)
 
     ref_last, ref_num_rej = _reference(sampled, num_draft_cpu, VOCAB_SIZE,
                                        max_num_seq)
 
     np.testing.assert_array_equal(np.asarray(last_sampled), ref_last)
     np.testing.assert_array_equal(np.asarray(num_rejected), ref_num_rej)
+    padded_num_draft = np.zeros(max_num_seq, dtype=np.int32)
+    padded_num_draft[:len(num_draft_cpu)] = num_draft_cpu
+    np.testing.assert_array_equal(np.asarray(mamba_read_offsets),
+                                  padded_num_draft - ref_num_rej)
 
 
 def test_filter_speculative_logprobs():
--- a/tpu_inference/layers/common/attention_metadata.py
+++ b/tpu_inference/layers/common/attention_metadata.py
@@ -55,26 +55,16 @@ class AttentionMetadata(object):
     # None for models without mamba layers; pure-mamba models would also
     # use this field, only hybrid models exercise it today.
     mamba_state_indices: jax.Array | None = None
-    # (mamba_num_blocks,) int32 — per-*slot* read offset for speculative
-    # decoding with mamba layers. `mamba_slot_read_offsets[base_slot]` is
-    # `num_accepted - 1` from the request's most recent verify step: the GDN
-    # kernel reads the request's initial state from
-    # `base_slot + offset` (the checkpoint of the last accepted token) and
-    # writes fresh checkpoints starting at `base_slot`. Indexed by physical
-    # slot (not batch position) so the value survives requests being
-    # rescheduled, condensed, or skipped for several steps. Updated on
-    # device after each rejection-sampling step; None unless the model has
-    # mamba layers *and* speculative decoding is enabled.
+    # (mamba_num_blocks,) int32 — mamba + spec decode only, else None. Per-slot
+    # state read offset (num_accepted - 1 from the last verify step): the GDN
+    # kernel resumes from checkpoint `base_slot + offset` and writes new
+    # checkpoints from `base_slot`. Indexed by physical slot so it survives
+    # rescheduling.
     mamba_slot_read_offsets: jax.Array | None = None
-    # (3 * dp_size,) int32 — GDN-specific request distribution, same
-    # 3-counters-per-rank format as `request_distribution` but with the
-    # first segment covering all *windowed* sequences (plain decodes and
-    # speculative verify windows of up to num_spec + 1 tokens) instead of
-    # only 1-token decodes. The persistent batch is ordered
-    # [decode][verify][prefill/mixed] so both segmentations hold at once:
-    # ragged paged attention keeps its 1-token decode front segment while
-    # the GDN kernel runs its windowed mode over the first two groups.
-    # None unless the model has mamba layers and spec decoding is enabled.
+    # (3 * dp_size,) int32 — mamba + spec decode only, else None. Like
+    # `request_distribution`, but its first segment counts all windowed
+    # sequences (decodes + verify windows), so the GDN kernel runs its windowed
+    # mode over the [decode][verify] prefix of the batch.
     mamba_request_distribution: jax.Array | None = None
 
     # (max_num_seqs, ) int32 — PCP only. For a single request, it is [rank*C, (2*pcp-1-rank)*C].
--- a/tpu_inference/layers/common/attention_metadata.py
+++ b/tpu_inference/layers/common/attention_metadata.py
@@ -27,6 +27,8 @@ import jax
         "query_start_loc",
         "request_distribution",
         "mamba_state_indices",
+        "mamba_slot_read_offsets",
+        "mamba_request_distribution",
         "pcp_q_pos_offsets",
         "pcp_kv_cache_lens",
     ],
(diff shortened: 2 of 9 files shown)


tests/spec_decode/test_utils.py
```diff
@@ -136,14 +136,19 @@ def test_extract_last_sampled_tokens_matches_parse_output(
```
gxd3, line 136: maybe add a e2e test case in spec-decoding-test?

gxd3 (approved):

tests/spec_decode/test_utils.py
Lumosis: Yes, we have MTP test running with Qwen3.5-4B. But the max_tokens is only 32 which is too small to catch this issue. I'll fix that in a separate PR

tpu_inference/layers/common/gdn_attention.py
```diff
@@ -121,15 +133,51 @@ def run_jax_gdn_attention(
...
         n_v=n_v // tp_size,
         d_k=d_k,
         d_v=d_v,
         kernel_size=kernel_size,
+        num_spec_tokens=num_spec_tokens,
     )
 
+    if slot_read_offsets is None:
+        p_run_jax_gdn_attention_local = functools.partial(
+            wrapper.fused_conv1d_gdn, **kernel_kwargs)
```
kyuyeunk: See here https://softwareengineering.stackexchange.com/questions/294320/is-kwargs-an-antipattern

Lumosis: Done. Thanks!

tpu_inference/layers/common/attention_metadata.py
```diff
@@ -53,6 +55,27 @@ class AttentionMetadata(object):
     # None for models without mamba layers; pure-mamba models would also
     # use this field, only hybrid models exercise it today.
     mamba_state_indices: jax.Array | None = None
+    # (mamba_num_blocks,) int32 — per-*slot* read offset for speculative
```
kyuyeunk: these comments are really verbose. please tighten them.

tpu_inference/layers/common/gdn_attention.py
```diff
@@ -152,6 +199,7 @@ def run_jax_gdn_attention(
         state_indices,
         distribution,
         seq_lens,
+        *args,
```
kyuyeunk: same comment

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
kyuyeunk, line 64: @pritha90 i thought we fixed this and changed into shared attention metadata?

tpu_inference/layers/vllm/custom_ops/gdn_attention_op.py
```diff
@@ -68,6 +68,13 @@ def gdn_attention_core_tpu(
     query_start_loc = first_attn_metadata.query_start_loc
     seq_lens = first_attn_metadata.seq_lens
     state_indices = first_attn_metadata.mamba_state_indices
+    slot_read_offsets = first_attn_metadata.mamba_slot_read_offsets
+    if first_attn_metadata.mamba_request_distribution is not None:
+        # Spec decoding: the GDN windowed segment covers both 1-token
+        # decodes and speculative verify windows (the batch is ordered
+        # [decode][verify][prefill/mixed]); RPA keeps using
+        # `request_distribution` with its 1-token decode front segment.
+        request_distribution = first_attn_metadata.mamba_request_distribution
```
kyuyeunk, line 77: > `request_distribution` with its 1-token decode front segment.

why? i had similar discussion with @amanseervi that decode and spec decode shouldn't behave any differently.

Lumosis: We keep two distributions because the same speculative verify window has opposite admissibility in the two kernels that read it. Both are built per DP-rank in runner/tpu_runner.py:2893-2896 — RPA's request_distribution uses num_decode (1-token decodes only) while GDN's mamba_request_distribution uses num_windowed (decodes + verify windows) — and the GDN one is only populated when has_mamba_layers and speculative_config is not None (tpu_runner.py:2899). RPA consumes request_distribution when it's handed to the ragged-paged-attention kernel at layers/jax/attention/attention.py:293, and its decode segment is strictly 1-token-per-request, so a multi-token window there would be incorrect, not merely slow — hence windows must stay out of num_decode. GDN instead swaps in mamba_request_distribution at layers/vllm/custom_ops/gdn_attention_op.py:77 (replacing the value read at :67), and its wrapper runs two passes at kernels/gdn/v3/wrapper.py:553-555 — BATCHED then PER_SEQ, split at distribution[0] (wrapper.py:488) — where only the batched pass uses window_size = num_spec+1 (wrapper.py:457) and thus emits the per-token state checkpoints rollback needs, while the per-seq pass uses window_size = 1 (wrapper.py:454) and produces only the final state; so windows must stay in GDN's num_windowed segment. Over the same reordered batch the fast-segment boundary genuinely differs per kernel, and unifying them would need a larger refactor (an RPA batched multi-token-window mode, or a GDN per-seq path that emits per-token checkpoints, which defeats the windowed design). It isn't different request-level behavior between decode and spec-decode — just a per-kernel dispatch boundary — so I'd keep the two-distribution approach here and track unification separately.

kyuyeunk: > RPA consumes request_distribution when it's handed to the ragged-paged-attention kernel at layers/jax/attention/attention.py:293, and its decode segment is strictly 1-token-per-request

which will never get triggered when spec decoding is enabled. when spec decoding is enabled, even RPA decode will have `mtp-tokenss + 1` number of tokens,.

PR #3182, opened 2026-07-14
[CI] Migrate multi-host benchmark

boe20211:
# Description

This PR migrates the legacy scripts/multihost/nightly_benchmarking.sh logic into the new .buildkite/benchmark unified structure. By consolidating the multi-host benchmarking flow into run_multihost.sh and run_job.sh. 
Additionally, it updates the configurations and execution commands for the Multihost benchmark, adds corresponding parameter explanations to the README, and includes specific usage instructions for  the Accuracy benchmark.

**1. Migration Plan & Legacy Scripts**
The rest of the legacy cases will be migrated in a follow-up PR very soon. Specifically, we will be porting the following cases next:
* `run_torchax_deepseek_v3_1k_8k`
* `run_torchax_deepseek_v3_1k_8k_random_routing`
* `run_torchax_deepseek_v3_8k_1k`
* `run_torchax_qwen3_coder_480b_1k_8k`
* `xprof_torchax_deepseek_v3_1k_1k`
* `xprof_torchax_deepseek_v3_1k_8k`
* `xprof_torchax_deepseek_v3_1k_8k_random_routing`
* `xprof_torchax_deepseek_v3_8k_1k`
* `run_jax_deepseek_v3_1k_8k`
* `run_jax_deepseek_v3_8k_1k`
* `xprof_deepseek_v3_1k_8k`

Once all of these cases are added and validated, we will completely delete the legacy `scripts/multihost/benchmarks/*` scripts. 

For the GCS log naming scheme: The new framework is already updated to use `REMOTE_LOG_ROOT="gs://vllm-bm-bk-storage/job_logs/$RECORD_ID/"`. This naming format aligns perfectly with our other standard benchmarks. Meanwhile, the legacy wrappers will continue to use their original naming scheme so there won't be any conflicts or drifting between the two parallel flows. When we are ready to flip to production, it will just be a single switch of the bucket variables without any naming conflicts.

**2. Performance Gate Intent (Daily Cases):**
*Note:* It is completely intentional that the new `daily` cases operate without a performance gate (`RUN_ONCE=true` and no `EXPECTED_THROUGHPUT`). These cases act as report-only metrics to monitor daily trends without blocking the CI, which perfectly matches the existing legacy cron behavior.

# Tests

[BuildKite(multihost/benchmarks/jax/run_deepseek_v3_1k_8k.sh)](https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/1824/canvas)

# Checklist

### DIFF
--- a/.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
+++ b/.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
@@ -1,0 +1,97 @@
+{
+  "global_env": {
+    "GCP_PROJECT_ID": "cloud-tpu-inference-test",
+    "GCS_BUCKET": "vllm-cb-storage2",
+    "GCP_INSTANCE_ID": "vllm-bm-inst",
+    "GCP_DATABASE_ID": "vllm-bm-bk-runs",
+    "RUN_ONCE": "true",
+    "IS_MULTI_HOST_BENCH": true,
+    "BK_TIMEOUT_IN_MINUTES": 240
+  },
+  "benchmark_cases": [
+    {
+      "case_name": "run_jax_deepseek_v3_1k_8k",
+      "ci_queue": [
+        "tpu_v7x_16_queue"
+      ],
+      "env": {
+        "RUN_TYPE": "DAILY",
+        "INPUT_LEN": 1024,
+        "OUTPUT_LEN": 8192,
+        "PREFIX_LEN": 0,
+        "MODEL_NAME": "DeepSeek-R1",
+        "DEVICE": "tpu7x-16",
+        "BENCHMARK_TIMEOUT": "3h"
+      },
+      "server_command_options": {
+        "command_type": "vllm_serve",
+        "env": {
+          "VLLM_DISABLE_SHARED_EXPERTS_STREAM": "0",
+          "NEW_MODEL_DESIGN": "1",
+          "VLLM_MLA_DISABLE": "0",
+          "MOE_REQUANTIZE_BLOCK_SIZE": "512",
+          "MOE_REQUANTIZE_WEIGHT_DTYPE": "fp4",
+          "MODEL_IMPL_TYPE": "flax_nnx",
+          "USE_UNFUSED_MEGABLOCKS": "0",
+          "MOE_ALL_GATHER_ACTIVATION_DTYPE": "fp8",
+          "TPU_BACKEND_TYPE": "jax",
+          "VLLM_ENGINE_READY_TIMEOUT_S": "10800",
+          "RAY_LOG_TO_STDERR": "1",
+          "AUTOSCALER_UPDATE_INTERVAL_S": "300"
+        },
+        "args": {
+          "model-path": "gs://tpu-commons-ci/deepseek/r1",
+          "model": "deepseek-ai/DeepSeek-R1",
+          "served-model-name": "deepseek-ai/DeepSeek-R1",
+          "tokenizer": "deepseek-ai/DeepSeek-R1",
+          "load-format": "runai_streamer",
+          "seed": 42,
+          "max-num-seqs": 160,
+          "max-num-batched-tokens": 512,
+          "tensor-parallel-size": {
+            "v7x-16": 16
+          },
+          "max-model-len": 9216,
+          "kv-cache-dtype": "fp8",
+          "gpu-memory-utilization": 0.95,
+          "enable-expert-parallel": true,
+          "no-enable-prefix-caching": true,
+          "async-scheduling": true,
+          "additional-config": "{\"sharding\": {\"sharding_strategy\": {\"enable_dp_attention\": true, \"expert_parallelism\": 16, \"tensor_parallelism\": 1}}, \"replicate_attn_weights\": \"True\", \"sparse_matmul\": \"True\", \"compilation_sizes\": [2560]}",
+          "generation-config": "gs://tpu-commons-ci/deepseek/multi-host-bench/DeepSeek-R1",
+          "trust-remote-code": true,
+          "hf-config-path": "deepseek-ai/DeepSeek-R1"
+        }
+      },
+      "client_command_options": {
+        "command_type": "vllm_bench_serve",
+        "args": {
+          "model": "deepseek-ai/DeepSeek-R1",
+          "backend": "vllm",
+          "request-rate": "inf",
+          "dataset-name": "random",
+          "random-input-len": 1024,
+          "random-output-len": 8192,
+          "num-prompts": 2560,
+          "ignore-eos": true,
+          "trust-remote-code": true,
+          "percentile-metrics": "ttft,tpot,itl,e2el"
+        }
+      },
+      "accuracy_command_options": {
+        "command_type": "local_benchmark_serving",
+        "args": {
+          "backend": "vllm",
+          "model": "deepseek-ai/DeepSeek-R1",
+          "dataset-name": "mmlu",
+          "dataset-path": "/workspace/mmlu/data/test",
+          "num-prompts": 14000,
+          "run-eval": true,
+          "temperature": 0,
+          "mmlu-output-len": 4,
+          "trust-remote-code": true
+        }
+      }
+    }
+  ]
+}
--- a/.buildkite/benchmark/scripts/parse_benchmark_log.py
+++ b/.buildkite/benchmark/scripts/parse_benchmark_log.py
@@ -1,0 +1,107 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+import json
+import re
+import sys
+
+
+def parse_benchmark_log(log_path, result_path):
+    # Mapping of what vllm prints vs what Spanner column expects
+    METRIC_MAPPING = {
+        "request throughput": "Throughput",
+        "output token throughput": "OutputTokenThroughput",
+        "total token throughput": "TotalTokenThroughput",
+        "median ttft": "MedianTTFT",
+        "p99 ttft": "P99TTFT",
+        "median tpot": "MedianTPOT",
+        "p99 tpot": "P99TPOT",
+        "median itl": "MedianITL",
+        "p99 itl": "P99ITL",
+        "median e2el": "MedianETEL",
+        "p99 e2el": "P99ETEL"
+    }
+
+    results = {}
+    in_results = False
+
+    try:
+        with open(log_path, "r") as f:
+            lines_generator = f
+            recent_lines = []
+
+            for line in lines_generator:
+                line = line.strip()
+
+                # Keep a sliding window of the last 10 lines
+                recent_lines.append(line)
+                if len(recent_lines) > 10:
+                    recent_lines.pop(0)
+
+                if "============ Serving Benchmark Result ============" in line:
+                    in_results = True
+                    continue
+                if "==================================================" in line and in_results:
+                    in_results = False
+
+                if in_results and ":" in line:
+                    key, val = line.split(":", 1)
+                    val = val.strip()
+
+                    # Remove units like (ms) or (tok/s) or (req/s)
+                    clean_key = re.sub(r"\(.*?\)", "", key).strip().lower()
+
+                    if clean_key in METRIC_MAPPING and METRIC_MAPPING[
+                            clean_key] not in results:
+                        if val != "N/A":
+                            results[METRIC_MAPPING[clean_key]] = val
+
+                # Handle explicit AccuracyMetrics: json (used by some benchmark wrappers)
+                if line.startswith("AccuracyMetrics:"):
+                    try:
+                        json_str = line.split("AccuracyMetrics:")[1].strip()
+                        # Verify it is valid JSON
+                        json.loads(json_str)
+                        results["AccuracyMetrics"] = json_str
+                    except Exception:
+                        pass
+
+                # Fallback: Parse legacy Accuracy result dict printed by benchmark_serving.py
+                if "Results" in recent_lines and line.startswith("{"):
+                    try:
+                        acc_dict = ast.literal_eval(line)
+                        if isinstance(acc_dict,
+                                      dict) and "accuracy" in acc_dict:
+                            results["AccuracyMetrics"] = json.dumps(
+                                {"accuracy": acc_dict["accuracy"]})
+                    except Exception:
+                        pass
+
+    except FileNotFoundError:
+        print(f"Warning: log file {log_path} not found.")
+
+    with open(result_path, "w") as out:
+        for k, v in results.items():
+            out.write(f"{k}={v}\n")
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(f"Usage: {sys.argv[0]} <log_file_path> <result_file_path>")
+        sys.exit(1)
+
+    log_file = sys.argv[1]
+    result_file = sys.argv[2]
+    parse_benchmark_log(log_file, result_file)
--- a/.buildkite/benchmark/scripts/parser_case.py
+++ b/.buildkite/benchmark/scripts/parser_case.py
@@ -20,7 +20,8 @@ import sys
 CMD_MAP = {
     "vllm_serve": "vllm serve",
     "vllm_bench_serve": "vllm bench serve",
-    "lm_eval": "lm_eval"
+    "lm_eval": "lm_eval",
+    "local_benchmark_serving": "python3 scripts/vllm/benchmarking/benchmark_serving.py"
 }
 
 # Helper function to print export statement only if value is valid
@@ -29,12 +30,27 @@ def export_env_if_valid(opts_dict, json_key, env_var_name):
 
     # Validate value: allow 0, but exclude None, empty dict, or empty string
     if val not in (None, {}, ""):
-        print(f"export {env_var_name}=\"{val}\"")
+        val_str = str(val).lower() if isinstance(val, bool) else val
+        print(f"export {env_var_name}=\"{val_str}\"")
 
-def get_current_machine_type():
+def normalize_device_name(device_name):
+    """
+    Normalizes device name from cases config (e.g., 'tpu7x-16' -> 'v7x-16')
+    """
+    if not device_name:
+        return None
+    normalized = device_name.lower()
+    if normalized.startswith("tpu"):
+        normalized = normalized[3:]
+    if not normalized.startswith("v") and len(normalized) > 0:
+        normalized = f"v{normalized}"
+    return normalized
+
+
+def get_current_machine_type(fallback_device=None):
     """
     Returns the current machine type string (e.g., 'v6e-8', 'v7x-2') 
-    using the tpu_info library.
+    using the tpu_info library, falling back to the case config if needed.
     """
     try:
         from tpu_info import device
@@ -54,14 +70,29 @@ def get_current_machine_type():
                 num_devices = num_chips * chip_type.value.devices_per_chip
 
             machine_type = f"{name}-{num_devices}"
-            print(f"echo '[DEBUG] Detected machine type: {machine_type}' >&2")
+            print(f"echo '[DEBUG] Detected local machine type via tpu_info: {machine_type}' >&2")
+
+            # Check if fallback_device matches the detected family but specifies a larger cluster count (multi-host setup)
+            if fallback_device and fallback_device.startswith(name):
+                try:
+                    fallback_count = int(fallback_device.split("-")[1])
+                    if fallback_count > num_devices:
+                        print(f"echo '[DEBUG] Multi-host TPU cluster detected. Local chips: {num_devices}, Cluster chips: {fallback_count}. Using cluster device: {fallback_device}' >&2")
+                        return fallback_device
+                except (ValueError, IndexError):
+                    pass
+
             return machine_type
         else:
             print(f"echo '[WARNING] No TPU chips detected: chip_type={chip_type}, num_chips={num_chips}' >&2")
     except ImportError:
-        print("echo '[WARNING] tpu_info library not found. Cannot determine machine type.' >&2")
+        print("echo '[WARNING] tpu_info library not found. Cannot determine machine type via library.' >&2")
     except Exception as e:
-        print(f"echo '[WARNING] Failed to determine machine type: {e}' >&2")
+        print(f"echo '[WARNING] Failed to determine machine type via library: {e}' >&2")
+
+    if fallback_device:
+        print(f"echo '[DEBUG] Falling back to case config device: {fallback_device}' >&2")
+        return fallback_device
     return None
 
 
@@ -99,11 +130,16 @@ def build_command(cmd_type, args_dict):
         return shlex.join(cmd_parts)
 
     for key, value in args_dict.items():
+        if key == "model" and "model-path" in args_dict:
+            continue
+
+        output_key = "model" if key == "model-path" else key
+
         if isinstance(value, bool):
             if value:
-                cmd_parts.append(f"--{key}")
+                cmd_parts.append(f"--{output_key}")
         else:
-            cmd_parts.append(f"--{key}")
+            cmd_parts.append(f"--{output_key}")
             cmd_parts.append(str(value))
 
     return shlex.join(cmd_parts)
@@ -116,9 +152,6 @@ def main():
 
     config_file = sys.argv[1]
     target_case = sys.argv[2] if len(sys.argv) > 2 else None
-
-    # Determine current machine type from tpu_info
-    current_machine = get_current_machine_type()
 
     try:
         with open(config_file, 'r') as f:
@@ -142,6 +175,15 @@ def main():
             f"echo 'Error: Case \"{target_case}\" not found in \"{config_file}\".' >&2; exit 1")
         sys.exit(1)
 
+    # Resolve fallback device from config
+    device_env = case_data.get("env", {}).get("DEVICE")
+    if not device_env:
+        device_env = data.get("global_env", {}).get("DEVICE")
+    fallback_machine = normalize_device_name(device_env)
+
+    # Determine current machine type
+    current_machine = get_current_machine_type(fallback_machine)
+
     merged_env = data.get("global_env", {}).copy()
     merged_env.update(case_data.get("env", {}))
 
@@ -153,7 +195,8 @@ def main():
 
     # Export environment variables securely
     for k, v in merged_env.items():
-        print(f"export {k}={shlex.quote(str(v))}")
+        v_str = str(v).lower() if isinstance(v, bool) else str(v)
+        print(f"export {k}={shlex.quote(v_str)}")
 
     srv_opts = case_data.get("server_command_options", {})
     cli_opts = case_data.get("client_command_options", {})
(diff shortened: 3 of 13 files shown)


.buildkite/scripts/setup_docker_env.sh
```diff
@@ -135,7 +135,13 @@ setup_environment() {
...
   fi
 
-  local CI_IMAGE_REPO="us-central1-docker.pkg.dev/cloud-ullm-inference-ci-cd/tpu-inference-ci/${IMAGE_NAME}"
+  local CI_IMAGE_REPO
+  if [[ "${IMAGE_NAME}" == *"/"* ]]; then
+    # If IMAGE_NAME is a full GCR path, replace the product repo name with the CI cache repo name
+    CI_IMAGE_REPO="${IMAGE_NAME/tpu-inference/tpu-inference-ci}"
+  else
+    CI_IMAGE_REPO="us-central1-docker.pkg.dev/cloud-ullm-inference-ci-cd/tpu-inference-ci/${IMAGE_NAME}"
+  fi
```
CienetStingLin: There might be an issue with the Docker image used here. Should we just use the same image as the one in `run_in_docker`?

scripts/multihost/run_cluster.sh
```diff
@@ -120,9 +120,13 @@ fi
...
 # Force cleanup of the image to ensure we pull the absolute latest
-echo "Ensuring we have the latest image for ${DOCKER_IMAGE}..."
-docker rmi "${DOCKER_IMAGE}" > /dev/null 2>&1 || true
-docker pull "${DOCKER_IMAGE}"
+if [[ "${IS_MULTI_HOST_BENCH,,}" != "true" ]]; then
+    echo "Ensuring we have the latest image for ${DOCKER_IMAGE}..."
+    docker rmi "${DOCKER_IMAGE}" > /dev/null 2>&1 || true
+    docker pull "${DOCKER_IMAGE}"
+else
+    echo "IS_MULTI_HOST_BENCH is true, skipping image pull to use prebuilt image ${DOCKER_IMAGE}."
```
CienetStingLin, line 128: The `IS_MULTI_HOST_BENCH` parameter alone does not explain why this is skipped. Please add more comments to explain the reason for skipping.

.buildkite/scripts/setup_docker_env.sh
boe20211: We are using the same image logic. Both scripts share the setup_environment function from setup_docker_env.sh, which automatically pulls the prebuilt image for CI environments. We use EXPORTED_CI_CACHE_IMAGE here so that all the remote worker nodes can pull the exact same prebuilt image from the registry.

.buildkite/scripts/run_multihost.sh
```diff
@@ -194,17 +218,112 @@ docker system prune -a --volumes -f || true
 # Source the environment setup script
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR/setup_docker_env.sh"
-# Pass "true" to enable pushing to GCR
-setup_environment "${IMAGE_NAME}" "true"
 
-DOCKER_IMAGE="${IMAGE_NAME}:${BUILDKITE_COMMIT:-latest}"
+# Determine the Docker image name and registry path.
+# Use the local image for multi-host benchmarks; otherwise, default to the remote GCR image.
+if [[ "$IS_MULTI_HOST_BENCH" == "true" ]]; then
```
QiliangCui: **[blocker]** `$IS_MULTI_HOST_BENCH` is referenced unguarded under `set -euo pipefail`, so every caller that does not set the env var now dies with "unbound variable". Live callers that do not set it: `.buildkite/pipeline_dev.yml:136`, `.buildkite/features/multi-host.yml:32`, and `pipeline_jax.yml:467` (test_26 — currently `skip:`-gated, so this PR's CI cannot catch it; it will fail as soon as the skip is lifted).

Suggest `if [[ "${IS_MULTI_HOST_BENCH:-false}" == "true" ]]; then`.

.buildkite/benchmark/scripts/run_job.sh
```diff
@@ -122,29 +126,81 @@ declare -a BENCHMARK_DOCKER_ARGS=(
 BENCHMARK_DOCKER_ARGS_STR="$(printf '%s\n' "${BENCHMARK_DOCKER_ARGS[@]}")"
 export BENCHMARK_DOCKER_ARGS_STR
 
-echo "--- Running job in docker via run_in_docker.sh"
-BM_JOB_STATUS=$EXIT_SUCCESS
+# Determine if it is a multi-host run.
+IS_MULTI_HOST_BENCH="${IS_MULTI_HOST_BENCH:-false}"
+if [[ ( "${VERSION:-}" == "7x" && ${COUNT:-0} -gt 8 ) || "${TPU_MULTIHOST_BACKEND:-}" == "ray" ]]; then
+    IS_MULTI_HOST_BENCH="true"
```
QiliangCui, line 132: **[blocker]** This auto-detect assignment is not `export`ed, so the child `run_multihost.sh` never sees it. For a 16-chip case whose JSON does not set `IS_MULTI_HOST_BENCH` in `global_env`, run_job.sh takes the multi-host branch here but `run_multihost.sh` then crashes on the unset variable at its `if [[ "$IS_MULTI_HOST_BENCH" == "true" ]]` check (or, once a `:-false` default is added there, silently takes the single-host GCR image path). Build 1775 only exercised the flag-in-JSON path.

Add `export IS_MULTI_HOST_BENCH` after this block.

.buildkite/scripts/run_multihost.sh
```diff
@@ -118,14 +126,17 @@ cleanup() {
     echo "==================== START OF VLLM SERVE LOG ===================="
     cat /tmp/vllm_serve.log || true
     echo "==================== END OF VLLM SERVE LOG ===================="
+    rm -f /tmp/vllm_serve.log
```
QiliangCui: **[major]** `cleanup()` runs on EXIT — i.e. before control returns to run_job.sh — and deletes `/tmp/vllm_serve.log` right after catting it. That makes run_job.sh's handoff (`mv /tmp/vllm_serve.log "$LOG_FOLDER/vllm_log.txt"`, run_job.sh:181-183) dead code, so on failures before run_bm.sh runs (server never became healthy — the most common multi-host failure) no vllm log artifact is uploaded at all.

Keep the file for the caller (delete it only in the start-of-run cleanup invocation), or copy it to a location run_job.sh owns before removing.

.buildkite/scripts/run_multihost.sh
```diff
@@ -312,3 +464,18 @@ else
 fi
 
 echo "--- Tests completed successfully"
+
+# Write GCS upload metadata variables to a temp file for run_job.sh
+METADATA_FILE="/tmp/multihost_run_metadata.sh"
```
QiliangCui: **[major]** The metadata file is only written on full success (the script exits on any earlier failure), but run_job.sh needs it precisely on failure: the legacy-GCS upload names the log `${MODEL_NAME:-unknown}_${INPUT_LEN:-0}_${OUTPUT_LEN:-0}_${impl_type}_..._FAILED_...`, so failed runs upload as `unknown_0_0_vllm_...` — wrong model and wrong impl_type (flax_nnx falls back to "vllm").

Write this file right after `source "$TEMP_EXPORT_FILE"` (~L273), where MODEL_NAME/INPUT_LEN/OUTPUT_LEN/SERVER_CMD_ENVS are already known.

.buildkite/benchmark/scripts/report_result.sh
```diff
@@ -137,12 +137,29 @@ RUN_TYPE="${RUN_TYPE:-DAILY}"
 RESULT_FILE="${ARTIFACT_FOLDER}/${RECORD_ID}.result"
 
 # Upload logs to GCS if bucket is provided
-if [[ -n "${GCS_BUCKET:-}" ]]; then
+if [[ -n "${GCS_BUCKET:-}" && "${SERVER_ALREADY_RUNNING:-false}" != "true" ]]; then
```
QiliangCui: **[major]** Skipping the unified GCS upload in multi-host mode means multi-host Spanner rows have no `job_logs/$RECORD_ID/` folder — log lookup by RecordId breaks for exactly these cases — and the "legacy GCS location" mentioned in the message only receives the server log (run_job.sh uploads just `$VLLM_LOG`), not bm_log or the result file.

If the reason is missing GCS credentials inside the head-node container, mounting `/etc/boto.cfg` there (as single-host's `BENCHMARK_DOCKER_ARGS` does) would let this unified upload work unchanged. If the skip must stay, the DB row should record where the logs actually live.

.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -135,10 +141,52 @@ export ARTIFACT_FOLDER
...
+
+    # Download dataset using wget if it is not present in the workspace
+    DATASET_DIR="/workspace/mmlu"
+    if [ ! -d "$DATASET_DIR/data/test" ]; then
+      echo "MMLU dataset test folder not found, downloading via wget..."
+      mkdir -p "$DATASET_DIR"
+      cd "$DATASET_DIR" || exit 1
+      if [ ! -f data.tar ]; then
+        echo "Downloading data.tar..."
+        wget https://people.eecs.berkeley.edu/~hendrycks/data.tar -P .
```
QiliangCui: **[major]** Robustness of the accuracy phase:
- Under `set -e`, an unguarded `wget`/`tar` failure (or the `cd ... || exit 1`) bypasses `report_and_exit` entirely — no DB report at all. This mirror is a personal university homepage; transient failures are likely. Add `--tries`/`--timeout` and route failures through `report_and_exit`.
- When accuracy fails after a successful perf phase, `report_and_exit 1` discards the already-measured perf metrics (report_result.sh skips parsing/DB when exit_code != 0). Consider reporting perf results even when the accuracy phase fails.
- The MMLU download is unconditional whenever `ACCURACY_CMD` is set — gate it on the accuracy args actually using mmlu (the JSON already hardcodes `dataset-path=/workspace/mmlu/data/test`, duplicated here).

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
```diff
@@ -0,0 +1,96 @@
...
+            "v7x-16": 16
+          },
+          "max-model-len": 9216,
+          "kv-cache-dtype": "fp8",
+          "gpu-memory-utilization": 0.95,
+          "enable-expert-parallel": true,
+          "no-enable-prefix-caching": true,
+          "async-scheduling": true,
+          "additional-config": "{\"sharding\": {\"sharding_strategy\": {\"enable_dp_attention\": true, \"expert_parallelism\": 16, \"tensor_parallelism\": 1}}, \"replicate_attn_weights\": \"True\", \"sparse_matmul\": \"True\", \"compilation_sizes\": [2560]}",
+          "generation-config": "gs://gpolovets-inference/deepseek/generation_configs/DeepSeek-R1",
```
QiliangCui: **[minor]** `generation-config` points at a personal bucket (`gs://gpolovets-inference/...`). A daily CI case now depends on a user-owned bucket that can disappear or change permissions — copy the config next to the weights under `gs://tpu-commons-ci/`.

.buildkite/scripts/run_multihost.sh
```diff
@@ -262,25 +387,27 @@ sleep 120
...
+if [[ "${VLLM_SERVE_CMD}" =~ --generation-config[=\ ]+gs://([^ ]+) ]]; then
+    gcs_path="gs://${BASH_REMATCH[1]}"
+    config_url="${gcs_path%/*}"
+    config_dir_name=$(basename "${config_url}")
+    config_file_name=$(basename "${gcs_path}")
+
+    echo "--- Detected GCS generation-config: $gcs_path"
+    echo "--- Pre-downloading config to workspace inside container..."
+
+    download_cmd="if ! command -v gsutil &> /dev/null; then curl -sSL https://sdk.cloud.google.com | bash -s -- --disable-prompts > /dev/null; export PATH=\"\$PATH:/root/google-cloud-sdk/bin\"; fi && mkdir -p /workspace && gsutil -m cp -r ${config_url} /workspace/ && "
```
QiliangCui: **[minor]** This adds a network install (`curl -sSL https://sdk.cloud.google.com | bash`) to the server-start path as a fallback. The CI image already ships gcloud/gsutil — if it is ever missing, failing explicitly beats a silent multi-minute SDK install spliced into the serve command. Suggest dropping the fallback.

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
```diff
@@ -0,0 +1,96 @@
...
+          "trust-remote-code": true,
+          "hf-config": "deepseek-ai/DeepSeek-R1"
```
QiliangCui: **[minor]** `hf-config` is not an actual vllm flag — this only parses because argparse abbreviation resolves it to `--hf-config-path`. Any future `--hf-config-*` flag in vllm makes the abbreviation ambiguous and kills server startup. Spell out `hf-config-path`.

.buildkite/benchmark/scripts/run_bm.sh
```diff
@@ -516,11 +610,16 @@ run_benchmark(){
   echo "[DEBUG] Executing client_cmd: ${CLIENT_CMD_ENVS[*]} ${CLIENT_CMD[*]} > $BM_LOG" >&2
   set +e
   # Execute the array directly, preserving strict argument boundaries
-  env "${CLIENT_CMD_ENVS[@]}" "${CLIENT_CMD[@]}" > "$BM_LOG" 2>&1
+  timeout 2h env "${CLIENT_CMD_ENVS[@]}" "${CLIENT_CMD[@]}" > "$BM_LOG" 2>&1
```
QiliangCui: **[minor]** This `timeout 2h` (and the one on the warmup path) applies to ALL existing single-host cases too, not just multi-host — a global behavior change riding along in a migration PR. A long daily case that legitimately exceeds 2h now dies mid-run. Suggest an env knob (default 2h) so cases can override. Also inconsistent: the accuracy exec in `run_accuracy_if_needed` gets no timeout at all. Related config tension in the new case: `VLLM_ENGINE_READY_TIMEOUT_S=10800` equals the entire `BK_TIMEOUT_IN_MINUTES=180` step budget, leaving zero time for the benchmark if startup is that slow.

.buildkite/scripts/run_multihost.sh
```diff
@@ -290,10 +417,35 @@ docker exec \
...
+      -e SERVER_ALREADY_RUNNING="true" \
+      -e VLLM_PORT="${VLLM_PORT}" \
+      node bash -c "cd /workspace/tpu_inference && chmod +x .buildkite/benchmark/scripts/run_bm.sh && .buildkite/benchmark/scripts/run_bm.sh $CASE_FILE $TARGET_CASE_NAME"
+  else
+    echo "--- Running Benchmark Command on Head Node"
+    docker exec \
+      -e HF_HOME=/root/.cache/huggingface \
+      node bash -c "cd /workspace/tpu_inference && ${CLIENT_BENCH_CMD}"
+  fi
+elif [ "$#" -gt 0 ]; then
```
QiliangCui: **[minor]** Dead/hazardous branch: in legacy string mode, whenever remaining CLI args exist `CLIENT_BENCH_CMD="$*"` was already set, so the first branch consumes them and this `elif` is unreachable. In JSON mode the positional args are never shifted, so if a case yields an empty `CLIENT_CMD` (e.g. an lm_eval-style case), this branch executes `<case.json> <target>` as a shell command inside the container. Also: `${EXTRA_DOCKER_ARGS:-}` / `${DOCKER_ENV_STR}` rely on word-splitting — any env value containing a space breaks the docker invocation, unlike the `DOCKER_ENV_ARGS` array used alongside them.

.buildkite/benchmark/scripts/run_job.sh
```diff
@@ -122,29 +126,81 @@ declare -a BENCHMARK_DOCKER_ARGS=(
...
+        pip install --upgrade -r tools/kernel/tuner/v1/storage_management/requirements.txt && \
+        source .buildkite/benchmark/scripts/kernel_autotune.sh && \
+        update_all_tuned_params_py || exit 1
+      fi && \
+      if [[ \"${KERNEL_AUTOTUNE_STAGE:-}\" == \"POST_KERNEL_AUTOTUNE_BM_RERUN\" ]]; then
+        pip install --upgrade -r tools/kernel/tuner/v1/storage_management/requirements.txt && \
+        source .buildkite/benchmark/scripts/kernel_autotune.sh && \
+        checkout_updated_tuned_params_py_branch || exit 1
+      fi && \
+      echo always > /sys/kernel/mm/transparent_hugepage/enabled && \
```
QiliangCui, line 199: **[minor]** The multi-host branch skips the `transparent_hugepage=always` setup that this single-host branch performs, so head/worker nodes run without it — multi-host perf numbers will not be apples-to-apples with single-host. If THP matters for these measurements, add it to run_cluster.sh's container startup.

.buildkite/benchmark/scripts/parse_benchmark_log.py
```diff
@@ -0,0 +1,107 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+import json
+import re
+import sys
+
+
+def parse_benchmark_log(log_path, result_path):
```
QiliangCui, line 21: **[minor]** This parser is now load-bearing for every benchmark case (single- and multi-host) and is trivially unit-testable, but ships with no tests. A small pytest with a canned `vllm bench serve` output block plus the two accuracy formats would catch format drift — the shell logic being replaced used `grep -i` for a reason. Same applies to the new parser_case.py helpers (`normalize_device_name`, the multi-host device fallback, the model-path→model mapping).

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
```diff
@@ -0,0 +1,96 @@
+{
+  "global_env": {
+    "GCP_PROJECT_ID": "cloud-tpu-inference-test",
+    "GCS_BUCKET": "vllm-cb-storage2",
+    "GCP_INSTANCE_ID": "vllm-bm-inst",
+    "GCP_DATABASE_ID": "vllm-bm-bk-runs",
+    "RUN_ONCE": "true",
+    "IS_MULTI_HOST_BENCH": true,
```
QiliangCui, line 8: **[nit]** JSON `true` here is exported by parser_case.py's env loop as the Python string `True` (`export IS_MULTI_HOST_BENCH=True`), while every shell consumer compares lowercase `"true"` — run_cluster.sh only survives via `${IS_MULTI_HOST_BENCH,,}`, and run_bm.sh's `== "false"` check passes by accident. Normalize booleans to lowercase in the export loop (or quote "true" in the JSON).

.buildkite/benchmark/scripts/parser_case.py
```diff
@@ -213,6 +253,18 @@ def main():
             shlex.quote(arg) for arg in shlex.split(cli_cmd))
         print(f"CLIENT_CMD=({quoted_cli_cmd})")
 
+        acc_opts = case_data.get("accuracy_command_options")
+        if acc_opts:
+            acc_cmd_type = acc_opts.get("command_type", "local_benchmark_serving")
+            acc_resolved_args = resolve_device_args(acc_opts.get("args", {}), current_machine)
+
+            # If the user sets dataset-path with env var references like $DATASET_DIR/test,
+            # we need to be careful with shlex.quote. build_command already handles it but we should pass it correctly.
```
QiliangCui: **[nit]** Stale comment — there is no env-var substitution handling for `dataset-path` here; `build_command` just quotes the literal value. Drop or reword.

.buildkite/benchmark/scripts/parser_case.py
```diff
@@ -20,7 +20,8 @@
 CMD_MAP = {
     "vllm_serve": "vllm serve",
     "vllm_bench_serve": "vllm bench serve",
-    "lm_eval": "lm_eval"
+    "lm_eval": "lm_eval",
+    "local_benchmark_serving": "python3 /workspace/tpu_inference/scripts/vllm/benchmarking/benchmark_serving.py"
```
QiliangCui: **[nit]** This hardcodes the in-container absolute path, so `local_benchmark_serving` only works inside the CI image — and the README calls it "built-in `vllm/benchmarking/benchmark_serving.py`" while it is actually `scripts/vllm/benchmarking/benchmark_serving.py` from this repo. Consider resolving relative to the repo root instead.

QiliangCui (commented):
Reviewed the diff plus the surrounding scripts at the PR head (355302d). The migration direction is good and the happy path is proven — build 1775 passed end-to-end on `tpu_v7x_16_queue` (server start incl. the runai/GCS path, in-container run_bm + report, DB write). The `wait_for_server` hardening (curl `--max-time`, EngineCore fast-fail), the docker `--log-opt` caps, and reusing the CI cache image for workers are all solid improvements.

Inline: 2 blockers (same root cause — `IS_MULTI_HOST_BENCH` unbound / not exported, which bites exactly the caller paths this PR's CI doesn't exercise), 4 major, rest minor/nit.

Two cross-cutting asks:

1. **Migration plan.** `scripts/multihost/nightly_benchmarking.sh` and the `scripts/multihost/benchmarks/{jax,torchax}/*.sh` wrappers stay in-tree with only the jax 1k/8k case ported, and the prod upload destinations are commented-out TODOs in two separate places (report_result.sh, run_job.sh). Please spell out in the description: which cases migrate next, when the legacy scripts get deleted, and that flip-to-prod is a single switch — otherwise two parallel flows keep writing the same legacy GCS naming scheme and drift.
2. **Perf gate intent.** With `RUN_ONCE=true` and no `EXPECTED_THROUGHPUT`, the new daily case has no performance gate at all (report-only). That matches the legacy cron behavior, but worth stating explicitly that this is intentional.

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
boe20211: Done!
1. Updated parser_case.py to explicitly normalize Python boolean values (True/False) to lowercase ("true"/"false") during the environment variable export loop.
2. Also took the opportunity to unify the downstream checking logic across all shell scripts (run_cluster.sh, run_bm.sh, run_job.sh, run_multihost.sh). They now consistently use the strict format "${IS_MULTI_HOST_BENCH:-false}" == "true" without needing the hacky ${var,,} workaround.
Thanks for catching this!

.buildkite/benchmark/scripts/parse_benchmark_log.py
boe20211: Thanks for the feedback! I've implemented the suggested unit tests and improvements:
1. Added test_parse_benchmark_log.py: Added a small pytest with a canned vllm bench serve output block (using a recent DeepSeek V3 log), along with tests for both the JSON and legacy accuracy metric formats.
2. Restored case-insensitive matching: Updated parse_benchmark_log.py to match metrics case-insensitively (bringing back the robustness of the old grep -i logic to catch format drift), and added a specific test case to verify this behavior.
3. Added test_parser_case.py: Included unit tests for the helper functions, covering normalize_device_name, the multi-host device fallback logic, and the model-path → model mapping.

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
boe20211: Done! I've copied the generation configs to be right next to the weights under gs://tpu-commons-ci/deepseek/r1 and updated all the references in our shell scripts, JSON configurations, and README to point to this new stable bucket.

.buildkite/benchmark/cases/daily/run_jax_deepseek_v3_1k_8k.json
boe20211: "Good point. I've updated it to explicitly use hf-config-path in the latest commit. Thanks!"

.buildkite/benchmark/scripts/parser_case.py
boe20211: "Yes, this has been resolved! I've updated local_benchmark_serving in CMD_MAP to use the relative path (scripts/vllm/benchmarking/benchmark_serving.py) instead of the hardcoded absolute path. It will now work correctly outside of the CI container as long as it's executed from the repository root."

.buildkite/benchmark/scripts/parser_case.py
boe20211: I have dropped the stale comment as suggested, since build_command only quotes the literal value and doesn't actually handle env-var substitution."

.buildkite/scripts/run_multihost.sh
boe20211: Good catch! I've updated the condition to [[ "${IS_MULTI_HOST_BENCH:-false}" == "true" ]] as suggested. This should properly guard against unbound variable errors for callers that don't set it.

.buildkite/benchmark/scripts/run_job.sh
boe20211: "Good catch! You are right, the auto-detect assignment wasn't exported so the child scripts never saw the updated value. I've added export IS_MULTI_HOST_BENCH right after the auto-detect block to fix this issue.

.buildkite/scripts/run_multihost.sh
boe20211: Good catch! You were completely right—the cleanup() function running on EXIT was deleting /tmp/vllm_serve.log right after printing it, leaving nothing for run_job.sh to move to the artifacts folder.

I've applied a fix based on your suggestion. The script now deletes /tmp/vllm_serve.log before docker cp copies the new log. This successfully clears out any stale logs at the start of the cleanup process, but leaves the fresh log intact on exit so run_job.sh can properly archive it when the server fails to become healthy.

.buildkite/scripts/run_multihost.sh
boe20211: Thanks for catching this! 
I've updated the script to write the metadata file immediately after `source "$TEMP_EXPORT_FILE"` (where the `MODEL_NAME`, `INPUT_LEN`, `OUTPUT_LEN`, and `SERVER_CMD_ENVS` are resolved).

.buildkite/benchmark/scripts/report_result.sh
boe20211: Thanks for the advice! I have updated the code to use the unified GCS upload location for multi-host as well. The previous separate vLLM log upload logic has been removed to keep everything consistent.

.buildkite/benchmark/scripts/run_bm.sh
boe20211: Thanks for the thorough review! I've addressed all the points regarding the accuracy phase robustness:
1. MMLU Download Gating: The dataset download logic in run_bm.sh is now properly gated behind a check that scans ACCURACY_CMD for the mmlu string. It will no longer download data.tar unconditionally for non-MMLU runs.
2. Download Robustness: I added --tries=3 and --timeout=15 to the wget command to handle transient network issues from the university mirror. Additionally, I've routed the cd, wget, and tar commands to || report_and_exit 1 so that any failure during the download/extraction phase is caught and reported to the DB, instead of silently bypassing the logic due to set -e.
3. Persisting Perf Metrics on Accuracy Failure: I refactored report_result.sh. Instead of exiting immediately when EXIT_CODE != 0, it now continues to run the metric parser and inserts the previously extracted performance metrics into the database. The job will still correctly record its Status as FAILED, but we will no longer lose the valuable performance data from the preceding phase.

.buildkite/benchmark/scripts/run_bm.sh
boe20211: Thanks for catching these! You're completely right. I've pushed a commit that addresses all your points:

Configurable Timeout: Replaced the hardcoded timeout 2h in run_bm.sh with a BENCHMARK_TIMEOUT env knob (defaults to 2h) so long-running cases don't die mid-run.
Accuracy Fix: Applied the exact same BENCHMARK_TIMEOUT logic to run_accuracy_if_needed, ensuring it now gracefully catches timeouts (exit code 124) instead of hanging the CI.

.buildkite/scripts/run_multihost.sh
boe20211: Good catch. You're right—having a multi-minute, silent SDK install spliced into the server start path is risky and masks underlying environment issues. I've removed the curl | bash fallback. It now strictly relies on the CI image's pre-installed gsutil, and will fail-fast with a clear error if it's ever missing. Updated in the latest commit

.buildkite/benchmark/scripts/run_job.sh
boe20211: Good catch! I've added the THP setup to run_cluster.sh's container startup command to ensure apples-to-apples measurements across all nodes.

.buildkite/scripts/run_multihost.sh
boe20211: Thanks for catching this! I've addressed both issues in the latest update:

I removed the dead/hazardous elif branch and refactored the conditional logic. Now, JSON configurations (including lm_eval-style cases with an empty CLIENT_CMD) will correctly and safely delegate to run_bm.sh.
I fixed the word-splitting issue by parsing ${EXTRA_DOCKER_ARGS:-} into an array and using array expansion ("${ARRAY[@]}") for the local Docker invocation, ensuring spaces inside variables are perfectly preserved."

CienetStingLin (approved):

PR #3188, opened 2026-07-14
[Spec Decoding] Fix scoped-VMEM OOM in spec-decode logits gather

Lumosis:
# Description

With speculative decoding, `_select_from_array_fn` gathers up to `max_num_seqs * (num_speculative_tokens + 1)` logits rows in one XLA gather, whose output is buffered entirely in scoped VMEM. Once the padded row count crosses ~64 (e.g. --max-num-seqs=16 with num_speculative_tokens=4 -> 128 rows x 62080 vocab shard in bf16 = 37.9M), compilation fails with:

  jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: E1001:
  CompileTimeScopedVmemOom ... exceeded scoped vmem limit (32.00M)

and the engine core dies mid-serving.

Gather in fixed 32-row chunks via lax.map so each gather's output stays small; small batches keep the original single-gather path. 

# Tests

Verified with a standalone repro of the failing [128, 62080] shape (fails before, passes after) and a full 100-prompt spec_bench run at --max-num-seqs=16 with qwen3_next_mtp (previously 53/100 requests failed; now 100/100 with no engine tracebacks).

# Checklist

### DIFF
--- a/tpu_inference/runner/tpu_runner.py
+++ b/tpu_inference/runner/tpu_runner.py
@@ -2007,7 +2007,19 @@ class TPUModelRunner(KVConnectorModelRunnerMixin, LoRAModelRunnerMixin):
     def _select_from_array_fn(array, indices_to_select, mesh):
 
         def select_local_fn(local_array, local_indices):
-            return local_array[local_indices]
+            # XLA keeps the whole row-gather output in scoped VMEM, so one
+            # gather of [n, vocab_shard] logits rows exceeds the 32M scoped
+            # limit once n grows (e.g. spec decode with max_num_seqs=16:
+            # [128, 62080] bf16 = 37.9M -> CompileTimeScopedVmemOom).
+            # Gather in fixed-size row chunks so each gather stays small.
+            chunk = 32
+            n = local_indices.shape[0]
+            if n <= chunk or local_array.ndim < 2:
+                return local_array[local_indices]
+            num_pads = (-n) % chunk
+            idx = jnp.pad(local_indices, (0, num_pads)).reshape(-1, chunk)
+            out = jax.lax.map(lambda ix: local_array[ix], idx)
+            return out.reshape(-1, *local_array.shape[1:])[:n]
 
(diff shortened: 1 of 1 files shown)


tpu_inference/runner/tpu_runner.py
```diff
@@ -2007,7 +2007,19 @@ def _sample_from_logits(
     def _select_from_array_fn(array, indices_to_select, mesh):
 
         def select_local_fn(local_array, local_indices):
-            return local_array[local_indices]
+            # XLA keeps the whole row-gather output in scoped VMEM, so one
+            # gather of [n, vocab_shard] logits rows exceeds the 32M scoped
+            # limit once n grows (e.g. spec decode with max_num_seqs=16:
+            # [128, 62080] bf16 = 37.9M -> CompileTimeScopedVmemOom).
+            # Gather in fixed-size row chunks so each gather stays small.
+            chunk = 32
```
jrplatin, line 2015: How was this chosen?

Lumosis: chunk = 32 was actually chosen empirically.

Why not larger: 
The scoped-VMEM limit is 32 MB and XLA's fused gather showed ~2.4× overhead over the raw output (the crashing 128-row × 62080 bf16 gather was 15.9 MB raw but a 37.9 MB scoped allocation). 
At 64 rows (~19 MB scoped) it fits today's bf16 case but leaves little margin if the logits are fp32 or a model has a bigger per-shard vocab. 
32 is safe. 32 rows ≈ 4 MB raw / ~10 MB scoped, keeping ~3× headroom.

Why not smaller: 
Each chunk is a sequential lax.map iteration, so smaller chunks just add steps without saving anything meaningful. This gather runs once per sampling step over at most a few hundred rows, so 32 keeps the loop relatively short (128 rows → 4 iterations) — measured throughput is unchanged vs. the original single gather.

sixiang-google (approved):

PR #3191, opened 2026-07-15
feat(kv_cache): env-gated KV-cache persistence across weight sync

lokic233:
# Description

Adds `KV_CACHE_PERSIST_ACROSS_WEIGHT_SYNC` (default **off**). When set, `delete_kv_cache` and `reinitialize_kv_cache` become no-ops **while a KV cache is allocated**, persisting the allocation across a weight update instead of freeing + reallocating it.

**Motivation.** In RL rollout weight sync (e.g. tunix `vllm_sampler.update_params`), each sync does:
```
reset_prefix_cache()  ->  delete_kv_cache()  ->  <transfer new weights>  ->  reinitialize_kv_cache()
```
`reset_prefix_cache()` invalidates every prefix→block mapping, so no block computed under the old weights is reachable under the new weights — the next request re-prefills from scratch and overwrites its blocks before any decode read. The delete+reallocate is therefore **correctness-neutral** (it frees + reallocates HBM whose contents are already logically invalid) but costs a dispatch-bound realloc on every sync. On TPU v7x DAPO rollout this is ~30s/sync (per-step setup 37s → 3s with the flag).

**Correctness (measured).** Paired A/B on TPU v7x (Qwen3-0.6B, continue_decode, true-8k), KV-persist ON vs OFF, pinned data order:
- Completion-length divergence (ON vs OFF): mean 0.93%, max 2.22%.
- **A/A noise floor** (two OFF runs, both `KV_CACHE_PERSIST=0`): mean 0.64%, max 1.70%.
- The ON-vs-OFF divergence sits **within** the OFF-vs-OFF noise floor — it does not grow across steps and reward/aborted-ratio distributions track identically. i.e. KV-persist introduces no divergence beyond the backend's own run-to-run nondeterminism (the JAX vLLM backend does not support a per-request seed, so two identical runs already differ at ~0.6%). This matches the mechanism argument above.

Not claimed: bit-exactness (the backend cannot deliver it). Loss-parity under real gradients is a natural follow-up (needs a policy that is actually learning; not required for this correctness argument).

**Safety.** Opt-in; the guard only fires when `kv_caches` is non-empty (the first allocation still happens). Enable only when the caller resets the prefix cache on every weight sync and HBM headroom does not require the free for resharding. Default off preserves stock free+reallocate behavior exactly.

# Tests

- Added a bool-env regression test in `tests/test_envs.py`.
- Verified on TPU v7x 4x4x4: patch fires (`[kv-persist] SKIP delete_kv_cache/reinitialize_kv_cache`), setup 37s→3s, A/B within A/A floor as above.

# Checklist

- [x] I have added comments in hard-to-understand areas (mechanism + safety conditions).
- [x] I have added a regression test.

### DIFF
--- a/tests/test_envs.py
+++ b/tests/test_envs.py
@@ -64,6 +64,7 @@ def test_boolean_env_vars(monkeypatch: pytest.MonkeyPatch):
     monkeypatch.setenv("USE_MOE_EP_KERNEL", "0")
     monkeypatch.setenv("LAYOUT_Q_PROJ_AS_NDH", "0")
     monkeypatch.setenv("USE_BATCHED_RPA_KERNEL", "0")
+    monkeypatch.setenv("KV_CACHE_PERSIST_ACROSS_WEIGHT_SYNC", "0")
     monkeypatch.setenv("DISABLE_WEIGHT_REQUANTIZATION", "0")
 
     # Test SKIP_JAX_PRECOMPILE (default False)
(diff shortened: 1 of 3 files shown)


wenxindongwork (commented):
hi, could you please check the usage in https://github.com/vllm-project/tpu-inference/pull/3362 and https://github.com/vllm-project/tpu-inference/pull/3231? 

i don't think this env flag is needed if you do `start_weight_update( free_kv_cache=False) `

lokic233:
Thanks @wenxindongwork — confirmed. `start_weight_update(free_kv_cache=False)` (from #3231) covers this exact case: it skips `delete_kv_cache()` on start and `reinitialize_kv_cache()` on finish, persisting the allocation across the sync, with the same `reset_prefix_cache()`-first precondition. That's cleaner than a global env flag (per-call intent, and part of the standard weight-transfer API RL orchestrators already drive), so closing this in favor of the standard API. Appreciate the pointer.

— via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_comment&utm_campaign=PR3191&ref=dengcchi) on behalf of Loki Chen (@lokic233)

PR #3159, opened 2026-07-11
Scope xla_tpu_scoped_vmem_limit_kib to MLA kernel tests (fixes Multimodal Inputs similarity + e2e slowdown from #3011)

QiliangCui2023:
## What
Fixes `tpu7x Correctness tests for Multimodal Inputs` (`test_multi_modal_inference[False]`, Qwen2.5-VL-3B) failing with text similarity 0.8385 vs the 0.85 gate in the v0.25.0 release nightly ([build 22159](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22159)) and in main nightlies since Jul 7 (21709, 21954, 22076).

## Root cause (proven by single-commit A/B)
#3011 moved `LIBTPU_INIT_ARGS += --xla_tpu_scoped_vmem_limit_kib=65536` from `tests/kernels/mla_v2_test.py` into the **global** `tests/conftest.py`. Every pytest-driven e2e test — full engine runs included — now executes under a 64MB scoped-vmem limit, which changes XLA codegen for unrelated models and deterministically shifts Qwen2.5-VL-3B decoded text (similarity 0.985 → 0.8385, identical digits on every run).

A/B on `tpu-inference-dev` (same forced vLLM `fa4321de` on all sides, exact nightly test, tpu7x):
- parent `8bad85bed` (#2880): **PASS 3/3** (build 661)
- `0bbffcecb` (#3011): **FAIL 3/3, identical 0.8385** (build 663)
- also exonerated by A/B: #2977 (builds 657/658), Jul-6 control green (build 660)

## Fix
Move the env setting back into the two MLA kernel test modules that need it (`mla_v2_test.py`, `mla_tuned_vs_baseline_test.py`), before the `jax` import. `tests/conftest.py` no longer mutates `LIBTPU_INIT_ARGS` globally.

## Validation
Dev build on this branch running the exact nightly Multimodal Inputs job (3x) is in flight; results will be posted here.

The same fix applies to `main` (this PR targets the release branch per the release-blocker process).

Do not merge until validation is posted and the release owner approves.

PR #3134, opened 2026-07-09
[libtpu] Bump to `0.0.43`

jrplatin:
# Description

See title.

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: #123456

# Tests

Passing nightly build: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22192

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3011, opened 2026-06-26
Remove total pages from MLA tuning key

BirdsOfAFthr:
# Description
This PR removes `total_num_pages` from the `TuningKey` in MLA attention tuning. 
`total_num_pages` represents the overall capacity of the global page pool in HBM/DRAM. Since page indirection lookup calculations compile statically, the concrete total page size has no impact on VMEM tile allocation boundaries or local execution performance of MLA kernel blocks. 
Including it in the `TuningKey` caused lookups to fragment unnecessarily (resulting in tuning cache misses) when the global pool was scaled up or down, even when the underlying batch size and active sequence lengths were identical.
# Tests
Unit tests have been updated to remove `total_num_pages` from testcase naming and calculate input page requirements dynamically using `pages_per_seq * max_num_seqs`.
Tests were run successfully:
```bash
uv run pytest tests/kernels/mla_tuned_vs_baseline_test.py
```
*(All 20 test cases collected and completed successfully).*

# Checklist

Before submitting this PR, please make sure:
[x]  I have performed a self-review of my code.
[x]  I have necessary comments in my code, particularly in hard-to-understand areas.
[x]  I have made or will make corresponding changes to any relevant documentation.

PR #3194, opened 2026-07-15
[MLA] Pass vmem_limit_bytes to xpose_pipeline instead of relying on LIBTPU_INIT_ARGS

QiliangCui2023:
## Problem

The MLA v2 kernel's q_nope and output transposes (`xpose_pipeline` pallas_call) need ~33.3–33.6 MiB of scoped VMEM, which exceeds XLA's 32 MiB default (`--xla_tpu_scoped_vmem_limit_kib`) and fails compilation with `E1001: CompileTimeScopedVmemOom` on every cold-cache run. This is the "tpu7x/tpu6e JAX unit tests - kernels" failure on main builds 22345 and 22385 (`mla_v2_test.py::test_ragged_paged_attention_basic`, 33.56M vs 32.00M limit).

The previous mitigation (#3159, before that #3011's conftest setting) appended `--xla_tpu_scoped_vmem_limit_kib=65536` to `LIBTPU_INIT_ARGS` at MLA-test-module import. That is **order-dependent**: under `pytest tests/kernels`, `deepseek_v4/compressor_test.py` evaluates `jax.devices()` at import inside its `skipif` decorator, and `deepseek_v4/` collects alphabetically before `mla_*`, so libtpu initializes **before** the env var is appended and the flag is silently dropped. Whether the kernels CI job passed then depended purely on JAX compilation-cache warmth — warm cache skips the compile (no OOM), cold cache always fails. A 10x A/B/C stress matrix at the pre-#3134, #3134, and #3028 commits showed a flat ~2/10 cold-failure rate at all three, confirming the flake is environmental and none of those PRs caused it.

## Fix

`mla_ragged_paged_attention` already resolves `vmem_limit_bytes` (`None` → full VMEM capacity via `pltpu.get_tpu_info().vmem_capacity_bytes`, [kernel.py#L2268-L2269](https://github.com/vllm-project/tpu-inference/blob/589c3f9ea2b68c598bc5b4a7e2017b1184a0e11e/tpu_inference/kernels/mla/v2/kernel.py#L2268-L2269) — 64 MiB on tpu7x) and applies it to the main pallas_call through Mosaic `CompilerParams`. This PR threads the same value through `prepare_q_nope_inputs` / `prepare_outputs` into `xpose_pipeline`'s `CompilerParams`. The per-kernel Mosaic limit overrides the global XLA flag for that custom call only, so:

- no environment variable is needed at all (the fragile `LIBTPU_INIT_ARGS` blocks in `mla_v2_test.py` / `mla_tuned_vs_baseline_test.py` are removed), and
- nothing changes for any other compiled program (the #3159 concern — a global flag altering XLA codegen for unrelated e2e tests — cannot recur).

Both OOM-ing call sites are covered: `prepare_outputs` (384x128x512 bf16, 33.56M) and `prepare_q_nope_inputs` (128x352x512 f32, 33.30M).

## Verification

On a tpu7x-8 host, libtpu 0.0.43, with the JAX compilation cache disabled (`JAX_ENABLE_COMPILATION_CACHE=false`) to force cold compiles, over `tests/kernels/{mla_v2_test,mla_tuned_vs_baseline_test,transpose_test,deepseek_v4}`:

| | result |
|---|---|
| before, no flag | **8 failed** (all `CompileTimeScopedVmemOom` in `xpose_pipeline`: basic, basic_with_small_page_size, deepseekv3_batch_decode, mixed, prefill_only, sliding_window0/1/2), 173 passed |
| after (this PR) | **182 passed, 0 failed** |

A regression test is added in `transpose_test.py` with the exact failing shape (384x128x512 bf16, n_tile=128, m_tile=64) and `vmem_limit_bytes` set.

Follow-up (separate change): `deepseek_v4/compressor_test.py`'s import-time `jax.devices()` remains a collection-order trap for any future module-level env configuration; it should move to a lazily-evaluated skip condition.

### DIFF
--- a/tests/kernels/mla_tuned_vs_baseline_test.py
+++ b/tests/kernels/mla_tuned_vs_baseline_test.py
@@ -12,19 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-# MLA-kernel-only vmem budget. Two constraints (see #3011 / #3159):
-# 1. Must run before libtpu/TPU backend initialization (module top,
-#    before the jax import below).
-# 2. Must NOT move into tests/conftest.py: a global setting changes XLA
-#    codegen for every e2e test in the suite (that is exactly how #3011
-#    caused a deterministic multimodal output drift and e2e slowdowns).
-if "--xla_tpu_scoped_vmem_limit_kib" not in os.environ.get(
-        "LIBTPU_INIT_ARGS", ""):
-    os.environ["LIBTPU_INIT_ARGS"] = (os.environ.get("LIBTPU_INIT_ARGS", "") +
-                                      " --xla_tpu_scoped_vmem_limit_kib=65536")
-
 import time
 
 import jax
--- a/tests/kernels/mla_v2_test.py
+++ b/tests/kernels/mla_v2_test.py
@@ -12,19 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
-# MLA-kernel-only vmem budget. Two constraints (see #3011 / #3159):
-# 1. Must run before libtpu/TPU backend initialization (module top,
-#    before the jax import below).
-# 2. Must NOT move into tests/conftest.py: a global setting changes XLA
-#    codegen for every e2e test in the suite (that is exactly how #3011
-#    caused a deterministic multimodal output drift and e2e slowdowns).
-if "--xla_tpu_scoped_vmem_limit_kib" not in os.environ.get(
-        "LIBTPU_INIT_ARGS", ""):
-    os.environ["LIBTPU_INIT_ARGS"] = (os.environ.get("LIBTPU_INIT_ARGS", "") +
-                                      " --xla_tpu_scoped_vmem_limit_kib=65536")
-
 import jax
 import jax.numpy as jnp
 import numpy as np
--- a/tests/kernels/transpose_test.py
+++ b/tests/kernels/transpose_test.py
@@ -242,6 +242,26 @@ class TestXposePipelineTiling(parameterized.TestCase):
         expected = jnp.transpose(input_data, (1, 0, 2))
         self.assertTrue(jnp.allclose(result, expected))
 
+    def test_vmem_limit_bytes_allows_large_tiles(self):
+        # Regression test: compiling this bf16 384x128x512 transpose with
+        # n_tile=128, m_tile=64 needs ~33.6 MiB of scoped VMEM, which exceeds
+        # XLA's 32 MiB default (--xla_tpu_scoped_vmem_limit_kib) and fails
+        # compilation with E1001 CompileTimeScopedVmemOom. This is the exact
+        # shape the MLA v2 kernel's prepare_outputs transpose produces.
+        # vmem_limit_bytes raises the limit for this pallas_call only, without
+        # any global XLA flag.
+        shape = (384, 128, 512)
+        input_data = jax.random.normal(jax.random.PRNGKey(0),
+                                       shape,
+                                       dtype=jnp.bfloat16)
+        result = xpose_pipeline(input_data,
+                                transpose_axes=(1, 0, 2),
+                                n_tile=128,
+                                m_tile=64,
+                                vmem_limit_bytes=64 * 1024 * 1024)[0]
+        expected = jnp.transpose(input_data, (1, 0, 2))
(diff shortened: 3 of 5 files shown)


gxd3 (approved):

QiliangCui2023:
## Test evidence

All local runs on a tpu7x-8 host, libtpu 0.0.43 (via `TPU_LIBRARY_PATH`), with the persistent compilation cache hard-disabled (`JAX_ENABLE_COMPILATION_CACHE=false`, `JAX_COMPILATION_CACHE_DIR` unset) so every compile is real and warm-cache masking is impossible.

**1. Baseline A (pre-fix code, `LIBTPU_INIT_ARGS` hacks removed, cold):**

```
pytest -q tests/kernels/mla_v2_test.py tests/kernels/mla_tuned_vs_baseline_test.py \
          tests/kernels/transpose_test.py tests/kernels/deepseek_v4/
=> 8 failed, 173 passed in 1196.79s (0:19:56)
```

All 8 failures are `E1001 CompileTimeScopedVmemOom` in `jit(mla_ragged_paged_attention)/jit(xpose_pipeline)`:
`test_ragged_paged_attention_basic`, `basic_with_small_page_size`, `deepseekv3_batch_decode`, `mixed`, `prefill_only`, `sliding_window0/1/2` — covering both call sites:
- `prepare_outputs`: shape 384x128x512 bf16, scoped alloc **33.56M** vs 32.00M limit
- `prepare_q_nope_inputs`: shape 128x352x512 f32, scoped alloc **33.30M** vs 32.00M limit

**2. Verification B (this PR applied — identical command, identical env, only the code change):**

```
=> 182 passed, 0 failed in 1310.55s (0:21:50)
```

The 8 previous failures pass, plus the new regression test. The ~22-min runtime confirms real cold compiles (a warm run of the mla portion takes seconds). A/B is strict: same machine, same libtpu, same pytest invocation; the only variable is the code change.

**3. Regression test** (`transpose_test.py::test_vmem_limit_bytes_allows_large_tiles`): compiles the exact failing shape (384x128x512 bf16, n_tile=128, m_tile=64) with `vmem_limit_bytes=64 MiB` and checks numerical equality against `jnp.transpose`. Compiled cold and passed in run B.

**4. CI build [22497](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/22497) — passed.** The tpu7x kernels job's single first-attempt failure was `test_tuned_vs_baseline` (perf assertion, one -9.18% speedup sample) with **no OOM anywhere in the log** and 1044 tests passed before the `-x` stop; the retry passed the full suite. That perf flake is pre-existing and unrelated: in a 10x stress matrix on unmodified main-history commits it failed 3/30 jobs with the same signature.

**Effective limit note:** with callers passing `vmem_limit_bytes=None`, the kernel resolves it to `pltpu.get_tpu_info().vmem_capacity_bytes` = 64 MiB on tpu7x — byte-identical to the value the old `--xla_tpu_scoped_vmem_limit_kib=65536` flag set when it worked, so no new codegen behavior is introduced for previously-passing paths; the limit is just applied per-custom-call instead of process-globally.

**Not covered:** transpose kernel perf was not re-measured (compiled output should match the old flag-applied path per the note above); no local MLA model serve — CI e2e jobs cover that indirectly.

QiliangCui2023:
**Addendum — import-order + cold-start verified directly on this branch:**

The pre-fix failure was order-dependent (libtpu initializing during collection of `deepseek_v4/compressor_test.py`, before the mla test modules could set `LIBTPU_INIT_ARGS`) and only manifested on cold compiles. Two additional runs on this branch, compilation cache hard-disabled:

1. **CI's exact directory form** (`pytest tests/kernels --ignore=... -k "...basic..."`), where `deepseek_v4/` collects before `mla_*`. A collection-hook trace confirms the worst case occurs — TPU backend already initialized before `mla_v2_test.py` is imported, and no scoped-vmem flag present in `LIBTPU_INIT_ARGS` at any point — and the previously-failing test **passes with a real cold compile** (35.9s). This exact configuration failed deterministically (5/5) before the fix.
2. **Minimal two-file pair with the bad import order** (`compressor_test.py` before `mla_v2_test.py`), the smallest reliable pre-fix repro: **passes cold** (35.5s).

Order independence is structural: the limit now travels inside the traced program (Mosaic `CompilerParams` on the pallas custom call) rather than through a process env var consumed once at libtpu load.

BirdsOfAFthr (approved):

kyuyeunk (approved):

PR #3198, opened 2026-07-16
rpa_v3_cp: support head_dim>128 

wenxindongwork:
## Summary
`rpa_v3_cp` fails to compile at `head_dim=256`; this makes it work. Two independent fixes, both no-ops at `head_dim=128`:

1. **Strided KV-cache write (`_update_kv_cache_partial`)** — the strided store bitcasts packed KV to uint32 and issues a strided vector load whose lane (last) dim is `head_dim`. Mosaic requires the base memref's lane dim to be exactly 128, so `head_dim=256` failed with:
   ```
   Not Implemented: The last dim size is not 128 in original base memref
   ```
   Fix: `head_dim` is always `align_to(.., 128)`, so reshape it to `(head_dim // 128, 128)` unconditionally — the strided load always sees a 128-wide lane dim, and `head_dim == 128` just gets a degenerate leading `1` and takes the same path (one code path for all head_dims).

2. **VMEM block sizing (`get_default_block_sizes`)** — the heuristic ignored `head_dim`, so the `head_dim=128`-tuned `bkv_sz=2048` overflowed VMEM at 256 (`Ran out of memory in memory space vmem. Used 65.71M of 63.94M`) for configs with several local KV heads. Fix: scale `bkv_sz` by `head_dim // 128` (factor 1 at 128).

## Testing — please read
These fixes were developed and validated against the **PCP** version of this kernel (on the `wxd-pcp` branch), where:
- the full PCP kernel suite passes at **both hd=128 and hd=256** (24/24 each), including the exact-equality strided/merged cache-write tests and PCP-vs-TP equivalence;
- `examples/offline_inference.py` on `Qwen/Qwen3-8B-Base` at tp=4 / pcp=2 produces correct outputs;
- a context sweep (16k->2M, chunk 16k, tp4/tp8/pcp2_tp4/pcp8_tp1) shows **no perf change at hd=128** (all configs within run-to-run noise; `tp4`/`tp8` don't touch this kernel and served as the noise control).

### DIFF
--- a/tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
+++ b/tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
@@ -1835,6 +1835,12 @@ def get_default_block_sizes(
     max_q = next_power_of_2(max_num_tokens)
     max_kv = pages_per_seq * page_size
 
+    # The KV compute/prefetch buffers scale with head_dim, but the default
+    # bkv_sz below is tuned for head_dim=128. For larger head_dim shrink the
+    # prefetch block proportionally so VMEM scratch stays within budget
+    # (head_dim=128 -> factor 1, so this is a no-op there).
+    hd_blocks = max(1, head_dim // 128)
+
     min_bkv_sz_to_peak = (16 * 1024 * 1024 * kv_packing // 4 // head_dim //
                           num_kv_heads_x2)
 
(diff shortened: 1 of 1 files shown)


tpu_inference/kernels/experimental/rpa_v3_cp/kernel.py
```diff
@@ -1821,6 +1832,12 @@ def get_default_block_sizes(
     max_q = next_power_of_2(max_num_tokens)
     max_kv = pages_per_seq * page_size
 
+    # The KV compute/prefetch buffers scale with head_dim, but the default
+    # bkv_sz below is tuned for head_dim=128. For larger head_dim shrink the
+    # prefetch block proportionally so VMEM scratch stays within budget
```
weiyu0824, line 1837: Discuss offline, and we think this is needed because previous bkv calculation didn't consider hd_blocks.

weiyu0824 (approved):

PR #3199, opened 2026-07-16
[Feat][KV Cache] Unified block-major KV cache for Qwen2/3: block id as the major dim for O(1) block transfer

yiw-wang:
## Description

This PR stores the KV cache of Qwen2/Qwen3 models as **one unified block-major array** instead of one array per layer:

```
per-layer (before): num_layers x (num_blocks, block_size, num_kv_heads*2/packing, packing, head_dim)
unified   (this PR): (num_blocks, num_layers, block_size, num_kv_heads*2/packing, packing, head_dim)
```

With `num_blocks` as the major dim, a block's KV across **all layers is contiguous in memory** and dim0 slice `pool[b]` *is* logical block `b`. The KV transfer engine can move a whole block with a **single DMA**, addressed natively by block id, instead of issuing `num_layers` DMAs per block — cutting per-layer transfer overhead in P/D disaggregation.

### Key Changes 

- Scope is gated to Qwen2/Qwen3 (one homogeneous attention group) via `model_uses_unified_kv_cache`. Other archs, speculative decoding, and hybrid (attention+Mamba) models keep the per-layer path unchanged (regression-tested).
- The RPA kernel is unchanged and still consumes its 5D `(num_pages, ...)` cache: the 6D->5D fold happens per device inside the attention `shard_map`, where it compiles to a **bitcast** (verified in HLO: all folded 5D views alias the pool at offset 0, zero copy).
- Page addressing is redirected in `attention()`: block `b`, layer `L` maps to flat page `b * num_layers + L` in the folded pool (`page_indices = block_tables * num_layers + layer_idx`), so each layer reads/writes its own slot through the unchanged 5D kernel with no data movement.

## Changes along the KV cache lifecycle

**1. Allocation** — `runner/kv_cache.py`, `runner/kv_cache_manager.py`
`create_unified_kv_cache` allocates the 6D pool with rank-6 sharding `P(ATTN_DATA, None, None, ATTN_HEAD, None, None)` (kv heads on dim3). `initialize_kv_cache` takes the unified path when the gate passes and all layers are regular attention, and records `attn_flat_indices` (layer -> pool slot).

**2. Model forward — the JAX and torchax paths converge at `attention()`:**
- *JAX (flax_nnx)* `models/jax/qwen2.py` / `qwen3.py`: a single-element `kv_caches` list is the unified pool; the layer loop threads the one pool through every layer with `layer_idx` / `num_layers`.
- *torchax (vLLM models)* `layers/vllm/backends/flash_attn.py`: the Pallas backend detects unified layers via `attn_flat_indices` (threaded through the model-wrapper context) and derives `layer_idx` the same way — vLLM model code untouched.

**3. Attention** — `layers/common/attention_interface.py`
`attention()` redirects page addressing: block `b`, layer `L` -> flat page `b * num_layers + L`. `sharded_ragged_paged_attention` folds the pool 6D->5D **inside its shard_map body** (per-device bitcast) and unfolds on return.

**4. Output sharding** — `models/common/model_loader.py`
`run_model`'s kv cache `out_shardings` pin is now rank-aware, selected by the same shared gate, so the pin always matches the allocated layout(mentioned in the Tests section).

**5. KV transfer / P-D handoff** — `distributed/kv_transfer.py`, `runner/kv_cache_manager.py`
The connector DMAs dim0 slices directly off the pool. The in-engine JAX transfer helpers unpack per-layer *views* (`pool[:, i]`, no copy) for extraction, and insert through the unified write path. The chunked-DMA jit gets a `jax.lax.optimization_barrier` on its output for the single-array case so the dest donation stays inplace.

## Tests

Key findings while landing this (each one was an OOM), all root-caused and verified on the real `run_model` HLO:

1. **Fold placement.** Folding 6D->5D in GSPMD-traced code lets the SPMD partitioner implement the reshape as a whole-pool relayout (TP=1) or all-to-all reshard (TP>1), ~2x pool size -> OOM. Moving the fold **inside the shard_map body** makes it a bitcast — HLO buffer assignment shows the 6D param and all folded 5D views at offset 0 of a single allocation.
2. **Rank-sensitive out_shardings.** `PartitionSpec` right-pads with `None`, so the legacy rank-3 pin on the rank-6 pool landed `ATTN_HEAD` on the page_size dim -> whole-pool all-to-all at the graph root -> OOM. Fixed with the rank-aware pin. 
3. **Donation aliasing.** With a single donated dest array, XLA's aliasing verifier rejected the transfer jit (memory-space mismatch on the entry output). `optimization_barrier` keeps the donation in place — verified same per-shard buffer pointers before/after, no extra copy.
4. **Single-host P/D disaggregation e2e on the Raiden connector** (Ref to tpu-raiden/examples/single_host_disagg, prefill and decode on separate chips): Benchmark 10/10 requests (median TTFT 84 ms). Correctness check: a ~900-token prompt with a secret number in the first block and the question at the end — decode answers correctly

## Next steps

- **Mixed KV dtype**: support mixed/quantized per-layer cache dtypes in the unified layout.
- **Mamba state**: extend the unified path to hybrid attention+Mamba models (they currently keep the per-layer path via the gate).

### DIFF
--- a/tpu_inference/runner/kv_cache_manager.py
+++ b/tpu_inference/runner/kv_cache_manager.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import dataclasses
 import os
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Tuple
 
 import jax
 import jax.numpy as jnp
@@ -46,7 +46,9 @@ from tpu_inference.offload.utils import get_kv_connector_cache_layout
 from tpu_inference.runner import utils as runner_utils
 from tpu_inference.runner.input_batch import CachedRequestState, InputBatch
 from tpu_inference.runner.kv_cache import (KVCacheMetadata, create_kv_caches,
-                                           get_attention_page_size_bytes)
+                                           create_unified_kv_cache,
+                                           get_attention_page_size_bytes,
+                                           model_uses_unified_kv_cache)
 
 if TYPE_CHECKING:
     from vllm.v1.request import Request
@@ -71,6 +73,25 @@ def is_ds_v4(vllm_config):
     return "DeepseekV4ForCausalLM" in (vllm_config.model_config.architectures)
 
 
+def _unpack_kv_caches(kv_caches: List[jax.Array],
+                      attn_flat_indices: Tuple[int, ...],
+                      num_total_caches: int) -> List[jax.Array]:
+    """Rebuild a per-layer KV cache list from the unified block-major array.
+
+    Compatibility path for consumers that still expect a per-layer list
+    (notably the in-engine JAX KV transfer). The dim1 slices are views, not
+    copies. No-op when attn_flat_indices is empty (per-layer allocation).
+    """
+    if not attn_flat_indices:
+        return kv_caches
+
+    large_attn_cache = kv_caches[0]
+    unpacked = [None] * num_total_caches
+    for i, flat_idx in enumerate(attn_flat_indices):
+        unpacked[flat_idx] = large_attn_cache[:, i]
+    return unpacked
+
+
 class KVCacheManager:
 
     def __init__(self, runner: "TPUModelRunner"):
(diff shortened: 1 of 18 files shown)


tpu_inference/runner/kv_cache_manager.py
```diff
@@ -784,6 +801,22 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+        # Allocate the unified block-major KV cache only when the shared gate
+        # passes (allowlisted arch, no spec decode; model_loader uses the same
+        # helper to pick the matching out_shardings pin) AND all layers are
+        # regular attention -- the unified pool holds one homogeneous
+        # attention group, so Mamba/hybrid layouts keep the per-layer path.
+        has_mamba_layers = any(
+            isinstance(spec, MambaSpec)
+            for spec in layer_name_to_spec.values())
+        use_unified = (model_uses_unified_kv_cache(self.runner.vllm_config)
+                       and not has_mamba_layers)
```
fenghuizhang, line 818: To play safe, we may want to disable use_unified if there are more than one cache dtypes. What do you think?

PR #3200, opened 2026-07-16
feat(rpa-v3): tuned 4-tuple block-size table (make the +48% decode split the default)

lokic233:
## Summary

Follow-up to #3102. Adds the **tuned block-size entry** kyuyeunk asked for, in the form that can actually hold the win.

#3102 exposes the decode fetch/compute split via an env override (self-contained kernel, no kernel changes). This PR makes the measured value the **default** — no env needed — by adding a v3-native tuned table and seeding it with the shape we measured on real tpu7x.

## Why a new table instead of an entry in `tuned_block_sizes.py`

The existing `tuned_block_sizes.py` stores a **2-tuple** `(num_kv_pages_per_block, num_queries_per_block)`, from which the kernel derives `bkv_sz == bkv_csz`. The +49% decode win comes precisely from a **fetch≠compute split** — a deep KV *fetch* block (`bkv_sz`) with a *smaller* compute block (`bkv_csz`) for deeper Pallas double-buffering. A 2-tuple entry has no slot for that; it would just reproduce today's default.

So this adds `tuned_block_sizes_v3.py`, a v3-native table keyed identically (reuses `get_lookup_keys`) but storing the full **4-tuple** `(bq_sz, bkv_sz, bq_csz, bkv_csz)`. The legacy 2-tuple table (consumed by the v2 / hd64 kernels) is **untouched**.

## Behavior

Lookup precedence in `get_default_block_sizes`:

```
caller override (d/p/m_block_sizes)  ->  v3 tuned table  ->  heuristic default
```

A table **miss returns `None`** and falls through to the existing heuristic, so no untuned shape is ever regressed — the table only ever *replaces* the heuristic for a shape measured on silicon.

## Seed entry

Qwen3-0.6B, DP16xTP4 (per-shard `q_head=4, kv_head=2, head_dim=128`, bf16 KV), tpu7x 4x4x4 decode:

| block sizes | tok/s | vs heuristic |
|---|---|---|
| heuristic default (`bkv_sz == bkv_csz == 16384`) | 64,929 | — |
| **`(1, 16384, 1, 4096)`** (deep fetch / small compute) | **96,292** | **+48%** |

Reproduced 4x (±0.5%); full inverted-U sweep confirms 4096 as the compute-block optimum.

## Testing

Device-free (runs on CPU CI). Asserts:
- every table entry passes the kernel's own `_validate_block_sizes` constraints (a malformed entry is caught at test time, not on silicon),
- a miss returns `None` (heuristic fallback),
- the seeded shape resolves under the exact production lookup key.

## Relationship to #3102

Independent — this touches `kernel.py` + the new table + a new test; #3102 touches `envs.py` + the attention layer + a different test. They don't overlap and can land in either order. #3102 is the minimal, self-contained exposure of the knob and is ready now; this PR is the "bake the proven value in as the default" step, split out so the small in-kernel table lookup can be weighed on its own.

---
_— drafted via [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_description&utm_campaign=tpu-inference&ref=lokic233) on behalf of @lokic233_

### DIFF
--- a/tests/kernels/ragged_paged_attention_v3_tuned_block_sizes_test.py
+++ b/tests/kernels/ragged_paged_attention_v3_tuned_block_sizes_test.py
@@ -1,0 +1,161 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Device-free tests for the RPA v3 tuned 4-tuple block-size table.
+
+These run on any host (including CPU CI): the table and its lookup are pure
+Python. Two things are guaranteed here:
+
+  1. Every entry satisfies the SAME constraints the kernel's caller-side
+     validator (`_validate_block_sizes`) enforces on operator overrides, so a
+     malformed tuned entry is caught at test time, never on silicon.
+  2. A miss returns None so the kernel falls back to the heuristic (an untuned
+     shape is never regressed), and the one seeded shape resolves to its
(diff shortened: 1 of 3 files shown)


kyuyeunk (commented):
i'm going to defer reviewing this because
1. change is too big
2. we are going to switch to batched rpa anyway.

PR #3202, opened 2026-07-16
Fix lm-eval accuracy jobs hanging on exit

theminghuang:
## Summary

- Construct the lm-eval vLLM adapter explicitly and always call `engine_core.shutdown()` in a `finally` block.
- Use the same lifecycle helper for both the pytest accuracy test and the shell-based `lm_eval` accuracy checks.
- Add regression coverage verifying shutdown runs after both successful evaluation and evaluation failure.

## Problem

Several Buildkite accuracy jobs finish all evaluation work but leave the Docker container running until the job timeout. The final output is either the lm-eval result aggregation message or a successful pytest summary, while the wrapper never reaches its post-Docker completion log.

The synchronous vLLM client relies on weak-reference/garbage-collection cleanup to stop its EngineCore subprocess. If the lm-eval model remains reachable through a reference cycle, the finalizer does not run and EngineCore keeps the container alive.

This change makes the accuracy entry points own the model lifecycle explicitly, so cleanup no longer depends on garbage-collection timing.

## Why this started appearing recently

The last known-good build and the first consistently affected build used the same pinned vLLM revision, which rules out the subsequent vLLM LKG update as the original regression. The only tpu-inference change between them was a multimodal ViT SDPA fix, while text-only Qwen3 and GPT-OSS jobs also hung.

The lifecycle flaw therefore predates the recent failures. CI images are rebuilt with `--no-cache` for each repository commit and resolve a mutable base image and unpinned transitive dependencies, so dependency/runtime drift or cache/network-dependent reference retention can expose the flaw without a corresponding engine change. The exact retaining object cannot be reconstructed because the historical images do not contain a recorded dependency manifest.

`lm-eval==0.4.12` is still the latest stable release, and its current vLLM adapter does not provide explicit engine cleanup.

## Testing

- `python3 -m py_compile scripts/vllm/integration/lm_eval_accuracy.py scripts/vllm/integration/test_accuracy.py tests/scripts/test_lm_eval_accuracy.py`
- `bash -n tests/e2e/check_lm_eval.sh`
- `git diff --check`
- Dependency-free mocked validation of cleanup after both success and exception

Full pytest was not run locally because the host Python environment does not have pytest installed.

### DIFF
--- a/scripts/vllm/integration/lm_eval_accuracy.py
+++ b/scripts/vllm/integration/lm_eval_accuracy.py
@@ -1,0 +1,110 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Run lm-eval with explicit ownership of the vLLM engine."""
+
+import argparse
+import json
+from typing import Any
+
+
+def _create_vllm_model(model_args: str | dict[str, Any], batch_size: str | int,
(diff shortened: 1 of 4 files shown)


theminghuang:

- #3202 ◀

wenxindongwork (approved):

QiliangCui (approved):

PR #3207, opened 2026-07-17
Optional block-diffusion (masked-diffusion) decode path

ethannnnnn:
## Summary

  Adds an optional **block-diffusion** (a.k.a. masked-diffusion / discrete-diffusion) decode path to the
  TPU engine, alongside the existing autoregressive (AR) decode.

  Block-diffusion LMs (e.g. Fast-dLLM, BD3-LM) generate a block of positions in parallel rather than one
  token at a time: the block is seeded with masked tokens and iteratively denoised, committing the
  high-confidence positions on each pass until the block is filled. Because multiple positions are
  committed per forward pass, far fewer *sequential* forward passes are needed for a given output length
  — the main latency win on decode-bound workloads.

  This PR is the **inference side only**. It is fully gated behind `enable_diffusion_decode` and is **off
  by default**; the AR decode path is unchanged.

  ## What's in this PR

  - **`use_causal_mask` plumbing** — an optional `use_causal_mask: bool = True` kwarg so a caller can
  request bidirectional (within-block) attention:
    - threaded through `attention()` / `sharded_ragged_paged_attention`
  (`layers/common/attention_interface.py`) and `Attention.attention`
  (`layers/jax/attention/attention.py`);
    - carried into the model `__call__` as a static `AttentionMetadata` field (default `True`), so the AR
  path stays byte-identical and only the diffusion caller flips it to `False` on the canvas;
    - the v3 ragged-paged-attention kernel already accepts the flag — this just forwards it. The hd64
  path (which lacks the param) keeps a guard.
  - **Threshold-commit sampler** — `diffusion_commit(logits, mask, threshold, temperature)` in
  `layers/jax/sample/sampling.py`: softmax → per-position top-token probability → commit positions above
  `threshold`, plus a forced arg-max of the highest-confidence still-masked position to guarantee forward
  progress. Pure JAX.
  - **Denoising decode loop** — `runner/diffusion_decode.py`: seeds a canvas (first token + `mask_id`),
  runs an on-device `jax.lax.while_loop` inner denoise per block and a host-side outer loop across blocks
  (block count is EOS/length-bounded), using the shifted-logit convention (token `i` predicted from
  hidden `i-1`). The model is injected as a `forward_fn`, so the loop is testable independently of the
  kernel.
  - **Config + dispatch** — `enable_diffusion_decode` / `diffusion_block_size` /
  `diffusion_commit_threshold` / `diffusion_max_denoise_steps` in the runner's `additional_config`, and a
  gated branch in `_execute_model` → `_execute_diffusion_decode`. `mask_id` is read from model config
  (not hardcoded).

  ## Design notes (TPU)

  - Bidirectional-within-block attention is expressed by running the v3 RPA kernel with
  `use_causal_mask=False` over `[prefix + canvas]`, so no log-sum-exp / `merge_state` combine is needed
  (unlike a paged-prefix + ragged-canvas FlashInfer path on GPU).
  - Injecting the model as a `forward_fn` keeps the denoise loop independent of both the specific model
  and the Pallas kernel — which is what makes the CPU stub tests possible.

  ## Back-compat / gating

  Entirely gated behind `enable_diffusion_decode` (default `False`). No change to the autoregressive
  decode path.

  ## Testing

  - **CPU unit tests**
    - `diffusion_commit`: threshold + forced-argmax progress + already-committed positions untouched (5
  tests).
    - Full denoise loop end-to-end with a stub `forward_fn`: full-commit, progress guarantee,
  shifted-logit indexing, multi-block seed chaining, EOS/max-len truncation (17 tests).
    - `use_causal_mask` threading and the config knobs verified via signature/AST checks.
  - **End-to-end on TPU** — with `use_causal_mask` threaded through the model `__call__`, the full path
  (bidirectional attention on the RPA kernel + on-device `_execute_diffusion_decode`) runs end-to-end on
  a real TPU against a block-diffusion Qwen3-0.6B checkpoint and produces coherent, on-topic output.
  Automated CI tests stay CPU-only because the RPA Pallas kernel and a checkpoint require TPU hardware
  not available in CI.

  ## Limitations / follow-ups

  - **Throughput** — the outer block loop is currently host-driven (one dispatch per denoise pass);
  fusing the block loop on-device into a single dispatch is a performance follow-up.
  - **Paged-cache K/V reuse** — denoise iterations currently overwrite the canvas K/V in the paged-cache
  tail each pass; writing K/V once per committed block (via an `update_kv_cache` flag threaded the same
  way) is a perf follow-up.
  - **Multi-block canvases** — `use_causal_mask=False` is fully bidirectional over `[prefix + canvas]`,
  which matches block-causal semantics only when exactly one block is on the canvas; multi-block canvases
  need a custom mask the binary flag can't express.
  - **Production serving throughput** — batched multi-request serving and the v1 scheduler/block-manager
  changes to reserve `block_size` KV slots per diffusion step (and advance `seq_len` by `block_size`) are
  tracked separately.

### DIFF
--- a/tests/layers/common/test_use_causal_mask_plumbing.py
+++ b/tests/layers/common/test_use_causal_mask_plumbing.py
@@ -1,0 +1,247 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Verifies that ``use_causal_mask`` is threaded through the attention stack.
+
+Actually executing the ragged-paged-attention (RPA) kernel requires a TPU
+(the Pallas kernel does not run on CPU), so these tests do NOT invoke the
+kernel. Instead they assert the flag is correctly threaded through every layer
+of the attention stack:
+
+  * declared as a parameter on the public entry points, and
+  * forwarded into the next call in the chain.
+
+The core checks parse the source with ``ast`` and therefore run on CPU without
+importing the heavy (TPU-only) dependency graph. An additional
+``inspect.signature`` check imports the real modules when possible and is
+skipped otherwise (e.g. when the TPU/vLLM deps are unavailable on CPU).
+"""
+
+import ast
+import inspect
+import pathlib
+
+import pytest
+
+_REPO_ROOT = pathlib.Path(__file__).resolve().parents[3]
+_ATTENTION_INTERFACE = (_REPO_ROOT / "tpu_inference" / "layers" / "common" /
+                        "attention_interface.py")
+_JAX_ATTENTION = (_REPO_ROOT / "tpu_inference" / "layers" / "jax" /
+                  "attention" / "attention.py")
+_ATTENTION_METADATA = (_REPO_ROOT / "tpu_inference" / "layers" / "common" /
+                       "attention_metadata.py")
+_QWEN3 = (_REPO_ROOT / "tpu_inference" / "models" / "jax" / "qwen3.py")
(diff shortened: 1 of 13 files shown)


Lumosis:
This PR is too large. Could you break it down into a few self-contained PRs?

ethannnnnn:
> This PR is too large. Could you break it down into a few self-contained PRs?

i will break the down diff with design doc for this work

PR #3208, opened 2026-07-17
support ipv6 and ipv4 parse

eating-chen:
# Description
This PR fixes a critical crash in tpu_connector.py when running on IPv6 env by safely formatting IPv6 addresses with brackets before passing them to the underlying JAX network transfer layer.    

- Why is this change being made?    
  - Currently, tpu_connector.py blindly formats IPs using f"{ip}:{port}". When deployed on an IPv6 cluster, passing a native raw IP like fd00::1 results in fd00::1:8200, which immediately crashes  jax.experimental.transfer.start_transfer_server and connect() with:
  ```                       
      jax.errors.JaxRuntimeError: INVALID_ARGUMENT: Encountered non-numeric characters while parsing port:          
      '0:0:15::7925:5000'    
  ```
- The problem being solved and relevant context:    
  - If we don't fix this inside tpu-inference, we force all downstream deployment repositories (e.g., llm-d) to inject messy if/else bash scripts into their Kubernetes YAML manifests. Like this [yaml setting](https://github.com/llm-d/llm-d/blob/main/guides/pd-disaggregation/modelserver/tpu/v7/vllm/patch-prefill.yaml).

- Why this is a good solution: 
  - I've created `format_ip_address_port` helper in `distributed/utils.py` that uses Python's ipaddress  module to automatically wrap IPv6 addresses in brackets while leaving IPv4 unchanged.

- Specific implementation details: 
  - Added format_ip_address_port to `tpu_inference/distributed/utils.py`.
  - Replaced the vulnerable f"{ip}:{port}" string interpolations across tpu_connector.py (in `_maybe_start_p2p_server`, `_maybe_build_kv_connection`).

Another Question: 
  A similar error might eventually be needed in jax_parallel_state.py for Pipeline Parallelism (init_pp_distributed_environment), as it uses the same underlying JAX transfer method.

# Tests

- Deployed a vLLM server on an IPv6 k8s cluster using a disaggregated prefill/decode topology (TPUConnectorHMA). 
- Verified that `_maybe_start_p2p_server` and `_maybe_build_kv_connection` successfully established the JAX transfer connections without the JaxRuntimeError.

# Checklist


tpu_inference/distributed/utils.py
```diff
@@ -63,6 +64,18 @@ def get_kv_transfer_port() -> str:
     return port
 
 
+def format_ip_address_port(ip: str, port: int | str) -> str:
```
richardsliu, line 67: Can we add a unit test for this?

richardsliu:
Consider adding a follow up change for https://github.com/vllm-project/vllm-torchtpu/tree/main/src/vllm_torchtpu/distributed/kv_transfer.

eating-chen:
> Consider adding a follow up change for https://github.com/vllm-project/vllm-torchtpu/tree/main/src/vllm_torchtpu/distributed/kv_transfer.
hi @richardsliu 
I can not access this link

tpu_inference/distributed/utils.py
eating-chen: @richardsliu 
I've added unit test in `test_distributed_utils.py`

sixiang-google (approved):

PR #3170, opened 2026-07-13
[BATCHED_RPA] Add SEQ_ON_LANE kv cache layout to remove padding when kv_heads = 1, kv d_type = fp8

a1yssan13:
# Description

In RPA, the kv_cache layout is packed along the number of KV heads. Thus, when `num_kv_heads = 1`, and kv dtype = fp8, we have to waste half the KV cache with padding. (k + v = 2, but 2 < kv_packing = 4). 

We allow the option to keep the cache in a layout -- seq_along_lane, with minimal performance degradation. In this layout, we put page_size on the last dimension (lane), and the head dimension on the sublane dimension.

Feature in progress to remove any overhead introduced by using `SEQ_ON_LANE` 

You can enable this via flag: `USE_BATCHED_RPA_SEQ_ON_LANE=1`

# Features 

With `SEQ_ALONG_LANE` enabled: 

1. You can now use `batched_rpa` with models that have`head_dim = 64`, since we no longer have to pad the head dimension up to 128. To be specific, SEQ_ALONG_LANE relaxes the constraint of having a head dimension that must be a multiple of 128 to only a multiple of 32. 
2. KV cache utilization is halved in models with the above characteristics (kv_heads per device = 1, kv_packing = 4)

# Constraints 

Since PAGE_SIZE is now on the last dimension of our cache, the configured block size must be 128 (for now). 

# Tests

All tests are run on Qwen-Coder. But any model that uses `BATCHED_RPA` should benefit. 

KV Cache capacity is doubled: 

`USE_BATCHED_RPA_SEQ_ON_LANE=0`:
```text
(EngineCore pid=2929407)   0.1.0: Array(shape=(5141, 128, 8, 4, 128), dtype=float8_e4m3fn, sharding=NamedSharding(mesh=Mesh('data': 1, 'model': 8, axis_types=(Auto, Auto)), spec=P(None, None, 'model'), memory_kind=device) [NamedSharding, id=123473401299648])
```

`USE_BATCHED_RPA_SEQ_ON_LANE=1`:
```text
(EngineCore pid=3053252)   0.1.0: Array(shape=(10283, 2, 256, 4, 128), dtype=float8_e4m3fn, sharding=NamedSharding(mesh=Mesh('data': 1, 'model': 8, axis_types=(Auto, Auto)), spec=P(None, None, 'model'), memory_kind=device) [NamedSharding, id=128670224217152])
```

[...]

PR #3214, opened 2026-07-17
[Batched_RPA] Fix/flash attn reshape

a1yssan13:
Minor bug that was introduced in the last PR #3170

### DIFF
--- a/tpu_inference/kernels/experimental/batched_rpa/flash_attention.py
+++ b/tpu_inference/kernels/experimental/batched_rpa/flash_attention.py
@@ -54,8 +54,8 @@ def flash_attention_qk_softmax(
     else:
         s = k.shape[-2]
         qk = lax.dot(
-            q.reshape(b, tq, h_size),
-            k.reshape(b, s, h_size),
+            q.reshape(-1, tq, h_size),
+            k.reshape(-1, s, h_size),
             dimension_numbers=(([2], [2]), ([0], [0])),
             preferred_element_type=jnp.float32,
         )


kyuyeunk (commented):
there were no tests to catch this? if not, can you add them?

kyuyeunk (approved):
approved for now because it's blocking and it's obvious fix. but please consider adding unit test in follow up.

lk-chen:
> approved for now because it's blocking and it's obvious fix. but please consider adding unit test in follow up.

@kyuyeunk it's because USE_BATCH_RPA flag is checked at [import time](https://github.com/vllm-project/tpu-inference/blob/86db923ea04a85cd51ea79d8222b0cec5eccced7/tpu_inference/layers/common/attention_interface.py#L53-L58) so it's tricky to use monkeypatch etc. to force using batch RPA during test. I used to try adding test (for Gemma4) but realized this blocker

kyuyeunk:
> > approved for now because it's blocking and it's obvious fix. but please consider adding unit test in follow up.
> 
> @kyuyeunk it's because USE_BATCH_RPA flag is checked at [import time](https://github.com/vllm-project/tpu-inference/blob/86db923ea04a85cd51ea79d8222b0cec5eccced7/tpu_inference/layers/common/attention_interface.py#L53-L58) so it's tricky to use monkeypatch etc. to force using batch RPA during test. I used to try adding test (for Gemma4) but realized this blocker

makes sense. can you create a pr that change this to be an argument of the kernel - and not have env checking be done inside the kernel?

lk-chen:
> > > approved for now because it's blocking and it's obvious fix. but please consider adding unit test in follow up.
> > 
> > 
> > @kyuyeunk it's because USE_BATCH_RPA flag is checked at [import time](https://github.com/vllm-project/tpu-inference/blob/86db923ea04a85cd51ea79d8222b0cec5eccced7/tpu_inference/layers/common/attention_interface.py#L53-L58) so it's tricky to use monkeypatch etc. to force using batch RPA during test. I used to try adding test (for Gemma4) but realized this blocker
> 
> makes sense. can you create a pr that change this to be an argument of the kernel - and not have env checking be done inside the kernel?

I think we can default to use batch RPA, given current status is quite mature?

kyuyeunk:
> > > > approved for now because it's blocking and it's obvious fix. but please consider adding unit test in follow up.
> > > 
> > > 
> > > @kyuyeunk it's because USE_BATCH_RPA flag is checked at [import time](https://github.com/vllm-project/tpu-inference/blob/86db923ea04a85cd51ea79d8222b0cec5eccced7/tpu_inference/layers/common/attention_interface.py#L53-L58) so it's tricky to use monkeypatch etc. to force using batch RPA during test. I used to try adding test (for Gemma4) but realized this blocker
> > 
> > 
> > makes sense. can you create a pr that change this to be an argument of the kernel - and not have env checking be done inside the kernel?
> 
> I think we can default to use batch RPA, given current status is quite mature?

that sounds good to me - as long as we have a proper fall-back mechanism for unsupported features (like attention sink).

PR #3216, opened 2026-07-17
Fix cpu_mesh() when no CPU backend exists (e.g. on Pathways)

andytwigg:
# Description

On backends like Pathways where only remote accelerators are registered, there is no local CPU backend, so jax.local_devices(backend=cpu) raises RuntimeError.

This change updates cpu_mesh() to return None if no CPU backend is available, and updates cpu_mesh_context() to return a nullcontext() in that case.

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/utils.py
+++ b/tpu_inference/layers/common/utils.py
@@ -168,13 +168,44 @@ def general_device_put(tensor: jax.Array,
     return jax.tree_util.tree_map(_put, tensor)
 
 
-def cpu_mesh() -> Mesh:
+_cpu_backend_available = None
+
+
+def _has_cpu_backend() -> bool:
+    """Check whether a CPU backend is registered in JAX (cached)."""
+    global _cpu_backend_available
+    if _cpu_backend_available is None:
+        try:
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/common/utils.py
```diff
@@ -168,13 +168,44 @@ def _put(t):
...
+def cpu_mesh():
+    """Return a single-device CPU mesh, or None if no CPU backend exists.
+
+    On backends like Pathways where only remote accelerators are registered,
+    there is no local CPU backend, so this returns None. Callers should use
+    cpu_mesh_context() which handles None gracefully.
+    """
     global _cpu_mesh
     if _cpu_mesh is None:
-        _cpu_mesh = Mesh(jax.devices("cpu")[:1], ("cpu", ))
```
lk-chen, line 174: it doesn't do harm to have
```
try:
  _cpu_mesh = Mesh(...)
except RuntimeError:
  return None
return _cpu_mesh
```
inside `cpu_mesh()` right? code looks simpler this way

PR #3220, opened 2026-07-19
[Gemma4][SWA] Enable sliding-window KV cache via flat shared-pool

prashant182:
## Summary

Enables hybrid **sliding-window KV cache for Gemma-4** on the TPU `flax_nnx` path. It is
currently force-disabled for Gemma-4 (`kv_cache_manager.get_kv_cache_spec` hardcodes
`sliding_window = None`, TODO(kwang3939)) because the model's sliding and global attention layers
have **heterogeneous KV shapes** (sliding `num_kv_heads=16, head_dim=256`; global
`num_kv_heads=4, head_dim=512`), and a period-tensor's `shared_by` layers therefore cannot back
onto a single typed `jax.Array`.

Without windowing, every layer keeps the full context, so a v6e-4 fits only ~2-3 concurrent
32K-context sequences. With windowing the 50 sliding layers cap at `sliding_window` tokens, which
is a large decode-concurrency win for long-context serving.

## Approach

The key observation: after `unify_kv_cache_spec_page_size` scales `block_size` (sliding 16 vs
global 32), **every layer's per-chip page is the same size** (e.g. `16·4·2·256 == 32·1·2·512 ==
32768` elem/chip on TP=4) — same bytes, different shape.

So the heterogeneous `shared_by` layers share **one native 5D KV pool** allocated in the first
shared layer's shape (`num_blocks = kv_cache_config.num_blocks`, no extra memory). In
`sharded_ragged_paged_attention`, each layer computes its own target 5D layout from its local `k`
and, **only if it differs from the buffer shape**, reshapes the buffer to its layout *inside the
`shard_map`* (a free per-chip reshape, no cross-device movement), runs the kernel, and reshapes
the updated cache back so the `out_spec` still matches. The pool stays native 5D, so its rank and
sharding match the rest of the runner. For homogeneous models the target equals the buffer shape
and the path is a no-op.

Two files change (+75/-6):
- `runner/kv_cache_manager.py`: honor `sliding_window` for sliding layers; stop pinning
  `page_size_padded` to the per-spec size so the hybrid page unification runs.
- `layers/common/attention_interface.py`: the in-`shard_map` reshape.

No kernel changes.

## Results (v6e-4, TP=4, Gemma-4-31B bf16, `max_model_len=32768`)

- **Correctness**: coherent short output and a 20K-token needle-in-a-haystack recall with real
  `sliding_window=1024` (no SparseCore fault).
- **Concurrency**: 64 concurrent 24K-in/1500-out requests all resident at ~78% KV usage
  (limited by `max_num_seqs`, not HBM), vs ~2-3 without windowing.

## Known limitations / for discussion

- The KV-offload / transfer helpers (`_jitted_insert*/_gather*`) still assume the per-layer 5D
  shape and are not updated here; they are not exercised on the standard prefill/decode path but
  would need the same reshape for a complete solution.
- Assumes the DCP/CONTEXT axis is 1 for the shared-pool case (block_size is folded into the page).
- Would appreciate review from the TPU team on the sharding/reshape approach and whether a
  first-class hybrid-shared-pool abstraction is preferable to the per-call reshape.

Happy to iterate. This unblocks correct long-context Gemma-4 serving on TPU today.


lk-chen:
Hi, what command did you test with? I used
```
VLLM_ENGINE_READY_TIMEOUT_S=6000 \
MOE_REQUANTIZE_WEIGHT_DTYPE=float8_e4m3fn \
USE_BATCHED_RPA_KERNEL=1 \
MODEL_IMPL_TYPE=flax_nnx \
examples/external_lb/run_dp_external_lb.sh serve google/gemma-4-26B-A4B-it \
    --max-model-len=1524 \
    --max-num-seqs=818 \
    --data-parallel-size 4 \
    --tensor-parallel-size 2 \
    --no-enable-prefix-caching \
    --additional_config='{"quantization": { "qwix": { "rules": [{ "module_path": ".*", "weight_qtype": "float8_e4m3fn", "act_qtype": "float8_e4m3fn"}]}}, "compilation_sizes": [818]}' \
    --kv-cache-dtype=fp8 \
    --gpu-memory-utilization=0.9 \
    --async-scheduling \
    --limit-mm-per-prompt '{"image": 0, "video": 0, "audio": 0}' \
    --block-size=256
```

and hit 
```
ValueError: Expected kv_cache.shape=(29454, 256, 2, 4, 256) to be equal to
  expected_kv_cache_shape=(29454, 256, 1, 4, 512)
```

prashant182:
PR updated — pushed the second file (apologies for not including it earlier). Here's how I ran it:
```
vllm serve <gemma-4-31b-it> --tensor-parallel-size 4 --max-model-len 32768 \
  --num-gpu-blocks-override 10000 --max-num-seqs 64 --max-num-batched-tokens 8192 \
  --limit-mm-per-prompt '{"image":0}' --disable_chunked_mm_input
```
Dense 31B, bf16, single v6e-4, default (v3) kernel — no batched kernel, no fp8, no `--block-size`.

PR #3230, opened 2026-07-21
Update jax to 0.11.0, libtpu to 0.0.44 and flax

patrickji2014:
# Description

Update dependencies version. 

# Tests

Rely on CI Test.
Nightly test https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23152/canvas
This is a reference nightly test from main: https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23149/canvas?sid=019f92ec-ab66-4584-b94e-90ee65268cb5&tab=output
Note: TPU 7x Qwen Perf Regression Test already failed in main branch's nightly test, other failed nightly test also happens in the reference nightly test.

# Checklist


lk-chen:
FYI https://github.com/vllm-project/tpu-inference/pull/3232

patrickji2014:
> FYI #3232
thanks, i am updating similar items as in new flax version, the sharding is not setable to metadata. param.sharding always refer to the underlying jax array's sharding.

tests/layers/jax/test_linear.py
```diff
@@ -146,9 +147,13 @@ def test_parameter_names_match_vllm_unquantized(self):
                 for k, v in vllm_mlp.named_parameters()
             })
 
-    def test_sharding_assignment(self):
+    @parameterized.expand([('eager_sharding_on', True),
```
lk-chen: we could always assume eager_sharding=False

patrickji2014: I update the test for testing non eager sharding mode only

lk-chen (approved):

tpu_inference/models/common/model_loader.py
```diff
@@ -192,7 +192,7 @@ def create_jit_model(
                 mesh,
                 apply_to_abstract_model=True)
 
-            model = nnx.eval_shape(abstract_model_fn)
+            model = nnx.eval_shape(abstract_model_fn, graph_updates=False)
```
jrplatin, line 195: Can you explain why this is needed?

jrplatin (approved):

tpu_inference/models/common/model_loader.py
patrickji2014: talked to Flax team, without it flax eval_shape will eagerly update Param's sharding via using mesh and spec. since in our case we want to load the param into cpu and do quantization and shard it later, the mesh is set to cpu mesh, which is not compatible with the spec info in the param. this cause error in flax when it try to build the sharding via NamedSharding(mesh, spec).

kyuyeunk (approved):

mrjunwan-lang (approved):

PR #3231, opened 2026-07-21
[RL] Add vLLM-native weight transfer API for RL weight sync

wenxindongwork:
# Description

This PR adds weight-update API vLLM standardized on: `init_weight_transfer_engine`, `start_weight_update`, `update_weights`, `finish_weight_update` to `TPUWorker`, so RL frameworks and orchestrators can drive a TPU sampler through a standard interface instead of a bespoke entry point.

## Usage

```python
llm.reset_prefix_cache()                      # scheduler-side; caller's job
llm.start_weight_update()                     # frees the KV cache
llm.update_weights()    
llm.finish_weight_update()                    # reallocates the KV cache
```

On TPU, update_weights is currently an no-op. Under Raiden, RL trainer transfers weights directly into the sampler's PJRT buffer (push mode). Under Pathways reshard, RL trainer directly modifies sampler weights. 

# Tests

Added unit tests.

# Checklist

### DIFF
--- a/tpu_inference/worker/tpu_worker.py
+++ b/tpu_inference/worker/tpu_worker.py
@@ -10,6 +10,7 @@ from typing import Any, Callable, Dict, Optional, Tuple
 import jax
 import jaxlib
 import jaxtyping
+import vllm.envs as vllm_envs
 from vllm.config import VllmConfig, set_current_vllm_config
 from vllm.distributed.kv_transfer import (ensure_kv_transfer_initialized,
                                           has_kv_transfer_group)
@@ -177,6 +178,9 @@ def _parse_profile_options(
 
 
 class TPUWorker(WorkerBase):
+
+    _weight_update_active: bool = False
+    _kv_cache_freed: bool = False
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


tpu_inference/worker/tpu_worker.py
```diff
@@ -733,6 +737,64 @@ def sync_weights(
...
+        """
+        if not self._weight_update_active:
+            raise RuntimeError(
+                "start_weight_update must be called before update_weights.")
+
+        if not vllm_envs.VLLM_TPU_USING_PATHWAYS:
+            return
+
+        try:
+            self.model_runner._sync_weights(
```
wang2yn84: I think we don't really use this path any more in Tunix. The initial design was to embed weight sync in vLLM. But later on we handles in Tunix with the exposed vLLM model weights: https://github.com/google/tunix/blob/7a7d9ea129023d0c900b892c11929dd970987cc1/tunix/generate/vllm_sampler.py#L207

tpu_inference/worker/tpu_worker.py
```diff
@@ -733,6 +737,64 @@ def sync_weights(
...
+        logger.info("init_weight_transfer_engine: %s", sorted(init_info or {}))
+
+    def start_weight_update(self, free_kv_cache: bool = True) -> None:
+        """Open a weight update session.
+
+        Freeing the KV cache is on by default: an incoming set of weights
+        roughly doubles peak HBM, and decoding from KV computed under the old
+        weights is incorrect. `finish_weight_update` reallocates it.
+
+        The prefix cache is scheduler-side and cannot be dropped here, so
```
wang2yn84, line 739: @khatwanimohit let's clear prefix cache from sampler.

tpu_inference/worker/tpu_worker.py
wenxindongwork: sounds like it is safe to delete the _sync_weights function in model_runner.py. if so, i will go ahead and delete it.

tpu_inference/worker/tpu_worker.py
```diff
@@ -733,6 +737,64 @@ def sync_weights(
...
+        callers must `reset_prefix_cache()` first -- otherwise the scheduler
+        serves hits from blocks this frees.
+        """
+        if self._weight_update_active:
+            raise RuntimeError(
+                "start_weight_update called while an update is already "
+                "active. Call finish_weight_update first.")
+        if free_kv_cache:
+            self.model_runner.delete_kv_cache()
+            self._kv_cache_freed = True
```
khatwanimohit: in the event if we pass free_kv_cache=false `self._kv_cache_freed` would be having previous stale value. We can set `self._kv_cache_freed = free_kv_cache` instead

khatwanimohit (approved):

tpu_inference/worker/tpu_worker.py
wang2yn84: Yes please.

sixiang-google (approved):

PR #3232, opened 2026-07-21
Propagate sharding metadata from Flax NNX correctly.

niting:
This PR fixes the propagation of sharding metadata from Flax NNX to weight loaders and quantization configurations in the JAX path.

Flax NNX deprecated the `"sharding"` metadata key in favor of `"out_sharding"`. While `weight_utils.py` was updated to use `"out_sharding"` in #3004, several other files still attempt to read from `"sharding"`, which returns empty and causes weights to be replicated during loading (leading to TPU HBM OOMs).

This PR updates the remaining occurrences to read from `"out_sharding"`.

**Affected files:**
*   `tpu_inference/models/common/pathways_dummy_loader.py`: Fixes dummy weight loading via Pathways.
*   `tpu_inference/layers/jax/quantization/configs.py`: Fixes quantization config weight sharding.
*   `tpu_inference/layers/jax/quantization/unquantized.py`: Fixes merged linear layers sharding retrieval.

### DIFF
--- a/tpu_inference/layers/jax/quantization/configs.py
+++ b/tpu_inference/layers/jax/quantization/configs.py
@@ -190,7 +190,7 @@ class QuantLinearConfig(CommonQuantLinearConfig):
         # Update weight_sharding and bias_sharding for 2D matmul compatibility
         if self.batch_features:
             self.weight_sharding = _to_partition_spec(
-                weight.get_metadata().get("sharding", ()))
+                weight.get_metadata().get("out_sharding", ()))
         else:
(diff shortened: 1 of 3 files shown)


lk-chen:
we can remove `fallback to "sharding"` as the current code is not compatible with previous version of jax

niting:
Done. I removed the fallback.

lk-chen (approved):

lk-chen:
LGTM, plz fix DCO

wenxindongwork (approved):

PR #3233, opened 2026-07-21
Add microbenchmarks for the collective-matmul kernels

guowei-dev:
Two benchmark scripts under scripts/benchmarking/kernels/ measure the auto-partitioned XLA baseline for the Llama-70B tensor-parallel collective-matmul patterns —
all-gather + matmul and matmul + reduce-scatter — at the MLP projection shapes (K=8192, N=57344), sweeping the token dimension across the serving padding buckets. Each reports mean per-call latency and per-core TFLOP/s; timing is wall-clock over dispatched calls blocked once, which tracks the profiler's module device time within a few percent. Pass --profile-dir to capture an xprof trace.

A shared collective_bench_lib holds the mesh, per-call timing, and sweep/report driver, so each script is just its collective and an IMPLEMENTATIONS registry. A fused kernel plugs in with a single entry and is benchmarked against the baseline; this version registers only the baseline and depends on no kernel.

# Description

the past.

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/scripts/benchmarking/kernels/collective_bench_lib.py
+++ b/scripts/benchmarking/kernels/collective_bench_lib.py
@@ -19,8 +19,11 @@ shardings, the einsum, and an IMPLEMENTATIONS registry. Not runnable on its own.
 """
 
 import argparse
-import contextlib
-import time
+import glob
+import os
+import shutil
+import statistics
+import tempfile
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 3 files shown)


scripts/benchmarking/kernels/collective_bench_lib.py
```diff
@@ -0,0 +1,148 @@
...
+                        help="comma-separated token counts to sweep")
+    parser.add_argument("--k", type=int, default=8192, help="hidden dim")
+    parser.add_argument("--n", type=int, default=default_n)
+    parser.add_argument("--dtype", default="bfloat16")
+    parser.add_argument("--reps",
+                        type=int,
+                        default=100,
+                        help="dispatched calls per cell (blocked once)")
+    parser.add_argument("--warmup", type=int, default=5)
+    parser.add_argument(
```
yaochengji: I think the kernel should be able to figure out the physical order themselves, which would make it easier for the kernel integration.

scripts/benchmarking/kernels/collective_bench_lib.py
```diff
@@ -0,0 +1,148 @@
...
+        "--device-order",
+        default=None,
+        help="comma-separated device-index permutation for the "
+        "mesh; default is natural order (order-sensitive ring "
+        "kernels want a ring-friendly permutation)")
+    parser.add_argument(
```
yaochengji, line 124: We should use the latency extracted from trace by default, just remember to remove the overhead of `barrier-core`

scripts/benchmarking/kernels/collective_bench_lib.py
guowei-dev: I see, delegated to the future kernel and removed the order from build_mesh

scripts/benchmarking/kernels/collective_bench_lib.py
guowei-dev: Sure, updated in device_time. Now it reports [median(per-call XLA-module device time) - median ( barrier-cores span)] from the tracer.

yaochengji: How about in every call, let's do `XLA-module device time - barrier_cores`?

guowei-dev: That makes more sense. Added a matching logic there and now [median(per-call XLA-module device time - barrier-cores span)]

QiliangCui (approved):

PR #3234, opened 2026-07-22
Apply fp8 on ragged gather

dawnhan1111:
# Description

Keep MoE activations in fp8 through the expert-parallel ragged gather

Why this change is being made

On the expert-parallel MoE path, activations are quantized to fp8 for the ICI all-gather (MOE_ALL_GATHER_ACTIVATION_DTYPE=fp8), but then dequantized back to bf16 right after the `all-gather` — before the SparseCore ragged gather (the expert permute) — and GMM1 subsequently re-quantizes the top-k-expanded activations to fp8 internally. That's two conversions that cancel, plus a ragged gather that moves bf16 (2 bytes/elem) instead of fp8 (1 byte/elem).

This change keeps activations in fp8 all the way into GMM1, deleting both the dequant and GMM1's internal re-quant.

The problem being solved, and why it's non-trivial

An fp8 tensor is two things: the codes and a per-token scale. To defer dequant past the ragged gather, the scale has to stay aligned with its row through the permute. The obvious implementations all cost more than the fp8 conversion saves:
- a second gather for the scale serializes against the activation gather on SparseCore;
- a separate scale array forces layout/re-tile copies that outweigh the arithmetic.

The whole engineering problem is getting the scale to ride along for free.

Why this is a good solution

Pack the scale into the codes' trailing columns, permute both in one gather, and let the kernel read the scale out of the buffer it already holds. Four properties make it cheap:

- A bf16 is two fp8-typed bytes. bitcast_convert_type reinterprets the bits; nothing is converted. The gather moves bytes and doesn't care what they mean.
- The gather is row-bound, not byte-bound. Widening each row from hidden to the next 128-multiple costs ~nothing, so the scale rides inside the same permute.
- GMM1 contracts size_k (which comes from the weights), so handing it a physically wider lhs is free — it never reads past the codes. No slice, no re-tile.
- The scale is constant across k, so applying it once to the fully-reduced accumulator is exact. It lands before the bias and activation (scaling across a nonlinearity is not equivalent).

Result: one all-gather instead of two, one ragged gather instead of two, no host-side dequant, no re-tile.

# Test 

Measured on Kimi-K2.6 (MLA attention, fp4 MoE), tpu7x-8, 8-way expert parallel, 544 prompts / 8192 in / 1024 out:

- +1.87% total throughput vs the same commit with the flag off (2 interleaved pairs: +1.98%, +1.76%; baseline reproduced across sessions to 0.04%).
- Quality unchanged: greedy output bit-identical, MMLU indistinguishable (0.87675 both arms).

Serve:
QUANTIZE_ON_LOAD_PREFIXES="self_attn" DP_SCHED_BATCH_PREFILL=0 \
MOE_REQUANTIZE_WEIGHT_DTYPE=fp4 MOE_ALL_GATHER_ACTIVATION_DTYPE=fp8 \
MOE_REQUANTIZE_BLOCK_SIZE=512 MODEL_IMPL_TYPE=vllm \
MLA_TRANSPOSE_KV_CACHE=1 NEW_MODEL_DESIGN=1 \
MOE_RAGGED_GATHER_FP8=1 \
vllm serve \
  --model "gs://tpu-commons-ci/moonshootai/kimi/2.6" \
  --served-model-name="moonshotai/Kimi-K2.6" \
  --max-model-len=9216 --max-num-batched-tokens=1024 --max-num-seqs=68 \
  --no-enable-prefix-caching --gpu-memory-utilization=0.98 \
  --tensor-parallel-size=8 --async-scheduling --enable-expert-parallel \
  --kv-cache-dtype=fp8 --seed=42 \
  --limit-mm-per-prompt='{"image": 0, "video": 0, "vision_chunk": 0}' \
  --additional-config '{"compilation_sizes": [544], "sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' \
  --generation-config="vllm" \
  --override-generation-config='{"do_sample": false, "temperature": 0.0, "model": "moonshotai/Kimi-K2.6"}' \
  --trust_remote_code 2>&1 | tee ~/kimi_tp_moe.log

Baseline (original path — flag OFF)

Identical, just flip the one flag:

... MOE_RAGGED_GATHER_FP8=0 ...

Bench:

vllm bench serve --model moonshotai/Kimi-K2.6 --dataset-name random \
  --random-input-len 8192 --random-output-len 1024 \
  --num-prompts 544 --ignore-eos --skip-chat-template --trust-remote-code

variant (fp8 — flag ON) — MOE_RAGGED_GATHER_FP8=1

## Serving Benchmark Result 
Successful requests:                     544
Benchmark duration (s):                  254.41
Total input tokens:                      4456448
Total generated tokens:                  557056
Request throughput (req/s):              2.14
Output token throughput (tok/s):         2189.63
Peak output token throughput (tok/s):    7600.00
Peak concurrent requests:                544.00
Total token throughput (tok/s):          19706.65
Mean TTFT (ms):                          97389.12
Median TTFT (ms):                        87410.16
P99 TTFT (ms):                           209323.15
Mean TPOT (ms):                          116.54
Median TPOT (ms):                        125.44
P99 TPOT (ms):                           155.22
Mean ITL (ms):                           116.54
Median ITL (ms):                         54.63
P99 ITL (ms):                            329.23

http://xprof.corp.google.com/trace_viewer/dawnhan-5437574570183876168

baseline (original — flag OFF) — MOE_RAGGED_GATHER_FP8=0

## Serving Benchmark Result 
Successful requests:                     544
Benchmark duration (s):                  259.46
Total input tokens:                      4456448
Total generated tokens:                  557056
Request throughput (req/s):              2.10
Output token throughput (tok/s):         2147.01
Peak output token throughput (tok/s):    7200.00
Peak concurrent requests:                544.00
Total token throughput (tok/s):          19323.12
Mean TTFT (ms):                          99238.52
Median TTFT (ms):                        89022.10
P99 TTFT (ms):                           213407.45
Mean TPOT (ms):                          118.82
Median TPOT (ms):                        128.02
P99 TPOT (ms):                           158.32
Mean ITL (ms):                           118.82
Median ITL (ms):                         55.84
P99 ITL (ms):                            334.95

http://xprof.corp.google.com/trace_viewer/dawnhan-18152971302305309911

### DIFF
--- a/tpu_inference/kernels/megablox/gmm_v2.py
+++ b/tpu_inference/kernels/megablox/gmm_v2.py
@@ -23,6 +23,11 @@ from jax import lax
 from jax.experimental import pallas as pl
 from jax.experimental.pallas import tpu as pltpu
 
+# Lane width of the tile the packed lhs scale is read from. A block narrower
+# than the lane width is not expressible, so the packed path loads this many
+# columns and slices the two transport bytes out in-register.
+_SCALE_TILE_LANES = 128
+
 # Util.
 
 
@@ -185,6 +190,26 @@ class FusedWeightsRef(RhsRef):
 
 @jax.tree_util.register_dataclass
 @dataclasses.dataclass(frozen=True)
+class LhsRef:
+    """Dataclass for lhs, optionally carrying a per-token dequant scale.
+
+    When ``scale`` is None, ``data`` is the (unquantized) lhs and the kernel
+    quantizes it internally, exactly as before. When ``scale`` is not None,
+    ``data`` is assumed to already be quantized (e.g. to fp8) and ``scale``
+    holds the per-token multiplier applied to the accumulator after the matmul
+    (the scale is constant across k, so it factors out of the k reduction).
+    """
+
+    data: Any
+    scale: Any | None = None
+
+    def get_scale(self) -> jax.Array:
+        assert self.scale is not None
+        return self.scale[...]
+
+
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass(frozen=True)
 class MetadataRef:
     gm_id_to_group_id: jax.Array
     gm_id_to_m_offset: jax.Array
@@ -215,6 +240,11 @@ class InputConfigs:
     dtype: jnp.dtype
     has_bias: bool = False
     has_scale: bool = False
+    # lhs only: data is already quantized and its per-token scale rides in the
+    # lhs buffer's trailing columns (two fp8-typed transport bytes at column
+    # size_k). Skip internal quantization, read the scale from the buffer the
+    # kernel already holds, and apply it after the matmul.
+    prequantized: bool = False
 
     @property
     def should_dequantize_before_matmul(self) -> bool:
@@ -280,6 +310,20 @@ class IndexMaps:
 
         return (pl.ds(row_start, row_size), 0, k_id)
 
+    def lhs_scale_index_map(self, _: jax.Array, gm_id: jax.Array,
+                            __: jax.Array):
+        # Scale rides in the lhs buffer itself, in the lane tile that starts at
+        # column size_k. Same row tiling as lhs; independent of k.
+        m_start = self.metadata_ref.gm_id_to_m_offset[gm_id]
+        m_end = self.metadata_ref.gm_id_to_m_offset[gm_id + 1]
+
+        row_start = m_start // self.cfgs.dims.size_lhs_sublane
+        row_end = pl.cdiv(m_end, self.cfgs.dims.size_lhs_sublane)
+        row_size = row_end - row_start
+
+        return (pl.ds(row_start,
+                      row_size), 0, self.cfgs.dims.size_k // _SCALE_TILE_LANES)
+
     def rhs_weight_index_map(self, n_id: jax.Array, gm_id: jax.Array,
                              k_id: jax.Array):
         group_id = self.metadata_ref.gm_id_to_group_id[gm_id]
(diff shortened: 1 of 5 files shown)


tpu_inference/kernels/megablox/gmm_v2.py
```diff
@@ -183,6 +188,26 @@ def get_bias(self) -> jax.Array:
         return interleave_lane(b_gate, b_up)
 
 
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass(frozen=True)
+class LhsRef:
+    """Dataclass for lhs, optionally carrying a per-token dequant scale.
+
+    When ``scale`` is None, ``data`` is the (unquantized) lhs and the kernel
```
bigeye, line 196: It would be more straightforward if we present this in a shape. We have [M, K], and expect [M, 1]?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -681,36 +710,134 @@ def _process_tokens_locally(hidden_states_local, topk_indices_local):
...
         return x, group_sizes_local, topk_argsort_revert_indices
 
-    if all_gather_fp8:
-        hidden_states = _apply_all_gather_fp8(hidden_states, mesh, dtype)
+    def _process_tokens_locally_fp8(combined_local, topk_indices_local):
+        # fp8 ragged-gather path: pack the per-token scale into fp8-typed
+        # transport bytes appended to the fp8 activations and permute both in a
+        # SINGLE SparseCore gather. The gather is row-bound, so the few extra
+        # columns are ~free, and one gather avoids serializing a second gather
+        # on the SparseCore.
```
bigeye, line 736: I agree that having a packed scale as a extra column would be cheap.

But I expected the second gather really cheap, because we're only talking about doing a gather on 1D array [M, 1]. Isn't this negligible enough?

If this is negligible, just passing them separately would make the code clearer and more straightforward. We may need to implement subchannel in the future for example, then this packing/unpacking logic would become much more complicated.

So first, could you provide more data points about the second gather?

PR #3238, opened 2026-07-22
Fix RaidenOffloadConnector for single-RaidenBlockID KVCacheStore API

yiw-wang:
Fix RaidenOffloadConnector for single-RaidenBlockID KVCacheStore API
  
## Summary
`tpu_raiden`'s `KVCacheStore` was updated to map a **single** `RaidenBlockID` per block hash (previously a `list[list[RaidenBlockID]]`, one hash → many replicas; a single record now carries both host and device residency). The `RaidenOffloadConnector` was still written against the old list-of-list API, so the offload path crashed at runtime as soon as a save completed:
 
```
  File ".../tpu_raiden/api/jax/kv_cache_store.py", in insert 
      raw_slices.append(s._impl)
  AttributeError: 'list' object has no attribute '_impl'
```
The crash killed EngineCore, which surfaced downstream as the benchmark client failing to connect to the server. And this is the reason why we see following error in our daily release test against HEAD.

```
Traceback (most recent call last):
 ...
  File "/home/yiweiw_google_com/raiden_release_test/tpu-raiden/.venv312/lib/python3.12/site-packages/aiohttp/connector.py", line 1324, in _wrap_create_connection
    raise client_error(req.connection_key, exc) from exc
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host [127.0.0.1:8000](http://127.0.0.1:8000/) ssl:default [Connect call failed ('127.0.0.1', 8000)]
```

## Changes

Aligned the three call sites that consumed the old nested shape:
  
- **`insert()`**: pass a flat `[locator.to_raiden_id()]` (one id per hash) instead of the nested `[[...]]`.
- **eviction handling**: each evicted entry is now `(hash, RaidenBlockID)`, so drop the inner per-slice loop and read the chunk id off the single `RaidenBlockID` (its `data_replica_idx` convenience property delegates to the wrapped `RaidenId`).
- **`lookup()` load path**: `matched[i]` now yields a single `RaidenBlockID`, so unpack it directly instead of indexing `[0]`.

## Testing
Ran the `kv_host_offloading` example end-to-end on tpu7x-8 (Qwen3-32B, tp=8) against `tpu_raiden` built from current `main`:

- **Before**: first request → `AttributeError` → EngineCore dead → 500s.
- **After**: 275 requests across two runs, **0 failures**, no `AttributeError`, clean shutdown. The offload store is populated and lookups hit (metrics show hundreds of successful inserts), confirming the fixed insert path runs repeatedly.

### DIFF
--- a/tests/offload/test_raiden_offload_connector.py
+++ b/tests/offload/test_raiden_offload_connector.py
@@ -87,13 +87,17 @@ def create_request(request_id: str, num_tokens: int,
 
 
 def make_matched(num_blocks: int, start_chunk: int = 100):
-    """Build a kv_store.lookup() result: list[(hash_bytes, [RaidenId])]."""
-    return [(f"hash_{i}".encode(), [
-        MockRaidenId(job_name="tpu_inference",
-                     job_replica_id="0",
-                     data_name="kv_cache",
-                     data_replica_idx=start_chunk + i)
-    ]) for i in range(num_blocks)]
+    """Build a kv_store.lookup() result: list[(hash_bytes, RaidenBlockID)].
+
+    KVCacheStore maps a single RaidenBlockID per block hash (not a list of
+    replicas), so each entry's second element is one id, not a list.
+    """
+    return [(f"hash_{i}".encode(),
+             MockRaidenId(job_name="tpu_inference",
+                          job_replica_id="0",
+                          data_name="kv_cache",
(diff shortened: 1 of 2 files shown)


juncgu-google (approved):
LGTM.

QiliangCui (approved):

PR #3239, opened 2026-07-22
Incremental weight loading and sharding for FP8 MoE models

BirdsOfAFthr:
# Description

This PR introduces an **Incremental FP8 Weight Streaming Loader** (`VLLM_INCREMENTAL_FP8_LOADING`) for FP8 MoE models.

### Context & Problem
Under standard loading for FP8 models, vLLM loads and retains all checkpoint shards in host memory simultaneously and then shard them to TPUs. 

TPU8i has strict host RAM limits (**~384 GB RAM limit**). As a result, standard loading triggers host Out-Of-Memory (OOM) crashes or severe memory pressure when initializing large FP8 MoE models.

### Solution & Implementation
This PR adds support for layer-by-layer incremental weight processing and sharding in FP8 MoE layers (`VllmFp8MoEMethod`):
- `VllmFp8MoEMethod` now inherits from `VllmQuantizationMethod` and implements `maybe_process_weights` and `process_weights_after_loading`.
- Weight parameters for each MoE layer are converted into JAX arrays as soon as their respective safetensor shards finish loading.
- PyTorch CPU tensor storage for each layer is immediately freed via `untyped_storage().resize_(0)`, `delattr`, `gc.collect()`, and `malloc_trim(0)`.
- Drops PyTorch CPU tensor memory usage, keeping peak host RAM safely beneath the host RAM limit.

---

# Tests

Verified on TPU VM (TPU v7x-8, 4 TPU chips) using `Qwen/Qwen3.5-397B-A17B-FP8`:

### 1. Host Memory & Weight Load Time Benchmarks
| Metric / Resource | Standard Baseline (`VLLM_INCREMENTAL_FP8_LOADING=False`) | Incremental FP8 Streaming (`VLLM_INCREMENTAL_FP8_LOADING=True`) | Trade-off / Savings |
| :--- | :--- | :--- | :--- |
| **Peak Host RAM (RSS)** | **418.95 GB** *(Exceeds 384GB limit)* | **219.40 GB** *(Safely within host RAM)* |  **~200 GB RAM Savings (~47.6% Peak Reduction)** |
| **Max PyTorch CPU Memory** | **314.40 GB** | **12.8 GB** | **-301.6 GB PyTorch CPU Storage Saved** |
| **Model Weight Load Time** | **431.77 s** (~7.2 min) | **652.11 s** (~10.8 min) |  **+3.6 min overhead for layer-by-layer CPU freeing** |

### 2. Quality & Accuracy Evaluation Results
| Benchmark | Standard Baseline (VLLM_INCREMENTAL_FP8_LOADING=0) | Incremental FP8 Streaming (VLLM_INCREMENTAL_FP8_LOADING=1) |
| :--- | :--- | :--- |
| GPQA Diamond (CoT 0-shot) | 7.58% (0.0758 ± 0.0189) | 8.59% (0.0859 ± 0.0200) |
| MMLU-Pro (5-shot, 700 sample) | 80.14% (0.8014 ± 0.0147) | 81.29% (0.8129 ± 0.0144) |
---
# Checklist

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -71,6 +71,7 @@ if TYPE_CHECKING:
     MIN_TOKEN_BUCKET: int = 16
     MOE_ROUTE_PADDING_TO_EXPERT0: bool = False
     VLLM_TPU_BUCKET_PADDING_GAP: int = 0
+    VLLM_INCREMENTAL_FP8_LOADING: bool = False
     TPU_MESH_SORT_BY_COORDS: bool = False
 
 
@@ -440,6 +441,9 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Currently, it only supports a single host set up.
     "TPU_MESH_SORT_BY_COORDS":
     env_bool("TPU_MESH_SORT_BY_COORDS", default=False),
+    # Controls whether FP8 layers perform incremental weight loading and sharding.
+    "VLLM_INCREMENTAL_FP8_LOADING":
+    env_bool("VLLM_INCREMENTAL_FP8_LOADING", default=False),
 }
 
 
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -441,7 +441,12 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Currently, it only supports a single host set up.
     "TPU_MESH_SORT_BY_COORDS":
     env_bool("TPU_MESH_SORT_BY_COORDS", default=False),
-    # Controls whether FP8 layers perform incremental weight loading and sharding.
+    # Controls whether FP8 linear and MoE layers perform incremental weight
+    # loading, sharding, and immediate host RAM cleanup. When enabled, weights
+    # are sharded and transferred to TPU device memory layer-by-layer (or per
+    # MoE expert shard), freeing PyTorch CPU storage and trimming glibc malloc
+    # arenas after each layer. This prevents host CPU memory exhaustion (OOM)
+    # when initializing large FP8 models on smaller RAM TPUs such as TPU8i.
     "VLLM_INCREMENTAL_FP8_LOADING":
     env_bool("VLLM_INCREMENTAL_FP8_LOADING", default=False),
 }
--- a/tpu_inference/layers/vllm/quantization/base.py
+++ b/tpu_inference/layers/vllm/quantization/base.py
@@ -19,7 +19,50 @@ import torch
 
 class VllmQuantizationMethod(ABC):
 
+    def maybe_process_linear_weights(
+        self,
+        layer: torch.nn.Module,
+        param_name: str,
+        args,
+        kwargs,
+        num_proj: int,
+        log_prefix: str = "",
+    ):
+        """Shared tracking logic for incremental sharding of linear layers."""
+        if isinstance(layer, vllm_linear.QKVParallelLinear):
+            if len(args) == 1:
+                shard_id = args[0]
+                layer._loaded_weights.add((param_name, shard_id))
+            else:
+                layer._loaded_weights.add((param_name, "q"))
+                layer._loaded_weights.add((param_name, "k"))
+                layer._loaded_weights.add((param_name, "v"))
+        elif isinstance(layer, vllm_linear.MergedColumnParallelLinear):
+            if len(args) == 1:
+                shard_id = args[0]
+                layer._loaded_weights.add((param_name, shard_id))
+            else:
+                for i in range(len(layer.output_sizes)):
+                    layer._loaded_weights.add((param_name, i))
+        else:
+            layer._loaded_weights.add(param_name)
+
+        expected_count = num_proj * len(
+            dict(layer.named_parameters(recurse=False)))
+        if len(layer._loaded_weights) == expected_count:
+            prefix_str = f"[{log_prefix}] " if log_prefix else ""
+            logger.debug(
+                f"{prefix_str}Start sharding weights for layer {type(layer)}")
+            self.process_weights_after_loading(layer)
+            logger.debug(
+                f"{prefix_str}Complete sharding weights for layer {type(layer)}"
+            )
+
     @abstractmethod
     def maybe_process_weights(self, layer: torch.nn.Module, param_name: str,
                               args, kwargs):
+        raise NotImplementedError
+
+    @abstractmethod
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
         raise NotImplementedError
--- a/tpu_inference/layers/vllm/quantization/base.py
+++ b/tpu_inference/layers/vllm/quantization/base.py
@@ -15,6 +15,10 @@
 from abc import ABC, abstractmethod
 
 import torch
+from vllm.logger import init_logger
+from vllm.model_executor.layers import linear as vllm_linear
+
+logger = init_logger(__name__)
 
 
 class VllmQuantizationMethod(ABC):
--- a/tpu_inference/layers/vllm/quantization/fp8.py
+++ b/tpu_inference/layers/vllm/quantization/fp8.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ctypes
 import gc
 from typing import Optional, Union
 
@@ -59,6 +60,39 @@ from tpu_inference.logger import init_logger
 P = PartitionSpec
 
 logger = init_logger(__name__)
+
+
+def _free_torch_storage(tensor: Optional[torch.Tensor]) -> None:
+    """Safely frees the underlying CPU memory storage of a PyTorch tensor.
+
+    Tries `untyped_storage().resize_(0)` first, with fallback to `set_(torch.storage.UntypedStorage())`
+    for 0-dim scalars or float8 dtypes that cannot be resized in-place.
+    """
+    if tensor is None:
+        return
+    try:
+        tensor.untyped_storage().resize_(0)
+    except Exception:
+        tensor.set_(torch.storage.UntypedStorage())
+
+
+try:
+    _libc = ctypes.CDLL("libc.so.6")
+except Exception:
+    _libc = None
+
+
+def _release_host_memory() -> None:
+    """Frees CPU host memory and trims malloc arena if incremental FP8 loading is enabled."""
+    if not envs.VLLM_INCREMENTAL_FP8_LOADING:
+        return
+    gc.collect()
+    jax.effects_barrier()
+    if _libc is not None and hasattr(_libc, "malloc_trim"):
+        try:
+            _libc.malloc_trim(0)
+        except Exception as e:
+            logger.debug(f"[fp8-incremental] malloc_trim failed: {e}")
 
 
 @register_quantization_config(FP8)
@@ -144,27 +178,17 @@ class VllmFp8LinearMethod(
         if not envs.VLLM_INCREMENTAL_FP8_LOADING:
             return
 
-        if isinstance(layer, vllm_linear.QKVParallelLinear):
-            if len(args) == 1:
-                shard_id = args[0]
-                layer._loaded_weights.add((param_name, shard_id))
-            else:
-                layer._loaded_weights.add((param_name, "q"))
-                layer._loaded_weights.add((param_name, "k"))
-                layer._loaded_weights.add((param_name, "v"))
-        elif isinstance(layer, vllm_linear.MergedColumnParallelLinear):
-            if len(args) == 1:
-                shard_id = args[0]
-                layer._loaded_weights.add((param_name, shard_id))
-            else:
-                for i in range(len(layer.output_sizes)):
-                    layer._loaded_weights.add((param_name, i))
-        else:
-            layer._loaded_weights.add(param_name)
+        logger.info_once(
+            "[fp8-incremental] VLLM_INCREMENTAL_FP8_LOADING is enabled")
 
-        if len(layer._loaded_weights) == self.linear_config.num_proj * len(
-                dict(layer.named_parameters(recurse=False))):
-            self.process_weights_after_loading(layer)
+        self.maybe_process_linear_weights(
+            layer,
+            param_name,
+            args,
+            kwargs,
+            self.linear_config.num_proj,
+            log_prefix="fp8-incremental",
+        )
 
     def create_weights(
         self,
@@ -197,13 +221,10 @@ class VllmFp8LinearMethod(
         )
 
         p_weight = layer.weight
-        p_scale = getattr(
-            layer, "weight_scale_inv" if self.block_quant else "weight_scale",
-            None)
 
         weight = _load_weight_for_layer(layer, "weight", loading_sharding)
         weight = jnp.transpose(weight)
-        p_weight.set_(torch.storage.UntypedStorage())
+        _free_torch_storage(p_weight)
         delattr(layer, "weight")
 
         if self.block_quant:
@@ -211,22 +232,24 @@ class VllmFp8LinearMethod(
             # Float8_e8m0fnu (ue8m0) scales cannot be converted via numpy in t2j.
             # TODO: consider get rid of the f32 conversion, to optimize the HBM usage.
             if weight_scale_inv.dtype == torch.float8_e8m0fnu:
-                layer.weight_scale_inv = weight_scale_inv.to(torch.float32)
+                layer.weight_scale_inv = Parameter(
+                    weight_scale_inv.to(torch.float32), requires_grad=False
+                )
             weight_scale = _load_weight_for_layer(layer, "weight_scale_inv",
                                                   loading_sharding)
             weight_scale = jnp.transpose(weight_scale)
-            if p_scale is not None:
-                p_scale.set_(torch.storage.UntypedStorage())
+            _free_torch_storage(layer.weight_scale_inv)
             delattr(layer, "weight_scale_inv")
         else:
             weight_scale_tensor = layer.weight_scale
             if weight_scale_tensor.dtype == torch.float8_e8m0fnu:
-                layer.weight_scale = weight_scale_tensor.to(torch.float32)
+                layer.weight_scale = Parameter(
+                    weight_scale_tensor.to(torch.float32), requires_grad=False
+                )
             scale_sharding = NamedSharding(self.linear_config.mesh, P(None))
             weight_scale = _load_weight_for_layer(layer, "weight_scale",
                                                   scale_sharding)
-            if p_scale is not None:
-                p_scale.set_(torch.storage.UntypedStorage())
+            _free_torch_storage(layer.weight_scale)
             delattr(layer, "weight_scale")
 
         if layer.bias is not None and not layer.skip_bias_add:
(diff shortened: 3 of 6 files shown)


Lumosis (approved):

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -228,18 +305,35 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
                 per_tensor=not self.block_quant,
             ))
 
+        scale_name = "weight_scale_inv" if self.block_quant else "weight_scale"
```
QiliangCui2023: **Blocker: block-quant linear layers will crash at first forward — `apply` reads `layer.weight_scale`, but this now stores the processed scale as `weight_scale_inv`.** `apply` (below, both branches) does `jax_view(layer.weight_scale)` / `zip(layer.weight, layer.weight_scale)` unconditionally, so after this `setattr(layer, scale_name, ...)` a block-quant FP8 linear has no `weight_scale` attribute → `AttributeError` on the first forward pass, regardless of `VLLM_INCREMENTAL_FP8_LOADING`. The previous code deliberately normalized the processed scale to `layer.weight_scale` for both quant types. Either keep storing under `weight_scale`, or update `apply` to read via `scale_name`. (If the Qwen3.5 benchmark passed, its attention linears presumably didn't go through this class as block-quant — worth confirming which configs exercise this path.)

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -151,32 +194,63 @@ def create_weights(
...
         delattr(layer, "weight")
 
         if self.block_quant:
             weight_scale_inv = layer.weight_scale_inv
             # Float8_e8m0fnu (ue8m0) scales cannot be converted via numpy in t2j.
             # TODO: consider get rid of the f32 conversion, to optimize the HBM usage.
             if weight_scale_inv.dtype == torch.float8_e8m0fnu:
-                weight_scale_inv = weight_scale_inv.to(torch.float32)
-            weight_scale = t2j(weight_scale_inv, use_dlpack=False)
+                layer.weight_scale_inv = weight_scale_inv.to(torch.float32)
```
QiliangCui2023: **Blocker: this raises `TypeError` for UE8M0 checkpoints — assigning a plain `Tensor` to a registered `Parameter` name.** vLLM registers `weight_scale_inv` via `layer.register_parameter("weight_scale_inv", scale)`, and `.to(torch.float32)` returns a plain `Tensor`, so `nn.Module.__setattr__` raises `TypeError: cannot assign ... as parameter 'weight_scale_inv' (torch.nn.Parameter or None expected)`. This is on the **default** (flag-off) path too — a regression vs the old local-variable conversion, hit by any DeepSeek-style e8m0-scale checkpoint. Same issue on the non-block branch (`layer.weight_scale = weight_scale_tensor.to(...)` at L238). Wrap in `Parameter(..., requires_grad=False)` or keep the conversion in a local and pass it through.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -130,6 +139,40 @@ def __init__(
...
+        else:
+            layer._loaded_weights.add(param_name)
+            num_projs = self.linear_config.num_proj
+
+        has_bias = getattr(layer, "bias", None) is not None and not getattr(
+            layer, "skip_bias_add", False)
+        num_params = 3 if has_bias else 2
+        expected_shards = num_projs * num_params
+
+        if len(layer._loaded_weights) >= expected_shards:
```
QiliangCui2023: **Early-fire risk: the hand-computed `expected_shards` can trigger processing before all shards are loaded → silently wrong weights.** `num_params = 3 if has_bias else 2` assumes weight+scale(+bias) are the only parameters, but FP8 with static activation quant also creates `input_scale` — its loader calls add entries to `_loaded_weights`, so the `>=` threshold is crossed while weight/scale shards are still missing, and `process_weights_after_loading` runs on a partially-loaded layer (later loader calls then hit the freed/deleted params). The merged unquantized method avoids this class of bug by counting dynamically: `num_proj * len(dict(layer.named_parameters(recurse=False)))`. Recommend the same here.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -228,18 +305,35 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+                layer,
+                scale_name,
+                to_parameter_list(weights.weight_scale),
+            )
+            if has_bias:
                 layer.bias = to_parameter_list(weights.bias)
 
+        import ctypes
+
+        gc.collect()
```
QiliangCui2023: **This cleanup block runs per layer even with `VLLM_INCREMENTAL_FP8_LOADING=False`.** On the default bulk path vLLM calls `process_weights_after_loading` for every layer back-to-back, so this adds a full `gc.collect()` (+`malloc_trim`) per layer across hundreds of layers — a real load-time regression for all FP8 models (the unquantized method does none of this). Please gate it on the env flag. Also: the same block is duplicated in the MoE method with an inline `import ctypes` in both — extract a small `_release_host_memory()` helper with a module-level import, and log the `except Exception` at debug instead of swallowing it silently.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -130,6 +139,40 @@ def __init__(
                 "Blockwise quantization is supported by quantized matmul kernel. Please enable quantized_matmul_kernel or unset the quantize block size to trigger XLA per-channel quantization."
             )
 
+    def maybe_process_weights(self, layer: torch.nn.Module, param_name: str,
```
QiliangCui2023, line 177: **Near-verbatim duplication of `VllmUnquantizedLinearMethod.maybe_process_weights`.** The QKV/MergedColumn/else tracking branches here are copied from the unquantized method with only the trigger arithmetic changed. Suggest lifting the shared body into `VllmQuantizationMethod` (with a hook for the expected-params count) — that also fixes the counting divergence flagged above in one place, since the unquantized version's dynamic `named_parameters` count is the correct arithmetic for both.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -151,32 +194,63 @@ def create_weights(
...
         assert isinstance(layer, vllm_linear.LinearBase)
 
-        weight = t2j(layer.weight, use_dlpack=False)
+        loading_sharding = NamedSharding(
+            self.linear_config.mesh,
+            PartitionSpec(*self.linear_config.weight_sharding[::-1]),
+        )
+
+        p_weight = layer.weight
+        p_scale = getattr(
```
QiliangCui2023: Ordering nit that matters once the e8m0 fix lands: `p_scale` is captured here **before** the e8m0→f32 reassignment below, so the tensor freed later is the stale pre-conversion one and the f32 copy's storage is never freed. Capture after conversion (or free via `getattr(layer, name)` at free time).

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -333,6 +457,19 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
             # Convert to tuple so jax jit can hash it
             weight_block_size=weight_block_size,
         )
+
+        # Free CPU memory now that weights have been safely transferred to TPU
+        layer.w13_weight.untyped_storage().resize_(0)
+        layer.w2_weight.untyped_storage().resize_(0)
+        p_w13_scale.set_(torch.storage.UntypedStorage())
```
QiliangCui2023: Inconsistent free idioms: weights use `untyped_storage().resize_(0)` but scales use `set_(torch.storage.UntypedStorage())` — both work, but pick one (the rest of the file and the unquantized method use `resize_(0)`), ideally via a shared `_free_torch_storage()` helper; the ParameterList-vs-tensor branching is also repeated three times in the linear method.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -342,10 +479,24 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+            Parameter(weights.w13_weight_scale, requires_grad=False),
+        )
+        setattr(
+            layer,
+            scale_w2_name,
+            Parameter(weights.w2_weight_scale, requires_grad=False),
+        )
+        import ctypes
+
+        gc.collect()
```
QiliangCui2023: Observability: the unquantized incremental path logs debug start/complete per layer; this path is completely silent. Please add matching `[fp8-incremental]`-tagged debug logs (and log the flag state once at load start) so an incremental-load OOM or early-fire can be triaged from the log alone.

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -295,23 +389,53 @@ def __init__(self,
...
+            return
+
+        expert_id = kwargs.get("expert_id")
+        shard_id = kwargs.get("shard_id")
+        if expert_id is not None and shard_id is not None:
+            layer._loaded_weights.add((param_name, expert_id, shard_id))
+        else:
+            layer._loaded_weights.add(param_name)
+
+        num_experts = getattr(layer, "global_num_experts", 0)
```
QiliangCui2023: **MoE counting: two robustness issues.** (1) `getattr(layer, "global_num_experts", 0)` — defaulting to 0 makes `expected_shards` ≈ 2, so the first scale load would trigger processing on an almost-empty layer; if the attribute is guaranteed, read it directly (like the unquantized method), otherwise assert. (2) For **non-block** FP8, `scale_w13_expected/scale_w2_expected = 1` is wrong if per-tensor scales are loaded per expert with `expert_id`/`shard_id` kwargs (as block scales are) — entries reach 3E, crossing the `>=` threshold at ~half the shards → early processing with missing experts. The unquantized method also *asserts* the kwargs are present rather than silently falling back to bare `param_name` tracking. Suggest dynamic counting via `named_parameters`, and/or explicitly rejecting non-block-quant for incremental mode until it's validated.

QiliangCui2023 (commented):
Reviewed at max depth (verified against the merged unquantized incremental path and the pinned vLLM). The approach is sound and mirrors proven code, but two issues break the **default flag-off path** (scale-name mismatch vs `apply`, e8m0 `Parameter` assignment) and the hand-rolled shard counting can fire early and silently corrupt weights — details inline.

Also: no unit tests accompany this change. The trigger arithmetic, fire-exactly-on-last-shard behavior, already-processed no-op guard, and e8m0 conversion are all cheaply unit-testable with mock layers, and such tests would have caught the two blockers. Please add coverage before merge.

Verified as sound, for the record: block-quant MoE shard counting (exactly 6E), the immediate-free-after-t2j pattern (identical to the shipped unquantized path), `_loaded_weights` initialization coverage, and the `_tensor_is_in_cpu` reprocessing guards.

tpu_inference/envs.py
```diff
@@ -456,6 +441,9 @@ def _get_int_list_env() -> list[int]:
     # Currently, it only supports a single host set up.
     "TPU_MESH_SORT_BY_COORDS":
     env_bool("TPU_MESH_SORT_BY_COORDS", default=False),
+    # Controls whether FP8 layers perform incremental weight loading and sharding.
+    "VLLM_INCREMENTAL_FP8_LOADING":
+    env_bool("VLLM_INCREMENTAL_FP8_LOADING", default=False),
```
lk-chen, line 451: should we default turn on it?

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done

QiliangCui2023 (approved):
All review comments verified as addressed at d695ae7d4f19bef6fc197a75039877ec3a81ec7e — scale normalization back to `weight_scale`, e8m0 Parameter wrapping, shared dynamic shard counting in `VllmQuantizationMethod`, flag-gated `_release_host_memory()`, post-conversion freeing, unified `_free_torch_storage()`, `[fp8-incremental]` logging, and the hardened MoE trigger (direct `global_num_experts`, kwargs asserts, non-block rejected). The new tests cover the trigger arithmetic, idempotency guard, and e8m0 conversion. Approving.

One non-blocking follow-up (pre-existing pattern shared with the unquantized incremental path, not introduced here): with incremental loading enabled and a non-fused linear (`fuse_matmuls=False`), the end-of-load sweep in `vllm_model_loader.py` re-calls `process_weights_after_loading`, and the idempotency guard doesn't catch it — after processing, `layer.weight` is a `ParameterList`, which `_tensor_is_in_cpu` treats as still-in-CPU (it isn't a torchax tensor), so the method re-processes and fails at `_load_weight_for_layer` with `AttributeError: 'ParameterList' object has no attribute 'device'`. Unreachable today (`fuse_matmuls` defaults to True and the non-fused truth table has no FP8 models), but worth hardening `_tensor_is_in_cpu` (or the shared guard) to treat `ParameterList` as already-processed — that fixes both the FP8 and unquantized methods in one place.

QiliangCui (approved):

tpu_inference/envs.py
BirdsOfAFthr: I wanted to leave this off until we tried it on more models.

tpu_inference/layers/vllm/quantization/base.py
```diff
@@ -15,11 +15,58 @@
...
+                layer._loaded_weights.add((param_name, shard_id))
+            else:
+                for i in range(len(layer.output_sizes)):
+                    layer._loaded_weights.add((param_name, i))
+        else:
+            layer._loaded_weights.add(param_name)
+
+        expected_count = num_proj * len(
+            dict(layer.named_parameters(recurse=False)))
+        if len(layer._loaded_weights) == expected_count:
```
jrplatin, line 56: What if this isn't the case? What happens?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -41,20 +45,57 @@
...
+    _load_weight_for_layer)
 from tpu_inference.logger import init_logger
-from tpu_inference.utils import t2j
 
 P = PartitionSpec
 
 logger = init_logger(__name__)
 
 
+def _free_torch_storage(tensor: Optional[torch.Tensor]) -> None:
```
jrplatin, line 69: Is this the cannonical / typical way of doing this? Seems a little hacky to me?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -41,20 +45,57 @@
...
+    if tensor is None:
+        return
+    try:
+        tensor.untyped_storage().resize_(0)
+    except Exception:
+        tensor.set_(torch.storage.UntypedStorage())
+
+
+try:
+    _libc = ctypes.CDLL("libc.so.6")
```
jrplatin: Remove?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -41,20 +45,57 @@
...
+except Exception:
+    _libc = None
+
+
+def _release_host_memory() -> None:
```
jrplatin, line 83: Same here as above -- this seems a bit hacky to me

tpu_inference/envs.py
```diff
@@ -456,6 +441,9 @@ def _get_int_list_env() -> list[int]:
     # Currently, it only supports a single host set up.
     "TPU_MESH_SORT_BY_COORDS":
     env_bool("TPU_MESH_SORT_BY_COORDS", default=False),
+    # Controls whether FP8 layers perform incremental weight loading and sharding.
```
jrplatin: Can we please make this comment more expansive? What is "incremental weight loading and sharding"? How is it different from what we have now by default?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -151,26 +212,45 @@ def create_weights(
...
             weight_scale = jnp.transpose(weight_scale)
+            _free_torch_storage(layer.weight_scale_inv)
             delattr(layer, "weight_scale_inv")
         else:
-            weight_scale = layer.weight_scale
-            if weight_scale.dtype == torch.float8_e8m0fnu:
-                weight_scale = weight_scale.to(torch.float32)
-            weight_scale = t2j(weight_scale, use_dlpack=False)
+            weight_scale_tensor = layer.weight_scale
+            if weight_scale_tensor.dtype == torch.float8_e8m0fnu:
```
jrplatin, line 246: Can you explain what all of this is doing?

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -295,23 +380,62 @@ def __init__(self,
...
+                "Incremental loading for non-block FP8 MoE is not supported.")
+
+        expert_id = kwargs.get("expert_id")
+        shard_id = kwargs.get("shard_id")
+        assert expert_id is not None, "Expecting expert_id argument"
+        assert shard_id is not None, "Expecting shard_id argument"
+
+        layer._loaded_weights.add((param_name, expert_id, shard_id))
+
+        expected_shards = 6 * layer.global_num_experts
```
jrplatin, line 409: Why 6?

tpu_inference/layers/vllm/quantization/base.py
BirdsOfAFthr: it will continue loading until we meet the requirement. We can only process after all the weights for that layer are loaded. Note this code was moved from the unquantized.py without any chnages.

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Yes, this is how we do this in unquantized.py as well.

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: This is necessary here because gc.collect() only releases Python/PyTorch objects—glibc retains the freed memory pages in its CPU heap arenas rather than returning them to the OS kernel. For large models (e.g. Qwen 3.5 397B FP8) loading on sfs, calling malloc_trim(0) after layer weight processing is required to return CPU RAM back to the kernel and prevent host memory exhaustion during incremental loading.

I've cleaned this up by removing the top-level global _libc handle and calling ctypes.CDLL("libc.so.6").malloc_trim(0) directly inside _release_host_memory().

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Same comment above

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: for the code I added:

* Replaces t2j with _load_weight_for_layer(..., scale_sharding) to stream the scale tensor to device memory with proper sharding.

* Calls _free_torch_storage(layer.weight_scale) immediately after loading to free the PyTorch CPU memory buffer before proceeding to the next layer.

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: For block-quantized FP8 MoE, each expert has 6 parameter shards:
 w1_weight, w1_weight_scale_inv, w3_weight, w3_weight_scale_inv,
w2_weight, and w2_weight_scale_inv.

added a docstring

tpu_inference/envs.py
BirdsOfAFthr: done

tpu_inference/layers/vllm/quantization/fp8.py
kyuyeunk: it's true that's how unquantized.py is doing it, but it's not a correct way.

tpu_inference/layers/vllm/quantization/fp8.py
jrplatin: Ok, please add a TODO that we should clean this up, thanks!

tpu_inference/layers/vllm/quantization/fp8.py
```diff
@@ -77,23 +77,16 @@ def _free_torch_storage(tensor: Optional[torch.Tensor]) -> None:
...
         return
     gc.collect()
     jax.effects_barrier()
-    if _libc is not None and hasattr(_libc, "malloc_trim"):
-        try:
-            _libc.malloc_trim(0)
-        except Exception as e:
-            logger.debug(f"[fp8-incremental] malloc_trim failed: {e}")
+    try:
+        ctypes.CDLL("libc.so.6").malloc_trim(0)
```
jrplatin: This still seems hard-coded? What 's `libc.so.6`?

BirdsOfAFthr: refactored and added docstring.

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Added a TODO. Thanks!

tpu_inference/layers/vllm/quantization/fp8.py
BirdsOfAFthr: Done. Thanks!

jrplatin (approved):

lk-chen (approved):

PR #3241, opened 2026-07-23
Multi-request prefill context parallelism (PCP)

wenxindongwork:
## Multi-request prefill context parallelism (PCP)

Generalizes PCP from **one request per model step** to **R fixed request lanes**, so prefill can batch multiple requests while still context-parallelizing each one head-tail across the `pcp` ranks. Targets disagg mode (**prefill-only, no mixed prefill/decode**).

### Design: wrapper request-loop, no kernel changes
Rather than teaching the RPA kernel to batch `2N` head-tail seqs (which would need a per-seq current-KV base offset and a per-request tail-write mask), the wrapper **loops over R lanes and runs the proven single-request head-tail body per lane**. Each per-lane kernel call still sees exactly one request = two seqs (head, tail), so `write_last_seq_only` and the scalar `pcp_chunk_size` remap stay valid **with zero kernel changes**. The KV cache is threaded through the loop; requests write disjoint pages via per-request block tables.

### Layout
Uniform per-lane chunk size `C` (compile-stable). The token buffer is `R * Lreq` (R = `max_num_reqs` lanes, `Lreq = 2*pcp*C` a power-of-2 lane length from the **longest single request**), laid out **RANK-MAJOR → LANE → head|tail**:
```
c<pcp : rank=c,      dest = c*region      + i*2C + (p%C)
c>=pcp: rank=2P-1-c, dest = (2P-1-c)*region + i*2C + C + (p%C)   (region = R*2C = tokens//pcp)
```
So each pcp rank's contiguous shard is `concat_i [head_r,i | tail_r,i]`, which the wrapper slices per lane. `R=1` reduces exactly to the previous single-request layout.

### Changes
- **`attention_interface.pcp_ragged_paged_attention`**: static loop over R lanes; per-lane head/tail geometry (`cu`, `q_pos_offsets`, `tail_real`) derived inside the shard_map from `C`, rank and per-request `seq_len`/`kv_cache_len` — dropping the `cu_q_lens`/`pcp_q_pos_offsets` metadata arrays. New signature takes `num_reqs`.
- **`tpu_runner`**: PCP token buckets = `R * (pow2 lane)`; pack tokens rank-major/lane/head-tail; per-lane `kv_cache_lens` and packed `logits_indices`; drop the single-request assert and the block-table duplication hack. `seq_lens`/`block_tables`/`query_start_loc` reuse the generic per-request fill (already lane-ordered).
- **`compilation_manager`**: precompile dummy metadata drops `pcp_q_pos_offsets` and the PCP `query_start_loc` override; `attn_num_reqs_paddings` pinned to `[max_num_reqs]` for PCP.
- **test**: `test_pcp_attention_interface.py` rewritten for the new signature with genuine multi-request cases (varied lengths, empty padded lane, per-request KV-write check).

### Validation status
- ✅ Packing ⇄ reconstruction index math and logits indices cross-checked by a standalone numpy simulation (pcp 2/4, single/multi/empty-lane/tiny).
- ✅ `pyflakes` / AST clean.
- ⏳ **Not yet TPU-validated** (TPU was occupied by a running `vllm serve`). Needs: offline LLM correctness vs single-request + TP baseline, and a successful precompile of the PCP token buckets.

### Known limitations / follow-ups
- Fixed `R = max_num_reqs` ⇒ buffer `= R * max_lane`; intended for PCP's natural regime (**modest `max_num_seqs`, long contexts**). A large `max_num_seqs` over-allocates — bucketed-R is a follow-up.
- `R` sequential kernel launches per layer (correctness-first v1; batching is a perf follow-up).
- No M-RoPE, no mixed prefill/decode, no DP (asserted).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wenxindongwork:
## ✅ TPU-validated (tp2 pcp2 vs tp4 baseline, Qwen3-0.6B, fp32, greedy)

**One bug found + fixed during validation** (91cba84): the cache-phase `cu_q_lens` must be length `max_num_seqs+1 = 3` (each per-lane call has 2 fused seqs); `cu_cache`/`cu_kv` were length 2.

### Results
- **Short prompts** (8 prompts, `max_num_seqs=8`, all batched into ONE prefill step → genuine 8-request PCP): **8/8 first tokens bit-exact** vs the tp4 baseline.
- **Long prompts** (~300 tok, deeper head-tail) with **chunked prefill** (`num_computed>0`, exercises the cache-read path): **3/4 bit-exact**; the 4th is a **numerical argmax tie, not a bug** —

  | token | TP logprob | PCP logprob |
  |---|---|---|
  | `' __'` (1304) | -2.28116 | -2.30591 |
  | `' _'` (716)  | -2.28604 | -2.32454 |

  The top-2 differ by only ~0.005 logprob (TP) / ~0.019 (PCP); TP itself flips between them across its own code paths (plain greedy → 716, logprobs path → 1304). All top-8 logprobs match within ~0.03 — the expected fp32 reduction-order noise from PCP's head-tail LSE merges + all-gather/reduce-scatter.

Precompile of the `R*L` token buckets succeeded; `num_reqs=8` lanes compiled and ran. Also confirmed `R=1` reduces exactly to the previous single-request layout.

wenxindongwork:
## ⚠️ Correction: multi-request DECODE is broken — this PR is prefill-only for now

My earlier "TPU-validated" comment was **overstated**: it only validated multi-request **prefill** (`max_tokens=1`), which never exercises the KV-cache read on a decode step. Running gsm8k on a single combined server (prefill+decode) exposes a real failure.

### gsm8k (Qwen3-8B-Base, tp4×pcp2, 200 ex, this branch)
| config | score |
|---|---|
| no PCP (tp4) | **0.845** |
| base `wxd-pcp` + PCP, `max_num_seqs=1` | 0.725 |
| **this branch** + PCP, `max_num_seqs=1` | **0.70** (no single-request regression) |
| **this branch** + PCP, `max_num_seqs=4` | **0.03** ❌ |

### Root cause
Offline (Qwen3-0.6B, multi-token): multi-request **prefill is correct** (prefill/first token bit-matches TP for every lane), but the multi-request **decode step corrupts lanes ≥ 1** — the *first decode token* is already wrong for lanes 1..3 (lane 0 fine), then cascades. A discriminator with long prompts (each prefilled alone in lane 0, then decoded together) still garbles lanes ≥1, so it is **not** the prefill cache-write — it's the multi-request decode path.

Architecturally: the wrapper routes *every* step through head-tail packing. For decode `num_current=1` → `C=1`, a degenerate "split" where each request's single token lands entirely on pcp rank 0, and lanes ≥1 aren't reconstructed correctly. Decode was outside the disagg-prefill scope this PR targets; single-request decode only works because `R=1` reduces to the base path.

### Status
- **Prefill (disagg): correct and validated** — bit-exact vs TP for single- and multi-request prefill.
- **Single combined server with `max_num_seqs>1`: NOT supported yet** (multi-request decode bug).
- Fix direction (needs a decision): give decode its own non-head-tail path that reads the pcp-strided KV cache (DCP-style gather), rather than forcing decode through the prefill wrapper.

wenxindongwork:
## ✅ Fixed: multi-request decode now works — gsm8k 0.835 with `max_num_seqs=4`

Root cause of the earlier 0.03 collapse was **not** the attention path — it was **async token substitution**. With async scheduling (on by default), the previous step's sampled tokens are written into `input_ids` at **natural** (contiguous) token positions, but PCP repacks tokens into the lane layout (request `i` → `dest = i*2C`). Request 0 aligns (0==0); requests ≥1 had their decode token written to the wrong slot → every lane after the first decoded garbage (prefill correct, *first decode token* wrong, cascading). This is why it was multi-request-only, positional, and reproduced even with identical prompts.

**Fix** (`tpu_runner.py`, small): build a natural→packed index map during packing and remap the async substitution indices for PCP. No kernel/attention changes; prefill path untouched.

### gsm8k (Qwen3-8B-Base, 200 ex)
| config | score |
|---|---|
| no PCP (tp4, seqs=4) | 0.845 |
| **this branch + PCP (tp4×pcp2, seqs=4)** — before fix | 0.03 ❌ |
| **this branch + PCP (tp4×pcp2, seqs=4)** — after fix | **0.835** ✅ |

Offline (Qwen3-0.6B tp2×pcp2) also confirms: identical prompts now give identical per-lane output, and 4-request decode matches the tp4 baseline token-for-token.

Multi-request PCP now works end-to-end (prefill **and** decode) on a single combined server.

PR #3242, opened 2026-07-23
[tpu-inference] Add block-diffusion strategy configuration

ethannnnnn:
> [!WARNING]
> This is an experimental, opt-in contract under
> `tpu_inference.runner.experimental.diffusion`. It has no compatibility
> guarantee yet and does not introduce or replace a vLLM runner.

## Motivation

Block-diffusion models need explicit model semantics and runtime controls without coupling the TPU runner to one model configuration.

## Scope

- Add typed generation-strategy, model, and runtime configuration.
- Add a model-adapter hook and a built-in seeded/shifted adapter.
- Validate block size, mask token, sub-block size, algorithm, confidence, temperature, and denoising limits.

## Design

`generation_strategy` defaults to `autoregressive`. Selecting `block_diffusion` resolves an immutable model specification plus runtime configuration from vLLM and Hugging Face configuration. The contracts live under the explicit `runner.experimental.diffusion` namespace. Policy enums make attention, logit alignment, canvas construction, prompt remainders, and next-block behavior explicit.

## Compatibility

Autoregressive behavior is unchanged: it remains the default, resolves without diffusion configuration, and this change does not wire a new runner path. The experimental package may change without a compatibility transition before it is promoted.

## Extensibility

Model adapters are callables that produce the shared model specification, so future models can supply different policies without adding model-name conditionals to the runner.

## Tests

- Relevant CPU test selection at this stack point: 10 passed.
- Stack checks passed: `isort`, `yapf`, `ruff`, `py_compile`, and `git diff --check`.

## Known limitations

This foundation declares the seeded/shifted adapter contract and low-confidence algorithm only. It does not add serving integration.

## Stack

- First PR in the stack; no previous PR dependency.
- [Public design document](https://docs.google.com/document/d/1d3qtGrvdcJ0ptluE0EwAaGj4AQyOlIYkYvKo9xPguX8/edit)

PR #3243, opened 2026-07-23
[tpu-inference] Drive attention masks from metadata

ethannnnnn:
## Motivation

Block diffusion needs masking policy to travel with attention metadata instead of being hard-coded at individual attention call sites.

## Scope

- Add a static causal-mask flag to the existing attention metadata.
- Resolve masking from metadata while retaining an explicit caller override.
- Reject bidirectional masking on attention paths that cannot implement it correctly.

## Design

`use_causal_mask` is static pytree metadata and defaults to `True`. The common attention entry point uses an explicit caller override when present; otherwise it passes the metadata value to the selected implementation.

## Compatibility

Autoregressive behavior is unchanged because newly constructed metadata is causal by default. Existing callers that pass `use_causal_mask` explicitly retain precedence.

## Tests

- Relevant CPU tests passed.

## Known limitations

Bidirectional attention is not supported with DCP or the `head_dim == 64` RPA path; those combinations fail early.

## Stack

- Depends on merged upstream PR https://github.com/vllm-project/tpu-inference/pull/3242, which is now included in `main`.

### DIFF
--- a/tests/layers/common/test_attention_interface.py
+++ b/tests/layers/common/test_attention_interface.py
@@ -23,7 +23,8 @@ from jax.sharding import Mesh
 from tpu_inference.layers.common.attention_interface import (
     attention, mla_attention, sharded_ragged_paged_attention)
 from tpu_inference.layers.common.attention_metadata import (
-    AttentionMetadata, SharedAttentionMetadata)
+    AttentionMaskKind, AttentionMaskSpec, AttentionMetadata,
+    SharedAttentionMetadata)
 from tpu_inference.layers.common.sharding import ShardingAxisName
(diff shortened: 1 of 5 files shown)


ethannnnnn:
Restacked the remaining experimental diffusion series after #3242 merged.

- #3243 is now a single focused commit on current `main`.
- #3244-#3247 remain ordered behind their immediate predecessor.
- The merged #3242 branch and PR were not modified.
- The existing `TPUModelRunner` remains the default; block diffusion is opt-in and lazy-loaded from the experimental namespace.

Validation: 44 focused CPU tests passed; isort, YAPF, Ruff, compilation, and `git diff --check` passed.

Lumosis (approved):

kyuyeunk (approved):

PR #3246, opened 2026-07-23
[tpu-inference] Add block-diffusion serving strategy

ethannnnnn:
> [!WARNING]
> Block-diffusion serving is experimental and opt-in. It extends the existing
> `TPUModelRunner` through a strategy object; it does not add or register a new
> runner. Its configuration and API may change without notice.

## Motivation

The configuration, attention, denoising, and scheduler foundations need an opt-in serving path that translates vLLM requests into block-diffusion work and returns generated tokens with correct request accounting.

## Scope

- Add prompt-block planning, per-request diffusion state, batched execution, and output assembly.
- Wire the experimental block-diffusion strategy into TPU platform configuration and the existing `TPUModelRunner`.
- Use a nested-JIT-safe model callable inside the denoising program while retaining the optimized callable for normal autoregressive execution.
- Expose the standard `modules()` traversal API on JAX modules so vLLM lifecycle cleanup can inspect Flax NNX models safely.
- Validate unsupported request, model, scheduler, kernel, and parallelism combinations before execution.

## Design

The platform resolves the generation strategy, emits an experimental-feature warning, sizes request capacity and multi-token lookahead from the model block size, and installs request validation. The existing runner lazy-imports and constructs `BlockDiffusionStrategy` only when selected. The strategy handles full and partial prompt blocks, invokes the jitted denoising program, buffers multi-token output, and removes finished request state. Because JAX compiler options are valid only on a top-level `jit`, the nested diffusion program uses an equivalent model callable without those options; the existing autoregressive path keeps its optimized callable unchanged. JAX modules provide the same `modules()` traversal shape expected by vLLM cleanup, while yielding only the existing JAX module graph.

## Compatibility

Autoregressive behavior is unchanged: it remains the default and follows the existing runner path when block diffusion is not selected. Continue-decode and block diffusion are mutually exclusive.

## Extensibility

Serving consumes the shared model specification and algorithm boundary rather than adapter-specific branches, allowing future adapters and commit algorithms to reuse the request lifecycle.

## Tests

- Current-main focused CPU gate across the stack: 44 passed.
- Block-diffusion CPU tests: 33 passed after the nested-JIT fix.
- JAX module traversal tests: 7 passed, including the vLLM-compatible `modules()` API.
- Existing runner model-loading fallback tests: 5 passed offline with the model configuration cache mounted read-only.
- Selected vLLM 0.23 platform and attention tests: 21 passed.
- Full platform test file before the fork-safety follow-up: 43 passed; 2 environment-only failures occurred because `tpu_info` was unavailable.
- A real 4x4x4 allocation ran 16 independent local 2x2x1 TPU workers. All 16 loaded the Flax NNX model, initialized the KV cache, compiled block diffusion, and exited successfully.
- Every worker completed two concurrent requests with exactly 64 output tokens per request: two 32-token blocks, 128 tokens per worker, no mask-token leakage, and non-empty decoded output.
- All 16 workers logged the explicit pass marker and exit code zero, with no traceback or cleanup `AttributeError`.
- Steady-state two-request generation took 0.647-0.688 seconds per worker (mean 0.665 seconds), or 186.1-197.8 aggregate output tokens/s per worker (mean 192.6). This excludes model initialization and is not a multi-host aggregate or an autoregressive comparison.
- E2E iteration history: isolated the installed runtime from source-checkout shadowing; made per-worker local topology explicit; gated unsupported prefix-cache, chunked-prefill, and async paths; removed top-level-only compiler options from the nested diffusion callable; then made the two-block smoke EOS-independent and added JAX lifecycle traversal compatibility. The final rerun passed after each failure was converted into a guarded contract or regression test.

## Known limitations

Chunked prefill is unsupported. `enable_chunked_prefill=False` is required for block-diffusion serving.

Pipeline parallelism, data parallelism greater than one, KV transfer, async scheduling, speculative decoding, LoRA, multimodal models, prefix caching, resumable input, streaming input, the batched RPA kernel, and `head_dim == 64` are not supported by this initial serving path.

The completed smoke validates the initial two-request, two-block path on real TPU hardware. It is a functional and baseline-performance result, not a saturation benchmark, dynamic-arrival/refill test, or multi-host request result.

## Stack

- Depends on the previous upstream PR in this stack: https://github.com/vllm-project/tpu-inference/pull/3245. Until that PR merges, GitHub's `main`-based comparison includes the preceding stack commits.

### DIFF
--- a/tests/core/test_multi_token_scheduler.py
+++ b/tests/core/test_multi_token_scheduler.py
@@ -1,0 +1,205 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib.util
+import pathlib
+import sys
+import types
+from types import SimpleNamespace
+
+
+def _load_utils_with_fake_vllm(monkeypatch):
+
+    class Scheduler:
+
+        def __init__(self, vllm_config):
+            del vllm_config
+            self.num_lookahead_tokens = 2
+
+        def _update_request_with_output(self,
+                                        request,
+                                        new_token_ids,
+                                        is_stale=False,
+                                        **kwargs):
+            del is_stale, kwargs
+            retained = list(new_token_ids[:request.retain_tokens])
+            request.output_token_ids.extend(retained)
+            return retained, len(retained) != len(new_token_ids)
+
+    class AsyncScheduler:
+
+        def _update_request_with_output(self,
+                                        request,
+                                        new_token_ids,
+                                        is_stale=False,
+                                        **kwargs):
+            del kwargs
+            if is_stale:
+                return list(new_token_ids), False
+            request.num_output_placeholders -= len(new_token_ids)
+            return list(new_token_ids), False
+
+    modules = {
+        "vllm":
+        types.ModuleType("vllm"),
+        "vllm.v1":
+        types.ModuleType("vllm.v1"),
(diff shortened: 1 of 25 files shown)


Lumosis (approved):

Lumosis:
Please add the "ready" label so that the presubmits can be triggered.

PR #3247, opened 2026-07-23
[tpu-inference] Make TPU device detection fork-safe

ethannnnnn:
## Motivation

TPU device-name discovery must not initialize a new process while Python worker creation is already in progress.

## Scope

- Route device-name detection through the existing fork-safe `tpu_info` wrapper.
- Normalize v5e and v7x accelerator names.
- Preserve the generic TPU fallback when discovery fails.

## Design

The platform imports the wrapper with its other TPU modules and calls `get_tpu_type()` from `get_device_name()`. A small normalization map converts accelerator identifiers to the names exposed through the vLLM platform API.

## Compatibility

Autoregressive and block-diffusion execution behavior is unchanged. This change affects platform device-name discovery only.

## Extensibility

Device probing remains centralized in the wrapper, and additional public accelerator-name mappings can be added without changing worker lifecycle behavior.

## Tests

- Full platform test file after this change: 50 passed.
- A real 4x4x4 TPU allocation completed on all 16 independent local workers; each process resolved the TPU platform, ran two block-diffusion requests, and exited zero without a traceback.

## Known limitations

Discovery failures still return the generic `TPU` device name after logging a warning.

## Stack

- Depends on the previous upstream PR in this stack: https://github.com/vllm-project/tpu-inference/pull/3246. Until that PR merges, GitHub's `main`-based comparison includes the preceding stack commits.

### DIFF
--- a/tests/core/test_multi_token_scheduler.py
+++ b/tests/core/test_multi_token_scheduler.py
@@ -1,0 +1,205 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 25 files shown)


Lumosis:
Please update the PR to resolve the conflicts

PR #3252, opened 2026-07-23
Add Gemma3-4B MaxText inference test to RL CI matrix

wenxindongwork:
## Summary

Adds a nightly CI test that runs MaxText's Gemma3-4B inference path, catching tpu-inference changes that would break MaxText's RL rollouts.

Ported from MaxText's [`test_gemma3_rl.sh`](https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh).

### Why this belongs in tpu-inference CI

MaxText pins tpu-inference at a SHA in [`post_train_github_deps.txt`](https://github.com/AI-Hypercomputer/maxtext/blob/main/src/dependencies/extra_deps/post_train_github_deps.txt):

```
tpu-inference @ https://github.com/vllm-project/tpu-inference/archive/<sha>.zip
vllm @ git+https://github.com/vllm-project/vllm@<sha>
```

Breakage is therefore only discovered when MaxText *bumps* that pin. This test surfaces it at head instead.

Versions: 

- MaxText HEAD 
- tpu-inference HEAD 
-  vLLM at the latest `.buildkite/vllm_lkg.version`
- v7x-1

## Files

| File | Purpose |
|---|---|
| `tests/e2e/test_gemma3_maxtext_inference.sh` | The test |
| `.buildkite/rl/gemma3_maxtext_inference.yml` | Nightly `rl/` support-matrix entry (`soft_fail: true`) |

## What the test does

1. **Preflight** — checks the checkpoint is readable via `gcsfs`, failing in seconds instead of after the install and model load.
2. **Install maxtext from HEAD** with the `[tpu-post-train]` extra.
3. **Verify the env survived** — imports `vllm`, then `maxtext`.
4. **Run `maxtext.inference.vllm_decode`** on the pre-converted Gemma3-4B checkpoint.
5. **Assert non-empty output**, so an early no-op exit can't pass as green.

## Implementation notes

Several non-obvious constraints shaped this; recording them so they aren't rediscovered.

<details>
<summary><b>torch is upgraded 2.10 → 2.11, deliberately</b></summary>

maxtext HEAD pins `torch==2.11.0+cpu` exactly. The image ships `2.10.0+cpu` because that is what its source-built vLLM was compiled against.

Holding torch back would pin us to maxtext `0.2.1` (2026-03-23) — months behind the tpu-inference under test. Instead torch is allowed to move, and the script **verifies vLLM still imports afterwards**, failing with an actionable message if its native extensions don't load:

```bash
torch: 2.10.0+cpu -> 2.11.0+cpu
```

`ALLOW_TORCH_UPGRADE=0` restores the old behaviour for bisecting. The JAX/TPU stack (`jax`, `jaxlib`, `libtpu`, `flax`) stays pinned to the image's versions via a generated constraints file, since those are lower bounds that pip would otherwise resolve to the newest release.
</details>

<details>
<summary><b><code>VLLM_XLA_CHECK_RECOMPILATION</code> is disabled for this test</b></summary>

`run_in_docker.sh` hard-codes it to `1` for all CI jobs. The guard raises if any JAX compilation happens inside `execute_model`, assuming warmup precompiled every shape.

That assumption doesn't hold here: maxtext drives the vLLM `LLM` API through its own wrapper rather than tpu-inference's `offline_inference.py`, so warmup covers nothing (`warmup model took 0.27 s`) and the first `generate()` trips it.

```bash
SKIP_JAX_PRECOMPILE=1 VLLM_XLA_CHECK_RECOMPILATION=0 python3 ...
```

**Trade-off:** this test no longer detects recompilation regressions. That property remains covered by the model matrix.
</details>

## Checkpoint

Reads a pre-converted MaxText checkpoint from the bucket CI already uses for model weights:

```
gs://tpu-commons-ci/gemma3-4b/to_maxtext/unscanned/items
```

Copied from MaxText's `gs://runner-maxtext-logs`. Override with `GEMMA3_MT_CKPT`.

## Testing

Passing end-to-end on `tpu-inference-dev` (tpu7x): install → maxtext import → checkpoint load → KV cache init → decode → non-empty output.

### DIFF
--- a/tests/e2e/test_gemma3_maxtext_inference.sh
+++ b/tests/e2e/test_gemma3_maxtext_inference.sh
@@ -1,0 +1,215 @@
+#!/bin/bash
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ---------------------------------------------------------------------------
+# Gemma3-4B MaxText inference smoke test.
+#
+# This is the tpu-inference-side counterpart of MaxText's RL validation script
+#   https://github.com/AI-Hypercomputer/maxtext/blob/main/tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh
+# The MaxText script validates a full RL pipeline (inference -> GRPO training ->
+# inference). tpu-inference only participates in that pipeline as the vLLM TPU
+# backend that MaxText drives via the `MaxTextForCausalLM` plugin, so here we
+# port only the inference legs (steps 1 & 3 of that script): run
+# `maxtext.inference.vllm_decode` against a pre-converted MaxText Gemma3-4B
+# checkpoint and assert that decoding succeeds and produces non-empty output.
+#
+# This guards against tpu-inference regressions that would break MaxText's RL
+# rollouts, without baking maxtext into the shared CI image.
+#
+# maxtext is installed from HEAD (see MAXTEXT_REF), so this tracks the upstream
+# script rather than a months-old PyPI release. Doing so requires upgrading the
+# image's torch 2.10.0+cpu to the torch==2.11.0+cpu that maxtext HEAD pins; the
+# image's vLLM is compiled against 2.10, so the script verifies vLLM still
+# imports after the upgrade and fails loudly if it does not.
+#
+# Usage (inside the CI docker image):
+#   BUILDKITE_COMMIT=<sha> .buildkite/scripts/run_in_docker.sh \
+#     bash /workspace/tpu_inference/tests/e2e/test_gemma3_maxtext_inference.sh
+#
+# Requires:
+#   - HF_TOKEN in the environment (Gemma3 tokenizer/config are gated on HF).
+#   - Read access to the checkpoint GCS path (see GEMMA3_MT_CKPT below).
+#   - Network access to github.com and download.pytorch.org from the agent.
+# ---------------------------------------------------------------------------
+
+set -euo pipefail
+
+MODEL_NAME="${MODEL_NAME:-gemma3-4b}"
+
+# maxtext is not baked into the shared CI image; install it in-step so this test
+# does not affect the image size / build time of every other CI job.
+#
+# Install maxtext from HEAD rather than a PyPI release, so this tracks the same
+# code as the upstream script this test is ported from
+# (tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh on maxtext main). Pin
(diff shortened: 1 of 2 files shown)


tests/e2e/test_gemma3_maxtext_inference.sh
```diff
@@ -0,0 +1,215 @@
...
+MODEL_NAME="${MODEL_NAME:-gemma3-4b}"
+
+# maxtext is not baked into the shared CI image; install it in-step so this test
+# does not affect the image size / build time of every other CI job.
+#
+# Install maxtext from HEAD rather than a PyPI release, so this tracks the same
+# code as the upstream script this test is ported from
+# (tests/end_to_end/tpu/gemma3/4b/test_gemma3_rl.sh on maxtext main). Pin
+# MAXTEXT_REF to a SHA if a green run needs to be reproducible.
+MAXTEXT_REF="${MAXTEXT_REF:-main}"
```
pv97, line 58: If maxtext is broken at HEAD, would it block our release?

tests/e2e/test_gemma3_maxtext_inference.sh
```diff
@@ -0,0 +1,215 @@
...
+    echo "Override the location with GEMMA3_MT_CKPT if the checkpoint has moved." >&2
+    exit 1
+fi
+
+echo "--- Recording pre-install versions"
+TORCH_BEFORE="$(python3 -c 'import torch; print(torch.__version__)' 2>/dev/null || echo unknown)"
+echo "torch before install: ${TORCH_BEFORE}"
+
+echo "--- Installing ${MAXTEXT_SPEC}"
+CONSTRAINTS_FILE="$(mktemp)"
```
pv97, line 138: nit: trap the cleanup 

```
CONSTRAINTS_FILE="$(mktemp)"
trap 'rm -f "${CONSTRAINTS_FILE}"' EXIT
```

tests/e2e/test_gemma3_maxtext_inference.sh
```diff
@@ -0,0 +1,215 @@
...
+    scan_layers=false \
+    enable_single_controller=false \
+    2>&1 | tee "${LOG_FILE}"
+set +x
+
+# `set -o pipefail` above means a non-zero exit from vllm_decode already fails
+# the test. As an extra correctness guard, require that the run actually emitted
+# generated text rather than exiting 0 after an early no-op.
+echo "--- Verifying decode produced output"
+if ! grep -qiE "output|generated|completion" "${LOG_FILE}"; then
```
pv97: Is this condition strict enough to avoid false positives?

tests/e2e/test_gemma3_maxtext_inference.sh
wenxindongwork: i set the nightly to soft-fail: true, so i hope it will not block our release

tests/e2e/test_gemma3_maxtext_inference.sh
wenxindongwork: thanks for the catch!

pv97 (approved):

yiw-wang (commented):
Im not sure if this is the best place to put such tests as MaxText seems like a downstream user to the tpu-inference, why not add a nightly build on MaxText side.

tests/e2e/test_gemma3_maxtext_inference.sh
QiliangCui: +1. 

I think we should not rely on maxtext healthness.

PR #3255, opened 2026-07-23
[DCP][Batched RPA] Enable DCP for batched RPA kernel

weiyu0824:
# Description

Mirror https://github.com/vllm-project/tpu-inference/pull/2842. 
The difference between this PR and and https://github.com/vllm-project/tpu-inference/pull/2842 is that in this PR, we restrict cp_kv_cache_interleave_size to block_size to avoid using additional kv_shuffle_vmem and save vmem.

This PR integrates DCP into the batched RPA kernel.

#### AttentionScope
other than `FULL`, this PR introduced
- `NEW_TOKENS_ONLY`: all ranks attend to new tokens identically and write their slice of new KV to cache
- `CACHE_ONLY`: each rank attends to its own local cache partition; results are merged across ranks using 

#### Scheduler: 
- `HeadAlongSublaneDmaNewCP`: new DMA descriptor with separate fetch/writeback flags. Writeback is conditional on per-page ownership under interleaved paging (p_idx % cp_group_size == rank), so no CP-specific logic is needed in bref_override. 
- compute_metadata: 
   - `CACHE_ONLY` shrinks k_len and end_k_idx to this rank's local_cache_len, switching k_idx from global to LOCAL  coordinates. 
   - `NEW_TOKENS_ONLY` skips pure-cache k blocks via start_k_idx.
– fill_dma_kv_new: CP path maps global token index to interleaved page layout and sets wb_val = 0 for pages not owned by this rank.

#### Kernel changes (kernel.py, flash_attention.py):
- sequence_len (renamed from effective_kv_len): bound only on q. 
- cache_len: per-sequence local cache length for `CACHE_ONLY` mask (kv_idx < cache_len)
- kv_new_start: global kv_start_pos for `NEW_TOKENS_ONLY` mask (kv_idx >= kv_new_start)
- lse output: needed for cross-rank merge return when return_lse=True

#### New utilities:
-  `cp_local_cache_len`: computes per-rank token count under interleaved page allocation

#### Plan:
- Introduced addition kv_cache_lens parameters, this can be used for sharded_cache (cp) and gemma-shared KV layers. so that we don't need update_kv_cache parameter.

# Tests
### Unit test
```bash
# Both head-on-sublane and seq-on-lane
USE_BATCHED_RPA_KERNEL=1 python3 tests/kernels/rpa_v3_cp/ragged_paged_attention_kernel_dcp_test.py
USE_BATCHED_RPA_SEQ_ON_LANE=1 USE_BATCHED_RPA_KERNEL=1 python3 tests/kernels/rpa_v3_cp/ragged_paged_attention_kernel_dcp_test.py

```

### Offline inference 
it does output the correct results
```bash
# Command
# Test head-on-sublane
USE_BATCHED_RPA_KERNEL=1 NEW_MODEL_DESIGN=1 python3 examples/offline_inference.py --max-model-len=9216 --max-num-batched-tokens=256 --max-num-seqs=128 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --decode-context-parallel-size 2 --model Qwen/Qwen3-235B-A22B-Instruct-2507 --block-size 128

# Test seq-on-lane
USE_BATCHED_RPA_SEQ_ON_LANE=1 USE_BATCHED_RPA_KERNEL=1 NEW_MODEL_DESIGN=1 python3 examples/offline_inference.py --max-model-len=9216 --max-num-batched-tokens=256 --max-num-seqs=128 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --decode-context-parallel-size 2 --model Qwen/Qwen3-235B-A22B-Instruct-2507 --block-size 128

# Test Non-cp pass
USE_BATCHED_RPA_KERNEL=1 NEW_MODEL_DESIGN=1 python3 examples/offline_inference.py --max-model-len=9216 --max-num-batched-tokens=256 --max-num-seqs=128 --no-enable-prefix-caching --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --model Qwen/Qwen3-235B-A22B-Instruct-2507 --block-size 128

```

### E2E correctness test: 
```bash
# Server
USE_BATCHED_RPA_KERNEL=1 NEW_MODEL_DESIGN=1 vllm serve Qwen/Qwen3-235B-A22B-Instruct-2507  --max-model-len=132072 --max-num-seqs=128 --gpu-memory-utilization=0.95 --tensor-parallel-size=8 --decode-context-parallel-size 2 --block_size 128

# Bench
lm_eval --model local-chat-completions     --model_args base_url="http://localhost:8000/v1/chat/completions",model="Qwen/Qwen3-235B-A22B-Instruct-2507",eos_string="<|im_end|>",num_concurrent=64     --tasks gsm8k     --apply_chat_template     --log_samples     --output_path ~/eval_results
```

TP8 DCP2
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8097|±  |0.0108|
|     |       |strict-match    |     5|exact_match|↑  |0.7051|±  |0.0126|

TP8 
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8120|±  |0.0108|
|     |       |strict-match    |     5|exact_match|↑  |0.7036|±  |0.0126|

# Checklist

### DIFF
--- a/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
+++ b/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
@@ -250,11 +250,13 @@ class KVBufferedRefHeadAlongSublane(_BypassRef):
 
     def copy_in(
         self,
-        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref],
+        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref,
+                       jax.Ref | None, jax.Ref | None],
         grid_indices: tuple[int | jax.Array, ...],
     ):
-        # src_ref: (kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref)
-        kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref = src_ref
+        # src_ref: (kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref,
+        #           [cp_rank_ref, kv_shuffle_ref]) last two only when CP is set.
+        kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref = src_ref[:4]
         slot = self.current_copy_in_slot
         sem = self.sem_recvs.at[slot]
         block_idx = jnp.maximum(grid_indices[0], 0)
@@ -302,41 +304,103 @@ class KVBufferedRefHeadAlongSublane(_BypassRef):
 
     def copy_out(
         self,
-        dst_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref],
+        dst_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref,
+                       jax.Ref | None, jax.Ref | None],
         grid_indices: tuple[int | jax.Array, ...],
     ):
-        kv_out_ref, _, schedule_ref, page_indices_ref = dst_ref
+        kv_out_ref, _, schedule_ref, page_indices_ref, cp_rank_ref, kv_shuffle_ref = dst_ref
+        kv_out_ref_flat = kv_out_ref.reshape(-1, *kv_out_ref.shape[2:])
         slot = self.current_copy_out_slot
         sem = self.sem_sends.at[slot]
         block_idx = grid_indices[0]
+        cp_group_size = self.cfgs.serve.cp_group_size
 
-        kv_out_ref_flat = kv_out_ref.reshape(-1, *kv_out_ref.shape[2:])
         vmem_src = self.window_ref.at[slot, :, :, :self.cfgs.kv_hbm_stride]
 
-        for b in range(self.cfgs.batch_size):
-            do_writeback = schedule_ref.do_writeback[block_idx, b] == 1
-            for i in range(self.cfgs.bkv_p_new):
-                dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
-                encoded_dst_hbm_off = dma_entry.wb_hbm[...]
-                src_vmem_off = dma_entry.wb_vmem[...]
-                new_sz = dma_entry.wb_val
-                global_p_idx = encoded_dst_hbm_off >> self.cfgs.serve.page_size_log2
-                p_off = encoded_dst_hbm_off & self.cfgs.serve.page_size_mask
-                dst_hbm_off = (page_indices_ref[global_p_idx] <<
-                               self.cfgs.serve.page_size_log2) | p_off
-                sz = jnp.where(do_writeback, new_sz, 0)
-                pltpu.make_async_copy(
-                    vmem_src.at[b, pl.ds(src_vmem_off, sz)],
-                    kv_out_ref_flat.at[pl.ds(dst_hbm_off, sz)],
-                    sem,
-                ).start()
+        if cp_group_size is not None and self.cfgs.serve.update_kv_cache:
+            cp_rank = cp_rank_ref[0]
+            # Static capacity: ceiling-div of block size by group size.
+            n_strided = pl.cdiv(self.cfgs.bkv_sz, cp_group_size)
+            for b in range(self.cfgs.batch_size):
+                do_writeback = schedule_ref.do_writeback[block_idx, b] == 1
+
+                # first_idx: offset within [0, new_sz) where this rank's tokens
+                # start. new_tok_off tells which rank owns the first new token.
+                new_tok_off = schedule_ref.new_tok_off[block_idx, b]
+                first_idx = (cp_rank - new_tok_off +
+                             cp_group_size) % cp_group_size
+
+                # Calculate total new tokens in this block
+                global_new_sz = jnp.zeros((), jnp.int32)
+                for i in range(self.cfgs.bkv_p_new):
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    global_new_sz = global_new_sz + dma_entry.wb_val
+                local_sz_total = jnp.maximum(
+                    0, (global_new_sz - first_idx + cp_group_size - 1) //
+                    cp_group_size)
+
+                # strided VMEM gather per batch slot to kv_shuffle_ref[slot, b]..
+                src_vmem_off_base = schedule_ref.dma_kv_new[block_idx, b,
+                                                            0].wb_vmem[...]
+                vmem_u32 = vmem_src.at[b].bitcast(jnp.uint32)
+                shuf_u32 = kv_shuffle_ref.at[slot, b].bitcast(jnp.uint32)
+                shuf_u32[pl.ds(0, n_strided)] = vmem_u32[
+                    pl.ds(src_vmem_off_base +
+                          first_idx, n_strided, cp_group_size),
+                    :self.cfgs.kv_hbm_stride,
+                ]
+
+                # DMA each page from the shuffle buffer.
+                local_done = jnp.zeros((), jnp.int32)
+                for i in range(self.cfgs.bkv_p_new):
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    encoded_dst_hbm_off = dma_entry.wb_hbm[...]
+                    global_p_idx = (
+                        encoded_dst_hbm_off >> self.cfgs.serve.page_size_log2)
+                    p_off = (encoded_dst_hbm_off
+                             & self.cfgs.serve.page_size_mask)
+                    dst_hbm_off = (page_indices_ref[global_p_idx] <<
+                                   self.cfgs.serve.page_size_log2) | p_off
+
+                    slot_local_sz = jnp.maximum(
+                        0,
+                        jnp.minimum(
+                            self.cfgs.serve.page_size,
+                            local_sz_total - i * self.cfgs.serve.page_size))
+                    local_sz = jnp.where(do_writeback, slot_local_sz, 0)
+                    pltpu.make_async_copy(
+                        kv_shuffle_ref.at[slot, b,
+                                          pl.ds(local_done, local_sz)],
+                        kv_out_ref_flat.at[pl.ds(dst_hbm_off, local_sz)],
+                        sem,
+                    ).start()
+                    local_done = local_done + local_sz
+        else:
+            for b in range(self.cfgs.batch_size):
+                do_writeback = schedule_ref.do_writeback[block_idx, b] == 1
+                for i in range(self.cfgs.bkv_p_new):
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    encoded_dst_hbm_off = dma_entry.wb_hbm[...]
+                    src_vmem_off = dma_entry.wb_vmem[...]
+                    new_sz = dma_entry.wb_val
+                    global_p_idx = encoded_dst_hbm_off >> self.cfgs.serve.page_size_log2
+                    p_off = encoded_dst_hbm_off & self.cfgs.serve.page_size_mask
+                    dst_hbm_off = (page_indices_ref[global_p_idx] <<
+                                   self.cfgs.serve.page_size_log2) | p_off
+                    sz = jnp.where(do_writeback, new_sz, 0)
+                    pltpu.make_async_copy(
+                        vmem_src.at[b, pl.ds(src_vmem_off, sz)],
+                        kv_out_ref_flat.at[pl.ds(dst_hbm_off, sz)],
+                        sem,
+                    ).start()
 
     def wait_in(
         self,
-        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref],
+        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref,
+                       jax.Ref | None, jax.Ref | None],
         grid_indices: tuple[int | jax.Array, ...],
     ):
-        _, _, schedule_ref, _ = src_ref
+        _, _, schedule_ref, _ = src_ref[:4]
         slot = self.current_wait_in_slot
         sem = self.sem_recvs.at[slot]
         vmem_dst = self.window_ref.at[slot]
@@ -363,22 +427,45 @@ class KVBufferedRefHeadAlongSublane(_BypassRef):
 
     def wait_out(
         self,
-        dst_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref],
+        dst_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref,
+                       jax.Ref | None, jax.Ref | None],
         grid_indices: tuple[int | jax.Array, ...],
     ):
-        kv_out_ref, _, schedule_ref, page_indices_ref = dst_ref
+        kv_out_ref, _, schedule_ref, _, cp_rank_ref, _ = dst_ref
         slot = self.current_wait_out_slot
         sem = self.sem_sends.at[slot]
         block_idx = grid_indices[0]
+        cp_group_size = self.cfgs.serve.cp_group_size
 
         total_sz = 0
         for b in range(self.cfgs.batch_size):
             do_writeback = schedule_ref.do_writeback[block_idx, b] == 1
-            for i in range(self.cfgs.bkv_p_new):
-                dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
-                new_sz = dma_entry.wb_val
-                sz = jnp.where(do_writeback, new_sz, 0)
-                total_sz += sz
+            if cp_group_size is not None and self.cfgs.serve.update_kv_cache:
+                # Mirror copy_out
+                new_tok_off = schedule_ref.new_tok_off[block_idx, b]
+                first_idx = ((cp_rank_ref[0] - new_tok_off + cp_group_size) %
+                             cp_group_size)
+                global_new_sz = jnp.zeros((), jnp.int32)
+                for i in range(self.cfgs.bkv_p_new):
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    global_new_sz = global_new_sz + dma_entry.wb_val
+                local_sz_total = jnp.maximum(
+                    0, (global_new_sz - first_idx + cp_group_size - 1) //
+                    cp_group_size)
+                for i in range(self.cfgs.bkv_p_new):
+                    slot_local_sz = jnp.maximum(
+                        0,
+                        jnp.minimum(
+                            self.cfgs.serve.page_size,
+                            local_sz_total - i * self.cfgs.serve.page_size))
+                    sz = jnp.where(do_writeback, slot_local_sz, 0)
+                    total_sz += sz
+            else:
+                for i in range(self.cfgs.bkv_p_new):
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    new_sz = dma_entry.wb_val
+                    sz = jnp.where(do_writeback, new_sz, 0)
+                    total_sz += sz
 
         # Flatten to 2D: (Total_Rows, Head_Dim)
         flat_ref = kv_out_ref.reshape((-1, *kv_out_ref.shape[2:]))
--- a/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
+++ b/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
@@ -482,6 +482,102 @@ class BatchingORef(pltpu.BufferedRef):
 
 @jax.tree_util.register_dataclass
 @dataclasses.dataclass(frozen=True)
+class BatchingLSERef(pltpu.BufferedRef):
+    """Handles writing LSE values to HBM, overlapped with compute via double buffering.
+
+    dst_ref tuple: (lse_hbm_ref, schedule_ref, cu_q_lens_ref).
+    VMEM block shape matches cfgs.lm_scratch_shape:
+      (batch_size, num_kv_heads, bq_sz * model.num_q_heads_per_kv_head, num_lanes).
+    """
+
+    cfgs: configs.RpaConfigs = dataclasses.field(default=None,
+                                                 metadata=dict(static=True))
+
+    @classmethod
+    def create(
+        cls,
+        spec: pl.BlockSpec,
+        dtype_or_type: jax.Array,
+        buffer_type,
+        buffer_count: int,
+        use_lookahead: bool,
+        cfgs: configs.RpaConfigs,
+    ):
+        standard_ref = pltpu.BufferedRef.create(
+            spec=spec,
+            dtype_or_type=dtype_or_type,
+            buffer_type=buffer_type,
+            buffer_count=buffer_count,
+            grid_rank=1,
+            use_lookahead=use_lookahead,
+        )
+        return cls(
+            cfgs=cfgs,
+            **{
+                f.name: getattr(standard_ref, f.name)
+                for f in dataclasses.fields(pltpu.BufferedRef)
+            },
+        )
+
+    def copy_out(
+        self,
+        dst_ref: tuple[jax.Ref, schedule.RpaSchedule, jax.Ref],
+        grid_indices: tuple[int | jax.Array, ...],
+    ):
+        lse_hbm, schedule_ref, cu_q_lens_ref = dst_ref
+        slot = self.current_copy_out_slot
+        sem = self.sem_sends.at[slot]
+        vmem_src = self.window_ref.at[slot]
+        block_idx = grid_indices[0]
+
+        dma_list = []
+        for b in range(self.cfgs.batch_size):
+            is_last_k = schedule_ref.is_last_k[block_idx, b] == 1
+            s_idx = schedule_ref.s_idx[block_idx, b]
+            q_idx = schedule_ref.q_idx[block_idx, b]
+            _, q_sz = schedule_ref.get_dma_q(block_idx, b)
+            q_src_flat = pl.multiple_of(
+                (cu_q_lens_ref[s_idx] + q_idx * self.cfgs.bq_sz) * self.cfgs.model.num_q_heads_per_kv_head, 8)
+            q_sz_flat = pl.multiple_of(q_sz * self.cfgs.model.num_q_heads_per_kv_head, 8)
+            q_sz_flat = jnp.where(is_last_k, q_sz_flat, 0)
+            dma_list.append((q_src_flat, q_sz_flat, b))
+
+        for i in range(len(dma_list)):
+            q_src_flat, q_sz_flat, b = dma_list[i]
+            pltpu.make_async_copy(
+                vmem_src.at[b, :, pl.ds(0, q_sz_flat), :],
+                lse_hbm.at[:, pl.ds(q_src_flat, q_sz_flat), :],
+                sem,
+            ).start()
+
+    def wait_out(
+        self,
+        dst_ref: tuple[jax.Ref, schedule.RpaSchedule, jax.Ref],
+        grid_indices: tuple[int | jax.Array, ...],
+    ):
+        lse_hbm, schedule_ref, _ = dst_ref
+        slot = self.current_wait_out_slot
+        sem = self.sem_sends.at[slot]
+        block_idx = grid_indices[0]
+
+        total_sz = 0
+        for b in range(self.cfgs.batch_size):
+            is_last_k = schedule_ref.is_last_k[block_idx, b] == 1
+            _, q_sz = schedule_ref.get_dma_q(block_idx, b)
+            q_sz_flat = pl.multiple_of(q_sz * self.cfgs.model.num_q_heads_per_kv_head, 8)
+            total_sz += jnp.where(is_last_k, q_sz_flat, 0)
+
+        flat_ref = lse_hbm.reshape((-1, lse_hbm.shape[-1]))
+        total_rows = pl.multiple_of(total_sz * lse_hbm.shape[0], 8)
+        pltpu.make_async_copy(
+            flat_ref.at[pl.ds(0, total_rows)],
+            flat_ref.at[pl.ds(0, total_rows)],
+            sem,
+        ).wait()
+
+
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass(frozen=True)
 class BatchingQRef(pltpu.BufferedRef):
     """Handles fetching Q blocks using precomputed metadata."""
 
--- a/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
+++ b/tpu_inference/kernels/experimental/batched_rpa/bref_override.py
@@ -250,13 +250,11 @@ class KVBufferedRefHeadAlongSublane(_BypassRef):
 
     def copy_in(
         self,
-        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref,
-                       jax.Ref | None, jax.Ref | None],
+        src_ref: tuple[jax.Ref, jax.Ref, schedule.RpaSchedule, jax.Ref],
         grid_indices: tuple[int | jax.Array, ...],
     ):
-        # src_ref: (kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref,
-        #           [cp_rank_ref, kv_shuffle_ref]) last two only when CP is set.
-        kv_cache_hbm, new_kv_hbm, schedule_ref, page_indices_ref = src_ref[:4]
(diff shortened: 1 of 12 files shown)


tpu_inference/layers/common/attention_interface.py
```diff
@@ -454,7 +455,7 @@ def _ragged_paged_attention(*args):
         # is a no-op so we don't forward it to the hd64 signature.
         if not use_hd64:
             kwargs["update_kv_cache"] = update_kv_cache
-            kwargs["use_causal_mask"] = use_causal_mask
+            # kwargs["use_causal_mask"] = use_causal_mask
```
wenxindongwork: nit: do you want to delete this line?

tpu_inference/layers/common/attention_interface.py
```diff
@@ -553,19 +554,21 @@ def _ragged_paged_attention_wrapper(*args):
...
+            q_scale=q_scale,
+            k_scale=k_scale,
+            v_scale=v_scale,
+            cp_rank=cp_rank,
+            cp_group_size=cp_group_size,
+            update_kv_cache=update_kv_cache,
+            skip_cache_attn=skip_cache_attn,
+            skip_current_attn=skip_current_attn,
+            return_lse=return_lse,
+            #   use_causal_mask=use_causal_mask
```
wenxindongwork: why is this line commented out?

tpu_inference/kernels/experimental/batched_rpa/flash_attention.py
```diff
@@ -28,8 +28,10 @@ def flash_attention_qk_softmax(
     processed_q_len: list[jax.Array],  # [B]
     processed_kv_len: list[jax.Array],  # [B]
     effective_kv_len: list[jax.Array],  # [B]
+    kv_cache_len_local: list[jax.Array],  # [B]
     cfgs: configs.RpaConfigs,
     bq_start: int,
+    cp_rank: jax.Array | None = None,
```
wenxindongwork: is cp_rank needed here?

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -454,28 +511,37 @@ def rpa_kernel(
...
         # Inputs.
         schedule_hbm_ref: schedule.RpaSchedule,
         q_hbm_ref: jax.Ref,
         new_kv_hbm_ref: jax.Ref,
         kv_cache_hbm_ref: jax.Ref,
+        lse_hbm_ref: jax.Ref | None,
         # Outputs.
         o_hbm_ref: jax.Ref,
         o_kv_cache_hbm_ref: jax.Ref,
+        o_lse_hbm_ref: jax.Ref | None = None,
```
wenxindongwork, line 549: is this var needed here?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -460,9 +516,22 @@ def q_loop(q_idx, _, *, s_idx, q_start, q_end, k_len, q_len, num_k):
...
         q_end = cu_q_lens_ref[s_idx + 1]
-        k_len = kv_lens_ref[s_idx]
+        k_len_global = kv_lens_ref[s_idx]
         q_len = q_end - q_start
 
+        global_cache_len = k_len_global - q_len
+
+        # Convert to LOCAL kv_len for CP: cache is sharded (1/cp_group_size per rank),
+        # but new KV is NOT sharded (all ranks hold all q_len new tokens).
+        if cfgs.serve.cp_group_size is not None and cp_rank_ref is not None:
```
wenxindongwork: when would one be none and the other one not none?

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
wenxindongwork, line 268: can you add descriptions in the comments for 

```
cp_group_size: int | None = None,
    cp_rank: jax.Array | None = None,
    skip_cache_attn: bool = False,
    skip_current_attn: bool = False,
    return_lse: bool = False,
```

wenxindongwork, line 287: should we also add a guard somewhere rejecting `cp_group_size is not None `together with `KVLayout.SEQ_ALONG_LANE`?

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
```diff
@@ -333,10 +346,23 @@ def ragged_paged_attention(
...
     )
 
+    # Pre-allocate LSE buffer.
+    lse_hbm_init: jax.Array | None = None
+    if return_lse:
+        gqh = num_q_heads // num_kv_heads
+        num_lanes = pltpu.get_tpu_info().num_lanes
+        max_tokens = queries.shape[0]
+        lse_hbm_init = jnp.full(
+            [num_kv_heads, max_tokens * gqh, num_lanes],
```
wenxindongwork, line 359: curiosu if you have experimented with different shapes to get rid of the num_lanes padding?

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
```diff
@@ -302,41 +304,103 @@ def copy_in(
...
-                sz = jnp.where(do_writeback, new_sz, 0)
-                pltpu.make_async_copy(
-                    vmem_src.at[b, pl.ds(src_vmem_off, sz)],
-                    kv_out_ref_flat.at[pl.ds(dst_hbm_off, sz)],
-                    sem,
-                ).start()
+        if cp_group_size is not None:
+            cp_rank = cp_rank_ref[0]
+            # Static capacity: ceiling-div of block size by group size.
+            n_strided = (self.cfgs.bkv_sz + cp_group_size - 1) // cp_group_size
```
wenxindongwork: for these kind of ceiling div ops, can we use `pl.cdiv`?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -172,6 +172,10 @@ class RpaSchedule:
     k_idx: SmemWrapper  # [steps, batch]
     is_last_k: SmemWrapper  # [steps, batch]
     do_writeback: SmemWrapper  # [steps, batch]
+    # global_cache_len % cp_group_size; 0 when CP is disabled.  Tells
+    # copy_out which rank owns the first new token in this block so the
+    # strided shuffle reads the correct interleaved offset.
+    new_tok_off: SmemWrapper  # [steps, batch]
```
wenxindongwork: is the intention of introducing this variable to reduce scalar computation inside the kernel? since we can also derive the new token offset from global cache length inside the kernel itself.

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
```diff
@@ -302,41 +304,103 @@ def copy_in(
...
+                                   self.cfgs.serve.page_size_log2) | p_off
+
+                    slot_local_sz = jnp.maximum(
+                        0,
+                        jnp.minimum(
+                            self.cfgs.serve.page_size,
+                            local_sz_total - i * self.cfgs.serve.page_size))
+                    local_sz = jnp.where(do_writeback, slot_local_sz, 0)
+                    pltpu.make_async_copy(
+                        kv_shuffle_ref.at[slot, b,
```
wenxindongwork: would it be clearer to be shuf_u32 instead of kv_shuffle_ref here?

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
```diff
@@ -363,22 +427,45 @@ def wait_in(
...
+                    0, (global_new_sz - first_idx + cp_group_size - 1) //
+                    cp_group_size)
+                for i in range(self.cfgs.bkv_p_new):
+                    slot_local_sz = jnp.maximum(
+                        0,
+                        jnp.minimum(
+                            self.cfgs.serve.page_size,
+                            local_sz_total - i * self.cfgs.serve.page_size))
+                    sz = jnp.where(do_writeback, slot_local_sz, 0)
+                    total_sz += sz
```
wenxindongwork: do you think we can put some of the wait_out and copy_out shared logic into a helper function?

tpu_inference/layers/common/attention_interface.py
wenxindongwork: actually, why are we removing the use_causal_mask flag?

weiyu0824: I think batched rpa already remove this flag. but rpa still contain this flag. I probably have to add this back.

tpu_inference/layers/common/attention_interface.py
weiyu0824: Add back.

tpu_inference/layers/common/attention_interface.py
weiyu0824: add this back.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
weiyu0824: I think so, just like  kv_cache_hbm_ref and o_kv_cache_hbm_ref. even though we don't use o_kv_cache_hbm_ref

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
weiyu0824: Added.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
weiyu0824: Do you mean reorder [num_kv_heads, max_tokens * gqh, num_lanes] ?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: This two arguments should present together. removed cp_rank_ref check

tpu_inference/kernels/experimental/batched_rpa/flash_attention.py
weiyu0824: Removed.

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
weiyu0824: Good suggestion. changed to pl.cdiv()

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Yes, this is the intention, this is also what I did in rpa_v3_cp kernel. the update_kv_cache_partial() fn contain additional variable.

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
```diff
@@ -302,41 +304,103 @@ def copy_in(
...
+                    dma_entry = schedule_ref.dma_kv_new[block_idx, b, i]
+                    encoded_dst_hbm_off = dma_entry.wb_hbm[...]
+                    global_p_idx = (
+                        encoded_dst_hbm_off >> self.cfgs.serve.page_size_log2)
+                    p_off = (encoded_dst_hbm_off
+                             & self.cfgs.serve.page_size_mask)
+                    dst_hbm_off = (page_indices_ref[global_p_idx] <<
+                                   self.cfgs.serve.page_size_log2) | p_off
+
+                    slot_local_sz = jnp.maximum(
```
kyuyeunk: this is very difficult to read

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -109,18 +113,42 @@ def _accum(b_idx: int):
...
+            q_sz_flat = pl.multiple_of(
+                q_sz_task * cfgs.model.num_q_heads_per_kv_head, 8)
+            m_scratch_ref[b_idx] = lse_val.astype(cfgs.serve.dtype_out)
+            cp = pltpu.make_async_copy(
+                m_scratch_ref.at[b_idx, :, pl.ds(0, q_sz_flat), :],
+                lse_hbm_ref.at[:, pl.ds(q_src_flat, q_sz_flat), :],
+                lse_dma_sem_ref.at[0],
+            )
+            cp.start()
+            cp.wait()
```
kyuyeunk: ummm there is no buffering? that doesn't look good?

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -109,18 +113,42 @@ def _accum(b_idx: int):
...
         # on every block. In decode, because of the large block sizes / and or
         # batch sizes, we almost always use accumulation on every block. Please
         # tune `fuse_accum` for your use case.
         if not cfgs.fuse_accum:
-            is_last_k = schedule_ref.is_last_k[step_idx, b] == 1
             jax.lax.cond(is_last_k, jax.named_call(_accum, name="accum"),
                          lambda _: None, b)
         else:
             _accum(b)
+        jax.lax.cond(is_last_k, _write_lse, lambda _: None, b)
```
kyuyeunk: this will cause performance regression for all workloads.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
```diff
@@ -262,12 +268,20 @@ def ragged_paged_attention(
         debug_mode: Not used.
         out_dtype: Dtype of output. Defaults to dtype of queries.
         use_causal_mask: Not used.
+        cp_group_size: Size of the context parallelism (CP) group. KV cache is sharded across devices in this group. Defaults to None.
+        cp_rank: Rank of the current device within the CP group, which determine the token ownership. Defaults to None. 
+        skip_cache_attn: If True, skip attention over the tokens in kv cache. Defaults to False.
```
kyuyeunk: w.r.t. `skip_cache_attn` and `skip_current_attn`. is there a use case where both can be true? if not, does it make sense to have it as enum instead?

similar to `return_lse`. is that value effectively automatically determined by `skip_cache_attn` or `skip_current_attn`?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -348,9 +355,38 @@ def k_loop(
...
 
+        # Pre-compute CP layout quantities used in both dma_kv_new branches.
+        # j_start: new tokens before this block in [0, q_len).
+        # j_start_rank: rank's new tokens before this block.
+        # new_tok_off_block: which rank owns the first new token in this block.
+        if cfgs.serve.cp_group_size is not None:
+            cp_group_size = cfgs.serve.cp_group_size
+            rank = cp_rank_ref[0]
+            local_cache_len = k_len - q_len
+            j_start = jnp.maximum(0, kv_len_start - local_cache_len)
```
kyuyeunk: what is the purpose of `j` prefix?

tpu_inference/kernels/experimental/batched_rpa/utils.py
```diff
@@ -24,6 +24,14 @@ def align_to(a, b):
     return pl.cdiv(a, b) * b
 
 
+def cp_local_cache_len(global_cache_len, cp_group_size, cp_rank):
```
kyuyeunk: out of curiosity, what happens if you just pass the kernel a precompute value of `cp_local_cache_len` and not global value of `kv_lens`? and if that's the case, doesn't it also remove the need for `cp_size` and `cp_rank` and just completely reuse the existing batched rpa logic?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -637,4 +716,4 @@ def generate_rpa_metadata(
         ),
         interpret=interpret,
         name="rpa_metadata_schedule",
-    )(cu_q_lens, kv_lens, distribution)
+    )(*scalar_prefetches)
```
kyuyeunk: as i mentioned in other pr, let's not do something like this.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -615,18 +687,25 @@ def generate_rpa_metadata(
     distribution: jax.Array,
     cfgs: configs.RpaConfigs,
     *,
+    cp_rank: jax.Array | None = None,
     interpret=False,
     update_kv_cache: bool = True,
 ) -> RpaSchedule:
     schedule_shaped_dtype = RpaSchedule.create_shape_dtype(cfgs)
+    # Same reason as rpa_kernel: None has 0 pytree leaves, so pallas would steal
+    # the first schedule output leaf as the 4th scalar. Use a dummy array instead.
```
kyuyeunk: i don't think this is a good approach. i think it's better to just reduce `num_scalar_prefetch` to 1 if there is `None` vs. allocating memory that is completely unused.

kyuyeunk (commented):
@wenxindongwork would it be okay for me to rely on your to do a first pass review? you can ping me if you think the pr is in a good state. i'm juggling between many code reviews so any help i can get to reduce them would be greatly appreciated.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
weiyu0824: Thanks for this suggestion. Added a attention_scope enum and removed this 2 parameters, skip_cache_attn and skip_current_attn.   I think return_lse is not automatically decide. For example if everything is write to kv_cache first, we still need to attend to cache only, but we don't have to return lse.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: No problem. This is my mistake. Fixed.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Fixed.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
weiyu0824: Thanks for pointing out. Guarded this line under return_lse.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
weiyu0824: This is my concern too. lse is very small, do we really have to create a double buffer for it? @wenxindongwork Any thoughts on this suggestion?

tpu_inference/kernels/experimental/batched_rpa/utils.py
weiyu0824: The main reason why we have to pass cp_size and cp_rank is because we have to do strided write for new_kv. But I think it is possible that we don't have to passed this 2 parameters to RPA, instead just passed it to scheduler kernel. I will investigate this more.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Rewrite the schedule, so don't need this part anymore.

tpu_inference/kernels/experimental/batched_rpa/bref_override.py
weiyu0824: Use new methods, so wait_out and copy_out almost rely on existing bref_overide.py infra.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Removed this part.

tpu_inference/kernels/experimental/batched_rpa/configs.py
```diff
@@ -308,6 +325,9 @@ def kv_vmem_shape(self):
     def dma_kv_new_size(self) -> int:
         if self.serve.kv_layout == KVLayout.SEQ_ALONG_LANE:
             return 5
+        # HEAD_ALONG_SUBLANE
+        if self.serve.cp_group_size is not None:
```
wenxindongwork: instead of commenting can you be explict about it in the check? 

`if self.serve.kv_layout == KVLayout.HEAD_ALONG_SUBLANE and self.serve.cp_group_size is not None:`

tpu_inference/layers/common/cp_attention.py
```diff
@@ -68,10 +78,27 @@ def _rpa_cp_call(
...
+        from tpu_inference.kernels.experimental.batched_rpa import \
+            configs as brpa_configs
+        if skip_cache_attn:
+            scope = brpa_configs.AttentionScope.NEW_TOKENS_ONLY
+        elif skip_current_attn:
+            scope = brpa_configs.AttentionScope.CACHE_ONLY
+        else:
+            scope = brpa_configs.AttentionScope.FULL
+        flags['attention_scope'] = scope
+        flags['use_causal_mask'] = True
```
wenxindongwork, line 97: why do we need to pass in `use_causal_mask` for batched rpa but not for rpa v3?

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
wenxindongwork: yes, because currently we only need `[num_kv_heads, max_tokens * gqh, 1]` but we are using `[num_kv_heads, max_tokens * gqh, num_lanes] `, is my understanding correct?

tpu_inference/kernels/experimental/batched_rpa/configs.py
```diff
@@ -412,3 +432,19 @@ def validate_inputs(
...
+                )
+            if self.serve.attention_scope == AttentionScope.FULL:
+                raise ValueError(
+                    "Context Parallel does not support AttentionScope.FULL where cache is sharded but current tokens is sequential"
+                )
+        # Attention Scope
+        if self.serve.kv_layout == KVLayout.SEQ_ALONG_LANE and self.serve.attention_scope != AttentionScope.FULL:
+            raise ValueError(
+                f"SEQ_ALONG_LANE do not supports {self.serve.attention_scope} yet."
+            )
```
wenxindongwork: should we also add a check for `sliding_window` + CP?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -201,9 +236,11 @@ def create_shape_dtype(cls, cfgs: configs.RpaConfigs):
                     cfgs.batch_size,
                     cfgs.bkv_p_new,
                 ),
-                struct_cls=(SeqAlongLaneDmaNew if cfgs.serve.kv_layout
-                            == configs.KVLayout.SEQ_ALONG_LANE else
-                            HeadAlongSublaneDmaNew),
+                struct_cls=(
```
wenxindongwork: this is a bit hard to read, can we move this logic to the func body ?

tpu_inference/kernels/experimental/batched_rpa/configs.py
weiyu0824: OK.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -146,10 +177,13 @@ def rpa_body(
     # Step 1: Fetch metadata.
     processed_q_len = []
     processed_kv_len = []
-    effective_kv_len = []
+    sequence_len = []
```
wenxindongwork, line 162: asked Weiyu offline regarding this renaming: the previous name (effective_kv_len) is confusing because it is just the sequence length.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -146,10 +177,13 @@ def rpa_body(
...
     # Lists to hold the 2 variables needed for stitching
     bkv_sz_frm_cache_list = []
     new_kv_len_start_list = []
+    # List to support AttentionScope.CACHE_ONLY and AttentionScope.NEW_TOKENS_ONLY.
+    kv_new_start = []  # for new tokens.
+    cache_len = []  # for cache only.
```
wenxindongwork, line 168: can we use cache_len for both `CACHE_ONLY` and `NEW_TOKENS_ONLY`. It seems like there is no need to create two lists where one would always be empty.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -101,6 +101,41 @@ def set_flags(self, fetch_val, wb_val):
...
+
+    @property
+    def wb_val(self):
+        return self._wb_flags[...]
+
+    @property
+    def wb_vmem(self):
+        return self.fetch_vmem
+
+    def set_flags(self, fetch_val, wb_val):
```
wenxindongwork, line 134: we also can do some bit shifting  logic inside`set_flag` and reuse `HeadAlongSublaneDmaNew` instead of creating a new dataclass for CP. 

judgement call for Kyuyeun.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -431,16 +501,41 @@ def q_loop(q_idx, _, *, s_idx, q_start, q_end, k_len, q_len, num_k):
...
 
-        end_k_idx_causal = (k_len - q_len + q_idx * cfgs.bq_sz + q_sz_task -
+        end_k_idx_causal = (cache_len + q_idx * cfgs.bq_sz + q_sz_task -
                             1) // cfgs.bkv_sz + 1
         end_k_idx = jnp.minimum(num_k, end_k_idx_causal)
 
+        if cfgs.serve.attention_scope == configs.AttentionScope.NEW_TOKENS_ONLY:
+            # Skip pure-cache blocks; start at the first block containing new tokens.
+            start_k_idx = jnp.maximum(start_k_idx,
+                                      (k_len - q_len) // cfgs.bkv_sz)
```
wenxindongwork: can we use `cache_len// cfgs.bkv_sz` here?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -386,18 +435,39 @@ def fill_dma_kv_new(i, dst_vmem, dma_sz, slot_start):
...
+                if cfgs.serve.cp_group_size is not None:
+                    # NOTE: p_idx should be relaxed
+                    # (cfgs.serve.pages_per_seq * cfgs.serve.cp_group_size) to avoid clamp.
+                    p_idx = jnp.minimum(
+                        tok_idx >> cfgs.serve.page_size_log2,
+                        cfgs.serve.pages_per_seq * cfgs.serve.cp_group_size -
+                        1,
+                    )
+                    p_off = tok_idx & cfgs.serve.page_size_mask
+                    local_slot = jnp.minimum(
```
wenxindongwork: do we need this `jnp.minimum` clamping for local_slot?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -386,18 +435,39 @@ def fill_dma_kv_new(i, dst_vmem, dma_sz, slot_start):
...
-                    (kv_len_start + dst_vmem) >> cfgs.serve.page_size_log2,
-                    cfgs.serve.pages_per_seq - 1,
-                )
-                p_off = (kv_len_start + dst_vmem) & cfgs.serve.page_size_mask
-                dst_hbm = ((s_idx * cfgs.serve.pages_per_seq + p_idx) <<
-                           cfgs.serve.page_size_log2) | p_off
+                tok_idx = kv_len_start + dst_vmem
+                if cfgs.serve.cp_group_size is not None:
+                    # NOTE: p_idx should be relaxed
+                    # (cfgs.serve.pages_per_seq * cfgs.serve.cp_group_size) to avoid clamp.
```
wenxindongwork: what does this note mean?

tpu_inference/kernels/experimental/batched_rpa/schedule.py
```diff
@@ -386,18 +435,39 @@ def fill_dma_kv_new(i, dst_vmem, dma_sz, slot_start):
...
+                        jnp.int32(0))
+                else:
+                    p_idx = jnp.minimum(
+                        tok_idx >> cfgs.serve.page_size_log2,
+                        cfgs.serve.pages_per_seq - 1,
+                    )
+                    p_off = tok_idx & cfgs.serve.page_size_mask
+                    dst_hbm = ((s_idx * cfgs.serve.pages_per_seq + p_idx) <<
+                               cfgs.serve.page_size_log2) | p_off
+                    wb_val = dma_sz
```
wenxindongwork, line 496: can we merge these two branches into one? 

for nonCP path, cp_group_size=1

tpu_inference/kernels/experimental/batched_rpa/kernel.py
```diff
@@ -168,9 +202,22 @@ def rpa_body(
...
+            if (cfgs.serve.cp_group_size is not None):
+                cp_group_size = cfgs.serve.cp_group_size
+                rank = cp_rank_ref[0]
+                local_cache_len = utils.cp_local_cache_len(
+                    offset, cp_group_size, rank, cfgs.serve.page_size)
+            # We add this to make sure flash_attention.py can mask non-cache tokens out.
+            cache_len.append(local_cache_len)
+        elif cfgs.serve.attention_scope == configs.AttentionScope.NEW_TOKENS_ONLY:
+            kv_new_start.append(kv_len - q_len)
+
```
wenxindongwork, line 199: seems like we can delete this entire block if pass in `local_kv_cache_lens` and `global_kv_pos_offset`

tpu_inference/layers/common/cp_attention.py
weiyu0824: Batched RPA do not support use_causal_mask=False.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
weiyu0824: yes.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Since we already have a guard above for p_idx, so we don't need this here. Thanks.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: I'd perfor not to merge them together, because cp_group_size=None actually means we don't need any cp-related stuff. In addition, in cp mode, wb_val and dma_sz are different, so I think the current separation is more self-explanatory.

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Done

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: Done

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: I update the note. Basically, it means that previous clamping is incorrect in cp_mode. 
```
# NOTE: tok_idx is a global token index, but cfgs.serve.pages_per_seq
# is the calculated from num_page_indices. The global upper
# bound for tok_idx is pages_per_seq * cp_group_size.
```

tpu_inference/kernels/experimental/batched_rpa/schedule.py
weiyu0824: OK, done.

tpu_inference/kernels/experimental/batched_rpa/configs.py
weiyu0824: Added.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
wenxindongwork: I think double buffer is a good idea

weiyu0824: Fixed. Add double buffer supports.

tpu_inference/kernels/experimental/batched_rpa/kernel.py
weiyu0824: discuss offline, think current implementation is still a better.

tpu_inference/kernels/experimental/batched_rpa/wrapper.py
weiyu0824: Yes, you are correct. I think my understanding is that we have to copy entire lane from vmem to hbm, so current approach is a tradeoff, especially in decode where we only have 1 token.

PR #3257, opened 2026-07-23
[RL] VllmSampler rollout class

khatwanimohit:
# Description

This PR introduces a standalone, high-throughput rollout sampling engine (`VllmSampler`) inside `tpu-inference` for Reinforcement Learning

The `VllmSampler` implementation satisfies the Tunix `Sampler` Protocol defined in [`tunix/experimental/sampler/sampler.py`](https://github.com/google/tunix)

### Key Features
1. **Dynamic Attribute Extraction (`VllmSampler.sample`)**:
   - Accepts raw strings, dictionaries, or duck-typed sampling request objects.
   - Extracts sampling parameters (`max_tokens`, `temperature`, `top_p`, `top_k`, `stop_sequences`) and returns `.text`, `.token_ids`, `.logprobs`, `.cumulative_logprob`, and `.route_key`.
   - Supports sticky prefix-cache routing headers (`route_key`) for multi-turn rollout turns.
2. **Integration with `TPUWorker` Weight Sync Lifecycle**:
   - **`pre_weight_sync()`**: Pauses engine background loops and calls `TPUWorker.start_weight_update(free_kv_cache=True)`.
   - **`weight_sync()`**: Executes sharded weights DMA via `TPUWorker.update_weights()`.
   - **`post_weight_sync()`**: Finalizes parameter updates via `TPUWorker.finish_weight_update()` and resumes inference intake.
3. server script for launching tpu_inference.rl.vllm_sampler as a server

# Tests

- Added `tests/rl/test_vllm_sampler.py`
- Agentic benchmarking result: https://paste.googleplex.com/4658655990513664   

# Checklist

### DIFF
--- a/tpu_inference/rl/vllm_sampler.py
+++ b/tpu_inference/rl/vllm_sampler.py
@@ -1,0 +1,463 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the tpu-inference project
+"""Asynchronous vLLM Sampler engine implementation for tpu-inference.
+
+This module provides a standalone vLLM serving and rollout sampler (`RLVllmSampler`)
+tailored for Reinforcement Learning (RL) workloads on TPUs.
+Satisfies the open-source Tunix `Sampler` Protocol defined in:
+https://github.com/google/tunix/blob/main/tunix/experimental/sampler/sampler.py
+without importing or depending on Tunix.
+"""
+
+import asyncio
+import logging
+import os
+import time
+
+os.environ["VLLM_USE_V1"] = "0"
+from types import SimpleNamespace
+from typing import Any
+
+import numpy as np
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+from vllm.sampling_params import SamplingParams as VllmSamplingParams
+
+logger = logging.getLogger(__name__)
+
+
+def _get_val(obj: Any, key: str, default: Any = None) -> Any:
+    """Helper to extract an attribute or dict key seamlessly from any duck-typed request."""
+    if obj is None:
+        return default
+    if isinstance(obj, dict):
+        return obj.get(key, default)
+    return getattr(obj, key, default)
+
+
+class RLVllmSampler:
+    """Asynchronous vLLM sampler for RL inside `tpu-inference`.
+
+  Satisfies the open-source Tunix `Sampler` Protocol without importing Tunix.
+  Handles:
+
+    - Direct binding to vLLM's `AsyncLLMEngine`.
+    - Dynamic attribute extraction (`_get_val`) for arbitrary request inputs.
+    - Non-blocking asynchronous batch generation (`sample`).
+    - Sticky routing header support (`route_key`) for multi-turn prefix cache hits.
+    - Native `TPUWorker` 3-phase weight synchronization (`pre_weight_sync`, `weight_sync`, `post_weight_sync`).
+  """
+
+    def __init__(self, engine_args: AsyncEngineArgs):
+        self.engine_args = engine_args
+        self._engine: Any | None = None
+        self._is_running = False
+        self._is_paused = False
+        self._cache_valid = True
+        self._mesh: Any | None = None
+        self._transfer_statuses: dict[str, str] = {}
+        self._policy_version = 0
+
+    def _get_tpu_workers(self) -> list[Any]:
+        """Retrieves active TPUWorker instances from underlying model executor."""
+        if self._engine is None:
+            return []
+        llm_engine = getattr(self._engine, "engine", None)
+        model_executor = getattr(llm_engine, "model_executor",
+                                 None) if llm_engine else None
+        if model_executor is None:
+            return []
+        workers = getattr(model_executor, "workers", None)
+        if workers:
+            return list(workers)
+        driver_worker = getattr(model_executor, "driver_worker", None)
+        return [driver_worker] if driver_worker else []
+
+    async def start(self, **kwargs: Any) -> None:
+        """Initializes the vLLM engine and execution environment."""
+        if self._is_running:
+            logger.warning("RLVllmSampler is already running.")
+            return
+
+        logger.info("Initializing RLVllmSampler with model: %s",
+                    self.engine_args.model)
+
+        self._engine = AsyncLLMEngine.from_engine_args(self.engine_args)
+        self._is_running = True
+
+        init_info = {
+            "model_path": self.engine_args.model,
+            "tp_size": self.engine_args.tensor_parallel_size,
+        }
+        if self._engine:
+            try:
+                await self._engine.init_weight_transfer_engine(init_info)
+            except Exception as e:
+                logger.warning(
+                    "Failed to initialize weight transfer engine: %s", e)
+
+        logger.info("RLVllmSampler started successfully.")
+
+    async def stop(self, **kwargs: Any) -> None:
+        """Stops the sampler and releases resources."""
+        if not self._is_running:
+            return
+        logger.info("Stopping RLVllmSampler...")
+        self._is_paused = True
+        self._engine = None
+        self._is_running = False
+        logger.info("RLVllmSampler stopped.")
+
+    async def pause(self, **kwargs: Any) -> None:
+        """Pauses request intake and drains active batch iterations during weight updates."""
+        if self._is_paused:
+            return
+        self._is_paused = True
+        # Note: vLLM's pause_background_loop stops the scheduler from taking new requests from the queue.
+        # Ongoing requests in the batch will be completed or drained depending on internal vLLM state.
+        if self._engine and hasattr(self._engine, "pause_background_loop"):
+            logger.info(
+                "Pausing RLVllmSampler inference intake for weight sync...")
+            await self._engine.pause_background_loop()
+        await asyncio.sleep(0.01)
+
+    async def resume(self, **kwargs: Any) -> None:
+        """Resumes inference processing after weight sync completion."""
+        if not self._is_paused:
+            return
+        if self._engine and hasattr(self._engine, "resume_background_loop"):
+            logger.info("Resuming RLVllmSampler inference serving...")
+            await self._engine.resume_background_loop()
+        self._is_paused = False
+
+    async def get_mesh(self, **kwargs: Any) -> Any | None:
+        """Returns the JAX device mesh."""
+        return self._mesh
+
+    # ----------------------------------------------------------------------------
+    # Helper Methods for Sample Stream Processing
+    # ----------------------------------------------------------------------------
+
+    def _build_vllm_params(
+        self,
+        req: Any,
+        kwargs: dict[str, Any],
+    ) -> VllmSamplingParams:
+        """Builds a vLLM SamplingParams object from any duck-typed request."""
+        sparams = _get_val(req, "sampling_params")
+        return VllmSamplingParams(
+            temperature=_get_val(sparams, "temperature",
+                                 kwargs.get("temperature", 0.7)),
+            top_p=_get_val(sparams, "top_p", kwargs.get("top_p", 0.95)),
+            top_k=_get_val(sparams, "top_k", kwargs.get("top_k", -1)),
+            max_tokens=_get_val(sparams, "max_tokens",
+                                kwargs.get("max_tokens", 128)),
+            stop=_get_val(sparams, "stop_sequences")
+            or _get_val(sparams, "stop") or kwargs.get("stop"),
+            logprobs=1
+            if _get_val(sparams, "return_logprobs",
+                        kwargs.get("return_logprobs", False)) else None,
+        )
+
+    async def _process_request_output(
+        self,
+        req_id: str,
+        task: Any,
+    ) -> SimpleNamespace:
+        """Consumes an AsyncLLMEngine output stream and formats output result."""
+        try:
+            final_output = None
+            async for step_output in task:
+                final_output = step_output
+
+            if final_output and final_output.outputs:
+                output_choice = final_output.outputs[0]
+                text = output_choice.text
+                token_ids_arr = np.array(output_choice.token_ids,
+                                         dtype=np.int32)
+                cum_logprob = float(
+                    getattr(output_choice, "cumulative_logprob", 0.0) or 0.0)
+
+                logprobs_arr = None
+                if getattr(output_choice, "logprobs", None):
+                    logprob_vals = [
+                        lp_dict[next(iter(lp_dict.keys()))].logprob
+                        if lp_dict else 0.0
+                        for lp_dict in output_choice.logprobs
+                    ]
+                    logprobs_arr = np.array(logprob_vals, dtype=np.float32)
+
+                routed_experts = getattr(output_choice, "routed_experts", None)
+                if routed_experts is not None and not isinstance(
+                        routed_experts, np.ndarray):
+                    routed_experts = np.array(routed_experts)
+
+                return SimpleNamespace(
+                    request_id=req_id,
+                    text=text,
+                    token_ids=token_ids_arr,
+                    logprobs=logprobs_arr,
+                    cumulative_logprob=cum_logprob,
+                    routed_experts=routed_experts,
+                    finish_reason=getattr(output_choice, "finish_reason",
+                                          "stop") or "stop",
+                    error=None,
+                )
+
+            err_obj = SimpleNamespace(
+                error_type="EmptyOutput",
+                message="No output generated by vLLM",
+                retryable=False,
+            )
+            return SimpleNamespace(
+                request_id=req_id,
+                text="",
+                token_ids=np.zeros(0, dtype=np.int32),
+                logprobs=None,
+                cumulative_logprob=0.0,
+                routed_experts=None,
+                finish_reason="stop",
+                error=err_obj,
+            )
+        except Exception as e:
+            logger.exception("Error generating sampling result for req_id=%s",
+                             req_id)
+            err_obj = SimpleNamespace(
+                error_type=type(e).__name__,
+                message=str(e),
+                retryable=True,
+            )
+            return SimpleNamespace(
+                request_id=req_id,
+                text="",
+                token_ids=np.zeros(0, dtype=np.int32),
+                logprobs=None,
+                cumulative_logprob=0.0,
+                routed_experts=None,
+                finish_reason="stop",
+                error=err_obj,
+            )
+
+    # ----------------------------------------------------------------------------
+    # Dynamic Batch Sampling
+    # ----------------------------------------------------------------------------
+
+    async def sample(
+        self,
+        sampling_requests: Any,
+        **kwargs: Any,
+    ) -> Any:
+        """Generates completions for a batch of sampling requests or raw prompts.
+
+    Accepts raw string lists, dictionaries, or duck-typed objects from callers
+    (such as `tunix.experimental.common.datatypes.SamplingRequest`).
+    """
+        while self._is_paused:
+            await asyncio.sleep(0.05)
+
+        if not self._is_running or self._engine is None:
+            await self.start()
+
+        raw_input_mode = False
+        if isinstance(sampling_requests, (str, list)) and (
+                isinstance(sampling_requests, str) or not sampling_requests
+                or isinstance(sampling_requests[0], str) or
+            (isinstance(sampling_requests[0], list)
+             and not _get_val(sampling_requests[0], "prompt"))):
+            raw_input_mode = True
+            items = sampling_requests if isinstance(
+                sampling_requests, list) else [sampling_requests]
+            req_list = [{
+                "prompt": item,
+                "request_id": f"req_{idx}_{time.time_ns()}",
+            } for idx, item in enumerate(items)]
+        elif _get_val(sampling_requests, "prompt") is not None:
+            req_list = [sampling_requests]
+        else:
+            req_list = list(sampling_requests)
+
+        pending_tasks = []
+        for idx, req in enumerate(req_list):
+            vllm_params = self._build_vllm_params(req, kwargs)
+            req_id = _get_val(req,
+                              "request_id") or f"req_{time.time_ns()}_{idx}"
+            prompt_val = _get_val(req, "prompt")
+            prompt_text = prompt_val if isinstance(prompt_val,
+                                                   str) else str(prompt_val)
+
+            task_gen = self._engine.generate(prompt_text,
+                                             vllm_params,
+                                             request_id=req_id)
+            pending_tasks.append((req_id, task_gen))
+
+        results = await asyncio.gather(*[
+            self._process_request_output(req_id, task_gen)
+            for req_id, task_gen in pending_tasks
+        ])
+
+        if raw_input_mode:
+            return [r.text for r in results]
+        return list(results)
+
+    # ----------------------------------------------------------------------------
+    # Cache Management
+    # ----------------------------------------------------------------------------
+
+    async def _clear_prefix_cache(self) -> None:
+        """Purges prefix caching blocks from HBM."""
+        logger.info("Clearing vLLM RadixAttention prefix cache...")
+        self._cache_valid = False
+        if self._engine and hasattr(self._engine, "reset_prefix_cache"):
+            await self._engine.reset_prefix_cache()
+
+    # ----------------------------------------------------------------------------
+    # Weight Synchronization (Integrated with TPUWorker)
+    # ----------------------------------------------------------------------------
+
+    async def get_weight_sync_metadata(self, **kwargs: Any) -> dict[str, Any]:
+        """Returns PyTree of weight sharding rules, dtype, and shape across devices."""
+        return self.get_weight_metadata()
+
+    def get_weight_metadata(self) -> dict[str, Any]:
+        """Synchronous alias for get_weight_sync_metadata."""
+        worker_ips = [
+            getattr(w, "ip_address", getattr(w, "ip", "127.0.0.1"))
+            for w in self._get_tpu_workers()
+        ]
+        tp_size = getattr(self.engine_args, "tensor_parallel_size", 1)
+        dp_size = getattr(self.engine_args, "data_parallel_size", 1)
+        ep_size = getattr(self.engine_args, "expert_parallel_size", 1)
+        return {
+            "sharding": {
+                "tensor_parallel_size": tp_size,
+                "data_parallel_size": dp_size,
+                "expert_parallel_size": ep_size,
+            },
+            "dtype": str(getattr(self.engine_args, "dtype", "bfloat16")),
+            "model_path": getattr(self.engine_args, "model", ""),
+            "policy_version": self._policy_version,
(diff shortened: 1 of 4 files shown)


tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+        **self.config.additional_engine_args,
+    )
+    self._engine = AsyncLLMEngine.from_engine_args(engine_args)
+    self._is_running = True
+
+    init_info = {
+        "model_path": self.config.model_path,
+        "tp_size": self.config.tensor_parallel_size,
+    }
+    for w in self._get_tpu_workers():
```
wang2yn84: I think vLLM provide high level API for init_weight_transfer_engine and other 3?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+        temperature=_get_val(sparams, "temperature", defaults["temperature"]),
+        top_p=_get_val(sparams, "top_p", defaults["top_p"]),
+        top_k=_get_val(sparams, "top_k", -1),
+        max_tokens=_get_val(sparams, "max_tokens", defaults["max_tokens"]),
+        stop=_get_val(sparams, "stop_sequences") or _get_val(sparams, "stop") or None,
+        prompt_logprobs=1 if _get_val(sparams, "return_logprobs", defaults["return_logprobs"]) else None,
+        logprobs=1 if _get_val(sparams, "return_logprobs", defaults["return_logprobs"]) else None,
+    )
+
+  async def _process_stream_output(
```
wang2yn84: I'd probably avoid using "stream" because it's a bit confusing that we are not using streaming mode.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+  ) -> VllmSamplingParams:
+    """Builds a vLLM SamplingParams object from any duck-typed request."""
+    sparams = _get_val(req, "sampling_params")
+    return VllmSamplingParams(
+        temperature=_get_val(sparams, "temperature", defaults["temperature"]),
+        top_p=_get_val(sparams, "top_p", defaults["top_p"]),
+        top_k=_get_val(sparams, "top_k", -1),
+        max_tokens=_get_val(sparams, "max_tokens", defaults["max_tokens"]),
+        stop=_get_val(sparams, "stop_sequences") or _get_val(sparams, "stop") or None,
+        prompt_logprobs=1 if _get_val(sparams, "return_logprobs", defaults["return_logprobs"]) else None,
```
wang2yn84: We should remove this option, we don't need prompt_logprobs at all.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+        or isinstance(sampling_requests[0], str)
+        or (isinstance(sampling_requests[0], list) and not _get_val(sampling_requests[0], "prompt"))
+    ):
+      raw_input_mode = True
+      items = sampling_requests if isinstance(sampling_requests, list) else [sampling_requests]
+      req_list = [
+          {
+              "prompt": item,
+              "request_id": f"req_{idx}_{time.time_ns()}",
+              "route_key": kwargs.get("route_key"),
```
wang2yn84: Let's put some comments here, "route key" is used for rollout manager talk to remote sampler.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+    elif _get_val(sampling_requests, "prompt") is not None:
+      req_list = [sampling_requests]
+    else:
+      req_list = list(sampling_requests)
+
+    defaults = {
+        "max_tokens": kwargs.get("max_tokens", 128),
+        "temperature": kwargs.get("temperature", 0.7),
+        "top_p": kwargs.get("top_p", 0.95),
+        "return_logprobs": kwargs.get("return_logprobs", True),
```
wang2yn84: It should default to False.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+    self._kv_cache_valid = False
+    if self._engine and hasattr(self._engine, "reset_prefix_cache"):
+      await self._engine.reset_prefix_cache()
+    self._kv_cache_valid = True
+
+  # ----------------------------------------------------------------------------
+  # Weight Synchronization (Integrated with TPUWorker)
+  # ----------------------------------------------------------------------------
+
+  async def get_weight_sync_metadata(self, **kwargs: Any) -> dict[str, Any]:
```
wang2yn84: dedup?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+    """Returns PyTree of weight sharding rules, dtype, and shape across devices."""
+    return {
```
wang2yn84: The weight sync metadata has been updated after our last meeting. Ptal. We need the ip address for all the TPU workers.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+      cid = cid or _get_val(sync_request, "controller_id")
+      self._policy_version = _get_val(sync_request, "policy_version", self._policy_version)
+
+    ip = ip or "10.0.0.1"
+    logger.info("Executing pre_weight_sync from Trainer IP: %s (policy_version=%d)", ip, self._policy_version)
+    self._src_controller_ip = ip
+    if cid:
+      self._dst_controller_id = cid
+
+    await self.pause()
```
wang2yn84: What will happen if there are on going requests in vLLM when we pause?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+      sync_request: Any = None,
+      src_controller_ip: str | None = None,
+      dst_controller_id: str | None = None,
+      free_kv_cache: bool = True,
+      **kwargs: Any,
+  ) -> None:
+    """Phase 1: Pauses intake, resets prefix cache, and calls start_weight_update()."""
+    ip = src_controller_ip
+    cid = dst_controller_id
+    if sync_request is not None:
```
wang2yn84: So sync_reqeust also contains id and ip? Is it redundant? I wrap them in WeightSyncRequest in the doc.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+    logger.info("Executing pre_weight_sync from Trainer IP: %s (policy_version=%d)", ip, self._policy_version)
+    self._src_controller_ip = ip
+    if cid:
+      self._dst_controller_id = cid
+
+    await self.pause()
+    await self.clear_cache()
+
+    for w in self._get_tpu_workers():
+      if hasattr(w, "start_weight_update"):
```
wang2yn84: ditto, do we have top level APIs?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+
+  # ----------------------------------------------------------------------------
+  # Cache Management
+  # ----------------------------------------------------------------------------
+
+  async def clear_cache(self) -> None:
+    """Purges KV-cache and prefix caching blocks from HBM."""
+    logger.info("Clearing vLLM PagedAttention KV cache and prefix cache...")
+    self._kv_cache_valid = False
+    if self._engine and hasattr(self._engine, "reset_prefix_cache"):
```
wang2yn84: We probably should also clear kv cache here?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,450 @@
...
+  async def pre_weight_sync(
+      self,
+      sync_request: Any = None,
+      src_controller_ip: str | None = None,
+      dst_controller_id: str | None = None,
+      free_kv_cache: bool = True,
+      **kwargs: Any,
+  ) -> None:
+    """Phase 1: Pauses intake, resets prefix cache, and calls start_weight_update()."""
+    ip = src_controller_ip
```
wang2yn84: We don't need to pass sampler controller ip and id to trainer for weight sync any more. The Raiden controller setup has been simplified. We only need to get the TPU worker ip now.

tpu_inference/rl/vllm_sampler.py
khatwanimohit: I have used weight transfer APIs which were added in [PR#3231](https://github.com/vllm-project/tpu-inference/pull/3231/changes)

tpu_inference/rl/vllm_sampler.py
khatwanimohit: Done. Completely removed the prompt_logprobs config mapping

tpu_inference/rl/vllm_sampler.py
khatwanimohit: Fixed! I renamed the internal parsing function to _process_request_output

tpu_inference/rl/vllm_sampler.py
khatwanimohit: Added a helpful inline comment clarifying the purpose of the route_key parameter for sticky routing from the Rollout Worker to the Remote Sampler

tpu_inference/rl/vllm_sampler.py
khatwanimohit: Fixed

tpu_inference/rl/vllm_sampler.py
khatwanimohit: fixed

tpu_inference/rl/vllm_sampler.py
khatwanimohit: Added! I injected a tpu_worker_ips list field into the metadata payload that dynamically pulls the active IPs from the workers list

tpu_inference/rl/vllm_sampler.py
khatwanimohit: vllm's pause_background_loop() is called blocks the scheduler from dispatching new incoming requests off its incoming queue, while ongoing generation requests are systematically drained

tpu_inference/rl/vllm_sampler.py
khatwanimohit: added in https://github.com/vllm-project/tpu-inference/pull/3231/changes

tpu_inference/rl/vllm_sampler.py
khatwanimohit: I see the doc has controller_id, does it refer to the source controller ip ? if so, should I remove src_controller_ip fully ?

tpu_inference/rl/vllm_sampler.py
khatwanimohit: done!

tpu_inference/rl/vllm_sampler.py
khatwanimohit: removed

tpu_inference/rl/vllm_sampler.py
wang2yn84: I'm thinking of calling the engine level API: https://vllm.ai/blog/2026-05-28-native-rl-apis, instead of the TPU worker level. They exposed here: https://github.com/search?q=repo%3Avllm-project%2Fvllm%20start_weight_update&type=code

tpu_inference/rl/vllm_sampler.py
wang2yn84: Do they wait until the request is completed? Or do they return partial rollout result?

tpu_inference/rl/vllm_sampler.py
wang2yn84: We don't need the source controller ip any more. Only port is needed.

tpu_inference/rl/vllm_sampler.py
khatwanimohit: moved to using APIs provided by engine

tpu_inference/rl/vllm_sampler.py
khatwanimohit: done

tpu_inference/rl/vllm_sampler.py
khatwanimohit: The request will be kept in the queue and won't be removed from the queue unless `abort_request()` is called

tpu_inference/rl/offline_benchmark.py
```diff
@@ -0,0 +1,197 @@
+# SPDX-License-Identifier: Apache-2.0
```
wenxindongwork: we already have https://github.com/vllm-project/tpu-inference/blob/main/scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py, wondering if we still need this script?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,435 @@
...
+  weight_dtype: str = "bfloat16"
+  additional_engine_args: dict[str, Any] = field(default_factory=dict)
+
+
+# ==============================================================================
+# Concrete VllmSampler Implementation (Duck-Typed for any RL orchestrator)
+# ==============================================================================
+
+
+class VllmSampler:
```
wenxindongwork: could we rename `VllmSampler` to `TunixVllmSampler` or something like that to make it clear that this is a sampler interface satisfying the Tunix requirement?

tpu_inference/rl/vllm_sampler.py
wang2yn84: Thank you for the information.

tpu_inference/rl/vllm_sampler.py
khatwanimohit: renamed to RLVllmSampler

tpu_inference/rl/offline_benchmark.py
khatwanimohit: removed this script and tested out with agentic benchmarking script

wenxindongwork (approved):

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,435 @@
...
+  return getattr(obj, key, default)
+
+
+# ==============================================================================
+# Configuration for RLVllmSampler
+# ==============================================================================
+
+
+@dataclass
+class VllmSamplerConfig:
```
wenxindongwork: nit: could we rename this to `RLVllmSamplerConfig`?

and curious do need need this class? it seems like it maps 1-to-1 to `AsyncEngineArgs`. 

would it be possible to do something like

```
class RLVllmSampler:
  def __init__(self, engine_args: AsyncEngineArgs):
    self.engine_args = engine_args

  async def start(self, **kwargs):
    self._engine = AsyncLLMEngine.from_engine_args(self.engine_args)
```
?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,435 @@
...
+
+  async def start(self, **kwargs: Any) -> None:
+    """Initializes the vLLM engine and execution environment."""
+    if self._is_running:
+      logger.warning("RLVllmSampler is already running.")
+      return
+
+    logger.info("Initializing RLVllmSampler with model: %s", self.config.model_path)
+
+    engine_args = AsyncEngineArgs(
```
wenxindongwork: should `data_parallel_size, expert_parallel_size, and enable_continue_decode` also be passed in here?

tpu_inference/rl/vllm_sampler.py
khatwanimohit: cool, changed it to using `AsyncEngineArgs`

tpu_inference/rl/vllm_sampler.py
khatwanimohit: using AsyncEngineArgs now

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+        or isinstance(sampling_requests[0], str)
+        or (isinstance(sampling_requests[0], list) and not _get_val(sampling_requests[0], "prompt"))
+    ):
+      raw_input_mode = True
+      items = sampling_requests if isinstance(sampling_requests, list) else [sampling_requests]
+      req_list = [
+          {
+              "prompt": item,
+              "request_id": f"req_{idx}_{time.time_ns()}",
+              # route_key is used by the rollout manager to talk to a specific remote sampler
```
wang2yn84: Router key is no longer needed based on the latest design.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+    logger.info("Clearing vLLM PagedAttention KV cache and prefix cache...")
+    self._kv_cache_valid = False
+    # Explicitly clear KV cache using tpu-inference TPUWorker method
+    for w in self._get_tpu_workers():
+      if hasattr(w, "delete_kv_cache"):
+        w.delete_kv_cache()
+
+    if self._engine and hasattr(self._engine, "reset_prefix_cache"):
+      await self._engine.reset_prefix_cache()
+    self._kv_cache_valid = True
```
wang2yn84: Cache is no longer valid until we reinit them, so should keep this signal as False.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+    - Sticky routing header support (`route_key`) for multi-turn prefix cache hits.
+    - Native `TPUWorker` 3-phase weight synchronization (`pre_weight_sync`, `weight_sync`, `post_weight_sync`).
+  """
+
+  def __init__(self, engine_args: AsyncEngineArgs):
+    self.engine_args = engine_args
+    self._engine: Any | None = None
+    self._is_running = False
+    self._is_paused = False
+    self._kv_cache_valid = True
```
wang2yn84: Consider rename to _cache_valid because we have kv cache and prefix cache?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+    """Returns PyTree of weight sharding rules, dtype, and shape across devices."""
+    return self.get_weight_metadata()
+
+  def get_weight_metadata(self) -> dict[str, Any]:
+    """Synchronous alias for get_weight_sync_metadata."""
+    worker_ips = [
+        getattr(w, "ip_address", getattr(w, "ip", "127.0.0.1"))
+        for w in self._get_tpu_workers()
+    ]
+    tp_size = getattr(self.engine_args, "tensor_parallel_size", 1)
```
wang2yn84: We also need EP here, basically all the sharding information.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+    dp_size = getattr(self.engine_args, "data_parallel_size", 1)
+    return {
+        "sharding": f"{tp_size}x{dp_size}",
+        "dtype": str(getattr(self.engine_args, "dtype", "bfloat16")),
+        "model_path": getattr(self.engine_args, "model", ""),
+        "policy_version": self._policy_version,
+        "layers_valid": self._kv_cache_valid,
```
wang2yn84: Do we need this?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+  ) -> None:
+    """Phase 1: Pauses intake, resets prefix cache, and calls start_weight_update()."""
+    self._policy_version = _get_val(sync_request, "policy_version", self._policy_version)
+    logger.info("Executing pre_weight_sync (policy_version=%d)", self._policy_version)
+
+    await self.pause()
+    await self.clear_cache()
+
+    if self._engine:
+      self._engine.start_weight_update(free_kv_cache=free_kv_cache)
```
wang2yn84: If we set free_kv_cache here, do we still need to reset the prefix cache in clear_cache()?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+
+    self._kv_cache_valid = False
```
wang2yn84: Don't need this one any more

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,404 @@
...
+    logger.info("Executing weight_sync update on TPU workers...")
+    u_info = update_info or _get_val(sync_request, "extra_config") or {}
+    if self._engine:
+      self._engine.update_weights(u_info)
+    await asyncio.sleep(0.01)
+
+  async def post_weight_sync(
+      self,
+      sync_request: Any = None,
+      req_id: str | None = None,
```
wang2yn84: We don't have this field in the API contract?

tpu_inference/rl/vllm_sampler.py
khatwanimohit: discussed offline, we are going to keep one rolloutworker per sampler and routing is going to happen in orchestrator. removing this

tpu_inference/rl/vllm_sampler.py
khatwanimohit: moved `self._cache_valid = True` to post_weight_sync

tpu_inference/rl/vllm_sampler.py
khatwanimohit: done

tpu_inference/rl/vllm_sampler.py
khatwanimohit: done

tpu_inference/rl/vllm_sampler.py
wang2yn84: BTW, the sync_request should have the request_id field

tpu_inference/rl/vllm_sampler.py
khatwanimohit: start_weight_update only delete's the kv-cache today so keeping prefix cache clearing separately

tpu_inference/rl/vllm_sampler.py
khatwanimohit: removed req_id argument and directly taking from sync_request

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+        self._is_paused = True
+        self._engine = None
+        self._is_running = False
+        logger.info("RLVllmSampler stopped.")
+
+    async def pause(self, **kwargs: Any) -> None:
+        """Pauses request intake and drains active batch iterations during weight updates."""
+        if self._is_paused:
+            return
+        logger.info(
```
wang2yn84: If engine is not available due to stop(), but it still prints pause information, it's misleading. Consider only print it when calling pause_background_loop.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+        # Ongoing requests in the batch will be completed or drained depending on internal vLLM state.
+        if self._engine and hasattr(self._engine, "pause_background_loop"):
+            await self._engine.pause_background_loop()
+        await asyncio.sleep(0.01)
+
+    async def resume(self, **kwargs: Any) -> None:
+        """Resumes inference processing after weight sync completion."""
+        if not self._is_paused:
+            return
+        logger.info("Resuming RLVllmSampler inference serving...")
```
wang2yn84: Ditto

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+        elif _get_val(sampling_requests, "prompt") is not None:
+            req_list = [sampling_requests]
+        else:
+            req_list = list(sampling_requests)
+
+        defaults = {
+            "max_tokens": kwargs.get("max_tokens", 128),
+            "temperature": kwargs.get("temperature", 0.7),
+            "top_p": kwargs.get("top_p", 0.95),
+            "return_logprobs": kwargs.get("return_logprobs", False),
```
wang2yn84: These should come from sampling_request.sampling_params instead of kwargs.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+
+        if raw_input_mode:
+            return [r.text for r in results]
+        return list(results)
+
+    # ----------------------------------------------------------------------------
+    # Cache Management
+    # ----------------------------------------------------------------------------
+
+    async def clear_cache(self) -> None:
```
wang2yn84: This API clears prefix cache only. We are clearing KV cache in start_weight_update. Consider rename this API to _clear_prefix_cache() and keep it local, not public any more.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+        logger.info("Executing pre_weight_sync (policy_version=%d)",
+                    self._policy_version)
+
+        await self.pause()
+        await self.clear_cache()
+
+        if self._engine and hasattr(self._engine, "start_weight_update"):
+            self._engine.start_weight_update(free_kv_cache=free_kv_cache)
+
+        self._cache_valid = False
```
wang2yn84: Remove this.

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+
+    async def weight_sync(
+        self,
+        sync_request: Any = None,
+        update_info: dict[str, Any] | None = None,
```
wang2yn84: Do we still need this when we have sync_request.extra_config?

tpu_inference/rl/vllm_sampler.py
```diff
@@ -0,0 +1,411 @@
...
+            self._engine.finish_weight_update()
+
+        self._cache_valid = True
+        if rid:
+            self._transfer_statuses[rid] = "SUCCESS"
+        await self.resume()
+
+    async def get_transfer_status(self, req_id: Any, **kwargs: Any) -> str:
+        """Returns status of weight transfer or KV-cache migration request."""
+        return self._transfer_statuses.get(str(req_id), "SUCCESS")
```
wang2yn84: Default value should be something different from SUCCESS? Maybe "MISSING"?

khatwanimohit: done

tpu_inference/rl/vllm_sampler.py
khatwanimohit: directly taking it from sync_request only

tpu_inference/rl/vllm_sampler.py
khatwanimohit: checking self.engine before prints

wang2yn84 (approved):

vipannalla (approved):

tianshub:
@khatwanimohit thanks for working on this. 
A few high level questions (cc @wenxindongwork ): 
1. Compare to offline inference, anything special for RL workload besides weight sync?
2. Do you think it's a good practice to have the RL specific code live in tpu-infernece? would it be more difficult to change sampler code related to RL workflow in the future since there are multiple codebases?
3. Do you expect to use it anywhere outside of Trellis/Tunix dist orchestrator? if not do you think it worth to consider putting it directly next to orchestrator?

PR #3259, opened 2026-07-24
compressed tensor nvfp4 support improvement

lxhfirenking:
# Description

Increase the fidelity of underlying matmul behavior for compressed tensor checkpoints with nvfp4 weights

1. Only attempt to quantize activation during quantized matmul when the checkpoint config actually asks for it.
2. Add flag that allow user to requantize nvfp4 weights into a larger block size so that to avoid being fully dequantized in gmm_v2 due to the nvfp4's block size of 16.

# Tests

Unit tests.
Tested with nvfp4 quantized gemma4 checkpoint
```
USE_BATCHED_RPA_KERNEL=1 \
REQUANTIZE_COMPRESSED_TENSOR_NVFP4_BLOCK_SIZE=256 \
ENABLE_QUANTIZED_MATMUL_KERNEL=1 \
JITTED_MM_MODULE_KEYS="model.vision_tower.encoder" \
REGISTER_MM_MODULE_CUSTOM_PYTREE_CLASSES="transformers.modeling_outputs.BaseModelOutputWithPast" \
MODEL_IMPL_TYPE="vllm" \
vllm serve \
    --max-num-seqs=127 \
    --max-model-len=4096 \
    --no-enable-prefix-caching  \
    --model=RedHatAI/gemma-4-31B-it-NVFP4 \
    --limit-mm-per-prompt '{"image": 8, "video": 0}' \
    --tensor-parallel-size=2 \
    --kv-cache-dtype=fp8
```
Benchmark run
```
    --backend vllm-chat \
    --model RedHatAI/gemma-4-31B-it-NVFP4\
    --dataset-name mmmu_pro \
    --mmmu-pro-output-len=2000 \
    --run-eval \
    --num-prompts 200 \
    --max-concurrency 32
```
Result
```
Successful requests:                     200       
Benchmark duration (s):                  399.53    
Total input tokens:                      109015    
Total generated tokens:                  164395    
Request throughput (req/s):              0.50      
Output token throughput (tok/s):         411.47    
Total token throughput (tok/s):          684.32    
Mean TTFT (ms):                          9577.18   
Median TTFT (ms):                        1672.52   
P99 TTFT (ms):                           66840.00  
Mean TPOT (ms):                          68.08     
Median TPOT (ms):                        69.12     
P99 TPOT (ms):                           145.87    
Mean ITL (ms):                           64.43     
Median ITL (ms):                         9.37      
P99 ITL (ms):                            1504.09   
Mean E2EL (ms):                          62527.34  
Median E2EL (ms):                        44519.74  
P99 E2EL (ms):                           193340.94 
```
Results without nvfp4 block size changes
```
Successful requests:                     200       
Benchmark duration (s):                  529.31    
Total input tokens:                      109015    
Total generated tokens:                  128500    
Request throughput (req/s):              0.38      
Output token throughput (tok/s):         242.77    
Total token throughput (tok/s):          448.72    
Mean TTFT (ms):                          12293.43  
Median TTFT (ms):                        1727.48   
P99 TTFT (ms):                           77980.13  
Mean TPOT (ms):                          112.68    
Median TPOT (ms):                        116.48    
P99 TPOT (ms):                           192.41    
Mean ITL (ms):                           108.57    
Median ITL (ms):                         15.61     
P99 ITL (ms):                            1621.25   
Mean E2EL (ms):                          82049.97  
Median E2EL (ms):                        66919.90  
P99 E2EL (ms):                           252723.92 
```

### DIFF
--- a/tests/layers/vllm/test_compressed_tensors_w4a4_nvfp4.py
+++ b/tests/layers/vllm/test_compressed_tensors_w4a4_nvfp4.py
@@ -38,8 +38,8 @@ from tests.layers.common import utils as test_utils
 from tests.layers.vllm.nvfp4_utils import (NVFP4_GROUP_SIZE, quantize_to_nvfp4,
                                            ref_dequant_nvfp4)
 from tpu_inference.layers.vllm.quantization import get_tpu_quantization_config
-from tpu_inference.layers.vllm.quantization.compressed_tensors.schemes.compressed_tensors_w4a4_nvfp4 import \
-    VllmCompressedTensorsW4A4Fp4
+from tpu_inference.layers.vllm.quantization.compressed_tensors.schemes.compressed_tensors_w4a4_nvfp4 import (
+    VllmCompressedTensorsW4A4Fp4, W4A4ActivationType)
 from tpu_inference.layers.vllm.quantization.configs import \
     VllmQuantLinearConfig
 
@@ -80,8 +80,10 @@ def override_activation_quant_config(vllm_config):
     vllm_config.model_config.hf_text_config.quantization_config = vllm_config.model_config.hf_config.quantization_config
 
 
-def initialize_layer_weights(layer: torch.nn.Module,
-                             use_a16: bool = True) -> torch.Tensor:
+def initialize_layer_weights(
+    layer: torch.nn.Module,
+    activation_type: W4A4ActivationType = W4A4ActivationType.BF16
+) -> torch.Tensor:
     assert isinstance(layer, LinearBase)
     scheme = layer.scheme
     assert isinstance(scheme, VllmCompressedTensorsW4A4Fp4)
@@ -123,7 +125,7 @@ def initialize_layer_weights(layer: torch.nn.Module,
     layer.weight_global_scale = torch.nn.Parameter(weight_global_scale,
                                                    requires_grad=False)
 
-    if not use_a16:
+    if activation_type == W4A4ActivationType.NVFP4:
         layer.input_global_scale = torch.nn.Parameter(torch.tensor(
             [1.0], dtype=torch.float32),
                                                       requires_grad=False)
@@ -133,11 +135,13 @@ def initialize_layer_weights(layer: torch.nn.Module,
     return weight_ref
 
 
-def return_ref_and_layer_output(layer: torch.nn.Module,
-                                use_a16: bool = True,
-                                batch_size: int = 16):
+def return_ref_and_layer_output(
+        layer: torch.nn.Module,
+        activation_type: W4A4ActivationType = W4A4ActivationType.BF16,
+        batch_size: int = 16):
 
-    weight_ref = initialize_layer_weights(layer, use_a16=use_a16)
+    weight_ref = initialize_layer_weights(layer,
+                                          activation_type=activation_type)
     assert isinstance(layer, LinearBase)
     scheme = layer.scheme
     assert isinstance(scheme, VllmCompressedTensorsW4A4Fp4)
@@ -151,9 +155,9 @@ def return_ref_and_layer_output(layer: torch.nn.Module,
     input_tensor = input_tensor.to('cpu')
 
     # Run reference implementation
-    if not use_a16:
+    if activation_type == W4A4ActivationType.NVFP4:
         # Quantize activation to FP4
-        # Since use_a16=False tests FP4xFP4, we need to quantize the activation
+        # Since activation_type=NVFP4 tests FP4xFP4, we need to quantize the activation
         # However, TPU does not natively support FP4xFP4 and raises an error
         # So we skip actual activation quantization in reference for now
         pass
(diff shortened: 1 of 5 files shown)


tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a4_nvfp4.py
```diff
@@ -39,22 +43,33 @@
...
 
+class W4A4ActivationType(str, Enum):
+    BF16 = "bf16"
+    NVFP4 = "nvfp4"
+    FP8 = "fp8"
+
+
 class VllmCompressedTensorsW4A4Fp4(CompressedTensorsW4A4Fp4):
 
     def __init__(
```
jrplatin, line 61: Can you explain the rationale behind this a bit? I am just a little nervous because we are changing the expected interface here such that it won't be matching upstream vLLM any more -- I am wondering if there might be a cleaner way of doing these (e.g. using a new kwarg or something)

tpu_inference/envs.py
```diff
@@ -290,6 +290,11 @@ def _get_int_list_env() -> list[int]:
     "REQUANTIZE_BLOCK_SIZE":
     lambda: int(block_size) if
     (block_size := os.getenv("REQUANTIZE_BLOCK_SIZE")) is not None else None,
+    # Specify requantization block size for compressed tensor NVFP4 weights
```
jrplatin, line 295: Is this all weights? MoE weights? Linear weights?

lxhfirenking: This is linear weights (dense MLP + attention project) for now. w4a4_nvfp4 moe will be implemented in a separate PR, which will also contain this flag behavior

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a4_nvfp4.py
lxhfirenking: It comes down to w4a8 (where weights are nvfp4) being a non existent combination in upstream vllm, so we have to modify an existing `compressed_tesnors_*.py` somewhat to make it work without copy pasting too much code.

It does not make much sense for GPU stack to have this because there is native nvfp4 x nvfp4 support both in kernel and in hardware, there is no benefit in keeping activation in fp8 instead of nvfp4 when the weights are already nvfp4. For TPU it is a different storage because underlying matmul will happen in fp8.


The original upstream interface support a4 (nvfp4) and a16, so I dont think it is too out of place to add a8 here.

jrplatin: I see ok, that makes sense thanks! Can you make a quick note indicating how / why we are deviating from upstream vLLM? Thanks!

jrplatin (approved):

lk-chen (approved):

tpu_inference/layers/vllm/quantization/compressed_tensors/schemes/compressed_tensors_w4a4_nvfp4.py
lxhfirenking: done.

tests/layers/vllm/test_compressed_tensors_w4a4_nvfp4.py
```diff
@@ -214,9 +219,17 @@ def setup_environment():
 @pytest.mark.parametrize("enable_sp", [False, True])
 @pytest.mark.parametrize("enable_attn_dp", [False, True])
 @pytest.mark.parametrize("model", MODELS)
-@pytest.mark.parametrize("use_a16", [True])
+@pytest.mark.parametrize("activation_type", [W4A4ActivationType.BF16])
```
QiliangCui: The new FP8 activation path has no test coverage. `activation_type` is parametrized only over `[W4A4ActivationType.BF16]` here (and at the equivalent lines in the column/qkv/merged-column tests), so every test still exercises the pre-existing bf16 behavior — nothing constructs a scheme with `W4A4ActivationType.FP8`, and nothing reaches the new `get_scheme` branch in `compressed_tensors.py` that maps an fp8 dynamic `input_quant` onto it.

That branch is the substantive new behavior in this PR and it is untested in both directions:

1. **Scheme selection** — a cheap unit test over `get_scheme` asserting that nvfp4 weights + fp8 dynamic input yields `activation_type == FP8`, that nvfp4 input still yields `NVFP4`, that `input_quant is None` still yields `BF16`, and that an unsupported `input_quant` (e.g. int8, or fp8 with `dynamic=False`) raises. Note the fp8-static case currently falls through to the `raise ValueError` because the branch requires `input_quant.dynamic`, so pinning it prevents a later refactor from silently accepting it.
2. **Apply path** — at least one parametrization that actually runs a layer with `activation_type=FP8`, since that is the only configuration where `maybe_quantize_x=True` is passed down to `sharded_quantized_matmul`. Today `maybe_quantize_x` is exercised only at its default value, so a regression in the new plumbing would not be caught. Adding `W4A4ActivationType.FP8` to the existing `activation_type` parametrization is likely enough if the reference computation can tolerate it; the tolerance may need to be scoped per activation type.

Worth confirming as well that the intended production configuration is covered end to end: the description's repro sets both `REQUANTIZE_COMPRESSED_TENSOR_NVFP4_BLOCK_SIZE=256` and `ENABLE_QUANTIZED_MATMUL_KERNEL=1`, but the tests here parametrize the requantize block size at 32 with the kernel flag unset, which is the combination where gmm_v2 still dequantizes the weights before the matmul (block size below the MXU column size) — i.e. the tested configuration is not the one the flag is meant for.

QiliangCui (commented):
One test-coverage gap on the new FP8 activation path — details in the inline comment.

QiliangCui (approved):

tests/layers/vllm/test_compressed_tensors_w4a4_nvfp4.py
lxhfirenking: Done.

PR #3267, opened 2026-07-26
[DSv4] Adapt KV cache creation to vLLM packed byte-offset overlay groups

QiliangCui2023:
## Problem

The nightly job `tpu7x E2E MMLU for DeepSeek-V4-Flash torchax backend` crashes at server startup (red on `main` nightlies since build 23029, 2026-07-23; inherited by `releases/v0.26.0` nightly build 23216):

```
File "tpu_inference/runner/kv_cache_manager.py", line 920, in initialize_kv_cache
AssertionError:  kv_cache_tensor.shared_by: ['model.layers.2.attn.indexer.k_cache', 'model.layers.0.attn.swa_cache', 'model.layers.1.attn.swa_cache', 'model.layers.2.attn.indexer.compressor.state_cache', 'model.layers.3.attn.compressor.state_cache']
```

Root cause: the vLLM LKG bump from `05781e21dd` to `191146dba5` picked up vllm-project/vllm#48993 (`f3a920a076`), which rewrote `_get_kv_cache_config_packed`. DSv4 cache groups are now laid out densely in one shared block slab and each `KVCacheTensor.shared_by` lists the layers from *different* cache groups that start at the same byte offset — so mixed indexer/SWA/compressor-state groups are normal and expected. The TPU-side creation logic still assumed the old slot-based grouping (indexer caches only group with indexer caches) and asserted on it.

## Fix

JAX arrays cannot be byte views into one slab, so instead of consuming vLLM's byte offsets, `initialize_kv_cache` now rebuilds the DSv4 overlay plan from the specs (works with both old and new vLLM groupings):

- Every `MLAAttentionSpec` layer (CSA main latent `*.attn`, indexer `*.attn.indexer.k_cache`) gets its own array shaped by its own spec.
- Every compressor state cache maps to its compressed-KV layer's array (mirrors how vLLM wires `DeepseekCompressor.k_cache_prefix`; the compressor kernel writes f32 state and compressed KV through one buffer and asserts index equality at runtime). Byte fit is validated at init with a self-contained `[kv-cache]` error.
- Every `swa_cache` overlays a distinct CSA main-latent array within its cache group; overflow layers (DSv4-Flash: 43 SWA vs 21 CSA) get standalone arrays. Cross-group array sharing is safe because a block ID is owned by one cache group at a time.

## Tests

- New unit tests in `tests/runner/test_kv_cache_manager.py`: the exact #48993 packed layout (mixed shared_by, the crashing shape), the legacy slot-based layout (same resulting plan), and the tagged error path. All pass on a tpu7x-8 host; full `test_kv_cache_manager.py` suite 37/37.
- E2E validation: dev-pipeline run of the failing MMLU job on this branch (link to follow in comments).

Nightly (release branch): https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23216

### DIFF
--- a/tests/runner/test_kv_cache_manager.py
+++ b/tests/runner/test_kv_cache_manager.py
@@ -28,7 +28,9 @@ from vllm.v1.attention.backend import AttentionType
 from vllm.v1.kv_cache_interface import (FullAttentionSpec, KVCacheConfig,
                                         KVCacheGroupSpec, KVCacheTensor,
                                         MambaSpec, MLAAttentionSpec,
-                                        SlidingWindowSpec)
+                                        SlidingWindowMLASpec,
+                                        SlidingWindowSpec,
+                                        UniformTypeKVCacheSpecs)
 from vllm.v1.request import Request
 
 from tpu_inference import utils as common_utils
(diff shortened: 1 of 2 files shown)


QiliangCui2023:
Validated on the dev pipeline: https://buildkite.com/tpu-commons/tpu-inference-dev/builds/696 — `tpu7x unit tests kv_cache_manager` (37 tests) and the exact failing CI job `tpu7x E2E MMLU for DeepSeek-V4-Flash torchax backend` both **passed** on this branch (the MMLU job crashes at server startup on releases/v0.26.0 and main without this fix).

tests/runner/test_kv_cache_manager.py
```diff
@@ -28,7 +28,9 @@
 from vllm.v1.kv_cache_interface import (FullAttentionSpec, KVCacheConfig,
```
gxd3, line 28: qq: does the E2E test still fails even after this PR?: 
https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23216/canvas?sid=019f9fe7-98ae-448a-89e9-a0da3993f479&tab=output

gxd3: oh, saw it passed: https://buildkite.com/tpu-commons/tpu-inference-dev/builds/696/canvas?sid=019fa022-4c74-407d-b088-260d6b51571e&tab=output

QiliangCui2023:
@gxd3 correct — build 23216 predates this fix (it's the release nightly that surfaced the crash). With this branch, the exact same job passes: dev build 696 ran both the new unit tests and the full `tpu7x E2E MMLU for DeepSeek-V4-Flash torchax backend` job, both green.

gxd3 (approved):

QiliangCui2023:
Retargeted this PR to `main` — the bug exists identically on both branches (broken on every `main` nightly since build 23029 / Jul 23, including today's #23225 with the jax 0.11 bump; inherited by `releases/v0.26.0` nightly #23216). The branch is based on `7bad4320f` (an ancestor of `main`), so the diff is unchanged. After this merges to `main`, it can be cherry-picked to `releases/v0.26.0`.

PR #2991, opened 2026-06-24
Add correctness e2e for mistral small 4

karan:
# Description

Adds a basic correctness test for `mistralai/Mistral-Small-4-119B-2603` (which was supported as of #2611).

# Tests

```

    Processed prompts: 100%|██████████| 3/3 [00:16<00:00,  5.49s/it, est. speed input: 1.46 toks/s, output: 11.65 toks/s]

    --- Prompt: Hello, my name is ---
    Generated:  Brenda and I am a student at the University of North Carolina at Chapel Hill. I am currently taking a class called "The History of the English Language" and I am working on a project about the history of the word "okay." I was wondering if you could help me with some information about the etymology

    --- Prompt: The capital of France is ---
    Generated:  Paris. Paris is the largest city in France and serves as its political, economic, and cultural center. It is known for its iconic landmarks such as the Eiffel Tower, Louvre Museum, and Notre-Dame Cathedral. Paris is also a major global city and a popular tourist destination.

    Rate this question:
    - 2

    --- Prompt: Write a short poem about artificial intelligence on TPUs. ---
    Generated:  (Tensor Processing Units)

    **Artificial minds in silicon bright,**
    **TPUs hum with logic’s light.**
    **Neural nets in parallel dance,**
    **Learning fast, they swiftly prance.**

    **Data flows like rivers wide,**
    **Through the circuits, they abide.**
    **Patterns deep in circuits spun

    PASSED

    ================== 1 passed, 19 warnings in 312.58s (0:05:12) ==================
```

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3268, opened 2026-07-26
[CI] Fix Mistral-Small-4 record step failing on tpu6e skip

QiliangCui2023:
## Problem

Every nightly, the job `tpu6e Record correctness result for mistralai/Mistral-Small-4-119B-2603/text-only` hard-fails (`soft_fail` is not set on record steps) even though the upstream `tpu6e Correctness` step passes. Red on `main` nightlies (e.g. build 23204) and inherited by the `releases/v0.26.0` nightly build 23216.

Root cause (introduced with the model config in #2991): on tpu6e the Correctness step self-skips and writes the outcome to buildkite meta-data as `"skipped (only for tpu7x)"`. `record_step_result.sh` matches outcomes exactly — `passed` / `skipped` / `unverified` / `not enough HBM` — so this string falls through to `Failing` and `exit 1`.

## Fix

Write the exact `skipped` token the recorder expects (renders as "Untested" in the support matrix).

Log evidence from build 23216:
```
--- Checking tpu6e_mistralai_Mistral-Small-4-119B-2603_text-only_Correctness Outcome (Hardware: tpu6e)
Step tpu6e_..._Correctness outcome: skipped (only for tpu7x)
Error: The command exited with status 1
```

Nightly (release branch): https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23216

### DIFF
--- a/.buildkite/models/mistralai_Mistral-Small-4-119B-2603_text-only.yml
+++ b/.buildkite/models/mistralai_Mistral-Small-4-119B-2603_text-only.yml
@@ -26,7 +26,7 @@ steps:
     commands:
       - |
         if [ "${TPU_VERSION:-tpu6e}" != "tpu7x" ]; then
-          buildkite-agent meta-data set "${TPU_VERSION:-tpu6e}_mistralai_Mistral-Small-4-119B-2603_text-only_Correctness" "skipped (only for tpu7x)"
+          buildkite-agent meta-data set "${TPU_VERSION:-tpu6e}_mistralai_Mistral-Small-4-119B-2603_text-only_Correctness" "skipped"
         else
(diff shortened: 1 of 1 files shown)


QiliangCui2023:
Retargeted to `main` — the bug exists identically on both branches: `main` still has the `"skipped (only for tpu7x)"` meta-data string, and today's main nightly #23225 fails `tpu6e Record correctness result for mistralai/Mistral-Small-4-119B-2603` the same way. The branch is based on `7bad4320f` (ancestor of main), so the diff is unchanged. After merge to main, this can be cherry-picked to `releases/v0.26.0`.

CienetStingLin (approved):

PR #3074, opened 2026-07-04
Optimize AOT precompilation in tpu-inference

igorts-git:
# Description

This PR optimizes Ahead-Of-Time (AOT) precompilation and dummy weight loading in the TPU inference runner to reduce HBM consumption and eliminate redundant compilation overhead.

### Key Changes

**Memory-efficient AOT Precompilation**: Replaces concrete dummy array allocations (`_create_dummy_tensor`) with lightweight `jax.ShapeDtypeStruct` during JIT function lowering (`fn.lower`). This avoids allocating High Bandwidth Memory (HBM) on the TPU for dummy inputs during compilation. As a result, the `MAX_PRECOMPILE_PROMPT_TOKENS` limit (previously capped at 1024) is removed, allowing precompilation of larger prompt token shapes without HBM exhaustion.

### Implementation Details
- Added `compile_only` to `_run_compilation` and `_precompile` methods in `CompilationManager` to support lowering and compiling JIT functions without adding them to warmup execution passes.
- Migrated placeholder tensor creations in `compilation_manager.py` to `jax.ShapeDtypeStruct`.

FIXES: b/528378191

# Tests

Manual invocations of MaxText `train_rl.py`.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3274, opened 2026-07-27
[AOT Optimization] Multi-threaded JAX Lowering & Helper Precompilation Pruning (b/528378191)

igorts-git:
# Description

This PR improves Ahead-Of-Time (AOT) precompilation time in `tpu-inference` engine initialization (b/528378191), reducing startup latency of Llama under MaxText/Tunix RL from ~15 minutes down to < 45 seconds.

FIXES: b/528378191

### Key Changes

1. **Parallel Host CPU JAX Lowering**: Refactored `_run_compilation` in `CompilationManager` to execute `fn.lower()` in parallel inside `ThreadPoolExecutor(max_workers=8)` worker threads with elevated stack size (`64MB`).
2. **Off-Diagonal Precompilation Pruning**: Restricted non-speculative helper precompilation loops (`_precompile_substitute_placeholder_token`) to diagonal shape pairs (`next_tokens_size == num_tokens`), eliminating ~300 redundant binary compilation targets.
3. **Macro Compilation Flush Barriers**: Consolidated 11 fine-grained micro-flushes into 3 macro flush barriers (Backbones -> Async Manipulators -> Auxiliary Kernels).
4. **Prompt Logprobs Precompilation Guard**: Restricted auxiliary prompt logprobs precompilation to sequence lengths <= 1024 tokens (`MAX_PRECOMPILE_PROMPT_TOKENS`), eliminating host CPU JAX tracing overhead for long prompt lengths. In PR #3074 we removed this guard, but it still very beneficial for start up time, so I am re-adding it.
6. **AWQ Quantization Import Fallback**: Added backward and forward compatibility import fallback (`auto_awq` vs `awq`) for AWQ quantization across vLLM release reorganizations. This is unrelated to AOT optimization, just a fix to a failure that I observed while testing.

# Tests
Manual invocations of MaxText/Tunix RL and standalone tpu-inference.
CI tests.

# Checklist

### DIFF
--- a/tpu_inference/runner/compilation_manager.py
+++ b/tpu_inference/runner/compilation_manager.py
@@ -161,26 +161,22 @@ class CompilationManager:
                 reason)
             return
 
-        try:
-            with jax.set_mesh(self.runner.mesh):
-                lowered = fn.lower(*args, **call_kwargs)
-        except Exception as e:
-            if compile_only:
-                logger.error(
-                    f"Failed to lower {name} with compile_only=True: {e}")
-                raise
-            else:
-                # AOT lower not supported here (e.g. a jit whose body contains a
-                # nested jit with compiler_options). Fall back to warmup-only — the
-                # warmup pass will trigger inline compile.
-                logger.info(
-                    "AOT lower skipped for %s (%r); will compile in warmup.",
-                    name, e)
-                return
-
-        # Compilation is thread-safe
-        def _compile(lowered, name, mesh):
+        def _lower_and_compile(fn, args, call_kwargs, name, mesh,
+                               compile_only):
             with jax.set_mesh(mesh):
+                try:
+                    lowered = fn.lower(*args, **call_kwargs)
+                except Exception as e:
+                    if compile_only:
+                        logger.error(
+                            f"Failed to lower {name} with compile_only=True: {e}"
+                        )
+                        raise
+                    else:
+                        logger.info(
+                            "AOT lower skipped for %s (%r); will compile in warmup.",
+                            name, e)
+                        return None
                 start = time.perf_counter()
                 compiled = lowered.compile()
                 elapsed = time.perf_counter() - start
@@ -189,10 +185,13 @@ class CompilationManager:
                 return compiled
 
         if self._compile_executor is None:
-            _compile(lowered, log_name, self.runner.mesh)
+            _lower_and_compile(fn, args, call_kwargs, log_name,
+                               self.runner.mesh, compile_only)
         else:
-            future = self._compile_executor.submit(_compile, lowered, log_name,
-                                                   self.runner.mesh)
+            future = self._compile_executor.submit(_lower_and_compile, fn,
+                                                   args, call_kwargs, log_name,
+                                                   self.runner.mesh,
+                                                   compile_only)
             self._compile_futures.append(future)
 
     def _flush_compilations(self) -> None:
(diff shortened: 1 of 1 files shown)


tpu_inference/layers/vllm/quantization/awq.py
```diff
@@ -29,10 +29,21 @@
                                                set_weight_attrs)
 from vllm.model_executor.layers.quantization import \
     register_quantization_config
-from vllm.model_executor.layers.quantization.auto_awq import \
-    AutoAWQConfig as AWQConfig
-from vllm.model_executor.layers.quantization.auto_awq import \
-    AutoAWQLinearMethod as AWQLinearMethod
+
```
wenxindongwork: could we remove this change from this PR? let us ask the oncall to handle this change in another PR.

tpu_inference/runner/compilation_manager.py
```diff
@@ -238,53 +237,49 @@ def capture_model(self) -> None:
...
                 self._precompile_structured_decoding()
-                self._flush_compilations()
                 if self.runner.speculative_config:
                     self._precompile_speculative_decoding()
-                    self._flush_compilations()
                 if self.runner.enable_continue_decode:
                     self._precompile_continue_decode()
-                    self._flush_compilations()
+                # Barrier 3: Flush Auxiliary Kernels
+                self._flush_compilations()
```
wenxindongwork, line 282: curious if you have tested with large `NUM_PRECOMPILE_WORKERS`? I'm concerned that flushing compilation less frequently would introduce OOM

tpu_inference/runner/compilation_manager.py
```diff
@@ -577,6 +572,14 @@ def _compile_one(input_padding: int, input_sharding: NamedSharding,
         else:
             for num_tokens in all_token_sizes:
                 for next_tokens_size in all_token_sizes:
+                    # Off-diagonal pruning: Skip non-matching (next_tokens_size != num_tokens) compilation pairs.
+                    # Rationale: Off-diagonal shapes (where next_tokens_size differs from prompt num_tokens) are
+                    # only required when speculative draft token counts vary off-diagonally. For non-speculative
+                    # precompilation and standard RL rollouts, next_tokens_size matches num_tokens.
+                    # Eliminating these ~300 redundant binary compilation pairs saves significant engine startup time
+                    # and host memory while remaining 100% safe for standard RL rollouts.
+                    if next_tokens_size != num_tokens:
```
wenxindongwork: would it be possible to remove the nested for loop?

tpu_inference/runner/compilation_manager.py
```diff
@@ -1081,8 +1084,17 @@ def _precompile_gather_logprobs(self) -> None:
         logger.info(
             "Compiling compute_and_gather_prompt_logprobs with different input shapes."
         )
-        # Bypassed MAX_PRECOMPILE_PROMPT_TOKENS limit as ShapeDtypeStruct compilation allocates no HBM
+        # Restricting precompilation of auxiliary prompt logprobs to prompt lengths num_tokens <= 1024
+        # speeds up engine startup time by avoiding redundant host CPU JAX tracing and XLA lowering overhead
+        # for long prompt sequence lengths (> 1024 tokens).
+        MAX_PRECOMPILE_PROMPT_TOKENS = 1024
```
wenxindongwork, line 1083: where is the MAX_PRECOMPILE_PROMPT_TOKENS number obtained from?

tpu_inference/layers/vllm/quantization/awq.py
igorts-git: I removed this change from the PR.

tpu_inference/runner/compilation_manager.py
igorts-git: It should not cause problems anymore. I just ran it with NUM_PRECOMPILE_WORKERS=32 and saw about 80GB of RAM used. The main change in RAM usage happened in https://github.com/vllm-project/tpu-inference/pull/3074 where we replaced `_create_dummy_tensor` with `jax.ShapeDtypeStruct`. So, now instead of creating these huge tensors in RAM and placing them in devices, we never allocate them in the first place.

tpu_inference/runner/compilation_manager.py
igorts-git: Good point. I just removed the loop. Silly me.

tpu_inference/runner/compilation_manager.py
igorts-git: This is exactly how the code looked like before PR #3074. The old code had this constant, I removed it, but now I see that it is quite beneficial, so I am reverting that limit. Here is the state of the file before that PR: https://github.com/vllm-project/tpu-inference/blob/8905d1c35a539588516205f925b2dff211b9ac59/tpu_inference/runner/compilation_manager.py#L1049-L1056

wenxindongwork (approved):

PR #2878, opened 2026-06-12
[Qwen3.5][torchax][Multimoda] Remove vision tower patch

lk-chen:
# Description

Encourage use `--compilation-config="{\"cudagraph_mm_encoder\": true}"`

#2804 

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3283, opened 2026-07-28
Update vLLM model wrapper, Qwen3 VL patcher, and vision tower JIT for…

amitkumar307d:
# Description

## 📊 Technical Analysis 

This commit addresses several critical memory inefficiencies, OOMs, and runtime stalls when performing large-scale **video inference** with Qwen architectures on TPUs.

---

### 1. 🧮 Memory Inefficiency & OOMs in Multimodal Forward Pass

*   **❌ The Issue**: For large inputs like video, the `pixel_values` tensor was being **fully replicated** across all TPU devices. This led to severe **HBM/Host Out-of-Memory (OOM)** errors. Furthermore, input sequence lengths were often not divisible by the Tensor Parallelism (`tp_size`) factor, making efficient sharding impossible without crashing downstream spatial downsampling operations.
*   **🧠 Approach**: Instead of replicating huge video tensors, shard them across the Model (TP) axis. To do this safely, dynamically pad the sequence length to a multiple of `tp_size` and valid factoring dimensions.
*   **✅ The Fix**: 
    *   Padded `pixel_values` and `grid_thw` to be multiples of `tp_size`.
    *   Ensured padding factors into valid `(t, h, w)` dimensions (where $h \ge 2, w \ge 2$) to satisfy downstream kernel constraints.
    *   Applied explicit sharding using `PartitionSpec("model", None)`.
    *   Stripped padding post-forward pass using original lengths to maintain exact correctness.

<details>
<summary>🔍 Code (<code>tpu_inference/models/vllm/vllm_model_wrapper.py</code>)</summary>

```python
# Pad pixel_values sequence length to multiple of tp_size to avoid full replication OOM
seq_len = pixels.shape[0]
if seq_len % tp_size != 0:
    pad_seq = tp_size - (seq_len % tp_size)
    # ... logic to find valid dummy_grid_thw [t, h, w] ...
    kwargs[pixel_key] = torch.cat([pixels, pad_pixels_seq], dim=0)

# Shard pixel values over model (TP) axis if divisible
if param_name in ("pixel_values", "pixel_values_videos"):
    spec = PartitionSpec("model", None) if arr.shape[0] % tp_size == 0 else PartitionSpec()
```
</details>

---

### 2. 💥 VMEM OOM in Vision Tower Attention

*   **❌ The Issue**: Native PyTorch `scaled_dot_product_attention` (SDPA) consumed excessive **VMEM** ($>40\text{MB}$) for long video sequences in the Vision Tower, triggering hard TPU VMEM crashes.
*   **🧠 Approach**: Intercept the standard SDPA call and transparently redirect execution to a memory-efficient Flash Attention kernel (`encoder_only_flash_attention`) based on dynamic memory heuristics.
*   **✅ The Fix**: 
    *   Globally patched `F.scaled_dot_product_attention` for Qwen3 VL.
    *   Implemented `_chunked_sdpa` which estimates native VMEM usage ($B \times H \times S \times S \times 2\text{ bytes}$).
    *   If estimated usage exceeds **40MB**, it redirects to the highly optimized `encoder_only_flash_attention` kernel from `tpu-inference`.

<details>
<summary>🔍 Code (<code>tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py</code>)</summary>

```python
# Global Patch
F.scaled_dot_product_attention = _chunked_sdpa

def _chunked_sdpa(...):
    # ... heuristics ...
    if attn_mask is None and dropout_p == 0.0 and estimated_vmem_native > 40 * 1024 * 1024:
        # Fallback to custom memory-efficient kernel
        out_jax = encoder_only_flash_attention(q_jax, k_jax, v_jax, seq_lens, ...)
        return torch_view(out_jax)
```
</details>

---

### 3. ⏳ Stalls Due to Missing JIT Support & Precompilation

*   **❌ The Issue**: Newer Qwen 3.5 architectures were not registered as "Jittable", missing out on JAX optimizations. Additionally, the vision precompiler only generated graphs for static images ($t=1$), causing major runtime compilation stalls ("JIT lag") during the first video request.
*   **🧠 Approach**: Whitelist newer architectures and expand the precompiler heat-up routine to simulate video frame batches.
*   **✅ The Fix**: 
    *   Added `Qwen3_5MoeForConditionalGeneration` and `Qwen3_5ForConditionalGeneration` to `JITTABLE_ARCHS`.
    *   Expanded `maybe_precompile_vision_encoder_fn` to iterate over common video frame counts `[1, 2, 4, 8, 16]` to fully warm up the JIT cache before requests arrive.

<details>
<summary>🔍 Code (<code>tpu_inference/models/vllm/experimental/vision_tower_jit.py</code>)</summary>

```python
JITTABLE_ARCHS = {
    Qwen3OmniMoeThinkerForConditionalGeneration,
    Qwen3_5MoeForConditionalGeneration,
    Qwen3_5ForConditionalGeneration,
}

# Video Precompilation Warming Up
frame_counts = [1, 2, 4, 8, 16]
for t_val in frame_counts:
    # ... warmup execution ...
```
</details>

---

### 4. 🗄️ Missing Multimodal Cache Integration

*   **❌ The Issue**: Redundant computation and data transfers occurred for multimodal features because the TPU worker and engine cores were not actively retrieving features from the `mm_receiver_cache`.
*   **🧠 Approach**: Activate and plumb the receiver cache directly into the core execution paths.
*   **✅ The Fix**: Initialized and invoked `mm_receiver_cache` in `TPUWorker` and `DisaggEngineCoreProc` before running model execution to reuse cached features.

<details>
<summary>🔍 Code (<code>tpu_inference/worker/tpu_worker.py</code>)</summary>

```python
if getattr(self, "_apply_mm_cache", None) is not None:
    self._apply_mm_cache(scheduler_output)
elif self.mm_receiver_cache is not None:
    for req in scheduler_output.scheduled_new_reqs:
        if getattr(req, "mm_inputs", None):
            req.mm_inputs = self.mm_receiver_cache.get_and_update_features(req.mm_inputs)
```
</details>

---

# Tests

### Model server command

```bash
python3 -m vllm.entrypoints.openai.api_server \
  --host=0.0.0.0 \
  --port=8000 \
  --download-dir=/data \
  --kv-cache-dtype=fp8 \
  --max-model-len=4096 \
  --gpu-memory-utilization=0.9 \
  --tensor-parallel-size=8 \
  --max-num-seqs=128 \
  --max-num-batched-tokens=1024 \
  --no-enable-prefix-caching \
  --enable-expert-parallel \
  --async-scheduling \
  --model=gs://amitmkumar-bucket/gcloud/Qwen3.5-397B-A17B-FP8 \
  --served-model-name=Qwen/Qwen3.5-397B-A17B-FP8 \
  --load-format=auto \
  --enable-auto-tool-choice \
  --reasoning-parser=qwen3 \
  --limit-mm-per-prompt='{"image": 0, "video": 1}' \
  --tool-call-parser=qwen3_coder \
  --block-size=256 \
  --additional-config='{"compilation_sizes": [512, 1024, 2048, 4096],"sharding": {"sharding_strategy": {"enable_dp_attention": true}}}' || true

```

### Benchmark command

```bash
vllm bench serve \
  --backend openai-chat \
  --dataset-name random-mm \
  --base-url http://qwen-3-5-397b-tpu-9k-video-svc:8000 \
  --num-prompts 640 \
  --num-warmups 640
  --max-concurrency 128 \
  --random-input-len 500 \
  --random-output-len 1024 \
  --random-mm-base-items-per-request 1 \
  --random-mm-limit-mm-per-prompt '{"image": 0, "video": 1}' \
  --random-mm-bucket-config '{"(256, 256, 16)": 1.0}' \
  --request-rate inf \
  --percentile-metrics ttft,tpot,itl,e2el \
  --ignore-eos 
```

### Performance metrics

```bash
Successful requests:                     640       
Maximum request concurrency:             128       
Benchmark duration (s):                  202.84    
Total input tokens:                      420098    
Total generated tokens:                  655360    
Request throughput (req/s):              3.16      
Output token throughput (tok/s):         3230.92   
Peak output token throughput (tok/s):    4480.00   
Peak concurrent requests:                192.00    
Total token throughput (tok/s):          5302.00   
Mean TTFT (ms):                          2609.97   
Median TTFT (ms):                        1219.77   
P99 TTFT (ms):                           11138.57  
Mean TPOT (ms):                          37.05     
Median TPOT (ms):                        37.37     
P99 TPOT (ms):                           41.15     
Mean ITL (ms):                           38.05     
Median ITL (ms):                         32.10     
P99 ITL (ms):                            96.49     
Mean E2EL (ms):                          40507.93  
Median E2EL (ms):                        39579.70  
P99 E2EL (ms):                           45238.58  
```

# Checklist

-[X] I have performed a self-review of my code.
-[X] I have necessary comments in my code, particularly in hard-to-understand areas.
-[X] I have made or will make corresponding changes to any relevant documentation.

### DIFF
--- a/tpu_inference/core/core_tpu.py
+++ b/tpu_inference/core/core_tpu.py
@@ -16,6 +16,7 @@ import jax
 # ======================================================================================
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
+from vllm.multimodal import MULTIMODAL_REGISTRY
 from vllm.tasks import POOLING_TASKS, SupportedTask
 from vllm.v1.core.kv_cache_utils import (get_request_block_hasher,
                                          init_none_hash)
@@ -513,7 +514,8 @@ class DisaggEngineCore(vLLMEngineCore):
         self.step_fn = (self.step if self.batch_queue is None else
                         self.step_with_batch_queue)
 
-        self.mm_receiver_cache = None
+        self.mm_receiver_cache = MULTIMODAL_REGISTRY.engine_receiver_cache_from_config(
+            vllm_config.model_config)
         self._orchestrator = _DisaggOrchestrator(
             config=vllm_config,
             output_queue=self.output_queue,
@@ -739,7 +741,8 @@ class DisaggEngineCoreProc(vLLMEngineCoreProc):
             self.request_block_hasher = get_request_block_hasher(
                 block_size, caching_hash_fn)
 
-        self.mm_receiver_cache = None
+        self.mm_receiver_cache = MULTIMODAL_REGISTRY.engine_receiver_cache_from_config(
+            vllm_config.model_config)
         self._orchestrator = _DisaggOrchestrator(
             config=vllm_config,
             output_queue=self.output_queue,
@@ -753,6 +756,18 @@ class DisaggEngineCoreProc(vLLMEngineCoreProc):
         if not isinstance(request.request_id, str):
             raise TypeError(
                 f"request_id must be a string, got {type(request.request_id)}")
+
+        if self.mm_receiver_cache is not None and getattr(
+                request, "mm_inputs", None):
+            request.mm_inputs = self.mm_receiver_cache.get_and_update_features(
+                request.mm_inputs)
+        elif self.mm_receiver_cache is not None and getattr(
+                request, "mm_positions", None) and hasattr(
+                    request, "mm_features"):
+            # Older vllm might use mm_features, checking both to be safe
+            if request.mm_features:
+                request.mm_features = self.mm_receiver_cache.get_and_update_features(
+                    request.mm_features)
 
         if pooling_params := request.pooling_params:
             supported_pooling_tasks = [
--- a/tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
+++ b/tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
@@ -45,8 +45,10 @@ import jax
 import jax.numpy as jnp
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
 import vllm.model_executor.models.qwen3_vl as qwen3_vl_mod
 import vllm.model_executor.models.utils as vllm_utils
+from jax.experimental.pallas import tpu as pltpu
 from torchax.interop import jax_view, torch_view
 from vllm.model_executor.models.qwen3_vl import Qwen3VLForConditionalGeneration
 from vllm.multimodal import NestedTensors
@@ -54,9 +56,66 @@ from vllm.sequence import IntermediateTensors
 
 from tpu_inference.distributed.jax_parallel_state import \
     get_pp_group as jax_get_pp_group
+from tpu_inference.kernels.flash_attention.kernel import \
+    encoder_only_flash_attention
 from tpu_inference.logger import init_logger
+from tpu_inference.utils import align_to
 
 logger = init_logger(__name__)
+
+_orig_sdpa = F.scaled_dot_product_attention
+
+
+def _chunked_sdpa(query,
+                  key,
+                  value,
+                  attn_mask=None,
+                  dropout_p=0.0,
+                  is_causal=False,
+                  scale=None):
+    if query.ndim == 4:
+        B, H, S, D = query.shape
+
+        # 1. Fallback for GQA (our custom kernel requires num_heads == num_kv_heads)
+        if query.shape != key.shape:
+            return _orig_sdpa(query, key, value, attn_mask, dropout_p,
+                              is_causal, scale)
+
+        # 2. Heuristic for native SDPA memory usage: B * H * S * S * element_size bytes
+        # We account for hardware alignment (e.g. 128 lanes) to prevent VMEM OOM.
+        lane_count = pltpu.get_tpu_info().num_lanes
+        S_padded = align_to(S, lane_count)
+        estimated_vmem_native = B * H * S_padded * S_padded * query.element_size(
+        )
+
+        # 3. Only use chunked flash attention if we will OOM native SDPA (>40MB)
+        if attn_mask is None and dropout_p == 0.0 and estimated_vmem_native > 40 * 1024 * 1024:
+            q_jax = jax_view(query)
+            k_jax = jax_view(key)
+            v_jax = jax_view(value)
+
+            # [B, H, S, D] -> [B, S, H, D] -> [B*S, H, D]
+            q_jax = jnp.reshape(jnp.swapaxes(q_jax, 1, 2), (B * S, H, D))
+            k_jax = jnp.reshape(jnp.swapaxes(k_jax, 1, 2), (B * S, H, D))
+            v_jax = jnp.reshape(jnp.swapaxes(v_jax, 1, 2), (B * S, H, D))
+
+            seq_lens = jnp.full((B, ), S, dtype=jnp.int32)
+            if scale is None:
+                scale = 1.0 / (D**0.5)
+
+            out_jax = encoder_only_flash_attention(q_jax,
+                                                   k_jax,
+                                                   v_jax,
+                                                   seq_lens,
+                                                   causal=is_causal,
+                                                   sm_scale=scale)
+
+            # [B*S, H, D] -> [B, S, H, D] -> [B, H, S, D]
+            out_jax = jnp.swapaxes(jnp.reshape(out_jax, (B, S, H, D)), 1, 2)
+            return torch_view(out_jax)
+
+    return _orig_sdpa(query, key, value, attn_mask, dropout_p, is_causal,
+                      scale)
 
 
 def _patched_set_deepstack(vllm_model, deepstack_input_embeds):
@@ -222,6 +281,12 @@ def _patched_embed_input_ids(vllm_model, orig_embed_input_ids, *args,
     """
     # 1. Get the base text embeddings from the native model.
     inputs_embeds = orig_embed_input_ids(*args, **kwargs)
+
+    # 1.5. Bypass deepstack packing during autoregressive decode or text-only prefill.
+    # In vllm_model_wrapper, mm_embeds is passed as args[1] during prefill. If len(args) == 1,
+    # it means mm_embeds is None (decode phase or text-only). We also check size(0) == 1 just in case.
+    if len(args) == 1 or inputs_embeds.size(0) == 1:
+        return inputs_embeds
 
     # 2. Check if there are any deepstack features to pack.
     # Read from _deepstack_tensors (our JAX-compatible cache) rather than
@@ -427,6 +492,9 @@ def apply_qwen3_vl_patches(vllm_model):
         logger.info(
             "Disabled dynamo for Qwen3LLMModel; JAX JIT handles outer compilation"
         )
+
+    # 8. Patch PyTorch SDPA globally to catch Qwen Vision Tower calls and prevent VMEM OOM
+    F.scaled_dot_product_attention = _chunked_sdpa
 
 
 def is_qwen3_vl(vllm_model) -> bool:
--- a/tpu_inference/models/vllm/experimental/vision_tower_jit.py
+++ b/tpu_inference/models/vllm/experimental/vision_tower_jit.py
@@ -24,9 +24,12 @@ import torch
 from transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe import \
     Qwen3OmniMoeConfig
 from vllm.config import VllmConfig
+from vllm.model_executor.models.qwen3_5 import (
+    Qwen3_5ForConditionalGeneration, Qwen3_5MoeForConditionalGeneration)
 from vllm.model_executor.models.qwen3_omni_moe_thinker import \
     Qwen3OmniMoeThinkerForConditionalGeneration
 
+from tpu_inference import envs
 from tpu_inference.logger import init_logger
 from tpu_inference.utils import to_jax_dtype
 
@@ -35,6 +38,8 @@ logger = init_logger(__name__)
 # Architectures whose embed_multimodal function is safe to wrap with jax.jit.
 JITTABLE_ARCHS = {
     Qwen3OmniMoeThinkerForConditionalGeneration,
+    Qwen3_5MoeForConditionalGeneration,
+    Qwen3_5ForConditionalGeneration,
 }
 
 
@@ -82,6 +87,7 @@ def maybe_jit_embed_multimodal_func(embed_multimodal_func_jax: Callable,
         return embed_multimodal_func_jax
 
 
+@jax.tree_util.register_pytree_node_class
 class GridTHW(tuple):
     """Tensor-like wrapper for image/video grid_thw arguments.
 
@@ -103,6 +109,12 @@ class GridTHW(tuple):
         flat: tuple = _nested_to_tuple(values)
         return super().__new__(cls, flat)
 
+    def __getitem__(self, key):
+        val = super().__getitem__(key)
+        if isinstance(key, slice):
+            return type(self)(val)
+        return val
+
     # ---- tensor-like API expected by _process_image_input ----
 
     @property
@@ -123,6 +135,13 @@ class GridTHW(tuple):
 
     def __repr__(self):
         return f"GridTHW({tuple(self)})"
+
+    def tree_flatten(self):
+        return (), tuple(self)
+
+    @classmethod
+    def tree_unflatten(cls, aux_data, children):
+        return cls(aux_data)
 
 
 def maybe_precompile_vision_encoder_fn(
@@ -151,7 +170,12 @@ def maybe_precompile_vision_encoder_fn(
     spatial_merge_unit = vc.spatial_merge_size**2
     max_patches = (vllm_config.scheduler_config.max_num_batched_tokens //
                    spatial_merge_unit)
-    min_shift = 4  # 1 << 4 = 16 patches minimum
+    min_shift = envs.VISION_MIN_SHIFT
+    if min_shift < 6:
+        logger.warning(
+            f"VISION_MIN_SHIFT is set to {min_shift} (< 6). "
+            "This may cause JAX divisibility errors on TPUs with 8+ devices "
+            "if spatial merging is active.")
     max_shift = max(min_shift, (max(max_patches, 1) - 1).bit_length())
     num_patches_paddings = [1 << i for i in range(min_shift, max_shift + 1)]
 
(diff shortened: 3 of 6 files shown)


tpu_inference/core/core_tpu.py
```diff
@@ -513,7 +514,8 @@ def __init__(
         self.step_fn = (self.step if self.batch_queue is None else
                         self.step_with_batch_queue)
 
-        self.mm_receiver_cache = None
+        self.mm_receiver_cache = MULTIMODAL_REGISTRY.engine_receiver_cache_from_config(
```
QiliangCui, line 517: **[blocker]** `engine_receiver_cache_from_config` takes the full `VllmConfig`, not a `ModelConfig`: at the pinned vLLM, `MultiModalRegistry._get_cache_type` immediately does `model_config = vllm_config.model_config` (vllm `multimodal/registry.py`), so passing `vllm_config.model_config` here raises `AttributeError` as soon as this class is constructed — for any model, multimodal or not. Same issue at line 744. PR CI doesn't catch it because these classes are only instantiated under `PREFILL_SLICES` disagg mode.

vLLM's own `EngineCore.__init__` calls it with the full config: `engine_receiver_cache_from_config(vllm_config)`.

tpu_inference/core/core_tpu.py
```diff
@@ -754,6 +757,18 @@ def add_request(self, request: EngineCoreRequest, request_wave: int = 0):
...
                 f"request_id must be a string, got {type(request.request_id)}")
 
+        if self.mm_receiver_cache is not None and getattr(
+                request, "mm_inputs", None):
+            request.mm_inputs = self.mm_receiver_cache.get_and_update_features(
+                request.mm_inputs)
+        elif self.mm_receiver_cache is not None and getattr(
+                request, "mm_positions", None) and hasattr(
+                    request, "mm_features"):
+            # Older vllm might use mm_features, checking both to be safe
```
QiliangCui, line 767: **[blocker]** This block can never execute at the vLLM commit this repo pins: `EngineCoreRequest` has neither `mm_inputs` nor `mm_positions` — multimodal data lives in `mm_features` (`vllm/v1/engine/__init__.py`). So the first branch is always falsy and `getattr(request, "mm_positions", None)` is always `None`. (The comment is also inverted — `mm_features` is the *newer* schema, `mm_inputs` the older one.)

If engine-side feature caching is the goal, mirror vLLM's `EngineCore.preprocess_add_request`:

```python
if self.mm_receiver_cache is not None and request.mm_features:
    request.mm_features = self.mm_receiver_cache.get_and_update_features(request.mm_features)
```

Also note the sibling `DisaggEngineCore` now creates the cache in `__init__` but its `add_request` never applies it.

tpu_inference/worker/tpu_worker.py
```diff
@@ -567,6 +573,18 @@ def execute_model(
         self,
         scheduler_output: SchedulerOutput,
     ) -> Optional[ModelRunnerOutput]:
+        if getattr(self, "_apply_mm_cache", None) is not None:
```
QiliangCui, line 576: **[blocker]** This whole block is dead code at the pinned vLLM:

- `_apply_mm_cache` is defined on vLLM's `WorkerWrapperBase`; `TPUWorker` subclasses the abstract `WorkerBase`, so `getattr(self, "_apply_mm_cache", None)` is always `None` here. When the executor path goes through `WorkerWrapperBase`, its `execute_model` already applies the cache before delegating (`vllm/v1/worker/worker_base.py`).
- In the `elif`, `NewRequestData` has no `mm_inputs` and no `mm_positions` — only `mm_features` (`vllm/v1/core/sched/output.py`) — so neither branch can fire.

Net effect: nothing runs. Either rewrite against `req.mm_features` (as `WorkerWrapperBase._apply_mm_cache` does) or drop this.

tpu_inference/worker/tpu_worker.py
```diff
@@ -262,6 +262,12 @@ def __init__(
         # step_counter is used to calculate uuid to transfer intermediate tensors.
         self.step_counter = 0
 
+        import multiprocessing
+
+        from vllm.multimodal import MULTIMODAL_REGISTRY
+        self.mm_receiver_cache = MULTIMODAL_REGISTRY.worker_receiver_cache_from_config(
```
QiliangCui, line 268: **[minor]** Two issues: (1) `worker_receiver_cache_from_config` only returns a cache for `mm_processor_cache_type == "shm"`, and vLLM deliberately refuses to build it without the executor-provided `shared_worker_lock` (it raises in `WorkerWrapperBase.init_worker`) — a fresh private `multiprocessing.Lock()` can't provide cross-process mutual exclusion for the shm object store. (2) This is now the third place a worker receiver cache is created (vLLM's `WorkerWrapperBase`, `disagg_executor.py:82`). Plumb the shared lock through the executor or drop this. Also move the function-level `multiprocessing` / `MULTIMODAL_REGISTRY` imports to module top.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -541,23 +549,223 @@ def embed_multimodal_func_torch(params_and_buffers: Any,
...
+                    pixels = kwargs[pixel_key]
+                    seq_len = pixels.shape[0]
+                    if seq_len % tp_size != 0:
+                        padded_anything = True
+                        pad_seq = tp_size - (seq_len % tp_size)
+
+                        # Find a pad_seq that can be factored into t, h, w where h>=2 and w>=2
+                        # to prevent spatial_merge_size downsampling from crashing on 0.
+                        dummy_grid_thw = None
+                        while dummy_grid_thw is None:
```
QiliangCui, line 644: **[major]** This search only enforces `h >= 2, w >= 2`, but the vision tower also needs `h` and `w` divisible by `spatial_merge_size`: e.g. `pad_seq=6` yields `(1, 2, 3)`, and Qwen3-VL's `rot_pos_ids` does `reshape(h//merge, merge, w//merge, merge)` → ValueError (6 elements into (1,2,1,2)).

Today this is unreachable only because the qwen processors emit `h`/`w` as multiples of `merge_size`, so patch counts are ≡ 0 mod 4 and `pad_seq` always lands on a multiple of 4 (the h=2, w=2 case) — but nothing asserts that invariant, and anything that violates it crashes deep in the vision tower.

Simpler and safe: pad to a multiple of `lcm(tp_size, merge_size**2)` and use `dummy_grid_thw = [pad_seq // merge_size**2, merge_size, merge_size]`; the nested search loop then goes away entirely.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -541,23 +549,223 @@ def embed_multimodal_func_torch(params_and_buffers: Any,
...
+                                pad_seq += tp_size
+
+                        pad_pixels_seq = torch.zeros(
+                            (pad_seq, *pixels.shape[1:]),
+                            dtype=pixels.dtype,
+                            device=pixels.device)
+                        kwargs[pixel_key] = torch.cat([pixels, pad_pixels_seq],
+                                                      dim=0)
+
+                        # Add a dummy grid entry for the sequence padding so the length matches
```
QiliangCui, line 664: **[major]** The batch-pad branch above extends `second_per_grid_ts` / `timestamps` to match the padded grid, but this seq-pad branch appends a grid entry without extending them. When it fires (e.g. TP=8 with total patches ≡ 4 mod 8), `grid` has b+1 entries vs b in those arrays — breaking models that consume them per grid entry (qwen2.5-VL's `second_per_grid_ts`).

It also invalidates the `valid_batch_sizes` bucketing computed just above (b → b+1 → a fresh XLA compile per shape). The two paddings should be computed together so the final grid length stays a bucketed size and every parallel array is extended consistently.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -428,6 +493,9 @@ def apply_qwen3_vl_patches(vllm_model):
             "Disabled dynamo for Qwen3LLMModel; JAX JIT handles outer compilation"
         )
 
+    # 8. Patch PyTorch SDPA globally to catch Qwen Vision Tower calls and prevent VMEM OOM
+    F.scaled_dot_product_attention = _chunked_sdpa
```
QiliangCui, line 497: **[major]** Patching `F.scaled_dot_product_attention` process-globally has a wide blast radius:

- The replacement drops `enable_gqa` (and any future kwargs) from the signature — any caller passing it gets a `TypeError` (`molmo2.py` does at the pinned vLLM). Please accept and forward `*args, **kwargs`.
- It intercepts every SDPA call in the process, not just the Qwen vision tower, and is never un-patched.
- The redirect is silent — a `logger.info_once` when the flash path engages would make the behavior swap visible in logs.

The vision callsite (`vllm/v1/attention/ops/vit_attn_wrappers.py`) resolves `F.scaled_dot_product_attention` at call time, so patching that module's binding (or the vision attention class) achieves the same effect with a much smaller radius.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -45,19 +45,78 @@
...
+    encoder_only_flash_attention
 from tpu_inference.logger import init_logger
+from tpu_inference.utils import align_to
 
 logger = init_logger(__name__)
 
+_orig_sdpa = F.scaled_dot_product_attention
+
+
+def _chunked_sdpa(query,
```
QiliangCui, line 69: **[nit]** `_chunked_sdpa` doesn't chunk — it redirects to `encoder_only_flash_attention`. Rename to reflect behavior (e.g. `_flash_attn_sdpa`), and lift `40 * 1024 * 1024` into a named constant. Side note: the GQA bail-out comment says the kernel requires `num_heads == num_kv_heads`, but the kernel signature accepts distinct `num_kv_heads` — the fallback is fine, the comment overstates the constraint.

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -165,20 +189,49 @@ def precompile_fn(run_compilation_fn: Callable) -> None:
...
-                },
-                num_patches=num_patches,
-            )
+            # By default, we precompile for common small frame counts to balance startup time.
+            # Users can override this via the VISION_PRECOMPILE_FRAMES environment variable
+            # (e.g., VISION_PRECOMPILE_FRAMES="1,2,4,8,16,64") to support specific video lengths.
+            #
+            # ⚠️ WARNING: Adding more frames or larger buckets here will significantly increase
+            # server startup time (XLA compilation) and can cause Host CPU OOMs during boot.
+            frame_counts = [1, 2, 4, 8, 16]
```
QiliangCui, line 198: **[major]** Startup-cost / coverage concerns with this expansion:

- Fan-out per patch bucket goes 1 → 12 compilations for the listed model types, and 18 for everything else. Qwen3-Omni's `model_type` (`qwen3_omni_moe`) is *not* in the tuple, so the currently-supported omni path now gets `video_grid_thw` and bare-`grid_thw` dummy calls it never received before, on top of the multiplier — and the `VISION_MIN_SHIFT` default change (4 → 6) silently drops the 16/32-patch buckets it used to warm.
- The grid args are jit `static_argnames`, so a precompiled `(t, 2^a, 2^b)` grid only matches media whose processed grid is exactly that. Real-world resolutions still compile at first request; the warm-up helps pow-2 shapes (like the 256x256x16 benchmark), but "fully warm up the JIT cache" overstates it.

Suggest scoping `frame_counts` to the video-capable archs that need it and keeping the previous defaults (min_shift=4, single image-grid call) for omni and image-only deployments.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -541,23 +549,223 @@ def embed_multimodal_func_torch(params_and_buffers: Any,
...
+                            (int, float, str, bool, jax.Array
+                             )) or v is None or type(v).__name__ == "GridTHW":
+                            return v
+                        if isinstance(v, (tuple, list)):
+                            return type(v)(move(x) for x in v)
+                        if not isinstance(v, torch.Tensor):
+                            logger.warning(
+                                f"Expect torch.Tensor, got {type(v)}")
+                            return v
+                        arr = t2j(v.contiguous(), use_dlpack=True)
```
QiliangCui, line 692: **[major]** `use_dlpack=False` was chosen explicitly when this conversion was introduced (#2396); flipping it to `True` (+ `.contiguous()`) changes the tensor-conversion path for **every** multimodal model, and is unrelated to the padding/sharding this PR is about. Was `False` load-bearing (dtype/layout issues)? If the flip is intentional, add a comment with the rationale and validation; otherwise keep `False` here and make the dlpack switch its own change.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -541,23 +549,223 @@ def embed_multimodal_func_torch(params_and_buffers: Any,
...
+                                    kwargs["timestamps"] = torch.cat(
+                                        [ts_vals, pad_ts_vals], dim=0)
+
+                    # 2. Pad pixel_values sequence length to multiple of tp_size to avoid full replication OOM
+                    # and pad grid_thw to match the sequence length
+                    if hasattr(self, "vllm_config") and hasattr(
+                            self.vllm_config, "parallel_config"):
+                        tp_size = self.vllm_config.parallel_config.tensor_parallel_size
+                    else:
+                        tp_size = 8
```
QiliangCui, line 633: **[minor]** `self.vllm_config` is always set in `__init__`, so the `hasattr` chain and this `tp_size = 8` fallback are dead — and wrong if they ever fired (a TP=2/TP=4 deployment would pad/shard as if TP=8). Also, divisibility for `PartitionSpec("model", None)` should be checked against the actual mesh axis size (`self.mesh.shape["model"]`), not `tensor_parallel_size` — they aren't guaranteed equal under all sharding configs. The same block is duplicated at ~705; extract one helper.

tpu_inference/models/vllm/vllm_model_wrapper.py
```diff
@@ -541,23 +549,223 @@ def embed_multimodal_func_torch(params_and_buffers: Any,
...
+                        if target_b is None:
+                            target_b = 1 << (b - 1).bit_length()
+
+                        if target_b > b:
+                            pad_b = target_b - b
+                            padded_anything = True
+
+                            # 1. Homogeneous padding to maximize cache hits
+                            grid_list = list(grid) if type(
+                                grid).__name__ == "GridTHW" else grid.tolist(
```
QiliangCui, line 579: **[minor]** `GridTHW` is imported at the top of this file — use `isinstance(grid, GridTHW)` instead of the `type(grid).__name__ == "GridTHW"` string check (4 occurrences), and this `list(grid) if ... else grid.tolist() if ... else list(grid)` ternary appears twice — worth a small `_grid_to_list(grid)` helper.

QiliangCui (changes requested):
Thanks — the problems this targets (host/HBM replication OOM on large video inputs, VMEM OOM in vision-tower SDPA, cold video compiles) are real, and the pixel-shard + strip-after-forward design and the GridTHW pytree/slice changes look sound. I reviewed this against the vLLM commit this repo pins (`.buildkite/vllm_lkg.version` = `0ba2aa35a`), and there are three things that need to be addressed before this can merge, plus inline comments.

**1. The mm receiver-cache integration (item 4 in the description) doesn't function at the pinned vLLM.** `engine_receiver_cache_from_config` is called with a `ModelConfig` where it takes the full `VllmConfig` → `AttributeError` at init of both disagg engine classes (any model, only under `PREFILL_SLICES`, which PR CI doesn't run). The application loops in `add_request` / `execute_model` check `mm_inputs` / `mm_positions`, which don't exist on the pinned request/scheduler types (only `mm_features` does), and `_apply_mm_cache` only exists on vLLM's `WorkerWrapperBase`, not on `TPUWorker` — so every branch is dead code. Details inline.

**2. No unit tests.** ~400 lines of production change across 5 modules with no test coverage — this repo requires unit tests for `tpu_inference/**` changes. The padding factorization, output-strip logic, `_chunked_sdpa` routing, and the GridTHW pytree round-trip are all cheaply unit-testable, and several of the inline findings would have been caught by them.

**3. No accuracy validation.** The PR validation is a throughput benchmark only. The padding/strip changes sit on the numerics path; please include an accuracy run (e.g. the CI accuracy harness or lm-eval on a VL benchmark) for at least one image and one video workload. Note the green CI here doesn't exercise the new seq-pad branch (its inputs are 4-aligned at small TP) nor the `PREFILL_SLICES` path.

Given the four changes are independent (pixel sharding+padding / SDPA flash redirect / precompile expansion / mm cache), splitting them into separate PRs would make each easy to test and land — the sharding+padding and SDPA pieces are closest to ready.

tpu_inference/models/vllm/experimental/vision_tower_jit.py
```diff
@@ -24,9 +24,12 @@
 from transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe import \
     Qwen3OmniMoeConfig
 from vllm.config import VllmConfig
+from vllm.model_executor.models.qwen3_5 import (
+    Qwen3_5ForConditionalGeneration, Qwen3_5MoeForConditionalGeneration)
```
lk-chen, line 28: these were intentionally removed in https://github.com/vllm-project/tpu-inference/pull/2878/changes

lk-chen: if you add a space around this line https://github.com/vllm-project/tpu-inference/blob/068cd11c3aa85fc48948f50492b14b528bb8a06e/.buildkite/benchmark/cases/daily/daily_qwen3_5_397b_fp8_tpu7x_8.json#L176, CI will trigger perf. benchmark, current throughput is [13.81 req/s](https://buildkite.com/tpu-commons/tpu-inference-benchmark/builds/1816#019fc23d-1d1e-4252-9021-25d1db76d4af/L2191) so you have a sense how much your change impacts

amitkumar307d: @lk-chen we can definitely leverage the benchmarking capability. One thing to note is the client args are different than what is needed. We need a video input with text and 1k output tokens.

amitkumar307d:
Split this PR into 3 PRs - 
https://github.com/vllm-project/tpu-inference/pull/3442
https://github.com/vllm-project/tpu-inference/pull/3443
https://github.com/vllm-project/tpu-inference/pull/3444

PR #3290, opened 2026-07-29
Add Batched RPA Kernel Decode Tuning/Gmm_v2 Kernel Tuner, Update Tuning Infra to Support Error Recovery/Resume

patrickji2014:
# Description
This PR introduces two new kernel tuners (Batched RPA Kernel Decode and Gmm_v2) and upgrades the underlying tuning infrastructure to implement a Multi-Process Execution Model for robust error recovery and resume capabilities.

## Why is this change being made?
In existing design, the tuning process becomes increasingly susceptible to hardware faults, Out-Of-Memory (OOM) crashes or device hangs. Previously, any kernel-level failure would crash the entire tuning process, requiring start the tuning sweep from scratch.

## The problem being solved and any relevant context
Kernel tuning often runs through thousands of configurations. In a single-process architecture, a single bad kernel run crashing the TPU runtime takes down the main tuner script with it. There was no clean way to isolate the execution faults from the state management of the tuner, resulting in lost progress and wasted compute time.

## Why this is a good solution
By transitioning the tuning infrastructure to an Architecture & Multi-Process Execution Model, the tuner's state management (driver process) from the actual kernel execution (worker process) are decoupled. If a kernel execution fails and crashes the worker, the main driver process remains alive, records the error for that specific configuration, recovers the state, and resumes tuning with the next configuration. This makes large sweeps significantly more resilient and time-efficient.

## Some information about the specific implementation

- Infrastructure Update: Refactored the core tuning loop to spawn isolated worker processes for kernel execution. Added state checkpointing to allow the tuner to resume from the last known good state in case of catastrophic failures.

- Batched RPA Kernel Decode Tuning: Update Batched RPA kernel tuner to support tuning decode cases

- Gmm_v2 Kernel Tuner: Added tuning support for Gmm_v2.

# Tests

Command to run in local TPU VMs
1. gmm_v2 kernel tuner
```
python -m tools.kernel.tuner.v1.kernel_tuner_runner --run_locally --kernel_tuner_name=gmm_v2_kernel_tuner --case_set_desc=bring_up_gmm_v2 --case_set_id=exp_gmm_v2_bring_up_0 --run_id=0 --tpu_version=tpu7x --tpu_cores=2
```
2. batched rpa kernel tuner
```
python -m tools.kernel.tuner.v1.kernel_tuner_runner --run_locally --kernel_tuner_name=batched_rpa_kernel_tuner --case_set_desc=batched_rpa_kernel_tuning_decode_q_128w --case_set_id=batched_rpa_decode_0 --run_id=0 --tpu_version=tpu7x --tpu_cores=2
```

Run in BuildKite
https://buildkite.com/tpu-commons/tpu-inference-kernel-tuning/builds/194/canvas?sid=019fe3fc-a1ae-4db4-81fb-0641593d5fab&open=false
```
 bkv_sz  --show  batch_size  --show  n_buffer --show latency_us --show-baseline
case     total_q_tokens  sliding_window  num_seqs  bq_sz  bq_c_sz  bkv_sz  batch_size  n_buffer  latency_us  baseline_latency  latency_improvement%
-------  --------------  --------------  --------  -----  -------  ------  ----------  --------  ----------  ----------------  --------------------
decode   512             1024            256       1      1        256     16          3         325         379               +14.2%              
decode   1024            1024            818       1      1        256     16          3         715         762               +6.2%               
decode   128             1024            818       1      1        256     16          3         117         125               +6.4%               
decode   16384           1024            818       1      1        256     16          3         716         761               +5.9%               
decode   16              1024            818       1      1        256     8           3         20          23                +13.0%              
decode   2048            1024            818       1      1        256     16          3         721         765               +5.8%               
decode   256             1024            818       1      1        256     16          3         225         242               +7.0%               
decode   32              1024            818       1      1        256     8           3         34          38                +10.5%              
decode   4096            1024            818       1      1        256     16          3         716         761               +5.9%               
decode   512             1024            818       1      1        256     16          3         443         477               +7.1%               
decode   64              1024            818       1      1        256     16          3         62          68                +8.8%               
decode   8192            1024            818       1      1        256     16          3         716         763               +6.2%               
decode   512             None            256       1      1        1536    1           3         357         363               +1.7%               
decode   1024            None            818       1      1        1536    2           3         752         757               +0.7%               
decode   128             None            818       1      1        1536    2           3         122         126               +3.2%               
decode   16384           None            818       1      1        1536    2           3         753         759               +0.8%               
decode   16              None            818       1      1        768     2           3         19          22                +13.6%              
decode   2048            None            818       1      1        1536    2           3         751         757               +0.8%               
decode   256             None            818       1      1        1536    2           3         239         244               +2.0%               
decode   32              None            818       1      1        768     2           3         34          37                +8.1%               
decode   4096            None            818       1      1        1536    2           3         752         758               +0.8%               
decode   512             None            818       1      1        1536    2           3         473         477               +0.8%               
decode   64              None            818       1      1        1536    1           3         64          68                +5.9%               
decode   8192            None            818       1      1        1536    2           3         752         758               +0.8%               
prefill  512             1024            256       256    64       256     1           3         9           12                +25.0%              
prefill  1024            1024            818       256    64       512     1           3         16          19                +15.8%              
prefill  128             1024            818       256    32       256     1           2         6           11                +45.5%              
prefill  16384           1024            818       512    64       512     1           3         159         162               +1.9%               
prefill  16              1024            818       256    64       256     1           2         5           10                +50.0%              
prefill  2048            1024            818       512    64       512     1           3         25          32                +21.9%              
prefill  256             1024            818       256    64       256     1           2         6           11                +45.5%              
prefill  32              1024            818       256    64       256     1           2         5           10                +50.0%              
prefill  4096            1024            818       512    64       512     1           2         45          51                +11.8%              
prefill  512             1024            818       256    64       256     1           3         9           12                +25.0%              
prefill  64              1024            818       256    32       256     1           2         6           11                +45.5%              
prefill  8192            1024            818       512    64       512     1           2         83          88                +5.7%               
prefill  512             None            256       256    64       256     1           3         11          13                +15.4%              
prefill  1024            None            818       256    32       512     1           3         23          25                +8.0%               
prefill  128             None            818       256    16       256     1           2         7           8                 +12.5%              
prefill  16384           None            818       256    32       256     3           3         241         243               +0.8%               
prefill  16              None            818       256    16       256     1           2         6           7                 +14.3%              
prefill  2048            None            818       256    64       256     2           3         39          39                +0.0%               
prefill  256             None            818       256    128      256     1           2         7           9                 +22.2%              
prefill  32              None            818       256    16       256     1           3         6           7                 +14.3%              
prefill  4096            None            818       256    64       256     2           3         68          69                +1.4%               
prefill  512             None            818       256    128      256     1           3         11          13                +15.4%              
prefill  64              None            818       256    64       256     1           2         6           8                 +25.0%              
prefill  8192            None            818       256    64       256     2           3         127         127               +0.0%               
```

# Checklist


tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -332,22 +301,23 @@ def _build_step(self, case_id_start: int, case_id_end: int,
...
                     f'  --use_bayesian_optimization={self.run_config.use_bayesian_optimization} '
                     f'  --n_bayesian_trials={self.tuner_config.n_bayesian_trials} '
+                    f'  --min_cases_for_bayesian={self.tuner_config.min_cases_for_bayesian} '
                     f'  --run_locally=False '
                     f'  --tpu_queue_multi={self.run_config.tpu_queue_multi} '
                     f'  --max_execution_minutes={self.run_config.max_execution_minutes} '
                     f'  --job_priority={self.run_config.job_priority} '
                     f'  --begin_case_id={case_id_start} --end_case_id={case_id_end}\''
-                ),
+                    f'  --worker_id={self.worker_id}'),
```
QiliangCui: This `--worker_id` lands outside the `bash -c '...'` payload: the `\'` at the end of the previous line closes the quote, so it is passed as an extra argument to `bash -c` (becomes `$0`) and never reaches `kernel_tuner_runner`. It works today only by accident: the flag is silently dropped and the worker falls back to env resolution (`TPU_WORKER_ID`/`HOSTNAME`), which is per-agent and correct.

Note that moving it inside the quotes would introduce a different bug: `self.worker_id` is resolved on the pipeline-generation machine, so every fanned-out job would be stamped with the generator's hostname instead of the agent that actually ran it. Suggest dropping the flag from the generated command entirely and letting each worker resolve its own id.

QiliangCui (commented):
Reviewed the multi-process tuner infra. Leaving the major findings as inline comments: the generated-step `--worker_id` quoting, two executor IPC edge cases (blank stdout line, unprefixed JSON), measurement-iters regression for timer-based tuners, the orphaned-executor kill path, and a bucket-id mismatch on the Bayesian-to-sweep fallback.

tools/kernel/tuner/v1/executor_process_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        while True:
+            elapsed = time.time() - start_time
+            remaining_timeout = max(1, int(timeout - elapsed))
+            if elapsed >= timeout:
+                raise TimeoutError(
+                    f"Executor did not respond within {timeout} seconds.")
+
+            line = self._read_with_timeout(remaining_timeout)
+            if not line:
+                raise EOFError("Executor stdout reached EOF unexpectedly.")
```
QiliangCui: A blank line on executor stdout crashes the worker: `_read_with_timeout` strips the trailing newline and returns `""`, so this `raise EOFError` fires — and `execute_run` doesn't catch `EOFError` (only BrokenPipeError/OSError/JSONDecodeError/TimeoutError), so it propagates and kills the worker process. The `if not line_str: continue` branch below is unreachable as a result.

Real EOF is already distinguished inside `_read_with_timeout` (`readline()` returning `""` raises BrokenPipeError). Suggest returning the raw line from `_read_with_timeout` (no strip) and stripping only here, treating an empty stripped line as `continue`.

tools/kernel/tuner/v1/executor_process_manager.py
```diff
@@ -0,0 +1,330 @@
...
+
+            line_str = line.strip()
+            if not line_str:
+                continue
+
+            if line_str.startswith("__JSON__"):
+                line_str = line_str[len("__JSON__"):].strip()
+
+            try:
+                return json.loads(line_str)
```
QiliangCui, line 276: Lines without the `__JSON__` prefix still fall through to `json.loads`, so any third-party stdout line that happens to be valid JSON (e.g. a library printing a dict/config) is consumed as the protocol response. `execute_run` then hits `result["status"]` -> KeyError, which is not in its except list and propagates uncaught — or worse, a wrong-but-plausibly-shaped dict is accepted as a real result.

Since the prefix exists precisely to disambiguate protocol messages from log noise, suggest `continue` unless the line starts with `__JSON__`.

tools/kernel/tuner/v1/executor_process_manager.py
```diff
@@ -0,0 +1,330 @@
...
+        logger.debug(f"Command: {' '.join(command)}")
+        self._proc = subprocess.Popen(
+            command,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            bufsize=1,  # line-buffered
+            env=env,
+            start_new_session=True,  # own process group for clean kill
```
QiliangCui, line 224: Because the executor gets its own session, the runner's `killpg` on a timed-out worker can't reach it. The stdin-EOF exit only helps when the executor is back at the request loop — an executor stuck in a hung TPU call (the main scenario this PR targets) survives as an orphan holding the exclusive TPU device, so every replacement executor fails to initialize and the 3-consecutive-start-failure abort fires repeatedly until the job container exits.

Options: set `preexec_fn` to call `prctl(PR_SET_PDEATHSIG, SIGKILL)` so the executor dies with the worker (CI runs on Linux), or record the executor pid somewhere the runner can kill it on worker timeout.

tools/kernel/tuner/v1/optimizer/sweep_optimizer.py
```diff
@@ -32,53 +33,189 @@ def generate_tuning_jobs(self, cases: list) -> list[tuple[int, int]]:
...
+            return status, 0
+
+        warmup_us = int(warmup_ns // 1000)
+
+        # --- Measurement Run ---
+        use_xprof = tuner.tuner_config.jit_kernel_pattern is not None
+        status, avg_latency_ns, _ = executor_mgr.execute_run(
+            tuning_key,
+            tunable_params,
+            iters=tuner._measurement_iters if not tuner.lightweight else 5,
```
QiliangCui: The worker always constructs the tuner with `lightweight=True`, so this always evaluates to 5 in production. For tuners without a `jit_kernel_pattern` (mla, flash_attention, example, rpa_v3) that silently drops measurement iterations from 100 to 5, making the timer-based latency numbers noisier — including for tuners used by the autotune pipeline.

`_measurement_iters` is set unconditionally in `KernelTunerBase.__init__` (5 with a jit pattern, 100 without), so it is already valid in lightweight mode: `iters=tuner._measurement_iters` should be enough here and at line 114.

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
```diff
@@ -85,81 +86,91 @@ def generate_tuning_jobs(self, cases: list) -> list[tuple[int, int]]:
...
-                "Cannot run Bayesian optimization; falling back to sequential sweep."
+                f"Tuning key {tuning_key} has {len(search_space)} cases in search space or "
+                f"less than {tuner.tuner_config.min_cases_for_bayesian} cases in bucket {begin_case_id}_{end_case_id}. "
+                "Cannot run Bayesian search strategy; falling back to sequential sweep. Update min_cases_for_bayesian and n_bayesian_trials via command line flags to use Bayesian search strategy."
             )
             from tools.kernel.tuner.v1.optimizer.sweep_optimizer import \
                 SweepOptimizer
-            SweepOptimizer(tuner).measure_latency(begin_case_id, end_case_id)
-            return
+            return SweepOptimizer(tuner, storage_manager,
```
QiliangCui, line 159: Bucket-id mismatch on this fallback path: in Bayesian mode `generate_buildkite_pipeline` creates bucket rows with `bucket_id = case_id_start`, but `SweepOptimizer.measure_latency` computes `bucket_id = begin_case_id // job_bucket_size`. The status updates from the fallback therefore target a row that doesn't exist (the Spanner UPDATE silently affects 0 rows; the local db appends a phantom row) and the real bucket stays IN_PROGRESS forever.

With the default `min_cases_for_bayesian=200` this fallback fires for every decode bucket (~100 cases each), so bucket status tracking is wrong for the common decode-tuning path. Suggest passing the bucket_id into `measure_latency` (or resolving it from storage by begin/end) instead of deriving it differently in each optimizer.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: thanks. drop the worker_id so each TPU VM resolve based on its env vars.

tools/kernel/tuner/v1/executor_process_manager.py
patrickji2014: fixed via removing the strip in read_with_timeout and apply the strip in the caller _read_json_response and treat it as a continue if empty.

tools/kernel/tuner/v1/executor_process_manager.py
patrickji2014: fixed via ignoring lines that don't start with __JSON__.

tools/kernel/tuner/v1/optimizer/sweep_optimizer.py
patrickji2014: fixed via dropping the iters' dependency on ligthweight config so the iters should always set to 100 for  jit_kernel_pattern None case

tools/kernel/tuner/v1/optimizer/bayesian_optimizer.py
patrickji2014: thanks. fixed via passing the bucket_id into the measure_latency function which update the bucket status based on the bucket id

tools/kernel/tuner/v1/executor_process_manager.py
patrickji2014: thanks. fixed by adding preexec_fn to kill the executor session/process when worker process exits.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -100,82 +125,74 @@ def __init__(self,
...
         self.run_config = run_config
         if run_config.n_bayesian_trials is not None:
             self.tuner_config.n_bayesian_trials = run_config.n_bayesian_trials
-        self.worker_id = run_config.worker_id or 'unknown_worker'
-        self.xprof_dir = os.path.join("/tmp/kernel_tuning",
-                                      self.tuner_config.kernel_tuner_name,
-                                      "xprof")
+        if run_config.min_cases_for_bayesian is not None:
+            self.tuner_config.min_cases_for_bayesian = run_config.min_cases_for_bayesian
         self.use_bayesian_optimization = tuner_config.support_bayesian_optimization and run_config.use_bayesian_optimization
```
QiliangCui, line 152: The last push removed `self.worker_id = get_worker_id(run_config.worker_id)` from `__init__`, but the optimizers still read it — `sweep_optimizer.py` lines 53 and 176 and `bayesian_optimizer.py` line 124 do `worker_id = tuner.worker_id` — so every real tuner run now raises `AttributeError`. It reproduces in this PR's own test:

```
$ pytest tools/kernel/tuner/v1/tests/test_kernel_tuner_base.py::WorkerIdResolutionTest::test_component_worker_id_initialization
E   AttributeError: 'MockKernelTuner' object has no attribute 'worker_id'
```

(The optimizer tests still pass because they use MagicMock tuners, which auto-create the attribute.) Either restore the assignment here, or switch the optimizers to read `self.storage_manager.worker_id` — the storage managers still resolve it via `get_worker_id`.

tools/kernel/tuner/v1/kernel_tuner_factory.py
```diff
@@ -0,0 +1,160 @@
...
+
+    Returns:
+        The kernel tuner class (a subclass of KernelTunerBase).
+
+    Raises:
+        ValueError: If the module or class cannot be found.
+    """
+    module_name = f'tools.kernel.tuner.v1.{kernel_tuner_name}'
+    try:
+        module = importlib.import_module(module_name)
```
QiliangCui, line 56: Lazy tuner loading breaks the documented tuner-specific flag mechanism. `.buildkite/pipeline_kernel_tuning.yml` converts every `KERNEL_TUNING_*` env var into a `--<flag>` on the `kernel_tuner_runner` command line, and the README documents naming tuner-specific flags `{kernel_name}_{flag}` / `KERNEL_TUNING_{KERNEL_NAME}_{FLAG}` (e.g. the ten `mla_*` flags in `mla_kernel_tuner.py`). Before this PR the registry imported every tuner module at import time, so those flags were registered before absl parsed argv. Now the tuner module is only imported here, after `app.run` has already parsed — so any tuner-specific env var fails the build with `FATAL Flags parsing error: Unknown command line flag`.

A related gap: even when such a flag does parse (local runs), it is never forwarded to the executor subprocess — `_get_flag_value` returns the flag default there — so an override changes case generation in the runner but not `generate_inputs` in the executor, producing inconsistent tuning inputs.

Suggestions: peek `--kernel_tuner_name` / `KERNEL_TUNING_KERNEL_TUNER_NAME` from `sys.argv`/env at import time and import the selected tuner module before flag parsing (or use a `flags_parser` with `known_only=True`), and forward tuner-specific flags to the worker/executor argv — or fold these knobs into `RunConfig` so they serialize through the existing JSON handoff.

QiliangCui (commented):
Thanks for the quick fixes in 914ea5c7 — the step-generation `--worker_id`, blank-line handling, measurement iters, pdeathsig, and the BO-fallback bucket_id all look right now. Two follow-ups from a second pass: the worker_id removal broke the optimizers (repros in this PR's own unit test), and the lazy tuner loading breaks the documented `KERNEL_TUNING_{TUNER}_{FLAG}` mechanism.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: thanks. I removed the unittest for testing the worker_id in the KernelTunerBase as it's not used anymore.

tools/kernel/tuner/v1/kernel_tuner_factory.py
patrickji2014: thanks. resolved by centralizing all the flag to one file and import flags before app.run. forward all flags according.

tools/kernel/tuner/v1/executor_process_manager.py
```diff
@@ -0,0 +1,350 @@
...
+        command = [
+            sys.executable,
+            "-m",
+            executor_module,
+            f"--run_config_path={self._run_config_path}",
+        ]
+
+        from tools.kernel.tuner.v1.kernel_tuner_flags import \
+            get_present_flag_args
+        command.extend(get_present_flag_args())
```
QiliangCui: This forwards **all** present flags with no exclusions. The runner always passes `--result_path` to the worker, so it is `present` in the worker's FLAGS and gets forwarded to the executor — which never registers that flag:

```
$ python -m tools.kernel.tuner.v1.kernel_tuner_executor --run_config_path=... --result_path=...
FATAL Flags parsing error: Unknown command line flag 'result_path'
```

So every executor spawn dies at parse, the 3-consecutive-start-failure cap aborts the worker, and every bucket fails. A second `--run_config_path` (the worker's) is also forwarded and silently overrides the explicit one above via absl last-wins.

Suggested fix — make the forwarding an allowlist instead of patching a blacklist (the next worker-only flag would re-break this the same way): have `get_present_flag_args` forward only flags defined in `kernel_tuner_flags.py` itself, via `flags.FLAGS.flags_by_module_dict().get(__name__, [])`, and use `flag.serialize()` rather than `f'--{name}={flag.value}'` (correct for booleans/lists). Then keep the explicit `--kernel_tuner_name={self._kernel_tuner_name}` in the command (the constructor arg is currently ignored) and exclude `{'kernel_tuner_name', 'begin_case_id', 'end_case_id'}` here.

A regression test that would have caught this and keeps the contract: a real-subprocess round-trip through `ExecutorProcessManager` with `example_kernel_tuner`, with `flags.FLAGS['result_path'].present = 1` set under `flagsaver` to simulate worker argv — the parse failure surfaces as "failed to start", no TPU needed.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
```diff
@@ -249,61 +266,21 @@ def get_search_space(self, tuning_key: TuningKey) -> dict:
...
+        yml_file_path = self.run_config.subbucket_yml_path(case_id_end)
+        safe_remove_files(yml_file_path)
+        from tools.kernel.tuner.v1.kernel_tuner_flags import \
+            get_present_flag_args
+        extra_flags = get_present_flag_args(exclude_flags={
+            'run_locally',
+            'begin_case_id',
+            'end_case_id',
+        })
+        extra_flags_str = '  '.join([''] +
```
QiliangCui: The forwarded flags are embedded **unquoted** into the single-quoted `bash -c '...'` payload. The previous hand-written command had `--case_set_desc=\"...\"` precisely because the inner shell word-splits values with spaces. Generated command on this head:

```
bash -c '... --case_set_desc=My tuning run description ...'
# inner bash argv: ['--case_set_desc=My', 'tuning', 'run', 'description']
```

The documented Buildkite flow passes `KERNEL_TUNING_CASE_SET_DESC="My tuning run description"`, so every generated bucket job fails on the stray args.

Suggested fix — double-quote the value of each forwarded arg when building this string (only here; the runner/manager subprocess paths use argv lists and are safe as-is). Inside the outer single-quoted `bash -c '...'`, double quotes pass through literally to the inner shell, which is what the old code relied on. Note `shlex.quote` is the wrong tool here: it emits single quotes, which terminate the outer `bash -c '...'` context and break the command differently.

```python
def _embed_flag_in_bash_c(arg: str) -> str:
    name, sep, value = arg.partition('=')
    if not sep:
        return arg  # boolean form '--name' / '--noname'
    assert "'" not in value, (
        f'{name} value must not contain single quotes when generating '
        'Buildkite steps (it would terminate the bash -c quoting)')
    escaped = (value.replace('\\', '\\\\').replace('\"', '\\\"')
                    .replace('$', '\\$').replace('`', '\\`'))
    return f'{name}=\"{escaped}\"'

extra_flags_str = ''.join(f'  {_embed_flag_in_bash_c(a)}' for a in extra_flags)
```

The single-quote limitation on values is pre-existing (the old command had it too); the assert makes it loud instead of silent. Also consider adding `worker_id` to the exclude set here — a `--worker_id` passed at pipeline-generation time would otherwise be stamped into every fanned-out job, re-creating the generator-hostname attribution problem through the generic forwarding.

QiliangCui (commented):
Two issues in the new flag-forwarding commit (6f9fcc5a), both verified by running against this head: the executor now fails to start because worker-only flags are forwarded to it, and the generated Buildkite step embeds forwarded flags unquoted so any case_set_desc with spaces breaks the bucket jobs. Inline comments have repros and suggested fixes.

tools/kernel/tuner/v1/executor_process_manager.py
patrickji2014: fixed via adding the flag names into exclude list. And update get_present_flag_args to use flags_by_module_dict to query module owned flags which will avoid passing un-owned flags to subprocess.

tools/kernel/tuner/v1/common/kernel_tuner_base.py
patrickji2014: thanks. fixed by using the _embed_flag_in_bash_c above.

QiliangCui (approved):

PR #3295, opened 2026-07-30
[DSv4] sync latest attn kernels and small fixes to compressor kernels

gxd3:
[DSv4] sync the latest core-attention and indexer kernels to tpu-inference, small fixes for compressor kernels.

* For the core-attention and indexer kernels, just copy paste from g3. The main diff from before: have a pipelined-gather-and-attend kernel for DSA. gather is offload to a SC kernel.

* The fixes for compressor kernels is found during integration of new kernels e2e.

E2E integration is not done in this PR, this PR just add / fix the new kernels, on caller yet.

(The E2E mostly working PR is in https://screenshot-v2.corp.google.com/6wJEcD48v6T4i6M; test cmd + MMLU eval: https://paste.googleplex.com/5954943549833216; But there is 1-1.5% MMLU score drop vs current HEAD, need some time to debug; there may additional issue in the integration code or in the new kernels.)


tpu_inference/kernels/experimental/deepseek_v4/core_attention/csa_gather.py
```diff
@@ -0,0 +1,346 @@
...
+        nope_in_specs = tuple(
+            pl.BlockSpec(
+                (pl.Indirect(row_subchunk_size), nope_in_cols),
+                lambda r, s=s: (idx_window(r, s), 0),
+            ) for s in range(num_streams))
+        # rope: gather int32 row == index // in_packing (in_packing entries/row).
+        rope_in_specs = tuple(
+            pl.BlockSpec(
+                (pl.Indirect(row_subchunk_size), rope_in_cols),
+                lambda r, s=s: (lax.div(idx_window(r, s), in_packing), 0),
```
inucool12, line 230: Are you assuming 4 tokens in the same row? I think ROPE cache is 1 token per 128B row (non-interleaved). If so, seems to me we shouldn't divide idx by 4 in rope_in_specs or shift by idx % 4 in bf16_bits.

gxd3: For ROPE, each token is (1, 128) uint8 (which encodes 64 bf16).
But the sparse core gather need to operate on 32-bit-words, so it gather (1, 128) int32 at a time, which is (4, 128) uint8.
Does it answer your question?

inucool12: Thanks for clarifying! Just one more clarification.

I think rope_in_cols is actually 32 int32s, not 128 int32s. So gathering 1 row (rope_in_cols) reads 32 int32s = 128B = 1 token (not 4 tokens). 
Since 1 row is 1 token, the row index should be idx. Am I missing something?

gxd3: about "rope_in_cols = rope_in_i32.shape[1]"
rope_in_cols is 128 not 32.

an example shape for `rope_in_hbm_ref` & `rope_in_i32`:
rope_in_hbm_ref:  (256000, 128) uint8
rope_in_i32:  (64000, 128) int32

does this address your confusion?

a1yssan13 (approved):

tpu_inference/kernels/experimental/deepseek_v4/core_attention/csa_gather.py
inucool12: Ah, got it. thanks.

inucool12 (approved):

PR #3298, opened 2026-07-30
Use server-reported token counts in agentic benchmark

wenxindongwork:
## Problem

The throughput this benchmark reports was derived from client-side tokenizer estimates rather than the token counts the engine actually processed. 

**Prompt.** The count was:

```python
"input_history_tokens": len(tokenizer.encode(str(messages[:-1]))),
```

`str(messages[:-1])` is the **Python repr** of a list of dicts, so every `{`, `'role':`, `'content':`, quote and comma is tokenized as a prompt token, while the chat template's own control tokens and generation prompt are absent.  

| messages | server (truth) | `str(messages)` |
|---:|---:|---:|
| 2 | 824 | 831 (+0.8%) |
| 22 | 3,444 | 3,591 (+4.3%) |
| 52 | 7,374 | 7,731 (+4.8%) |
| 100 | 13,662 | 14,355 (+5.1%) |

## Fix

Request usage explicitly:

```python
payload["stream_options"] = {"include_usage": True}
```

vLLM then emits a final chunk with an empty `choices` list carrying `usage`; take `prompt_tokens` and `completion_tokens` from it. 

A server that does not report usage is a **hard error** .

### DIFF
--- a/scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
+++ b/scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
@@ -124,10 +124,14 @@ async def run_grpo_stream(
         }
         if args.ignore_eos:
             payload["ignore_eos"] = True
+        # Ask the server to report the token counts it actually processed.
+        # Sent as a final chunk with an empty "choices" list.
+        payload["stream_options"] = {"include_usage": True}
 
         headers = {"Authorization": f"Bearer {os.getenv('HF_TOKEN', '')}"}
         start_time = time.perf_counter()
         ttft = None
+        usage = None
         full_response_text = []
 
         try:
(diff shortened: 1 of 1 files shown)


scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
```diff
@@ -168,6 +187,9 @@ async def run_grpo_stream(
                 "content": assistant_response
             })
 
+            if prompt_tokens is None:
+                prompt_tokens = len(tokenizer.encode(str(messages[:-1])))
```
sierraisland: Are we include chat template tokens here?

sierraisland (approved):
LGTM

scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
wenxindongwork: good catch, updated!

PR #3299, opened 2026-07-30
[Gemma4] Do not allocate K/V params for KV-shared layers

### LINKED ISSUE #3225: [Bug]: Gemma 4 E2B QAT checkpoints fail to load: k_norm required for KV-shared layers (num_kv_shared_layers ignored by loader)
### Summary

The Gemma 4 E2B QAT checkpoints fail to load on TPU in every currently serveable form, on both the JAX and torchax model paths. The root cause for the bf16 QAT checkpoint appears to be that the Gemma 4 model implementation instantiates `self_attn.k_norm` for **all** layers and requires them from the checkpoint, without accounting for `num_kv_shared_layers` — the KV-shared upper layers have no `k_norm` of their own, and the QAT export (correctly, per the config) omits them.

### Environment

- Image: `vllm/vllm-tpu:nightly` (vLLM `v0.23.1rc1.dev1076+g5c342876a`), pulled 2026-07-21
- Hardware: v6e-1 (`ct6e-standard-1t` GCE flex-start VM, `europe-west4-a`)
- `--tensor-parallel-size 1 --max-model-len 65536 --gpu-memory-utilization 0.9 --max_num_batched_tokens 4096 --disable_chunked_mm_input`

### Failure 1: `google/gemma-4-E2B-it-qat-q4_0-unquantized`, JAX path (default)

```
ValueError: Following weights were not initialized from checkpoint:
{'model.language_model.layers.15.self_attn.k_norm.weight', ...
 'model.language_model.layers.34.self_attn.k_norm.weight'}   # layers 15–34, i.e. exactly the 20 KV-shared layers
```

(With `--load-format runai_streamer` the same underlying problem surfaces later and more cryptically, as `TypeError: Argument 'model.states[0][223]' of shape bfloat16[256] of type <class 'jax.ShapeDtypeStruct'> is not a valid JAX type` during the first jitted call — the un-loaded param is still an abstract value.)

### Failure 2: same checkpoint, `MODEL_IMPL_TYPE=vllm` (torchax path)

```
File ".../model_loader/default_loader.py", line 467, in track_weights_loading
ValueError: Following weights were not initialized from checkpoint:
{'language_model.model.layers.15.self_attn.k_norm.weight', ... 'language_model.model.layers.34.self_attn.k_norm.weight'}
```

### Checkpoint forensics (why this is a loader bug, not a broken export)

Read directly from the safetensors headers of both HF repos:

| checkpoint | `k_norm` tensors present | config |
|---|---|---|
| `google/gemma-4-E2B-it` | all 35 language-model layers | `num_hidden_layers: 35`, `num_kv_shared_layers: 20` |
| `google/gemma-4-E2B-it-qat-q4_0-unquantized` | only the 15 non-KV-shared layers (+ towers) | identical |

The configs are byte-identical on these fields. Layers 15–34 are the 20 KV-shared layers; they reuse K/V from lower layers, so the QAT export ships no `k_norm` for them. The plain checkpoint happens to include (unused) `k_norm` tensors for those layers, which is the only reason it loads while the QAT one does not. The plain `google/gemma-4-E2B-it` serves fine in the same environment.

**Expected behavior:** the Gemma 4 implementation should not create/require `k_norm` (or K/V projections) for the `num_kv_shared_layers` upper layers, so that architecturally-minimal exports like the QAT one load.

### Failure 3 (related): `google/gemma-4-E2B-it-qat-w4a16-ct`

The compressed-tensors int4 QAT variant fails earlier, in scheme resolution:

```
File ".../tpu_inference/layers/jax/quantization/compressed_tensors.py", line 146, in get_quant_method
NotImplementedError: compressed-tensors scheme for layer 'model.language_model.per_layer_model_projection' is not yet supported in the JAX path.
```

Happy to provide full logs or test candidate fixes on live v6e capacity.

### Repro

```bash
docker run --privileged --net=host --shm-size 10gb -v /dev/shm:/dev/shm \
  -e HF_HOME=/dev/shm -e HF_TOKEN=... \
  vllm/vllm-tpu:nightly vllm serve google/gemma-4-E2B-it-qat-q4_0-unquantized \
  --tensor-parallel-size 1 --max-model-len 65536
```

xbill9:
## Summary

KV-shared layers in Gemma 4 (the last `num_kv_shared_layers`) read K/V out of the source layer's cache slot. `Gemma4Attention.__call__` already skipped `k_norm` and RoPE on them and called `attention(..., update_kv_cache=False)`, and the RPA v3 kernel sets `kv_left_frm_cache = kv_left` in that mode — so the layer's own k/v operands are never read. The layers were nevertheless allocating `k_proj`, `v_proj` and `k_norm`, and the loader then required those tensors from the checkpoint.

Gemma 4 QAT exports omit them, exactly as the config implies, so every QAT checkpoint failed to load:

```
ValueError: Following weights were not initialized from checkpoint:
{'model.language_model.layers.15.self_attn.k_norm.weight', ...
 'model.language_model.layers.34.self_attn.k_norm.weight'}
```

Layers 15–34 are the 20 KV-shared layers of E2B. Both released KV-share geometries are affected — read straight from the safetensors headers, the QAT exports carry K/V-side tensors for exactly the non-shared layers:

| checkpoint | layers | KV-shared | `k_proj`/`v_proj`/`k_norm` present |
| --- | --- | --- | --- |
| `google/gemma-4-E2B-it` | 35 | 15–34 | all 35 |
| `google/gemma-4-E2B-it-qat-q4_0-unquantized` | 35 | 15–34 | layers 0–14 only |
| `google/gemma-4-E4B-it` | 42 | 24–41 | all 42 |
| `google/gemma-4-E4B-it-qat-q4_0-unquantized` | 42 | 24–41 | layers 0–23 only |

Fixes #3225 on the JAX path.

## Changes

- **`Gemma4Attention.__init__`** — KV-shared layers build a standalone `q_proj` and no `qkv_proj`/`k_proj`/`v_proj`/`k_norm`. `__call__` passes zeros for k/v to satisfy the `attention()` signature; the kernel discards them in `update_kv_cache=False` mode.
- **`kv_share.drop_shared_kv_weights`** — filters those tensors out of the weight stream, so plain BF16 exports (which do ship them, unused) keep loading through the same model definition instead of hitting "is not a valid param path". Anchored on `language_model.layers.{i}.self_attn.`, so the vision and audio towers' own per-layer `k_proj`/`v_proj` are untouched — dropping those would have left a partly random encoder rather than raising.
- Applied in both `Gemma4ForCausalLM.load_weights` and `Gemma4ForConditionalGeneration.load_weights`.

Weight routing needs no change: `JaxAutoWeightsLoader._route` falls through when no fused `qkv_proj` param exists at that path, so the standalone `q_proj` loads normally — the same mechanism the existing `use_k_eq_v` branch relies on.

## Test plan

New cases in `tests/models/jax/test_gemma4_kv_share.py`:

- `test_shared_layers_allocate_no_kv_params` — non-shared layers keep the fused QKV and `k_norm`; shared layers have `q_proj` only.
- `test_drop_shared_kv_weights_filters_only_shared_layers` — BF16 stream: shared layers' K/V-side tensors dropped, everything else kept.
- `test_drop_shared_kv_weights_is_noop_for_qat_checkpoint` — the #3225 checkpoint shape passes through unchanged.
- `test_drop_shared_kv_weights_noop_without_sharing` — `num_kv_shared_layers=0` (26B/31B path).
- `test_drop_shared_kv_weights_spares_the_vision_tower` — vision/audio tower attention weights survive.
- `test_drop_shared_kv_weights_matches_unprefixed_names` — anchors with or without the leading `model.`.
- `test_drop_shared_kv_weights_on_released_geometries` — parametrized over the real E2B (35/20) and E4B (42/18) configs, asserting the shared range and that exactly those layers lose their K/V-side tensors.

### Verified on hardware (v6e-1, `ct6e-standard-1t`, europe-west4-a)

`vllm/vllm-tpu:nightly` (tpu-inference at `30dcf7ad0`) vs. the same image with this patch applied, `--tensor-parallel-size 1 --max-model-len 65536`:

| image | checkpoint | result |
| --- | --- | --- |
| nightly | `gemma-4-E2B-it` (BF16) | serves |
| nightly | `gemma-4-E2B-it-qat-q4_0-unquantized` | ❌ `ValueError: Following weights were not initialized from checkpoint` — `k_norm` for layers 15–34; engine exits 1 |
| **patched** | `gemma-4-E2B-it-qat-q4_0-unquantized` | ✅ weights load in 43.2 s, `Application startup complete`, generates correctly |
| **patched** | `gemma-4-E2B-it` (BF16) | ✅ weights load in 44.8 s — the path where `drop_shared_kv_weights` actually filters — generates correctly |
| nightly | `gemma-4-E4B-it-qat-q4_0-unquantized` | ❌ same failure, `k_norm` for layers 24–41 |
| **patched** | `gemma-4-E4B-it-qat-q4_0-unquantized` | ✅ weights load in 57.4 s, serves, generates correctly |

The failing layer ranges match each config's KV-share geometry exactly (E2B 35/20 → 15–34, E4B 42/18 → 24–41).

`pytest tests/models/jax/test_gemma4_kv_share.py tests/models/jax/test_gemma4_ple.py` → **28 passed on the v6e-1**.

Unaffected by construction: `gemma-4-31B-it` and `gemma-4-26B-A4B-it` both declare `num_kv_shared_layers: 0`, so `compute_kv_share_map` returns `{}`, the filter is a pass-through, and every layer takes the pre-existing `use_k_eq_v` branch.

Scope note: this fixes the JAX path only. The same checkpoint also fails on the torchax path (Failure 2 in #3225), which lives in vllm's own Gemma 4 model, and the `-qat-w4a16-ct` variant fails earlier in compressed-tensors scheme resolution (Failure 3) — both left for follow-ups.

### DIFF
--- a/tpu_inference/models/jax/gemma4.py
+++ b/tpu_inference/models/jax/gemma4.py
@@ -42,7 +42,8 @@ from tpu_inference.layers.jax.rope_interface import (apply_rope,
                                                      normalize_rope_scaling)
 from tpu_inference.layers.vllm.quantization.configs import VllmQuantConfig
 from tpu_inference.logger import init_logger
-from tpu_inference.models.common.kv_share import compute_kv_share_map
+from tpu_inference.models.common.kv_share import (compute_kv_share_map,
+                                                  drop_shared_kv_weights)
 from tpu_inference.models.jax.jax_intermediate_tensor import \
     JaxIntermediateTensors
 from tpu_inference.models.jax.utils.weight_utils import (
@@ -314,7 +315,50 @@ class Gemma4Attention(JaxModule):
                            None) if _shard_kv_on_k else (None, None, "model")
         _kv_bias_spec = ("model", None) if _shard_kv_on_k else (None, "model")
 
-        if use_k_eq_v:  # TODO: Add QKV fusion logic for k == v case.
+        # KV-cache sharing (mirrors vllm-pytorch `Gemma4Attention.__init__`
+        # KV-share derivation). Layers in the last `num_kv_shared_layers`
+        # reuse K/V from earlier layers of matching attention type. The
+        # runner side populates the redirect mapping; this layer just needs
+        # to set is_kv_shared_layer + kv_sharing_target_layer_name.
+        # Derived before the projections are built because shared layers do
+        # not own K/V parameters at all (see below).
+        kv_share_map = compute_kv_share_map(config)
+        self.is_kv_shared_layer = layer_idx in kv_share_map
+        self.kv_sharing_target_layer_name: Optional[str] = None
+        if self.is_kv_shared_layer:
+            # The runner uses unprefixed "layer.{i}" keys; this string must
+            # match the keys produced by KVCacheManager's spec-creation loop
+            # and Gemma4Model's layer-name iteration.
+            self.kv_sharing_target_layer_name = (
+                f"layer.{kv_share_map[layer_idx]}")
+
+        if self.is_kv_shared_layer:
+            # Shared layers never use their own K/V: `__call__` reads the
+            # source layer's already-normed-and-roped K/V straight from the
+            # redirected cache slot, so k_proj/v_proj/k_norm are dead weight
+            # and dead compute. Gemma 4 QAT exports omit them entirely for
+            # these layers, so allocating them makes those checkpoints fail
+            # to load (vllm-project/tpu-inference#3225). BF16 exports do ship
+            # them; every `load_weights` that builds this module drops them on
+            # the way in via `drop_shared_kv_weights`.
+            self.qkv_proj = None
+            self.q_proj = JaxEinsum(
+                "TD,DNH->TNH",
+                (self.hidden_size, self.num_heads, self.head_dim),
+                bias_shape=(self.num_heads,
+                            self.head_dim) if config.attention_bias else None,
+                param_dtype=dtype,
+                kernel_init=nnx.with_partitioning(init_fn,
+                                                  (None, "model", None)),
+                bias_init=nnx.with_partitioning(init_fn, ("model", None))
+                if config.attention_bias else None,
+                rngs=rng,
+                quant_config=quant_config,
+                prefix=prefix + ".q_proj",
+            )
+            self.k_proj = None
+            self.v_proj = None
+        elif use_k_eq_v:  # TODO: Add QKV fusion logic for k == v case.
             self.qkv_proj = None
             self.q_proj = JaxEinsum(
                 "TD,DNH->TNH",
@@ -370,15 +414,21 @@ class Gemma4Attention(JaxModule):
             prefix=prefix + ".q_norm",
         )
 
-        self.k_norm = JaxRmsNorm(
-            self.head_dim,
-            epsilon=self.rms_norm_eps,
-            param_dtype=dtype,
-            scale_init=nnx.with_partitioning(init_fn, (None, )),
-            rngs=rng,
-            quant_config=quant_config,
-            prefix=prefix + ".k_norm",
-        )
+        # Only non-shared layers norm their own K (see `__call__`); shared
+        # layers read K/V that the source layer already normed. Assigned once
+        # per branch: nnx rejects re-binding a static None to a Module.
+        if self.is_kv_shared_layer:
+            self.k_norm = None
+        else:
+            self.k_norm = JaxRmsNorm(
+                self.head_dim,
+                epsilon=self.rms_norm_eps,
+                param_dtype=dtype,
+                scale_init=nnx.with_partitioning(init_fn, (None, )),
+                rngs=rng,
+                quant_config=quant_config,
+                prefix=prefix + ".k_norm",
+            )
         # V norm: no learnable scale (pure normalization only)
         self.v_norm = JaxRmsNorm(
             self.head_dim,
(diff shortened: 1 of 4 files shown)


tpu_inference/models/jax/gemma4.py
```diff
@@ -314,7 +315,50 @@ def __init__(self,
...
+            # The runner uses unprefixed "layer.{i}" keys; this string must
+            # match the keys produced by KVCacheManager's spec-creation loop
+            # and Gemma4Model's layer-name iteration.
+            self.kv_sharing_target_layer_name = (
+                f"layer.{kv_share_map[layer_idx]}")
+
+        if self.is_kv_shared_layer:
+            # Shared layers never use their own K/V: `__call__` reads the
+            # source layer's already-normed-and-roped K/V straight from the
+            # redirected cache slot, so k_proj/v_proj/k_norm are dead weight
```
lk-chen: what does "dead" mean here

tpu_inference/models/jax/gemma4.py
```diff
@@ -314,7 +315,50 @@ def __init__(self,
...
+        kv_share_map = compute_kv_share_map(config)
+        self.is_kv_shared_layer = layer_idx in kv_share_map
+        self.kv_sharing_target_layer_name: Optional[str] = None
+        if self.is_kv_shared_layer:
+            # The runner uses unprefixed "layer.{i}" keys; this string must
+            # match the keys produced by KVCacheManager's spec-creation loop
+            # and Gemma4Model's layer-name iteration.
+            self.kv_sharing_target_layer_name = (
+                f"layer.{kv_share_map[layer_idx]}")
+
```
lk-chen: I think we can remove line 334 and 335?

lk-chen (approved):
some nits, LGTM

tpu_inference/models/jax/gemma4.py
xbill9: nothing ever reads them.  maybe this is better wording?  nothing ever reads what k_proj/
 v_proj/k_norm would produce

tpu_inference/models/jax/gemma4.py
xbill9: good catch! yes!

PR #3300, opened 2026-07-30
[Multi-Host] Fix HBM memory stats on worker hosts and propagate dtype in sharding callback

tyuchn:
## Description

This PR resolves two multi-host initialization issues in `tpu-inference` and adds comprehensive unit test coverage:

### 1. Fix HBM Memory Stats on Non-Zero Multi-Host Workers
- **Issue**: In `hbm_usage_bytes()`, iterating over the global `devices` list (`jax.devices()`) causes non-zero multi-host workers (Host 1, Host 2, etc.) to query `device.memory_stats()` on remote devices (Host 0 chips). This causes PjRt to throw `INVALID_ARGUMENT: MemoryStats is only supported for addressable PjRt devices`, leading to an empty `usage = []` return and calculating `0.0 GiB` available HBM on worker nodes.
- **Fix**: Query `jax.local_devices()` when `TPU_MULTIHOST_BACKEND == 'ray'`, ensuring each worker only queries its local addressable chips.
- **Improvement**: Replaced bare `except Exception: pass` with `logger.warning()` so any device failure is logged with the error reason.

### 2. Propagate `dtype` in `general_device_put` Callback
- **Issue**: `jax.make_array_from_callback` in `general_device_put` omitted the `dtype` argument, which could lead to type mismatch during distributed tensor sharding callbacks.
- **Fix**: Explicitly pass `dtype=t.dtype` to `jax.make_array_from_callback`.

---

## Test Plan
- Added `test_hbm_usage_bytes_ray_backend_leader` and `test_hbm_usage_bytes_ray_backend_worker_host` in `tests/test_utils.py` verifying behavior on worker nodes where remote devices precede local devices.
- Added `test_hbm_usage_bytes_single_host_success` and `test_hbm_usage_bytes_single_host_partial_failure` verifying fallback behavior in single-host setups.
- Updated `tests/layers/common/test_utils.py` to verify `dtype` propagation.
- Verified on a live 4-node Trillium TPU cluster with multi-host 480B model serving.

### DIFF
--- a/tests/layers/common/test_utils.py
+++ b/tests/layers/common/test_utils.py
@@ -73,12 +73,10 @@ class UtilsTest(jtu.JaxTestCase):
                 result = general_device_put(tensor, self.sharding)
 
                 mock_make_array.assert_called_once()
-                args, _ = mock_make_array.call_args
+                args, kwargs = mock_make_array.call_args
                 self.assertEqual(args[0], tensor.shape)
                 self.assertEqual(args[1], self.sharding)
-                # Check that x_split contains device_put result
-                # Since we didn't mock device_put, it runs on the tensor slice.
-                # We can check the length of x_split
+                self.assertEqual(kwargs.get('dtype'), tensor.dtype)
                 self.assertAllClose(result, tensor)
 
     def test_general_device_put_multihost_pytree(self):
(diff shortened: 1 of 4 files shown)


tyuchn:
cc @Lumosis

PR #3305, opened 2026-07-31
[JAX] Add int8 W8A8 compressed-tensors support to the JAX path

shungcp:
# Description

The JAX path only supported fp8 compressed-tensors schemes; int8 W8A8
(`int-quantized`) checkpoints hit `NotImplementedError` in
`CompressedTensorsConfig.get_quant_method`.

This adds `Int8ChannelwiseLinearMethod` for the llm-compressor style scheme
(int8 per-channel static weights + dynamic per-token int8 activations). It
reuses the existing `xla_quantized_matmul`, which already handles integer
weight dtypes, so only the int8 weight dtype and channelwise `weight_scale`
are new.

Fused `JaxMergedColumnParallelLinear` int8 raises `NotImplementedError` for
now, matching the fp8 blockwise branch.

# Tests

New `test_int8.py` (5 tests: dispatch, ignore-regex, static-act rejection,
param shapes, forward vs fp32 dequant reference).

```
JAX_PLATFORMS=cpu python -m pytest tests/layers/jax/quantization/test_int8.py -v
```

All 5 pass; also validated on TPU v6e-1 with a real int8 W8A8 checkpoint
(parity with the torchax int8 path).

# Checklist

### DIFF
--- a/tests/layers/jax/quantization/test_int8.py
+++ b/tests/layers/jax/quantization/test_int8.py
@@ -1,0 +1,163 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for the JAX-native int8 W8A8 compressed-tensors linear method."""
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+import pytest
+from flax import nnx
+from jax.sharding import Mesh
+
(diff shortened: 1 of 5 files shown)


lk-chen:
did you test against e.g. https://huggingface.co/RedHatAI/Qwen3.5-4B-quantized.w8a8

could you add to https://github.com/vllm-project/tpu-inference/tree/main/tests/models/jax

shungcp:
Added a model-level test under `tests/models/jax/` (`test_qwen3_int8.py`) that loads a real public `int-quantized` checkpoint — `RedHatAI/Qwen3-4B-Instruct-2507-quantized.w8a8` (same channelwise-int8-W / dynamic-per-token-int8-A scheme) — and asserts the linears route to `Int8ChannelwiseLinearMethod` with int8 weights + a 1-D channelwise float scale, then runs a forward producing finite outputs.

I went with the dense `Qwen3ForCausalLM` checkpoint rather than the Qwen3.5 VLM you suggested: `Qwen3_5ForConditionalGeneration` isn't in the JAX `_MODEL_REGISTRY` (it's routed to the vLLM/torchax path via `_VLLM_PREFERRED_ARCHITECTURES`), so it wouldn't exercise this JAX method — the Qwen3 checkpoint uses the identical quant scheme and hits the code path directly.

PR #3312, opened 2026-07-31
Fix nightly failures against new vLLM LKG (is_stale kwarg + MM test fault_tolerance_config)

QiliangCui:
## Problem

The 07-31 nightly (build 23558) fails on both tpu6e and tpu7x with two distinct regressions introduced by the vLLM LKG advancing to `a7a204cc6`:

1. **moe_expert_ids / continue_decode scheduler patch**: vLLM added an `is_stale` kwarg to `Scheduler._update_request_with_output` and `AsyncScheduler._update_request_with_output` (the caller in `update_from_output` now passes `is_stale=output_is_stale`). The continue_decode monkeypatch wrappers in `tpu_inference/core/sched/utils.py` kept the old 2-arg signature, so any engine with the patched scheduler dies:
   `TypeError: patch_vllm_scheduler_for_continue_decode.<locals>.patched_update_base() got an unexpected keyword argument 'is_stale'` -> `EngineDeadError`.

2. **test_multi_modal_inference**: the test round-trips `asdict(EngineArgs(...))` into `LLM(...)`. `asdict` converts the default `FaultToleranceConfig` dataclass into a plain dict, and `EngineArgs.__post_init__` now treats a dict `fault_tolerance_config` as an explicit opt-in and force-enables fault tolerance, which validation then rejects:
   `ValueError: Fault tolerance requires external load balancer mode ... Internal LB mode is not supported.`

## Fix

- `tpu_inference/core/sched/utils.py`: both patched wrappers accept and forward `is_stale`. The `num_computed_tokens` advance and the `num_output_placeholders` pre-compensation are skipped for stale deliveries, mirroring the `if not output_is_stale` guards vLLM applies around its own counter adjustments (stale output predates the preemption rollback of those counters).
- `tests/e2e/test_multi_modal_inference.py`: drop `fault_tolerance_config` from the asdict dict before `LLM(...)` (same class of workaround as the existing pass_config None-stripping) so the default disabled config is used.
- New `tests/core/test_sched_utils.py` covering the wrapper signatures and both is_stale guards.

## Validation (local tpu7x, vllm at LKG a7a204cc6)

- `tests/runner/test_moe_expert_ids.py::TestMoEExpertIds::test_moe_expert_ids_sync_path_prompt_populated` (the failing nightly test): **1 passed** (was EngineDeadError).
- `tests/e2e/test_multi_modal_inference.py` both params: **2 passed** (was ValueError at engine init).
- `tests/core/test_sched_utils.py`: 3 passed.
- Verified the mechanism directly: `EngineArgs(fault_tolerance_config=<dict from asdict>)` -> `enable_fault_tolerance=True`; field omitted -> `False`.

## Dev-pipeline verification (test evidence)

Dev build [tpu-inference-dev #716](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/716) — **PASSED** — ran this branch's fixes with the exact commands of the failing nightly steps (`rl/moe_expert_ids.yml`, `features/Multimodal_Inputs.yml`) on both platforms, plus the new unit tests:

| Job | Result | Nightly 23558 (before fix) |
|---|---|---|
| tpu6e `pytest tests/core/test_sched_utils.py tests/runner/test_moe_expert_ids.py` | **7 passed** (47:19) | FAILED (`EngineDeadError`, is_stale TypeError) |
| tpu7x `pytest tests/core/test_sched_utils.py tests/runner/test_moe_expert_ids.py` | **7 passed** (31:45) | FAILED (same) |
| tpu6e `pytest tests/e2e/test_multi_modal_inference.py` | **2 passed** (5:12) | FAILED (fault-tolerance ValueError at init) |
| tpu7x `pytest tests/e2e/test_multi_modal_inference.py` | **2 passed** (3:52) | FAILED (same) |

### DIFF
--- a/tests/core/test_sched_utils.py
+++ b/tests/core/test_sched_utils.py
@@ -1,0 +1,93 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import inspect
+from unittest import mock
+
+from tpu_inference.core.sched.utils import \
+    patch_vllm_scheduler_for_continue_decode
+
+
+def _make_request(num_computed_tokens=10, num_output_placeholders=1):
+    request = mock.MagicMock()
+    request.num_computed_tokens = num_computed_tokens
+    request.num_output_placeholders = num_output_placeholders
+    return request
+
+
+class TestContinueDecodeSchedulerPatch:
+
+    def test_patched_signatures_accept_is_stale(self):
+        """vLLM passes is_stale=... as a kwarg from update_from_output; the
+        patched wrappers must accept it or the EngineCore dies with a
+        TypeError."""
+        patch_vllm_scheduler_for_continue_decode()
+        from vllm.v1.core.sched.async_scheduler import AsyncScheduler
+        from vllm.v1.core.sched.scheduler import Scheduler
+
+        for cls in (Scheduler, AsyncScheduler):
+            params = inspect.signature(
+                cls._update_request_with_output).parameters
+            assert "is_stale" in params, (
+                f"{cls.__name__}._update_request_with_output must accept "
+                f"is_stale (got {list(params)})")
+
+    def test_sync_patch_advances_num_computed_tokens_unless_stale(self):
+        patch_vllm_scheduler_for_continue_decode()
+        from vllm.v1.core.sched.scheduler import Scheduler
+
+        scheduler = mock.MagicMock()
+        with mock.patch("vllm.v1.core.sched.scheduler.check_stop",
+                        return_value=False):
+            # Fresh output: 3 tokens returned for 1 scheduled step -> advance
+            # num_computed_tokens by the extra 2 on-device tokens.
+            request = _make_request(num_computed_tokens=10)
+            Scheduler._update_request_with_output(scheduler,
+                                                  request, [1, 2, 3],
+                                                  is_stale=False)
+            assert request.num_computed_tokens == 12
+
+            # Stale output predates the preemption rollback -> no advance.
+            request = _make_request(num_computed_tokens=10)
+            Scheduler._update_request_with_output(scheduler,
+                                                  request, [1, 2, 3],
+                                                  is_stale=True)
+            assert request.num_computed_tokens == 10
+
+    def test_async_patch_placeholder_compensation_respects_is_stale(self):
+        patch_vllm_scheduler_for_continue_decode()
+        from vllm.v1.core.sched.async_scheduler import AsyncScheduler
+
+        # spec= so the zero-arg super() in the wrapped original resolves
+        # (isinstance check); instance attributes read by the base
+        # implementation must be set explicitly.
+        scheduler = mock.MagicMock(spec=AsyncScheduler)
+        scheduler.max_model_len = 1024
+        with mock.patch("vllm.v1.core.sched.scheduler.check_stop",
+                        return_value=False):
+            # Fresh output: pre-compensate placeholders by (N - 1) so the
+            # original's -= N lands back at 0 without underflowing.
+            request = _make_request(num_output_placeholders=1)
+            AsyncScheduler._update_request_with_output(scheduler,
+                                                       request, [1, 2, 3],
+                                                       is_stale=False)
+            assert request.num_output_placeholders == 0
+
+            # Stale output: placeholders were zeroed at preemption; neither the
+            # pre-compensation nor the original's decrement may touch them.
+            request = _make_request(num_output_placeholders=1)
+            AsyncScheduler._update_request_with_output(scheduler,
+                                                       request, [1, 2, 3],
+                                                       is_stale=True)
+            assert request.num_output_placeholders == 1
--- a/tests/core/test_sched_utils.py
+++ b/tests/core/test_sched_utils.py
@@ -15,8 +15,29 @@
 import inspect
 from unittest import mock
 
+import pytest
+
 from tpu_inference.core.sched.utils import \
     patch_vllm_scheduler_for_continue_decode
+
+
+@pytest.fixture(autouse=True)
+def restore_scheduler_patch():
+    """The continue_decode patch is a process-global, irreversible monkeypatch
+    (including Scheduler.__init__, which forces num_lookahead_tokens >= 9).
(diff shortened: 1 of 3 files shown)


wenxindongwork (approved):

tpu_inference/core/sched/utils.py
```diff
@@ -71,16 +79,22 @@ def patched_init(scheduler_self, vllm_config, *args, **kwargs):
...
                 # num_output_placeholders cleanly decrements by 1 without underflowing < 0.
+                # Placeholders are zeroed at preemption, so a stale delivery must
+                # not be pre-compensated (the original skips its decrement too).
                 request.num_output_placeholders += (len(new_token_ids) - 1)
-            return original_async_update_req(scheduler_self, request,
-                                             new_token_ids)
+            return original_async_update_req(scheduler_self,
+                                             request,
+                                             new_token_ids,
+                                             is_stale=is_stale)
```
QiliangCui: **blocker — async stale path bypasses the new guard.**

vLLM's `AsyncScheduler._update_request_with_output` calls `super()._update_request_with_output(request, new_token_ids)` **without** forwarding `is_stale` (`vllm/v1/core/sched/async_scheduler.py` L51-70 at `a7a204cc6`). That super() call resolves to our patched base, which then runs with its `is_stale=False` default — so a stale async delivery still advances `num_computed_tokens` by N-1 on a preempted request whose counter was just reset to 0 and whose blocks were freed (`_preempt_request`, `scheduler.py` L1287/L1300). Verified against vLLM at the LKG: a stale 3-token async delivery moves `num_computed_tokens` 0 -> 2; on reschedule those positions are treated as computed with no KV backing.

Note stale deliveries only arise under async scheduling (see the `_preempt_request` comment), so the sync-side guard protects a path that never sees stale output in practice, while the async path that does see it bypasses the guard.

Fix: thread staleness to the base via an instance flag, since a parameter can't survive vLLM's super() call:

```python
# patched_async_update_request_with_output
scheduler_self._cd_stale_in_flight = is_stale
try:
    return original_async_update_req(scheduler_self, request,
                                     new_token_ids, is_stale=is_stale)
finally:
    scheduler_self._cd_stale_in_flight = False

# patched_update_base
stale = is_stale or getattr(scheduler_self, "_cd_stale_in_flight", False)
if diff > 0 and not stale:
    request.num_computed_tokens += diff
```

Verified this closes the hole (async stale -> num_computed_tokens unchanged) with sync-stale and async-fresh behavior unaffected.

tests/core/test_sched_utils.py
```diff
@@ -0,0 +1,93 @@
...
+                                                       is_stale=False)
+            assert request.num_output_placeholders == 0
+
+            # Stale output: placeholders were zeroed at preemption; neither the
+            # pre-compensation nor the original's decrement may touch them.
+            request = _make_request(num_output_placeholders=1)
+            AsyncScheduler._update_request_with_output(scheduler,
+                                                       request, [1, 2, 3],
+                                                       is_stale=True)
+            assert request.num_output_placeholders == 1
```
QiliangCui, line 130: **major — assert `num_computed_tokens` here too.**

This test executes exactly the chain where the base advance escapes the stale guard (async wrapper -> vLLM original -> super() -> patched base with default `is_stale=False`), but only asserts placeholders. Add `assert request.num_computed_tokens == 10` to this stale case (it currently ends at 12) and `assert request.num_computed_tokens == 12` to the fresh case above, so the guard's end-to-end behavior is locked in.

tests/core/test_sched_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    return request
+
+
+class TestContinueDecodeSchedulerPatch:
+
+    def test_patched_signatures_accept_is_stale(self):
+        """vLLM passes is_stale=... as a kwarg from update_from_output; the
+        patched wrappers must accept it or the EngineCore dies with a
+        TypeError."""
+        patch_vllm_scheduler_for_continue_decode()
```
QiliangCui, line 66: **minor — no teardown for a process-global patch.**

These tests apply the irreversible monkeypatch (including `Scheduler.__init__`, which forces `num_lookahead_tokens >= max_decode_steps - 1 = 9`), and CI part2 runs all of `tests/` in a single pytest process with `tests/core` collected first — every later test inherits the patched classes. Elsewhere the suite deliberately avoids this (`tests/platforms/test_tpu_platform.py` mocks the patch call). Suggest capturing the three originals and restoring them (plus deleting the `_continue_decode_patched` flags) in a module-scoped autouse fixture.

tpu_inference/core/sched/utils.py
```diff
@@ -36,15 +36,23 @@ def patch_vllm_scheduler_for_continue_decode():
     if not getattr(Scheduler, "_continue_decode_patched", False):
         original_update_base = Scheduler._update_request_with_output
 
-        def patched_update_base(scheduler_self, request, new_token_ids):
+        def patched_update_base(scheduler_self,
+                                request,
+                                new_token_ids,
+                                is_stale=False):
```
QiliangCui: **minor — harden against the next added kwarg.**

This nightly broke because vLLM added one kwarg to a monkeypatched signature. Adding `**kwargs` pass-through to both wrappers (keeping `is_stale` explicit) turns the next additive upstream kwarg into a no-op instead of an EngineDeadError.

QiliangCui (commented):
Self-review notes from verifying this against vLLM at the LKG (`a7a204cc6`):

- Confirmed both claimed upstream mechanisms: `update_from_output` passes `is_stale=` as a kwarg (`vllm/v1/core/sched/scheduler.py` L1802-1804), and a dict `fault_tolerance_config` force-enables fault tolerance (`vllm/engine/arg_utils.py` L763-769). The TypeError fix, the MM test fix, and the signature test are correct.
- One real gap: the async stale path bypasses the new `num_computed_tokens` guard — inline comment with a verified fix.
- Two smaller test notes inline.

tpu_inference/core/sched/utils.py
QiliangCui: Fixed in 14913749e via the suggested instance flag: the async wrapper sets `_cd_stale_in_flight` around the wrapped call (try/finally so it can't leak), and the patched base checks `is_stale or getattr(scheduler_self, "_cd_stale_in_flight", False)`. Verified the new end-to-end test reproduces the 10->12 advance on the unfixed code and passes with the fix.

tests/core/test_sched_utils.py
QiliangCui: Done in 14913749e — the async test now asserts `num_computed_tokens` (12 fresh / 10 stale) plus placeholders, and that the flag resets after the call. Confirmed the stale assert fails on the pre-fix code (catches exactly this bypass).

tests/core/test_sched_utils.py
QiliangCui: Done in 14913749e — autouse fixture captures the three originals before each test and restores them (and removes the `_continue_decode_patched` class flags) afterward.

tpu_inference/core/sched/utils.py
QiliangCui: Done in 14913749e — both wrappers take and forward `**kwargs` (with `is_stale` still explicit), and the signature test asserts VAR_KEYWORD is present so it can't regress.

PR #3314, opened 2026-08-01
Trim noisy blocks out of the CI job logs

QiliangCui:
trim logs


theminghuang:
Since we are here, can we also trim the logs on pulling docker image and cleanup docker images?

Lumosis (approved):

theminghuang (approved):

PR #3310, opened 2026-07-31
[DSv4] e2e integration of the new kernels

gxd3:
[DSv4] e2e integration of the new kernels

* Deleted the old kernels; 
* The previous saw MMLU 1.5% drop (https://screenshot-v2.corp.google.com/5cBbKb2Q3rPVtkE) is fixed, the fix is in sparse_mla.py.

E2E Verified with v4 Flash:
https://paste.googleplex.com/4818686866227200

PR #3316, opened 2026-08-01
[DSv4] Fix SMEM overflow in compress_norm_rope_store: prefetch per-token window pages, not the full block table

QiliangCui:
## Problem

Since #3310, `compress_norm_rope_store` prefetches the entire `[num_reqs, max_blocks]` state-cache block table into scoped SMEM as a scalar-prefetch operand. With the state cache's 2-token pages this scales as `num_reqs * max_model_len / 2`, and the nightly `tpu7x E2E MMLU for DeepSeek-V4-Flash torchax backend` config (`max_num_seqs=64`, `max_model_len=9216`) needs a 64x4608 int32 table = 1.125 MiB — over the 1 MiB scoped-SMEM budget. Engine init dies at compile:

```
jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: Allocation (size=1179648) would exceed memory (size=1048576)
:: [shape = 'u8[1179648]{0}', space=smem, scoped, tag = 'prefetched SMEM operand 0'] :: compress_norm_rope_store...
```

Nightly red since build [23643](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/23643); root cause proven by commit-vs-parent A/B on the dev pipeline (build [737](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/737) @ `8ead48e6e` FAIL vs [738](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/738) @ parent PASS, same vLLM pin). The job is nightly-gated so #3310's PR CI could not catch it.

## Fix

The kernel only ever reads `pages_to_buffer_per_token` consecutive block-table entries per token (`PageBufferRef.copy_in`). Gather exactly those page numbers outside the kernel — the same host-side gather `derive_metadata` already does for `slot_mapping_slots` — and prefetch the resulting `[num_tokens, pages_to_buffer_per_token]` array instead of the full table:

- `kernel.py` (`compress_norm_rope_store`): build `token_window_pages[t, p] = block_table[req[t], max(pos[t]//state_block_size - (P-1) + p, 0)]` and pass it as scalar-prefetch operand 0. Public signature unchanged (still takes `block_table`).
- `buffered_ref.py` (`PageBufferRef.copy_in`): read `token_window_pages_ref[idx_n, p]` directly; `page_buffer_args` shrinks to `(cache_ref, token_window_pages_ref)`.

The SMEM operand now scales with the scheduled-token count (~20 KB at the failing config) instead of `num_reqs * max_model_len`. The per-token lookups are bit-identical to what the kernel computed in-SMEM before, including the `max(idx, 0)` clamp for window prefixes.

## Verification

| Check | Result |
|---|---|
| `tests/kernels/deepseek_v4/compress_store_test.py` (pallas-vs-reference parity), local tpu7x-8 | **14/14 passed** |
| Dev build [749](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/749) = this fix: kernel tests + exact nightly test_37 MMLU config | **PASSED** — MMLU accuracy **0.8715** (>= 0.86), throughput >= 1500 PASSED |
| Dev build [750](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/750) = control (unfixed main + identical pipeline) | **FAILED** with the byte-identical RESOURCE_EXHAUSTED |

The 749-vs-750 pair is a controlled before/after on the same harness: only the two-file fix differs.

### DIFF
--- a/tests/kernels/deepseek_v4/compress_store_test.py
+++ b/tests/kernels/deepseek_v4/compress_store_test.py
@@ -465,6 +465,39 @@ class CompressStoreTest(jtu.JaxTestCase):
             self.skipTest("skip csa_decode_batch_large on TPUv6e")
         self.run_compress_store_correctness(**cfg)
 
+    def test_compress_store_production_scale_block_table(self):
+        """Regression test for b/541619368 (SMEM overflow).
+
+        A [64, 4608] int32 block table — the nightly DeepSeek-V4-Flash MMLU
+        engine shape (max_num_seqs=64, max_model_len=9216,
+        state_block_size=2) — is 1.125 MiB, over the 1 MiB scoped-SMEM
+        budget. A kernel that prefetches the full table (as before the
+        per-token window-page gather) fails to compile here with
+        RESOURCE_EXHAUSTED; the gathered form must stay bounded by the
+        token count. The case also pins the request axis of the gather:
+        all tokens map to request row 37 and every other row points at
+        page 0, so indexing the wrong row breaks output parity.
+        """
+        num_reqs, max_blocks, req_idx = 64, 4608, 37
+        num_tokens = 8
+        block_table = np.zeros((num_reqs, max_blocks), dtype=np.int32)
+        # Identity mapping on the real row; entries past the pages the test
+        # actually allocates are never dereferenced by valid positions.
+        block_table[req_idx, :] = np.arange(max_blocks, dtype=np.int32)
+        self.run_compress_store_correctness(
+            num_tokens=num_tokens,
+            head_dim=512,
+            rope_head_dim=64,
+            compress_ratio=4,
+            overlap=True,
+            physical_page_size=256,
+            hidden_size=4096,
+            token_to_req_indices=np.full((num_tokens, ),
+                                         req_idx,
+                                         dtype=np.int32),
+            block_table=block_table,
+        )
+
     def test_derive_aliases(self):
         # HCA: has_rope=True, has_rope_cache=False, num_scalar_prefetch=5
         self.assertEqual(
--- a/tpu_inference/kernels/experimental/deepseek_v4/compress_and_store/buffered_ref.py
+++ b/tpu_inference/kernels/experimental/deepseek_v4/compress_and_store/buffered_ref.py
@@ -115,7 +115,10 @@ class PageBufferRef(_BufferedRef):
         def loop_body(idx):
             n = idx // pages_to_buffer_per_token
             p = idx % pages_to_buffer_per_token
-            idx_n = global_idx + n
+            # Clamp for the final partial tile (same convention as
+            # gather_from_page_buffer); bounds checks are disabled, so an
+            # unclamped read would fetch a garbage page number.
+            idx_n = jnp.minimum(global_idx + n, self.cfgs.dims.size_n - 1)
 
             # Page numbers for each token's window are gathered outside the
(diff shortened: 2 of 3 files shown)


QiliangCui:
Addressed an in-depth review pass in `dbaef3a74`:

- **Regression test added** (`test_compress_store_production_scale_block_table`): builds the production-scale `[64, 4608]` block table with all tokens on request row 37 and garbage elsewhere. Verified both directions on tpu7x: pre-fix kernel fails it with the byte-identical `RESOURCE_EXHAUSTED (size=1179648 > 1048576, prefetched SMEM operand 0)`; fixed kernel passes (suite now 15/15). This moves the bug class from nightly-only to PR-CI-detectable, and pins the request axis of the relocated gather.
- **Dead prefetch operand removed**: `token_to_req_indices` is consumed by the host-side gather and nothing in-kernel reads it — dropped from `scalar_prefetch` / `kernel_fn` / `create_allocs_and_specs` (aliases key off `len(scalar_prefetch)`, so they adjust automatically).
- **Partial-tile clamp**: `PageBufferRef.copy_in` now clamps the token index like `gather_from_page_buffer` does (bounds checks are disabled, so the unclamped tail read fetched a garbage page number — pre-existing, discarded, but now clean).
- **SMEM ceiling documented**: the gathered operand re-hits the 1 MiB budget around ~13K scheduled tokens for CSA — noted at the gather site so a future `max_num_batched_tokens` bump doesn't rediscover b/541619368.

Re-verification after these changes: local kernel suite 15/15; dev build [757](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/757) re-running the kernel tests + exact nightly test_37 MMLU config on the updated branch.

gxd3:
> Addressed an in-depth review pass in `dbaef3a74`:
> 
> * **Regression test added** (`test_compress_store_production_scale_block_table`): builds the production-scale `[64, 4608]` block table with all tokens on request row 37 and garbage elsewhere. Verified both directions on tpu7x: pre-fix kernel fails it with the byte-identical `RESOURCE_EXHAUSTED (size=1179648 > 1048576, prefetched SMEM operand 0)`; fixed kernel passes (suite now 15/15). This moves the bug class from nightly-only to PR-CI-detectable, and pins the request axis of the relocated gather.
> * **Dead prefetch operand removed**: `token_to_req_indices` is consumed by the host-side gather and nothing in-kernel reads it — dropped from `scalar_prefetch` / `kernel_fn` / `create_allocs_and_specs` (aliases key off `len(scalar_prefetch)`, so they adjust automatically).
> * **Partial-tile clamp**: `PageBufferRef.copy_in` now clamps the token index like `gather_from_page_buffer` does (bounds checks are disabled, so the unclamped tail read fetched a garbage page number — pre-existing, discarded, but now clean).
> * **SMEM ceiling documented**: the gathered operand re-hits the 1 MiB budget around ~13K scheduled tokens for CSA — noted at the gather site so a future `max_num_batched_tokens` bump doesn't rediscover b/541619368.
> 
> Re-verification after these changes: local kernel suite 15/15; dev build [757](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/757) re-running the kernel tests + exact nightly test_37 MMLU config on the updated branch.

We have a TODO in https://screenshot-v2.corp.google.com/8hdmqyFUrRGbos3
let me give that approach a try a bit, it should fix this problem.

Maybe we could hold this PR a bit?

PR #3319, opened 2026-08-03
[DSv4] Overlay HCA compressor state onto CSA's NOPE cache

gxd3:
[DSv4] Overlay HCA compressor state onto CSA's NOPE cache (instead of HCA's kv cache)
fix the TODO in https://screenshot-v2.corp.google.com/7wzqnyBVfpsjHfp

With this change, HCA compressor state's block_size change from 2 -> 32; so that the block_table for HCA compressor state become 16x smaller, avoid the SMEM OOM in https://b.corp.google.com/issues/541619368.

Verified with MMLU & MMLU Pro for V4Flash:
https://paste.googleplex.com/6250520430903296
https://paste.googleplex.com/4908257251557376

also piggyback 2 additional small changes:
* let compressor_forward take 1D block_table instead of 2D, so that caller don't need to reshape (tpu-runner's attn-metadata contains 1D block table)
* small perf improvement in mla_swa.py + mla.py

### DIFF
--- a/tpu_inference/kernels/experimental/deepseek_v4/core_attention/mla_swa.py
+++ b/tpu_inference/kernels/experimental/deepseek_v4/core_attention/mla_swa.py
@@ -571,16 +571,6 @@ def _mla_sliding_window_ragged_paged_attention_kernel(
         # hi byte). `pltpu.bitcast` fuses consecutive sublanes little-endian, so the
         # reconstruction is a plain bitcast + reshape.
         bkv = pltpu.bitcast(bkv_u8, jnp.bfloat16).reshape(bkv_sz, head_dim)
-
-        # In vLLM, multiple caches may overlay on the same KV Tensor. For example,
-        # compressor state cache write data in bfloat16 / float32 format, certain
-        # byte pattern are interpreted as NaN in FP8, e.g. float8_e8m0fnu byte 0xFF
-        # decodes to NaN.
-        # We need to mask out the data by the actual kv_len to avoid NaN propagting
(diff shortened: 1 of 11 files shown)


a1yssan13 (approved):
thanks guangxiang!

tpu_inference/kernels/experimental/deepseek_v4/core_attention/mla_swa.py
```diff
@@ -571,16 +571,6 @@ def load_bkv(bkv_sem_idx, bkv_idx, start_offset):
         # hi byte). `pltpu.bitcast` fuses consecutive sublanes little-endian, so the
         # reconstruction is a plain bitcast + reshape.
         bkv = pltpu.bitcast(bkv_u8, jnp.bfloat16).reshape(bkv_sz, head_dim)
-
```
inucool12, line 574: Since SWA caches still overlay the CSA NoPE array (which is shared with CSA/HCA states), isn't  there any risk of removing NaN mask here ?

gxd3: The reason for no risk:
in fetch_bkv(), we only DMA the valid kv (must not NaN) into `bkv_x2_ref` (not DMA the whole page, but DMA the valid tokens' kv)

during init: we zero the buffer: https://screenshot-v2.corp.google.com/6YvRhHWwNbxEFa6 at the begining.

therefore, we never bring NaN into bkv_x2_ref ever; so it's safe.

PR #3250, opened 2026-07-23
Add experimental stacked RPA attention kernel

rupengliu-meta:
# Description
New Stacked rpa kernel that support up to 1M context length and use seq_alone_lane kv layout by default.
Upload an initial version as discussed, more optimization will be added later
initial results compared with batched rpa on decode performance. Batched rpa is still faster on short context, while stack is faster on long context. 
截屏2026-07-23 上午10 13 10

Start with a short description of what the PR does and how this is a change from
the past.

The rest of the description includes relevant details and context, examples:
- why is this change being made,
- the problem being solved and any relevant context,
- why this is a good solution,
- some information about the specific implementation,
- shortcomings of the solution and possible future improvements.

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: #123456

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3330, opened 2026-08-03
[stacked_rpa] Speed up decode across the context range

qizzzh:
> Depends on #3250 (the base stacked_rpa kernel). Until #3250 merges, this PR's diff also shows the commits from that branch; those will fall off automatically once #3250 lands on `main`.

Five related changes to the stacked_rpa decode path. They compose to speed
up decode across short, mid, and long context on Qwen3-Coder-480B TP=8.
Prefill and mixed paths are unchanged.

1. **Context-aware block-size heuristic.**
   Short context keeps the legacy `(batch=8, nbuf=3)` pick. Long context
   picks the `(batch, nbuf)` combination that maximises useful tokens
   streamed per step (`batch * min(bkv, max_model_len)`), and `bkv` is
   capped at `max_model_len` so we never pay MXU cost on a masked-zero
   tail.

2. **VMEM estimator no longer pads `num_kv_heads*2` to `packing_kv`.**
   That dimension is stored on a regular axis, not sublane/lane, so it
   was never actually padded in the allocation. The old estimator
   over-counted KV bytes for narrow-KV FP8 configs and starved the
   block-size heuristic. The bank-conflict bump that only applied to the
   over-padded stride was dropped for the same reason.

3. **FP8 KV tuned entries for Qwen3-Coder-480B TP=8 v7x-8.**
   Cover `max_model_len` 64k / 128k / 256k / 512k / 1M at
   `batched_tokens_bucket=64`. The 64k–256k picks use triple-buffered
   `bkv=65536` (fits the actual allocation even though the estimator's
   80% cap would forbid it); 512k and 1M use a wider tile with `nbuf=2`,
   which wins once enough outer steps amortise the buffer count.

4. **Replace `pltpu.einshape` with `tensor.reshape` in the flash matmuls.**
   `pltpu.einshape` is a hijax op, which prevents `pjit` from caching the
   compiled executable and forces a Python re-trace on every call. Every
   `einshape` here is a merge or split of adjacent leading axes, so plain
   `.reshape` produces the same layout at zero runtime cost. Same fix as
   #3156, applied here to `flash_attention.py` and
   `flash_attention_chunked.py`.

5. **Opt-in chunked-KV flash with 5-D VMEM.**
   New decode-only path behind `STACKED_RPA_CHUNKED_FLASH` (default off).
   The 5-D layout matches HBM's outer-page axis, which lets each step
   issue one DMA descriptor spanning all cache pages
   (`pltpu.make_async_copy` requires src/dst shapes to match, so the 4-D
   VMEM cannot coalesce). On the workloads measured here the descriptor
   saving does not offset the per-page loop overhead, so it stays off by
   default; kept in tree for workloads with heavily fragmented paged KV.

## Bench

Kernel-only microbench on TPU v7x-8 with FP8 KV, prior kernel vs this change:

| shape             | base_us | new_us | delta   |
|-------------------|--------:|-------:|--------:|
| decode_1k_c32     |   952.5 |  744.3 | -21.9%  |
| decode_1k_c64     |  1036.2 |  935.4 |  -9.7%  |
| decode_4k_c32     |   938.6 |  824.9 | -12.1%  |
| decode_4k_c64     |  1076.0 |  987.9 |  -8.2%  |
| decode_16k_c32    |   972.2 |  894.2 |  -8.0%  |
| decode_16k_c64    |  1122.3 | 1013.4 |  -9.7%  |
| decode_64k_c32    |  1322.7 |  905.0 | -31.6%  |
| decode_64k_c64    |  1675.9 | 1206.4 | -28.0%  |
| decode_128k_c32   |  1576.5 | 1096.2 | -30.5%  |
| decode_128k_c64   |  2331.6 | 1504.6 | -35.5%  |
| decode_256k_c8    |  1225.2 |  882.5 | -28.0%  |
| decode_256k_c16   |  1564.1 | 1142.7 | -26.9%  |
| decode_256k_c32   |  2256.6 | 1457.0 | -35.4%  |
| decode_256k_c64   |  3651.7 | 2142.7 | -41.3%  |
| decode_512k_c8    |  1520.8 | 1061.4 | -30.2%  |
| decode_512k_c16   |  2276.5 | 1427.2 | -37.3%  |
| decode_512k_c32   |  3573.0 | 2171.1 | -39.2%  |
| decode_512k_c64   |  9477.4 | 3679.9 | -61.2%  |
| decode_1m_c8      |  2179.7 | 1400.1 | -35.8%  |
| decode_1m_c16     |  4038.7 | 2125.2 | -47.4%  |
| decode_1m_c32     |  9128.2 | 3574.1 | -60.8%  |
| decode_1m_c64     | 23452.9 | 6486.6 | -72.3%  |

No shape regresses. BF16 shapes are within run-to-run noise; the tuned
entries added here are FP8-only, and the einshape port is neutral on
BF16 in these measurements.

## Test plan

- [x] `python3 -m pytest tests/kernels/stacked_rpa/` — 7/7 pass
- [ ] Reviewer runs the same test suite on their setup
- [ ] Reviewer confirms the FP8 tuned entries match their bench (optional; the
      heuristic-only path is validated by tests either way)

## Follow-ups

- The chunked flash path stays off by default and can be revisited if a
  workload appears where DMA descriptor count is the bottleneck (opt in with
  `STACKED_RPA_CHUNKED_FLASH=1`).
- Tuned entries cover TPU v7x only; other generations need their own sweeps.


kyuyeunk (approved):
approving as it is in experimental folder without potential impact for other workloads.

PR #3334, opened 2026-08-04
deepseek_v4 compressor: keep the boundary-batch scatter in bounds by construction

QiliangCui:
## What

`prepare_boundary_batch` scatters the combined boundary batch with `.at[dest].set(src, mode="drop")`, routing every invalid packed row to `dest == output_size` and relying on drop-mode to discard it. This change keeps every index in bounds by construction instead: scatter into a buffer with one extra sentinel row, clamp any out-of-range destination to that row, and slice it off. Output is element-exact identical to the old code (verified on randomized cases); the sentinel row costs one element.

## Why

On TPU, drop-mode scatter semantics for out-of-bounds indices are not honored at every row width. Concretely, on jax 0.10.2 we hit this in `ragged_conv1d`'s boundary fixup: `-1` sentinel indices under `mode="drop", wrap_negative_indices=False` are correctly dropped at row width 128 but silently **wrap** at width 1024, overwriting row 0 — an eager+jitted micro-repro exists, and it corrupted first-token conv outputs in a shipping model until fixed by removing the OOB indices entirely (commit 6e51acbe5). Whether this file's 1-D int scatter is affected on any current width/toolchain is unverified — this change makes the question moot rather than answering it per-toolchain.

## Tests

`tests/kernels/deepseek_v4/boundary_batch_scatter_test.py` (CPU-runnable, 7 cases): the all-sentinel canary at several sizes incl. 1024 (a wrapped index corrupts row 0 there), exactly-once landing of valid tokens with defaults everywhere else, and the decode packed-prefix layout. The pre-existing `compress_store_test.py` results are unchanged (same 12F/1P/1S on CPU at base and with this change; the failures are the pallas kernels needing a TPU).

### DIFF
--- a/tests/kernels/deepseek_v4/boundary_batch_scatter_test.py
+++ b/tests/kernels/deepseek_v4/boundary_batch_scatter_test.py
@@ -1,0 +1,147 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Scatter-bounds tests for prepare_boundary_batch.
+
+The combined-batch scatter routes every non-boundary / invalid token to a
+sentinel destination one past the end of the output. These tests pin that
+sentinel rows can never corrupt live output rows, at several output sizes:
+on TPU, drop-mode scatter semantics for out-of-bounds indices have been
+observed (jax 0.10.2) to hold at one row width and silently wrap at
+another, so the packing must keep every index in bounds by construction
+rather than rely on them being dropped.
+"""
+
+import jax.numpy as jnp
+import numpy as np
+from absl.testing import absltest, parameterized
+
+from tpu_inference.kernels.experimental.deepseek_v4.compress_and_store.compressor_v1 import \
+    prepare_boundary_batch
+
+
+def _run(positions, req_indices, kv_slot_mapping, is_boundary, is_decode,
+         compress_ratio, num_reqs):
+    return prepare_boundary_batch(
+        jnp.asarray(positions, dtype=jnp.int32),
+        jnp.asarray(req_indices, dtype=jnp.int32),
+        jnp.asarray(kv_slot_mapping, dtype=jnp.int32),
+        jnp.asarray(is_boundary, dtype=bool),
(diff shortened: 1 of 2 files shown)


tests/kernels/deepseek_v4/boundary_batch_scatter_test.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+
+class BoundaryBatchScatterTest(parameterized.TestCase):
+
+    @parameterized.parameters(64, 128, 1024, 1030)
+    def test_no_valid_tokens_leaves_outputs_at_defaults(self, num_tokens):
+        """All-sentinel scatter: every output row must keep its default.
+
+        This is the wrap canary: if the sentinel destinations are not
```
QiliangCui, line 52: Triage note: on CPU the old drop-mode code also passes this canary (CPU honors drop semantics for OOB indices), so a green CPU run does not exercise the TPU wrap bug — the canary only has teeth on the TPU kernels CI job. Consider one docstring sentence to that effect so a future triager doesn't mistake a CPU pass for coverage of the miscompile.

QiliangCui (commented):
Reviewed with independent verification:

- **Correctness**: `safe_dest` is in `[0, output_size]` by construction, the buffer is `output_size + 1` rows, and the static slice drops the sentinel row — the default `promise_in_bounds` scatter mode is genuinely satisfied, with no reliance on OOB-drop or negative-wrap semantics. Duplicate indices can only target the sentinel row (sliced off), so scatter write-order nondeterminism is harmless; valid destinations keep the pre-existing uniqueness invariant. The `dest >= 0` guard is strictly safer than the old code: the old scatter's default `wrap_negative_indices=True` would have wrapped a hypothetical negative destination into a live row, while this routes it to the sentinel.
- **Equivalence**: an independent 300-trial randomized old-vs-new comparison on CPU (mixed valid/sentinel destinations, defaults 0 and -1) produced 0 mismatches, matching the element-exact claim in the description.
- **Tests**: 7/7 new cases pass on CPU.

Two notes beyond this diff:

1. `tpu_inference/layers/common/ragged_conv1d_jax.py` (lines 162-166) still carries the exact pattern this PR's description documents as miscompiling: `-1` sentinel scatter indices with `mode="drop", wrap_negative_indices=False`, observed to wrap at row width 1024 on jax 0.10.2. It is currently unreachable on main (`ragged_conv1d` has no production caller yet), but the gather-based fix (6e51acbe5) should land with or before whatever wires it in.
2. The description's "until fixed by removing the OOB indices entirely (commit 6e51acbe5)" cites a commit that lives on a dev branch, not main — worth rewording so readers don't look for it in main history.

tests/kernels/deepseek_v4/boundary_batch_scatter_test.py
QiliangCui: Done in 330d029ac — the canary's docstring now states that CPU honors drop semantics, so a green CPU run checks only the packing contract; the miscompile coverage is the TPU kernels CI job.

PR #3336, opened 2026-08-04
Cache Mamba state allocators during KV cache reinit

AntonyMei:
# Description

Jax is re-jitting the mamba state allocation function every time, leading to slow weigh transfer in RL post-training in MaxText / Tunix. Similar to [PR 3180] (https://github.com/vllm-project/tpu-inference/pull/3180), we apply the same fix to mamba state allocation function.

# Checklist

### DIFF
--- a/tpu_inference/runner/kv_cache.py
+++ b/tpu_inference/runner/kv_cache.py
@@ -116,6 +116,27 @@ def _get_kv_cache_allocator(
     return _allocate
 
 
+@cache
+def _get_mamba_cache_allocator(
+        cache_shape: tuple, cache_dtype: jnp.dtype,
+        sharding: NamedSharding) -> Callable[[], jax.Array]:
+
+    @partial(jax.jit, out_shardings=sharding)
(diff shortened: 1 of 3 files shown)


tpu_inference/runner/kv_cache.py
```diff
@@ -116,6 +116,27 @@ def _allocate() -> jax.Array:
     return _allocate
 
 
+@cache
+def _get_mamba_cache_allocator(
```
wenxindongwork, line 120: curious why do we need a `_get_mamba_cache_allocator` wrapper? what if we simply make `_allocate` a standalone function?

AntonyMei: Previously in #3180, I tried to make _allocate a standalone function, instead of adding a cached wrapper. However, some unit tests will fail (e.g. Jax compilation is triggered when disabled, etc.) and the fix is to use a wrapper like this.

wenxindongwork (approved):

PR #3358, opened 2026-08-09
Advance vLLM LKG to 83ad767eed (latest) with hybrid prefix-caching and MLA compat fixes

QiliangCui:
## Why

Second step of re-advancing the LKG after #3357 (Aug 3). This moves the pin to vLLM HEAD `83ad767eed3be3ee7f2df63be693bfaca5c7c922` (Aug 9) together with fixes for the three known breaks in the Aug 3 -> HEAD window.

## Fixes

| Break | vLLM change | Fix here |
|---|---|---|
| DeepSeek MLA crashes at engine init: `'VllmMLAAttention' object has no attribute 'calculate_kv_scales'` | vllm-project/vllm#49389 removed the attribute the parent `__init__` used to set | default it to `False` when absent; `tests/layers/vllm/test_mla_attention.py` 7/7 against both old and new vLLM |
| `test_gemma4_mtp` weight download 404s (`RevisionNotFoundError`) | vllm-project/vllm#49990 resolves `model_config.revision` to the repo's commit sha at config time; the test redirects `.model` to the `-assistant` repo afterwards | clear the resolved revision on redirect |
| **Qwen3.5-397B-A17B-FP8 gsm8k accuracy 0.01** (integration builds 1909+) | vllm-project/vllm#50991 enables prefix caching by default for hybrid (mamba/linear-attention) models; the TPU hybrid path garbles cached-prefix reuse | TPU platform force-disables prefix caching for `is_hybrid` models (and resets the derived `mamba_cache_mode`/`mamba_block_size`), restoring the pre-#50991 default; parametrized unit tests added |

## Proof for the prefix-caching fix (dev pipeline, exact CI job config)

- Build 797: vLLM `7cab4368f2`, no fix — stock (APC on) **0.01/0.01** flex/strict; same commit with `enable_prefix_caching=False` **0.97/0.96**. Same-commit flag-level A/B isolates prefix caching as the trigger.
- Build 799: with this platform guard, the stock job (no flag) passes at **0.98/0.97**.

## Also validated locally (tpu7x-8, vLLM at 83ad767eed)

- `tests/runner/test_mm_encoder_jit_manager.py` 16/16 (no further encoder-protocol drift at HEAD)
- `tests/platforms/test_tpu_platform.py` 40/40
- Full `tests/` collection clean


tests/models/jax/test_gemma4_mtp.py
```diff
@@ -106,6 +106,11 @@ def test_model_loading(
         # For HF loader testing, we redirect the model to point to the real assistant draft checkpoint
         if load_format == "skip_layers_model_loader_for_test":
             vllm_config.model_config.model = "google/gemma-4-31B-it-assistant"
+            # Newer vLLM (vllm#49990) resolves model_config.revision to the
```
QiliangCui: remove this comments like this. We won't care which vllm change introduced this change in the future.

tpu_inference/layers/vllm/custom_ops/mla_attention.py
```diff
@@ -93,6 +93,12 @@ def __init__(
         self.attn_type = AttentionType.DECODER
         self.sliding_window = None
 
+        # vLLM removed the runtime KV-scale calculation (and with it the
+        # calculate_kv_scales attribute set by MLAAttention.__init__) in
+        # vllm#49389; our forward still branches on it for older vLLM.
```
QiliangCui: remove comment like this. we won't care which vllm change caused this change in the future.
put this comment in PR description is good enough as a record.

tpu_inference/platforms/tpu_platform.py
```diff
@@ -284,6 +284,34 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         cls._initialize_sharding_config(vllm_config)
 
         cache_config = vllm_config.cache_config
+        # vLLM (>= vllm#50991) enables prefix caching by default for hybrid
```
QiliangCui: remove comment like this. we won't care which vllm change caused this change in the future.
put this comment in PR description is good enough as a record.

QiliangCui:
Addressed: dropped the vLLM-history comments in all three places (dbcb56547); the code comments now state only the current constraint, with the change history kept in the PR description.

sixiang-google (approved):

yiw-wang (approved):

jrplatin (approved):

PR #3366, opened 2026-08-10
[DSv4] avoid params being copied in fwd pass

gxd3:
[DSv4] avoid params being copied in fwd pass

3 params were being copied in the forward pass before:
* cos_sin_cache for ROPE
* MOE tid_to_eid-hash-table for the bottom 3 layers of MOE
* compressor's projection weigthts.

For `cos_sin_cache` & `tid_to_eid-hash-table`, copy is due to the last dim of the array is not multiple of 128; XLA chose a column-major layout (to avoid padding, thus save space); in each forward pass, XLA's relayout (to row major layout) trigger a copy;
* to avoid copy, added environment variable to force a row-major layout for those 2 arrays at model loading time.

For compressor's projection weights;
* explicitly perform a weight matrix tranpose at the end of weight loading so that we don't need to call ".T" within each forward pass.

Quality validated:
MMLU: https://paste.googleplex.com/6592322820243456 
MMLU Pro: https://paste.googleplex.com/4992211245727744

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -75,6 +75,8 @@ if TYPE_CHECKING:
     LORA_MODULE_PATH: str = ""
     SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES: str = "auto"
     SLICE_ROPE_CACHE: bool = False
+    ROPE_CACHE_ROW_MAJOR: bool = True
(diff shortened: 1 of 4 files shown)


inucool12 (approved):
LGTM Clever performance optimization for minimal memory overhead increase.

jrplatin (approved):

QiliangCui (approved):

PR #3374, opened 2026-08-11
Fix continue_decode on the flax_nnx path and stop dropping compile options

### LINKED ISSUE #3373: [Bug]: continue_decode crashes on the flax_nnx path — `step_fn_no_options` is only defined for torchax
### Your current environment

<details>
<summary>Environment</summary>

```
vLLM Version : 0.23.1rc1.dev1474+g0ba2aa35a.tpu (git sha 0ba2aa35a)
tpu_inference: in-image build; contains deea9a4 (verified: model_loader.py:386 has
               compiler_options=get_step_fn_compiler_options())
jax / jaxlib : 0.10.2
libtpu       : 0.0.43
torch        : 2.10.0+cpu    torchax 0.0.13
flax         : 0.12.8        transformers 5.14.1    numpy 2.1.3
Python       : 3.12.13       Debian 12 (bookworm), glibc 2.36

Hardware     : TPU v7x (Ironwood), 4x4x4 = 64 chips, DP16 x TP4, via Pathways
Model        : Qwen3-0.6B (flax_nnx path, driven by MaxText post-training RL)
```

Note: `collect_env.py` was run in a CPU-only pod of the same image, so its TPU/CUDA fields are
empty. The actual runtime is the TPU v7x topology above.
</details>

────────────────────────────────────────────────────

### 🐛 Describe the bug

With `vllm_additional_config={"enable_continue_decode": true}` on a **flax_nnx** model, the fused
decode loop raises:

```
ValueError: `compiler_options` can only be passed to top-level `jax.jit`.
Got compiler_options={
    'xla_tpu_all_gather_collective_matmul_mode': 'post_spmd_conservative',
    'xla_tpu_reduce_scatter_collective_matmul_mode': 'post_spmd_conservative',
    'xla_tpu_use_minor_sharding_for_major_trivial_input': 'true',
    'xla_tpu_sparse_core_all_reduce_offload_min_size_in_bytes': '67108864',
    'xla_tpu_sparse_core_all_gather_offload_min_size_in_bytes': '67108864'
} specified on a nested jit with name: run_model
```

Failing frames:

```
tpu_runner.py:1801   _execute_continue_decode  -> continue_decode(
decode_loop.py:469   continue_decode           -> _decode_core(
decode_loop.py:325   _decode_core              -> jax.lax.while_loop(cond_fn, body_fn, init_carry)
decode_loop.py:299   body_fn                   -> _run_one_step(...)
decode_loop.py:183   _run_one_step             -> model_fn(
model_loader.py:471  wrapped_model_fn          -> jitted_model_fn(*args, **kwargs)
```

### Root cause

`tpu_runner.py:1811` and `compilation_manager.py:2082` request an options-free model fn:

```python
model_fn=getattr(self.model, "step_fn_no_options", self.model_fn)
```

**That attribute is set on the torchax path but not the flax_nnx path.**

`models/vllm/vllm_model_wrapper.py:507-519` builds both variants:

```python
step_fun_no_options   = step_fun_jit(step_fun_impl)
step_fun_with_options = step_fun_jit(step_fun_impl,
                                     compiler_options=get_step_fn_compiler_options())
...
self.step_fn_no_options = step_fun_no_options    # for the fused decode loop
return step_fun_with_options                     # normal path keeps the options
```

`models/common/model_loader.py:375-391` builds only the options-carrying `run_model`. So on
flax_nnx the `getattr` falls back to `model_fn`, which carries `compiler_options`, and JAX rejects
it when it is staged inside `jax.lax.while_loop`.

Introduced by deea9a4 / #3158 ("Apply same compile options on flax_nnx / torchax path"), which
applied the options to both paths but added the no-options twin only to torchax. It is not caught
in CI because `enable_continue_decode` defaults to `False`.

### Second, related issue: the fused path loses 2 of 5 compile options

`decode_loop.py:120-125` hoists the options onto the outer `_decode_core` jit as a hardcoded
**3-entry literal**, while `get_step_fn_compiler_options()` returns **5** — it also adds
`xla_tpu_sparse_core_all_{reduce,gather}_offload_min_size_in_bytes` whenever
`SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES` is non-zero. That env var **defaults to `"auto"`**
(`envs.py:76`), which resolves to VMEM size (64 MiB here) — see the 5 options in the error above.

So once the crash is fixed, the fused decode path still silently compiles without those two
options. The literal predates `compiler_options.py`, which deea9a4 created.

### Suggested fix

1. `models/common/model_loader.py` — mirror the torchax pattern: build a second jit of `run_model`
   without `compiler_options` and expose it as `step_fn_no_options`. Both consumers already
   `getattr` it, so nothing downstream changes.
2. `runner/decode_loop.py` — replace the hardcoded literal with a call to
   `get_step_fn_compiler_options()`. It must be lazy (e.g. an `lru_cache`d factory building the
   jit on first call) because the decorator runs at import and the `"auto"` path needs
   initialized devices.

### Verification

Patching only `model_loader.py:386` to `compiler_options=None` in the installed package lets the
same workload run — 4 global steps, no error, where it previously died during the first rollout.
That is a diagnostic only (it discards the options rather than sourcing them from `_decode_core`),
but it confirms the mechanism.

The error is raised at trace time, so it is not specific to multi-host or multi-chip setups and
should reproduce on a single chip with a flax_nnx model and `enable_continue_decode: true`.

────────────────────────────────────────────────────

gutianyu-google:
Fixes #3373.

## What

`continue_decode` calls the model step fn from inside a `jax.lax.while_loop` body, where JAX
forbids `compiler_options` on a nested jit. `runner/tpu_runner.py:1811` and
`runner/compilation_manager.py:2082` therefore ask for an options-free variant:

```python
model_fn=getattr(self.model, "step_fn_no_options", self.model_fn)
```

`models/vllm/vllm_model_wrapper.py` sets that attribute; `models/common/model_loader.py` did not.
So on the **flax_nnx** path the `getattr` fell back to the options-carrying `model_fn` and
`enable_continue_decode=True` failed with:

```
ValueError: `compiler_options` can only be passed to top-level `jax.jit`.
... specified on a nested jit with name: run_model
```

This builds an options-free twin of `run_model` and exposes it as `step_fn_no_options`, mirroring
the torchax path. The draft model's jit already carries no compiler options, so it is its own twin.

## Second, related fix

`runner/decode_loop.py` hoisted the options onto the loop jit as a hardcoded **3-entry literal**,
while `get_step_fn_compiler_options()` returns **5** — it also adds
`xla_tpu_sparse_core_all_{reduce,gather}_offload_min_size_in_bytes` whenever
`SC_ALLREDUCE_ALLGATHER_OFFLOAD_MIN_BYTES` is non-zero, and that env var **defaults to `"auto"`**.
So the fused decode program was compiled without those two options on every default TPU deployment.

The literal predates `compiler_options.py`; it was correct when written and went stale when that
function was introduced. This calls the function instead of duplicating its result, so both decode
paths always receive the same options and cannot drift again.

The jit is now built lazily through an `lru_cache`d factory rather than at import, because the
`"auto"` path calls `get_tpu_vmem_size_bytes()` and needs initialized devices.

Fixing only the first half would leave continue_decode working but silently dropping two live
compile options, so both changes are in one PR.


tpu_inference/models/common/model_loader.py
```diff
@@ -543,6 +568,11 @@ def compute_pooler_output(
 
     state_leaves = tuple(jax.tree_util.tree_leaves(state))
 
+    # `runner/tpu_runner.py` and `runner/compilation_manager.py` look this up
+    # with `getattr(self.model, "step_fn_no_options", <fallback>)`, where
```
lk-chen: both torchax/flax_nnx path should have this field, could you refactor those 2 files to use self.model.step_fn_no_options?

torchax added in https://github.com/vllm-project/tpu-inference/pull/2507

tpu_inference/models/common/model_loader.py
```diff
@@ -372,7 +373,13 @@ def get_flax_model(
     # costs ~17 ms/step on Gemma-4-31B decode at TP=2.
     _state_treedef = jax.tree_util.tree_structure(state)
 
-    @jax.jit(
+    def run_model_impl(state_leaves, *args):
+        state = jax.tree_util.tree_unflatten(_state_treedef, state_leaves)
+        model = nnx.merge(graphdef, state)
+        return model(*args)
+
+    run_model_jit = functools.partial(
```
lk-chen: nit - rename to `_wrap_with_jit`

tpu_inference/models/common/model_loader.py
gutianyu-google: Done

tpu_inference/models/common/model_loader.py
gutianyu-google: Done

lk-chen (approved):

sixiang-google (approved):

PR #3365, opened 2026-08-10
Replay recorded rollout traces in the agentic benchmark

wenxindongwork:
## Problem

This pull request enables the replay of recorded rollout traces in the agentic benchmark. 

example usage: 

```
python benchmark_agentic.py \
  --model Qwen/Qwen3.5-397B-A17B-FP8 \
  --trace-file /tmp/gbs1024_trace_file.jsonl \
  --global-prefix-len 6476 \
  --num-groups 16 \
  --concurrency 16
```

The changes include:

* Added support for a `--trace-file` argument to replay recorded rollout traces, allowing for deterministic benchmarking using pre-recorded prompt and output lengths.  
* Implemented the `load_trace` function to parse trace files and group trajectories for replay. 
* Added a `--global-prefix-len` argument and logic to generate a prompt prefix shared across all trajectories, mimicking real-world agent system prompts and tool definitions.  

## Trace file format

One JSON object per line, one line per trajectory:

```json
{"group": 0, "stream": 0, "prompt_len": 7570, "out_lens": [399, 65, 57], "obs_lens": [23, 942, 2126]}
```

| field | meaning |
| --- | --- |
| `group` | Problem index. Streams sharing a `group` share one initial prompt, which is what exercises prefix caching. |
| `stream` | Generation index within the group, `0..G-1`. |
| `prompt_len` | Initial prompt length in tokens. Taken from the first stream of each group. |
| `out_lens` | Assistant tokens generated on each turn, in order. Its length sets the turn count. |
| `obs_lens` | Environment observation tokens appended after each turn, in order. |

PR #3383, opened 2026-08-13
Replay tool-call idle time in the agentic benchmark

wenxindongwork:
## Problem

The trace replay added in #3365 reproduces the token structure of real SWE-agent rollouts (prompt sharing, turn counts, per-turn output/observation lengths) but launches every turn back-to-back. Real rollouts leave each stream idle between turns while the environment executes the tool call.

## Changes

* Traces may carry an optional per-gap `tool_times` list (seconds, aligned with `obs_lens`); replay sleeps that long between consecutive turns of a stream.
* Synthetic mode gains `--tool-time-min` / `--tool-time-max` (uniform, default 0 = previous behavior).
* The report prints a tool-idle summary line (total, p50/p99 per gap) when any idle time was injected.

## Example

```
python benchmark_agentic.py \
  --trace-file gbs1024_trace_file_tool_time.jsonl \
  --global-prefix-len 6476 \
  --num-groups 16 \
  --concurrency 16
```

### DIFF
--- a/scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
+++ b/scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py
@@ -6,6 +6,9 @@ This script simulates a continuous stream of GRPO training requests:
 - Each stream is a multi-turn conversation (10-100 turns).
 - Each turn generates 200-2k tokens, followed by a simulated environment
   response of 10-100 tokens.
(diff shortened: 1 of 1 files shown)


pv97 (approved):

andytwigg (approved):
@wenxindongwork what format is being used for the trajectories? Can we align it with the ATIF 1.7 format that is being used elsewhere?

PR #3040, opened 2026-06-29
[kernels][fused_moe] Add another fused EP MoE kernels

rupengliu-meta:
Summary:
Upload a new kernel by @rupengliu-meta @bangshengtang @zongweiz @yaochengji  . Bring the fused expert-parallel MoE TPU kernels into the repo as a self-contained, dependency-free package. The main kernel runs the entire EP MoE (gather, GMM1, activation, GMM2) with the ICI A2A fused into a single Pallas call, replacing the usual post-kernel all-reduce and keeping the collective overlapped with compute. A second, smaller kernel is included as a worked example of fusing the upstream all-gather into the grouped matmul (per-round push schedule) and is not performance-tuned.

The code is fully standalone: device-specific tuned block-size tables and all internal references were stripped for external release, the two kernels share a single trimmed grouped-matmul base, and a README documents usage and future optimization directions.

Perf gain:
Currently we see 15-55% perf gain for entire MoE layer compared with existing MoE implementation; Around 10%-30% e2e gain on different batch_sizes/prefill_decode mode with this kernel

# Description

Start with a short description of what the PR does and how this is a change from
the past.

The rest of the description includes relevant details and context, examples:
- why is this change being made,
- the problem being solved and any relevant context,
- why this is a good solution,
- some information about the specific implementation,
- shortcomings of the solution and possible future improvements.

If the change fixes a Github issue, please include a link, e.g.,:
FIXES: #123456

# Tests

Please describe how you tested this change, and include any instructions and/or
commands to reproduce.

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3388, opened 2026-08-13
Fused MoE Kernel integration and optimization

helloworld1:
# Description

Add integration, optimization to the fused MoE kernel based on https://github.com/vllm-project/tpu-inference/pull/3040

Performance is optimized for large m (> 1024). Tested on 10 run median of qwen coder 480b moe layer setup.

qwen_coder_1024
fused_moe_gmm 854 us
Original kernel 560.93 us
Optimized kernel 456.98 us
-18.53%

qwen_coder_2048
fused_moe_gmm: 1959 us
Original kernel 916.30 us
Optimized kernel 581.96 us
-36.49%

qwen_coder_4096
fused_moe_gmm:  6264 us
Original kernel  1694.44 us
Optimized kernel  1024.17 us
-39.56%

qwen_coder_8192
fused_moe_gmm 3513 us
Original kernel  3140.65 us
Optimized kernel  2001.39 us
-36.28%

*Ablation of changes*:
qwen_coder_2048

Original kernel 916.30 us
Gather optimization: 902 us
Scatter optimization: 878 us
Tile m size min 64 -> 128: 582 us

FIXES: #123456

# Tests

commands to reproduce.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -22,9 +22,8 @@ from jax.sharding import PartitionSpec as P
 
 import tpu_inference.envs as envs
 from tpu_inference.kernels.collectives.hierrs_sc import wrapper as hier_rs_sc
-from tpu_inference.kernels.experimental.fused_moe.fused_moe_rs import (
-    fused_moe_func_rs,
-)
+from tpu_inference.kernels.experimental.fused_moe.fused_moe_rs import \
+    fused_moe_func_rs
 from tpu_inference.kernels.megablox.gmm_v2 import gmm_v2
 from tpu_inference.kernels.sparse_core.dense_gather_reduce import \
     dense_gather_reduce
@@ -595,21 +594,20 @@ def fused_moe_func(
     """
 
     if use_gmm_fused_rs_kernel:
-        return fused_moe_func_rs(
-            hidden_states=hidden_states,
-            w1=w1,
-            w2=w2,
-            w1_scale=w1_scale,
-            w2_scale=w2_scale,
-            w1_bias=w1_bias,
-            w2_bias=w2_bias,
-            gating_output=gating_output,
-            topk=topk,
-            renormalize=renormalize,
-            mesh=mesh,
-            activation=activation,
-            scoring_fn=scoring_fn,
-            fp8_post_gather=all_gather_fp8)
+        return fused_moe_func_rs(hidden_states=hidden_states,
+                                 w1=w1,
+                                 w2=w2,
+                                 w1_scale=w1_scale,
+                                 w2_scale=w2_scale,
+                                 w1_bias=w1_bias,
+                                 w2_bias=w2_bias,
+                                 gating_output=gating_output,
+                                 topk=topk,
+                                 renormalize=renormalize,
+                                 mesh=mesh,
+                                 activation=activation,
+                                 scoring_fn=scoring_fn,
+                                 fp8_post_gather=all_gather_fp8)
 
     num_tokens, hidden_size = hidden_states.shape
     global_num_experts, padded_hidden_size, _ = w1.shape
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -593,7 +593,8 @@ def fused_moe_func(
         Output of moe operation [num_tokens, hidden_size]
     """
 
-    if use_gmm_fused_rs_kernel:
+    if use_ep and use_gmm_fused_rs_kernel:
+        logger.info("fused_moe_rs kernel in use")
         return fused_moe_func_rs(hidden_states=hidden_states,
                                  w1=w1,
                                  w2=w2,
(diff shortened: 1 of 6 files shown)


qizzzh (approved):

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -22,6 +22,8 @@
 
 import tpu_inference.envs as envs
 from tpu_inference.kernels.collectives.hierrs_sc import wrapper as hier_rs_sc
+from tpu_inference.kernels.experimental.fused_moe.fused_moe_rs import \
```
QiliangCui: This module-level import is the cause of the `tpu7x E2E speculative decoding test` failure (`test_eagle3_correctness[True-True]` OOM) that has failed in every CI build of this PR (24439/24476/24544/24560) while main stays green.

**Chain:**
1. `fused_moe_rs.py` reads `ShardingAxisName.EXPERT` / `ShardingAxisName.MLP_DATA` at module top (lines 35-36). That file landed via #3040 but nothing imported it until now — this import makes those lines run whenever `fused_moe_gmm` is imported.
2. `ShardingAxisName` is a `LazyShardingAxisName` singleton (`tpu_inference/layers/common/sharding.py`) that latches its backing class on **first attribute access**: `NEW_MODEL_DESIGN=1` → `ShardingAxisNameBase`, otherwise → `ShardingAxisName2D`. It never re-evaluates.
3. In the spec-decoding pytest job, model code gets imported while `NEW_MODEL_DESIGN` is still unset (the test helper only sets it right before constructing the LLM), so the singleton latches `ShardingAxisName2D` (`MLP_TENSOR='model'`). The forked EngineCore inherits the latched singleton.
4. In the `enable_dp_attention` case the mesh is `model=1, attn_dp=2`, so with the 2D axis names the MLP/vocab weights shard over a size-1 axis → fully replicated. Llama-3-8B loads at **15.75 GiB/chip instead of 9.36** (log line `Init model | hbm=...`), and KV-cache allocation then fails: `Attempting to allocate 4.33G ... 1.09G free`.

**Verification:** reproduced the exact failure locally by running this PR's CI image (`vllm-tpu:4d2c6af...-tpu7x`) — 15.75 GiB + the identical OOM. Mechanism A/B without a TPU:

```python
import os
os.environ["NEW_MODEL_DESIGN"] = "0"          # simulates import during earlier tests
import tpu_inference.layers.common.fused_moe_gmm
os.environ["NEW_MODEL_DESIGN"] = "1"          # what the dp_attention test sets later
from tpu_inference.layers.common.sharding import ShardingAxisName
print(ShardingAxisName.MLP_TENSOR)
# this PR:  'model'  (wrong — replicates MLP under dp_attention)
# main:     ('attn_dp', 'attn_dp_expert', 'expert', 'model', 'dcp', 'pcp')
```

**Fix options:**
- Move this import inside `fused_moe_func` (deferred import), or
- Remove the module-level `EXPERT`/`MLP_DATA` constants in `fused_moe_rs.py` and read `ShardingAxisName.*` inside functions instead.

Separately, `LazyShardingAxisName`'s latch-once behavior (plus its silent `except Exception → 2D` fallback) is a footgun — any future import-time access reintroduces this bug class; might be worth a follow-up hardening (e.g. assert-not-yet-latched or re-evaluate per access).

QiliangCui: Verified the deferred-import fix end-to-end on a tpu7x host, same harness before/after (import model code with `NEW_MODEL_DESIGN=0`, then set `=1`, then build the `enable_dp_attention` eagle3 engine — mirrors the pytest job):

| | `ShardingAxisName.MLP_TENSOR` after env=1 | `Init model \| hbm` | KV cache init |
|---|---|---|---|
| PR head as-is | `'model'` (2D) | **15.75 GiB/chip** | OOM (`4.33G ... 1.09G free` in CI) |
| PR head + deferred import | `('attn_dp', 'attn_dp_expert', 'expert', 'model', 'dcp', 'pcp')` | **9.36 GiB/chip** | allocates 19314 blocks, engine init completes |

The fix applied was exactly:

```python
# remove the module-level import, and inside fused_moe_func:
    if use_ep and use_gmm_fused_rs_kernel:
        from tpu_inference.kernels.experimental.fused_moe.fused_moe_rs import \
            fused_moe_func_rs
```

9.36 GiB and 19314 KV blocks match what main's green runs of this job report, so this restores parity.

QiliangCui: Fix also verified on the dev pipeline against this PR's exact HEAD (`4d2c6af` + only the deferred-import change, commit [`d94d699`](https://github.com/vllm-project/tpu-inference/commit/d94d6998ea09e75ff6de6aed5b26518b5ee2fbec) on branch `cuiq-pr3388-latch-fix`), running the same tpu7x E2E speculative decoding job as CI: [tpu-inference-dev build 912](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/912) — **passed**, 11/11 tests green including `test_eagle3_correctness[True-True]`, which now loads at 9.36 GiB/chip (matching main) instead of 15.75.

Feel free to cherry-pick the `fused_moe_gmm.py` hunk from that commit into this PR (ignore its `pipeline_dev.yml` change — that's just the test scaffolding).

helloworld1: Thanks a lot @QiliangCui for identifying this very hidden issue in this PR. Fixing it now and re-running the CI

jrplatin (approved):

QiliangCui (approved):

sixiang-google (approved):

PR #3389, opened 2026-08-13
Honour --group-size when replaying a trace, and read traces from GCS

wenxindongwork:
Two small usability fixes to the agentic benchmark, both hit while sweeping batch size against the MLPerf traces.

## `--group-size` was ignored when replaying a trace

`run_group` took its stream count from `len(specs)`, so `--group-size` did nothing whenever `--trace-file` was supplied. A run intending 8 streams silently replayed the trace's full 16, and the only visible sign was the stream count buried in the report — easy to miss, and it quietly invalidates a batch-size sweep.

Each group is now trimmed to `--group-size` at load time, so the startup summary and the run agree. 

## `--trace-file` now accepts `gs://`

```bash
python benchmark_agentic.py \
  --trace-file gs://... \
  --global-prefix-len 6476 --num-groups 1 --group-size 8 --concurrency 1
```


sierraisland (approved):
LGTM

PR #3391, opened 2026-08-13
[DSv4] add safe guard to avoid cache being mistakenly copied

gxd3:
[DSv4] Add safe guard to avoid cache being mistakenly copied due to XLA compiler HLO transformation (more details in the comments)

Observed cache being copied in 8 layers microbenchmark:
before this fix:
http://xprof.corp.google.com/trace_viewer/gxd-9010194911334918722
https://screenshot-v2.corp.google.com/66aittuuf2bhg

After the fix:
http://xprof/trace_viewer.html?session_id=gxd-900400297238143778 

This issue seem not triggered in real e2e run; didn't see perf diff e2e, 
but still add a safeguard to avoid such issue.


inucool12 (approved):
LGTM

PR #3395, opened 2026-08-14
Forward RPA v3 block-size overrides on the MaxText serving path

gutianyu-google:
## Summary

`RPA_V3_{DECODE,PREFILL,MIXED}_BLOCK_SIZES` (added in #3102) were wired into only one of the two entry points to the RPA v3 kernel:

| entry point | forwards the override? | used by |
|---|---|---|
| `layers/jax/attention/attention.py` — `_rpa_block_size_kwargs()` | yes | tpu-inference's own JAX models |
| `layers/common/attention_interface.py` — `sharded_ragged_paged_attention` | **no** | **MaxText**, via `Attention.forward_serve_vllm` |

On the MaxText path `d_block_sizes` therefore reached the kernel as `None`, `_prepare_block_sizes` fell back to `get_default_block_sizes()`, and for tpu7 `DECODE` that sets **both** `bkv_sz` and `bkv_csz` to `min(min_bkv_sz_to_peak, max_kv)` — so setting the env var had no effect whatsoever.

This is consistent with the kernel test's own note, which says the caller-side plumbing "is covered in the attention-layer tests" — true for `attention.py`, but there was no equivalent for `attention_interface.py`.

## Change

Forward the three tuples as kwargs when non-empty, mirroring `attention.py`. Guarded by the existing `not use_hd64` check, because the hd64 kernel derives its own sizes via `get_tuned_block_sizes` and does not accept these arguments. An unset var is an empty list and is omitted, so the default call is unchanged.

## Evidence

TPU v7x 4x4x4 (64 chips), qwen3-0.6b GRPO rollout, `RPA_V3_DECODE_BLOCK_SIZES=1,16384,1,4096`. Kernel actually executed, read from the decode-phase xplane:

```
before   RPAd-p_64-bq_1_1-bkv_16384_16384    138.4 ms   77.0% of XLA-Ops time
after    RPAd-p_64-bq_1_1-bkv_16384_4096     117.5 ms   56.6%
```

The compute block is 4096 only after this change; before it was silently 16384.

End to end over 19 warm steps:

| | rollout | global step | train (control) |
|---|---|---|---|
| before | 184.3 s | 267.9 s | 48.3 s |
| after | 134.7 s | 218.6 s | 48.4 s |
| delta | −49.6 s | **−49.3 s (−18.4%)** | 0.0 |

The entire saving lands in the rollout phase and the training phase is unchanged to 0.1 s, which is what a decode-kernel change should look like.

Net effect: this makes #3102's measured +49% decode throughput actually reachable for MaxText serving, which currently gets none of it.

## Testing

Two device-free regression tests in `tests/layers/common/test_attention_interface.py`, both stubbing the kernel and asserting on the kwargs it receives:

- `test_rpa_block_sizes_absent_when_env_unset` — unset env leaves the stock call untouched
- `test_rpa_block_sizes_forwarded_from_env` — decode/mixed forwarded, empty prefill omitted

They use `monkeypatch.setenv` rather than patching the `envs` module, since `envs` resolves these lazily via a module-level `__getattr__`; this also exercises the real `env_int_list` parsing.

Opened as a draft: I do not have a TPU/vLLM environment to execute the tests locally, so CI is their first run.

**Rebase note (2026-09-04):** rebased onto main after the batched-RPA-kernel change landed. The override is forwarded only on the default v3 kernel path; with `USE_BATCHED_RPA_KERNEL=1` the batched kernel takes its own `BlockSizes` configs and the v3 4-tuples are not passed to it (covered by `test_rpa_block_sizes_not_forwarded_to_batched_kernel`).

### DIFF
--- a/tpu_inference/layers/common/attention_interface.py
+++ b/tpu_inference/layers/common/attention_interface.py
@@ -403,6 +403,25 @@ def sharded_splash_attention(
         ))
 
 
+def _rpa_block_size_kwargs() -> dict[str, tuple[int, int, int, int]]:
+    """Optional RPA v3 block-size overrides from env, for the call site to
+    forward with ``**``.
+
+    The kernel is self-contained and never reads env, so the caller must supply
+    these; layers/jax/attention/attention.py does the same. Without it the
+    RPA_V3_*_BLOCK_SIZES vars are silently ignored on this path and the kernel
+    falls back to get_default_block_sizes(). Each var is a comma-separated
+    4-tuple ``(bq_sz, bkv_sz, bq_csz, bkv_csz)``; a key is included only when
+    its var is non-empty, so the default call is unchanged.
+    """
+    env_to_kwarg = {
+        "d_block_sizes": envs.RPA_V3_DECODE_BLOCK_SIZES,
+        "p_block_sizes": envs.RPA_V3_PREFILL_BLOCK_SIZES,
+        "m_block_sizes": envs.RPA_V3_MIXED_BLOCK_SIZES,
+    }
+    return {k: tuple(v) for k, v in env_to_kwarg.items() if v}
+
+
 def sharded_ragged_paged_attention(
     mesh: Mesh,
     q: jax.Array,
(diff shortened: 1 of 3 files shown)


gxd3 (approved):

tpu_inference/layers/common/attention_interface.py
```diff
@@ -403,6 +403,25 @@ def sharded_splash_attention(
         ))
 
 
+def _rpa_block_size_kwargs() -> dict[str, tuple[int, int, int, int]]:
```
lk-chen: shall we remove https://github.com/vllm-project/tpu-inference/blob/67d7e0b35340f2e6112885c7b0960a7ecaff4bbf/tpu_inference/layers/jax/attention/attention.py#L38 and reuse this ?

gutianyu-google: Done, thanks! Moved it to `attention_interface.py` as a public `rpa_block_size_kwargs` and `attention.py` now imports it.

lk-chen (approved):

PR #3401, opened 2026-08-16
[continue_decode] Make EOS_CHECK_INTERVAL<=0 a correct continuous-batching window

gutianyu-google:
## Summary

`CONTINUE_DECODE_EOS_CHECK_INTERVAL <= 0` currently just disables the EOS early-exit inside the fused decode loop. That configuration is broken today: sequences stop terminating and run to their token limit. This PR turns `interval <= 0` into a correct **continuous-batching decode window** — the fused loop runs to its static depth regardless of EOS, finished slots stay masked and inert, and the host reseeds every sequence from its own newest token.

On a decode-heavy RL post-training workload this removes the dominant host-side overhead of continue-decode (the window currently early-exits after only a handful of iterations once EOS events become frequent, paying the host gap on almost every decode step): **rollout 130 s → 98 s per training step (−25%)** with output quality identical to the default path.

## Why `interval <= 0` is broken today

Three independent issues, each invisible until the window runs past an EOS:

1. **Reseed reads the window's final row.** `tpu_runner._execute_continue_decode` derives every sequence's newest token from `generated_tokens[step_counter - 1]`. That is only correct because the default window stops immediately after any EOS. Once the window keeps running, a sequence that finished mid-window has **padding** in that row, so it is reseeded from a padding token, continues generating from garbage, and never terminates.
2. **Masked slots overwrite their own KV with padding-KV.** `_update_loop_state` re-feeds finished slots `padding_token_id` at a frozen position, so every masked iteration recomputes and rewrites that slot's KV from a padding embedding. With the default every-step check no masked iteration ever executes, so this path has never run in production.
3. **The window never exits when all sequences are done** (`return not_done` ignores `active_mask`).

## Changes

- `decode_loop.cond_fn` (`interval <= 0` branch): `return i < max_decode_steps AND any(active_mask)`.
- `decode_loop._run_one_step`: masked slots re-feed their **current input token** instead of the padding token — same token, same frozen position, same context, so the KV rewrite is byte-identical to what is stored and the slot is provably inert.
- `tpu_runner`: reseed from each sequence's **last non-padding row** within the executed steps. Whenever the window early-exits this reduces exactly to the final row, so the default path is unchanged.
- Warn if this mode is enabled together with `async_scheduling` (see limitation below).

**Default behavior (`interval >= 1`) is byte-identical**: with the every-step check the window exits before any masked iteration executes, and the last-non-padding-row reseed reduces to the final row.

## Known limitation: requires `async_scheduling=False` for now

With async scheduling enabled, deep mixed (active+masked) windows degrade generation quality: the EOS-termination rate roughly halves and 30%+ of sequences run to their token limit (measured at 54% truncated-at-cap vs 24% baseline on the workload below, reproducible at inference step 0 with byte-identical prompts). Extensive instrumentation shows every host-visible mechanism is correct in that regime — per-request retirement, extraction, `num_computed_tokens`, per-DP-rank layout, and page bookkeeping all audit clean, and there are no vLLM preemptions — so the failure is an interaction between the async machinery and deep windows that has not been pinpointed yet. The same window under `async_scheduling=False` produces generation statistically identical to the default path.

Since a deep window cuts the number of scheduler/runner round-trips per decode chain by ~20x (e.g. 64 windows instead of ~1400 for an 8192-token chain at depth 128), sync scheduling costs very little in this mode: on the workload below, everything except ~15 s of prefill serialization per step is recovered. The warning added in this PR points users at the constraint. I'm happy to file the async interaction as a separate issue with the full bisection data.

## Measurements

Qwen3-0.6B GRPO post-training (MaxText/Tunix), 2048 sequences/step, generation cap 8192, `max_decode_steps=128`, TPU v7x 4x4x4, DP16×TP8 rollout sharding, `async_scheduling=False`, 11 training steps:

| | default early-exit | this PR (`interval=0`) |
|---|---|---|
| rollout time / step | 130 s | **98 s** |
| training-step time | 190 s | **157 s** |
| reward mean | ~0.146 | 0.137–0.155 |
| truncated-at-cap | 22–26% | 22–27% |
| mean completion length | ~3750 | 3600–4020 |

Termination-health check at inference step 0 (identical prompts, identical base checkpoint, so directly comparable across configs): 22.7% truncated-at-cap vs 24.0% baseline; per-sequence EOS-termination counts match the baseline across the full decode chain.

## Test notes

- Unit tests (`tests/runner/test_decode_loop.py`): `test_continue_decode_no_exit_on_eos` now expects the window to exit once every sequence has finished (`step_counter == 3`, not 5); new `test_continue_decode_no_exit_on_eos_while_any_active` covers run-to-depth with a masked slot (padding in the buffer, frozen position/seq_len, last input token re-fed); `test_continue_decode_early_exit` expects a finished slot to carry its last input token instead of padding (the carry is the only observable difference on the default path — the window exits before any masked iteration runs, so KV and outputs are unchanged). All 8 decode-loop tests pass on CPU.
- Verified the default path is unchanged by construction (masked iterations cannot execute at `interval=1`; reseed reduction argument above) and by an 11-step baseline run.
- Correctness of the continuous window validated end-to-end on the RL workload above: reward, reward-component, and output-length distributions match the default path step-for-step.
- Edge case: a legitimately sampled token equal to `pad_token_id` in the window's final row would make the reseed pick an earlier row for that sequence. For models where `pad_token_id` is a plausible generation (id 0 here), this is rare; models where pad is also an EOS are unaffected (the sequence is finished either way). Flagging for reviewer judgment — an explicit per-row step counter carried through the loop would eliminate it at the cost of one more carried array.

### DIFF
--- a/tests/runner/test_decode_loop.py
+++ b/tests/runner/test_decode_loop.py
@@ -217,9 +217,13 @@ def test_continue_decode_early_exit():
     )
     assert np.array_equal(token_buffer, expected_tokens)
 
-    # Verify final state (state at the start of step 2, which did not run)
+    # Verify final state (state at the start of step 2, which did not run).
+    # A finished slot keeps its last input token (43) rather than padding:
+    # masked slots re-feed that token at their frozen position so any masked
+    # iteration rewrites the same KV it already holds. With the every-step EOS
+    # check no masked iteration runs, so this only shows up in the carry.
     assert np.array_equal(final_state.active_mask, [True, False])
-    assert np.array_equal(final_state.current_tokens, [44, -1])
+    assert np.array_equal(final_state.current_tokens, [44, 43])
     assert np.array_equal(final_state.attn_metadata.input_positions, [2, 1])
     assert np.array_equal(final_state.attn_metadata.seq_lens, [3, 2])
     assert all_expert_indices is None
@@ -385,15 +389,17 @@ def test_continue_decode_no_exit_on_eos():
         continue_decode_eos_check_interval=-1,
     )
 
-    # Verify loop ran all 5 steps despite EOS hit
-    assert int(final_state.step_counter) == 5
+    # With interval <= 0 an EOS never breaks the window: req 1 finishing at
+    # step 1 does not stop the loop. The loop only exits once every sequence
+    # has finished (after step 2) or at the static depth, whichever is first.
+    assert int(final_state.step_counter) == 3
 
     # Expected tokens:
     # Step 0: [42, 43]
     # Step 1: [44, 99] (req 1 hits EOS)
     # Step 2: [99, -1] (req 0 hits EOS; req 1 inactive -> -1)
-    # Step 3: [-1, -1] (both inactive)
-    # Step 4: [-1, -1] (both inactive)
+    # Step 3: [-1, -1] (not executed: all sequences finished)
+    # Step 4: [-1, -1] (not executed)
     expected_tokens = np.array(
         [
             [42, 43],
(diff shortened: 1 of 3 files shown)


gutianyu-google:
Rebased on current main (clean), fixed DCO sign-off and the yapf pre-commit failure, and updated the decode-loop unit tests for the new `interval <= 0` semantics (8/8 pass). Marking ready for review.

@pv97 since you added the EOS-check interval (#3068/#3174), your take on the semantics here would be much appreciated. Could a maintainer also apply the `ready` label so CI runs? Thanks.

gutianyu-google:
Closing this PR. Re-checked against current main:

- The fixed-depth window is already there: with `CONTINUE_DECODE_EOS_CHECK_INTERVAL <= 0`, `cond_fn` returns `i < max_decode_steps` only. The `AND any(active_mask)` exit this PR added would re-introduce a cross-DP reduction on every iteration, which measured as a net loss on the workload above.
- The inert-masked-slot change was tested in isolation and did not change generation; it is not needed.
- The reseed-from-last-real-row fix only matters on the async-scheduling branch (the sync branch trims each row at its first EOS on the host), and async + deep windows is not supported for the separate reason described here.

The one remaining piece worth landing from this line of work, dropping the now-dead `any_hit_eos` reduction when the check is off, is #3509.

PR #3402, opened 2026-08-16
Disable mm_device_do_normalize for JAX-native multimodal models

QiliangCui:
## What this PR contains

Two things, which is why the diff spans six files:

1. **Bug fix** — the JAX-native multimodal path silently ran the ViT on unnormalized pixels after the #3358 vLLM LKG advance; fixed in `tpu_platform.py` (+ platform test).
2. **Code refactor (review follow-up)** — extract the repeated env-read → `"auto"`-resolve pattern into a shared `resolve_model_impl_type` helper in `model_loader.py` and collapse the four open-coded copies onto it (`get_model`, `eagle3.py`, `dflash.py`, and this PR's new `tpu_platform.py` site), plus direct unit tests for the helper. No behavior change outside the bug fix.

## Problem

The nightly multimodal correctness test (Qwen2.5-VL-3B, `tests/e2e/test_multi_modal_inference.py`) has failed deterministically on both v6e and v7x since the vLLM LKG advance in #3358, on the default and `MODEL_IMPL_TYPE=flax_nnx` nightlies (e.g. builds 24350/24355): the model produces a fluent description of a *different* image (similarity 0.24 vs the 0.85 threshold). The `MODEL_IMPL_TYPE=vllm` variant passes.

## Root cause

vLLM commit `2fa490470d` ("[Model] Fused mm preprocess normalisation on the Device", vllm-project/vllm#50411), included in the #3358 LKG range, makes models that declare `supports_mm_device_do_normalize` (Qwen2.5-VL among them) skip `do_rescale`/`do_normalize` in the CPU processor; the vLLM model then normalizes on-device via `FusedInputNorm` in its vision tower.

JAX-native multimodal models consume the processor's `pixel_values` directly and have no device-side norm, so the ViT silently ran on unnormalized pixels.

## Fix (part 1: the bug)

Force `mm_device_do_normalize=False` in `TpuPlatform.check_and_update_config` when the model resolves to the JAX-native path, restoring CPU-side normalization. The vllm (torchax) path keeps the optimization (its model owns the `FusedInputNorm`).

## Refactor (part 2: why eagle3/dflash/model_loader are touched)

Per review, the impl-resolution dance (`envs.MODEL_IMPL_TYPE` / `DRAFT_MODEL_IMPL_TYPE`, then resolve `"auto"` from the architecture) existed in four places. It now lives in one helper:

- `model_loader.py`: new `resolve_model_impl_type(vllm_config, is_draft_model=False)` next to `resolve_model_architecture`; `get_model` uses it (keeps its "Loading model with MODEL_IMPL_TYPE=..." log).
- `spec_decode/jax/eagle3.py`, `spec_decode/jax/dflash.py`: their draft/target comparison blocks become two helper calls; unused imports dropped.
- `tpu_inference/platforms/tpu_platform.py`: the new mm_device_do_normalize site calls the helper (keeping the lazy import and the correctness-safe flax_nnx fallback).
- `tests/models/common/test_model_loader.py`: direct unit tests for the helper (explicit passthrough for target/draft env vars without architecture lookup; `"auto"` delegation with the draft flag).

## Verification

Commit-level A/B on the dev pipeline (`tests/e2e/test_multi_modal_inference.py` on tpu7x):
- [build 840](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/840) @ `a5596b27f` (before #3358): **pass**
- [build 841](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/841) @ `56c1eb0fb` (#3358): **fail** (regression enters here)
- [build 842](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/842) @ `e25c90b11` (current failure signature, bit-identical similarity scores to nightly): **fail**
- [build 845](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/845) @ main + this fix: **pass**

Unit tests: new parametrized `test_check_and_update_config_mm_device_do_normalize` covers auto→flax_nnx, flax_nnx, and vllm impl types (43/43 platform tests pass).

### DIFF
--- a/tpu_inference/platforms/tpu_platform.py
+++ b/tpu_inference/platforms/tpu_platform.py
@@ -318,19 +318,16 @@ class TpuPlatform(Platform):
             if vllm_config.model_config else None
         if mm_cfg is not None and getattr(mm_cfg, "mm_device_do_normalize",
                                           False):
-            impl = envs.MODEL_IMPL_TYPE
-            if impl == "auto":
-                from tpu_inference.models.common.model_loader import \
-                    resolve_model_architecture
-                try:
-                    impl = resolve_model_architecture(vllm_config,
-                                                      is_draft_model=False)
-                except Exception as e:
-                    logger.warning(
-                        "[tpu_platform] resolve_model_architecture failed "
-                        "(%s); assuming the JAX-native path for the "
-                        "mm_device_do_normalize check.", e)
-                    impl = "flax_nnx"
+            from tpu_inference.models.common.model_loader import \
+                resolve_model_impl_type
+            try:
+                impl = resolve_model_impl_type(vllm_config)
+            except Exception as e:
+                logger.warning(
+                    "[tpu_platform] resolve_model_impl_type failed (%s); "
+                    "assuming the JAX-native path for the "
+                    "mm_device_do_normalize check.", e)
+                impl = "flax_nnx"
             if impl != "vllm":
                 logger.warning(
                     "[tpu_platform] Disabling mm_device_do_normalize: the "
--- a/tpu_inference/platforms/tpu_platform.py
+++ b/tpu_inference/platforms/tpu_platform.py
@@ -320,14 +320,7 @@ class TpuPlatform(Platform):
                                           False):
             from tpu_inference.models.common.model_loader import \
                 resolve_model_impl_type
-            try:
-                impl = resolve_model_impl_type(vllm_config)
-            except Exception as e:
-                logger.warning(
-                    "[tpu_platform] resolve_model_impl_type failed (%s); "
-                    "assuming the JAX-native path for the "
-                    "mm_device_do_normalize check.", e)
-                impl = "flax_nnx"
+            impl = resolve_model_impl_type(vllm_config)
             if impl != "vllm":
                 logger.warning(
                     "[tpu_platform] Disabling mm_device_do_normalize: the "
--- a/tpu_inference/platforms/tpu_platform.py
+++ b/tpu_inference/platforms/tpu_platform.py
@@ -308,6 +308,37 @@ class TpuPlatform(Platform):
                 cache_config.mamba_block_size = (
                     vllm_config.model_config.max_model_len)
 
+        # vLLM's mm_device_do_normalize skips do_rescale/do_normalize in the
+        # CPU processor and instead normalizes inside the vLLM model's vision
+        # tower (FusedInputNorm). JAX-native multimodal models consume the
+        # processor's pixel_values directly and have no device-side norm, so
+        # they would silently run the ViT on unnormalized pixels. Keep the
+        # normalization in the CPU processor for the JAX-native path.
+        mm_cfg = getattr(vllm_config.model_config, "multimodal_config", None) \
+            if vllm_config.model_config else None
+        if mm_cfg is not None and getattr(mm_cfg, "mm_device_do_normalize",
+                                          False):
+            impl = envs.MODEL_IMPL_TYPE
+            if impl == "auto":
+                from tpu_inference.models.common.model_loader import \
+                    resolve_model_architecture
+                try:
+                    impl = resolve_model_architecture(vllm_config,
+                                                      is_draft_model=False)
+                except Exception as e:
+                    logger.warning(
+                        "[tpu_platform] resolve_model_architecture failed "
+                        "(%s); assuming the JAX-native path for the "
+                        "mm_device_do_normalize check.", e)
+                    impl = "flax_nnx"
+            if impl != "vllm":
+                logger.warning(
+                    "[tpu_platform] Disabling mm_device_do_normalize: the "
+                    "JAX-native multimodal path normalizes images in the CPU "
+                    "processor; device-side FusedInputNorm only exists in the "
+                    "vLLM model implementation.")
+                mm_cfg.mm_device_do_normalize = False
+
         # For v0, the default block size is 16.
         if cache_config and not cache_config.user_specified_block_size:
             if vllm_config.model_config:
--- a/tests/models/common/test_model_loader.py
+++ b/tests/models/common/test_model_loader.py
@@ -133,6 +133,39 @@ def test_get_model_architecture_unsupported():
         model_loader._get_model_architecture(config)
 
 
+@pytest.mark.parametrize("env_value,is_draft", [
+    ("flax_nnx", False),
+    ("vllm", False),
+    ("flax_nnx", True),
+    ("vllm", True),
+])
+def test_resolve_model_impl_type_explicit(env_value, is_draft):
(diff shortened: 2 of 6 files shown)


tpu_inference/platforms/tpu_platform.py
```diff
@@ -308,6 +308,37 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        # CPU processor and instead normalizes inside the vLLM model's vision
+        # tower (FusedInputNorm). JAX-native multimodal models consume the
+        # processor's pixel_values directly and have no device-side norm, so
+        # they would silently run the ViT on unnormalized pixels. Keep the
+        # normalization in the CPU processor for the JAX-native path.
+        mm_cfg = getattr(vllm_config.model_config, "multimodal_config", None) \
+            if vllm_config.model_config else None
+        if mm_cfg is not None and getattr(mm_cfg, "mm_device_do_normalize",
+                                          False):
+            impl = envs.MODEL_IMPL_TYPE
```
QiliangCui: the logic of getting "impl" is ugly.

QiliangCui:
**Refactor: extract the impl-resolution into a shared helper instead of open-coding it here.**

The env-read → `"auto"`-resolve dance this PR adds to `check_and_update_config` is the 4th copy of the same pattern in the repo:

- `tpu_inference/models/common/model_loader.py` — `get_model` (reads `envs.MODEL_IMPL_TYPE`/`DRAFT_MODEL_IMPL_TYPE`, resolves `"auto"`)
- `tpu_inference/spec_decode/jax/eagle3.py` — draft/target impl comparison
- `tpu_inference/spec_decode/jax/dflash.py` — same block as eagle3
- this PR's hunk in `tpu_platform.py`

Add one helper next to `resolve_model_architecture` in `model_loader.py` and collapse all four sites onto it:

```python
def resolve_model_impl_type(vllm_config: VllmConfig,
                            is_draft_model: bool = False) -> str:
    """Effective model impl type ("flax_nnx" | "vllm"), resolving "auto"
    from the model architecture."""
    impl = (envs.DRAFT_MODEL_IMPL_TYPE
            if is_draft_model else envs.MODEL_IMPL_TYPE)
    if impl == "auto":
        impl = resolve_model_architecture(vllm_config, is_draft_model)
        logger.info("Resolved MODEL_IMPL_TYPE 'auto' to '%s'", impl)
    return impl
```

Notes for the change:

- **`tpu_platform.py`**: keep the lazy function-level import and the try/except with the `flax_nnx` fallback — defaulting to "disable device norm" is the correctness-safe direction (with the flag off, the vLLM torch model installs `FusedInputNorm.identity()`, so a wrong fallback only costs the fusion optimization, never correctness). The site shrinks to:

  ```python
  from tpu_inference.models.common.model_loader import resolve_model_impl_type
  try:
      impl = resolve_model_impl_type(vllm_config)
  except Exception as e:
      logger.warning(
          "[tpu_platform] resolve_model_impl_type failed (%s); assuming "
          "the JAX-native path for the mm_device_do_normalize check.", e)
      impl = "flax_nnx"
  ```

- **`get_model`**: keep the "Loading model with MODEL_IMPL_TYPE=..." log; the auto-resolution log moves into the helper (drop the now-duplicate line there).
- **`eagle3.py` / `dflash.py`**: the two blocks become
  ```python
  draft_model_impl = resolve_model_impl_type(self.vllm_config, is_draft_model=True)
  target_model_impl = resolve_model_impl_type(self.vllm_config, is_draft_model=False)
  ```
  and drop their direct `resolve_model_architecture` imports if unused afterwards.
- **Tests**: the new `test_check_and_update_config_mm_device_do_normalize` should pass unchanged — it patches `tpu_platform.envs.MODEL_IMPL_TYPE` and `model_loader.resolve_model_architecture`, both of which the helper still reads/calls under those names (`envs` is the same module object in both files, and the helper calls `resolve_model_architecture` through `model_loader`'s module global, which the patch intercepts).
- Run yapf/isort/ruff on the touched files before pushing.

(Resolving once in `check_and_update_config` and storing the result on the config so `get_model`/spec-decode read it instead of re-resolving would be the nicer end-state, but that's a wider refactor — follow-up, not this PR.)

QiliangCui:
Addressed in `aede7398e`: added `resolve_model_impl_type` next to `resolve_model_architecture` and collapsed all four sites onto it (`get_model`, `eagle3.py`, `dflash.py`, and the `tpu_platform.py` hunk — the platform site keeps the lazy import and the flax_nnx-fallback try/except). The auto-resolution log moved into the helper; `get_model` keeps its "Loading model with MODEL_IMPL_TYPE=..." line. Local run: 75/75 across tests/platforms, tests/spec_decode (dflash + eagle3) and tests/models/common/test_model_loader.py on tpu7x.

QiliangCui:
**Drop the try/except around `resolve_model_impl_type` in `check_and_update_config` — replace it with a bare `impl = resolve_model_impl_type(vllm_config)`.**

Rationale:

1. **Any exception it catches re-throws in the worker anyway.** The helper is deterministic on (env value, hf_config, load_format), and `get_model()` calls the same function with the same inputs at model load. Swallowing here never saves the server — it just delays the identical crash until after engine startup and prepends a misleading "assuming JAX-native path" warning to the real error. Failing fast at config time is strictly better.
2. **Explicit-impl lanes can't throw.** `MODEL_IMPL_TYPE=vllm|flax_nnx` returns before any throwable code. Only the `auto` lane reaches `resolve_model_architecture`, whose failure modes are: an invalid env value (`env_with_choices` ValueError), the `assert len(architectures) == 1`, and the runai-streamer registry lookup.
3. **No current flow can hit those.** This code path only runs for models whose vLLM class declares `supports_mm_device_do_normalize` — today only the Qwen2-VL family, which has exactly one architecture, and none of those lanes use runai_streamer. Removing the guard cannot break any currently-green path.
4. **As written it doesn't guard the riskiest step anyway**: the `model_loader` import (which pulls the jax/flax import graph into the frontend process) sits outside the try.
5. **Keeping it can mask future bugs as silent de-optimization**: if resolution ever throws while the real impl is `vllm`, the fallback forces the flag off — output stays correct (the torch model installs `FusedInputNorm.identity()`), but the fused-norm optimization is silently lost behind a buried startup warning. The except branch is also untested.

Keep the explanatory comment block above the check; just remove the guard.

QiliangCui:
Done in `2d55c8af6` — dropped the try/except; the site is now a bare `impl = resolve_model_impl_type(vllm_config)` with the explanatory comment block kept. 43/43 platform tests pass locally (the parametrized mm_device_do_normalize cases exercise the bare call on all three impl lanes).

lk-chen (approved):

sixiang-google (approved):

kwang3939 (approved):

gxd3 (approved):

Lumosis (approved):

PR #3406, opened 2026-08-17
[gemma4] Fix torchax-path E2B/E4B accuracy: inline PLE + skip KV-cache writes for KV-shared layers

QiliangCui:
## Problem

gemma-4 E2B/E4B Accuracy has failed on every `MODEL_IMPL_TYPE=vllm` nightly since June (e.g. [build 24382](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/24382), both tpu6e and tpu7x). Two independent bugs:

**1. Engine-init crash (the visible failure).** vLLM's `Gemma4ForConditionalGeneration` caches per-layer embeddings (PLE) in a pre-allocated plain-attribute buffer (`self.per_layer_embeddings = torch.zeros(...)`, a CUDA-graph pattern). It is not a parameter or registered buffer, so `shard_model_to_tpu` never converts it to a torchax tensor and `embed_input_ids`'s `per_layer_embeddings[:n].copy_(...)` crashes in torchax's `_aten_copy`:

```
AttributeError: 'Tensor' object has no attribute '_elem'
```

Only the Accuracy job hits it because it runs the engine-init precompile (the UnitTest job's `offline_inference.py` sets `SKIP_JAX_PRECOMPILE=1`). Even if converted, the write cannot work here: the runner materializes `jax_view(params_and_buffers)` once at load, so a stateful slice-`copy_` never crosses the jit boundary.

**2. KV-share cache corruption (masked by the crash).** With (1) fixed, text generation is garbage. E2B/E4B mark their last 20/18 layers KV-shared: `kv_sharing_target_layer_name` is set and `layer_name_to_kvcache_index` maps them to the target layer's cache index. `PallasAttentionBackendImpl.forward` wrote `new_kv_cache` back unconditionally, so every shared layer overwrote its target's cache with its own (supposed-to-be-discarded) K/V. Consistent with the nightly matrix: 26B/31B (`num_kv_shared_layers=0`) Accuracy passes on the vllm variant; E2B/E4B (20/18 shared layers) never has.

## Changes

- `gemma4_mm_patcher.py`: drop the PLE buffer and compute PLE inline in the jitted forward from `input_ids` (a pure embedding lookup), masking multimodal placeholder token ids to 0 — exactly the reference `is_multimodal` masking (placeholder positions hold those ids; gemma-4 `image_token_id=258880 < vocab_size_per_layer_input=262144`, so `get_per_layer_inputs`'s internal mask does not cover them).
- `flash_attn.py`: pass `update_kv_cache=(kv_sharing_target_layer_name is None)` through to `attention()`, mirroring the JAX-native model (`models/jax/gemma4.py`: `update_kv_cache=not self.is_kv_shared_layer`). The returned (unchanged) cache is still stored: the argument is donated, and always threading the returned value keeps the cache use-chain linear so XLA input-output aliasing holds without copy insertion (a conditional skip produced a 170G HLO-temporaries OOM).

## Verification

- Unit tests: `test_gemma4_mm_patcher.py` (masking, buffer drop, forward wiring) and `test_flash_attn.py::test_forward_kv_shared_layer_does_not_write_cache` (real kernel: shared layer leaves the cache bit-identical, non-shared layer writes). All pass on tpu7x.
- A/B on tpu7x, `MODEL_IMPL_TYPE=vllm`, gemma-4-E2B offline_inference with precompile enabled: unpatched crashes with the `_aten_copy` AttributeError; with only the PLE fix it runs but generates token soup and scores gsm8k 0.0 ([dev 896](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/896)); with both fixes the output matches the JAX-native path ("The capital of France is **Paris**.") and gsm8k strict-match is **0.494** (threshold 0.35).
- [Dev 897](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/897): E2B Accuracy (tpu6e TP=1) and E4B Accuracy (tpu7x TP=2) both pass; MM correctness (Qwen2.5-VL) passes, i.e. no regression on non-gemma multimodal.

## Update (2026-08-21): rebased on main, reverts #3414, and a third torchax bug

- Rebased onto current `main`.
- Added a revert of #3414 (`9fd55bc67f`), which had marked the gemma-4 E2B/E4B **Accuracy** steps as `unverified` on the `MODEL_IMPL_TYPE=vllm` nightly variant while the torchax path was broken. With this PR the vllm variant runs those accuracy tests for real again, so merging this PR and lifting the gate are one atomic change. This also makes #3436 (which would copy the same gate onto the Benchmark step) unnecessary.
- **Third bug, found by the verification below:** with the PLE + KV-share fixes in place, text inference is correct, but any request carrying an *image* still killed the EngineCore: since vLLM #42662 (Aug 13) `Gemma4ForConditionalGeneration._process_image_input` sizes vision-encoder chunks with `torch.accelerator.get_memory_info()`, and on the torchax path (PrivateUse1 `jax` device) every `torch.accelerator` memory API raises `RuntimeError: PyTorch is not linked with support for jax devices`. The 26B/31B nightly perf steps don't hit it only because they set `cudagraph_mm_encoder`, which routes the encoder through `MMEncoderJitManager` and never enters `_process_image_input`. Fix: `tpu_platform.py` now shims `torch.accelerator.get_memory_info` to answer from `jax.local_devices()[0].memory_stats()`, mirroring the existing `empty_cache` shim; unit-tested in `tests/platforms/test_tpu_platform.py`.

### Verification on the dev pipeline

The complete per-model yml step set (Unit tests → Accuracy → Performance benchmarks), for **both** `MODEL_IMPL_TYPE=flax_nnx` and `MODEL_IMPL_TYPE=vllm`, on **both** tpu6e and tpu7x, with `soft_fail` removed so every step is hard-gated:

- [build 918](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/918) — `google_gemma-4-E2B-it.yml` (TP=1 on both chips): **12/12 passed**. gsm8k strict-match on the vllm path: tpu6e 0.485, tpu7x 0.492 (flax_nnx: 0.493 / 0.494; floor 0.35).
- [build 919](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/919) — `google_gemma-4-E4B-it.yml` (tpu6e TP=1, tpu7x TP=2): **12/12 passed**. gsm8k strict-match on the vllm path: tpu6e 0.715, tpu7x 0.717 (flax_nnx: 0.716 / 0.718; floor 0.65). flax_nnx perf: 5.1 req/s tpu6e, 6.9 req/s tpu7x.
- Caveat that led to the third fix: in 918/919 the *vllm* Performance steps reported PASSED but with `Total generated tokens: 0` — the EngineCore had died on the first image request (the `get_memory_info` crash above) and the recipe's 0.0 throughput floor let it through. The flax_nnx perf runs were real (E2B: 8.5 req/s tpu7x, 6.3 req/s tpu6e).
- [build 927](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/927) — the four vllm Performance steps (E2B/E4B × tpu6e/tpu7x) re-run with the shim: **4/4 real passes** — 128/128 requests, 16384 generated tokens each, no EngineCore errors. Request throughput: E2B 1.09 req/s tpu7x / 1.05 tpu6e, E4B 0.72 tpu7x / 0.91 tpu6e (floor 0.0). Slower than the JAX path (5–8.5 req/s) because these steps run the eager vision encoder without `cudagraph_mm_encoder`; perf tuning is a follow-up, correctness is what this PR restores.

### DIFF
--- a/.buildkite/models/google_gemma-4-E2B-it.yml
+++ b/.buildkite/models/google_gemma-4-E2B-it.yml
@@ -67,16 +67,7 @@ steps:
       MINIMUM_ACCURACY_THRESHOLD: 0.35
     commands:
       - |
-        if [ "${MODEL_IMPL_TYPE:-auto}" = "vllm" ]; then
-          echo "gemma-4 E2B on the vllm (torchax) path has never passed" \
-               "accuracy: the model's per_layer_embeddings CUDA-graph buffer" \
-               "is incompatible with torchax and its KV-shared layers corrupt" \
-               "the target layer's cache. Recording as unverified until the" \
-               "torchax path is fixed."
-          buildkite-agent meta-data set "${TPU_VERSION:-tpu6e}_google_gemma-4-E2B-it_Accuracy" "unverified"
-        else
-          .buildkite/scripts/run_in_docker.sh bash -c 'bash /workspace/tpu_inference/tests/e2e/benchmarking/test_accuracy.sh'
-        fi
+        .buildkite/scripts/run_in_docker.sh bash -c 'bash /workspace/tpu_inference/tests/e2e/benchmarking/test_accuracy.sh'
 
   - label: "${TPU_VERSION:-tpu6e} Record integration test result for google/gemma-4-E2B-it"
     key: "${TPU_VERSION:-tpu6e}_record_google_gemma-4-E2B-it_Accuracy"
--- a/.buildkite/models/google_gemma-4-E4B-it.yml
+++ b/.buildkite/models/google_gemma-4-E4B-it.yml
@@ -65,16 +65,7 @@ steps:
       MINIMUM_ACCURACY_THRESHOLD: 0.65
     commands:
       - |
-        if [ "${MODEL_IMPL_TYPE:-auto}" = "vllm" ]; then
-          echo "gemma-4 E4B on the vllm (torchax) path has never passed" \
-               "accuracy: the model's per_layer_embeddings CUDA-graph buffer" \
-               "is incompatible with torchax and its KV-shared layers corrupt" \
-               "the target layer's cache. Recording as unverified until the" \
-               "torchax path is fixed."
-          buildkite-agent meta-data set "${TPU_VERSION:-tpu6e}_google_gemma-4-E4B-it_Accuracy" "unverified"
-        else
-          .buildkite/scripts/run_in_docker.sh bash -c 'bash /workspace/tpu_inference/tests/e2e/benchmarking/test_accuracy.sh'
-        fi
+        .buildkite/scripts/run_in_docker.sh bash -c 'bash /workspace/tpu_inference/tests/e2e/benchmarking/test_accuracy.sh'
 
   - label: "${TPU_VERSION:-tpu6e} Record integration test result for google/gemma-4-E4B-it"
     key: "${TPU_VERSION:-tpu6e}_record_google_gemma-4-E4B-it_Accuracy"
--- a/tests/layers/vllm/backends/test_flash_attn.py
+++ b/tests/layers/vllm/backends/test_flash_attn.py
@@ -216,6 +216,54 @@ class TestPallasAttentionBackendImpl:
                 shared_attn_metadata=shared_metadata):
             impl.forward(layer, query, key, value, torch.tensor([]), metadata)
 
+    def test_forward_kv_shared_layer_does_not_write_cache(self, mesh):
+        """KV-shared layers (kv_sharing_target_layer_name set, e.g. gemma-4
+        E2B/E4B cross-decoder) map to their target layer's cache index, so
+        writing would overwrite the target's entries with this layer's
+        discarded K/V. The write must only happen for non-shared layers."""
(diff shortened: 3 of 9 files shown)


lk-chen (commented):
@anthonsu could you take a look? shall we use ViT cudagraph for Gemma?

anthonsu:
> @anthonsu could you take a look? shall we use ViT cudagraph for Gemma?

Yes, I believe we should. Vision encoder overhead could have a larger impact on E2B/E4B due to their smaller backbones. The first PLE fix enables ViT cudagraph for E2B and E4B precompile. The second fix addresses the kv_shared_layer issue specific to E2B/E4B in this case. 

Thanks Qiliang for making the changes!

anthonsu (approved):

gxd3 (approved):

lk-chen (approved):

theminghuang (approved):

Lumosis (approved):

sixiang-google (approved):

PR #3407, opened 2026-08-17
[multimodal] Raise ViT flash-attention scoped-vmem limit to 64MiB

QiliangCui:
## Problem

The `MODEL_IMPL_TYPE=vllm` nightly's "Performance tests for Multimodal Inputs" fails on both tpu6e and tpu7x (e.g. [build 24382](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/24382)): under the MM perf request mix, large flattened image sequences push the ViT flash-attention pallas kernel past the default 32MiB scoped-vmem limit and the engine dies mid-benchmark:

```
CompileTimeScopedVmemOom: Ran out of memory in memory space vmem ... for
%flash_attention.1 = bf16[1,16,25344,80] ... Scoped allocation with size
37.22M and limit 32.00M exceeded scoped vmem limit by 5.22M.
```

## Change

Pass `vmem_limit_bytes=64MiB` to `sharded_flash_attention` at the two vision SDPA entry points (`scaled_dot_product_attention`, `vllm_vit_sdpa`). The plumbing already existed down to the kernel; only these call sites never set it. 64MiB covers the observed 37–38M shapes with headroom, and the RPA kernels already run with up to 100MiB on both v6e and v7x.

## Verification

- New unit tests assert both ops pass the limit through and that it covers the failing shape.
- [Dev 897](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/897) (`MODEL_IMPL_TYPE=vllm`): MM perf passes on BOTH tpu7x and tpu6e (previously failed every nightly), and MM correctness passes (no regression).

### DIFF
--- a/tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
+++ b/tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
@@ -21,12 +21,10 @@ from torchax.ops.jtorch import register_function
 from tpu_inference.layers.common.attention_interface import \
     sharded_flash_attention
 
-# The pallas default scoped-vmem limit is 32MiB, which the ViT flash-attention
-# exceeds on large flattened image sequences: nightly 24382
-# (MODEL_IMPL_TYPE=vllm, Qwen2.5-VL-3B MM perf) failed with
-# CompileTimeScopedVmemOom "size 37.22M and limit 32.00M" at
-# bf16[1,16,25344,80]. 64MiB covers those shapes with headroom; the RPA
-# kernels already run with up to 100MiB on both v6e and v7x.
+# ViT flash-attention on large flattened image sequences needs more scoped
+# vmem than the 32MiB pallas default (~37MiB at bf16[1,16,25344,80]); 64MiB
+# covers such shapes with headroom. The RPA kernels already run with up to
+# 100MiB on both v6e and v7x.
 _VIT_FLASH_ATTENTION_VMEM_LIMIT_BYTES = 64 * 1024 * 1024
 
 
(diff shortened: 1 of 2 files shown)


lk-chen (approved):

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -21,6 +21,14 @@
 from tpu_inference.layers.common.attention_interface import \
     sharded_flash_attention
 
+# The pallas default scoped-vmem limit is 32MiB, which the ViT flash-attention
+# exceeds on large flattened image sequences: nightly 24382
```
QiliangCui: We should not mention nightly 24382 in incline code comment.

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -21,6 +21,14 @@
...
+# (MODEL_IMPL_TYPE=vllm, Qwen2.5-VL-3B MM perf) failed with
+# CompileTimeScopedVmemOom "size 37.22M and limit 32.00M" at
```
QiliangCui: remove this comments here and put it PR description.

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
QiliangCui: Done — removed the build reference; the comment is now self-contained (also scrubbed it from the test docstring).

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
QiliangCui: Done — trimmed the failure-log details from the comment; they were already in the PR description (Problem section).

gxd3 (approved):

vanbasten23 (approved):

PR #3408, opened 2026-08-17
[gemma4] Make 26B/31B benchmarks pass on the torchax path (HBM headroom, startup timeout, variant bars)

QiliangCui:
## Problem

The `MODEL_IMPL_TYPE=vllm` nightly's gemma-4-26B/31B performance benchmarks (tpu7x-8, TP=8) have failed every run (e.g. [build 24382](https://buildkite.com/tpu-commons/tpu-inference-ci/builds/24382)):

```
RuntimeProgramAllocationFailure: Error loading program 'jit__forward':
Attempting to allocate 1.13G. That was not possible. There are 528.72M free.
```

## Root cause

At the recipe default `gpu_memory_utilization=0.98`, KV sizing fills tpu7x HBM to ~92.9/94.7GiB. The torchax path's compiled programs (backbone shapes + mm-encoder cudagraph token budgets) total ~6GiB+ at load — measured by iterating: 0.95 (~4.7GiB free) still hit the same failure ~28 minutes into startup ([dev 898](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/898)). The jax-native path has a smaller program footprint and fits at 0.98, which is why only the vllm variant was red.

## Changes (both Benchmark steps)

1. `GPU_MEMORY_UTILIZATION: "0.90"` — frees ~9.5GiB for program load; KV concurrency only drops 153x → ~140x, far above benchmark needs.
2. `TIMEOUT_SECONDS` 960 → 2400 — cold-JAX-cache startup takes ~30min (vision-encoder cudagraph budgets compile at ~3min/shape); warm nightly agents start much faster.
3. Variant-aware throughput bars — with the OOM fixed, the torchax path measures 1.86–1.91 req/s (26B) and 1.59–1.62 (31B) across three runs ([dev 899](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/899)–[901](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/901)), ~8-10% below the jax-native-derived bars (2 / 1.80). `MODEL_IMPL_TYPE=vllm` is held to measured × ~0.85 (1.55 / 1.35) so the nightly gates regressions without demanding jax-native throughput from the torchax bridge. The jax-native bars are unchanged. Closing the gap is a separate optimization item.

## Verification

[Dev 901](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/901) (`MODEL_IMPL_TYPE=vllm`, tpu7x-8, final yml values): 26B **1.91 req/s ≥ 1.55 PASSED**, 31B **1.61 req/s ≥ 1.35 PASSED**.

### DIFF
--- a/.buildkite/models/google_gemma-4-26B-A4B-it.yml
+++ b/.buildkite/models/google_gemma-4-26B-A4B-it.yml
@@ -91,6 +91,15 @@ steps:
       TEST_MODEL: google/gemma-4-26B-A4B-it
       TENSOR_PARALLEL_SIZE: 8
       MINIMUM_THROUGHPUT_THRESHOLD: 2
+      # 0.98 (recipe default) leaves ~1.9GiB/chip after KV sizing on tpu7x
+      # (94.74GiB HBM), but the torchax-path compiled programs (backbone
+      # shapes + mm-encoder cudagraph budgets) total ~6GiB+ at load:
+      # nightly 24382 (MODEL_IMPL_TYPE=vllm) failed with "Error loading
+      # program 'jit__forward': Attempting to allocate 1.13G. There are
+      # 528.72M free.", and 0.95 (~4.7GiB free) still hit the same error
+      # late in startup (dev builds 897/898). 0.90 frees ~9.5GiB; KV
+      # concurrency only drops 153x -> ~140x, far above benchmark needs.
+      GPU_MEMORY_UTILIZATION: "0.90"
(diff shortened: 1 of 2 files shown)


lk-chen (approved):

anthonsu (approved):
LGTM. Just to confirm: are we aligning with gemma-4's native token budgets in CI testing? Previous local benchmark runs showed better results than budgets like [64, 128, ...], FYI.

theminghuang (approved):

PR #3416, opened 2026-08-18
[CI] Name the failing steps in TPU test notification + retry unit tests on infra failures

lokic233:
## What

Makes a red `buildkite/tpu-inference-ci/pr` build **actionable** and reduces false reds from transient TPU infra blips. Two small, self-contained changes:

### 1. `check_results.sh` — name the failing steps (shared by the jax + torch notification steps)

The TPU unit-test steps run with `soft_fail: true`, so an individual failure is reported to GitHub as a **green** status. The `tpu-test-notification` step (which runs `check_results.sh`) is what turns a soft-failed step into the real red on the umbrella `.../pr` context — but today it only emits a generic `"… Tests Failed"` label, so a red build **doesn't say which step failed**. You have to open Buildkite and scan ~37 steps to find it.

This change collects the non-passing step keys, prints them in the build log, and emits a Buildkite **error annotation** naming them. Behavior is otherwise identical:
- still `exit 1` on any non-`passed`/non-`skipped` outcome,
- still sanitizes the caller-supplied label before YAML interpolation (unchanged injection guard),
- still uses a fixed `command:` body,
- the annotation is best-effort and **never** changes the aggregator's exit status.

### 2. `pipeline_jax.yml` — retry the JAX unit-test steps on **infrastructure** failures only

part1/part2 get `retry.automatic` for `exit_status: -1` (agent lost) and `255` (conn/timeout), `limit: 2`. This recovers a transient TPU preemption / lost-agent blip without a manual re-run. It deliberately does **not** retry `exit_status: 1` (a deterministic pytest failure), so a real test failure still surfaces immediately — and part2 is a ~1h20m sweep, so blindly re-running it on every non-zero exit would be costly and pointless. This uses the same `retry.automatic` idiom already present in `pipeline_kernel_autotune_template.yml`.

## Why

Context: a fork PR's build went red on the umbrella context while **every** substantive test job was green. Root cause was a `soft_fail` in the large `part2` unit sweep (which flakes intermittently on `main` itself) → the `tpu-test-notification` step hard-failed → umbrella red, **with the failing step's identity hidden** behind the generic label. This makes triage slow and makes flaky infra look like a code failure.

## Tests

`check_results.sh` was validated with a mock `buildkite-agent` harness (included in the PR description below for reproducibility), covering:
- all-pass → `exit 0`, no annotation;
- one `soft_failed` → `exit 1`, log + annotation **name the failing key** (one line per `key (outcome)`);
- multiple failures → all named;
- `skipped` is allowed (→ `exit 0`);
- an adversarial label (`evil"; rm -rf / #`) → still `exit 1`, label **sanitized** in the uploaded YAML, fixed command body.

`shellcheck -s sh` is clean. Both `pipeline_jax.yml` and `pipeline_torch.yml` parse as valid YAML, and the two new `retry` blocks match Buildkite's `automatic: [{exit_status, limit}]` schema.

## Scope / follow-ups (intentionally NOT in this PR, to keep it reviewable)

- **Drop `-x` on part2** so a failing run reports *all* failures (full per-test attribution) instead of aborting at the first.
- **Rebalance the split**: part1 is 5 hard-coded files, part2 is ~119 — shard by count/time (e.g. `pytest-split`) to shrink the ~1h20m tail and blast radius.
- **Reconcile the `soft_fail`-step + hard-`notification` design** so "soft" and "hard" don't contradict at the umbrella level.

Happy to follow up on any of these in separate PRs.

## Review

The shell change was independently reviewed for correctness (exit-code semantics, label-injection safety, `retry`×`soft_fail` interaction, and POSIX-sh portability). That review surfaced one real portability nit — the failing-step accumulator was piped to `sed` without a trailing newline (`printf '%s'`), which strict POSIX `sed` (dash/BusyBox/BSD) can silently truncate, dropping the **last** failing step from the annotation. Fixed by using `printf '%s\n'` consistently across all three `printf | sed` sites. Re-validated: `shellcheck -s sh` clean, and a mock-`buildkite-agent` harness confirms all-pass→exit 0 / any-fail→exit 1 with every failing step named on its own line.

---
_Authored with [Navi](https://navibot.dev?utm_source=github&utm_medium=pr_description&utm_campaign=tpuinf_ci_stability&ref=dengcchi) on behalf of Loki Chen (@lokic233)._

### DIFF
--- a/.buildkite/pipeline_jax.yml
+++ b/.buildkite/pipeline_jax.yml
@@ -104,6 +104,15 @@ steps:
       - label: "${TPU_VERSION:-tpu6e} JAX unit tests part1"
         key: ${TPU_VERSION:-tpu6e}_test_7_1
         soft_fail: true
+        # Retry only on infrastructure-level failures (agent lost / preemption /
+        # timeout), NOT on pytest exit 1 (a deterministic test failure), so a real
+        # failure still surfaces without re-running the suite. -1 = agent lost.
+        retry:
+          automatic:
+            - exit_status: -1
+              limit: 2
+            - exit_status: 255
+              limit: 2
         artifact_paths: ".coverage.part1.${TPU_VERSION:-tpu6e}"
         env:
           USE_PREBUILT_IMAGE: "1"
@@ -119,6 +128,16 @@ steps:
       - label: "${TPU_VERSION:-tpu6e} JAX unit tests part2"
         key: ${TPU_VERSION:-tpu6e}_test_7_2
         soft_fail: true
+        # Retry only on infrastructure-level failures (agent lost / preemption /
+        # timeout), NOT on pytest exit 1 (a deterministic test failure). part2 is
+        # a large (~1h20m) sweep, so re-running on a real failure would be costly
+        # and pointless; re-running on a lost agent recovers a transient blip.
+        retry:
+          automatic:
+            - exit_status: -1
+              limit: 2
+            - exit_status: 255
+              limit: 2
         artifact_paths: ".coverage.part2.${TPU_VERSION:-tpu6e}"
         env:
           USE_PREBUILT_IMAGE: "1"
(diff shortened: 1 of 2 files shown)


kyuyeunk:
@QiliangCui do you thin this is a good change?

PR #3277, opened 2026-07-27
PCP + ring attention

wenxindongwork:
## Summary

This PR adds a third PCP cache-phase strategy — `cache_phase="ring"` in addition to the existing gather_kv/gather_q strategy. 

| phase | how every rank sees the cache | cost |
| --- | --- | --- |
| `gather_q` | all-gather Q, reduce-scatter the output | two collective rounds, comm independent of context |
| `gather_kv` | all-gather the striped cache into one buffer | comm and peak memory grow with context |
| **ring** | rotate KV around the PCP ring inside the kernel | no exposed collective outside of the kernel, ici traffic grows with context |

Ring is now the default behavior for PCP since it outcompetes gather_q and gather_kv in almost every setting. 

## Implementation details

Each rank keeps its local Q and cache stripe; the kernel rotates the cache one block at a time with `make_async_remote_copy`.

The conceptual change vs the non-ring path is how bkv is prefetched. In the non-ring path, we keep a bkv double buffer and prefetch the next bkv block from HBM. In the ring path, we still keep a bkv double buffer, but prefetch the next block via ICI from the direct neighbor (except for the first ring step which prefetches from HBM). 

We also added a custom block size calculation logic for the ring path, which maximizes bq_sz. 

## PCP strategy performance comparison. 

```
CH=4k, NQ=32 NKV=2 HD=256, bf16:

pcp4 × tp2 (baseline = tp8)

┌─────────┬─────────────┬────────────────────────┬───────────────────────┬────────────────────────┬────────────────────────┬───────────────────────┐
│ Context │ tp8 (8 dev) │      tp4 (4 dev)       │     pcp_ag (auto)     │        pcp_ring        │       gather_kv        │       gather_q        │
├─────────┼─────────────┼────────────────────────┼───────────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│      1k │        0.75 │     0.79 (5.3% slower) │   0.86 (14.7% slower) │    0.85 (13.3% slower) │    0.85 (13.3% slower) │   0.86 (14.7% slower) │
├─────────┼─────────────┼────────────────────────┼───────────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│      2k │        0.77 │     0.81 (5.2% slower) │   0.90 (16.9% slower) │    0.89 (15.6% slower) │    0.89 (15.6% slower) │   0.94 (22.1% slower) │
[...]

PR #3166, opened 2026-07-13
PCP kernel support for single request

wenxindongwork:
# Description

This pull request introduces support for PCP (Prefill Context Parallel) mode in the `experimental/rpa_v3_cp/kernel.py` kernel. Currently PCP only supports processing one request at a time (`max_num_seqs=1`). 

**Correctness test**

Tested with this PR + e2e PCP draft PR https://github.com/vllm-project/tpu-inference/pull/3096.

```
SKIP_JAX_PRECOMPILE=1 NEW_MODEL_DESIGN=1 vllm serve  Qwen/Qwen3-8B-Base  --max-model-len=4096   --max-num-batched-tokens=4096 --max_num_seqs=1  --gpu-memory-utilization=0.
9   --tensor-parallel-size=4   --prefill-context-parallel-size 2

lm_eval --model local-chat-completions     --model_args base_url="http://localhost:8000/v1/chat/completions",model="Qwen/Qwen3-8B-Base",eos_string="<|im_end|>",num_concurrent=256     --tasks gsm8k     --apply_chat_template    --fewshot_as_multiturn --limit 200

```

```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.845|±  |0.0257|
|     |       |strict-match    |     5|exact_match|↑  |0.845|±  |0.0257|
```

**TTFT Benchmark**

v7x-8, NQ=16, NKV=4 (GQA), HD=128, bf16

```

Chunk-size sweep (context = 128K):

┌───────┬──────────┬──────────┬───────────────┬─────────┬──────────┬──────────────┐
│ chunk │ tp4 (ms) │ tp8 (ms) │ pcp2_tp4 (ms) │ tp4/tp8 │ tp4/pcp2 │ pcp2 vs tp8  │
├───────┼──────────┼──────────┼───────────────┼─────────┼──────────┼──────────────┤
│    4k │     81.5 │     44.9 │          73.3 │   1.82× │    1.11× │ 1.63× slower │
├───────┼──────────┼──────────┼───────────────┼─────────┼──────────┼──────────────┤
│    8k │     77.3 │     41.1 │          54.8 │   1.88× │    1.41× │ 1.34× slower │
├───────┼──────────┼──────────┼───────────────┼─────────┼──────────┼──────────────┤
[...]

PR #3425, opened 2026-08-18
[PCP] Multi-request prefill context parallelism on the ring cache phase

bhuvanpkaruturi:
# Description

Prefill context parallelism (PCP, #3096 / #3166 / #3277) currently serves **exactly one request per step**: `_prepare_inputs` asserts a single scheduled request whenever PCP is on, so a prefill server with a queue leaves the slice idle for every client but one. This PR lets a PCP step carry **several prefill requests at once**, on top of the in-kernel ring cache phase from #3277.

The idea is to lay requests side by side on every rank, each one cut into its own zigzag head/tail chunk pair exactly as a single request is today, and to describe each request to the attention kernel separately. Nothing about the ring itself changes; the kernel already handles many sequences per launch, per sequence.

## What changes, at a high level

**Token layout (runner).** Request `i` with `n_i` scheduled tokens gets a per-request chunk `C_i = ceil(n_i / 2P)` (rounded so the strided KV write stays page-aligned) and occupies `2 C_i` rows on every rank: chunk `r` (head) and chunk `2P-1-r` (tail) on rank `r`. Requests are concatenated at offsets `off_i`; every rank holds `S = sum 2 C_i` rows, and the padded per-rank width `S_pad = T_pad / P`. A lone request keeps today's single-request layout (`chunk = T_pad / 2P`), so the `R = 1` code path and its numerics are untouched. Layout is built over live requests only; the static request-count bucket is `{1, max_num_seqs}`.

**Cache phase (queries vs. cached KV) = one ring seq per request.** Instead of one kernel seq covering the whole local buffer, the ring cache phase gets `R` seqs, one per request, each with its own cached length and page list. Its metadata is a strided view of the current-phase metadata (`cu_q_lens[0::2]`, `kv_lens[0::2]`, `kv_cache_lens[0::2]`, every other block-table row, `distribution[2] // 2`) — no new arrays, no collectives. Because all ranks hold identical `cu_q_lens`, cache lengths and `request_distribution`, every rank runs the identical (seq, tile, block, round) schedule, which is what the ring's matched remote copies require. `R == 1` keeps the PR's whole-buffer seq unchanged.

**Current phase (queries vs. this step's KV).** The all-gathered current K/V is permuted (`kv_token_order`) into request-major token order so each request's new KV is one contiguous run, and the kernel gets per-seq `kv_new_starts` (offset of each seq's new KV in the gathered buffer) plus a per-seq `kv_write_seq_mask` (which seqs write their strided share of the KV cache — replaces `write_last_seq_only`). `R = 1` keeps the kernel-side `pcp_chunk_size` remap and pays nothing.

**Precompile.** The multi-request shapes are added to the PCP precompile set. Since #3277 the `cache_pages` value only decides whether the cache phase runs at all (0 elides it) — every nonzero rung compiles the same ring kernel — so the ladder collapses from 5 rungs to `{0, max}` (`PCP_CACHE_PAGE_BUCKET_COUNT = 2`), which halves the PCP precompile set and roughly halves PCP server startup.

## Costs and trade-offs

- **Collectives:** none added. The current phase's all-gather of current K/V is exactly what runs today; the cache phase has no collective.
- **Ring ICI traffic:** additive across requests, each against its **own** cache — a short request batched with a 1M-token request streams its own small stripe, not the big one — plus at most one extra query tile per request from rounding each request's tile count separately.
- **Compute:** sum over requests of what each costs alone; padding rows (up to `2P-1` per request) are computed but never read, <0.1%.
- **HBM:** for `R > 1` one gather over `T_pad` rows of K and V per layer to put the current KV in request-major order (~1 ms/step at 16k tokens over 36 layers).
- **Memory:** attention metadata sized for `2 x max_num_seqs` seqs; block table is `2 x max_num_seqs x pages_per_request` (4 MiB at 1M context).
- **Not supported:** batches mixing prefill and decode (a decode request would occupy `2P` query rows for one token). Target is a P/D-disaggregated prefill server, where this does not arise. `P` must be even (from #3277).

## Commits

| # | commit | what |
|---|---|---|
| 1 | `PCPMetadata` multi-request fields | `kv_token_order`, `kv_new_starts`, `kv_write_seq_mask`, per-request layout fields |
| 2 | Kernel: per-seq `kv_new_starts`, `kv_write_seq_mask` | replaces `write_last_seq_only`; per-seq addressing into the gathered new KV |
| 3 | `pcp_forward` for R requests | ring cache phase = one seq per request via strided views; current-phase reorder + per-request KV write; `R == 1` unchanged |
| 4 | Runner: multi-request token layout | per-request zigzag chunks, permutation, logits indices, block-table row expansion, buffer sizing; lone request keeps the single-request layout |
| 5 | Precompile the multi-request shapes | |
| 6 | Layout over live requests only; `cache_pages` ladder -> `{0, max}` | |
| 7 | Tests | kernel + interface, see below |

## Known issues found on the way (fixed in a separate follow-up, not in this PR)

Running this at 32k+ context surfaced pre-existing issues that are independent of multi-request support and will go up separately: (a) the RPA CP kernel's decode-case variant is compiled though unreachable under PCP and its default `bkv` overflows VMEM from ~32k `max_model_len`; (b) in the single-request `pcp_chunk_size` remap path, a new-KV run that starts at `kv_cache_len_local % bkv != 0` straddles a chunk boundary and one contiguous DMA fetches the wrong tokens (corrupt K/V, intermittent NaN at 32k with chunked prefill); (c) `get_min_page_size` does not count the `2 x max(max_num_seqs, 8)` seq rows the runner allocates under PCP (SMEM overflow at 1M `max_model_len`); (d) vLLM allocates the KV cache at the raw block size under PCP while the TPU runner sizes rows for `block_size x P` — being addressed by scaling DCP = PCP inside tpu-inference. None of these are needed for the tests in this PR to pass; the end-to-end numbers below were taken with (a)–(d) applied locally.

# Tests

**Unit (v7x-8, both suites, run at this PR's tip on the current `main` base, without any of the follow-up fixes):**
- `tests/kernels/rpa_v3_cp/ragged_paged_attention_kernel_cp_test.py`: 38 passed, 1 skipped (the opt-in zero-length-seq hang repro). New: fused current phase with several requests (per-request `kv_new_starts`, `kv_write_seq_mask`, every request's strided KV share written exactly once, ragged lengths, mixed cached/uncached); ring cache phase over three seqs in one launch with different cached lengths (one zero), row counts and page lists, `P = 2, 4`, f32 and bf16, against a per-seq full-cache reference; a zero-length seq in range.
- `tests/layers/common/test_pcp_attention_interface.py`: 25 passed. New: `pcp_forward` with `R` requests — ragged lengths, mixed cached/uncached, live count below the static bucket, per-request KV cache write, a request spanning several ring query tiles, the batch shape from a 32k end-to-end run in the headroom token bucket, ~1M cached tokens next to a short request; single-request cases unchanged.

**End-to-end correctness (Qwen3-8B, TP4 x PCP2 vs TP8, greedy, `max_tokens=1`, logprobs compared):** 8k context (`--max-model-len 16384 --max-num-batched-tokens 16384`, 8 concurrent prompts of 120–5.5k tokens): 32/32 identical at `--max-num-seqs 8` (multi-request batches of 3–8 live requests, 5 of them chunked prefills with `num_computed > 0`) and 32/32 at `--max-num-seqs 1`. 32k context (`--max-num-batched-tokens 16384`): 32/32 identical with 19 chunked multi-request batches. Re-run on the rebased tip (current `main` base) at 8k: 8/8 identical, multi-request batches confirmed.

**TTFT, PCP2 x TP4 (`--max-num-seqs 8`) vs TP4 without PCP**, Qwen3-8B on one v7x-8 (PCP arm uses 8 chips, TP4 arm 4), prefill-only, prefix caching off, `--max-model-len = ctx + 4096`, `--max-num-batched-tokens = min(16384, ctx/2)` so every prompt is prefilled in at least two chunks and the ring cache phase runs at every point. TTFT p50 in seconds, gain = p50 improvement over TP4:

| ctx (chunk) | conc | TP4 p50 | PCP p50 | TP4 tok/s | PCP tok/s | TTFT gain |
|---|---|---|---|---|---|---|
| 4k (2048) | 1 | 0.088 | 0.124 | 46549 | 33013 | -41.0% |
| 4k | 8 | 0.409 | 0.884 | 50511 | 22723 | -116.3% |
| 32k (16384) | 1 | 0.882 | 0.518 | 37043 | 63171 | **+41.2%** |
| 32k | 4 | 3.357 | 2.152 | 38701 | 60836 | +35.9% |
| 32k | 8 | 4.242 | 2.686 | 38692 | 60926 | +36.7% |
| 128k (16384) | 1 | 7.319 | 3.808 | 17875 | 34329 | **+48.0%** |
| 128k | 4 | 28.444 | 14.806 | 18318 | 35110 | +47.9% |
| 128k | 8 | 35.768 | 18.590 | 18321 | 35098 | +48.0% |
| 256k (16384) | 1 | 24.346 | 11.704 | 10746 | 22354 | **+51.9%** |
| 256k | 2 | 47.762 | 22.887 | 10922 | 22759 | +52.1% |
| 512k (16384) | 1 | 87.622 | 39.907 | 5972 | 13112 | **+54.5%** |
| 512k | 2 | 173.243 | 78.978 | 6030 | 12488 | +54.4% |

The gain is the same at every concurrency measured — multi-request batches formed throughout — so it holds under load rather than only at concurrency 1, and prefill throughput scales the same way (32k: 61k vs 39k tok/s at conc 4–8; 512k: 13.1k vs 6.0k). PCP2 x TP4 on 2x the chips delivers 1.6–2.2x the TTFT of TP4, close to linear from 128k up. 4k is a loss: with a 2048-token chunk each PCP step is tiny and the ring's fixed per-step costs dominate (matches #3277's own tables below ~8k) — the regime where PCP should not be enabled, not a multi-request effect.

# Checklist


tpu_inference/layers/common/cp_attention.py
```diff
@@ -362,11 +407,29 @@ def to_token_order(x):  # rank-order chunks -> global token order
         # pcp_cu_q_lens_local[0] = [0, chunk, chunk+tail_real]; pcp_q_pos_offsets_local[0] = [head_offset, tail_offset].
         # remap_kv: if C aligns with page_size, all_gather_tokens() avoids an extra gather-reorder.
         page_size = kv_cache_local.shape[1]
-        remap_kv = (C >= page_size) and (C % page_size == 0)
-        k_curr = all_gather_tokens(k_local) if remap_kv else to_token_order(
-            k_local)
-        v_curr = all_gather_tokens(v_local) if remap_kv else to_token_order(
-            v_local)
+        if multi_req:
```
weiyu0824: This is one of the example.

wenxindongwork:
thanks a lot for the PR! could we also compare pcp2 tp4 ttft performance against tp=8 (without attn_dp)?

bhuvanpkaruturi:
> thanks a lot for the PR! could we also compare pcp2 tp4 ttft performance against tp=8 (without attn_dp)?

| ctx (chunk) | conc | TP8 p50 | PCP p50 | TP8 p90 | PCP p90 | TP8 tok/s | PCP tok/s | TTFT gain vs TP8 |
|---|---|---|---|---|---|---|---|---|
| 4k (2048) | 1 | 0.066 | 0.125 | 0.066 | 0.126 | 62413 | 32819 | **−89.9%** |
| 4k | 2 | 0.115 | 0.372 | 0.125 | 11.237* | 69776 | 2663* | −225.2% |
| 4k | 4 | 0.230 | 0.743 | 0.245 | 0.747 | 68638 | 22789 | −222.6% |
| 4k | 8 | 0.301 | 0.878 | 0.470 | 1.434 | 69203 | 22782 | −191.8% |
| 32k (16384) | 1 | 0.571 | 0.516 | 0.573 | 0.517 | 57278 | 63349 | **+9.5%** |
| 32k | 2 | 1.046 | 1.081 | 1.095 | 17.124* | 61804 | 12863* | −3.4% |
| 32k | 4 | 2.090 | 2.152 | 2.138 | 2.161 | 61879 | 60796 | −3.0% |
| 32k | 8 | 2.665 | 2.695 | 4.230 | 4.298 | 61774 | 60801 | −1.1% |
| 128k (16384) | 1 | 4.462 | 3.802 | 4.467 | 3.807 | 29333 | 34393 | **+14.8%** |
| 128k | 2 | 8.492 | 7.453 | 8.716 | 23.754* | 30616 | 22740* | **+12.2%** |
| 128k | 4 | 16.974 | 14.797 | 17.210 | 14.886 | 30613 | 35151 | **+12.8%** |
| 128k | 8 | 21.442 | 18.574 | 34.178 | 29.761 | 30609 | 35150 | **+13.4%** |
| 256k (16384) | 1 | 14.360 | 11.705 | 14.371 | 11.732 | 18217 | 22347 | **+18.5%** |
| 256k | 2 | 27.833 | 22.881 | 28.304 | 23.102 | 18695 | 22774 | **+17.8%** |
| 512k (16384) | 1 | 50.600 | 39.843 | 50.638 | 39.854 | 10339 | 13134 | **+21.3%** |
| 512k | 2 | 99.232 | 78.818 | 100.245 | 93.548* | 10511 | 12523 | **+20.6%** |

weiyu0824 (commented):
There are lots of multi_req flag, I am thinking is it possible to unify single and multi-requests?

tpu_inference/layers/common/cp_attention.py
bhuvanpkaruturi: As discussed offline, this is intentional as the multi-req and single-req token layouts differ in this implementation. This is needed to optimize padding and DMA transfers. In a follow PR will try to merge them by scaling the head/tail chunk size to be a multiple of block_size.

bhuvanpkaruturi:
> There are lots of multi_req flag, I am thinking is it possible to unify single and multi-requests?

As discussed offline, this is intentional as the multi-req and single-req token layouts differ in this implementation. This is needed to optimize padding and DMA transfers. In a follow PR will try to merge them by scaling the head/tail chunk size to be a multiple of block_size.

tpu_inference/layers/common/attention_metadata.py
```diff
@@ -166,7 +186,11 @@ def __init__(
         groups, layer_names_per_group),
 )
 
-PCP_CACHE_PAGE_BUCKET_COUNT = 5
+# After PR #3277 (in-kernel ring cache phase) `cache_pages` only decides whether
+# the cache phase runs at all (0 elides it); every nonzero rung compiles the
+# same kernel, so the ladder is just {0, max}. Each rung multiplies the PCP
+# precompile set, so keep this at 2 unless the value gains another use.
+PCP_CACHE_PAGE_BUCKET_COUNT = 2
```
wenxindongwork: could me make this a bool instead?

tpu_inference/layers/common/cp_attention.py
```diff
@@ -315,10 +315,16 @@ def pcp_forward(
...
+    multi_req = num_reqs > 1
 
     def _shard_fn(q_local, k_local, v_local, kv_cache_local, kv_lens_local,
                   kv_cache_lens_local, page_indices_local, distribution_local,
-                  pcp_cu_q_lens_local, pcp_q_pos_offsets_local):
+                  pcp_cu_q_lens_local, pcp_q_pos_offsets_local, *pcp_extra):
+        # Multi-request extras (see in_specs below); empty when num_reqs == 1.
+        kv_new_starts_local = kv_token_order_local = None
+        if multi_req:
+            kv_new_starts_local, kv_token_order_local = pcp_extra
```
wenxindongwork: would it be possible that we also supply  kv_new_starts_local, kv_token_order_local in the single request case?

tpu_inference/layers/common/cp_attention.py
```diff
@@ -395,6 +459,20 @@ def to_token_order(x):  # rank-order chunks -> global token order
...
+        if md.pcp.kv_new_starts is None or md.pcp.kv_token_order is None:
+            raise ValueError(
+                "PCP with num_reqs > 1 requires md.pcp.kv_new_starts and "
+                "md.pcp.kv_token_order; the runner must build both.")
+        extra_specs = (
+            P(),  # pcp.kv_new_starts: replicated
+            P(),  # pcp.kv_token_order: replicated
+        )
+        extra_args = (md.pcp.kv_new_starts, md.pcp.kv_token_order)
+
```
wenxindongwork: can we reduce branching by always supplying pcp.kv_new_starts and pcp.kv_token_order regardless of single or multi request

tpu_inference/layers/common/cp_attention.py
```diff
@@ -337,20 +343,59 @@ def to_token_order(x):  # rank-order chunks -> global token order
...
+                # global cache length.  q_pos_offsets, the one per-rank
+                # array, is not passed here.
+                cu_cache = pcp_cu_q_lens_local[0][0::2]
+                kv_lens_cache = kv_lens_local[0::2]
+                kv_cache_lens_cache = kv_cache_lens_local[0::2]
+                pps = page_indices_local.shape[0] // kv_lens_local.shape[0]
+                page_indices_cache = page_indices_local.reshape(
+                    -1, pps)[0::2].reshape(-1)
+                distribution_cache = jnp.zeros_like(
+                    distribution_local).at[2].set(distribution_local[2] // 2)
```
wenxindongwork, line 355: is this branching really necessary? feels like the multi-request branch covers the single request case.

tpu_inference/runner/tpu_runner.py
```diff
@@ -1048,6 +1049,12 @@ def _init_inputs(self) -> None:
         # The total number of requests is dp_size * max_num_seqs
         self.max_num_reqs = max(self.dp_size * scheduler_config.max_num_seqs,
                                 MIN_NUM_SEQS)
+        # PCP presents each request to the attention kernel as TWO fused seqs
+        # (its head chunk and its tail chunk), so the attention metadata
+        # buffers need twice the slots. Sampling/logits stay per-request.
+        pcp_size = self.vllm_config.sharding_config.prefill_cp_size
+        self.attn_max_num_seqs = self.max_num_reqs * (2 if pcp_size > 1 else 1)
+        self._pcp_multireq_seen = 0  # multi-request batches logged so far
```
wenxindongwork: is this for debugging purpose?

tpu_inference/runner/tpu_runner.py
```diff
@@ -2856,66 +2912,134 @@ def _prepare_inputs(self, scheduler_output: "VllmSchedulerOutput"):
...
+            # Announce genuinely concurrent PCP batches (the first 200): a
+            # multi-request build that the scheduler only ever feeds one
+            # request at a time is indistinguishable in every benchmark
+            # number otherwise.
+            if num_pcp_reqs > 1 and self._pcp_multireq_seen < 200:
+                logger.info(
+                    "PCP multi-request batch #%d: %d requests, "
+                    "scheduled tokens %s, computed %s",
+                    self._pcp_multireq_seen, num_pcp_reqs, counts, computed)
+                self._pcp_multireq_seen += 1
```
wenxindongwork: not sure if this is leftover from debugging

wenxindongwork (commented):
left some nit comments, thank you for the great work! 

Could we also cut down the amount of new comments this PR introduces, personally l feel like AI over explains a lot of times.

could we also measure whether the all-gather + permutation is overlapped by the cache phase? If permutation takes too much time, we should consider using ring attn for the current phase.

tpu_inference/layers/common/cp_attention.py
bhuvanpkaruturi: Done. Make `kv_new_starts_local, kv_token_order_local` mandatory for both single/multi-req paths and removed branching.

tpu_inference/layers/common/cp_attention.py
bhuvanpkaruturi: Great suggestion. Supplied the metadata to both paths and removed branching

tpu_inference/runner/tpu_runner.py
bhuvanpkaruturi: Ah my bad! Yes this was for debugging. Removed it now.

tpu_inference/runner/tpu_runner.py
bhuvanpkaruturi: Removed it now. Thanks.

tpu_inference/layers/common/cp_attention.py
bhuvanpkaruturi: Both differ in functionality only when handling clipped tails and `interface tests`, `pcp_vs_tp_benchmark.py` still pass single req metadata with per-rank clipped tails (`[0,C,C+actual_tail]`, instead of `[0,C,2C]`), and the general construction would then give ranks different cache-phase schedules, which the ring's lock-step cannot tolerate. In the follow up PR that aligns the chunk size with page size (block_size), this branch will be deleted and two call sites will be changed to full-length tails.

bhuvanpkaruturi:
> left some nit comments, thank you for the great work!
> 
> Could we also cut down the amount of new comments this PR introduces, personally l feel like AI over explains a lot of times.
> 
> could we also measure whether the all-gather + permutation is overlapped by the cache phase? If permutation takes too much time, we should consider using ring attn for the current phase.

Thanks for review. I have reduced the comment verbosity.

Regarding the overlap of all-gather + permutation with cache phase kernel, no, there are not overlapped. Off the two ops (all-gather, permutation), permutation takes the most time at ~135 us vs ~27us for all-gather (measured with PCP=2). The follow-up PR with page-aligned chunk removes the HBM-to-HBM permutation entirely. The remaining 27us for all-gather is negligible compared to the ~4ms kernel time. I will consider ring attention for the current phase as an optimization on top of the page-aligned chunks if it improves performance meaningfully.

wenxindongwork (approved):

jrplatin (approved):

PR #3427, opened 2026-08-19
[layers/common] Add dynamic activation padding to sharded_matmul for unaligned token sequences

amitkumar307d:
## Summary
When running `sharded_matmul` under Attention Data Parallelism (`enable_dp_attention: true`) or Prefill Context Parallelism (`pcp`), activations are partitioned along the sequence length dimension (Axis 0). If the token count (`x.shape[0]`) is not evenly divisible by the product of the active attention data mesh axes (e.g. dynamic image resolutions or unpadded multimodal prompts), JAX throws `IndivisibleError`.

```bash
(EngineCore pid=756) ERROR 08-18 11:04:56 [core.py:1343] jax._src.sharding.IndivisibleError: One of pjit outputs with pytree key path result was given the sharding NamedSharding(mesh=Mesh('data': 1, 'attn_dp': 8, 'attn_dp_expert': 1, 'expert': 1, 'model': 1, 'dcp': 1, 'pcp': 1, axis_types=(Auto, Auto, Auto, Auto, Auto, Auto, Auto)), spec=P(('data', 'pcp', 'attn_dp', 'attn_dp_expert'), None), memory_kind=device) implies that array axis 0 is partitioned 8 times, but the dimension size is 4 (full shape: (4, 4608), per-dimension tiling factors: (8, 1) should evenly divide the shape). This error occurs at source:  /workspace/tpu_inference/tpu_inference/layers/common/linear.py:112:8 (sharded_matmul)
```

This PR adds dynamic zero-padding to `sharded_matmul` along Axis 0 before `shard_map` execution, and slices the output back to the original unpadded sequence length before returning.

## Changes
- Compute the required partition divisor based on active `x_spec[0]` mesh axes present in `mesh.shape` (`ShardingAxisName.ATTN_DATA`).
- If `x.shape[0] % divisor != 0`, dynamically zero-pad `x` along Axis 0 to the nearest multiple of `divisor`.
- Slice `out[:-pad_len]` after `shard_map` to restore the exact original sequence length.
- Add unit tests covering:
  - Divisor computation and padding calculation for DP attention and hybrid PCP+DP attention meshes.
  - Numerical equivalence for `sharded_matmul` across various unaligned sequence lengths (`[5, 7, 13, 17, 33, 634]`).

## Testing
- Unit tests in `tests/layers/common/test_sharded_matmul_padding.py`.
- Regression check in `tests/layers/common/test_sharding.py`.

# Checklist

### DIFF
--- a/tpu_inference/layers/common/linear.py
+++ b/tpu_inference/layers/common/linear.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import math
+
 import jax
 import jax.numpy as jnp
 from jax.sharding import Mesh, NamedSharding
@@ -32,8 +34,7 @@ def xla_quantized_matmul(
     w_scale: jax.Array,
     quantize_activation=True,
 ) -> jax.Array:
-    """
-    Reference (pure JAX) implementation of the quantized matmul kernel below.
+    """Reference (pure JAX) implementation of the quantized matmul kernel below.
 
     Args:
         x:  Activation.
@@ -55,15 +56,15 @@ def xla_quantized_matmul(
 
         w_q_reshaped = w_q.reshape(in_blocks, block_size_in, out_blocks,
                                    block_size_out)
-        w_q = (w_q_reshaped.astype(jnp.float32) *
-               w_scale[:, jnp.newaxis, :, jnp.newaxis]).reshape(
-                   in_features, out_features).astype(x.dtype)
+        w_q = ((w_q_reshaped.astype(jnp.float32) *
+                w_scale[:, jnp.newaxis, :, jnp.newaxis]).reshape(
+                    in_features, out_features).astype(x.dtype))
 
         # in this case, we don't want to quantize the activations
         quantize_activation = False
         logger.info_once(
-            "Skipping activation quantization due to weight requantization being disabled."
-        )
+            "Skipping activation quantization due to weight requantization"
+            " being disabled.")
 
     if quantize_activation:
         acc_dtype = jnp.float32
@@ -90,6 +91,40 @@ def xla_quantized_matmul(
     return out.astype(x.dtype)
 
 
+def _pad_sharded_activation(
+    x: jax.Array,
+    mesh: Mesh | None,
+    axis_spec,
+    axis_idx: int = 0,
+) -> tuple[jax.Array, int]:
+    """Pads activation tensor `x` along `axis_idx` if unaligned to mesh sharding divisor."""
+    orig_len = x.shape[axis_idx]
+    if mesh is not None and hasattr(mesh, "shape") and orig_len > 0:
+        axes = (axis_spec if isinstance(axis_spec, (tuple, list, set)) else
+                (axis_spec, ))
+        divisor = math.prod(mesh.shape.get(ax, 1) for ax in axes if ax)
+        pad_len = -orig_len % divisor
+        if pad_len > 0:
+            pad_config = [(0, pad_len) if i == axis_idx else (0, 0)
+                          for i in range(x.ndim)]
+            x = jnp.pad(x, pad_config)
+    return x, orig_len
+
+
+def _unpad_sharded_activation(
+    out: jax.Array,
+    orig_len: int,
+    axis_idx: int = 0,
+) -> jax.Array:
+    """Unpads output tensor along `axis_idx` back to `orig_len`."""
+    if out.shape[axis_idx] == orig_len:
+        return out
+    slice_spec = tuple(
+        slice(0, orig_len) if i == axis_idx else slice(None)
+        for i in range(out.ndim))
+    return out[slice_spec]
+
+
 def sharded_matmul(x: jax.Array,
                    w: jax.Array,
                    weight_sharding: P | NamedSharding,
@@ -109,6 +144,9 @@ def sharded_matmul(x: jax.Array,
     # x may have extra leading batch dims.
     batch_dims = (None, ) * (x.ndim - 2)
     x_spec = P(ShardingAxisName.ATTN_DATA, *batch_dims, in_axis)
+
+    x, orig_len = _pad_sharded_activation(x, mesh, x_spec[0], axis_idx=0)
+
     x = jax.lax.with_sharding_constraint(
         x,
         NamedSharding(mesh, x_spec) if mesh else x_spec)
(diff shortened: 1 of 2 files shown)


wenxindongwork:
why are we performing this padding logic inside the layer instead of tpu_runner.py's `prepare_inputs`?
`sharded_matmul` may not be the only layer that requires this input padding.

amitkumar307d:
> why are we performing this padding logic inside the layer instead of tpu_runner.py's `prepare_inputs`? `sharded_matmul` may not be the only layer that requires this input padding.

Thanks @wenxindongwork for the review! I think in multimodal pipelines, the vision tower extracts and projects image/video features inside `multimodal_manager.py` before `tpu_runner` and the scheduler ever run. Because these dynamic vision patches go directly into `sharded_matmul` without passing through `tpu_runner.prepare_inputs`, padding inside `sharded_matmul` is needed to prevent `shard_map` crashes. For standard token batches that are already aligned by `tpu_runner`, this logic is a zero-overhead no-op (`pad_len = 0`).

kwang3939:
Hi @amitkumar307d , is this `sharded_matmul` the only place that needs this fix? I discussed with @wenxindongwork earlier, and our concern is this kind of layer level fix would be needed in multiple places.

But as I mentioned I earlier, I also have some concerns on scheduler level fix:  
- We need to update two paths: both `multimodal_manager.py` and `mm_encoder_jit_manager.py` to pass in dp/pcp size. Padding might break the budgets from `mm_encoder_jit_manager.py`, which are preset before precompilation.
- Also, I think both within vision encoder the vision patches and after vision encoder the generated vision tokens merged with text tokens would trigger this issue.

amitkumar307d:
@kwang3939
1. Fix needed at below mentioned places
   - `linear.py` is the only place in the repo where `shard_map` is called on dynamic activation sequences. Within `linear.py`, this applies to:
   - `sharded_matmul` (done in this PR)
   - `sharded_quantized_matmul` and `sharded_quantized_batched_matmul` (for quantized/einsum paths)
   - Other layers (RMSNorm, activations, attention) rely on implicit `NamedSharding` or pre-allocated KV tables and do not have this `shard_map` partition constraint.
2. Regarding the layer-level approach handling the scheduler & budget concerns:
   - Precompilation budgets is preserved because `sharded_matmul` slices the output back to `out[:orig_len]` before returning, tensor shapes remain exact externally and do not alter or break preset budgets in `mm_encoder_jit_manager.py`.
   - I think it transparently protects both raw vision patches (inside the encoder) and merged tokens (post-encoder) without requiring intrusive changes across `multimodal_manager.py` or scheduler plumbing.

Let me go ahead and extract a shared `_pad_to_mesh_sharding` helper in `linear.py` and apply it to `sharded_quantized_matmul` and `sharded_quantized_batched_matmul` as well to ensure all linear primitives are fully covered.

tpu_inference/layers/common/linear.py
```diff
@@ -90,6 +91,40 @@ def xla_quantized_matmul(
...
+def _pad_sharded_activation(
+    x: jax.Array,
+    mesh: Mesh | None,
+    axis_spec,
+    axis_idx: int = 0,
+) -> tuple[jax.Array, int]:
+    """Pads activation tensor `x` along `axis_idx` if unaligned to mesh sharding divisor."""
+    orig_len = x.shape[axis_idx]
+    if mesh is not None and hasattr(mesh, "shape") and orig_len > 0:
+        axes = (axis_spec if isinstance(axis_spec, (tuple, list, set)) else
```
kwang3939: Could you use existing `get_mesh_shape_product` instead so `import math` can be avoided?

tpu_inference/layers/common/linear.py
```diff
@@ -90,6 +91,40 @@ def xla_quantized_matmul(
...
+                (axis_spec, ))
+        divisor = math.prod(mesh.shape.get(ax, 1) for ax in axes if ax)
+        pad_len = -orig_len % divisor
+        if pad_len > 0:
+            pad_config = [(0, pad_len) if i == axis_idx else (0, 0)
+                          for i in range(x.ndim)]
+            x = jnp.pad(x, pad_config)
+    return x, orig_len
```
kwang3939, line 115: it seems `mesh` is None case is left unhandled here?

tpu_inference/layers/common/linear.py
```diff
@@ -90,6 +91,40 @@ def xla_quantized_matmul(
...
+
+
+def _unpad_sharded_activation(
```
kwang3939: I think you may use `jax.lax.slice_in_dim` instead of adding a helper func?

tpu_inference/layers/common/linear.py
```diff
@@ -366,18 +429,23 @@ def wrapper(x, w_q, w_s):
...
-    return jax.shard_map(
+    out = jax.shard_map(
         wrapper,
         mesh=mesh,
         in_specs=(x_sharding, weight_spec, scale_sharding),
         out_specs=out_sharding,
         check_vma=False,
     )(x, w_q, w_s)
+
+    out_axis_str = einsum_str.replace(" ", "").split("->")[1]
```
kwang3939: Could you reuse line 366 `lhs, _ = einsum_str.replace(" ", "").split("->")`?

kwang3939:
@amitkumar307d , thanks for the reply! I just double checked and agree that for layer level fix these should be the places that need to be addressed and I also prefer layer level fix.

One correction: I checked the merging funcs on vision tokens and text tokens, which I noticed that has already gone through `prepare_inputs` so the part can really trigger this issue is in vision encoder only. 

Also I left some comments and could you address them and also revert some unrelated format changes like `_parse_einsum_dims`? Thanks!

tpu_inference/layers/common/linear.py
amitkumar307d: Yes. Done.

tpu_inference/layers/common/linear.py
amitkumar307d: Thanks for the suggestion. Done.

tpu_inference/layers/common/linear.py
amitkumar307d: Definitely. Done.

tpu_inference/layers/common/linear.py
amitkumar307d: Retrieving the mesh if present, before calculateing the correct padding divisor. If no mesh exists, it gracefully leaves `x` untouched without crashing `_pad_sharded_activation`.

amitkumar307d:
@kwang3939 Thanks for the confirmation. 

Resolved the comments. Kindly check.

kwang3939:
Thanks for the update! One remaining issue in `_pad_sharded_activation`: I noticed it seems using `jax.sharding.get_mesh()` under jit would fall to exception and skip padding. so could you avoid using `try/except` and `hasattr` and use `jax.sharding.get_abstract_mesh()` instead?
The current test passes likely due to `self.mesh.shape.get(ShardingAxisName.ATTN_DATA,1)` returns 1 as `ATTN_DATA` is a tuple.

amitkumar307d:
@kwang3939 Simplified the logic and using get_abstract_mesh for jax.jit comptibility.

kwang3939 (approved):

lk-chen (approved):

PR #3435, opened 2026-08-20
[Fix] Reduce shared-expert output over in-group TP under attention DP

wenxindongwork:
## Problem

With attention DP and in-group TP > 1 (e.g. `--tensor-parallel-size 8` + `attn_dp_size=4` → mesh `model=2`), Qwen3.5-MoE produces degraded / garbled output. `attn_dp_size == tp` (model=1) and plain TP are fine. 

## Root cause

Qwen3.5-MoE's shared expert is built with `reduce_results=False`, so the shared-expert output is a partial sum over the in-group tp axis. Under attention DP the routed experts reduces and scatters its own output, but the shared experts do not. Net effect: `result = shared_partial + routed_full` — the shared expert contributes 1/`model` of its value.

## Fix

Reduce the shared-expert output over `ShardingAxisName.ATTN_HEAD` (the in-group tp axis) under attention DP instead of skipping it. 

## Validation

Qwen3.5-35B-A3B-FP8, tp=8, EP, gsm8k 5-shot, v7x-8:

| config | mesh `model` | before | after |
|---|---|---|---|
| plain tp=8 | 8 | 96.0% | — |
| attn_dp=8 | 1 | 96.1% | — (reduce is a no-op) |
| attn_dp=4 | 2 | 86.1% | **96.3%** |
| attn_dp=2 | 4 | 23.9% | **96.2%** |

Baseline (main branch): Qwen3.5-4B (dense: GDN + full attention, no MoE) scores 91.1% at both attn_dp=4 and attn_dp=8, and Qwen3-30B-A3B-FP8 (MoE without a shared expert, no GDN) scores 91.7% / 91.6% at attn_dp=4 / 8

### DIFF
--- a/tpu_inference/layers/vllm/custom_ops/fused_moe.py
+++ b/tpu_inference/layers/vllm/custom_ops/fused_moe.py
@@ -41,8 +41,16 @@ def _get_mesh() -> Mesh | None:
 
 def _all_reduce_over_tp(t: torch.Tensor,
                         mesh: Mesh,
-                        axis_name=ShardingAxisName.MLP_TENSOR) -> torch.Tensor:
-    """All-reduce an unreduced local sum over the TP axis."""
+                        axis_name=None) -> torch.Tensor:
+    """All-reduce an unreduced local sum over the TP axis.
+
+    ``axis_name`` is resolved inside the function on purpose: ``ShardingAxisName``
+    is lazily initialized from the environment on first attribute access, so
+    reading it in a default argument (at import time, before NEW_MODEL_DESIGN is
+    set) would latch the 2D axis set for the whole process.
+    """
+    if axis_name is None:
+        axis_name = ShardingAxisName.MLP_TENSOR
     spec = P(ShardingAxisName.ATTN_DATA, None)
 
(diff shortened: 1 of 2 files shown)


jrplatin (approved):
LGTM!! Thanks for this!

tpu_inference/layers/vllm/custom_ops/fused_moe.py
```diff
@@ -39,13 +39,15 @@ def _get_mesh() -> Mesh | None:
         return None
 
 
-def _all_reduce_over_tp(t: torch.Tensor, mesh: Mesh) -> torch.Tensor:
+def _all_reduce_over_tp(t: torch.Tensor,
+                        mesh: Mesh,
+                        axis_name=ShardingAxisName.MLP_TENSOR) -> torch.Tensor:
```
kyuyeunk: huh, was this issue always there? or was this due to some recent change?

wenxindongwork: i'm not sure, we discovered it this week

kyuyeunk (approved):

PR #3439, opened 2026-08-21
[Jax][Qwix] Quantize large NNX models submodule-by-submodule to avoid a peak-HBM OOM

shungcp:
### Description

Online Qwix quantization (`additional_config={"quantization": "int8_default.yaml"}` etc.) OOMs on large MoE models whose bf16 weights already use a large fraction of a chip's HBM, even though the quantized model would be much smaller.

Repro: GLM-4.7-Flash (~62GB bf16, MoE, 47 layers) on v6e-4, `--tensor-parallel-size 4`. bf16 weights already use ~17GB/31GB HBM per chip before quantization starts. Both `int8_default.yaml` (w8a8) and `int8_all_modules_w_only.yaml` (w8a16) fail identically:

```
jax.errors.JaxRuntimeError: RESOURCE_EXHAUSTED: Ran out of memory on HBM, the total memory
required for HLO temporaries (33.40G) exceeds available HBM (31.24G).
```

### Root cause

`qwix_quantize_nnx_model` calls `qwix.quantize_model(model, ...)` once on the whole model inside one `jax.jit(..., donate_argnums=(0,))`. `qwix.quantize_model`'s NNX path internally does `model = nnx.clone(model)`, so the original bf16 weights and the cloned bf16-on-its-way-to-int8 weights are resident at the same time for the entire model at once — roughly 2x the model's HBM footprint.

`donate_argnums` doesn't rescue this: donation only lets XLA alias a buffer when input and output have identical shape/dtype, and quantization changes both (bf16 array → int8 array + scale array). We also tried Qwix's documented large-model pattern (`quantize_params` + an abstract `nnx.eval_shape` template) — it doesn't help either, because this all runs inside `create_jit_model`'s outer `@nnx.jit(donate_argnums=(0,))` (`model_loader.py`), so it's still one fused program with the same non-aliasable structural change, and it OOMs at essentially the same HBM figure.

### Fix

Quantize one submodule at a time (`embed_tokens`, each decoder layer, `norm`), each via its own `jax.jit(..., donate_argnums=(0,))` call, dispatched and consumed before moving to the next. This bounds peak extra HBM to one submodule instead of the whole model. Works because Qwix's PTQ interception is per-instance (swaps `type(instance)`), so an unmodified parent's `self.submodule(...)` call picks up the quantized submodule automatically.

`lm_head` stays unquantized, matching this model's existing convention.

Only the concrete-model branch (`apply_to_abstract_model=False`) in `apply_qwix_quantization` changes. The abstract-model path (pre-quantized checkpoints) is untouched.

### Tests

- Updated `tests/layers/jax/test_qwix.py::TestApplyQwixQuantization::test_quantization_applied_from_dict` — it was asserting `jax.jit` gets called once, which no longer holds; now asserts the new streaming function is dispatched.
- `pytest tests/layers/jax/test_qwix.py`: 36 passed.
- Manually verified end-to-end on GLM-4.7-Flash (v6e-4, TP=4): both w8a8 and w8a16 now load and serve correctly (HBM flat at ~17GB through all 47 layers, vs. OOM before), output matches the bf16 baseline.
- Could not run `tests/models/jax/test_deepseek_v3.py` locally (`ModuleNotFoundError: tokamax`, unrelated to this change).

### DIFF
--- a/tpu_inference/models/jax/utils/qwix/qwix_utils.py
+++ b/tpu_inference/models/jax/utils/qwix/qwix_utils.py
@@ -22,6 +22,7 @@ if TYPE_CHECKING:
 from tpu_inference import utils
 from tpu_inference.layers.common.attention_metadata import AttentionMetadata
 from tpu_inference.layers.common.sharding import ShardingAxisName
+from tpu_inference.layers.jax.pp_utils import PPMissingLayer
 from tpu_inference.logger import init_logger
 from tpu_inference.runner.kv_cache import (DEFAULT_KV_CACHE_DTYPE,
                                            create_kv_caches)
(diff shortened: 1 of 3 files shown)


tpu_inference/models/jax/utils/qwix/qwix_utils.py
```diff
@@ -217,6 +218,155 @@ def qwix_quantize_nnx_model(model: nnx.Module, qwix_config: List[dict],
     return model
```
lk-chen, line 218: is `qwix_quantize_nnx_model` used anywhere after your fix? shall we delete it ?

lk-chen (commented):
mostly LGTM

tpu_inference/models/jax/utils/qwix/qwix_utils.py
shungcp: No, qwix_quantize_nnx_model is still live — it's used by llama4/gpt-oss in the apply_to_abstract_model=True branch. My fix only replaced the call in the apply_to_abstract_model=False (concrete model) branch, since that's the path that actually OOMs. Both paths need to coexist for now.

PR #2251, opened 2026-04-15
[torchax] Add ViT sdpa op

muskansh-google:
# Description

This PR implements the vllm_vit_sdpa function in scaled_dot_product_attention.py. This provides a custom JAX implementation of vLLM Scaled Dot Product Attention (SDPA) specifically for ViTs. `register_torchax_dispatch_op` intercepts the vLLM ViT SDPA wrapper and replaces it with custom JAX implementation

FIXES: #1850

# Tests

Manually tested accuracy and run by serving Qwen3-VL model on Torchax

# Checklist

Before submitting this PR, please make sure:
- I have performed a self-review of my code.
- I have necessary comments in my code, particularly in hard-to-understand areas.
- I have made or will make corresponding changes to any relevant documentation.

PR #3442, opened 2026-08-22
[Fix] Avoid O(S^2) attention bias allocation in ViT SDPA when attn_mask is None

amitkumar307d:
## Problem

`scaled_dot_product_attention` allocated a dense
`(batch, num_heads, q_seq_len, kv_seq_len)` float32 `attention_bias`
unconditionally — including when `attn_mask is None`, where its only remaining
purpose was masking the 128-alignment padding.

That spends O(S^2) memory to express an O(S) constraint. On long vision/video
sequences it dominates: for a flattened ViT sequence the bias alone is
`num_heads * S^2 * 4` bytes, which at `num_heads=16, S=30976` works out to
~61 GB. (That figure is arithmetic from the tensor shape, not a measurement.)

## Fix

When `attn_mask is None`, pass segment ids instead of a bias. Real tokens share
segment 0 and the alignment padding lands in the trailing segment, so real
queries cannot attend to padded keys — the same constraint the padded bias
encoded, in O(S) memory.

This reuses `segment_ids_from_cu_seqlens`, matching the construction already
used by `vllm_vit_sdpa` in this file, so both ops now build segment ids the
same way.

The explicit-`attn_mask` path is unchanged and still builds the bias. Both
paths continue to pass the `vmem_limit_bytes` added in #3407.

## Why the kernel no longer needs changing

An earlier revision of this PR raised the kernel's VMEM ceiling and added
model-specific logic to `qwen3_vl_patcher.py`. Both are reverted — those were
treating the symptom. Once the dense bias is gone, the allocation that caused
the OOM is gone, so `kernel.py` and `qwen3_vl_patcher.py` are untouched
(zero diff against main).

## Tests

Adds `TestSdpaSkipsBiasWithoutMask`:

- `test_unaligned_seq_len_uses_segment_ids_not_bias` — padding present →
  `use_attention_bias=False` and segment ids mark the padded tail.
- `test_aligned_seq_len_needs_no_segment_ids` — already 128-aligned → no
  segment ids needed.
- `test_explicit_attn_mask_still_uses_bias` — the mask path is unaffected.
- `test_segment_ids_match_bias_numerically` — the segment-id path and the old
  bias path produce the same output.

Full file: 9 passed, 3 skipped (the 3 skips need >=2 devices for the DP mesh).

## Checklist


tests/models/vllm/experimental/test_vision_flash_attn.py
kyuyeunk, line 1: Please remove model specific logic from kernel (even if it's just a test)

tpu_inference/kernels/flash_attention/kernel.py
```diff
@@ -332,8 +332,10 @@ def flash_attention(
             estimated_vmem = calculate_vmem_usage_bytes(
                 block_sizes, q.dtype, k.dtype, d_model, kv_seq_len, ab,
                 segment_ids)
-            vmem_limit = pltpu.get_tpu_info(
+            raw_limit = pltpu.get_tpu_info(
             ).vmem_capacity_bytes if vmem_limit_bytes is None else vmem_limit_bytes
+            # Cap vmem_limit to physical per-core Scoped VMEM budget (32 MB on TPU v6e/v7x)
+            vmem_limit = min(raw_limit, 32 * 1024 * 1024)
```
kyuyeunk: Why cap it?

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -372,7 +497,18 @@ def _patched_flatten_embeddings(embeddings: NestedTensors) -> torch.Tensor:
...
 
 def apply_qwen3_vl_patches(vllm_model):
-    """Apply Qwen3-VL specific patches for stateless Deepstack support."""
+    """Apply Qwen3-VL specific patches for stateless Deepstack support and Vision Flash Attention."""
+    # Patch SDPA specifically in ViT attention wrappers and Qwen vision modules
+    # to redirect large sequence video attention to encoder_only_flash_attention.
+    try:
+        from vllm.v1.attention.ops import vit_attn_wrappers
+        vit_attn_wrappers.F.scaled_dot_product_attention = _flash_attn_sdpa
+    except (ImportError, AttributeError):
```
kyuyeunk: This is really bad programming pattern.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -44,20 +44,145 @@
...
 from tpu_inference.distributed.jax_parallel_state import \
     get_pp_group as jax_get_pp_group
+from tpu_inference.kernels.flash_attention.kernel import \
+    encoder_only_flash_attention
 from tpu_inference.logger import init_logger
 
 logger = init_logger(__name__)
 
+# Architectures considered part of the Qwen3-VL / Qwen3.5 family
+QWEN3_VL_ARCHS = set()
```
kyuyeunk: Ummm I feel like I don't have to call out that this is a horrible design pattern?

tests/models/vllm/experimental/test_vision_flash_attn.py
amitkumar307d: Removed test_vision_flash_attn.py. Replaced it with dedicated unit tests in tests/kernels/flash_attention_kernel_test.py to directly test dynamic block sizing and the VMEM guard.

tpu_inference/kernels/flash_attention/kernel.py
amitkumar307d: The 32MB originated from XLA's compile-time limit (CompileTimeScopedVmemOom: budget: 33554432 bytes), which is the physical on-chip VMEM per core on TPU v5e/v6e/v7x. 

I understand the concern. Switched to pltpu.get_tpu_info().vmem_capacity_bytes to dynamically query the active TPU's actual capacity without hardcoding.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: Replaced it with an immutable static set _SUPPORTED_QWEN_VL_ARCHS.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: Noted. Updated to import vit_attn_wrappers at the top level. Also added an idempotency guard to apply_qwen3_vl_patches.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -44,20 +44,129 @@
...
+        is_self_attn = False
+
+    # Only engage flash attention path if:
+    # 1. Total elements exceed large attention threshold
+    # 2. Equal sequence lengths and matching head counts (standard encoder self-attention)
+    # 3. No dropout, non-causal (standard for vision encoder self-attention)
+    # 4. No arbitrary attention mask (standard for full bidirectional vision attention)
+    if (total_elements >= LARGE_ATTN_ELEMENT_THRESHOLD and is_self_attn
+            and dropout_p == 0.0 and not is_causal and attn_mask is None):
+        try:
```
kyuyeunk: why is this needed?

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -44,20 +44,129 @@
...
+            logger.info_once(
+                f"Redirecting large vision attention ({total_elements} elements) to encoder_only_flash_attention"
+            )
+            # Obtain JAX views of query, key, value
+            q_jax = jax_view(query)
+            k_jax = jax_view(key)
+            v_jax = jax_view(value)
+
+            if q_jax.ndim == 4:
```
kyuyeunk: can you try to remove this if/else statement? this is easier to read.

```
should_process = q_jax.ndim == 4
if should_process:
  # preprocess

# compute

if should_process:
  # postprocess
```

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -44,20 +44,129 @@
...
+                                                       sm_scale=scale)
+                # Reshape back: [S, H, D] -> [H, S, D]
+                out_jax = jnp.transpose(out_jax, (1, 0, 2))
+                return torch_view(out_jax)
+        except Exception as e:
+            logger.warning(
+                f"Failed to execute encoder_only_flash_attention ({e}), falling back to standard SDPA"
+            )
+
+    return _orig_sdpa(query,
```
kyuyeunk: this is difficult to read. there are 3 exit points of this function. try to see if you can only have one.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -371,8 +480,22 @@ def _patched_flatten_embeddings(embeddings: NestedTensors) -> torch.Tensor:
     return torch.cat(tuple(_patched_flatten_embeddings(t) for t in embeddings))
 
 
-def apply_qwen3_vl_patches(vllm_model):
-    """Apply Qwen3-VL specific patches for stateless Deepstack support."""
+def apply_qwen3_vl_patches(vllm_model: nn.Module):
+    """Apply Qwen3-VL specific patches for stateless Deepstack support and Vision Flash Attention."""
+    # Idempotency guard: prevent repeatedly wrapping vllm_model methods
+    # (such as forward and embed_input_ids) if patches are invoked multiple times.
+    if getattr(vllm_model, "_qwen3_vl_patches_applied", False):
```
kyuyeunk: don't do `getattr`

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
```diff
@@ -429,9 +552,12 @@ def apply_qwen3_vl_patches(vllm_model):
...
 
 
-def is_qwen3_vl(vllm_model) -> bool:
-    """Check if the given vLLM model is of architecture Qwen3VLForConditionalGeneration."""
-    return isinstance(vllm_model, Qwen3VLForConditionalGeneration)
+def is_qwen3_vl(vllm_model: nn.Module) -> bool:
+    """Check if the given vLLM model is of architecture Qwen3-VL or Qwen3.5."""
+    architectures = (getattr(getattr(vllm_model, "config", None),
+                             "architectures", ()) or ())
+    return (any(arch in _SUPPORTED_QWEN_VL_ARCHS for arch in architectures)
```
kyuyeunk: what is wrong with original logic?

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: It isn't — this was inside `_flash_attn_sdpa`, which I've removed entirely.
The whole redirect was redundant: torchax already intercepts
`torch.nn.functional.scaled_dot_product_attention` globally via `@register_function`
in `tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py`, so ViT SDPA was
already being routed to the flash attention path without any patching here.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: I'd changed it to an architecture-string check thinking the Qwen3.5 arches needed
explicit handling. They don't: `Qwen3_5ForConditionalGeneration` and
`Qwen3_5MoeForConditionalGeneration` both subclass `Qwen3VLForConditionalGeneration`
(as does `Qwen3VLMoeForConditionalGeneration`), so the original `isinstance` already
matched all four. The frozenset was redundant, and strictly more fragile — it would
silently miss any future subclass that wasn't added to it.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: Removed. This was the `getattr(vllm_model, "_qwen3_vl_patches_applied", False)`
idempotency guard I added, which is no longer needed now that
`apply_qwen3_vl_patches` is back to its original form.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: Fair — the three exits (4D path, 3D path, `_orig_sdpa` fallback) were a symptom of the
function trying to do too much. Rather than restructure it, I removed it: the redirect
was redundant with the existing torchax `@register_function` interception.

tpu_inference/models/vllm/experimental/qwen3_vl_patcher.py
amitkumar307d: Agreed on the pattern, and thanks for spelling it out. It turned out not to be needed.

amitkumar307d:
@kyuyeunk thanks for the review. Updating this PR to focus on pending issue since VMEM issue has been taken care of in another PR. Here's why the patcher changes / monkey-patching were reverted:

- Redundant — torchax has intercepted F.scaled_dot_product_attention globally via @register_function since the ViT SDPA ops landed (#2251 / #2252, April), so ViT SDPA was already on the flash-attention path; the redirect was a no-op from the start.
- Wrong layer — the OOM was a dense tensor allocated one level up, not routing or kernel block-sizing. Raising the VMEM ceiling moved the ceiling instead of removing the allocation.
- Once the redirect proved to be a no-op, everything attached to it went too: the 4D/3D reshaping, the three exit points, the getattr guard, _SUPPORTED_QWEN_VL_ARCHS.
- On the arch check — you were right. All four Qwen3-VL / 3.5 arches subclass Qwen3VLForConditionalGeneration, so isinstance already matched; the frozenset was redundant and more fragile.

Regarding the VMEM, the cap is gone — kernel.py is zero-diff, so this PR sets no VMEM limit of its own.
The ViT limit is now owned upstream by #3407 (scoped 64 MiB, in this same file). Both branches of the rewritten function pass it, so there's no regression, and the existing test covers it.

#### Current scope: 
when attn_mask is None, pass segment ids instead of a dense (B, H, S, S) float32 bias that only masked 128-alignment padding. This op has no cu_seqlens, so unlike vllm_vit_sdpa it can't segment per image — but with attn_mask is None the only thing left to mask is that padding, so a trivial [0, S] split suffices. It's built with #3450  segment_ids_from_cu_seqlens, which gives trailing positions a dedicated id rather than folding them into the last real sequence, so both ops in this file now construct segment ids the same way.

kwang3939:
Hi @amitkumar307d, I think qwen torchax path vision encoder does not go through `scaled_dot_product_attention` here? Since vLLM calls `torch.ops.vllm.torch_sdpa_wrapper`, it uses `vllm_vit_sdpa` already with segment ids.  
The earlier OOM issue should be addressed by #3407 and #3450. So I'm wondering what real use case this PR is trying to address?

amitkumar307d:
Hi @kwang3939, you're completely right about Qwen — Qwen's vision encoder routes through `MMEncoderAttention` -> `vit_torch_sdpa_wrapper` -> `torch.ops.vllm.torch_sdpa_wrapper`, so it dispatches directly to `vllm_vit_sdpa` (which already builds segment IDs) and never hits `scaled_dot_product_attention`. The Qwen-specific OOM was indeed resolved by #3407 and #3450.
The reason this PR was updated to this scope is:
1. **Origin:** This PR originally started from the Qwen video OOM investigation, but Kims previous comments prompted me to re-traced the execution and confirmed that the patcher redirect was a no-op and the VMEM cap was resolved upstream by #3407, I reverted the patcher and kernel changes.
2. **The Latent Issue in `scaled_dot_product_attention`:** While auditing `scaled_dot_product_attention.py`, I was noticed that while `vllm_vit_sdpa` had already adopted segment IDs to avoid dense bias materialization, its companion op `scaled_dot_product_attention` (the TorchAX handler for native `F.scaled_dot_product_attention`) was still unconditionally allocating a dense `(batch, num_heads, q_seq_len, kv_seq_len)` float32 `attention_bias` whenever 128-alignment padding was needed — even when `attn_mask is None`.
3. **Real Use Case:** Models in vLLM that do not use `MMEncoderAttention` and instead invoke standard PyTorch `F.scaled_dot_product_attention` (e.g. Pixtral, Molmo2, DeepEncoder, Phi4-MM, Granite Speech, Cheers, or custom ViTs) run through this function. For those models, long sequences with `attn_mask=None` would allocate an $O(S^2)$ float32 bias (~61 GB for $S=30,976$) solely to express the 128-byte padding constraint.
This PR brings `scaled_dot_product_attention` into parity with `vllm_vit_sdpa` by using 1D `SegmentIds` to handle 128-alignment padding in $O(S)$ memory when no mask is provided.
That said, if the team prefers to leave `scaled_dot_product_attention` as-is until one of those specific models explicitly raises an issue, I'm also happy to close this PR. Let me know what you think!

amitkumar307d:
This PR has become redundant. Closing it.

PR #3449, opened 2026-08-23
[continue_decode] Defer KV block free under async scheduling

gutianyu-google:
## Summary

Fixes generation corruption when `continue_decode` runs with `async_scheduling=True`. On a decode-heavy workload the EOS-termination rate roughly halves and 30%+ of sequences run to their token limit (measured 55% truncated-at-cap vs 24% baseline at inference step 0 with byte-identical prompts), while every host-side invariant stays correct. The identical window under `async_scheduling=False` is clean.

Companion to #3401, which documents this interaction and works around it by requiring `async_scheduling=False`. This PR removes that constraint.

## Root cause: a KV-block write-after-free

A `continue_decode` step runs up to `max_decode_steps` (N) decode iterations on-device in a single scheduling step. Under async scheduling the host overlaps output processing of the previous window with dispatch of the next. A request that samples EOS **mid-window** is still carried by the runner as a masked (zombie) row for one or two more windows (retirement lag), and the fused decode loop keeps **writing** that row's KV. If the scheduler frees and reallocates the request's blocks during the overlap, the in-flight window writes into **another request's block** and corrupts its generation. In sync mode the free happens strictly between device steps, so the device never references a freed block — which is why async-off is clean.

Diagnosis: a per-window device-vs-host probe (dump `init_tokens`, `input_positions`, `seq_lens` at every fused-window start and compare with the host `input_batch`) over **88,545 request-windows** under async showed `tok_mismatch ≈ 0`, `pos_mismatch = 0`, `sl_mismatch = 0`. So the device starts every window from the correct token, position and KV length — the bookkeeping is correct and the defect is in the KV **bytes**. By elimination the channel is the KV cache contents, and the only async-specific mutation of KV storage is block free/realloc during the overlap.

This matches every symptom: corruption scales with window depth, is uniform across DP ranks, has zero preemptions, and leaves `num_computed_tokens` / `seq_lens` / per-rank layout / extraction / retirement all correct.

## Fix

vLLM already guards this exact hazard with deferred block freeing (`Scheduler._free_request_blocks`): a finished request's blocks are held out of the pool until `last_sched_seq <= processed_step_seq`, i.e. until every in-flight step that referenced the request has been processed. The machinery (`sched_step_seq` / `processed_step_seq` / `deferred_frees` / the `_drain_deferred_frees` fence) is initialized unconditionally; only the `defer_block_free` flag gates it, and it is auto-enabled solely for KV-connector consumers (`multiple_inflight_batches and is_kv_consumer`).

Continue-decode + async is the same overlapping-write hazard, widened from 1 token to N. This PR sets `defer_block_free = True` in the continue-decode scheduler patch when `async_scheduling` is on. The retirement-lag zombie is re-scheduled in the next window, which advances its `last_sched_seq`, so the existing fence holds its blocks until that window's output is processed — exactly covering the vulnerable window.

Scope: the change is confined to `patch_vllm_scheduler_for_continue_decode` and to `async_scheduling=True`; sync scheduling and non-continue-decode paths are untouched. It reuses vLLM's own deferral state machine rather than adding a new one.

## Validation

- Correctness restored end-to-end on an RL post-training rollout (Qwen3-0.6B GRPO, 2048 seqs, `max_decode_steps=128`, v7x 4x4x4, DP16×TP8, `async_scheduling=True`): step-0 truncated-at-cap and output-length distribution return to the sync/default path (validation run in progress; numbers to be filled in).
- Default path unchanged: with `async_scheduling=False` the new branch does not execute, and `defer_block_free` stays at its upstream default.

## Note on performance

On a workload where continue_decode windows are already deep (N=128), enabling async scheduling recovers ~0 additional throughput (the host-overlap it provides is already amortized ~N× by the fused window; measured 11.3 vs 11.2 µs/token). This PR is therefore a **correctness** fix — it makes async usable with continue_decode — rather than a speedup.

### DIFF
--- a/tpu_inference/core/sched/utils.py
+++ b/tpu_inference/core/sched/utils.py
@@ -71,6 +71,34 @@ def patch_vllm_scheduler_for_continue_decode():
         def patched_init(scheduler_self, vllm_config, *args, **kwargs):
             original_init(scheduler_self, vllm_config, *args, **kwargs)
 
+            # Under async scheduling, a continue-decode window runs up to N
+            # decode iterations on-device while the host overlaps output
+            # processing of the previous window. A request that samples EOS
+            # mid-window is still carried by the runner as a masked row for one
+            # or two more windows (retirement lag), and the device keeps writing
+            # that row's KV. If the scheduler frees and reallocates the
+            # request's blocks during the overlap, the in-flight window writes
+            # into another request's block and corrupts its generation, while
+            # every host-side invariant stays correct because the defect is in
+            # the KV bytes, not the bookkeeping.
+            #
+            # vLLM already guards this with deferred block freeing
+            # (Scheduler._free_request_blocks): a finished request's blocks are
+            # held until last_sched_seq <= processed_step_seq, i.e. until every
+            # in-flight step that referenced the request has been processed. It
+            # is auto-enabled only for KV-connector consumers; enable it here for
+            # continue-decode under async, the same overlapping-write hazard. The
+            # retirement-lag zombie is re-scheduled in the next window, which
+            # advances its last_sched_seq, so the fence holds its blocks until
+            # that window's output is processed.
+            try:
+                if getattr(vllm_config.scheduler_config,
+                           "async_scheduling", False) and hasattr(
+                               scheduler_self, "deferred_frees"):
+                    scheduler_self.defer_block_free = True
+            except Exception:  # pragma: no cover - best effort, never block init
+                pass
+
             additional_config = getattr(vllm_config, "additional_config", {})
             max_decode_steps = additional_config.get("max_decode_steps",
                                                      DEFAULT_MAX_DECODE_STEPS)


gutianyu-google:
**Validation update — this minimal change is NOT sufficient on its own; keeping as draft.**

I tested `defer_block_free=True` on the actual workload (async continue-decode, DP16×TP8 via the tpu_inference `DPScheduler` with 16 per-rank `AsyncScheduler` sub-schedulers). Instrumentation confirms the flag is set on every sub-scheduler:

```
16x  cls=AsyncScheduler async=True defer=True has_machinery=True
```

but generation is still corrupted — step-0 truncated-at-cap stays at ~54% (vs ~23% with async off), unchanged from before the patch.

**Why it's insufficient here:** the deferral fence (`last_sched_seq` vs `processed_step_seq`) counts each per-rank scheduler's own `schedule()` / `update_from_output()` calls. In the `DPScheduler` architecture the overlapping write that corrupts is between the **device** (running the fused N-step window) and the **host** free path, which the per-rank scheduler's step counter does not appear to capture — so the fence resolves same-step and blocks are freed immediately, as if deferral were off.

The root-cause diagnosis in the description still holds (a per-window device-vs-host probe over 88,545 request-windows shows window-start token/position/seq_len all correct under async, so the defect is in KV contents, not bookkeeping). But the correct fix needs to defer the block free until the **device window** that referenced the request completes — i.e. a device-completion fence in `DPScheduler`/runner — or eliminate the zombie KV write in the fused loop (mask masked-row KV writes to a scratch block). Both are larger than this one-liner.

Leaving this as a draft/RFC documenting the mechanism; will follow up with a device-completion-based fix. Flagging @pv97 (continue-decode) for thoughts on the right layer.

gutianyu-google:
**Update: this mechanism is falsified — the block-free deferral is not the cause. Marking this PR for closure.**

I instrumented `Scheduler._free_request_blocks` on the failing workload (async continue-decode, DPScheduler + 16 per-rank AsyncSchedulers). With `defer_block_free=True` the fence engages correctly on **every** finished request:

```
960x  will_defer=True  last_sched_seq=4 processed=3 sched=4  deferred_q growing
```

So finished requests' blocks are held through the vulnerable window exactly as intended — yet generation stays corrupted (~54% truncated). The deferral works and does not fix it, so the KV-block write-after-free is **not** the mechanism.

Root cause is elsewhere in the async continue-decode path (leading suspect: `_execute_continue_decode` reconciles the previous window via `_modify_prev_results()` and then re-runs the async on-device token substitution in `_prepare_inputs` because `_pre_async_results` is not cleared, unlike the non-continue-decode path — a possible double reseed). I'll pursue that separately and close this PR; leaving the diagnosis here for reference. Apologies for the noise, @pv97.

PR #3450, opened 2026-08-23
[Bugfix] Compile the Qwen2.5-VL vision tower on the torchax path (mm-encoder JIT padding + ViT pad-segment fixes)

QiliangCui:
## Problem

On the torchax path (`MODEL_IMPL_TYPE=vllm`) Qwen2.5-VL-7B's multimodal benchmark has never genuinely passed. Nightly 24593 / 24660:

- **tpu7x**: 0.04 req/s (bar 0.76), mean TPOT 8.3 s, mean TTFT 27 min — the whole engine crawls, not just prefill.
- **tpu6e**: EngineCore dies with `RESOURCE_EXHAUSTED` inside `vllm_vit_sdpa` → `jit__flash_attention` (96 MiB needed, 52 MiB free); the run reported `Total generated tokens: 0` and was scored PASSED until #3447.

The JAX-native model runs the same recipe at 1.18–1.29 req/s.

## Root cause

The vision tower runs **eager** under torchax: every image goes op-by-op through all 32 ViT blocks, each `vllm_vit_sdpa` call launching a standalone flash-attention program, and every engine step waits behind it. Nothing enables `compilation_config.cudagraph_mm_encoder`, so `maybe_create_mm_encoder_jit_manager` returns `None` (log: `Qwen2_5_VLForConditionalGeneration's vision tower does NOT support JIT compilation`).

vLLM's `Qwen2_5_VLForConditionalGeneration` implements `SupportsEncoderCudaGraph` (since vllm-project/vllm@f5bb701fa2, in our LKG), so the compiled route is `MMEncoderJITManager` — the same mechanism the gemma-4 26B/31B benchmarks already use. Turning it on exposed two correctness bugs in that route, which this PR fixes first:

1. **`MMEncoderJITManager._pad_to_template` ignored `EncoderCudaGraphConfig.padding_logics`** and zero-filled every replay buffer. Qwen2.5-VL registers `_pad_cumulative_seqlens_buffer` (repeat the last offset) for `cu_seqlens`/`cu_window_seqlens` and returns `cu_window_seqlens` **unpadded** at replay, so the zero tail made `lens = cu[1:] - cu[:-1]` negative and the ViT built garbage window-attention segment ids. Now the per-key logic is applied (upstream `_copy_padded_buffer` fallback), a template-shaped buffer is cast to the template dtype (stable jit signature), and an over-large replay buffer raises a `[mm_encoder_jit]`-tagged error naming key/shapes/budget.
2. **`vllm_vit_sdpa` gave budget-padding tokens the last image's segment id.** `jnp.repeat(..., total_repeat_length=L)` fills positions past `sum(lens)` with the last id; on the padded JIT path `pixel_values` is padded to the budget, so for a full batch (no trailing empty `cu_seqlens` entry) the pad rows attended together with the last image. Segment ids are now `searchsorted(cu_seqlens[1:], pos, side="right")` over the padded length, so every position ≥ `cu_seqlens[-1]` lands in the dedicated trailing segment.
3. The manager's eager fallback (`encoder_eager_forward`) now enters `enable_torch_wrap(False)` like `run_budget_forward` does; without it vllm.ir ops in the ViT norms reach torchax as opaque OpOverloads and abort.

Then the benchmark step for Qwen2.5-VL-7B gets `COMPILATION_CONFIG: '{"cudagraph_mm_encoder": true}'` (gemma-4 convention). The flag is inert for the JAX-native model, which does not implement the protocol.

## Tests

- CPU unit tests: per-key padding logic / dtype cast / oversize error in `_pad_to_template`; `_segment_ids_from_cu_seqlens` (trailing pad segment, padded cu_seqlens with empty segments, leading empty segment, jit-traced == eager) and the ids `vllm_vit_sdpa` hands the kernel.
- TPU integration test (`TestQwen25VLMMEncoderJITIntegration`): real Qwen2.5-VL-3B config with dummy weights (shrunk via `hf_overrides` to 2 ViT blocks — one window, one full attention — and 1 text layer, ~80 s for the class), comparing the budget-padded JIT output with the eager forward for a mixed-grid batch and for a full `max_batch_size` batch. The nightly accuracy job is gsm8k (text only), so this is the only place the compiled vision path is checked against eager. On unfixed code the max deviation is 0.31–0.34 of max|out| (bug 1 corrupts item 0, bug 2 corrupts the last item); fixed: 0.007.

## Evidence (tpu-inference-dev)

| Build | Code | tpu7x perf | tpu6e perf | Other |
|---|---|---|---|---|
| [931](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/931) | main + flag only | 1.18 req/s, 16384 tok, 0 failed | 0.88 req/s, 16384 tok, 0 failed, no OOM | — |
| [932](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/932) | first PR SHA | 1.91 req/s, 16384 tok, 0 failed | 1.07 req/s, 16384 tok, 0 failed | gsm8k (vllm path, TP=2) 0.732 ≥ 0.72; unit tests pass on both chips |
| [934](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/934) | first PR SHA, gemma-4 | 26B: 6.72 auto / 2.12 vllm; 31B: 4.60 auto / 1.78 vllm (all 16384 tok, 0 failed) | — | `padding_logics` side effect on gemma-4's compiled encoder — no regression (bars 2.0/1.80 auto, 1.55/1.35 vllm) |
| [935](https://buildkite.com/tpu-commons/tpu-inference-dev/builds/935) | **final SHA 96e350b22** | 1.94 req/s, 16384 tok, 0 failed, TPOT 163 ms | 1.12 req/s, 16384 tok, 0 failed, TPOT 279 ms | JAX-path control with the flag present: 1.99 req/s, no manager created (flag inert on `auto`); gsm8k vllm path 0.734; 79 unit tests pass on tpu7x (170 s, incl. `tests/models/jax/test_qwen2_5_vl.py`), 76 + 3 skipped on tpu6e |

Nightly 24593 for reference: tpu7x 0.04 req/s; tpu6e dead engine / 0 tokens.

## Notes

- Supersedes #3446 (marks the benchmark `unverified` on the vllm variant); with this change that gating is not needed.
- Pairs with #3447 (benchmark fails on 0 generated tokens / failed requests).
- Known, not addressed here: `Qwen2_5_VisionTransformer.get_rope_by_thw` is `lru_cache`d on `(t, h, w)` and is reached from both the eager fallback (inside the torchax env) and the replay-buffer prep (outside). Unreachable in production: budget-vs-eager routing in `_execute_local` is a pure function of an item's token count, i.e. of its `(t, h, w)`, so a geometry that goes eager always goes eager and its cached torchax tensors never reach the replay-buffer prep. The integration test is the only place both paths see the same geometry; it clears the cache per test and runs `execute()` before its eager reference.
- `_pad_to_template` honouring `padding_logics` is model-agnostic, so it also activates gemma-4's `pad_pixel_position_ids` (`tpu_inference/models/jax/gemma4_mm.py`): pad images in a partial batch now carry the `-1` position sentinel (same as the capture template) instead of position (0,0). SigLIP encodes images independently along the batch axis and outputs are sliced per item, so real items are unaffected; verified on the dev pipeline (build 934: gemma-4 26B/31B benchmarks on both impl variants — see Evidence).

### DIFF
--- a/tests/runner/test_mm_encoder_jit_manager.py
+++ b/tests/runner/test_mm_encoder_jit_manager.py
@@ -25,9 +25,12 @@ from torchax.interop import jax_view
 from vllm.config import (CompilationConfig, ModelConfig, MultiModalConfig,
                          VllmConfig, set_current_vllm_config)
 from vllm.distributed.parallel_state import (ensure_model_parallel_initialized,
-                                             init_distributed_environment)
+                                             init_distributed_environment,
+                                             model_parallel_is_initialized)
 from vllm.engine.arg_utils import EngineArgs
 from vllm.model_executor.model_loader import get_model as vllm_get_model
+from vllm.model_executor.models.qwen2_5_vl import \
+    _pad_cumulative_seqlens_buffer
 from vllm.v1.worker.encoder_cudagraph_defs import (
     EncoderCudaGraphCaptureInputs, EncoderCudaGraphConfig,
     EncoderCudaGraphReplayBuffers)
@@ -181,15 +184,26 @@ class TestAdapterPostprocessEncoderOutput:
 
 
 class TestPadToTemplate:
-    """Test MMEncoderJITManager._pad_to_template — zero-padding, slicing, and dtype-casting.
+    """Test MMEncoderJITManager._pad_to_template — per-key padding logic,
+    slicing, and dtype-casting.
 
     Pure torch logic — tested without a full manager init by constructing a
-    bare instance via object.__new__ and setting only budget_templates.
+    bare instance via object.__new__ and setting the few attributes the method
+    reads (budget_templates, config, max_batch_size).
     """
 
-    def _manager(self, template: dict) -> MMEncoderJITManager:
+    def _manager(self,
+                 template: dict,
+                 padding_logics: dict | None = None) -> MMEncoderJITManager:
         m = object.__new__(MMEncoderJITManager)
         m.budget_templates = {256: template}
+        m.max_batch_size = 4
+        m.config = EncoderCudaGraphConfig(
+            modalities=["image"],
+            buffer_keys=list(template),
+            out_hidden_size=64,
+            padding_logics=padding_logics or {},
+        )
         return m
 
     def test_general_case_zeros_then_copies(self):
@@ -230,6 +244,66 @@ class TestPadToTemplate:
         result = self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
         assert result["pv"].dtype == torch.bfloat16
         assert torch.all(result["pv"][:4] == 1.0)
+
+    def test_padding_logic_from_config_is_used(self):
+        """A key with a registered padding logic must use it instead of the
+        zero-fill default.
+
+        Qwen2.5-VL registers `_pad_cumulative_seqlens_buffer` for cu_seqlens /
+        cu_window_seqlens and leaves cu_window_seqlens UNPADDED at replay time.
+        Zero-filling the tail makes `lens = cu[1:] - cu[:-1]` go negative and
+        the ViT attention builds garbage segment ids, so the tail must repeat
+        the last cumulative offset (= empty trailing sequences) instead.
+        """
+        tmpl = torch.zeros(6, dtype=torch.int32)
+        src = torch.tensor([0, 4, 10], dtype=torch.int32)
+        manager = self._manager({"cu_window_seqlens": tmpl},
+                                padding_logics={
+                                    "cu_window_seqlens":
+                                    _pad_cumulative_seqlens_buffer
+                                })
+        result = manager._pad_to_template({"cu_window_seqlens": src}, 256)
+        torch.testing.assert_close(
+            result["cu_window_seqlens"],
+            torch.tensor([0, 4, 10, 10, 10, 10], dtype=torch.int32))
+
+    def test_no_padding_logic_still_zero_pads(self):
+        """Keys without a registered logic keep the upstream default: zero the
+        buffer, slice-copy src onto its head."""
+        tmpl = torch.zeros(6, dtype=torch.int32)
+        src = torch.tensor([1, 2, 3], dtype=torch.int32)
+        result = self._manager({
+            "window_index": tmpl
+        })._pad_to_template({"window_index": src}, 256)
+        torch.testing.assert_close(
+            result["window_index"],
+            torch.tensor([1, 2, 3, 0, 0, 0], dtype=torch.int32))
+
+    def test_template_shaped_src_is_cast_to_template_dtype(self):
+        """src already template-shaped but a different dtype: cast, otherwise
+        the jit signature changes and every request recompiles."""
+        tmpl = torch.zeros(4, 2, dtype=torch.bfloat16)
+        src = torch.ones(4, 2, dtype=torch.float32)
+        result = self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
+        assert result["pv"].dtype == torch.bfloat16
+        assert result["pv"].shape == (4, 2)
+        assert torch.all(result["pv"] == 1.0)
+
+    def test_oversized_src_raises_with_shapes(self):
+        """A replay buffer longer than its template is a packing bug — fail
+        loudly with the key and both shapes, not a bare copy_ error."""
+        tmpl = torch.zeros(4, 2)
+        src = torch.ones(9, 2)
+        with pytest.raises(ValueError, match="leading-dim overflow"):
+            self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
+
+    def test_trailing_dim_mismatch_raises_with_shapes(self):
+        """Only the leading dim is padded, so a differing trailing dim is a
+        layout bug and must fail loudly rather than slice-copy silently."""
+        tmpl = torch.zeros(8, 2)
+        src = torch.ones(4, 3)
+        with pytest.raises(ValueError, match="trailing-dim mismatch"):
+            self._manager({"pv": tmpl})._pad_to_template({"pv": src}, 256)
 
 
 class TestMMEncoderJITManagerInit:
@@ -275,28 +349,78 @@ def _image_mm_kwargs(t=1, h=4, w=4):
     # Derived from Qwen3.5-0.8B vision_config:
     #   patch_size=16, temporal_patch_size=2, in_channels=3,
     _PATCH_FEAT = 2 * 16 * 16 * 3  # temporal_patch_size * patch_size^2 * in_channels
+    # Seeded on the grid so a failure reproduces run to run.
+    gen = torch.Generator().manual_seed(hash((t, h, w)) & 0xFFFFFFFF)
     return {
-        "pixel_values": torch.randn(t * h * w,
-                                    _PATCH_FEAT,
-                                    dtype=torch.bfloat16),
+        "pixel_values":
+        torch.randn(t * h * w,
+                    _PATCH_FEAT,
+                    dtype=torch.bfloat16,
+                    generator=gen),
         "image_grid_thw": [(t, h, w)],
     }
 
 
-@pytest.fixture(scope="module")
-def qwen35_mm_encoder():
-    engine_args = EngineArgs(
-        model="Qwen/Qwen3.5-0.8B",
-        max_model_len=256,
-        max_num_batched_tokens=256,
-        max_num_seqs=4,
-        dtype="bfloat16",
-        load_format="dummy",
-        limit_mm_per_prompt={"image": 1},
-    )
+def _reinit_params_for_sensitivity(params: dict, seed: int = 0) -> dict:
+    """Replace vLLM's dummy weights with a non-degenerate random init.
+
+    ``load_format="dummy"`` fills EVERY parameter — norm scales included — with
+    uniform(-1e-3, 1e-3). In that regime each ViT block's residual contribution
+    is ~1e-9 of the residual stream, so the encoder output is essentially
+    ``patch_embed`` + ``merger`` and is numerically INSENSITIVE to which tokens
+    the attention actually sees: a padded-vs-eager comparison then passes even
+    when budget-padding rows leak into a real image's attention segment
+    (measured on Qwen2.5-VL-3B: max|out| 1.0e-3, and the same 7.5e-6 max
+    difference with and without that bug).
+
+    Re-init norm scales to 1, biases to 0 and every other float parameter to
+    normal(0, 0.02) so the blocks actually shape the output and the comparison
+    has teeth. The same weights feed both the padded JIT path and the eager
+    reference, so equivalence still has to hold exactly.
+    """
+    key = jax.random.key(seed)
+    out: dict = {}
+    for i, (name, value) in enumerate(sorted(params.items())):
+        if not (hasattr(value, "dtype")
+                and jnp.issubdtype(value.dtype, jnp.floating)):
+            out[name] = value
+        elif name.endswith(".bias"):
+            out[name] = jnp.zeros_like(value)
+        elif "norm" in name or "ln_" in name:
+            out[name] = jnp.ones_like(value)
+        else:
+            out[name] = (jax.random.normal(
+                jax.random.fold_in(key, i), value.shape, dtype=jnp.float32) *
+                         0.02).astype(value.dtype)
+    return out
+
+
+def _build_mm_encoder_manager(engine_args: EngineArgs,
+                              params_transform=None,
+                              **compilation_overrides):
+    """Load a real model with dummy weights and build a live manager.
+
+    Shared by the per-model integration fixtures. The torch distributed init is
+    process-global, so it runs at most once even with several fixtures.
+
+    Args:
+      engine_args: Model/scheduler args; ``load_format="dummy"`` keeps this to
+        a config download plus random weights.
+      params_transform: Optional callable applied to the sharded jax parameter
+        dict before the manager is built (see
+        ``_reinit_params_for_sensitivity``).
+      compilation_overrides: Extra ``compilation_config`` attributes to set
+        (e.g. ``encoder_cudagraph_max_vision_items_per_batch``).
+
+    Returns:
+      ``(manager, mesh)`` — the mesh must be active (``jax.set_mesh``) around
+      any call that reaches the vision attention kernel.
+    """
     vllm_config = engine_args.create_engine_config()
     vllm_config.device_config.device = "cpu"
     vllm_config.compilation_config.cudagraph_mm_encoder = True
+    for key, value in compilation_overrides.items():
+        setattr(vllm_config.compilation_config, key, value)
 
     # sharded_flash_attention uses batch_axis="data" and head_axis="model";
     # both axes must be present. With one device both are size-1 (replicated).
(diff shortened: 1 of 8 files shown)


tpu_inference/runner/mm_encoder_jit_manager.py
```diff
@@ -356,13 +370,25 @@ def _pad_to_template(
...
+                    f"template: key={key} src_shape={tuple(src.shape)} "
+                    f"template_shape={tuple(tmpl.shape)} budget={budget} "
+                    f"max_batch_size={self.max_batch_size}; the batch was "
+                    f"packed into a token budget whose capture inputs are too "
+                    f"small for this key.")
+            # General case: zeroed buffer + the model's per-key padding logic.
             buf = torch.zeros_like(tmpl)
-            n = src.shape[0]
-            buf[:n] = src.to(dtype=tmpl.dtype, device=tmpl.device)
+            padding_logic = self.config.padding_logics.get(
```
QiliangCui, line 395: **[major]** Honouring `padding_logics` here is model-agnostic, so it also activates gemma-4's `pad_pixel_position_ids` (`tpu_inference/models/jax/gemma4_mm.py:857-869`): pad images in a partial batch go from position-(0,0) "real" images (old zero fill) to `-1`-masked ones. This looks benign — the capture template is already all-`-1`, items are independent along the batch axis, and outputs are sliced per item — but it is an unmentioned behavior change on gemma-4's compiled encoder path, and neither CI build 24696 nor dev 932 ran a gemma-4 job. Please add a sentence to the PR description and run the gemma-4 26B/31B accuracy + benchmark steps on the dev pipeline before merging.

tpu_inference/runner/mm_encoder_jit_manager.py
```diff
@@ -148,7 +148,12 @@ def encoder_eager_forward(self,
         # the inherited _execute_local can stay in plain-torch context (its
         # replay-buffer prep indexes model Parameters, which must NOT run
         # under the torchax dispatch).
-        with torchax.default_env():
+        # ``enable_torch_wrap(False)`` is required for the same reason as in
+        # ``run_budget_forward``: with the torch custom-op layer on, vllm.ir
+        # ops (e.g. ``vllm_ir.rms_norm`` in the ViT block norms) reach torchax
+        # as opaque OpOverloads it has no lowering for, and dispatch dies with
+        # "torchax Tensors can only do math within the torchax environment".
+        with torchax.default_env(), enable_torch_wrap(False):
```
QiliangCui, line 156: **[minor]** (PR description, "Notes") "TPU requires `--disable-chunked-mm-input`" overstates it — `tpu_inference/platforms/tpu_platform.py:427-433` only `logger.warning`s. The `get_rope_by_thw` lru_cache hazard is still unreachable, for a stronger reason: budget-vs-eager routing in `_execute_local` is a pure function of an item's token count, i.e. of its `(t, h, w)`, so a geometry that goes eager always goes eager and its cached torchax tensors never reach the replay-buffer prep. Suggest rewording so the guarantee doesn't hinge on a flag.

tests/runner/test_mm_encoder_jit_manager.py
```diff
@@ -398,3 +593,131 @@ def test_execute_within_budget(self, qwen35_mm_encoder):
...
+    """Build image mm_kwargs for a list of (t, h, w) Qwen2.5-VL grids.
+
+    h and w must be multiples of spatial_merge_size (2). pixel_values rows are
+    concatenated per item in grid order, matching how the model's
+    select_encoder_cudagraph_items slices a batch back apart.
+    """
+    total_patches = sum(t * h * w for t, h, w in grids)
+    return {
+        "pixel_values":
+        torch.randn(total_patches, _QWEN25VL_PATCH_FEAT, dtype=torch.bfloat16),
```
QiliangCui: **[minor]** `torch.randn` is unseeded, so the padded-vs-eager comparison runs on different inputs every CI run. The margins are wide (14x / 3x), but seed it (e.g. `torch.randn(..., generator=torch.Generator().manual_seed(0))`) so a future tolerance failure is reproducible.

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
```diff
@@ -28,6 +28,39 @@
 _VIT_FLASH_ATTENTION_VMEM_LIMIT_BYTES = 64 * 1024 * 1024
 
 
+def _segment_ids_from_cu_seqlens(cu_seqlens, total_len: int):
```
QiliangCui: **[minor]** `_segment_ids_from_cu_seqlens` duplicates `generate_window_segment_ids` (`tpu_inference/models/jax/qwen2_5_vl.py:190`) — same `searchsorted(cu[1:], pos, side="right")` construction, different id convention (0-based with pad = `num_segs` here vs 1-based with pad = 0 there). Consider lifting one helper into `tpu_inference/layers/common/` (next to `SegmentIds`) and having both call it, or at least cross-reference the two.

tpu_inference/runner/mm_encoder_jit_manager.py
```diff
@@ -356,13 +370,25 @@ def _pad_to_template(
...
-                # max_batch_size at metadata-prep time).
-                padded[key] = src
+                # max_batch_size at metadata-prep time). Still cast to the
+                # template dtype — a differing dtype would change the jit
+                # signature and force a recompile per request.
+                padded[key] = (src if src.dtype == tmpl.dtype else src.to(
+                    dtype=tmpl.dtype))
                 continue
-            # General case: zero buffer, then slice-copy src onto its head.
+            if src.shape[0] > tmpl.shape[0]:
```
QiliangCui: **[minor]** The oversize check only covers dim 0; a trailing-dim mismatch still dies as a bare `copy_` RuntimeError inside the padding logic. Extending the condition to `src.shape[1:] != tmpl.shape[1:]` gives that case the same tagged message.

tests/runner/test_mm_encoder_jit_manager.py
```diff
@@ -398,3 +593,131 @@ def test_execute_within_budget(self, qwen35_mm_encoder):
...
+
+    def _assert_padded_matches_eager(self, manager, mesh, grids):
+        mm_kwargs = _qwen25vl_mm_kwargs(grids)
+        with jax.set_mesh(mesh):
+            # Run the budget path BEFORE the eager reference. Both go through
+            # visual.get_rope_by_thw, which is @lru_cache'd on (t, h, w): the
+            # eager path runs inside the torchax env and would seed that cache
+            # with torchax tensors, which then blow up in the replay-buffer
+            # prep (plain torch, outside the env).
+            padded = manager.execute(mm_kwargs)
```
QiliangCui, line 703: **[nit]** Correctness here depends on `execute()` having seen every geometry before any eager call, and that ordering also holds *across* tests on the module-scoped fixture (`get_rope_by_thw` is lru_cached per `(t, h, w)`). Calling `manager.model.visual.get_rope_by_thw.cache_clear()` at the top of each test would make them order-independent, so adding/reordering a test can't poison the cache.

tests/runner/test_mm_encoder_jit_manager.py
```diff
@@ -310,27 +422,110 @@ def qwen35_mm_encoder():
                                     device=jax.devices()[0],
                                     need_pp=False)
 
+    # TPUPlatform.check_and_update_config attaches `sharding_config` to the
+    # VllmConfig as a plain attribute — it is not a declared dataclass field.
+    # Models with a nested text config (Qwen2.5-VL) call
+    # VllmConfig.with_hf_config -> vllm.config.utils.replace during init, which
+    # rebuilds the config from __dict__ and raises "Field 'sharding_config' not
+    # found in VllmConfig" on any extra key. Stash it across model construction.
+    sharding_config = vllm_config.__dict__.pop("sharding_config", None)
```
QiliangCui: **[nit]** This re-implements the production workaround in `tpu_inference/models/vllm/vllm_model_wrapper.py:200-205` (`delattr` on the load-time config). Using the same idiom and pointing the comment at it would keep the two in sync.

tests/runner/test_mm_encoder_jit_manager.py
```diff
@@ -283,20 +349,66 @@ def _image_mm_kwargs(t=1, h=4, w=4):
...
+    has teeth. The same weights feed both the padded JIT path and the eager
+    reference, so equivalence still has to hold exactly.
+    """
+    key = jax.random.key(seed)
+    out: dict = {}
+    for i, (name, value) in enumerate(sorted(params.items())):
+        if not (hasattr(value, "dtype")
+                and jnp.issubdtype(value.dtype, jnp.floating)):
+            out[name] = value
+        elif "norm" in name or "ln_" in name:
```
QiliangCui: **[nit]** `"norm" in name` is checked before `.bias`, so a LayerNorm bias would be re-initialised to ones. Harmless for Qwen2.5-VL (RMSNorm, no bias) — swapping the order avoids the surprise.

QiliangCui (commented):
Reviewed against the LKG vLLM (`d626108b18`): `_copy_padded_buffer(dst, src)` / `padding_logics` (default `{}`) / `self.config` + `self.max_batch_size` all come from the parent `__init__`; `cu_seqlens` replay padding and `_pad_cumulative_seqlens_buffer` both repeat the last offset, so `searchsorted`'s monotonic precondition holds on both paths; the new segment-id math checks out for empty/leading/trailing segments and kv is sized on its own length. CI build 24696 green, pre-commit/DCO pass, formatters clean.

No blockers. 1 major (gemma-4 side effect — needs a dev run), 4 minor, 3 nits inline.

tpu_inference/layers/vllm/ops/scaled_dot_product_attention.py
QiliangCui: Done in 94b1c7cfe: lifted one shared helper, `segment_ids_from_cu_seqlens(cu_seqlens, total_len)`, into `tpu_inference/layers/common/attention_interface.py` (0-based ids, positions >= cu[-1] get `num_segs`). `vllm_vit_sdpa` uses it directly; `generate_window_segment_ids` in the JAX Qwen2.5-VL is now `shared(cu, seq_len) + 1` followed by its unchanged zero-concat, which is the same expression it had before (for cu[-1] <= p < seq_len both give num_segs+1). Helper tests moved to `tests/layers/common/test_attention_interface.py`; `tests/models/jax/test_qwen2_5_vl.py` passes on TPU.

tpu_inference/runner/mm_encoder_jit_manager.py
QiliangCui: Done in 94b1c7cfe: the check is now `src.shape[0] > tmpl.shape[0] or src.shape[1:] != tmpl.shape[1:]`, and the tagged message says which (`leading-dim overflow` / `trailing-dim mismatch`). Added `test_trailing_dim_mismatch_raises_with_shapes`.

tests/runner/test_mm_encoder_jit_manager.py
QiliangCui: Done in 94b1c7cfe: both `_qwen25vl_mm_kwargs` and `_image_mm_kwargs` (also unseeded) now use `torch.Generator().manual_seed(...)` derived from the grid(s).

tests/runner/test_mm_encoder_jit_manager.py
QiliangCui: Done in 94b1c7cfe: `get_rope_by_thw.cache_clear()` at the top of `_assert_padded_matches_eager` and `test_execute_single_image`, so test order cannot poison the lru_cache. The execute-before-eager comment now uses the token-count routing argument instead of the flag.

tests/runner/test_mm_encoder_jit_manager.py
QiliangCui: Done in 94b1c7cfe: same `hasattr`/`delattr` idiom as `vllm_model_wrapper.py`, with the comment pointing at that production site (it strips the attribute from its load-time config copy; the test re-attaches it afterwards because the manager and sharding helpers run against the same config).

tests/runner/test_mm_encoder_jit_manager.py
QiliangCui: Done in 94b1c7cfe: `.bias` -> zeros is now checked before `norm`/`ln_` -> ones.

tpu_inference/runner/mm_encoder_jit_manager.py
QiliangCui: Agreed, reworded (PR description and the test comment): the routing in `_execute_local` is a pure function of an item's token count, i.e. of its (t, h, w), so in production a geometry that goes eager always goes eager and its cached torchax tensors never reach the replay-buffer prep. The integration test is the only place both paths see the same geometry; it clears the cache per test and runs execute() first.

tpu_inference/runner/mm_encoder_jit_manager.py
QiliangCui: Agreed, and added to the PR description. Ran gemma-4 on the dev pipeline with this branch's manager code (build 934, https://buildkite.com/tpu-commons/tpu-inference-dev/builds/934), tpu7x TP=8, `cudagraph_mm_encoder` on, both impl variants:

| model | impl=auto (JAX, bar 2.0/1.80) | impl=vllm (bar 1.55/1.35) |
|---|---|---|
| gemma-4-26B-A4B | 6.72 req/s, 16384 tok, 0 failed | 2.12 req/s, 16384 tok, 0 failed |
| gemma-4-31B | 4.60 req/s, 16384 tok, 0 failed | 1.78 req/s, 16384 tok, 0 failed |

Same regime as the last nightlies (24656 auto 26B 5.33; 24660 vllm 26B 2.16). The 26B gsm8k job in that build is red only because the dev yml used 0.70 instead of the model's 0.41 bar (measured 0.413, which passes the real gate; text-only anyway). On why it is benign: the capture template is already all `-1`, SigLIP encodes images independently along the batch axis, and outputs are sliced per item, so the pad images' position ids cannot reach a real item.

QiliangCui:
Re-reviewed at 96e350b22 — all 8 threads addressed, resolving.

Verified:
- #4 helper lifted to `tpu_inference/layers/common/attention_interface.py::segment_ids_from_cu_seqlens`; `generate_window_segment_ids` = shared + 1 is expression-equivalent to the old code (same int32 ids for every position); the 5 helper tests moved with it.
- #5 trailing-dim check + `leading-dim overflow` / `trailing-dim mismatch` reason in the tagged error; new `test_trailing_dim_mismatch_raises_with_shapes`.
- #3 seeded inputs (`hash` of int tuples is unsalted, so the seed is stable across processes); #6 `get_rope_by_thw.cache_clear()` per test; #7 `hasattr`/`delattr` idiom with the pointer to `vllm_model_wrapper.py`; #8 `.bias` before `norm`.
- #1/#2 description updated (gemma-4 `padding_logics` side effect; routing-by-token-count argument). Dev build 934 ran the first PR SHA, which already contained the `padding_logics` change: gemma-4 26B/31B perf green on both impl variants; the 26B gsm8k red is the dev yml's 0.70 bar — measured 0.413 strict-match vs the model yml's 0.41. Dev build 935's `tpu_inference/` + `tests/` tree is byte-identical to this head; all jobs green on tpu7x and tpu6e.
- pre-commit / DCO pass; yapf / isort / ruff clean locally; CI build 24701 in progress.

One leftover doc nit, no action needed for merge: the "Tests" bullet still names `_segment_ids_from_cu_seqlens`; it is now `segment_ids_from_cu_seqlens` in `layers/common/attention_interface.py`.

gxd3 (approved):

tpu_inference/runner/mm_encoder_jit_manager.py
anthonsu: This is great! This enables the Gemma-4 custom padding we added to vLLM (on GPU) to be seamlessly used on the TPU as well.

anthonsu (approved):

kwang3939 (approved):

theminghuang (approved):

jrplatin (approved):

PR #3452, opened 2026-08-24
Fix Gemma 4 MoE sharding, routed experts gating, and sampling microbatching

prishajain1:
# Description

Fixes several issues encountered when loading and running Gemma 4 MoE on v6e, along with Flax NNX compatibility and sampling microbatching.

### Summary of changes:
- **MoE loading & sharding fixes**: 
  - Shard expert weight params at creation time (`create_param`) so we don't replicate full weights across all devices during init (prevents HBM OOM on large MoE models).
  - Refactor `_get_weight_shardings` to reuse `_get_moe_weight_shardings` from `moe_weights.py`.
  - Avoid redundant top-k in `Gemma4MoE` when `enable_return_routed_experts` is not requested (defaults to `False`).
- **Sampling & Perf Optimizations**:
  - Add optional microbatching for TPU top-k / top-p sampling (`SAMPLING_MICROBATCH_SIZE`, defaults to `0` / disabled) via `jax.lax.map` to constrain peak VMEM during binary search reductions over large vocabularies.
     - **Result**: For a batch size of 32, the sampling time decreased from **6.08ms** (Baseline) to **1.105ms** (with microbatchsize of 16), a *5.5x reduction* in sampling time. (Gemma4 MoE on v6e-8)
 image image

# Tests
- `pytest tests/layers/common/test_fused_moe_gmm.py`
- `pytest tests/layers/jax/quantization/test_unquantized.py`
- `pytest tests/layers/jax/sample/test_sampling.py`
- `pytest tests/models/jax/test_gemma4.py`

# Checklist


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -102,6 +104,53 @@ def _apply_sampling_transforms(
...
+    microbatch_temperature = tpu_sampling_metadata.temperature.reshape(
+        (num_microbatches, microbatch_size))
+    microbatch_top_k = tpu_sampling_metadata.top_k.reshape(
+        (num_microbatches, microbatch_size))
+    microbatch_top_p = tpu_sampling_metadata.top_p.reshape(
+        (num_microbatches, microbatch_size))
+
+    def transform_microbatch(inputs):
+        chunk_logits, temperature, top_k, top_p = inputs
+        chunk_metadata = TPUSupportedSamplingMetadata(
```
muskansh-google: If the caller set do_sampling=False (for greedy decoding / argmax), this forces sampling on.

Also, any additional parameters on TPUSupportedSamplingMetadata (such as min_p, repetition penalties, or random seeds) are getting silently dropped here.

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -34,6 +35,7 @@
 _SAMPLING_EPS = 1e-5
 
 
+
```
muskansh-google: Nit: Remove extra line

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -102,6 +104,53 @@ def _apply_sampling_transforms(
...
+    """Apply sampling transforms in sequential, scratchpad-sized chunks.
+
+    The top-k and top-p implementations repeatedly reduce over the complete
+    vocabulary. Processing large request batches at once can push their working
+    set out of fast TPU memory. Keep the existing path for small or non-divisible
+    batches, and split larger divisible batches into fixed-size chunks.
+    """
+    batch_size = logits.shape[0]
+    microbatch_size = envs.SAMPLING_MICROBATCH_SIZE
+    if (microbatch_size <= 0
```
muskansh-google: Why do we keep the existing path for small or non-divisible batches?

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -41,6 +41,13 @@
 TARGET_SLOT_CHUNK_SIZE = 2048
 
 
+def _invert_permutation(permutation: jax.Array) -> jax.Array:
+    """Construct the exact inverse of a one-dimensional permutation."""
```
patrickji2014: i think this argsort replacement has some precondition,  could you add more detail for this function like "this is argsort replacement because ... "

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -440,6 +447,36 @@ def _compute_use_ep() -> bool:
...
+                         if mesh.shape.get(axis, 1) > 1)
+            if not axes:
+                return None
+            return axes[0] if len(axes) == 1 else axes
+
+        if "expert" in mesh.axis_names:
+            expert_axis = active_axis(
+                ("attn_dp", "attn_dp_expert", "expert", "model", "dcp",
+                 "pcp"))
+            tensor_axis = active_axis(
```
patrickji2014: In sharding.py there is 
```
MOE_TENSOR = ('attn_dp', 'model', 'dcp', 'pcp')
```

is the tensor_axis here intentionally created to be same as expert_axis?

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -440,6 +447,36 @@ def _compute_use_ep() -> bool:
         return (pc.data_parallel_size * pc.prefill_context_parallel_size *
                 pc.tensor_parallel_size) > 1 and pc.enable_expert_parallel
 
+    @staticmethod
+    def _get_weight_shardings(
```
patrickji2014: iiuc, https://github.com/vllm-project/tpu-inference/blob/5c1ed68b4235751f0974b61ecb3d915f2a78f3da/tpu_inference/layers/common/process_weights/moe_weights.py#L648 has similar functionality. if true, we should dedup them or just add a "TODO: refactor with moe_weights.py:_get_expert_shard_axis"  for later on code refactor or clean up.

tpu_inference/envs.py
```diff
@@ -498,9 +499,13 @@ def _get_int_list_env() -> list[int]:
     # RL weight sync: verify tensor checksums after each Raiden H2D transfer.
     "VERIFY_WEIGHTS":
     env_bool("VERIFY_WEIGHTS", default=False),
+    # Microbatch size for sampling block. Set to 0 to disable microbatching (disabled by default).
+    "SAMPLING_MICROBATCH_SIZE":
+    lambda: int(os.getenv("SAMPLING_MICROBATCH_SIZE", "0")),
 }
 
 
+
```
patrickji2014, line 518: nit, extra blank line. 

In case you don't know the pre-commit lint, have you run "pre-commit run --show-diff-on-failure --color=always --all-files".

tpu_inference/layers/jax/sample/sampling.py
prishajain1: - Small batches (< microbatch size) already fit on VMEM without microbatching and that path is already fast and optimized. 
- For non divisible batches:  Ideally the user would give a microbatch size that is a factor of the batch size (we can add this as a comment), hence somebody using this feature optimally won't run into such issues. If you think we still should support non divisible batches, we can potentially pad the last chunk upto the microbatchsize. This is just a fall back solution if the case of non divisible batch arises (falling back to existing implementation)

tpu_inference/layers/jax/sample/sampling.py
prishajain1: Done! Thanks for pointing this

tpu_inference/layers/jax/moe/moe.py
prishajain1: Corrected

patrickji2014 (approved):

tpu_inference/layers/common/fused_moe_gmm.py
```diff
@@ -683,7 +696,7 @@ def _process_tokens_locally(hidden_states_local, topk_indices_local):
         group_sizes_local = jax.nn.one_hot(topk_indices_flat,
                                            global_num_experts,
                                            dtype=jnp.int32).sum(axis=0)
-        topk_argsort_revert_indices = jnp.argsort(topk_argsort_indices)
+        topk_argsort_revert_indices = _invert_permutation(topk_argsort_indices)
```
lk-chen: This is on critical path, for GMM change, you need to check in CL and get approval in google3 https://source.corp.google.com/piper///depot/google3/third_party/py/tokamax/google/benchmarking/components/vllm_fused_moe_gmm/fused_moe_gmm.py;l=752;bpv=1;bpt=0;rcl=973144579 then port the change in Github.

Let's revert this file from this PR

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -440,6 +447,34 @@ def _compute_use_ep() -> bool:
         return (pc.data_parallel_size * pc.prefill_context_parallel_size *
                 pc.tensor_parallel_size) > 1 and pc.enable_expert_parallel
 
+    # TODO: refactor with moe_weights.py:_get_expert_shard_axis
+    @staticmethod
+    def _get_weight_shardings(mesh: jax.sharding.Mesh,
```
lk-chen: this static method is a bit hard to follow the logic, let's just reuse https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/process_weights/moe_weights.py#L564-L607

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -391,26 +391,33 @@ def __init__(
         D = hidden_size
         F = intermediate_size_moe
 
-        # Weights are initially unsharded; quant method shards them in
-        # process_weights_after_loading via shard_moe_weights.
+        # Assign the final loading layout up front. Otherwise each full expert
+        # tensor is temporarily replicated on every device before the quant
+        # method gets a chance to redistribute it, which can exhaust HBM.
+        self.use_ep = self._compute_use_ep()
+        self.edf_sharding, self.efd_sharding = self._get_weight_shardings(
```
lk-chen: we don't need to add 2 more member variable

tpu_inference/layers/jax/moe/moe.py
```diff
@@ -477,10 +512,13 @@ def _load_weights(self,
                 raise ValueError(f"Unexpected param type in {rel_name}, "
                                  "expected gate_proj, up_proj, or down_proj")
             assert isinstance(jax_param, nnx.Param)
+            permute_dims = ((0, 2, 1) if self.moe_backend
+                            == MoEBackend.FUSED_MOE else None)
```
lk-chen: do you encounter FUSED_MOE ? the expected behavior is that moe_method should be either GMM_TP or GMM_EP

tpu_inference/layers/jax/quantization/unquantized.py
```diff
@@ -189,6 +189,18 @@ def process_weights_after_loading(self, layer: JaxRoutedExperts, *args,
             layer: The layer to process.
         """
         if layer.moe_backend == MoEBackend.FUSED_MOE:
+            # The streaming loader may invoke post-processing before every
```
lk-chen: same as above, the expected behavior is that moe_method should be either GMM_TP or GMM_EP

tpu_inference/models/common/model_loader.py
```diff
@@ -142,13 +142,18 @@ def _get_nnx_model(
     model_config = (vllm_config.speculative_config.draft_model_config
                     if is_draft_model else vllm_config.model_config)
 
-    def create_abstract_model() -> nnx.Module:
+    def create_abstract_model(graph_updates: bool = False) -> nnx.Module:
         """
         Helper class to create an abstract model for `nnx.eval_shape`.
 
+        Some Flax NNX versions forward ``graph_updates`` to the evaluated
```
lk-chen: in that case our CI should catch this as regression, could you check your flax version?

tpu_inference/models/jax/gemma4.py
```diff
@@ -649,6 +650,10 @@ def __init__(self,
                                      mesh=mesh,
                                      rngs=rng,
                                      quant_config=quant_config,
+                                     enable_return_routed_experts=getattr(
+                                         config,
+                                         "enable_return_routed_experts",
```
lk-chen, line 655: but Gemma4 config doesn't have this field? https://huggingface.co/google/gemma-4-26B-A4B-it/blob/main/config.json

tpu_inference/layers/jax/moe/moe.py
prishajain1: No, default runs use GMM_TP / GMM_EP as expected. I only ran into this when experimenting with USE_MOE_EP_KERNEL=1.

I've reverted the FUSED_MOE handling from this PR so we stay strictly scoped to GMM.

tpu_inference/layers/jax/moe/moe.py
prishajain1: removed, made local vars

tpu_inference/layers/jax/moe/moe.py
prishajain1: Refactored `_get_weight_shardings` to reuse `_get_moe_weight_shardings` directly from `moe_weights.py`

tpu_inference/models/common/model_loader.py
prishajain1: Reverted

tpu_inference/models/jax/gemma4.py
prishajain1: config here is vLLM's ModelConfig, not the Hugging Face config. Upstream had this hardcoded to True, which ran an extra top_k even when routing metadata wasn't needed. Defaulting to False avoids that duplicate op

tpu_inference/layers/common/fused_moe_gmm.py
prishajain1: Reverted and opened this CL: https://critique.corp.google.com/cl/981105448

tpu_inference/models/jax/gemma4.py
lk-chen: is there a reason why you use `getattr` instead of just `config. enable_return_routed_experts`?

lk-chen (commented):
there's no more "loading" related changes in this PR, could you update the title?

tpu_inference/models/jax/gemma4.py
prishajain1: Changed

lk-chen (approved):

lk-chen:
@copilot list per-file owner, just name don't at them

QiliangCui (approved):

PR #3453, opened 2026-08-24
Fix GDN v3 metadata padding for partial sequence tiles

xuuuge:
# Description

Pad GDN v3 batched sequence metadata to complete sequence tiles before invoking the Pallas kernel.

I ran into this while trying Qwen3.5-9B on a Kaggle v5e-8 TPU with `max_num_seqs=1`, which is a pretty normal setup when testing or benchmarking a single request. The model loaded successfully, but generation crashed during decode inside `fused_conv1d_gdn_batched`:

```text
Scalar error PC ... HLO: fused_conv1d_gdn_batched.1
JaxRuntimeError: The program continuator has halted unexpectedly
SLICE_FAILURE_SW_INJECT_ERROR
Terminating the libtpu controller process because an anomalous TPUworker process is detected
```

`libtpu` terminated the TPU worker and the notebook kernel could not recover, so the Kaggle runtime had to be recycled. Changing only `max_num_seqs` from 1 to 4 allowed the same checkpoint and prompt to generate successfully (`TPU load succeeded`).

This pointed to a mismatch between the four-sequence decode tile and the one-entry metadata buffers. The kernel consumes complete metadata tiles with bounds checks disabled, but the previous implementation allocated metadata for only `max_seqs`. When `max_seqs` was not divisible by `tile_size`, the final tile could read past those buffers.

This change:

- pads sequence metadata to a complete tile;
- maps padded entries to the reserved null state slot with zero request length;
- clamps `num_tiles` to the actual sequence capacity; and
- adds direct metadata regressions for a single sequence and a partial second tile.

No documentation changes are needed because this fixes an internal kernel invariant without changing the public API. It follows the kernel's existing padding convention by representing unused positions as zero-length entries mapped to the reserved null state slot.

# Tests

- `pre-commit run --files tests/kernels/gdn_metadata_v3_test.py tpu_inference/kernels/gdn/v3/metadata.py` — all configured hooks passed.
- `pytest tests/kernels/gdn_metadata_v3_test.py` on CPU JAX — both new regression cases passed.
- Kaggle v5e-8, JAX/JAXLIB 0.11.0, libtpu 0.0.44:
  - the full GDN v3 suite plus temporary one-request and five-request end-to-end cases passed 15/15;
  - five additional repetitions of both temporary end-to-end cases passed 10/10.

Separately, in the linked notebooks, I called `compute_batched_seq_metadata` with 384 combinations of `tile_size`, `max_seqs`, and `end_seq`, then inspected the metadata it returned. Before the fix, 206 combinations returned metadata sized to `max_seqs` instead of a complete tile, while 178 were already tile-aligned. With the fix, all 384 combinations returned complete, safely padded metadata. This was a standalone diagnostic, not 384 repository tests.

Original failure and `max_num_seqs=1` vs. `4` confirmation: https://www.kaggle.com/code/xuanyangge/qwen3-5-9b-vllm-tpu-load-smoke

Fix evidence notebook: https://www.kaggle.com/code/xuanyangge/gdn-v3-partial-decode-tile-pr-evidence

Unpatched baseline notebook: https://colab.research.google.com/drive/1kPcz8ySPXiRptl5yoQor7eHT0dOzTxlL

# Checklist

- [x] I have made or will make corresponding changes to any relevant documentation. (No documentation change required.)

### DIFF
--- a/tests/kernels/gdn_metadata_v3_test.py
+++ b/tests/kernels/gdn_metadata_v3_test.py
@@ -1,0 +1,84 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jax.numpy as jnp
+import numpy as np
+from absl.testing import parameterized
+
+from tpu_inference.kernels.gdn.v3 import config, metadata
+
+
+class GDNMetadataTest(parameterized.TestCase):
+
+    @parameterized.named_parameters(
+        dict(testcase_name="single_sequence", max_seqs=1),
+        dict(testcase_name="partial_second_tile", max_seqs=5),
+    )
+    def test_batched_metadata_pads_partial_tile(self, max_seqs):
+        tile_size = 4
+        padded_max_seqs = ((max_seqs + tile_size - 1) // tile_size) * tile_size
+        padding = padded_max_seqs - max_seqs
+        dtypes = config.Dtypes(*([jnp.float32] * 5))
+        cfg = config.GDNConfig(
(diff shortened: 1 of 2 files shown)


xuuuge:
@kyuyeunk Could you add the ready label when you have a moment? The PR is now ready for review, but GitHub does not allow me to apply repository labels. Both current Buildkite failures are the missing-label bootstrap gate; no tests were scheduled.

kyuyeunk:
is this still an issue after https://github.com/vllm-project/tpu-inference/pull/3411 ?

PR #3472, opened 2026-08-27
bench: dual write benchmark results to BigQuery

theminghuang:
`report_result.sh` now mirrors each benchmark result into BigQuery right after the Spanner upsert, sharing the `RecordId` as `record_id` so the two sinks join. This lets dashboards migrate off Spanner without a flag day. Spanner stays the source of record; the BigQuery write is best-effort and never fails a build. `BQ_UPLOAD_ENABLED=false` turns it off.

Table is `llm_benchmark_analytics.benchmark_runs`, shared with vllm-torchtpu and keyed apart by `repository`. `bq_utils.py` is a port of that repo's copy and should stay in sync — with one fix worth porting back: BigQuery rejects the doubled `''` quote escape, parsing `'it''s'` as two adjacent literals, so `sql_escape` uses `\'`. That bug is latent in torchtpu only because its payloads have no apostrophes.

See the commit message for the metric-name mapping and retry handling.

Validated with `bq query --dry_run` against the live table. Not exercised end to end — `UPLOAD_DB=false` on PR builds, so the first real write is the first daily build after merge.


theminghuang:

- #3472 ◀

QiliangCui (approved):

PR #3476, opened 2026-08-27
Avoid arange().repeat() + gather when building token_indices_sorted in fused MoE GMM

shungcp:
## What

In `_process_tokens_locally`, `token_indices_sorted` was built by materializing `jnp.arange(num_tokens).repeat(topk)` and then gathering it with `topk_argsort_indices`. Since the flattened top-k indices are token-major, flat index `i` maps to token `i // topk`, so `token_indices[topk_argsort_indices]` is exactly `topk_argsort_indices // topk`. Compute it directly, using a bit-shift when `topk` is a power of two.

## Why

Removes an intermediate buffer allocation + a gather, replacing them with a single elementwise op. Microbenchmark shows this step drop from ~400us to ~40-50us. Mathematically equivalent (verified across power-of-two and non-power-of-two topk).

### DIFF
--- a/tpu_inference/layers/common/fused_moe_gmm.py
+++ b/tpu_inference/layers/common/fused_moe_gmm.py
@@ -672,12 +672,18 @@ def fused_moe_func(
             topk_indices, global_num_experts)
 
     def _process_tokens_locally(hidden_states_local, topk_indices_local):
-        num_tokens_local = hidden_states_local.shape[0]
         topk_indices_flat = topk_indices_local.flatten()
         topk_argsort_indices = jnp.argsort(topk_indices_flat)
-        token_indices = jnp.arange(num_tokens_local,
-                                   dtype=jnp.int32).repeat(topk)
-        token_indices_sorted = token_indices[topk_argsort_indices]
+        # topk_indices_flat is flattened token-major, so flat index i maps to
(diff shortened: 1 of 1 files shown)


shungcp:
End-to-end validation on v6e, across 4 MoE models × 2 model-definition paths (`MODEL_IMPL_TYPE=flax_nnx` / `=vllm`). Throughput change, ABAB alternation with an independent server start per arm:

| Model | topk | `flax_nnx` | `vllm` (torchax) |
|---|--:|---|---|
| GLM-4.7-Flash | 4 | +9.3% (4-chip TP, int8) / +11.2% (EP8, int8) | +10.65% (4-chip, bf16) |
| Qwen3-30B-A3B | 8 | +10.10% | +10.39% |
| Gemma-4-26B-A4B | 8 | +8.22% | +8.03% |
| Llama-4-Scout-17B-16E | 1 | +0.83% | +0.88% |

Isolated cost of the removed gather, ≈ 0.0112 µs × `idx.size` (= `num_tokens_local × topk`):

| idx.size | before | after |
|--:|--:|--:|
| 4,096 | 51.7 µs | 33.1 µs |
| 8,192 | 102.1 µs | 32.6 µs |
| 16,384 | 200.6 µs | 32.5 µs |
| 32,768 | 399.8 µs | 32.5 µs |

lk-chen (approved):

PR #3481, opened 2026-08-28
Fix vLLM Integration Failures

patrickji2014:
# Description

This PR attempts to fix all the failures in tpu-inference integration tests due to vLLM upstream change. 

### Changes in vLLM
1. In upstream vLLM, ```KVCacheTensor.shared_by``` replaced by ```KVCacheTensor.layers```
2. Upstream removed ```get_kv_cache_layout()``` and ```set_kv_cache_layout()``` from vllm.v1.attention.backends.utils, causing ImportError / AttributeError.
3. Field Renames in MLA Specs: ```compress_ratio``` was renamed to ```tokens_per_state```, and ```storage_block_size``` was renamed to ```num_states``` in ```MLAAttentionSpec```.
4. For Group-Aware Tensor Sizing: ```KVCacheTensor``` now groups all same-spec layers in a group into ```kv_cache_tensor.layers```, where ```tensor.size``` allocates memory for all layers in the group (```size >= num_layers * page_size_bytes * num_blocks```), rather than legacy sizing for only 1 layer.

### Fixes in tpu-inference
1. Updated all references from ```shared_by``` to ```layers``` in ```tpu_inference/runner/kv_cache_manager.py```
 and test suites. 
2. Replaced ```get_kv_cache_layout()``` / ```set_kv_cache_layout()``` with direct use of ```DEFAULT_KV_CACHE_LAYOUT```.
3. Updated DeepSeek-V4 attention and indexer cache to use ```tokens_per_state``` and ```num_states``` (with backwards-compatible fallbacks).
4. Updated ```KVCacheManager.initialize_kv_cache``` to set ```num_blocks = kv_cache_config.num_blocks``` and ```alloc_per_layer = True``` when ```tensor.size``` accommodates all layers in ```kv_cache_tensor.layers```.

### Changes in vLLM
Upstream standardized the KV cache layout around [L, B, H, N, C] using KVCacheTensor with layers, offset, layer_stride, block_stride, and spec.num_states.
In the old layout (assumed by main), a KVCacheTensor held cross-group aliasing layers via shared_by (e.g. [layer.0, layer.9, layer.52]).
In the new layout, a KVCacheTensor contains all same-spec layers within a group (e.g. [layer.0, layer.1, ..., layer.8]), with layer_stride spacing them out and multiple group tensors starting from offset = 0 to overlay memory.

### Why tpu-inference's current layer aliasing fails under the New Layout

Intra-group layer collision: tpu-inference ```main``` assigned layer_idx = i (the tensor index). Because a single tensor now contains all layers of a group, layers 0..8 in Group 0 all mapped to cache index 0, overwriting each other's KV data in every forward step.
Over-calculated num_blocks: main did tensor.size // page_size_bytes expecting 1 layer per tensor, multiplying the block count by num_layers.
Unsupported unequal group sizes: main had no way to map per-layer physical slices (offset + j * layer_stride) for hybrid models like Gemma 3 (with 9, 9, ..., 10 layers).

### Fix via mirroring GPU vLLM:
GPU vLLM: Calls [torch.as_strided(raw, storage_offset=raw.storage_offset() + kv_cache_tensor.offset)](https://github.com/vllm-project/vllm/blob/main/vllm/v1/kv_cache_interface.py#L358-L362), letting ```address(l, b) = offset + l * layer_stride + b * block_stride``` handle aliasing implicitly.
TPU Backend: Because JAX arrays are strongly typed and cannot do pointer reinterpreting, it can be done in 2 passes to mirror what GPU's change due to this vLLM refactor PR. 
Pass 1 explicitly computes ```layer_offset = offset + j * layer_stride``` to form a ```slot_key```. Layers with identical byte offsets collapse into the same slot. 
Pass 2 allocates exactly one physical JAX array per unique slot.

### Problem & Root Cause:
During engine pause operations (await engine.pause_generation(mode="keep")), vLLM's ```EngineCore._finish_pause``` triggers ```self.model_executor.collective_rpc("synchronize_device")```.
```WorkerBase.synchronize_device()``` falls back to ```torch.accelerator.synchronize()```.
Because torchax registers "jax" as a PrivateUse1 backend in PyTorch, ```torch.accelerator.is_available()``` returns True. However, ```torch.accelerator.synchronize()``` internally calls torch._C._accelerator_getDeviceIndex(), which fails with:
```RuntimeError: PyTorch is not linked with support for jax devices```
This unhandled exception caused the EngineCore background process to die (EngineDeadError).

The triggers of ```self.model_executor.collective_rpc``` is a introduced by this [pr code change](https://github.com/vllm-project/vllm/pull/52914/changes#diff-3dd8e96bc7c1aaf28faa13b3b705f4f7bdbf755aec34dcf4d9b67c933ddfb127R828). Before the vllm pr, no call to synchronize_device.

### Fix in tpu-inference:
Implemented ```synchronize_device(self)``` on TPUWorker in ```tpu_inference/worker/tpu_worker.py``` using ```jax.effects_barrier()```.
Added a monkeypatch in ```tpu_inference/platforms/tpu_platform.py``` for ```torch.accelerator.synchronize()``` to catch the "jax" runtime error and safely fall back to ```jax.effects_barrier()```.

### UnquantizedLinearMethod __init__ Signature:
Upstream vLLM changed ```UnquantizedLinearMethod.__init__``` to take no arguments. Calling ```super().__init__(linear_config)``` in ```VllmUnquantizedLinearMethod``` caused a TypeError.
333 Fix: 
Explicitly initialized ```vllm_linear.UnquantizedLinearMethod.__init__(self)``` and ```common_unquantized.UnquantizedLinearMethod.__init__(self, linear_config)```.
### LoRA Punica Wrapper:
Upstream removed ```extra_vocab_size``` from ```convert_mapping()``` and changed ```embeddings_indices``` to 1D.
Fix: Removed ```extra_vocab_size``` argument from ```convert_mapping``` and adjusted padding dimension in ```_pad_to_shape``` to use ```dims=self._embeddings_indices.ndim```.
### DeepSeek-V4 MLA Attention backend_cls:
Upstream introduced an abstract class attribute ```backend_cls``` on ```DeepseekV4Attention```.
Fix: Defined ```backend_cls = DeepseekV4SparseMLABackend``` on ```VllmDeepseekV4MLAAttention```.

### Compressed Tensors Unit Test:
Removed assertion on deleted attribute scheme.cutlass_block_fp8_supported.

# Tests

Nightly CI Tests Passed https://buildkite.com/tpu-commons/tpu-inference-ci/builds/24992/canvas

# Checklist

# Note
AI(Antigravity Gemini) Assisted and Review By Author

### CODE REVIEW

**Review by gxd3 — APPROVED** [review]

**Review by mrjunwan-lang — APPROVED** [review]

**Review by QiliangCui — APPROVED** [review]

**Review by vanbasten23 — APPROVED** [review]

File: tpu_inference/runner/kv_cache_manager.py [hunk 1]
```diff
@@ -734,7 +731,7 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
 
         # set the kv cache layout which is needed by kv connectors
         # NOTE(jcgu): please update the default value when the order changes
-        set_kv_cache_layout(DEFAULT_KV_CACHE_LAYOUT)
+        # set_kv_cache_layout(DEFAULT_KV_CACHE_LAYOUT)
```
jrplatin: I assume this is not needed, so just remove?

jrplatin (approved):

tpu_inference/runner/kv_cache_manager.py
patrickji2014: yeah, this should be removed. thanks.

PR #3487, opened 2026-08-30
feat: Add compressed-tensors W4A16 int4 support for vLLM-path MoE and linear layers

inkitori:
# Description

Add compressed-tensors W4A16 asymmetric int4 support for vLLM-path linear and MoE layers. Also adds an `axis` argument to `u32_unpack_i4` to correctly unpack compressed-tensors MoE zero points and fixes bug in `quantize_tensor` that truncated instead of rounding for int quants

# Tests

Unit tests:

```
pytest tests/layers/vllm/test_compressed_tensors_wna16.py tests/layers/vllm/test_compressed_tensors_moe_w4a16.py -v
```

Also tested Qwen3-30B-A3B ([FabioTrindade/Qwen3-30B-A3B-W4A16KV16-asym-GS-128](https://huggingface.co/FabioTrindade/Qwen3-30B-A3B-W4A16KV16-asym-GS-128)) with

```
MODEL_IMPL_TYPE=vllm vllm serve $MODEL --max-model-len 4096 --port 8000
```

GSM8K (5-shot, no cot):

```
local-completions ({'model': 'FabioTrindade/Qwen3-30B-A3B-W4A16KV16-asym-GS-128', 'base_url': 'http://localhost:8000/v1/completions', 'num_concurrent': 8, 'max_retries': 3, 'tokenized_requests': False}), gen_kwargs: ({}), limit: None, num_fewshot: None, batch_size: 8
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8810|±  |0.0089|
|     |       |strict-match    |     5|exact_match|↑  |0.8908|±  |0.0086|
```

Ran using

```
lm_eval \
  --model local-completions \
  --model_args "model=$MODEL,base_url=http://localhost:8000/v1/completions,num_concurrent=8,max_retries=3,tokenized_requests=False" \
  --tasks gsm8k \
  --batch_size 8
```

Also benchmarked performance with `vllm bench serve` on v6e-1 TPU

```
Successful requests:                     200
Maximum request concurrency:             8
Benchmark duration (s):                  49.85
Total input tokens:                      204800
Total generated tokens:                  25600
Request throughput (req/s):              4.01
Output token throughput (tok/s):         513.56
Peak output token throughput (tok/s):    688.00
Peak concurrent requests:                16.00
Total token throughput (tok/s):          4622.07
Mean TTFT (ms):                          250.21
Median TTFT (ms):                        268.11
P99 TTFT (ms):                           493.53
Mean TPOT (ms):                          13.72
Median TPOT (ms):                        13.72
P99 TPOT (ms):                           15.42
Mean ITL (ms):                           13.72
Median ITL (ms):                         10.76
P99 ITL (ms):                            89.85
```

Ran with

```
vllm bench serve \
  --model "$MODEL" \
  --dataset-name random \
  --random-input-len 1024 --random-output-len 128 \
  --num-prompts 200 --max-concurrency 8 \
  --port 8000
```

# Checklist


jt226ub:
Thanks for this — it closes a real hole. We hit the same gap porting a compressed-tensors int4
MoE checkpoint (`qwen4_exp`) to a v5e-8 and carried a local patch set for it, so a few notes
from that experience. Three are things this PR fixes and one is a regression it would introduce.

### 1. The rounding fix is right, and there is a second site

The added hunk is important beyond W4A16, and the `jnp.issubdtype` guard on it is the detail
that makes it safe:

```python
     tensor_q = jnp.clip(tensor * scale_inv, dtype_min, dtype_max)
+    if jnp.issubdtype(dtype, jnp.integer):
+        tensor_q = jnp.rint(tensor_q)
     tensor_q = tensor_q.reshape(orig_shape)
     tensor_q = tensor_q.astype(dtype)
```

Guarding matters because `quantize_tensor` is also called with float8 —
`compressed_tensors_w8a8_fp8.py:142` passes `jnp.float8_e4m3fn`, and
`layers/common/quantization/fp8.py:163` passes a `requant_weight_dtype` — where rounding to
integers would be a far worse bug than the truncation being fixed. Worth stating in the commit
message, because the obvious-looking version of this patch (unguarded `jnp.rint`) breaks fp8
requantisation, and a future reader deleting the guard would not learn that from the diff.

Before the patch, the integer path was `jnp.clip(tensor * scale_inv, min, max).astype(dtype)` — a
C-style truncating cast, so every magnitude is biased toward zero by up to a full quantisation
step.

Measured on a real checkpoint at int8 group-128: **57% of one expert's weights quantised to
zero**, and the block's routed output came out at **0.41×** the reference. It is not a rounding
nicety; it silently degrades any int-quantised model on the path.

`static_per_tensor_quantize_tensor`, immediately below it
(`tpu_inference/layers/common/quantization/__init__.py:283` on `main`), has the identical
unguarded pattern and is **not** in this PR's scope:

```python
return jnp.clip(tensor / scale, dtype_min, dtype_max).astype(dtype)
```

That one is on the **KV-cache** path, which is why we would not leave it: it is the helper behind
`quantize_kv` (`__init__.py:294,297`) and is called with `kv_cache_quantized_dtype` from
`flash_attn_mla.py:166-168` and `models/jax/deepseek_v3.py:697-699`. With an int8 KV cache, every
cached key and value is biased low by up to a full step, on every token — a wider blast radius
than weights, and much harder to notice, since it degrades long-context quality rather than
failing.

It needs the same `jnp.issubdtype` guard for the same reason: those call sites pass fp8 dtypes
too. (`tpu_inference/kernels/quantized_matmul/util.py:267` already does
`jnp.round(data / scale).astype(target_dtype)` — the correct form exists in-tree, in a third
copy of this operation, which is arguably the real finding here.)

### 2. The predicate reordering fixes a crash worth calling out explicitly

The new `_is_wNa16_group_channel` check placed **before** `_is_dynamic_token_w8a8`, with the
comment *"Weight-only schemes have no input quantization; check them before the w8a8
predicates, which assume input_quant is present"*, fixes an `AttributeError` that currently
kills any weight-only checkpoint before a kernel runs:
`_is_dynamic_token_w8a8` (vLLM `compressed_tensors.py:498`) evaluates
`weight_quant.num_bits == input_quant.num_bits == 8` with no `None` guard, unlike its sibling
`_is_fp8_w8a8`, which guards at `:542-543`.

The ordering makes the crash unreachable **for checkpoints this PR's predicate matches**. It
does not fix the predicate, so a weight-only config that falls outside `_is_wNa16_group_channel`
(W8A16, or an int4 layout the predicate rejects) still reaches `:142` and still crashes with a
`NoneType` attribute error rather than the intended `NotImplementedError` at `:149`. Filing that
separately as its own bug; flagging it here so the two are not read as duplicates.

### 3. `defer_all_reduce` will not reach the new linear scheme, and the MoE path will double-reduce

This is the one that would land as a regression, and it is not visible in the diff — it is in
the wiring around it.

`VllmRowParallelLinear.__init__` (`tpu_inference/layers/vllm/custom_ops/linear.py:39-42`) is the
only producer of `defer_all_reduce`:

```python
linear_config = getattr(self.quant_method, "linear_config", None)
if linear_config is not None:
    linear_config.defer_all_reduce = not self.reduce_results
```

For compressed-tensors, `get_quant_method` returns **vLLM's** `CompressedTensorsLinearMethod`
and stashes the scheme on the layer instead
(`compressed_tensors/compressed_tensors.py:169-170`):

```python
layer.scheme = scheme
return CompressedTensorsLinearMethod(self)
```

`CompressedTensorsLinearMethod` has no `linear_config`; the scheme does. So the `getattr`
returns `None`, the assignment is skipped, and `defer_all_reduce` keeps its default of `False`
(`layers/common/quantization/configs.py:29`) for **every** compressed-tensors layer — including
the new `VllmCompressedTensorsWNA16`, whose `if getattr(self.linear_config, "defer_all_reduce", False):`
branches (diff lines 1257, 1284) are therefore unreachable in a real engine.

The same is already true of the existing CT schemes, which read
`self.linear_config.defer_all_reduce` at `compressed_tensors_w8a8_int8.py:204,231`,
`compressed_tensors_w8a8_fp8.py:257,302` and `compressed_tensors_w4a8_fp8.py:249,274` — all dead
paths today. By contrast `nvfp4.py:97-103` returns `VllmNvfp4LinearMethod(self, linear_config)`,
a method object that *owns* the config, which is why the mechanism works there.

**Observable consequence** for a CT-quantised MoE with a shared expert at TP>1: the shared
expert's `RowParallelLinear` is constructed with `reduce_results=False` because the MoE runner
intends to merge that all-reduce into the fused-MoE one
(`layers/vllm/interface/moe.py:129`), but the layer all-reduces anyway, so the shared-expert
contribution is summed twice. We measured the block's output at **7.99× the reference at TP=8**
(≈ TP-size scaling, as expected for a doubled reduction) until we forced the plumbing through.

Two shapes of fix, either fine by us:

- in `VllmRowParallelLinear.__init__`, fall back to the scheme —
  `getattr(getattr(layer, "scheme", None), "linear_config", None)`; or
- have CT's `get_quant_method` return a tpu-inference linear method that owns
  `linear_config`, like nvfp4 does, rather than vLLM's wrapper.

A test that fails without the fix and passes with it: build a `RowParallelLinear` with
`reduce_results=False` under a CT int8 or W4A16 config and assert
`layer.scheme.linear_config.defer_all_reduce is True`. Happy to open that as a separate PR
against this branch or after it lands, whichever you prefer.

### 4. The JAX-native path is untouched

`tpu_inference/layers/jax/quantization/compressed_tensors.py:154` still carries
`# TODO: w4a8 / wNa16 schemes need their own JAX methods (not yet ported).`, so after this PR a
compressed-tensors W4A16 checkpoint works on the vLLM/torchax path and still has no scheme on
the JAX-native path. Filing that as its own issue rather than expanding this PR's scope —
mentioning it here only so the two paths' status is not conflated in the release notes.

---

**Evidence provenance.** Line numbers are `main` at `623e68a14`; the diff quotes are from
`gh pr diff 3487` as of 2026-09-08. The 57%/0.41×/7.99× figures were measured on a TPU v5e-8
(`vllm-tpu==0.28.0`, `MODEL_IMPL_TYPE=vllm`, TP=8) with a compressed-tensors int4 MoE
checkpoint; they are properties of the arithmetic and the reduction, not of that chip.

jt226ub:
Numbers for the two things this comment said were being filed separately: the unguarded `_is_dynamic_token_w8a8` predicate is #3539, and the JAX-path `wNa16` gap is #3540. The `defer_all_reduce` wiring — the one that would land as a regression with this PR — is #3537.

PR #3490, opened 2026-08-31
perf(sample): enable distributed top-k candidate sampling and keep vocab logits sharded

prishajain1:
### Summary
This PR introduces distributed candidate sampling and an option to retain vocabulary logits in sharded layout across TP shards during sampling to reduce latency and memory traffic:

1. **Retain Sharded Logits**: Avoids all-gathering the full 262k vocabulary logits prior to sampling, saving ~330µs. 
2. **Distributed Top-K Candidate Sampling (`DISTRIBUTED_SAMPLING_MAX_TOP_K`)**: Each TP shard extracts its top `2*DISTRIBUTED_SAMPLING_MAX_TOP_K` candidates locally. With the default maximum of, this gathers 128 candidates per shard (1,024 candidates across 8 shards), and applies each request's runtime top-k and top-p values.
3. **Exactness Guarantee**: Uses the requested top-k threshold and preserves boundary ties. It safely falls back to full-vocabulary sampling if a shard may have omitted candidates tied at threshold. 
4. **Safety & Fallbacks**:
   - `DISTRIBUTED_SAMPLING_MAX_TOP_K` defaults to 64 and controls the static compiled candidate capacity
   - Raw logprobs supported; processed logprobs modes use the full-vocabulary path when logprobs are requested
   - The full-vocabulary path explicitly unshards logits, preserving the existing fallback behaviour. 

### Benchmark Results
We tested on Gemma4 MoE model on v6e-8:
```
export DISTRIBUTED_SAMPLING_MAX_TOP_K=64
vllm serve google/gemma-4-26B-A4B-it \
  --max-model-len=4096 \
  --max-num-seqs=256 \
  --tensor-parallel-size=8 \
  --gpu-memory-utilization=0.85 \
  --async-scheduling \
  --kv-cache-dtype=fp8 \
  --limit-mm-per-prompt '{"image":0,"video":0,"audio":0}' \
  --block-size=128 \
  --additional-config \
  '{"sharding":{"sharding_strategy":{"tensor_parallelism":8,"expert_parallelism":1}}}'

vllm bench serve \
        --backend openai \
        --endpoint /v1/completions \
        --port 8000 \
        --dataset-name random \
        --random-input-len 1024 \
        --random-output-len 128 \
        --request-rate inf \
        --num-prompts 500 \
        --ignore-eos
```
- Reduced `jit(sample)` latency from 19.3 ms to 2.9 ms (-85%).
Screenshot 2026-09-10 at 10 20 17 PM

Screenshot 2026-09-10 at 10 19 07 PM

### Tests Added
- `test_distributed_candidates_match_full_vocab_filters`
- `test_distributed_candidates_preserve_topk_boundary_tie`
- `test_distributed_candidates_detect_truncated_tie_group`
- Env var tests in `tests/test_envs.py`

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -86,7 +86,6 @@ if TYPE_CHECKING:
     TPU_MESH_SORT_BY_COORDS: bool = False
     VERIFY_WEIGHTS: bool = False
     SAMPLING_MICROBATCH_SIZE: int = 0
-    SAMPLING_KEEP_SHARDED_LOGITS: bool = False
     DISTRIBUTED_SAMPLING_MAX_TOP_K: int = 64
 
 
@@ -514,10 +513,6 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Microbatch size for sampling block. Set to 0 to disable microbatching (disabled by default).
     "SAMPLING_MICROBATCH_SIZE":
     lambda: int(os.getenv("SAMPLING_MICROBATCH_SIZE", "0")),
-    # Keep vocabulary logits sharded across TP shards during sampling to avoid
-    # an expensive all-gather across TP before sampling.
-    "SAMPLING_KEEP_SHARDED_LOGITS":
-    env_bool("SAMPLING_KEEP_SHARDED_LOGITS", default=False),
     # Largest runtime top-k handled by distributed candidate sampling. This is
     # read at trace time so candidate tensor shapes remain static.
     "DISTRIBUTED_SAMPLING_MAX_TOP_K":
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -87,6 +87,7 @@ if TYPE_CHECKING:
     VERIFY_WEIGHTS: bool = False
     SAMPLING_MICROBATCH_SIZE: int = 0
     SAMPLING_KEEP_SHARDED_LOGITS: bool = False
+    DISTRIBUTED_SAMPLING_MAX_TOP_K: int = 64
     USE_DISTRIBUTED_TOPK_SAMPLING: bool = False
 
 
@@ -518,6 +519,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # an expensive all-gather across TP before sampling.
     "SAMPLING_KEEP_SHARDED_LOGITS":
     env_bool("SAMPLING_KEEP_SHARDED_LOGITS", default=False),
+    # Largest runtime top-k handled by distributed candidate sampling. This is
+    # read at trace time so candidate tensor shapes remain static.
+    "DISTRIBUTED_SAMPLING_MAX_TOP_K":
+    lambda: int(os.getenv("DISTRIBUTED_SAMPLING_MAX_TOP_K", "64")),
     # Sample supported top-k requests from compact sharded candidates. The
     # sampler falls back to its general path for unsupported or incomplete
     # candidate sets.
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -85,7 +85,6 @@ if TYPE_CHECKING:
     VLLM_INCREMENTAL_FP8_LOADING: bool = False
     TPU_MESH_SORT_BY_COORDS: bool = False
     VERIFY_WEIGHTS: bool = False
-    SAMPLING_MICROBATCH_SIZE: int = 0
     DISTRIBUTED_SAMPLING_MAX_TOP_K: int = 64
 
 
@@ -510,9 +509,6 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # RL weight sync: verify tensor checksums after each Raiden H2D transfer.
     "VERIFY_WEIGHTS":
     env_bool("VERIFY_WEIGHTS", default=False),
-    # Microbatch size for sampling block. Set to 0 to disable microbatching (disabled by default).
-    "SAMPLING_MICROBATCH_SIZE":
-    lambda: int(os.getenv("SAMPLING_MICROBATCH_SIZE", "0")),
     # Largest runtime top-k handled by distributed candidate sampling. This is
     # read at trace time so candidate tensor shapes remain static.
     "DISTRIBUTED_SAMPLING_MAX_TOP_K":
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -88,7 +88,6 @@ if TYPE_CHECKING:
     SAMPLING_MICROBATCH_SIZE: int = 0
     SAMPLING_KEEP_SHARDED_LOGITS: bool = False
     DISTRIBUTED_SAMPLING_MAX_TOP_K: int = 64
-    USE_DISTRIBUTED_TOPK_SAMPLING: bool = False
 
 
 def env_with_choices(
@@ -523,11 +522,6 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # read at trace time so candidate tensor shapes remain static.
     "DISTRIBUTED_SAMPLING_MAX_TOP_K":
     lambda: int(os.getenv("DISTRIBUTED_SAMPLING_MAX_TOP_K", "64")),
-    # Sample supported top-k requests from compact sharded candidates. The
-    # sampler falls back to its general path for unsupported or incomplete
-    # candidate sets.
-    "USE_DISTRIBUTED_TOPK_SAMPLING":
-    env_bool("USE_DISTRIBUTED_TOPK_SAMPLING", default=False),
 }
 
 
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -85,6 +85,9 @@ if TYPE_CHECKING:
     VLLM_INCREMENTAL_FP8_LOADING: bool = False
     TPU_MESH_SORT_BY_COORDS: bool = False
     VERIFY_WEIGHTS: bool = False
+    SAMPLING_MICROBATCH_SIZE: int = 0
+    SAMPLING_KEEP_SHARDED_LOGITS: bool = False
+    USE_DISTRIBUTED_TOPK_SAMPLING: bool = False
 
 
 def env_with_choices(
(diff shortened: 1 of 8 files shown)


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -32,6 +34,11 @@
     from tpu_inference.runner.input_batch import CachedRequestState
 
 _SAMPLING_EPS = 1e-5
+# Candidate tensor shapes must be static inside the compiled sampler. Keep the
+# supported top-k and candidate capacity explicit; requests outside this fast
+# path continue through the general full-vocabulary implementation.
+_DISTRIBUTED_SAMPLING_SUPPORTED_TOP_K = 64
+_DISTRIBUTED_SAMPLING_CANDIDATES_PER_SHARD = 128
```
muskansh-google: Why are these flags hard coded to values of 64 and 128?

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -102,6 +109,178 @@ def _apply_sampling_transforms(
     return logits
 
 
+def _apply_sampling_transforms_microbatched(
```
muskansh-google: I think this PR includes overlapping changes from other PRs like - https://github.com/vllm-project/tpu-inference/pull/3452

Can we raise independent non-overlapping PRs to make review easier?

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -177,16 +309,60 @@ def sample(
...
+                sampled_tokens, incomplete_candidates = (
+                    _distributed_topk_sample(
+                        rng,
+                        mesh,
+                        logits,
+                        tpu_sampling_metadata.temperature,
+                        tpu_sampling_metadata.top_p,
+                    ))
+
+                def use_candidate_result(_):
```
muskansh-google, line 323: Can we validate the statement - "No caller consumes processed logits when logprobs=False."

lk-chen:
You could leave just https://github.com/vllm-project/tpu-inference/pull/3490/changes/ac1bce833fd0053469f86bb2ff020a3eb7a0f9b5 in this branch, so the changes in #3452 don't carry over

tpu_inference/envs.py
```diff
@@ -498,9 +501,22 @@ def _get_int_list_env() -> list[int]:
     # RL weight sync: verify tensor checksums after each Raiden H2D transfer.
     "VERIFY_WEIGHTS":
     env_bool("VERIFY_WEIGHTS", default=False),
+    # Microbatch size for sampling block. Set to 0 to disable microbatching (disabled by default).
+    "SAMPLING_MICROBATCH_SIZE":
+    lambda: int(os.getenv("SAMPLING_MICROBATCH_SIZE", "0")),
+    # Keep vocabulary logits sharded across TP shards during sampling to avoid
+    # an expensive all-gather across TP before sampling.
+    "SAMPLING_KEEP_SHARDED_LOGITS":
```
lk-chen: if there's no drawback of this feature (and dist. topk sampling), let's remove the flag and always enable it?

tpu_inference/layers/jax/sample/sampling.py
lk-chen: same question here

prishajain1: top_k is passed dynamically per request (not during server startup), so we used a static size to prevent JAX from recompiling every time a request asks for a different k. 
If we want to generalise, we may need to add a max_top_k parameter and then mask down for the k required by each batch of requests. Let me know if we want to implement this change and if the direction sounds good?
The current hardcoded values were extracted for gemma4 from standard generation config.

tpu_inference/envs.py
prishajain1: Yes, we can make it the default behaviour. For pure top p, and for cases where request has top k > provided value by user, we will safely fall back to standard full vocab path even when this flag is enabled.

tpu_inference/layers/jax/sample/sampling.py
prishajain1: Added DISTRIBUTED_SAMPLING_MAX_TOP_K which defaults to 64. Now all requests with top k values <=DISTRIBUTED_SAMPLING_MAX_TOP_K are covered, and for higher we fall back to original path.

tpu_inference/envs.py
```diff
@@ -508,6 +510,13 @@ def _get_int_list_env() -> list[int]:
     # RL weight sync: verify tensor checksums after each Raiden H2D transfer.
     "VERIFY_WEIGHTS":
     env_bool("VERIFY_WEIGHTS", default=False),
+    # Microbatch size for sampling block. Set to 0 to disable microbatching (disabled by default).
+    "SAMPLING_MICROBATCH_SIZE":
+    lambda: int(os.getenv("SAMPLING_MICROBATCH_SIZE", "0")),
+    # Largest runtime top-k handled by distributed candidate sampling. This is
+    # read at trace time so candidate tensor shapes remain static.
+    "DISTRIBUTED_SAMPLING_MAX_TOP_K":
```
lk-chen, line 515: can we remove the flag and use vocab_size // TP (which is much larger than typical top_k) as threshold? If there's no drawback, let's enable it as much as possible.

In that case, we don't need `_distributed_sampling_fits`

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -34,6 +36,53 @@
...
+    """Whether sampling can return raw logits for the requested logprob mode."""
+    return not (logprobs and str(logprobs_mode).startswith("processed"))
+
+
+def _can_sample_distributed(
+        tpu_sampling_metadata: TPUSupportedSamplingMetadata) -> jax.Array:
+    """Whether every row is supported by distributed candidate sampling."""
+    is_greedy = tpu_sampling_metadata.temperature < _SAMPLING_EPS
+    supported = ((tpu_sampling_metadata.top_k > 0) &
+                 (tpu_sampling_metadata.top_k <=
```
lk-chen: This line (`top_k <= ...max_top_k()`) seems redundant check of `_distributed_sampling_fits`

cc @gutianyu-google

tpu_inference/layers/jax/sample/sampling.py
prishajain1: Validated! in both TPU runner and fused decode loop, logits are consumed only when logprobs requested.

tpu_inference/layers/jax/sample/sampling.py
prishajain1: `_distributed_sampling_fits checks` whether the vocab shard size is large enough to supply the candidate buffer

`top_k <= max_top_k()` is runtime check per request, check if this specific request in the batch fits within our compiled candidate capacity.  If user has a large K, we route it to the full vocab path. 

Hence we should keep both

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -102,42 +151,264 @@ def _apply_sampling_transforms(
...
+        use_distributed_candidates = (
+            allow_distributed_sampling
+            and _distributed_sampling_fits(mesh, logits.shape[-1]))
+        if use_distributed_candidates:
+            # Candidate shapes use a trace-time maximum; each request's top-k
+            # remains dynamic. Greedy and padded rows do not consume a sample.
+            supported = _can_sample_distributed(tpu_sampling_metadata)
+
+            def sample_candidates(_):
+                sampled_tokens, incomplete_candidates = (
```
lk-chen, line 313: IIUC, `incomplete_candidates` only happen when there's tie around top-k boundary.

Does vLLM also relax `k` when there's tie? If not we can simplify this logic (i.e. not consider incomplete/tie, and hard truncate at k) cc @gutianyu-google

tpu_inference/envs.py
prishajain1: That would defeat the purpose of this optimization right (unless im missing something). We gather the fixed max_top_k candidate capacity rather than the request’s actual top_k, so using vocab_size // TP would make each shard gather almost its entire local vocabulary and lose the performance benefit.

tpu_inference/layers/jax/sample/sampling.py
prishajain1: I checked the upstream vllm:
- [Pytorch path](https://github.com/vllm-project/vllm/blob/51da0ca66c8065619c79e35dff97aa99aeaf5644/vllm/v1/sample/ops/topk_topp_sampler.py#L385-L391) preserves the boundary ties
- [Triton path](https://github.com/vllm-project/vllm/blob/51da0ca66c8065619c79e35dff97aa99aeaf5644/vllm/v1/sample/ops/topk_topp_triton.py#L840-L861) keeps exactly k

Our existing TPU path was already preserving ties, so this PR is aligned to keep that behaviour the same

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -34,6 +36,53 @@
...
+
+
+def _can_sample_distributed(
+        tpu_sampling_metadata: TPUSupportedSamplingMetadata) -> jax.Array:
+    """Whether every row is supported by distributed candidate sampling."""
+    is_greedy = tpu_sampling_metadata.temperature < _SAMPLING_EPS
+    supported = ((tpu_sampling_metadata.top_k > 0) &
+                 (tpu_sampling_metadata.top_k <=
+                  _distributed_sampling_max_top_k()) &
+                 (tpu_sampling_metadata.top_p > 0.0))
```
patrickji2014: this enhanced sampling algorithm is not limited to top_p > 0 right? we check top_p > 0 just because the way you implement right now.

tpu_inference/layers/jax/sample/sampling.py
gutianyu-google: Both are needed. `_distributed_sampling_fits` is decided once at compile time: is each vocab shard big enough to hold the candidate buffer? `top_k <= max_top_k` is decided per batch at runtime: does this request's k fit in that buffer? Different questions, so I'd keep both.

tpu_inference/layers/jax/sample/sampling.py
gutianyu-google: Yes, keeping ties matches what we do today: the TPU `topk_mask` keeps every value tied at rank k, and so does vLLM's PyTorch sampler. Cutting hard at k would change results for tied rows. Ties at the shard boundary are rare with float32 logits, so the fallback almost never fires and only costs one scalar `cond`. I'd keep it.

lk-chen (approved):

tpu_inference/layers/jax/sample/sampling.py
prishajain1: yes, this algorithm works for pure topk , top k+top p. we check top_p > 0 as a sanity guard . Pure top p is supported by original fall back algorithm

muskansh-google (approved):

lk-chen:
@copilot show per-file owner, don't at them just name

gxd3 (approved):

patrickji2014 (approved):

jrplatin (approved):

QiliangCui (approved):

PR #3492, opened 2026-08-31
[Core][DP] Fix "Should not schedule a request that does nothing!" in DPScheduler

csgoogle:
# Description

Under KV cache pressure, the DP engine wedges and never recovers:

- It spins on empty steps, TPU idle.
- It logs `Should not schedule a request that does nothing!` hundreds of thousands of times.
- Requests sit preempted and are never scheduled again, even though KV usage has dropped back to 0%.
- Only a kill gets you out.

## Why it happens

**Step 1 — DP overwrites the token count.**

`tpu_inference/core/sched/dp_scheduler.py`, `UPDATE_FROM_OUTPUT`:

```python
if num_output_tokens_dict.get(req_id, 0) > 0:        # decode-phase guard
    num_sampled = len(model_runner_output.sampled_token_ids[req_idx])
    if num_sampled > 0:
        scheduler_output.num_scheduled_tokens[req_id] = num_sampled
result = scheduler.update_from_output(scheduler_output, model_runner_output)
```

This is on purpose: MoE routing offsets index by the tokens actually sampled.

**Step 2 — that breaks a counter pairing in vLLM.**

vLLM adds the **scheduled** count and subtracts whatever the dict holds — now the **sampled** count:

```python
# vllm/v1/core/sched/scheduler.py:1393, _update_after_schedule
request.num_in_flight_tokens += num_scheduled_token   # scheduled

# vllm/v1/core/sched/scheduler.py:1802, update_from_output
request.num_in_flight_tokens -= num_tokens_scheduled  # sampled, after our overwrite
```

In vanilla vLLM these are the same number and it balances to zero. With the overwrite in between, the difference stays on the request. Nothing clears it.

**Step 3 — where scheduled and sampled actually diverge.**

Not in steady-state decode: that schedules 1 and samples 1. The gap comes from a **chunked prefill step**, which computes hundreds of tokens and samples exactly one.

The guard `num_output_tokens > 0` is meant to select decode-phase requests. But a request **resumed after a preemption** passes it too:

- `_preempt_request` zeroes `num_computed_tokens`, so the request must re-prefill its whole prompt.
- It still carries the output tokens it generated before being preempted, so `num_output_tokens > 0`.
- Resumed requests ride in `CachedRequestData` (`scheduler.py:1509`, `resumed_req_ids`).

Non-final chunks sample nothing and are skipped by the `num_sampled > 0` guard. The **final** prefill chunk samples 1 token:

```
_update_after_schedule:   in_flight += 638     (chunk size)
overwrite:                dict[req]  =   1     (sampled)
update_from_output:       in_flight -=   1
                          ────────────────
                          residue    = 637
```

**Step 4 — the second preemption weaponizes the residue.**

```python
# vllm/v1/core/sched/scheduler.py:1369, _preempt_request
request.num_stale_output_tokens = request.num_in_flight_tokens
```

Stale output is a real mechanism: after a preemption, output from already-dispatched steps is still arriving, and resuming early could resample a position that output is about to deliver. So the request is fenced:

```python
# vllm/v1/core/sched/scheduler.py:780, schedule()
if (request.num_stale_output_tokens > 0
        and not request.drop_stale_output):
    # parked in skipped_waiting
```

**Step 5 — deadlock.**

A real stale debt is paid off by the in-flight output as it arrives. This one is **phantom** — no step was ever dispatched for it. So:

- Nothing will ever arrive to decrement it.
- While stale > 0 the request is parked, so it is never scheduled, so it never appears in a future `scheduler_output`, so `update_from_output` never visits it either.

**Step 6 — the engine spins instead of stopping.**

`get_num_unfinished_requests()` counts `skipped_waiting`, so `has_requests()` stays `True`. The core loop never blocks and never exits.

## The fix

Record how many tokens the overwrite removed, then subtract that from both counters after `update_from_output` returns, so they track the scheduled count.

- The overwrite stays — routing offsets need the sampled count.
- Done **after** `update_from_output`, so that call still sees the values it expects today. Scheduling behaviour is unchanged.
- Clamped at zero; `num_stale_output_tokens` has an `assert >= 0` at `:1807`.
- Applied in the **worker** process, where the rank's real `scheduler.requests` objects live. The parent holds only pipes and `SchedulerOutput`s — no request state.
- Requests that finish and get dropped during the update are skipped.

Note this is not async-scheduling-specific: all four sites above are in the base `Scheduler`, and `AsyncScheduler` overrides neither the pairing nor the stale logic. In sync mode `num_in_flight_tokens` should be exactly 0 at every `schedule()`, so any non-zero stale there is entirely phantom.

# Tests

## Why the repro needs a big context

Preemption both **creates** the residue (by forcing the re-prefill) and **triggers** the deadlock (by converting it to stale). No KV pressure, no bug.

| `max_model_len` | KV pressure | Preemption | Result |
| --- | --- | --- | --- |
| 1524 | none | never | passes 4 waves x 256 requests — **false negative** |
| 16384 | real | yes | wedges in wave 0, every run |

An early repro at 1524 passed and was briefly read as "no bug". It could not have reproduced it.

## What the wedge looks like

`max_model_len 16384`, dp2xtp4, engine idle, KV usage back at 0%:

```
rank 0 containers: running=0 waiting=0 skipped_waiting=3 requests_map=3 unfinished=3
  wave0-0124 status=PREEMPTED stale=7   drop=False in_flight=7   preemptions=3 computed=0/4096
  wave0-0122 status=PREEMPTED stale=11  drop=False in_flight=11  preemptions=2 computed=0/4096
rank 1 containers: running=0 waiting=0 skipped_waiting=5 requests_map=5 unfinished=5
  wave0-0115 status=PREEMPTED stale=637 drop=False in_flight=637 preemptions=2 computed=0/4096
```

Reading this:

- **`preemptions` is 2 or 3 on every request, never 1** — exactly as the mechanism predicts: the first preemption creates the residue via re-prefill, the second converts it to stale.
- **`stale == in_flight`** on every one — the assignment at `:1369`, never decremented since.
- **`placeholders=0`, `computed=0/4096`** — nothing was genuinely in flight, so the entire stale count is phantom.
- **The values are arbitrary** (4, 6, 6, 7, 8, 10, 11, 637) because each is `final_chunk_size - 1`, and the final chunk is whatever remainder that request got from a budget (`max_num_batched_tokens=2496`) shared with others.
- `running=0 waiting=0` with `unfinished` non-zero — all of them in `skipped_waiting`.

Run config: `enable_continue_decode: false`, `async_scheduling: true`, `speculative_config=None`. So every overwrite that fired was this unintended re-prefill case, which is why the residue was always positive.

## Result

| | Before | After |
| --- | --- | --- |
| Outcome | wedged in wave 0 | `PASS cycle=0`, 2 waves x 128 requests |
| Empty steps | 110,867 | 2 |
| Parked requests | 8 | 0 |

## Two things that make this hard to find

- **Preemption is never logged.** Nothing in `scheduler.py` logs a preemption, so grepping a wedged log for `preempt` finds nothing. That does not mean it did not happen.
- **No DP probe reports `skipped_waiting`.** `GET_REQUEST_COUNTS` returns only `len(running), len(waiting)`, so the wedge shows up as `running=0 waiting=0 unfinished=8`, which looks impossible. `GET_PENDING_PREFILL_TOKENS` misses it too, so a rank full of parked requests reports 0 pending tokens and looks **idle** to `_find_best_rank_for_request` — worth fixing separately.

The probe that produced the dump above, and the repro drivers, are left out of this PR to keep it to the fix. Happy to send them as a follow-up.

# Checklist


wenxindongwork (approved):

tpu_inference/core/sched/dp_scheduler.py
```diff
@@ -243,11 +246,43 @@ def _send_result(result):
...
+                    # is copied into num_stale_output_tokens (_preempt_request).
+                    # schedule() skips any request with stale tokens, and only
+                    # update_from_output clears them -- so the request can never
+                    # run again, and the engine spins on empty steps forever.
+                    #
+                    # We only fix the counters. The rewrite stays because
+                    # routed_experts slicing needs the sampled count. We do it
+                    # after update_from_output so that call still sees the
+                    # values it expects. Clamp at zero: a multi-step decode can
+                    # sample more than was scheduled.
```
wenxindongwork: can we simplify this comment?

pv97:
Thank you for this fix!

Maybe the original `case SchedulerCommand.UPDATE_FROM_OUTPUT` is not ideal. The only reason scheduler_output.num_scheduled_tokens was mutated in the first place was so that `DPScheduler._split_model_output_by_rank` could slice the routed_experts tensor for MoE models.

Instead of mutating the scheduler output, would it be possible to slice `routed_experts` using `sampled_token_ids` directly: In `dp_scheduler.py`, `_split_model_output_by_rank` can determine token counts directly from `len(g.sampled_token_ids[g.req_id_to_index[req_id]])` instead of reading `scheduler_output.num_scheduled_tokens`.
Then we can remove the fragile `if model_runner_output.sampled_token_ids:` block.

pv97:
> Thank you for this fix!
> 
> Maybe the original `case SchedulerCommand.UPDATE_FROM_OUTPUT` is not ideal. The only reason scheduler_output.num_scheduled_tokens was mutated in the first place was so that `DPScheduler._split_model_output_by_rank` could slice the routed_experts tensor for MoE models.
> 
> Instead of mutating the scheduler output, would it be possible to slice `routed_experts` using `sampled_token_ids` directly: In `dp_scheduler.py`, `_split_model_output_by_rank` can determine token counts directly from `len(g.sampled_token_ids[g.req_id_to_index[req_id]])` instead of reading `scheduler_output.num_scheduled_tokens`. Then we can remove the fragile `if model_runner_output.sampled_token_ids:` block.

Actually nevermind, this might break prefill

sixiang-google (approved):

PR #3493, opened 2026-08-31
perf(sparsecore): enable dense gather reduce for v6e via FP32 intermediate buffer

prishajain1:
### SparseCore Dense Gather Reduce for v6e

Tensor cores are primarily designed for regular matrix multiplication. Since gathering for each token by expert and subsequent multiplication with expert router weights can be an irregular operation, we leverage the dedicated SparseCore kernel for this. There already exists an implementation of this kernel which works for v7x, but not on v6e. This PR explains why it did not work on v6e, and implements the fix to enable SparseCore dense gather reduce on v6e.

---

### Why did it work for v7x and not v6e?
- **v7x has 16 SparseCore SIMD lanes**, while **v6e has only 8 SIMD lanes**.
- Gemma 4 uses $top\_k = 8$ experts per token.

#### On v7x (16 lanes):
```
16 lanes
----------------------------------------------------------
| 8 expert rows for token A  │ 8 expert rows for token B |
----------------------------------------------------------
```
Hence, we get 2 output rows for one SIMD step. These two output rows are stored as `bfloat16`, and packed into a 32-bit unit:
```
One 32-bit unit
----------------------------------------------
| 16-bit bf16 token A  │ 16-bit bf16 token B |
----------------------------------------------
```
- $\text{Output block size} = 2 \text{ output rows} / 2 \text{ bf16 values per unit} = 1$ (valid integer block size).

#### On v6e (8 lanes):
```
8 lanes
------------------------------
| 8 expert rows for token A  |
------------------------------
```
One SIMD step produces **1 output row**. But when attempting to store that output directly as `bfloat16`, 2 values are required to fill one 32-bit packed unit:
```
One 32-bit unit
----------------------------------------------
| 16-bit bf16 token A  │ missing             |
----------------------------------------------
```
- $\text{Output block size} = 1 \text{ output row} / 2 \text{ bf16 values per unit} = 0$.
- A zero-sized output block is invalid in Pallas/Mosaic and caused a fallback to TensorCore JAX execution.

---

### The Fix
Write the kernel's output buffer as fp32, which packs 1 row per word. v6e then
gets 1/1 = 1 and the block is valid.

```
One 32-bit unit
---------------------------------------------
| 32-bit value (fp32)                       |
---------------------------------------------
```
The operand stays bf16. The kernel used to derive a single packing factor from
the operand dtype and use it for both the gather addressing and the output
block; only the output block is affected by this, so it now has its own
`out_dtype`. That matters because the operand is `gmm2_res`, which is topk
times larger than the output — upcasting it would cost a full fp32 copy and
buy nothing, since the gather addresses HBM in 32-bit words and a bf16 gather
already fetches the whole word.

#### Why is this lossless?

- Precision is unchanged. The reduction accumulates in fp32 either way; the
only difference is where the fp32 result gets rounded to bf16. With a bf16
output buffer the kernel does it before the write, with an fp32 buffer it
happens on the result right after. Same value, rounded once, in both cases.

---

### Benchmark Results
We tested on Gemma4 MoE model on v6e-8:
```
export PHASED_PROFILING_DIR="$HOME/vllm_profile"
export USE_BATCHED_RPA_KERNEL=1
export USE_BATCHED_RPA_SEQ_ON_LANE=1
export MODEL_IMPL_TYPE=flax_nnx
export NEW_MODEL_DESIGN=1

export SAMPLING_MICROBATCH_SIZE=16
export USE_DISTRIBUTED_TOPK_SAMPLING=1
export GEMMA4_REPLICATE_GLOBAL_KV_WEIGHTS=1
export BATCHED_RPA_DECODE_SLIDING_BKV_SIZE=1024


vllm serve google/gemma-4-26B-A4B-it \
  --max-model-len=4096 \
  --max-num-seqs=256 \
  --tensor-parallel-size=8 \
  --gpu-memory-utilization=0.85 \
  --async-scheduling \
  --kv-cache-dtype=fp8 \
  --limit-mm-per-prompt '{"image":0,"video":0,"audio":0}' \
  --block-size=128 \
  --additional-config \
  '{"sharding":{"sharding_strategy":{"tensor_parallelism":8,"expert_parallelism":1}}}'

vllm bench serve \
        --backend openai \
        --endpoint /v1/completions \
        --port 8000 \
        --dataset-name random \
        --random-input-len 1024 \
        --random-output-len 128 \
        --request-rate inf \
        --num-prompts 500 \
        --ignore-eos
```
- **Prefill Phase Forward Pass Latency**: 99 ms → 79 ms (**20% reduction**)
- **`jit_dense_gather_reduce` Kernel Execution Time**: 1.198 ms → 522 µs (**56% reduction**)

image
image

---

### DIFF
--- a/tests/kernels/dense_gather_reduce_test.py
+++ b/tests/kernels/dense_gather_reduce_test.py
@@ -258,11 +258,9 @@ class IsCompatibleTest(parameterized.TestCase):
 
     # (num_lanes, dtype, reduce_group_size, expected_is_compatible)
     @parameterized.named_parameters(
-        # v6e SparseCore (num_lanes=8). bf16 packing=2: 8//8//2 = 0 -> the
-        # Qwen3-30B-A3B crash -> must fall back.
-        ("v6e_bf16_topk8_degenerate", 8, jnp.bfloat16, 8, False),
-        # Same v6e lanes but f32 (packing=1): 8//8//1 = 1 -> kernel is valid,
-        # must NOT be blocked just because it is v6e.
+        # v6e SparseCore (num_lanes=8), bf16, topk=8: enabled via FP32 output buffer (packing=1)
+        ("v6e_bf16_topk8_fp32_buffer", 8, jnp.bfloat16, 8, True),
+        # Same v6e lanes but f32 (packing=1): 8//8//1 = 1 -> kernel is valid.
         ("v6e_f32_topk8_ok", 8, jnp.float32, 8, True),
         # v6e lanes, bf16, smaller group: 8//4//2 = 1 -> valid.
         ("v6e_bf16_topk4_ok", 8, jnp.bfloat16, 4, True),
--- a/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
+++ b/tpu_inference/kernels/sparse_core/dense_gather_reduce.py
@@ -48,9 +48,15 @@ def is_compatible(
     if sc_info.num_lanes % reduce_group_size != 0:
         return False
 
-    # The output block has (num_lanes // reduce_group_size) // packing rows;
-    # fall back to JAX when that is 0 (the kernel can't emit a zero-row block).
-    packing = 32 // jax.dtypes.itemsize_bits(op.dtype)
+    # The output block has (num_lanes // reduce_group_size) // packing rows.
+    # When num_lanes // reduce_group_size == 1 on bf16 (e.g. 8 lanes on v6e with topk=8),
+    # packing=2 would yield a 0-sized output block. We enable the kernel by storing the
+    # intermediate output in FP32 (packing=1) and converting back to bf16.
+    can_use_fp32_output_buffer = (
+        op.dtype == jnp.bfloat16
+        and (sc_info.num_lanes // reduce_group_size) == 1)
+    packing = 1 if can_use_fp32_output_buffer else (
+        32 // jax.dtypes.itemsize_bits(op.dtype))
     if (sc_info.num_lanes // reduce_group_size) // packing < 1:
         return False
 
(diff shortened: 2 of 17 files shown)


tpu_inference/kernels/sparse_core/dense_gather_reduce.py
```diff
@@ -313,15 +350,21 @@ def dense_gather_reduce(
                 break
             col_chunk_size -= 128
         if col_chunk_size > 0:
+            # Same decision is_compatible() used to pick packing; the upcast
+            # must match it or the kernel emits a zero-row output block.
+            kernel_dtype = _get_kernel_dtype(x.dtype, sc_info.num_lanes,
+                                             reduce_group_size)
+            kernel_x = x.astype(kernel_dtype) if kernel_dtype != x.dtype else x
```
guowei-dev: I'd be cautious of this one. Only the output block would need fp32. Widening the operand materialises a full fp32 copy of gmm2_res and doesn't reduce gather bytes since bf16 already fetches the whole word.

tests/kernels/dense_gather_reduce_test.py
guowei-dev: Since this PR enables dense_gather_reduce on v6e, we should be able to enable these tests now.

prishajain1: Done

tpu_inference/kernels/sparse_core/dense_gather_reduce.py
prishajain1: Updated so only the output buffer is fp32 (out_dtype). The operand stays untouched in bf16 so we avoid the large copy, and we just cast the smaller result back to x.dtype after.

guowei-dev (approved):

a1yssan13 (approved):

PR #3501, opened 2026-09-01
[Bugfix] Tolerate non-resizable unquantized weight storage

### LINKED ISSUE #3400: Safetensors-backed CPU weights can fail cleanup because storage is not resizable
## Summary

After fixing a TPU weight-loading path so host tensors are transferred directly into their target sharding, unquantized weight cleanup began failing on safetensors-backed CPU tensors with:

```text
RuntimeError: Trying to resize storage that is not resizable
```

The failure comes from cleanup code that unconditionally does:

```python
tensor.untyped_storage().resize_(0)
```

Safetensors-backed tensors can use non-resizable storage, so this is not a universally valid way to release their CPU backing memory.

## Current affected pattern

Current `tpu_inference/layers/vllm/quantization/unquantized.py` contains direct calls of this form for several weights / biases after transfer, including linear and MoE paths.

Examples include cleanup for:

```text
layer.weight
layer.bias
layer.w13_weight
layer.w2_weight
layer.w13_bias
layer.w2_bias
```

The tensor is detached from the module immediately afterward, so a non-resizable safetensors backing store can safely be handled by dropping the final owning reference rather than forcing the storage to resize.

## Reproducer context

- single-host TPU v5e-8
- Qwen3.8-27B BF16
- safetensors checkpoint
- unquantized TPU weight path

Once the preceding TPU placement bug was bypassed, all checkpoint shards loaded and this cleanup call became the next deterministic blocker.

## Local fix

The working local patch wraps storage release in a small helper:

```python
def _release_cpu_storage(tensor: torch.Tensor) -> None:
    """Release CPU weight storage when the backing storage permits resizing."""
    try:
        tensor.untyped_storage().resize_(0)
    except RuntimeError as exc:
        # Safetensors-backed tensors can use non-resizable storage. The tensor
        # is detached from the module immediately afterward, so dropping its
        # reference is the correct cleanup for that case.
        if "not resizable" not in str(exc):
            raise
        logger.debug("CPU weight storage is non-resizable; dropping reference")
```

Then the direct `resize_(0)` calls at the relevant cleanup sites are replaced with `_release_cpu_storage(...)`.

The exception handling is intentionally narrow: unrelated `RuntimeError`s are still re-raised.

## Validation

After this change:

- the module compiled with `py_compile`,
- all Qwen3.8-27B checkpoint shards loaded successfully,
- the prior non-resizable-storage exception disappeared,
- XLA initialization / compilation completed,
- repeated `/v1/chat/completions` requests returned HTTP 200.

## Proposed upstream shape

I suggest keeping this separate from any TPU placement / sharding fix because the defect is a generic CPU-storage lifecycle assumption.

A focused unit test could provide a tensor backed by non-resizable storage (or mock `untyped_storage().resize_`) and verify that:

1. the known `"not resizable"` RuntimeError is tolerated;
2. other RuntimeErrors continue to propagate;
3. cleanup proceeds by allowing the tensor reference to be dropped.

GianluigiVitale:
Fixes #3400.

Qwen3.8 safetensors exposed an invalid assumption in the unquantized TPU
loader: every CPU backing storage can be resized to zero after transfer.
Safetensors-backed tensors may instead raise `RuntimeError: Trying to resize
storage that is not resizable`, aborting loading before the module can drop its
CPU tensor reference.

This change introduces a narrow helper that:

- preserves the existing zero-size fast path for resizable storage;
- tolerates only the known `RuntimeError` containing `not resizable`;
- re-raises unrelated runtime failures; and
- is used by every direct unquantized cleanup site, including Pathways dummy
  loading, linear weights/biases, and MoE weights/biases.

The helper does not promise that a non-resizable backing allocation is freed
immediately. It only permits the caller to continue and drop or replace its
tensor reference according to the existing lifecycle. FP8 cleanup is unchanged;
sharing its broader helper here would introduce the wrong dependency direction
and is outside this focused fix.

### Tests

Observed on TPU Inference
`76374c3c6d6a9575983e370b09ed49396367fc8d` with pinned vLLM LKG
`d626108b1841888ec90aced33367149a6bbc7e4b` and Python 3.12.13
(`jax==0.10.1`, `torch==2.10.0+cpu`, `torchax==0.0.11`,
`pytest==9.0.3`):

```text
pytest -q tests/layers/vllm/test_unquantized.py \
  -k 'release_cpu_storage or pathways_dummy_non_resizable'

4 passed, 244 deselected, 18 warnings in 24.75s
```

The focused tests cover real resizable storage, real `torch.frombuffer`
non-resizable storage, propagation of an unrelated `RuntimeError`, and the
production Pathways-dummy caller.

A CPU-backed whole-file attempt additionally reached 88 passes and 64 expected
skips. The remaining 96 existing `test_fused_moe` parametrizations uniformly
stopped at `Unsupported TPU device kind: cpu`; none of the storage-release
tests failed. Those unrelated TPU-kernel cases are left to upstream TPU CI.

Pinned static checks also passed:

```text
ruff 0.11.8: passed
YAPF 0.43.0: no diff
isort 6.0.1: passed
compileall: passed
git diff --check: passed
```

### Risk and rollback

Risk is limited to unquantized CPU-storage cleanup. The catch is deliberately
narrow and leaves FP8 behavior unchanged. Reverting this single commit restores
the prior behavior.

### Contributor checklist

- [x] Focused Linux pytest observed.
- [x] Pinned formatter/linter checks pass.
- [x] Commit includes `Signed-off-by`.
- [x] Adversarial review completed after caller-level regression coverage was
  added.
- [x] AI assistance disclosed: OpenAI Codex helped prepare this patch and PR
  text; the human submitter reviewed and remains responsible for every line.

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -64,6 +64,20 @@ P = PartitionSpec
 logger = init_logger(__name__)
 
 
+def _release_cpu_storage(tensor: torch.Tensor) -> None:
+    """Release resizable CPU storage and tolerate safetensors backing."""
+    try:
+        tensor.untyped_storage().resize_(0)
+    except RuntimeError as exc:
+        if "not resizable" not in str(exc):
+            raise
+        # Safetensors can expose non-resizable storage. Callers may still drop
+        # their tensor reference without mutating the backing storage.
+        logger.debug(
+            "CPU weight storage is non-resizable; deferring to reference cleanup"
+        )
+
+
 def _load_weight_for_layer(
     layer: torch.nn.Module,
     param_name: str,
@@ -107,7 +121,7 @@ def _load_weight_for_layer(
         # Dummy weights are created directly on the TPU mesh, no CPU→TPU transfer needed
         tensor_shape = tuple(tensor.shape)
         tensor_dtype = tensor.dtype
-        tensor.untyped_storage().resize_(0)
+        _release_cpu_storage(tensor)
         dtype = to_jax_dtype(tensor_dtype)
         return create_dummy_weights_on_tpu(
             sharding=sharding,
@@ -227,8 +241,8 @@ class VllmUnquantizedLinearMethod(vllm_linear.UnquantizedLinearMethod,
         weight = _load_weight_for_layer(layer, "weight", loading_sharding)
         weight = jnp.transpose(weight)
 
-        # Free CPU memory immediately
-        layer.weight.untyped_storage().resize_(0)
+        # Release resizable CPU storage before replacing the module parameter.
+        _release_cpu_storage(layer.weight)
         delattr(layer, 'weight')
         if layer.bias is not None and not layer.skip_bias_add:
             if layer.return_bias:
@@ -236,7 +250,7 @@ class VllmUnquantizedLinearMethod(vllm_linear.UnquantizedLinearMethod,
             bias_sharding = NamedSharding(self.linear_config.mesh,
                                           self.linear_config.bias_sharding)
             bias = _load_weight_for_layer(layer, "bias", bias_sharding)
-            layer.bias.untyped_storage().resize_(0)
+            _release_cpu_storage(layer.bias)
             delattr(layer, 'bias')
         else:
             bias = None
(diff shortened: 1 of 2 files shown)


GianluigiVitale:
Hi @kyuyeunk @gxd3 @jrplatin, this PR is ready for review.

It is a small fix for #3400 with tests for both resizable and non-resizable CPU storage.

I noticed the PR still has no `ready` label. Could someone take a look and add it if everything looks OK?

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -64,6 +64,20 @@
 logger = init_logger(__name__)
 
 
+def _release_cpu_storage(tensor: torch.Tensor) -> None:
+    """Release resizable CPU storage and tolerate safetensors backing."""
+    try:
+        tensor.untyped_storage().resize_(0)
+    except RuntimeError as exc:
+        if "not resizable" not in str(exc):
```
kyuyeunk: this doesn't look like a good fix? do you have any references on how this is typically resolved?

GianluigiVitale: Thanks for pointing this out. Updated in `10ab132c` to check `storage.resizable()` directly: resizable storage keeps the `resize_(0)` fast path; non-resizable storage uses `tensor.set_(torch.storage.UntypedStorage())` under `torch.no_grad()`.

The existing precedent is `_free_torch_storage` in `tpu_inference/layers/vllm/quantization/fp8.py`, which uses the same empty-storage detachment fallback. The `no_grad()` guard also makes this safe for leaf `Parameter` objects.

Removed exception-message matching and updated the tests to expect zero-sized tensor storage, including the Pathways dummy case. Added leaf `Parameter` coverage with and without gradients.

All 9 focused tests and formatting/lint checks pass. FP8 behavior is unchanged.

GianluigiVitale: @kyuyeunk

PR #3503, opened 2026-09-01
[Perf] Opt-in host staging for quantized MoE expert weights

lxhfirenking:
# Description

`process_quantized_moe_weights` opens by putting each of the four fp8 MoE weights at its expert sharding, but `_load_weight_for_layer` has already built the array with `t2j` — so the whole unsharded tensor lands on the default device first and is then sliced back apart. Every host stages the full tensor on its device 0 and keeps only the shards it owns; on a 32-device mesh that discards 31/32 of the transfer, and the discarded fraction grows with the mesh.

This adds a `stage_on_host` option that builds the array from a zero-copy numpy view of the torch storage instead, so every shard goes straight to the device that owns it. It is taken at the four fp8 MoE call sites — the ones whose very next step is a put at exactly that sharding, so the result is unchanged by construction.

This is what brings Qwen3.8-2.4T-A95B-FP8 weight loading on v7x-32 down from ~3 hours to ~30 minutes. Profiling a 4-host run put 76–83% of weight-load wall clock in this per-layer step; on one host, staging a 2.1 GB fp8 expert weight goes from 1.51s to 0.14s (1.4 → 14.9 GB/s).

Off by default. Set `MOE_STAGE_WEIGHTS_ON_HOST=1` to opt in — current weight-loading behavior is unchanged without it.

Weight-scale shapes are not always divisible along the axes their weight is, so an unshardable shape logs once and falls back to the existing path.

# Tests

New unit tests covering the three pieces:

- `_host_numpy_view` — shape/dtype/values match `t2j` for float32, bf16 and both fp8 formats; zero-copy for the dtype numpy represents natively; declines the cases the uint8 bit cast cannot serve (non-CPU storage, non-contiguous, 0-dim).
- `_load_weight_on_host` — bit-identical to staging on a device and then putting, across every dtype and three partition specs; ready before it returns, so the array survives the caller freeing the torch storage it aliases; returns `None` when the sharding does not divide the shape.
- The gate — `_load_weight_for_layer` stages on host only when asked and never under Pathways; `process_weights_after_loading` requests it for all four fp8 MoE weights per `MOE_STAGE_WEIGHTS_ON_HOST` and at the expert sharding that makes staging sound; the flag defaults off and is copied into the Ray workers.

```
pytest tests/layers/vllm/test_unquantized.py tests/layers/vllm/test_fp8.py \
       tests/platforms/test_tpu_platform.py tests/test_envs.py
```

Full runs of the two layer test files pass (346 passed, 100 skipped), and `test_fp8.py` also passes end-to-end with `MOE_STAGE_WEIGHTS_ON_HOST=1` — no numerical divergence.

# Checklist


tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -57,19 +58,94 @@
...
+    merely slower, though: the fallback is torchax's `t2j`, which widens bf16
+    and fp8 to float32 before it copies, so a weight that declines staging
+    needs 2-4x its own size in host memory on the way through.
+    """
+    try:
+        np_view = _host_numpy_view(tensor)
+        if np_view is None:
+            return None
+        array = general_device_put(np_view, sharding)
+    except Exception as e:
```
QiliangCui: This bare `except Exception` also swallows genuine runtime failures and reports them as a routine staging decline. Two concrete cases:

- HBM under pressure: `general_device_put` raising `RESOURCE_EXHAUSTED` (a `JaxRuntimeError`) gets one `warning_once`, then the t2j fallback re-attempts a strictly larger allocation (full unsharded tensor on one device, widened to float32 -- the 2-4x cost the docstring above prices out), so the eventual OOM is bigger and points away from the real cause.
- A genuine bug (e.g. a `general_device_put` signature change) would silently disable the optimization model-wide instead of failing loudly.

The graceful-decline cases this try is meant to catch raise only `ValueError`/`IndexError` (`IndivisibleError` subclasses `ValueError`; checked on jax 0.10.2), so narrowing keeps every fallback path working:

```suggestion
    except (ValueError, IndexError) as e:
```

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -57,19 +58,94 @@
...
+        logger.warning_once(
+            f"Host staging unavailable for {tuple(tensor.shape)}"
+            f" {tensor.dtype} at {sharding}: {e}")
+        return None
+    # np_view aliases the torch storage and callers free that storage as soon as
+    # this returns, so the asynchronous copy has to be forced here rather than
+    # left to whoever consumes the array.
+    jax.block_until_ready(array)
```
QiliangCui, line 134: `jax.block_until_ready(array)` sits outside the `try`, but an async H2D copy failure can surface at the await rather than at enqueue (PJRT attaches such errors to the buffer's definition event and raises at block time). In that case this line raises and the whole weight load crashes -- nothing up through `process_weights_after_loading` catches it -- whereas the same failure one line earlier would log and fall back to t2j, which is still sound here since the torch storage hasn't been freed yet. Moving the `block_until_ready` inside the `try` preserves the decline-not-raise contract for that case too.

tpu_inference/envs.py
```diff
@@ -317,6 +318,12 @@ def _get_int_list_env() -> list[int]:
     "MOE_REQUANTIZE_CLIP_PERCENTILE":
     lambda: float(pct)
     if (pct := os.getenv("MOE_REQUANTIZE_CLIP_PERCENTILE")) else None,
+    # Build each quantized MoE expert weight directly at its expert sharding
```
QiliangCui: This reads as if it covers quantized MoE methods generally, but only `VllmFp8MoEMethod` (and the deepseek_v4_fp8 method that reuses it) reads `MOE_STAGE_WEIGHTS_ON_HOST`. The nvfp4, compressed-tensors w4a8/w8a8-fp8, mxfp4, and awq MoE methods still stage the full unsharded tensor via plain t2j and ignore the flag with no warning -- e.g. `nvfp4.py` does its t2j loads and then `shard_moe_weights_to_tpu`, exactly the pattern this flag exists to avoid, and those checkpoints carry some of the largest expert weights. So setting the flag for an NVFP4/W4A8 model is a silent no-op today.

Either wire `stage_on_host` through the other quantized MoE methods (fine as a follow-up PR) or scope this comment to fp8 so the flag doesn't overpromise.

tests/layers/vllm/test_unquantized.py
```diff
@@ -816,3 +819,184 @@ def test_load_weight_for_layer_pathways_dummy(_, dtype):
...
+    result = _load_weight_on_host(tensor, sharding)
+
+    expected = jax.device_put(t2j(tensor, use_dlpack=False), sharding)
+    assert result is not None
+    assert result.dtype == expected.dtype
+    assert result.sharding == sharding
+    np.testing.assert_array_equal(np.asarray(result), np.asarray(expected))
+
+
+def test_load_weight_on_host_does_not_outlive_the_torch_storage():
```
QiliangCui: This test can't fail for the hazard it names: with `jax.block_until_ready` patched to a no-op, it still passes end-to-end (reproduced on a 4-device CPU backend). `set_(UntypedStorage())` only drops a reference -- the numpy-view/jax refchain keeps the bytes alive -- and the real in-place free (`resize_(0)`) is independently blocked because the numpy view marks the storage non-resizable. So if the `block_until_ready` is ever removed or deferred (a tempting tweak, since it serializes the four per-layer transfers), this guard stays green; the only remaining check is `assert result.is_ready()`, which is a timing probe.

Suggest asserting the mechanism instead (e.g. mock `jax.block_until_ready` and assert it was called on the staged array before return), or renaming the test to what it actually pins.

tests/platforms/test_tpu_platform.py
```diff
@@ -90,6 +91,17 @@ def test_basic_properties(self):
         assert TpuPlatform.get_device_communicator_cls(
         ) == expected_communicator
 
+    def test_additional_env_vars_covers_moe_stage_weights_on_host(self):
+        """A weight-loading flag the workers never see is a flag that does
+        nothing.
+
+        Ray copies only vLLM-prefixed variables plus this list into the worker
+        actors, and process_weights_after_loading runs there -- which is why
+        the MOE_REQUANTIZE_* flags beside it are listed too.
```
QiliangCui, line 100: Not quite complete: `MOE_REQUANTIZE_CLIP_PERCENTILE` -- read worker-side at `tpu_inference/layers/common/process_weights/moe_weights.py:741`, inside the same `process_quantized_moe_weights` this flag feeds -- is missing from `additional_env_vars`, and so is `DISABLE_WEIGHT_REQUANTIZATION` (read at `moe_weights.py:729`). By the mechanism this docstring describes, setting either on a Ray multihost deployment is honored on the driver and silently dropped in the workers, i.e. per-host quantization divergence. Pre-existing gap, but worth fixing while touching this list: add the two flags (and consider deriving the expectation from `envs.environment_variables` so worker-read flags can't silently drift out again).

tpu_inference/layers/vllm/quantization/unquantized.py
lxhfirenking: I dont quite agree wtih the accessment here. As your earlier comment suggested, the try-catch should not be a catch all, but rather only for cases where stage_on_host is not expected to work (e.g. due to dimension divisibility). If an exception happens during the H2D copy, we should fail loudly instead.

tests/platforms/test_tpu_platform.py
lxhfirenking: Will fix these two flags in a separate follow up PR. These additional  two flags are not relevent to the issues we try to fixed here.

tpu_inference/envs.py
lxhfirenking: Right, and thanks for checking the other methods -- I confirmed only `fp8.py` passes `stage_on_host`; nvfp4, compressed-tensors, mxfp4 and awq never do.

Scoped the comment to fp8 in 315dd5cf7 and named the methods that ignore the flag, so it stops overpromising. Wiring the rest through is worth doing given the expert-weight sizes there, but I would rather keep that a separate PR than grow this one.

tests/layers/vllm/test_unquantized.py
lxhfirenking: Updated.

Also dropped the value comparison and the `is_ready()` assert -- they are what made the old test look like a liveness check, and `test_load_weight_on_host_matches_device_staging` already covers correctness at that sharding.

tpu_inference/layers/vllm/quantization/unquantized.py
QiliangCui: This comment was written against the previous version, where the clause was `except Exception` — i.e. the function's contract was "any staging failure declines to t2j". Under that contract the placement was load-bearing, and both halves reproduce on a v6e-8 (jax 0.10.2): a 2 GiB sharded `device_put` returns from the try in 0.6 ms with `is_ready()=False` (the copy lands ~31 ms later at the block line), and a deferred failure passes the try cleanly and raises `jax.errors.JaxRuntimeError` at `block_until_ready` — past `except Exception`, since it's raised outside the try. Meanwhile an HBM OOM during the put surfaces at enqueue, so that version declined the common genuine failure to a `warning_once` while crashing on the deferred one: which behavior you got was PJRT surfacing timing, not policy.

With the narrowing from the other thread applied, agreed — the placement no longer changes behavior. Deferred errors are `RuntimeError`s, so they propagate loudly from either line, and enqueue-time OOM now propagates too: fail-loudly holds consistently at both surfacing points. That resolves this thread; the fix just landed in the other one.

QiliangCui (approved):

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -57,19 +58,94 @@
...
+    t = tensor.detach()
+    if t.device.type != "cpu":
+        return None
+
+    jax_dtype = _NUMPY_UNSUPPORTED_DTYPES.get(t.dtype)
+    if jax_dtype is None:
+        return t.numpy()
+    if not t.dim() or not t.is_contiguous():
+        return None
+    return t.view(torch.uint8).numpy().view(jax_dtype)
```
jrplatin, line 94: Do we want to hardcode torch.unit 8 here or should this be an arg to the function? Also, is this function useful elsewhere? Could move to a utils file

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -57,19 +58,94 @@
...
+    device. On one host that is merely wasteful. Across hosts it is much worse,
+    because every host stages the entire tensor on its device 0 and keeps only
+    the shards it owns -- on a 32-device mesh 31/32 of that transfer is thrown
+    away, and the discarded fraction grows with the mesh.
+
+    Staging on the host instead lets every shard go straight to the device that
+    owns it. Measured on one host against a 17.2 GB expert weight: 5.08s ->
+    0.99s (3.4 -> 17.3 GB/s), with more to gain multi-host.
+
+    Returns None if the weight cannot be staged this way -- most often a shape
```
jrplatin, line 113: If this returns None, what does that mean?

jrplatin (approved):

tpu_inference/layers/vllm/quantization/unquantized.py
lxhfirenking: See the function comments

> numpy has no scalar type for bf16 or the fp8 formats, so those travel as uint8 and are reinterpreted with the ml_dtypes type JAX already uses for them

This is just an intermediate type used for torch tensor to numpy view conversion, an implementation detail that should not be revealed to callers.

tpu_inference/layers/vllm/quantization/unquantized.py
lxhfirenking: It means it fails to stage it on host (e.g. due to dimension divisibility). This is mainly a fallback mechanism to default path for cases where not all tensors within the model can be staged on host. We only need this path to work properly for the routed experts weights to get most of its benefits.

sixiang-google (approved):

PR #949, opened 2025-10-27
Update docs to include installation guide with building from source.

RobMulla:
# Description

This PR restructures the getting started documentation for `tpu-inference`. The main change is the introduction of a new "Installation" guide that provides comprehensive instructions for installing the package. The "Quickstart" guide has been updated to focus on running the server and benchmarking, assuming the user has already installed the package.

The key changes are:
- A new `docs/getting_started/installation.md` file with detailed installation instructions.
- A new `docs/getting_started/tpu_setup.md` file with information on setting up Cloud TPUs.
- The `docs/getting_started/quickstart.md` file is now focused on running the server and benchmarking.
- The main `README.md` and `docs/README.md` files now point to the new installation guide.
- The `mkdocs.yml` file has been updated to reflect the new documentation structure.

This change improves the user experience by providing a clear and dedicated guide for installation, making it easier for new users to get started with `tpu-inference`.

# Tests

I have manually reviewed the rendered documentation to ensure that all links are correct and the content is displayed as expected.

PR #3504, opened 2026-09-01
Qwen3.8-2.4T-A95B-FP8 bring-up on v7x-32, and a 6.6x faster fp8 MoE weight load

lxhfirenking:
# Description

Brings up `Qwen/Qwen3.8-2.4T-A95B-FP8` on v7x-32 (model=16 x attn_dp=2, EP=32), plus the fixes that bring-up needed.

**Weight load: 3h34m -> 33m (6.6x).** `_load_weight_for_layer` ignored its `sharding` argument and used `t2j`, landing the whole 17.2 GB expert weight on one device; `general_device_put` then sliced it back apart, so every host staged the full tensor and kept 8 shards of 32. New `stage_on_host` option builds the array from a zero-copy numpy view so each shard goes straight to its owner. Taken only at the four fp8 MoE call sites whose next step is a put at exactly that sharding, and gated by `MOE_STAGE_WEIGHTS_ON_HOST` (default on; set 0 to revert). Per-layer processing 99-132s -> 2.5s.

**Weight-load profiling**, opt-in via `TPU_WEIGHT_LOAD_PROFILE=1`: splits load wall clock into stream / dispatch / load / process from inside the loader. Off by default costs one bool test.

**Eval harness.** MMLU/GPQA extractors took the *first* letter in the response, so a reasoning model that weighs "(B)" before settling on "(D)" was scored B. Strip the reasoning block before extraction, and report `unparsed` / `unparsed_rate` so a truncated-generation problem shows up as unparsed instead of as a silently lower accuracy.

**`run_multihost.sh`.** Tail the serve log every 60s while waiting (a multi-hour bring-up was previously indistinguishable from a hang), seed the health timeout from `VLLM_ENGINE_READY_TIMEOUT_S`, and add `MULTIHOST_ENV_VARS` / `MULTIHOST_DOCKER_ARGS` for env and `docker run` args that Ray workers need too.

**`.buildkite/dev/`**: dev-only pipelines for the v7x-32 run and the 397B de-risk steps. Not wired into PR CI.

# Tests

Build #953 on `tpu_v7x_32_queue`:
- Perf (in=8192, out=1024, n=128, concurrency 8): output throughput 178.1 tok/s, median TTFT 3167 ms, median TPOT 37.90 ms, 128/128, 0 failed. Repeatable to within 0.15% against #952.
- Accuracy (`reasoning_effort=low`): GPQA-Diamond 0.803, GSM8K 8-shot CoT 0.9363 flexible-extract, MMLU 0-shot 0.931 — all with <2% unparsed.
- Host staging verified numerically equivalent: every benchmark metric within 0.1% of the pre-change run (#949).

# Checklist


kyuyeunk:
what is the core changes vs. qwen3.5?

lxhfirenking:
This PR is created by mistake...it is just experimental changes not ready fot review.

PR #3511, opened 2026-09-02
[Sampler] Sample from vocab-sharded logits instead of all-gathering the vocab

gutianyu-google:
## Summary

`sample()` all-gathers the `(batch, vocab)` logits over the tensor-parallel axis on every decode step ("Unshard the logits explicitly…", with a TODO asking whether the vocab dimension could stay sharded) and then runs top-k / top-p / categorical on the replicated array, on every device. For a 151,936-token vocab that is a ~78 MB f32 gather per DP rank per step, followed by the 32–33 vocab-wide reduction sweeps of `topk_mask` and `topp_mask`, all redundant across the TP devices. In an xprof profile of a Qwen3-0.6B GRPO rollout (DP16 × TP8, TPU v7x) this was ~30% of the decode step's device time.

This PR makes sampling run on vocab-sharded logits. Each vocab shard keeps its top `VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES` (default 64) candidates and only those (values and global ids, ~0.4 MB) are gathered; top-k, top-p and the categorical draw run on the merged candidates.

**Exactness.** The kept set is the same as the replicated path's:
- the union of the per-shard top candidates contains the global top-k for `k <=` the candidate budget;
- ties at the k-th value are kept, like `topk_mask`;
- the top-p cutoff is the smallest prefix of the sorted probabilities with mass `>= p`, like `topp_mask`;
- candidates outside the top-k set are masked, so the softmax normalizes over exactly the same set (up to float32 summation order);
- greedy rows take the argmax with the lowest id among ties, like `jnp.argmax`.

The sampled tokens follow the same distribution as before but are not bitwise identical, since the random draw is over the candidates rather than the full vocab. Different batch shards fold their shard index into the key so they draw independent noise.

**Fallback.** Rows are exact when greedy, or when `0 < top_k <=` the candidate budget and `top_p > 0`. A batch containing any other sampling row takes the replicated path through a `lax.cond`; both branches leave the returned logits vocab-sharded, so a runtime fallback never adds the gather to the hot path. Padded request slots carry `DEFAULT_SAMPLING_PARAMS` (temperature −1, top_k 0), i.e. greedy, so they never force the fallback (in an earlier version of this change a guard on `top_k > 0` alone was silently tripped by the padded rows on every step).

**Constraints.** The returned logits are the raw, still-sharded logits. The only consumers of the processed logits are the `processed_*` logprobs modes, and only when logprobs were requested, so the runner and the decode loop pass `vocab_sharded=False` in exactly that case (`vocab_sharded_sampling_allowed`) and sampling stays on the replicated path. Sampled-token logprobs (`compute_and_gather_logprobs`) still run on the replicated logits; sharding that path is a natural follow-up.

**Measured** on TPU v7x GRPO rollouts (2048 sequences/step, generation cap 8192, top_k 50, top_p 1.0): −5.8 s/step at DP16 × TP8 (123.2 s → 118.0 s) and −9.5 s/step at DP32 × TP4 (99.7 s → 90.2 s), with reward and completion-length distributions unchanged across 20-step runs.

## Tests

`tests/layers/jax/sample/test_sampling.py::TestVocabShardedSampling` (mesh shards the vocab over every local device; run with `--xla_force_host_platform_device_count=4` to exercise the multi-shard merge on CPU):
- `_can_sample_vocab_sharded`: padded slots and greedy rows never force the fallback; a real row without a top-k filter or above the candidate budget does.
- greedy rows match `jnp.argmax`, including the lowest-id tie rule;
- sampled tokens always lie in the top-k / top-p set of `_apply_sampling_transforms`;
- over 8192 identical rows the empirical distribution matches the replicated path's probabilities within binomial noise, with zero mass outside the kept set;
- a batch with a `top_k=2000` row produces exactly the replicated path's tokens (same key);
- `vocab_sharded=False` keeps the replicated path;
- lowering `VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES` to 8 makes a `top_k=50` batch fall back to the replicated path's tokens;
- `vocab_sharded_sampling_allowed` picks the replicated path only for `processed_*` logprobs modes with logprobs on.

Also: `tests/test_envs.py` (`VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES` defaults to 64 and honors an override).

### DIFF
--- a/tpu_inference/envs.py
+++ b/tpu_inference/envs.py
@@ -44,6 +44,7 @@ if TYPE_CHECKING:
     CONTINUE_DECODE_EOS_CHECK_INTERVAL: int = 1
     USE_BATCHED_RPA_KERNEL: bool = False
     USE_BATCHED_RPA_SEQ_ON_LANE: bool = False
+    VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES: int = 64
     # Optional operator override for the RPA v3 kernel block sizes, one per
     # case. Each is a comma-separated 4-tuple (bq_sz, bkv_sz, bq_csz, bkv_csz).
     # Empty (default) = use the built-in tuned/heuristic sizes.
@@ -357,6 +358,12 @@ environment_variables: dict[str, Callable[[], Any]] = {
     env_bool("USE_BATCHED_RPA_KERNEL"),
     "USE_BATCHED_RPA_SEQ_ON_LANE":
     env_bool("USE_BATCHED_RPA_SEQ_ON_LANE"),
+    # Candidates each tensor-parallel shard keeps for vocab-sharded sampling.
+    # Rows with 0 < top_k <= this value are sampled exactly; a batch with a
+    # larger top_k falls back to sampling over the full vocab. Raise it if
+    # your requests use a larger top_k; larger values only grow the merge.
+    "VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES":
+    lambda: int(os.getenv("VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES") or "64"),
     # Optional operator override for RPA v3 kernel block sizes, per case.
     # Comma-separated 4-tuple: bq_sz,bkv_sz,bq_csz,bkv_csz. Empty = use the
     # built-in tuned/heuristic sizes. Lets operators retune the decode
--- a/tpu_inference/layers/jax/sample/sampling.py
+++ b/tpu_inference/layers/jax/sample/sampling.py
@@ -21,6 +21,7 @@ from jax.sharding import Mesh, NamedSharding
 from jax.sharding import PartitionSpec as P
 from vllm.v1.outputs import LogprobsTensors
 
+from tpu_inference import envs
 from tpu_inference.layers.common.binary_search import topk_mask, topp_mask
 from tpu_inference.layers.common.sharding import ShardingAxisName
 from tpu_inference.layers.jax.sample.sampling_metadata import \
@@ -32,6 +33,23 @@ if TYPE_CHECKING:
     from tpu_inference.runner.input_batch import CachedRequestState
 
 _SAMPLING_EPS = 1e-5
+
+
+def _num_vocab_sharded_candidates() -> int:
+    """Candidates each vocab shard contributes to vocab-sharded sampling.
+
+    Rows with 0 < top_k <= this many candidates are sampled exactly (see
+    _can_sample_vocab_sharded). Read from VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES
+    at trace time, so it is fixed for the life of the process.
+    """
+    num_candidates = envs.VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES
+    if num_candidates < 1:
+        raise ValueError("VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES must be >= 1, "
+                         f"got {num_candidates}")
+    return num_candidates
+
+
+_MASKED_LOGIT = -1e12
 
 
 @dataclass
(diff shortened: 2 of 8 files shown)


tpu_inference/envs.py
```diff
@@ -357,6 +358,16 @@ def _get_int_list_env() -> list[int]:
     env_bool("USE_BATCHED_RPA_KERNEL"),
     "USE_BATCHED_RPA_SEQ_ON_LANE":
     env_bool("USE_BATCHED_RPA_SEQ_ON_LANE"),
+    # Sample from vocab-sharded logits instead of all-gathering the full
```
lk-chen: Any drawback turning on this flag? if not we can always enable this and remove the flag

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -33,6 +34,12 @@
 
 _SAMPLING_EPS = 1e-5
 
+# Candidates each vocab shard contributes to vocab-sharded sampling. Rows with
+# 0 < top_k <= this many candidates are sampled exactly (see
+# _can_sample_vocab_sharded).
+VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES = 64
```
lk-chen: where does 64 come from

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -102,19 +109,220 @@ def _apply_sampling_transforms(
...
+    top_p > 0, since the union of the per-shard top candidates then contains
+    the whole top-k set and the top-p cutoff is computed over that set. Padded
+    request slots carry DEFAULT_SAMPLING_PARAMS (temperature -1, top_k 0), so
+    they count as greedy and never force the fallback.
+    """
+    is_greedy = tpu_sampling_metadata.temperature < _SAMPLING_EPS
+    top_k = tpu_sampling_metadata.top_k
+    exact = ((top_k > 0) & (top_k <= VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES) &
+             (tpu_sampling_metadata.top_p > 0.0))
+    return jnp.all(is_greedy | exact)
```
lk-chen: no need to `jnp.all` ? since all variables here are on CPU, you just need `is_greedy or exact`?

tpu_inference/envs.py
gutianyu-google: Good point, dropped the flag; sharded sampling is now the default. Two things to note: sampled tokens follow the same distribution as before but aren't bit-identical for a given seed, since the draw runs over the candidates; and rows with `top_k=0` or above the candidate budget still fall back to the old path.

tpu_inference/layers/jax/sample/sampling.py
gutianyu-google: It's how many candidates each shard keeps. It has to be fixed at compile time and must be at least the largest `top_k` you want handled exactly. 64 covers the usual values (20/40/50) and keeps the gather tiny. Made it configurable: `VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES`, default 64.

tpu_inference/layers/jax/sample/sampling.py
gutianyu-google: They're not CPU values: `temperature`, `top_k`, `top_p` are per-row device arrays inside the jitted `sample()`, so `is_greedy | exact` is a vector. `jnp.all` turns it into the single bool that `lax.cond` needs.

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -34,6 +35,23 @@
...
+    at trace time, so it is fixed for the life of the process.
+    """
+    num_candidates = envs.VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES
+    if num_candidates < 1:
+        raise ValueError("VOCAB_SHARDED_SAMPLING_NUM_CANDIDATES must be >= 1, "
+                         f"got {num_candidates}")
+    return num_candidates
+
+
+_MASKED_LOGIT = -1e12
```
patrickji2014, line 52: is this a common value for masked_logit, why don't we just use iinfo().min_val. thanks.

tpu_inference/layers/jax/sample/sampling.py
lk-chen: I mean can we use vocab_size // TP (which is much larger than typical top_k). If there's no drawback, let's enable it as much as possible.

lk-chen:
talked offline, close in favor of https://github.com/vllm-project/tpu-inference/pull/3490

PR #3514, opened 2026-09-02
[PCP] Move the host-side preprocessing into runner/pcp_utils.py

bhuvanpkaruturi:
> **Stacked on #3425.** Only the last commit (`[PCP] Move the host-side preprocessing into runner/pcp_utils.py`) belongs to this PR; the first seven are #3425 rebased onto current `main`. Kept as a draft until #3425 merges, then this rebases to a single commit.

# Description

Pure code motion, no behaviour change. Follows a review suggestion on #3425 to move the CP preprocessing logic into its own file in the runner folder.

Today the per-step PCP work that `_prepare_inputs` does inline — validating the batch as prefill-only, zigzag-chunking each request, permuting the token buffers into rank order, building the per-seq attention metadata and logits indices, and assembling `PCPMetadata` — is one of the largest single blocks in `tpu_runner.py`, and the pure layout helpers it uses sit in `layers/common/attention_metadata.py` next to the dataclass they build.

After this PR:

- **New `tpu_inference/runner/pcp_utils.py`** holds all of it. `PCPPreprocessor.prepare_inputs` is the old inline block; `_prepare_inputs` reduces to a single call. The layout helpers (`pcp_token_layout`, `pcp_seq_arrays`, `pcp_cache_page_buckets`, `round_up_pcp_cache_pages`) move here from `attention_metadata.py`, which keeps only the `PCPMetadata` dataclass — the layer-side contract stays under `layers/`, everything that builds it is runner-side. This mirrors how speculative decoding, structured decoding and multimodal preprocessing already live in their own runner modules.
- Two self-contained numpy loops become `pcp_token_permutation` and `pcp_last_token_slots`, and `pcp_batch_layout` folds the buffer-fit assert together with the single-request chunk override, so each step of the layout can be checked on its own without a runner.
- `PCPPreprocessor` is constructed in `_init_inputs` after the request-count ladder and owns the multi-request log counter. The token-bucket sizing uses `pcp_buffer_tokens`. The compilation manager and the interface test only change their imports.

`git diff --color-moved` shows the move; the runner-side call site is the only logic that is new.

# Tests

- New `tests/runner/test_pcp_utils.py` (39 checks, numpy plus one mesh test): every real token lands in exactly one slot, `kv_token_order` undoes the permutation on the live rows, the zigzag rank assignment, `cu_q_lens` / `q_pos_offsets` / `kv_new_starts` per seq, the logits slots, and `prepare_inputs` end to end on a mesh (skipped with fewer devices than `pcp_size`).
- On v7x: `tests/kernels/rpa_v3_cp/ragged_paged_attention_kernel_cp_test.py` (38 passed, 1 skipped) and `tests/layers/common/test_pcp_attention_interface.py` (25 passed) — identical counts to #3425 without this change — plus the new runner suite (39 passed).
- End-to-end greedy-token diff of a PCP=2 x TP=4 server against TP=8 (Qwen3-8B, 8k prompts, 4k chunk so the cache phase runs, up to 8 requests per step): behaviour is identical with and without this refactor on the same base, and 8/8 prompts match once the chunked-prefill fixes that follow #3425 are included.

# Checklist


bhuvanpkaruturi:
Superseded by #3557, which is opened from a branch in this repository so it can be stacked on #3425 and show only its own diff.

PR #3515, opened 2026-09-02
[Perf/Fix] Configure heterogeneous KV cache sharding for GDN/Mamba in model_loader

sierraisland:
# Description

Currently, `model_loader.py` defines a single attention-oriented `kv_cache_sharding = NamedSharding(mesh, PartitionSpec(ShardingAxisName.BATCH, ShardingAxisName.KV_CONTEXT, ShardingAxisName.KV_HEAD))` and applies it to all layers via JIT `out_shardings`.
For hybrid models with linear attention / GDN / Mamba layers (such as Qwen 3.5):
- GDN's `recurrent_state` (SSM state) has shape `[num_blocks, num_heads, d_k, d_v] = [1028, 32, 128, 128]`.
- Applying the attention spec forces dimension 2 (`d_k=128`) to be sharded across `KV_HEAD` (e.g. split into `[2, 64]` on TP=2), while `run_jax_gdn_attention` expects dimension 1 (heads) to be sharded.
- This layout/sharding mismatch forces XLA to insert a massive **257MB All-to-All** per GDN layer on every step, followed by an asynchronous SparseCore layout formatting offload (`sparse-core-data-format-call`, ~188 μs per layer). Across 30 GDN layers, this causes ~6 ms of pure overhead per step.

### Fix
Construct a heterogeneous `kv_cache_sharding` list matching the model's `layer_types`:
- Full Attention layers use `attn_kv_sharding`: `(BATCH, KV_CONTEXT, KV_HEAD)`.
- GDN layers use `gdn_kv_sharding`: `(conv_sharding, ssm_sharding)` with `recurrent_state` sharded as `PartitionSpec(ATTN_DATA, ATTN_HEAD, None, None)`.

Related Bug: b/549307194

# Tests

- Tested with Qwen-3.5-397B on TPU v7x-8.
- Traces collected with Xprof confirm that the 257MB All-to-All and SparseCore data formatting offloads have completely disappeared.

# Checklist

### DIFF
--- a/tpu_inference/models/common/model_loader.py
+++ b/tpu_inference/models/common/model_loader.py
@@ -353,10 +353,87 @@ def get_flax_model(
                                pooler=pooler,
                                is_draft_model=is_draft_model)
     vllm_config.model_config.dtype = original_dtype
-    kv_cache_sharding = NamedSharding(
+    attn_kv_sharding = NamedSharding(
         mesh,
         PartitionSpec(ShardingAxisName.BATCH, ShardingAxisName.KV_CONTEXT,
                       ShardingAxisName.KV_HEAD))
+    gdn_conv_sharding = NamedSharding(
+        mesh,
+        PartitionSpec(ShardingAxisName.ATTN_DATA, None,
+                      ShardingAxisName.ATTN_HEAD))
+    gdn_ssm_sharding = NamedSharding(
+        mesh,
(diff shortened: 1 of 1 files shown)


wenxindongwork:
thanks for the PR fix! 

I want to see this a simpler fix like this can achieve the same effect, do you mind trying it out? 
https://github.com/vllm-project/tpu-inference/pull/3519

Basically, mimicking what the torchax model loader does, which instead of explicitly setting the kv cache sharding just let XLA infer.

PR #3516, opened 2026-09-02
Fix the Raiden RL weight-sync path

A9isha:
# Description

Running tunix's GRPO example against this repo's RL sampler with `WEIGHT_SYNC_MODE=raiden` hit five separate faults. Each is small and independent; together they are the difference between the rollout worker aborting at startup and a clean end-to-end sync.

1. Engine-first Raiden load (`__init__.py`)

   The existing preload never ran engine-first and in fact never ran at all. It sat below `import tpu_inference.env_override`, and that module does `import vllm`, which transitively loads jaxlib's libjax_common.so -- so jaxlib's XLA copy was always already resident by the time the preload was reached. It also imported `tpu_raiden`, but the raiden_jax wheel's distribution is named `tpu_raiden_jax` and the package it installs is `tpu_sync`, so the import raised ModuleNotFoundError and was silently swallowed every time.

   Each Raiden extension module statically links its own XLA runtime and its own copy of the `execute_options.proto` descriptor, so whichever loads first wins the protobuf registry and a later one aborts the process:

       F message.cc:348] File is already registered:
         xla/pjrt/proto/execute_options.proto

   or, on newer Raiden builds, `free(): invalid pointer`. Move the preload above env_override, try both package spellings, and load all of Raiden's modules rather than just the engine -- the FFI modules are imported at module scope by the synchronizer, so they would otherwise arrive mid-run, long after libtpu.

2. Torn weights on the destination (`rl/raiden_worker_sync.py`)

   The destination bound with `auto_h2d` left at its default of False, so `h2d()` unpacked whatever happened to be staged at the moment it was called and tensors still in flight were installed torn. This is nastily silent: per-tensor checksums match the source for some tensors and sit partway between the initial and synced values for others, varying run to run as `parallelism` streams race. Bind with `auto_h2d=True` and ingest each slice as it lands, which is what tunix's in-process destination already does.

3. Unverifiable checksums (`rl/raiden_worker_sync.py`)

   `checksums()` sampled three tensors, which says nothing about how much of the model arrived intact. Add `__grand_total__` over every bound tensor plus `__tensor_count__` and `__element_count__`, so the two sides' totals are comparable only when they actually bound the same tensors rather than being assumed to be.

4. `None`-valued kwargs defeating sampling defaults (`rl/vllm_sampler.py`)

   `_get_val` already coalesces a present-but-None attribute to its default, but the default *expression* was `kwargs.get(key, fallback)`. Callers pass the whole generation-kwargs dict through with unset fields present as None, so `.get` finds the key, returns None, and defeats the default before `_get_val` ever sees it. Downstream this surfaces as `TypeError: '<' not supported between instances of 'NoneType' and 'int'` on `self.top_k < -1`. Route the kwargs lookups through `_get_val` too.

5. Missing prompt_token_ids (`rl/vllm_sampler.py`)

   `prompt_token_ids` lives on the RequestOutput, not the CompletionOutput, and was never copied out. Tunix's SamplingResponse carries the field and its collector feeds it straight into `np.asarray(..., dtype=np.int32)`, so omitting it surfaces as "int() argument must be ... not 'NoneType'". Populate it on all three construction sites, including the two empty/error paths.

# Tests

Verified end to end on a v5p-8 host running tunix's math_gsm8k_dist GRPO example (Qwen3-0.6B, MaxText trainer, vLLM rollout, TP=2): the trainer's source checksums and the vLLM destination's match exactly -- `__grand_total__` 13217973.204956055 over 310 tensors / 596049920 elements on both sides.Start with a short description of what the PR does and how this is a change from
the past.

# Checklist

### DIFF
--- a/tpu_inference/__init__.py
+++ b/tpu_inference/__init__.py
@@ -12,30 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Engine-first XLA load. Each Raiden module statically links its own XLA and its
+# own copy of `xla/pjrt/proto/execute_options.proto`; whichever is dlopened first
+# wins the protobuf registry and a later one aborts the process. Must precede
+# env_override, which does `import vllm` and brings up jaxlib's XLA. All modules,
+# not just the engine: the FFI ones are imported at module scope by tunix's
+# synchronizer, so they would otherwise land mid-run. The wheel is distributed
+# as `tpu_raiden_jax` but installs `tpu_sync`.
+for _mod in ("_tpu_raiden_jax", "_weight_synchronizer_ffi",
+             "_kv_cache_manager_ffi"):
+    try:
+        __import__(f"tpu_sync.frameworks.jax.{_mod}")
+    except ImportError:
+        # Raiden absent (e.g. the benchmark client), or an older build that
+        # does not ship this module. Only EngineCore workers need the load.
+        pass
+
 # The environment variables override should be imported before any other
 # modules to ensure that the environment variables are set before any
 # other modules are imported.
-import tpu_inference.env_override  # noqa: F401
-from tpu_inference import envs
-
-# Engine-first XLA load: import Raiden's engine extension here -- the earliest
-# tpu_inference entry point, before any submodule (e.g. platforms/tpu_platform.py)
-# runs `import jax` and loads jaxlib's libjax_common.so. This ensures Raiden's
-# embedded XLA runtime comes up before jaxlib's so the two XLA copies don't collide
-# in static initializers.
-try:
-    import tpu_raiden.frameworks.jax._tpu_raiden_jax  # noqa: F401
-except ModuleNotFoundError:
-    # Expected in processes without tpu_raiden on the path (e.g. the benchmark
-    # client). Only the EngineCore workers need the engine-first load; skip quietly.
-    pass
-except Exception as _raiden_exc:  # pragma: no cover - best-effort preload
-    import sys as _sys
-    print(f"[tpu_raiden] engine preload failed: {_raiden_exc}",
-          file=_sys.stderr)
-
-from tpu_inference import tpu_info as ti
-from tpu_inference.logger import init_logger
+import tpu_inference.env_override  # noqa: F401,E402
+from tpu_inference import envs  # noqa: E402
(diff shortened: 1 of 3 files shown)


tpu_inference/__init__.py
```diff
@@ -12,27 +12,28 @@
...
 # limitations under the License.
 
+# Engine-first XLA load. Each Raiden module statically links its own XLA and its
+# own copy of `xla/pjrt/proto/execute_options.proto`; whichever is dlopened first
+# wins the protobuf registry and a later one aborts the process. Must precede
+# env_override, which does `import vllm` and brings up jaxlib's XLA. All modules,
+# not just the engine: the FFI ones are imported at module scope by tunix's
+# synchronizer, so they would otherwise land mid-run. The wheel is distributed as
+# `tpu_raiden_jax` but installs `tpu_sync`; `tpu_raiden` is the older spelling.
+for _pkg in ("tpu_sync", "tpu_raiden"):
```
khatwanimohit: we can keep only tpu_sync here

wenxindongwork (approved):

tpu_inference/__init__.py
A9isha: done - thanks!

tpu_inference/__init__.py
```diff
@@ -12,30 +12,29 @@
...
 # limitations under the License.
 
+# Engine-first XLA load. Each Raiden module statically links its own XLA and its
+# own copy of `xla/pjrt/proto/execute_options.proto`; whichever is dlopened first
+# wins the protobuf registry and a later one aborts the process. Must precede
+# env_override, which does `import vllm` and brings up jaxlib's XLA. All modules,
+# not just the engine: the FFI ones are imported at module scope by tunix's
+# synchronizer, so they would otherwise land mid-run. The wheel is distributed
+# as `tpu_raiden_jax` but installs `tpu_sync`.
+for _mod in ("_tpu_raiden_jax", "weight_synchronizer_ffi",
```
sixiang-google: From Dateng: should be ```_weight_synchronizer_ffi```.

A9isha: that is correct, thank you fixed it now

QiliangCui (approved):

PR #3523, opened 2026-09-03
build(docker): build and package vllm-rs Rust frontend for TPU

aojea:
### Summary

When vLLM introduced the Rust frontend (`vllm-rs`) in [vllm#43283](https://github.com/vllm-project/vllm/pull/43283), commit `b9f1475d4e7` added `setuptools-rust` to fix the immediate Docker build break. However, `protobuf-compiler` (`protoc`) and invoking `build_rust.sh` were omitted from the build process.

Because vLLM's `setup.py` marks Rust extensions with `optional=True` when building without precompiled binaries, `setup.py` silently skipped building the `vllm-rs` binary when the required build steps were missing, leaving the published `vllm/vllm-tpu` images without `vllm-rs`.

This PR recovers the synchronization and alignment with upstream vLLM:
1. Installs `protobuf-compiler` in `docker/Dockerfile` and `docker/Dockerfile.pypi`.
2. Installs `requirements/build/rust.txt` and runs `bash build_rust.sh` before installing vLLM.
3. Ensures `build_vllm_tpu.sh` builds `vllm-rs` before wheel packaging.

### Validation
Verified that `vllm-rs` compiles successfully in the TPU Docker build and tested end-to-end inference serving on real Google Cloud TPU hardware (v5e).

### DIFF
--- a/.buildkite/scripts/build_vllm_tpu.sh
+++ b/.buildkite/scripts/build_vllm_tpu.sh
@@ -127,6 +127,9 @@ fi
 # --- Step 3: Execute the vLLM TPU build script ---
 echo "Ensuring 'build' package is installed..."
 pip install build
+echo "Installing Rust build requirements and building Rust frontend..."
+pip install -r requirements/build/rust.txt
+bash build_rust.sh
 echo "Executing the vLLM TPU build script..."
 bash tools/vllm-tpu/build.sh "${VLLM_TPU_VERSION}"
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,6 +37,7 @@ RUN apt-get update && apt-get install -y \
     bc \
     ca-certificates \
     unzip \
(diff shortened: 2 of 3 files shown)


.buildkite/scripts/build_vllm_tpu.sh
```diff
@@ -127,6 +127,9 @@ fi
 # --- Step 3: Execute the vLLM TPU build script ---
 echo "Ensuring 'build' package is installed..."
 pip install build
+echo "Installing Rust build requirements and building Rust frontend..."
+pip install -r requirements/build/rust.txt
+bash build_rust.sh
```
charlesli640, line 132: build_rust.sh does not exist in tpu-inference repo

aojea: Thanks, I was honestly mostly copy pasting and using gemini to figuring my way through this

PR #3546, opened 2026-09-09
Raiden rollout weight sync: KV-cache filtering, completion barrier

khatwanimohit:
# Description

The rollout side could register with Raiden but a sync round was neither complete nor correct. Four independent fixes, all on the destination (rollout) side; the trainer uses tunix's own synchronizer.

1. Don't bind KV-cache leaves. They live in the same state tree as the weights and are ordinary float arrays, so `_bindable` accepted them, but the trainer has no counterpart and the controller pairs by name. On qwen3.5-35b this surfaced 150 destination-only entries in the manifest preflight. `_filter_bindable` now drops any path containing `['cache']` and logs the count.

2. Wait for the H2D DMA to land. Raiden's `h2d()` is asynchronous and writes the bound buffers outside the JAX graph, so `jax.block_until_ready` is not a completion barrier -- the rollout could resume generating from half-written weights. `_wait_until_settled` polls a whole-model digest until it stops changing. Interim: `WeightSynchronizer` exposes no wait/join today. Gated by `RAIDEN_H2D_SETTLE` (default on).

3. Re-point the dispatch view after a sync. `model_fn` takes `state_leaves`, derived from `state` at load time; anything that rebinds `state` must re-derive it. Added `TPUModelRunner.refresh_state_leaves()` and exposed it as `TPUWorker.refresh_model_state_leaves` / `RLVllmSampler.refresh_model_state_leaves`, which tunix's `vllm_sampler_adapter` calls after `raiden_h2d()` (it warns loudly when the method is absent). `runner/lora_utils.py` now calls the same helper instead of open-coding it.

4. Thread `job_name` through `RLVllmSampler.bind_raiden_sync` -> `TPUWorker.bind_raiden_sync` so independent rollout replicas register as separate Raiden jobs rather than as hosts of one job. Re-binding updates the name on the existing synchronizer, which otherwise re-registered under the first one.

Supporting changes: `_grand_total` computes the verification checksum over every bound tensor in one stacked device sync accumulated in float64 (the previous per-tensor float32 sum lost the low digits at 35b scale); the module uses `init_logger` so its records reach vLLM's configured handlers; `RAIDEN_H2D_SETTLE` is declared in `envs.py` so it accepts `true`/`false` as well as `1`/`0`.

# Tests

qwen3.5-35b GRPO Raiden run on bodaborg-v5p-nap (MaxText/Pathways trainer 2x2x2 + 2 vLLM rollouts 2x2x1) completes with EXIT_CODE=0, final step 1, policy version 3, reward mean=0.7500 std=0.4330, 150 non-weight leaves dropped per bind. Unit tests: 20 passed.

# Checklist

### DIFF
--- a/tpu_inference/rl/raiden_worker_sync.py
+++ b/tpu_inference/rl/raiden_worker_sync.py
@@ -15,13 +15,21 @@ tunix's `weight_sync.dict_to_metadata` defines the metadata shape.
 from __future__ import annotations
 
 import socket
+import time
 from typing import Any, List, Optional, Tuple
 
 import jax
 import jax.numpy as jnp
+import numpy as np
+
+from tpu_inference import envs
+from tpu_inference.logger import init_logger
+
+logger = init_logger(__name__)
 
 _ws_lib: Any = None
 _RAIDEN_IMPORT_ERROR: Optional[Exception] = None
+
 try:
     from tpu_sync.api.jax import \
         weight_synchronizer as _ws_lib  # pylint: disable=g-import-not-at-top
@@ -106,17 +114,33 @@ def _bindable(arr: Any) -> bool:
         return False
 
 
+# KV-cache leaves are ordinary float arrays in the weight tree, so _bindable
+# accepts them, but the trainer has no counterpart to pair them with.
+_NON_WEIGHT_PATH_PARTS = ("['cache']", )
+
+
+def _is_weight(name: str) -> bool:
+    return not any(part in name for part in _NON_WEIGHT_PATH_PARTS)
+
+
 def _filter_bindable(names: List[str],
                      arrays: List[Any]) -> Tuple[List[str], List[Any]]:
-    """Drops leaves the native layer cannot bind (e.g. RNG-key arrays)."""
+    """Drops leaves the native layer cannot bind, and non-weight state."""
     keep_names: List[str] = []
     keep_arrays: List[Any] = []
+    dropped_non_weight = 0
     for name, arr in zip(names, arrays):
+        if not _is_weight(name):
+            dropped_non_weight += 1
+            continue
         if _bindable(arr):
             if hasattr(arr, "block_until_ready"):
                 arr.block_until_ready()
             keep_names.append(name)
             keep_arrays.append(arr)
+    if dropped_non_weight:
+        logger.info("skipped %d non-weight leaves (KV cache) when binding",
+                    dropped_non_weight)
     return keep_names, keep_arrays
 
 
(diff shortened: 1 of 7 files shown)


tpu_inference/rl/raiden_worker_sync.py
```diff
@@ -106,17 +114,36 @@ def _bindable(arr: Any) -> bool:
         return False
 
 
+# KV-cache leaves live in the same state tree as the weights and are ordinary
+# float arrays, so _bindable accepts them -- but the trainer has no counterpart,
+# and the controller pairs by name. qwen3.5-35b's attention layers surface
+# `attention.cache.cached_prefill_{key,value}`, which showed up as 240
+# destination-only entries in the manifest preflight.
```
wenxindongwork: nit: could we clean up the comments

tpu_inference/rl/raiden_worker_sync.py
```diff
@@ -246,6 +323,12 @@ def metadata_dict(self) -> dict:
             _tensor_metadata_dict(name, arr, idx)
             for idx, (name, arr) in enumerate(zip(self.names, self.arrays))
         ]
+        # Advertise the *raw* mesh: `_tensor_metadata_dict` derives each
+        # variable's `sharding_spec` from the array's own PartitionSpec, which
+        # names the original axes (and comma-joins multi-axis dims). Dropping
+        # size-1 axes here would leave those specs referring to axes absent
+        # from `mesh_axes`, which tunix's `_validate_metadata` rejects as
+        # "names unknown mesh axis".
```
wenxindongwork: nit: i think we can remove this comment

tpu_inference/rl/raiden_worker_sync.py
```diff
@@ -148,6 +175,25 @@ def _tensor_metadata_dict(name: str, arr: Any, layer_idx: int) -> dict:
     }
 
 
+def _grand_total(arrays: List[Any]) -> float:
```
wenxindongwork: nit: perhaps a better function name could be `_l1_norm` or `_sum_of_elements`?

tpu_inference/rl/raiden_worker_sync.py
```diff
@@ -232,7 +309,7 @@ def total(arr):
             name: total(arr)
```
wenxindongwork: nit: instead of total(arr) it could just be `_l1_norm([arr])`

tpu_inference/rl/raiden_worker_sync.py
```diff
@@ -210,6 +256,37 @@ def h2d(self) -> None:
         # transferred data.
         self._require_sync("h2d()").h2d()
         jax.block_until_ready(self.arrays)
+        # `block_until_ready` only orders JAX computations. Raiden's H2D is
+        # documented as asynchronous and writes these buffers via DMA outside
+        # the JAX graph, so it is not a completion barrier -- without the wait
+        # below the rollout can resume generating from half-written weights.
+        # Measured settle time is 2-4s for Qwen3-0.6B.
```
wenxindongwork: could we remove "Measured settle time is 2-4s for Qwen3-0.6B."?

wenxindongwork (approved):

kyuyeunk (approved):

PR #3548, opened 2026-09-09
Consider a name change to JAX-TPU

mtsokol:


mtsokol:
Hi @RobMulla,

This PR is a proposition of renaming TPU Inference to JAX-TPU.

We could have this migration in the docs regardless whether the project gets a rename.

Here's a preview: https://vllm--3548.org.readthedocs.build/projects/tpu/en/3548/
The new top-left corner logo design:
obraz

PR #3550, opened 2026-09-09
[Prefix Caching] Fix mamba cache replication issue 

wenxindongwork:
> [!NOTE]
> **This PR re-lands #3517 in addition to its own changes.**
> #3517 ("Decoupled Dual Block Pools for Hybrid Mamba/GDN Prefix Caching in
> Align Mode") was reverted from main by #3551, and this PR builds directly on
> it. 

## Summary

This PR mirrors mamba kv-cache groups onto one set of block ids, which effective reduces the mamba cache usage by 3x for Qwen3.5-397B. 

vLLM gives each kv-cache group its own block ids, so a model whose mamba layers are split into G groups spends G ids per logical checkpoint. Qwen3.5-397B has 45 mamba layers in 3 groups, so a pool of N ids held **N/3** checkpoints while the memory behind it holds N — every mamba layer has N slots and a checkpoint needs one slot in each. The split exists so an upstream `KVCacheTensor` can reinterpret one block's bytes as a different layer per group; TPU cannot overlay typed JAX arrays and already allocates one array per layer, so it bought nothing and cost the factor of 3.

Non-primary mamba groups are now bound to a `MirrorMambaBlockPool` that hands them the ids the primary group just allocated, and the pool canonicalises cache keys to the primary group so a lookup spanning all three groups resolves to one entry. `can_allocate_tokens` no longer counts mirrored demand. The lockstep between groups is an ordering property, so `replay_last_allocation` asserts the counts match rather than trusting it — aliasing the wrong id would corrupt state silently.

Effect on Qwen3.5-397B-A17B-FP8 (65-block pool per rank, multiplier 16, 371 requests of an agentic multi-turn trace):

| | before | after |
|---|---:|---:|
| prefill residency | 63 of 65 blocks | 19 |
| retained checkpoints | 10–13 | 52–61 |
| mamba lookup hit rate | 9.9% | 66–95% |
| engine prefix cache hit rate | 79.3% | 92.9% |
| prefill work | 89% of prompt tokens | 6.7% |

This PR also makes the following smaller changes. 

### 1. Simplify the mamba block reservation

The align-mode pool was sized by a chain of special cases: an absolute `custom_mamba_cache_size` checkpoint budget on top of active slots. Replaced with a per-request budget:

```
mamba_num_blocks = blocks_per_request * max_num_reqs + 1
```

`blocks_per_request` is `DEFAULT_MAMBA_CACHE_MULTIPLIER` (8) in align mode, overridable with `--additional-config '{"custom_mamba_cache_multiplier": N}'`. **This flag replaces `custom_mamba_cache_size`**, which is removed. Without prefix caching there is nothing to checkpoint, so the multiplier does not apply.  

### 2. Fix align-mode mamba over-allocation (HBM OOM)

`initialize_kv_cache` sized the mamba arrays at `tensor_num_blocks` — the *attention* pool — whenever `mamba_cache_mode == "align"`, blowing the budget the sizing pass had just computed. On Qwen3.5-397B-A17B-FP8 the pass budgeted 388 mamba blocks (69.4 GiB) and allocation then asked for 29588 (14.70 GiB per layer per device), dying on the first of 45 layers:

```
RESOURCE_EXHAUSTED: Attempting to allocate 14.45G. That was not possible. There are 903.14M free.
```

Align mode could not start on that model before this change.

## Benchmarks

Agentic benchmark, Qwen3.5-397B-A17B-FP8, TP=8 with DP attention (`attn_dp_size=4`) + expert parallel, `--max-num-seqs=16`, `--max-model-len=65536`, `--block-size=256`, `--prefix-cache-retention-interval 256`.

Total tokens/sec, across the five client workloads:

| client | mult8 | **mult16** | mult16 + `DP_SCHED_BATCH_PREFILL=true` |
|---|---:|---:|---:|
| `--num-groups 1 --concurrency 1 --group-size 8` | 38,379 | 38,110 | 35,118 |  
| `--num-groups 1 --concurrency 1` | 60,538 | 60,474 | 54,366 |  
| `--num-groups 2 --concurrency 2` | 75,271 | 76,691 | 68,307 |  
| `--num-groups 4 --concurrency 4` | 68,814 | **83,846** | 66,995 |  
| `--num-groups 8 --concurrency 8` | 31,206 | **51,580** | 43,025 | 

**Best configuration: `custom_mamba_cache_multiplier: 16`, DP+EP, `DP_SCHED_BATCH_PREFILL=false`.**

### Server command

```bash
NUM_PRECOMPILE_WORKERS=8 MODEL_IMPL_TYPE=vllm USE_MOE_EP_KERNEL=0 \
ATTN_BUCKETIZED_NUM_REQS=true ATTN_CUSTOM_NUM_REQS_BUCKETS=4 \
ONEHOT_MOE_PERMUTE_THRESHOLD=32768 VLLM_MOE_CHUNK_SIZE=256 SLICE_ROPE_CACHE=1 \
DP_SCHED_BATCH_PREFILL=false NEW_MODEL_DESIGN=1 \
LIBTPU_INIT_ARGS=' --xla_tpu_use_minor_sharding_for_major_trivial_input=true --xla_tpu_enable_sparse_core_collective_offload_reduce_scatter=false --xla_tpu_ars_combiner_threshold_in_bytes=0 --xla_tpu_enable_async_collective_merger=false --xla_tpu_check_legacy_constraints_in_reduce_scatter_legalizer=false' \
vllm serve Qwen/Qwen3.5-397B-A17B-FP8 \
  --max-model-len=65536 --max-num-batched-tokens=2048 --max-num-seqs=16 \
  --enable-prefix-caching \
  --gpu-memory-utilization=0.9 --tensor-parallel-size=8 --async-scheduling --port=8000 \
  --language-model-only --enable-auto-tool-choice --tool-call-parser=qwen3_coder \
  --reasoning-parser=qwen3 --default-chat-template-kwargs '{"enable_thinking": false}' \
  '--limit-mm-per-prompt={"image": 0, "video": 0}' --kv-cache-dtype=bfloat16 \
  --additional_config='{"sharding": {"sharding_strategy": {"enable_dp_attention": true, "attn_dp_size": 4}}, "custom_mamba_cache_multiplier": 16}' \
  --block-size=256 --enable-chunked-prefill \
  --prefix-cache-retention-interval 256 \
  --enable-expert-parallel
```

### Client command

```bash
python scripts/vllm/benchmarking/agentic_benchmark/benchmark_agentic.py \
  --model-path-or-id Qwen/Qwen3.5-397B-A17B-FP8 \
  --trace-file gs://wenxindong-vm/rl/mlperf2026/agentic_benchmark/gbs1024_trace_file_tool_time.jsonl \
  --global-prefix-len 6476 \
  --num-groups 4 --concurrency 4
```

## Quality (lm_eval)

gsm8k against the running server (mirroring on, align mode, prefix caching on):

| Task | Filter | n-shot | Metric | Value | Stderr |
|---|---|---:|---|---:|---:|
| gsm8k | flexible-extract | 5 | exact_match | **0.99** | ±0.01 |
| gsm8k | strict-match | 5 | exact_match | **0.99** | ±0.01 |

```bash
lm_eval --model local-chat-completions \
  --model_args base_url="http://localhost:8000/v1/chat/completions",model="Qwen/Qwen3.5-397B-A17B-FP8",eos_string="<|im_end|>",num_concurrent=1,max_retries=3,tokenized_requests=False \
  --tasks gsm8k --gen_kwargs max_gen_toks=2048 --apply_chat_template --fewshot_as_multiturn \
  --limit 100
```

### DIFF
--- a/tests/core/test_dp_scheduler.py
+++ b/tests/core/test_dp_scheduler.py
@@ -55,6 +55,7 @@ class TestDPScheduler:
         config.cache_config = MagicMock()
         config.cache_config.enable_prefix_caching = False
         config.cache_config.prefix_caching_hash_algo = "sha256"
+        config.cache_config.mamba_num_blocks = None
         return config
 
     @pytest.fixture
@@ -129,6 +130,41 @@ class TestDPScheduler:
                 # Verify processes were started
                 mock_process = mock_ctx.Process.return_value
                 assert mock_process.start.call_count == 2
+
+    def test_init_with_mamba_num_blocks_per_rank(
+        self,
+        mock_vllm_config,
+        mock_kv_cache_config,
+        mock_structured_output_manager,
+    ):
+        """Test that mamba_num_blocks is partitioned per DP rank from cache_config."""
+        mock_vllm_config.cache_config.mamba_num_blocks = 60
+        scheduler = self._create_scheduler(
+            mock_vllm_config,
+            mock_kv_cache_config,
+            mock_structured_output_manager,
+        )
+        assert len(scheduler.per_rank_kv_cache_configs) == 2
+        for rank_config in scheduler.per_rank_kv_cache_configs:
+            assert rank_config.num_blocks == 50
+            assert rank_config.mamba_num_blocks == 30
+
+    def test_init_without_mamba_num_blocks(
+        self,
+        mock_vllm_config,
+        mock_kv_cache_config,
+        mock_structured_output_manager,
+    ):
+        """Test that rank configs have no mamba_num_blocks if not configured."""
+        scheduler = self._create_scheduler(
+            mock_vllm_config,
+            mock_kv_cache_config,
+            mock_structured_output_manager,
+        )
+        assert len(scheduler.per_rank_kv_cache_configs) == 2
+        for rank_config in scheduler.per_rank_kv_cache_configs:
+            assert rank_config.num_blocks == 50
+            assert getattr(rank_config, "mamba_num_blocks", None) is None
 
     def test_init_with_prefix_caching_enabled(
         self,
--- a/tests/core/test_hybrid_coordinator.py
+++ b/tests/core/test_hybrid_coordinator.py
@@ -1,0 +1,630 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+from vllm.v1.core.block_pool import BlockPool
+from vllm.v1.core.kv_cache_utils import (BlockHash,
+                                         make_block_hash_with_group_id)
+from vllm.v1.kv_cache_interface import (FullAttentionSpec, KVCacheConfig,
+                                        KVCacheGroupSpec, MambaSpec)
+from vllm.v1.request import Request, RequestStatus
+
+from tpu_inference.core.hybrid_coordinator import (
+    MambaBlockPool, MirrorMambaBlockPool, TPUDualBlockPool,
+    TPUHybridKVCacheCoordinator, TPUKVCacheManager,
+    install_hybrid_coordinator_hooks, set_mamba_num_blocks)
+
+
+def _make_mock_hybrid_kv_cache_config(
+    num_attn_blocks: int = 500,
+    mamba_num_blocks: int | None = 50,
+    block_size: int = 16,
+) -> KVCacheConfig:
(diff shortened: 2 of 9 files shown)


sierraisland (approved):
LGTM

mrjunwan-lang (approved):

gxd3 (approved):

PR #3556, opened 2026-09-10
[PCP] Page-align chunks, unshuffle K/V in the kernel, merge the single-request path

bhuvanpkaruturi:
Stacked on #3425 (multi-request PCP) and #3514 (preprocessing module); only the last 4 commits are new. Review them in order:

1. kernel: page-granular indirection for the new-KV fetch
2. page-align multi-request chunks; unshuffle K/V in the kernel
3. merge the single-request path into the general layout
4. tests

## What this does

Rounds every request's zigzag chunk up to a whole number of KV pages (`C_i = round_up(ceil(n_i / 2P), block_size)`). That makes every token-order page of the all-gathered current K/V contiguous in rank order, so the kernel can unshuffle K/V through a small per-page map (`kv_page_order`) during the HBM→VMEM fetch it performs anyway — one page-sized DMA per page, exactly like the paged-cache side of a mixed fetch. The per-layer HBM→HBM gather pass (`kv_token_order` + `jnp.take` over the whole token buffer) is deleted.

With chunks page-aligned for any request count, the single-request special case (kernel arithmetic remap via `pcp_chunk_size`, the whole-buffer cache-phase span) becomes the R = 1 instance of the general layout and is deleted too: one layout, one metadata contract, one kernel fetch path (net −100 lines before tests).

## Measured (TPU v7x, Qwen3-8B shapes)

- Current-phase attention layer in isolation: **1.34–2.49× faster**; outputs bitwise identical to the gather path in every case. The deleted gather pass alone is ~450–540 µs/layer; the in-kernel map costs nothing measurable over the bare all_gather + attention.
- E2E serving (PCP2×TP4, ctx 8k): **−7% TTFT / +6% prefill tok/s** at concurrency 4–8 where multi-request batches form; exact parity on single-request steps (the condition for merging the paths).
- Padding: +2.15% attended rows at the observed batch, bounded by `2P·(block_size−1)` per request; same token bucket.
- Precompile variants: **−44–52%** (buckets that cannot give every request one page per chunk become unreachable and are skipped).

## Costs

- Up to `2P·(block_size−1)` padding rows per request (attended, since they sit inside `cu_q_lens` spans).
- The runner's headroom bucket grows from `+2P` to `+2P·block_size` slack per request, so an occasional batch lands one bucket higher.

## Testing

- Kernel CP suite (multi-request tests parameterized over the map path, including mid-page fetch starts), interface suite (single-request cases now drive the same map path), and the host-only layout-helper tests in `tests/runner/test_pcp_utils.py` — all green on TPU v7x.
- The page-contiguity invariant the kernel fetch relies on is pinned against a brute-force per-token expansion of the zigzag layout.


bhuvanpkaruturi:
Superseded by #3558, which is opened from a branch in this repository so it can be stacked on #3557 and #3425 and show only its own diff.

PR #3562, opened 2026-09-11
Quantize bf16 linear layers to fp8 at load time

lxhfirenking:
# Description

Qwen3.8-2.4T-A95B-FP8 quantizes only its 512 routed experts and lists
978 modules in `modules_to_not_convert`, so every attention and gated-delta-net
projection arrives bf16 and costs 4.27 GiB/chip.

Add `BF16_LINEAR_REQUANTIZE_PATTERNS`, which selects those layers by name and
quantizes their weights to fp8 on the way to the device, giving back 2.13
GiB/chip. We are doing this for NPI (Sunfish), which has way fewer bf16 flops than fp8 flops.

Selected layers get `VllmQuantizedBf16LinearMethod`, which inherits weight
loading and sharding from the unquantized path unchanged and only swaps in the
fp8 apply. Scales are per output channel by default, which stays valid under
both column-parallel and row-parallel sharding;
`BF16_LINEAR_REQUANTIZE_BLOCK_SIZE` switches to 1-D blocks along the input
axis. Both default off, so nothing changes unless the patterns are set.

The names follow the `<SCOPE>_REQUANTIZE_<KNOB>` convention the rest of
`envs.py` uses. The scope is `BF16_LINEAR` rather than plain `LINEAR` because
the bare `REQUANTIZE_WEIGHT_DTYPE` / `REQUANTIZE_BLOCK_SIZE` are already the
linear knobs for checkpoints that arrived quantized, and a `LINEAR_` pair
beside them would differ only by a prefix that does not say which case it
means.

Along the way, `process_weights_after_loading` is split into
`_load_linear_weights` / `_build_linear_weights` / `_store_linear_weights` on
`VllmUnquantizedLinearMethod`. This is a purely mechanical refactor to avoid code duplication for `VllmQuantizedBf16LinearMethod`

# Tests

`tests/layers/vllm/test_unquantized.py` covers pattern matching (exact, `re:`,
fused-shard expansion, the all-or-nothing guard), numerics against the bf16
matmul for merged-column-parallel and row-parallel layers across 1 and 8
devices with and without a block size, the block-size validation errors, and
that unselected layers still take the unquantized path and store no scale.

```
pytest tests/layers/vllm/test_unquantized.py tests/layers/vllm/test_fp8.py
```
382 passed, 100 skipped on v7x-8.

Also Tested E2E with Qwen3.8-2.4T-A95B-FP8 on v7x-32 with
```
'BF16_LINEAR_REQUANTIZE_PATTERNS=re:.*self_attn\..*,re:.*linear_attn.(in_proj_qkv|in_proj_z|in_proj_b|in_proj_a|out_proj)$$' \
'BF16_LINEAR_REQUANTIZE_BLOCK_SIZE=512' \
```
Verified quality with GPQA.

# Checklist

### DIFF
--- a/tpu_inference/layers/vllm/quantization/unquantized.py
+++ b/tpu_inference/layers/vllm/quantization/unquantized.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
 from typing import Any, Callable, Mapping, Optional, Sequence
 
 import jax
@@ -34,6 +33,8 @@ from vllm.model_executor.layers.quantization import \
     register_quantization_config
 from vllm.model_executor.layers.quantization.base_config import (
     QuantizationConfig, QuantizeMethodBase)
+from vllm.model_executor.layers.quantization.utils.config_utils import \
+    is_equal_or_regex_match
 from vllm.model_executor.layers.vocab_parallel_embedding import (
     ParallelLMHead, UnquantizedEmbeddingMethod, VocabParallelEmbedding)
 
@@ -209,18 +210,14 @@ def _load_weight_for_layer(
 def _prefix_matches(prefix: str, patterns: Sequence[str]) -> bool:
     """Match a layer prefix against one of the QUANTIZE_BF16_LINEAR_PATTERNS.
 
-    A bare pattern matches the whole prefix or a dotted suffix of it, so
-    `o_proj` selects every layer named that. A `re:`-prefixed pattern is a
-    regex that has to match the prefix in full, following the convention
-    compressed-tensors uses for its own ignore lists.
+    Patterns follow the same convention as the `ignore` /
+    `modules_to_not_convert` lists these names are usually copied from: a bare
+    pattern is the whole layer name, and a `re:`-prefixed one is a regex
+    anchored at the start. To select every layer with a given suffix, spell the
+    regex out -- `re:.*\\.o_proj`.
     """
-    for pattern in patterns:
-        if pattern.startswith("re:"):
-            if re.fullmatch(pattern[len("re:"):], prefix):
-                return True
-        elif prefix == pattern or prefix.endswith(f".{pattern}"):
-            return True
-    return False
+    return any(
+        is_equal_or_regex_match(prefix, pattern) for pattern in patterns)
 
 
 def should_quantize_bf16_linear(
@@ -358,6 +355,19 @@ class VllmUnquantizedLinearMethod(vllm_linear.UnquantizedLinearMethod,
         if not _tensor_is_in_cpu(layer.weight):
             # Already processed and sharded.
             return
+        weight, bias = self._load_linear_weights(layer)
+        self._store_linear_weights(
+            layer, self._build_linear_weights(layer, weight, bias))
+
+    def _load_linear_weights(
+            self,
+            layer: torch.nn.Module) -> tuple[jax.Array, jax.Array | None]:
+        """Move the layer's weight and bias onto the mesh, freeing the host copy.
+
+        How a weight is loaded does not depend on what is done to it afterwards,
+        so subclasses inherit this unchanged and override
+        `_build_linear_weights` instead.
+        """
         # Under Pathways, shard weights directly onto the TPU mesh to avoid
         # placing a full unsharded copy on a single device (OOM).
         loading_sharding = NamedSharding(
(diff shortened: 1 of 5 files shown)


kyuyeunk (commented):
is this a good idea? typically, we don't want to mess with model's bitwidth that is explicitly defined in the config.

lxhfirenking:
> is this a good idea? typically, we don't want to mess with model's bitwidth that is explicitly defined in the config.

Requantization on the fly is an existing pattern in tpu-inference. We already have MOE_REQUANTIZE_WEIGHT_DTYPE, so i dont think this is a brand new pattern. At the end of the day this is enabled by flags, so we can assume whoever wants this on knows what they are doing.

kyuyeunk:
yes but it always keep the bitwidth the same. common pattern is to change block size but not bitwidth.

kyuyeunk:
also, if we are going to introduce it, let's keep it consistent and have it `LINEAR_REQUANTIZE_WEIGHT_DTYPE`

lxhfirenking:
> yes but it always keep the bitwidth the same. common pattern is to change block size but not bitwidth.

I find configs like these in benchmark runs https://github.com/vllm-project/tpu-inference/blob/main/.buildkite/benchmark/cases/daily/deepseek.json#L31  the original checkpoint has moe weights in fp8, and they are requantized into fp4 for the benchmark

kyuyeunk:
hmmm that doesn't sound right. i'm not sure if that is used in production anywhere? but anyway, as long as it is 1. gated 2. follows prior design convention. it should be fine.

lxhfirenking:
> also, if we are going to introduce it, let's keep it consistent and have it `LINEAR_REQUANTIZE_WEIGHT_DTYPE`

Updated to `BF16_LINEAR_REQUANTIZE_*`. 

We have existing `REQUANTIZE_WEIGHT_DTYPE`/`REQUANTIZE_BLOCK_SIZE` flags. These are really misnomers because they are exclusively used for linear layers that are already fp8 in the checkpoint, so I used a more specific name prefix for our new flags to avoid naming/semantic collision with those.

kyuyeunk:
can you go more into detail on `REQUANTIZE_WEIGHT_DTYPE`? because we can just have it as a no-op for pre-existing fp8 linear checkpoint and only trigger it for here right?

lxhfirenking:
> can you go more into detail on `REQUANTIZE_WEIGHT_DTYPE`? because we can just have it as a no-op for pre-existing fp8 linear checkpoint and only trigger it for here right?

The entry point is [here](https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/common/quantization/configs.py#L55)

Then it got used by both regular fp8 linear path and compressed_tensor fp8 linear path https://github.com/search?q=repo%3Avllm-project%2Ftpu-inference+requant_weight_dtype&type=code 

The whole thing runs in parallel with the code path that handles unquantized weights from checkpoint.

In general, I don't think it is a good idea to completely change the semantic meaning of an existing set of flagslike this. We could unknowingly break existing use cases. If we dont want these REQUANTIZE_* flags any more we should just deprecate them instead.

kyuyeunk:
don't get it. it seems more like the existing naming of `REQUANTIZE_WEIGHT_DTYPE` is wrong?

kyuyeunk:
there is no messaging from the env name that it's only for compressed tensor so users would be confused why that variable would not work for other checkpoint types.

lxhfirenking:
> there is no messaging from the env name that it's only for compressed tensor so users would be confused why that variable would not work for other checkpoint types.

Yes I agree that the naming of `REQUANTIZE_WEIGHT_DTYPE` is just bad and non specific. It does not include either the layer or the checkpoint type is targeting. The only way user would know what it does is reading through code. This makes it extra difficult to know what existing users of these flags expect them to behave like. 

That is yet more reason to use the more specific `BF16_LINEAR_REQUANTIZE_ *` names instead of trying to reuse the old confusing names.

Let's try to clean up the `REQUANTIZE_WEIGHT_DTYPE` in a separate PR, through renaming and better documentation.

jrplatin (approved):

tpu_inference/layers/vllm/quantization/unquantized.py
```diff
@@ -392,6 +483,153 @@ def apply(self,
         return out
 
 
+class VllmQuantizedBf16LinearMethod(common_fp8.Fp8LinearMethod,
```
kyuyeunk, line 501: why create an entirly new class for this? if i understand correctly, shouldn't the core logic be exactly the same?

kyuyeunk (approved):
the general direction of this pr looks very wrong. isn't it better to override it at custom op level where quantization method gets initialization?: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/custom_ops/linear.py

you can see some existing example here: https://github.com/vllm-project/tpu-inference/blob/main/tpu_inference/layers/vllm/custom_ops/embedding.py

kyuyeunk (changes requested):
revert

kyuyeunk (commented):
suggest referring to online fp8 where it converts an checkpoint into fp8 during model loading time: https://github.com/vllm-project/vllm/blob/0384e72693985d0eddae70bd330b9cb283ff4313/vllm/model_executor/layers/quantization/online/fp8.py

kyuyeunk (approved):

tpu_inference/layers/vllm/quantization/unquantized.py
lxhfirenking: synced offline over design rationale. closing conversation

PR #3563, opened 2026-09-11
[CI] Run the P/D disaggregation benchmark on the Kueue fleet

theminghuang:
Adds a non-blocking kube lane for the 1P1D disaggregation benchmark, beside the nightly that runs the same thing on `tpu-v7x-cluster-c`.

**What**
- `.buildkite/kubernetes/manifests/v7x/kueue_disagg.yaml` — the three Deployments as one JobSet: prefill and decode on 2x2x1 tpu7x, proxy and client in a CPU pod. No Services, no per-run Hyperdisk claims, no privileged init container.
- `.buildkite/kubernetes/benchmark.Dockerfile` — upstream `vllm/vllm-tpu:nightly` republished to the project's Artifact Registry with `buildkite-agent` added. The fleet accepts a workload image only from that registry, and the benchmark pod has no agent to upload with otherwise.
- `.buildkite/pipeline_disagg_kube.yml` — a pipeline of its own, manually triggered.

**Why not change the existing pipeline**: both lanes measure the same benchmark on the same eight-chip reservation, so they must not run at once. Separate pipelines is what lets this one be triggered by hand while the nightly keeps the schedule and keeps owning the Spanner rows. Results here are Buildkite artifacts only.

`soft_fail: true` while the lane is being brought up, so a fleet problem does not read as a regression here.

Successful run: https://buildkite.com/vllm/tpu-inference-pd-disagg-kube/builds/7/list


.buildkite/kubernetes/toy_proxy_server.py
theminghuang, line 1: This is copied from https://github.com/vllm-project/tpu-inference/blob/main/examples/disagg/toy_proxy_server.py since we should take this opportunity to consolidate stuff related to buildkite into our own folders

QiliangCui (approved):

PR #3565, opened 2026-09-11
[Prefix Caching] Fix out-of-bounds GDN recurrence kernel crash in align mode

bhuvanpkaruturi:
## Summary

Fixes a TensorCore halt (`E0200: RuntimeUnexpectedCoreHalt`, HLO `fused_conv1d_gdn_batched`) in the GDN recurrence kernel when serving hybrid models such as `Qwen/Qwen3.5-397B-A17B` with prefix caching in align mode on a multi-host slice. The halt appeared once multi-turn conversations passed roughly 20k context tokens.

## Root cause

In align mode the mamba state pool is sized by the model runner during HBM profiling and sharded per attention-DP rank. With the Ray or multiproc executors the runner lives in a worker process, so the value it wrote to `cache_config.mamba_num_blocks` never reached the engine-core process. The scheduler sized its per-rank pools from a different number and handed out slot ids past the rows the workers had actually allocated. The Pallas kernel runs with bounds checks disabled, so an out-of-range slot id became an illegal DMA and halted the core.

## Changes

- **Publish the allocated pool size to the engine core.** `TPUWorker.get_mamba_num_blocks` reports the rows each worker allocated. `MambaPoolSyncExecutorMixin`, mixed into the multiproc and both Ray executors, fetches it over `collective_rpc` right after `initialize_from_config`, before the scheduler is built, and writes it to `cache_config`, `kv_cache_config` and the coordinator registry. Workers that disagree, or a hybrid model with no reported size, raise instead of guessing.
- **Install the hybrid coordinator hooks from the executors**, which run in the engine-core process; the platform config hook did not.
- **Share the sizing helpers** (`mamba_blocks_per_request`, `mamba_pool_size`) between the runner's `KVCacheManager` and the coordinator so both sides use one formula. `DPScheduler` partitions the propagated value per rank.
- **Clamp slot ids in `gdn_attention_op.py`** to the rank-local row count before they reach the kernel, as a last line of defense against a future sizing mismatch. Nothing under `tpu_inference/kernels/` changes.

## Verification

Cloud TPU v7x-16 (2 hosts, 16 devices), `Qwen/Qwen3.5-397B-A17B`, TP=16 with attention DP=8 and expert parallel, `--enable-prefix-caching`, `custom_mamba_cache_multiplier=16`:

- Runner allocated 1752 mamba rows; the engine core received 1752 from both workers; all 8 rank schedulers sized their pools at 219 rows, matching the arrays.
- Multi-turn agentic benchmark completed 240/240 turns with zero `E0200` halts. Earlier revisions of this branch also ran the full agentic trace (>50M tokens, >20k-token contexts) with zero halts and no numerical issues.
- Unit tests cover the propagation path, executor ordering, the DP partitioning and the coordinator's fail-fast behaviour.

### DIFF
--- a/tpu_inference/core/hybrid_coordinator.py
+++ b/tpu_inference/core/hybrid_coordinator.py
@@ -59,6 +59,37 @@ def get_mamba_num_blocks() -> int | None:
     """Get the registered Mamba block pool capacity."""
     return _GLOBAL_MAMBA_NUM_BLOCKS
 
+def derive_mamba_num_blocks(vllm_config: Any, divisor: int = 8) -> int:
+    """Deterministically derive the compact Mamba block pool capacity from VllmConfig.
+
+    This matches the sizing formula in KVCacheManager without requiring
+    worker-side runtime registration to have completed first.
+    """
+    scheduler_config = getattr(vllm_config, "scheduler_config", None)
+    additional_config = getattr(vllm_config, "additional_config", None) or {}
+
+    # Check explicit user override
+    if "mamba_num_blocks" in additional_config:
+        val = int(additional_config["mamba_num_blocks"])
+        return max(((val + divisor - 1) // divisor) * divisor, divisor)
+
+    max_num_reqs = getattr(scheduler_config, "max_num_seqs", 128)
+    num_spec = 0
+    if getattr(vllm_config, "speculative_config", None) is not None:
+        num_spec = getattr(vllm_config.speculative_config, "num_speculative_tokens", 0)
+
+    # active slots: 1 per concurrent request + 1 per speculative token + null block
+    active_slots = max_num_reqs * (num_spec + 1) + 1
+
+    # checkpoint budget for prefix caching in align mode
+    checkpoint_budget = additional_config.get("mamba_cache_checkpoint_budget")
+    if checkpoint_budget is None:
+        checkpoint_budget = max_num_reqs * 8
+    checkpoint_budget = int(checkpoint_budget)
+
+    total_slots = active_slots + checkpoint_budget
+    return max(((total_slots + divisor - 1) // divisor) * divisor, divisor)
+
 
 def is_mamba_spec(spec: Any) -> bool:
     """Check if a KV cache spec represents a Mamba layer."""
@@ -286,9 +317,10 @@ class TPUHybridKVCacheCoordinator(HybridKVCacheCoordinator):
         if mamba_num_blocks is None:
             mamba_num_blocks = get_mamba_num_blocks()
         if mamba_num_blocks is None:
-            raise ValueError(
-                "[TPUHybridKVCacheCoordinator] mamba_num_blocks must be registered "
-                "via set_mamba_num_blocks().")
+            mamba_num_blocks = derive_mamba_num_blocks(self.vllm_config)
+            logger.info(
+                "[TPUHybridKVCacheCoordinator] mamba_num_blocks was not registered via "
+                "set_mamba_num_blocks; derived %d from VllmConfig.", mamba_num_blocks)
         self.mamba_num_blocks = int(mamba_num_blocks)
 
         logger.info(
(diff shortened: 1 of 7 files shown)


tpu_inference/kernels/gdn/v3/wrapper.py
kyuyeunk, line 1: Don't introduce framework level external dependencies to the kernel

kyuyeunk (commented):
Don't get why all these changes are needed? Just clipping at the metadata generation level (not even inside the kernel) should suffice to fix hbm out of bounds error?

bhuvanpkaruturi:
> Don't get why all these changes are needed? Just clipping at the metadata generation level (not even inside the kernel) should suffice to fix hbm out of bounds error?

@kyuyeunk Thanks for the review. I have reverted or moved all the changes touching files in kernels/. 

Regarding the other changes: clipping only prevents the crash. The root cause is a mismatch in the mamba block count between the scheduler and the runner. Without these changes the scheduler sizes its mamba pool from the config (257 blocks per DP rank) while the runner allocates fewer after the HBM re-split (219 per rank). A request "A" that the scheduler assigns a slot id of 219 or above gets clipped to 218, so it reads its initial state from slot 218 and writes its updated state back there, overwriting the state of whatever request "B" the scheduler legitimately gave slot 218. That produces wrong outputs with no error. The fix is to make the scheduler use the runner's allocated block count,

tpu_inference/core/hybrid_coordinator.py
```diff
@@ -60,6 +78,51 @@ def get_mamba_num_blocks() -> int | None:
...
+    `dp_size * max_num_seqs` and the arrays are sharded over the DP axis. Use
+    only when the runner's registered value has not reached this process.
+    """
+    dp_size = vllm_config.sharding_config.total_dp_size
+    max_num_reqs = dp_size * vllm_config.scheduler_config.max_num_seqs
+    _, blocks_per_req = mamba_blocks_per_request(vllm_config,
+                                                 is_align_mode=True)
+    return mamba_pool_size(max_num_reqs, blocks_per_req, divisor=dp_size)
+
+
```
wenxindongwork: do we really need this fallback logic? would this ever be reached?

tpu_inference/core/hybrid_coordinator.py
```diff
@@ -283,12 +346,24 @@ def __init__(
...
+                    "[TPUHybridKVCacheCoordinator] mamba_num_blocks must be "
+                    "registered via set_mamba_num_blocks() or derivable from "
+                    "the VllmConfig passed to install_hybrid_coordinator_hooks()."
+                )
+            mamba_num_blocks = derive_mamba_num_blocks(_VLLM_CONFIG)
+            logger.warning(
+                "[TPUHybridKVCacheCoordinator] mamba_num_blocks did not "
+                "reach this process from the model runner; derived %d from "
+                "VllmConfig, which may not match the allocated arrays.",
+                mamba_num_blocks)
```
wenxindongwork: can we safely assume that we will never reach this path? or that if this is hit, something unexpected has happened?

wenxindongwork (commented):
thank you for the PR! could we make sure that this PR doesn't break the single-host (non Ray) path?

tpu_inference/core/hybrid_coordinator.py
bhuvanpkaruturi: Yes. We should not reach this block and the only scenario where this will execute is when the runner failed to broadcast the mamba block count it has computed.

wenxindongwork: thank you for confirming! i think then we can safely delete this block, and also `_VLLM_CONFIG`.

bhuvanpkaruturi: Done. removed the fallback and the _VLLM_CONFIG global.

wenxindongwork (approved):

Lumosis (approved):
Approved for ray executor

jrplatin (approved):

PR #3579, opened 2026-09-15
[Bugfix] Replicate logprobs tensors so multihost device_get() succeeds

bhuvanpkaruturi:
# Description

Requesting logprobs on a 2-host `tpu7x-16` slice with `enable_dp_attention` (`attn_dp_size=8`) crashes on the first decode step:

```
RuntimeError: Fetching value for `jax.Array` that spans non-addressable (non process local) devices is not possible.
```

`gather_logprobs()` returns its outputs with whatever sharding the input logits had, which under DP attention is the `attn_data` axis. On a multi-host mesh that axis spans the other host's devices, so the `jax.device_get()` in `_jax_logprobs_materialize()` can't read them. Runs without logprobs never hit this path.

The fix constrains the three outputs to `NamedSharding(mesh, P())` inside the jitted function, the same thing `sample()` already does for `next_tokens`. `mesh` is threaded through as a static jit arg and defaults to `None`, in which case nothing changes. The routed-expert indices in `vllm_model_wrapper.py` are fetched over the same path and fail the same way, so they get the same constraint.

Don't try to work around this on the host side by concatenating `addressable_shards`: it silently returns duplicated rows, and a different set on each host.

Stacked on #3565 since that's the tree the bug was reproduced on. Retarget to `main` after it merges.

# Tests

- `tests/layers/jax/sample/test_sampling.py` passes on CPU with 1 and 8 devices. New `test_logprobs_replicated_with_mesh` checks that `gather_logprobs`, `compute_and_gather_logprobs` and `compute_and_gather_prompt_logprobs` return fully replicated outputs when given a mesh, with values identical to the unconstrained call. `test_compute_prompt_logprobs_success` now passes a mesh too.
- A/B on `tpu7x-16`: unpatched reproduces the error; with this patch both hosts return the same top-k ids, logprobs and ranks.
- 480-turn agentic run with logprobs and routed-expert collection on, no errors.

# Checklist

- [x] I have made or will make corresponding changes to any relevant documentation (none needed).


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -354,24 +354,50 @@ def compute_logprobs(logits: jax.Array) -> jax.Array:
...
+# Stating it on the jit rather than asking each caller to opt in is the same
+# guarantee `sample()` gives for `next_tokens`, and it means precompilation
+# necessarily warms up the program the runner actually calls. It costs nothing
+# when it is not needed: for the raw logprobs modes the logits are sharded over
+# the vocab axis rather than over requests, so the outputs are already
+# replicated and the constraint folds away to zero extra collectives.
+#
+# Callers must run under `jax.set_mesh()` so the bare PartitionSpec resolves.
+
+
```
wenxindongwork: nit: could we reduce the amount of comments?

tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -389,6 +415,10 @@ def compute_prompt_logprobs(
     """Dispatches prompt logprob computation on TPU and snapshots per-request state.
     Returns PromptLogprobsAsyncData containing the async-copied tensors and
     the snapshotted state needed to safely slice them in get_output().
+
+    Must be called under `jax.set_mesh()`: the returned tensors are replicated
+    for the host fetch, and the bare PartitionSpec that does it resolves
+    against the mesh in context.
```
wenxindongwork: nit: i think it okay to remove this comment

tpu_inference/runner/tpu_runner.py
```diff
@@ -2100,22 +2100,21 @@ def _sample_from_logits(
         logits = logits.astype(jnp.float32)
         if full_logits is not None:
             full_logits = full_logits.astype(jnp.float32)
-        with self.maybe_forbid_compile:
+        # The logprobs jits below are declared with out_shardings=P(), whose
+        # bare PartitionSpec resolves against the mesh in context.
```
wenxindongwork: nit: same here, i think we can drop this comment

wenxindongwork (approved):

tpu_inference/layers/jax/sample/sampling.py
bhuvanpkaruturi: Done.

tpu_inference/runner/tpu_runner.py
bhuvanpkaruturi: Done. Shortened the comment to one line.

jrplatin (approved):

PR #3580, opened 2026-09-15
[CI] [sample] Make top_k=1 deterministic by routing it to argmax

JiriesKaileh:
# Description

Both top-k implementations (the distributed candidate sampler and the legacy 'topk_mask' binary search) are threshold based, so they retain every value tied with the k-th ranked value. With 'top_k=1' an exact tie at the maximum therefore leaves more than one candidate and 'jax.random.categorical' picks between them at random.

bf16 logits make such ties common. For the prompt used by 'tests/e2e/test_sampling_params.py::TestTopK::test_top_k_restricts_sampling', Llama-3.2-1B produces logits of exactly 21.375 for both " 5" and " 7" at the first decode step, so two 'generate' calls with 'top_k=1' disagreed roughly half the time. That is what has been flaking the 'sampling_params' correctness step.

'top_k=1' is exactly argmax, so take the argmax directly. Only token selection changes; 'is_greedy' still drives which logits are returned, so processed logprobs are unaffected.

FIXES: [b/561500965](https://b.corp.google.com/issues/561500965) , [b/561500849](https://b.corp.google.com/issues/561500849)

# Tests

Included `test_top_k_one_is_argmax_with_tied_logits` in `test_sampling.py`


tpu_inference/layers/jax/sample/sampling.py
```diff
@@ -288,6 +288,14 @@ def sample(
...
     else:
         is_greedy = tpu_sampling_metadata.temperature < _SAMPLING_EPS
+        # top_k == 1 leaves a single candidate, so it is exactly argmax. Both
+        # top-k implementations are threshold based and retain every value
+        # tied with the k-th rank, which would otherwise make these rows
+        # sample uniformly among tied maxima. Take the argmax directly so the
+        # result is deterministic and matches vLLM's exact-k top-k. Note this
+        # only selects the token; `is_greedy` still drives which logits are
+        # returned, so processed logprobs are unaffected.
+        takes_argmax = is_greedy | (tpu_sampling_metadata.top_k == 1)
```
lk-chen, line 298: @prishajain1 @gutianyu-google do we want to preserve tied logits still?

related discussion https://github.com/vllm-project/tpu-inference/pull/3490#discussion_r3981976919

PR #3587, opened 2026-09-16
[DP scheduler] Add a round_robin request routing policy

alisachen-google:
# Description

Adds a `round_robin` DP request-routing policy configurable via the `DP_SCHED_ROUTING` environment variable. When enabled, incoming requests are assigned to DP ranks cyclically, skipping the per-request all-rank state queries performed by the default `least_loaded` policy. `least_loaded` remains the default.

### Motivation & Problem
`DPScheduler` currently routes each new request by querying every DP rank for its pending-prefill tokens, in-flight request count, and remaining output tokens (`_find_best_rank_for_request`).

For workloads with large batches of similarly shaped requests (e.g., RL rollouts with thousands of prompts), this per-request IPC fan-out dominates host CPU time before the first decode step. In high-data-parallelism setups with homogeneous requests, cyclic assignment achieves balanced distribution with zero inter-rank query overhead.

### Key Changes
- **Environment Variable (`tpu_inference/envs.py`)**:
  - Added `DP_SCHED_ROUTING` (default: `"least_loaded"`, choices: `["least_loaded", "round_robin"]`), validated case-insensitively via `env_with_choices`.
- **Scheduler Implementation (`tpu_inference/core/sched/dp_scheduler.py`)**:
  - Initialized `self._routing_policy` and `self._round_robin_next_rank` during scheduler initialization.
  - Added `_pick_rank_for_request(request)`: returns `(rank + 1) % dp_size` without querying rank state when `round_robin` is active; otherwise delegates to `_find_best_rank_for_request`.
  - Updated `_route_and_forward_request` to use `_pick_rank_for_request`.

### Benchmark Results
- **Qwen3-0.6B** (`tpu7x`, DP=32 TP=4, 2048 prompts):
  - Rollout time per step: **64.1 s $\to$ 59.1 s** (~8% speedup).
- **Gemma4-E2B GRPO rollout** (`tpu7x`, DP=128 TP=1h):
  - Rollout time per step: **~130.5 s $\to$ ~58.4 s** (~55% speedup / 2.2x faster).
  - Demonstrates that rank query fan-out is the dominant host-side bottleneck at high DP scales ($DP \ge 128$).

### Trade-offs
- `round_robin` assumes relatively uniform request lengths and does not inspect dynamic rank utilization. It is intended for batch/rollout workloads rather than heterogeneous online serving.

---

# Tests

- **`tests/test_envs.py`**:
  - `test_dp_sched_routing`: Verifies default (`"least_loaded"`), `"round_robin"`, case-insensitivity (`"Round_Robin"`), and invalid policy values raising `ValueError`.
- **`tests/core/test_dp_scheduler.py`**:
  - `test_round_robin_routing_skips_rank_queries`: Verifies cyclic rank assignment `[0, 1, 0, 1, 0]` across requests and ensures `_find_best_rank_for_request` is never called.

**Run tests:**
```bash
pytest tests/test_envs.py -k test_dp_sched_routing
pytest tests/core/test_dp_scheduler.py -k test_round_robin_routing_skips_rank_queries

### CODE REVIEW

**Review by sixiang-google — APPROVED** [review]

File: tpu_inference/envs.py [hunk 1]
```diff
@@ -71,6 +71,7 @@
     NUM_PRECOMPILE_WORKERS: int = 1
     DP_SCHED_BATCH_PREFILL: bool = False
     DP_SCHED_BATCH_PREFILL_FLUSH_TIMEOUT_MS: int = 10000
+    DP_SCHED_ROUTING: str = "least_loaded"
```
wenxindongwork, line 77: do you mind also creating a Enum class  inside DPScheduler for the routing strategy so that it is clear to the users what the options are?

wenxindongwork (approved):

tpu_inference/envs.py
alisachen-google: Done. Thanks Wenxin!

QiliangCui (approved):

PR #3590, opened 2026-09-17
[CI] [sampling] Make top_k tests tie-aware

JiriesKaileh:
# Description

top_k=1 is not greedy: topk_mask keeps every token tied with the k-th largest logit, so a tied top logit leaves several candidates and sampling picks among them. Llama-3.2-1B ties " 5" and " 7" on "Pick a number between 1 and 10:", which flaked the nightly.

FIXES: [b/561500849](https://b.corp.google.com/issues/561500849), [561500965](https://b.corp.google.com/issues/561500965)

# Tests

I tested with the test itself.


tests/e2e/test_sampling_params.py
```diff
@@ -150,34 +178,58 @@ def test_top_p_one_skips_masking(self, llm: LLM):
...
+            completion = outputs[0].outputs[0]
+            for sampled_logprob, best_logprob in iter_step_logprobs(
+                    completion):
+                assert sampled_logprob == pytest.approx(
+                    best_logprob, abs=TIE_TOLERANCE), (
+                        "top_k=1 sampled a token that is not tied for the "
+                        f"highest probability: {sampled_logprob} vs "
+                        f"{best_logprob}")
+
+    def test_top_k_restricts_sampling(self, llm: LLM):
```
lk-chen, line 210: for original test that checks "top_k=1 restricts sampling", I think you can simply change the prompt to something more deterministic e.g. "Capital of France is: "

tests/e2e/test_sampling_params.py
```diff
@@ -150,34 +178,58 @@ def test_top_p_one_skips_masking(self, llm: LLM):
...
+        """
+        prompt = "Pick a number between 1 and 10:"
         sampling_params_k1 = SamplingParams(temperature=1.0,
                                             top_k=1,
-                                            max_tokens=5)
+                                            max_tokens=5,
+                                            logprobs=5)
 
-        # top_k=-1 considers all tokens
+        for _ in range(10):
```
lk-chen: SamplingParams has `n` attribute to determine how many responses to generate 

https://docs.vllm.ai/en/latest/api/vllm/#vllm.SamplingParams.n

JiriesKaileh: Added

tests/e2e/test_sampling_params.py
JiriesKaileh: Changed

PR #3592, opened 2026-09-17
Fix multi-host JAX crash: avoid device_get on non-addressable arrays (#3570)

### LINKED ISSUE #3570: Multi-host serving crashes at sampling: jax.device_get on cross-host (non-addressable) array in host_extract_sampled_tokens
## Summary

When serving on a **multi-host** TPU (a tensor-parallel / mesh group that spans more than one host), the worker crashes on the **first request**, during token sampling. `host_extract_sampled_tokens` calls `jax.device_get(...)` on a `jax.Array` that is sharded across hosts, which raises:

```
RuntimeError: Fetching value for `jax.Array` that spans non-addressable (non process local) devices is not possible.
```

This is **not model-specific** — the sampling path runs for every model, and the sampled-tokens array is sharded by the multi-host mesh, not by the model architecture. Single-host serving (all devices process-local/addressable) is unaffected.

## Environment

- vLLM: `0.28.1rc1.dev500+g51da0ca66` (TPU / JAX backend, `tpu-inference`)
- Model: `Qwen/Qwen3-32B-FP8` (behavior is model-independent; see Summary)
- TPU: multi-host slice — **2 hosts**, `tensor_parallel_size=8` (mesh spans both hosts), multi-host workers launched via Ray (`RayWorkerProc` on a remote host)
- Observed on the prefill worker in a disaggregated (P/D) setup; the failing code path is the generic per-step sampling path and is not specific to disaggregation.

## What happens

Startup succeeds fully — weight load, XLA compile/warmup, and KV-cache allocation all complete and the server reports ready. On the **first request**, a worker process raises and the `EngineCore` is force-killed (the HTTP frontend then returns 500 / connection errors).

## Traceback

```
WorkerProc hit an exception.
Traceback (most recent call last):
  File "/workspace/vllm/vllm/v1/executor/multiproc_executor.py", line 1047, in _execute_worker_rpc
    output = func(*args, **kwargs)
  File "/workspace/tpu_inference/tpu_inference/worker/tpu_worker.py", line 616, in sample_tokens
    return self.model_runner.sample_tokens(grammar_output)
  File "/workspace/tpu_inference/tpu_inference/runner/tpu_runner.py", line 1441, in sample_tokens
    return self._sample_from_logits(
  File "/workspace/tpu_inference/tpu_inference/runner/tpu_runner.py", line 2268, in _sample_from_logits
    valid_sampled_token_ids = runner_utils.host_extract_sampled_tokens(
  File "/workspace/tpu_inference/tpu_inference/runner/utils.py", line 950, in host_extract_sampled_tokens
    next_tokens = np.asarray(jax.device_get(next_tokens))
  File ".../jax/_src/api.py", line 2501, in device_get
    return tree_map(_device_get, x)
  File ".../jax/_src/api.py", line 2462, in _device_get
    return toarray()
  File ".../jax/_src/array.py", line 414, in __array__
    return np.asarray(self._value, dtype=dtype, **kwds)
  File ".../jax/_src/array.py", line 653, in _value
    raise RuntimeError(
RuntimeError: Fetching value for `jax.Array` that spans non-addressable (non process local) devices is not possible. You can use `jax.experimental.multihost_utils.process_allgather` to print the global array or use `.addressable_shards` method of jax.Array to inspect the addressable (process local) shards.
```

## Root cause

`host_extract_sampled_tokens` (`tpu_inference/runner/utils.py`) does:

```python
next_tokens = sampled_output
if spec_decode_metadata is None:
    next_tokens = np.asarray(jax.device_get(next_tokens))   # <-- fails on multi-host
    ...
```

`next_tokens` (the sampled token ids returned by `sample(...)`) is a **global** `jax.Array` sharded across the multi-host mesh. In a multi-process (multi-host) job, each process can only access its local (addressable) shards, so calling `jax.device_get` on the global array raises. There is no cross-host gather before the host copy.

## Reproduction

1. Serve any model on a multi-host TPU slice with tensor parallelism spanning **> 1 host** (e.g., 2 hosts, `tp=8`) on the JAX backend.
2. Send a single completion request.
3. The worker crashes with the traceback above and the engine dies.

joeywan-google:
## Summary

Fixes #3570.

On a **multi-host** mesh (e.g. 2 hosts × 4 chips, `tensor-parallel-size: 8`), the engine starts up, loads weights, compiles, and allocates KV cache — then **crashes on the first request** during host token extraction:

```
File "tpu_inference/runner/utils.py", host_extract_sampled_tokens
  next_tokens = np.asarray(jax.device_get(next_tokens))
RuntimeError: Fetching value for `jax.Array` that spans non-addressable
(non process local) devices is not possible.
```

## Root cause

Sampling/decoding produce global `jax.Array`s whose shards span more than one host. A plain `jax.device_get()` tries to fetch **non-addressable** (remote) shards, which JAX forbids in a multi-process job. Single-host runs never hit this; the first multi-host sampling step does.

## Fix

Add a `safe_device_get()` helper (`tpu_inference/utils.py`) that copies a `jax.Array` to host numpy correctly under multi-host SPMD:

- **Fully replicated** array → read the process-local shard directly (cheap, no collective). This is the common sampling case (outputs are replicated across tensor-parallel ranks).
- **Sharded** array (e.g. across the data-parallel axis) → reconstruct the full value with `multihost_utils.process_allgather`. Taking a single shard would silently return **partial** data, so this path is required for correctness.
- **Single-host / non-jax** input → fall back to `jax.device_get()` (behavior unchanged).

Apply it at the token-extraction sites that receive these global arrays:
- `runner/utils.py` → `host_extract_sampled_tokens`
- `runner/tpu_runner.py` → `_process_continue_decode_outputs` (sync) and `_modify_prev_results` (async continue-decode)

## Scope / notes

- Pure Python, `tpu_inference`-only; no vLLM or public API changes. Single-host behavior is unchanged.
- Remaining `jax.device_get` calls on feature-gated paths (logprobs, routed-expert indices) are **not** exercised by the default generation path and are left as follow-up.

## Test

`tests/test_utils.py::test_safe_device_get_*`:
- replicated array → reads the local shard, never calls `device_get`/`process_allgather`;
- sharded array → gathers via `process_allgather` (asserts partial shard is **not** used);
- shardless input → falls back to `device_get`;
- single-host `jax.Array` roundtrips correctly.


tpu_inference/utils.py
```diff
@@ -42,6 +43,31 @@
...
+      gather it across processes with ``process_allgather`` to reconstruct the
+      full value -- taking a single shard would silently return partial data.
+    - Otherwise (single-host arrays or plain numpy / non-jax inputs) we fall
+      back to ``jax.device_get()``.
+    """
+    if hasattr(arr, "addressable_shards") and arr.addressable_shards:
+        if getattr(arr, "is_fully_replicated", False):
+            return np.asarray(arr.addressable_shards[0].data)
+        return np.asarray(multihost_utils.process_allgather(arr, tiled=True))
+    return np.asarray(jax.device_get(arr))
```
anthonsu, line 68: We may want `is_fully_addressable` guard. Shall we check if the array is already fully addressable?

tpu_inference/utils.py
```diff
@@ -42,6 +43,31 @@
...
+    - If the array is **fully replicated**, every host holds the complete value
+      on its first process-local shard, so we read that shard directly (cheap,
+      no collective).
+    - If the array is **sharded** (e.g. across the data-parallel axis), we must
+      gather it across processes with ``process_allgather`` to reconstruct the
+      full value -- taking a single shard would silently return partial data.
+    - Otherwise (single-host arrays or plain numpy / non-jax inputs) we fall
+      back to ``jax.device_get()``.
+    """
+    if hasattr(arr, "addressable_shards") and arr.addressable_shards:
```
anthonsu, line 64: It looks like lack of Pytree support. Considering if user passes a tuple or dict, this check would eval to False and drop down to `jax.device_get(arr)` then crashes again.

tpu_inference/utils.py
```diff
@@ -42,6 +43,31 @@
...
+        if getattr(arr, "is_fully_replicated", False):
+            return np.asarray(arr.addressable_shards[0].data)
+        return np.asarray(multihost_utils.process_allgather(arr, tiled=True))
```
anthonsu, line 67: `process_allgather` is a collective SPMD operation, synchronously from my understand. If `safe_device_get` is called conditionally on rank 0 only, the TPU cluster may hang indefinitely in a deadlock.

`process_allgather` can trigger a `jax.jit` compilation which can cause unwanted latency spikes during token generation.

Also `tiled=True` concatenates along axis 0. If a tensor is partitioned across a data-parallel axis that isn't axis 0, or if local shapes differ, this may produce an unexpected shape. 

Regarding tensor replicated across TP ranks, please ensure this collective call is executed symmetrically across all hosts if we encounter a true sharded multi-host tensor here.

tpu_inference/runner/tpu_runner.py
anthonsu, line 250: These still call raw `jax.device_get` on global arrays, may cause the same crashes.

tpu_inference/runner/utils.py
anthonsu, line 958: The `parse_output` does `np.asarray`, which could raise the identical `non-addressable` error.

tpu_inference/utils.py
joeywan-google: Sure. Added is_fully_addressable as the first check

tpu_inference/utils.py
joeywan-google: Done. safe_device_get now maps over leaves with jax.tree_util.tree_map, so tuples/dicts work like jax.device_get. Added tests for both.

tpu_inference/utils.py
joeywan-google: Regarding deadlock, the allgather is now the last branch and only fires for a true multi-host sharded array. The guards (is_fully_addressable/is_fully_replicated) are global array properties, so every host takes the same branch — it's symmetric, and the call sites run in SPMD on all hosts (not rank-0-only).

joeywan-google: For latency, I think the common paths (single-process, replicated) never hit the collective, so the decode hot path avoids it.

tiled=True - the tokens here are 1-D sharded on axis 0 (batch/DP), so axis-0 concat is right. Agreed it's not general-purpose anyway.

tpu_inference/runner/utils.py
joeywan-google: Done

tpu_inference/runner/tpu_runner.py
joeywan-google: Done

Lumosis (approved):

anthonsu (approved):

PR #3593, opened 2026-09-17
Fix KeyError when a request finishes before its routed-experts output resolves

wenxindongwork:
## Summary

With `enable_return_routed_experts=True`, EngineCore dies mid-generation:

```
File "tpu_inference/runner/tpu_runner.py", line 478, in get_output
  routed_experts = _reconstruct_routed_experts(
File "tpu_inference/runner/tpu_runner.py", line 791, in _reconstruct_routed_experts
  req_state = runner.requests[req_id]
KeyError: '1-b742e2fc'
```

`_reconstruct_routed_experts` dereferences **live runner state at output-resolution time**:

```python
req_state = runner.requests[req_id]
```

That is unsafe under the engine's batch queue. `step_with_batch_queue` dispatches the next step before it blocks on the previous step's future:

1. `execute_model(step N)` returns an `AsyncTPUModelRunnerOutput` holding `expert_indices` and `scheduler_output(N)`
2. `execute_model(step N+1)` runs **first** → `_update_states` → `persistent_batch_manager.py`: `finished_req_states[req_id] = self.requests.pop(req_id, None)` evicts finished requests
3. *then* `future.result()` for step N → `get_output()` → `_reconstruct_routed_experts(scheduler_output N)` walks step N's req_ids and indexes `runner.requests` for a request step 2 already popped → `KeyError`, fatal to EngineCore

The trigger is a request **finishing while its output is still in flight**, so it needs at least two concurrent requests with different completion times. Single-request runs never open the window, which is why this survived the existing tests.

Observed on `Qwen/Qwen3.5-35B-A3B`, TP=8, two prompts at `max_tokens=512`, one finishing before the other.

## Fix

Snapshot the attention-group block IDs at **dispatch**, when the request is guaranteed live, and pass that into the reconstruction. This mirrors the `_re_block_ids` snapshot vLLM's scheduler already keeps for the same hazard ("A concurrent `schedule()` may preempt requests and free blocks before `update_from_output` runs; the snapshot survives that").

`_reconstruct_slots_for_request` now takes `block_ids` directly rather than a `CachedRequestState` — it only ever read `req_state.block_ids`, and the narrower signature means a caller cannot reintroduce the lifetime dependency. The snapshot copies the list, so later mutation of the request's block list can't change it either.

Both call sites are covered: the async path reads the snapshot stored on `AsyncTPUModelRunnerOutput`, and the sync path (`_sample_from_logits`) builds one inline, where the requests are still live.

## Testing

- `pytest tests/runner/test_routed_experts_slots.py` → **10 passed**.
- New `test_request_evicted_before_output_resolves` reproduces the reported crash: it snapshots while the request is live, pops it from `runner.requests` to simulate `_update_states`, and asserts the reconstruction still yields the correct slots. **Negative control:** reverting just that one line back to `runner.requests[req_id]` makes exactly this test fail with `KeyError: 'reqGone'`, the same failure mode as the report, while the other 9 pass.
- New `TestSnapshotBlockIds` covers the copy semantics and the absent-request case.
- `pre-commit run --files ...` clean on both files.
- Not yet re-run end to end on TPU with the two-request workload that produced the original crash.

## Notes

- **Interaction with #3583** (hybrid-model slot group). That PR makes the *group index* configurable, because on hybrid models the full-attention group is not group 0; this PR changes *where the block IDs come from*. They compose, and in fact compose well: once both land, the group selection belongs in `_snapshot_block_ids_for_routed_experts` (`block_ids[attn_gid]` instead of `block_ids[0]`), and the `kv_cache_group_id` parameter #3583 adds to `_reconstruct_slots_for_request` can be dropped. Whichever merges second should do that collapse — it's a one-line change at the snapshot site. This PR deliberately preserves main's current group-0 behaviour so the two fixes stay reviewable independently.
- **Adjacent, not fixed here:** the `continue_decode` path already tolerates a missing request (`requests_dict.get(req_id)` then `if req_state is not None`), so it doesn't crash — but it silently drops that request's routing rows instead. Worth a separate look.
- **Pre-existing, not changed here:** when a request's block list is empty, `_reconstruct_slots_for_request` returns all-zero slots, which makes `store_batch` write those rows into slot 0. That behaviour is unchanged by this PR.


wenxindongwork:
Folded into #3583 — the two fixes touch the same function and compose (the group selection now lives in the snapshot, so `_reconstruct_slots_for_request` takes plain `block_ids` with no group index). Closing in favour of that PR.

PR #3597, opened 2026-09-18
[Sampler] Unshard logits before argmax on full vocab path to avoid redundant AllReduce

prishajain1:
### Summary
In PR #3490, un-sharding of logits (`with_sharding_constraint(..., P(ATTN_DATA, None))`) was moved from the entry of `sample()` into the inner closure `sample_full_vocab(_)`.

As a result, whenever the full vocab sampling path is taken (distributed candidate sampling disabled or not fitting shard size):
1. `greedy_tokens = jnp.argmax(logits, axis=-1)` executes on **vocab-sharded logits**, triggering an unnecessary cross-chip **`AllReduce`**.
2. Immediately afterward, `sample_full_vocab(_)` executes an **`AllGather`** of the full logits.

This causes a performance regression compared to pre-#3490 by executing two sequential collective operations (`AllReduce` + `AllGather`) on every decode step instead of one.

### Fix
Restores the un-sharding constraint before `jnp.argmax` when `use_distributed_candidates` is `False` and `do_sampling` is `True`. This ensures:
- `greedy_tokens` is computed locally on each shard with **0 `AllReduce`**.
- `full_logits` is already gathered for sampling transforms, restoring the original single-`AllGather` performance.
- When `use_distributed_candidates` is `True`, logits remain sharded along the vocab axis as intended.


tpu_inference/layers/jax/sample/sampling.py
lk-chen, line 358: we still need this?

